PR #1, opened 2020-03-04
Review metadata models

axelboc-esrf:
- Add interfaces for HDF5 shape, type, attributes, etc.
- Rename interface used to type a mocked JSON metadata file from `HDF5Metadata` to `MockHDF5Metadata`, since this interface won't be used to type the data returned from HSDS.

### DIFF
--- a/src/demo-app/mock-data/models.ts
+++ b/src/demo-app/mock-data/models.ts
@@ -1,0 +1,18 @@
+import {
+  HDF5Id,
+  HDF5Collection,
+  HDF5Group,
+  HDF5Dataset,
+  HDF5Datatype,
+} from '../../h5web/providers/models';
+
+export interface MockHDF5Metadata {
+  id?: HDF5Id;
+  root: HDF5Id;
+  [HDF5Collection.Groups]: Record<HDF5Id, HDF5Group>;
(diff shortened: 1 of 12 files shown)


src/demo-app/mock-data/models.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  HDF5Id,
+  HDF5Collection,
+  HDF5Group,
+  HDF5Dataset,
+  HDF5Datatype,
+} from '../../h5web/providers/models';
+
+export interface MockHDF5Metadata {
```
axelboc-esrf: This interface is specific to mocked JSON data, so I move into the same folder and rename it.

src/h5web/App.tsx
```diff
@@ -18,11 +19,14 @@ function App(): JSX.Element {
         <ReflexElement className={styles.explorer} flex={0.3} minSize={250}>
           <Explorer
             filename="water_224.h5"
-            metadata={mockMetadata as HDF5Metadata}
+            metadata={mockMetadata as MockHDF5Metadata}
             onSelect={link => {
               setSelectedLink(link);
 
-              if (link.collection === HDF5Collection.Datasets) {
```
axelboc-esrf: A link can now be of multiple types, so I have to use a type guard `isHardLink` before assuming that `collection` is available.

PR #11, opened 2020-03-06
Support selecting domain in explorer

axelboc-esrf:
The domain (previously called the filename) is now part of the tree data structure that is rendered by the `Explorer` -- it's the root node of the tree and is associated with a fake `HDF5HardLink` object in order to be selectable (the `MetadataViewer` expects an `HDF5Link`).


src/h5web/explorer/Explorer.module.css
```diff
@@ -3,8 +3,35 @@
...
+
 .domain {
   margin: 0;
   padding: 1rem;
   font-weight: 600;
 }
+
+.domainBtn {
+  composes: btn domain;
+}
```
axelboc-esrf: I moved tree view styles to the explorer since I need access to the `btn` style in the `Explorer` component.

src/h5web/explorer/Explorer.tsx
```diff
@@ -18,25 +18,53 @@ function Explorer(props: Props): JSX.Element {
...
+  }, [tree]);
+
   useEffect(() => {
     if (selectedNode) {
+      // Propagate selected link to parent component
       onSelect(selectedNode.data);
     }
   }, [selectedNode, onSelect]);
 
+  if (!tree) {
```
axelboc-esrf: While the tree is loading (i.e. while the metadata is being fetched and the tree is being built), I render the domain and a loading text right away.

src/h5web/explorer/Explorer.tsx
```diff
@@ -18,25 +18,53 @@ function Explorer(props: Props): JSX.Element {
...
+        className={styles.domainBtn}
+        type="button"
+        role="treeitem"
+        aria-selected={selectedNode === tree}
+        onClick={() => {
+          setSelectedNode(tree);
+        }}
+      >
+        {domain}
+      </button>
```
axelboc-esrf: Even though the domain is part of the tree data structure, I decided not to render it within `TreeView` to avoid having to  deal with a root case in the recursion (`div` instead of `ul`, no expansion, no icon, etc.)

src/h5web/providers/silx/utils.ts
```diff
@@ -1,24 +1,49 @@
...
+export function buildTree(
+  metadata: SilxMetadata,
+  domain: string
+): TreeNode<HDF5Link> {
+  const rootLink: HDF5HardLink = {
+    class: HDF5LinkClass.Hard,
+    collection: HDF5Collection.Groups,
+    title: domain,
+    id: metadata.root,
+  };
```
axelboc-esrf: It's here that I create a fake `HDF5HardLink` for the root domain node.

loichuder:
What about adding an icon to the root node to emphasise the fact that it can be clickable ?

PR #13, opened 2020-03-10
Display link metadata

axelboc-esrf:
- Add component `LinkInfo` to display metadata for a `link`
- Add a new type to represent the root link, `HDF5RootLink`, with a custom class, `H5WEB_ROOT`, to make it clear that this is not a class of link defined by the HDF5 spec.
- Don't show the title and class info for the root link, as this doesn't make sense.

### DIFF
--- a/src/h5web/metadata-viewer/LinkInfo.tsx
+++ b/src/h5web/metadata-viewer/LinkInfo.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { HDF5Link, HDF5LinkClass } from '../providers/models';
+
+import styles from './MetadataViewer.module.css';
+
+interface Props {
+  link: HDF5Link;
+}
+
+function LinkInfo(props: Props): JSX.Element {
+  const { link } = props;
+
+  return (
+    <>
+      <tr>
+        <th className={styles.table_head} colSpan={2}>
+          Link info
+        </th>
+      </tr>
+      {link.class !== HDF5LinkClass.Root && (
+        <>
+          <tr>
+            <th scope="row">Title</th>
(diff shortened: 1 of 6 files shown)


src/h5web/metadata-viewer/LinkInfo.tsx
```diff
@@ -0,0 +1,56 @@
...
+  link: HDF5Link;
+}
+
+function LinkInfo(props: Props): JSX.Element {
+  const { link } = props;
+
+  return (
+    <>
+      <tr>
+        <th colSpan={2}>Link info</th>
```
loichuder: I would add the styles.table_head here for consistency with the other table heads.

src/h5web/providers/silx/utils.ts
```diff
@@ -34,14 +34,11 @@ function buildTreeNode(
...
-  domain: string
-): TreeNode<HDF5Link> {
-  const rootLink: HDF5HardLink = {
-    class: HDF5LinkClass.Hard,
+export function buildTree(metadata: SilxMetadata): TreeNode<HDF5Link> {
+  const rootLink: HDF5RootLink = {
+    class: HDF5LinkClass.Root,
     collection: HDF5Collection.Groups,
-    title: domain,
+    title: '',
```
loichuder: I understand that the root is a special case but I thought that having the domain as title was making sense.

Can we have `'/'` for the root node instead of the empty string ?

axelboc-esrf: As discussed, the title isn't used at the moment and putting something else than an empty string may make it more difficult to build a path for each node during the construction of the tree.

PR #16, opened 2020-03-11
Reorganise metadata viewer

axelboc-esrf:
- Normalize child components : `LinkInfo`, `EntityInfo`, `AttributesInfo`, `RawInfo`
- Show entity path(s) (i.e. = `alias`)


src/h5web/metadata-viewer/AttributesInfo.tsx
```diff
@@ -16,22 +16,20 @@ function AttributesRenderer(props: Props): JSX.Element {
   return (
     <>
       <tr>
-        <th className={styles.table_head} colSpan={2}>
+        <th className={styles.headingCell} colSpan={2}>
```
loichuder: Sorry, Python snakecase habits !

axelboc-esrf:
> From what I get, `alias` is the path inside the HDF5 file. What about renaming it `hdfPath` or something along these lines ?

Does it get returned by HSDS at all, or is it h5json specific?

axelboc-esrf:
Ok, I removed `alias` from the models for now, since it's not returned by HSDS. We'll have to think of a better way to show the path of each entity.

PR #20, opened 2020-03-13
Pass dataset instead of link to DatasetVisualizer

axelboc-esrf:
I think it makes more sense to pass a valid dataset to the visualizer rather than a link. In `App.tsx`, `selectedDataset` is now of type `HDF5Entity<HDF5Collection.Datasets>`, which is more logicial and means we don't have to re-check that the corresponding entity is indeed a dataset inside `DatasetVisualizer`.

### DIFF
--- a/src/h5web/App.tsx
+++ b/src/h5web/App.tsx
@@ -1,33 +1,33 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { ReflexContainer, ReflexSplitter, ReflexElement } from 'react-reflex';
 
 import Explorer from './explorer/Explorer';
 import DatasetVisualizer from './dataset-visualizer/DatasetVisualizer';
-import { HDF5Link, HDF5Collection, HDF5HardLink } from './providers/models';
+import { HDF5Link, HDF5Collection, HDF5Entity } from './providers/models';
 import MetadataViewer from './metadata-viewer/MetadataViewer';
 import styles from './App.module.css';
-import { isHardLink } from './providers/type-guards';
+import { isDataset } from './providers/type-guards';
+import { useEntity } from './providers/hooks';
 
 function App(): JSX.Element {
   const [selectedLink, setSelectedLink] = useState<HDF5Link>();
-  const [selectedDataset, setSelectedDataset] = useState<HDF5HardLink>();
+  const [selectedDataset, setSelectedDataset] = useState<
+    HDF5Entity<HDF5Collection.Datasets>
+  >();
+
+  const selectedEntity = useEntity(selectedLink);
+
+  useEffect(() => {
+    if (selectedEntity && isDataset(selectedEntity)) {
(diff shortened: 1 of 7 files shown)


src/h5web/App.tsx
```diff
@@ -1,33 +1,33 @@
...
+import { useEntity } from './providers/hooks';
 
 function App(): JSX.Element {
   const [selectedLink, setSelectedLink] = useState<HDF5Link>();
-  const [selectedDataset, setSelectedDataset] = useState<HDF5HardLink>();
+  const [selectedDataset, setSelectedDataset] = useState<
+    HDF5Entity<HDF5Collection.Datasets>
+  >();
+
+  const selectedEntity = useEntity(selectedLink);
```
axelboc-esrf: The entity is retrieved here, in `App.tsx`, and passed down to the child components.

src/h5web/App.tsx
```diff
@@ -1,33 +1,33 @@
...
   return (
     <div className={styles.app}>
       <ReflexContainer orientation="vertical" windowResizeAware>
         <ReflexElement className={styles.explorer} flex={0.3} minSize={250}>
-          <Explorer
-            onSelect={link => {
-              setSelectedLink(link);
-
-              if (
-                isHardLink(link) &&
```
axelboc-esrf: The hard link check is now performed inside `useEntity` => it returns undefined if the link is not a hard link, which is then checked inside `useEffect`.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -1,25 +1,22 @@
...
 }
 
 function DatasetVisualizer(props: Props): JSX.Element {
-  const { link } = props;
-  const { title } = link;
+  const { dataset } = props;
+  const { id } = dataset;
 
-  const value = useValue(link);
+  const value = useValue(id);
```
axelboc-esrf: `useValue` no longer relies on the link but on the entity's ID.

src/h5web/providers/context.ts
```diff
@@ -1,12 +1,12 @@
...
+import { HDF5Link, HDF5GenericEntity, HDF5Id } from './models';
 import { TreeNode } from '../explorer/models';
 
 interface DataProvider {
   getDomain: () => string;
   getMetadataTree: () => Promise<TreeNode<HDF5Link>>;
-  getEntity: (link: HDF5Link) => Promise<HDF5GenericEntity | undefined>;
-  getValue: (link: HDF5HardLink) => Promise<any>; // eslint-disable-line @typescript-eslint/no-explicit-any
+  getEntity: (link?: HDF5Link) => Promise<HDF5GenericEntity | undefined>;
+  getValue: (id: HDF5Id) => Promise<any>; // eslint-disable-line @typescript-eslint/no-explicit-any
```
axelboc-esrf: Note these two minor signature changes in the `DataProvider` interface.

PR #28, opened 2020-03-18
Implemented index display in MatrixVis

loichuder:


src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +14,162 @@ interface Props {
...
+        {index >= 0 && index}
+      </div>
+    );
+  }
+
+  function IndexColumn(): JSX.Element {
+    const { columnWidth, rowHeight, rowCount } = useContext(GridSettings);
+    return (
+      <div
+        className={`${styles.sticky} ${styles.indexColumn}`}
```
loichuder: I tried to separate the `index` CSS from the `sticky` CSS. Perhaps can we implement a toggle sticky ?

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +14,162 @@ interface Props {
...
+type CellRendererFunc = (props: GridChildComponentProps) => JSX.Element;
+
 function MatrixVis(props: Props): JSX.Element {
   const { dims, data } = props;
+  const gridDimensions = {
+    rowHeight: 32,
+    rowCount: dims[0] + 1,
+    columnWidth: 116,
+    columnCount: (dims.length === 2 ? dims[1] : 1) + 1,
+  };
```
loichuder: I am conflicted here. I choose to add 1 to `rowCount` and `columnCount` because the final grid has 1 more row and 1 more column. But it may also be possible to have `rowCount` and `columnCount` to be equal to the exact number of rows and columns of the *true* data and move the `+1` in the `Index` components.

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +14,162 @@ interface Props {
...
+  }
+
+  function IndexColumn(): JSX.Element {
+    const { columnWidth, rowHeight, rowCount } = useContext(GridSettings);
+    return (
+      <div
+        className={`${styles.sticky} ${styles.indexColumn}`}
+        style={{ width: columnWidth }}
+      >
+        {range(rowCount - 1).map(row => {
```
loichuder: As `lodash` is already used somewhere else in the code, I allowed myself to use the `range` method from `lodash`

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +14,162 @@ interface Props {
...
+
+  const innerElementType = forwardRef(({ children, ...rest }, ref) => {
+    return (
+      <div ref={ref as Ref<HTMLDivElement>} {...rest}>
+        <IndexRow />
+        <IndexColumn />
+        {children}
+      </div>
+    );
+  });
```
loichuder: I wonder if all functions that are only relevant to `IndexedGrid` should be moved inside this function.

src/h5web/dataset-visualizer/vis/MatrixVis.module.css
```diff
@@ -13,3 +13,33 @@
...
+  display: flex;
+  font-weight: bold;
+  align-items: center;
+  justify-content: center;
+  background-color: var(--primary-bg);
+}
+
+.indexContainer {
+  display: flex;
+  z-index: 2;
```
axelboc-esrf: Would `1` work?

src/h5web/dataset-visualizer/vis/MatrixVis.module.css
```diff
@@ -13,3 +13,37 @@
 .cell[data-bg] {
   background-color: var(--secondary-light-bg);
 }
+
+.indexCell {
+  display: flex;
+  font-weight: bold;
+  align-items: center;
+  justify-content: center;
+  background-color: var(--primary-bg);
```
axelboc-esrf: Just some visual suggestions, take it or leave it:

- `font-weight: 600`
- alternating the background colour like what we do for data cells (this tool might help finding a lighter or darker colour: https://calcolor.co/blend/f5fbef/ffffff)
- making the background colour very slightly transparent with `rgba`

src/h5web/dataset-visualizer/vis/MatrixVis.module.css
axelboc-esrf: Actually... When you scroll horizontally, it looks like `indexRow` goes above `indexColumn`, but when you scroll vertically it's actually `indexColumn` that goes above `indexRow`. This inconsistent behaviour looks a little strange.

It comes from the fact that:
- the empty index cell in the top left of the grid belongs to `indexRow`,
- `indexColumn` comes after `indexRow` in the DOM, so its the stacking order is higher than `indexRow`'s.

I think the simpler option to fix this would be for `indexRow` to have `z-index: 2` and `indexColumn` to have `z-index: 1`. This means that the row will always remain above the column. If the opposite behaviour is preferred, then an empty cell would have to be added to `indexColumn` in order to cover the row when scrolling horizontally (the column would also have to be negatively offset).

src/h5web/dataset-visualizer/vis/MatrixVis.module.css
```diff
@@ -13,3 +13,37 @@
...
+}
+
+.indexColumn {
+  composes: indexContainer;
+  flex-direction: column;
+}
+
+.indexRow {
+  composes: indexContainer;
+  flex-direction: row;
```
axelboc-esrf: pas besoin normalement (c'est la valeur par défaut)

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +21,179 @@ interface Props {
...
+      <div className={styles.indexCell} style={style}>
+        {index >= 0 ? (
+          index
+        ) : (
+          <button
+            className={styles.btn}
+            type="button"
+            onClick={() => {
+              setSticky(!sticky);
+            }}
```
axelboc-esrf: Ha, fun! The button needs to fill the whole cell, I think. Also, it could use a hover effect of some sort.

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
axelboc-esrf: Hmm, yeah tough one. I think it's easier to keep it here, since `react-window` does need to calculate the total width and height of the grid, index cells included.

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
axelboc-esrf: Sounds good, the alternative is ugly... There's `Array(length).fill(0)` now, but then you have to add the `index` argument to the arrow function in the `map`...

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +21,179 @@ interface Props {
...
+    return (
+      <div
+        className={`${styles.indexColumn} ${stickyStyle}`}
+        style={{ width: columnWidth }}
+      >
+        {range(rowCount - 1).map(row => {
+          return (
+            <IndexCell
+              index={row}
+              key={`${row}.0`}
```
axelboc-esrf: not sure I get the `.0`

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +21,179 @@ interface Props {
...
+                height: rowHeight,
+              }}
+            />
+          );
+        })}
+      </div>
+    );
+  }
+
+  const innerElementType = forwardRef(({ children, ...rest }, ref) => {
```
axelboc-esrf: You can give this a better name, and with a capital letter since it's a component. That's the convention they use [in the docs](https://reactjs.org/docs/forwarding-refs.html).

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +21,179 @@ interface Props {
...
+  const [sticky, setSticky] = useState(true);
+  const gridDimensions = {
+    rowHeight: 32,
+    rowCount: dims[0] + 1,
+    columnWidth: 116,
+    columnCount: (dims.length === 2 ? dims[1] : 1) + 1,
+  };
+  const GridSettings = createContext(gridDimensions);
+
+  const CellRenderer: CellRendererFunc = (
```
axelboc-esrf: I'd prefer to keep the `function` declaration syntax for readability and consistency: `function CellRenderer(...): ... { ... }` even if it means you can't reuse type `CellRendererFunc` (you already type the argument anyway).

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
axelboc-esrf: The sticky state is dynamic now, so there are a couple of ways to implement this a bit better than by conditionally concatenating classes:

- using a lib like `classnames`;
- adding a ternary and two classes: `indexColumn` and `stickyIndexColumn`, and composing `sticky` from the latter in CSS;
- moving the sticky state to a `data` attribute: `data-sticky={sticky || undefined}` and using an attribute selector: `.indexColumn[data-attribute]`.

The last solution is the cleanest in my opinion. The only downside it that it increases the specificity of the selector, which could be bad in some cases.

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
loichuder: Left over from the time where all index cells were in the same container. Fixed.

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +21,179 @@ interface Props {
...
+      </div>
+    );
+  }
+
+  function IndexCell(indexCellProps: IndexCellProps): JSX.Element {
+    const { index, style } = indexCellProps;
+    return (
+      <div className={styles.indexCell} style={style}>
+        {index >= 0 ? (
+          index
```
axelboc-esrf: If you remove the anchor button logic from here and introduce a generic component for `IndexRow` and `IndexColumn` as suggested in other comments, I don't think you'd need to keep `IndexCell` as a separate component.

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +21,179 @@ interface Props {
...
+        columnIndex={columnIndex - 1}
+        style={style}
+      />
+    );
+  };
 
   const accessor: AccessorFunc =
     dims.length === 1 ? row => data[row] : (row, col) => data[row][col];
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
axelboc-esrf: no longer required

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
```diff
@@ -13,34 +21,179 @@ interface Props {
...
+  // IndexedGrid based on https://codesandbox.io/s/0mk3qwpl4l
+  function IndexedGrid(matrixProps: MatrixProps): JSX.Element {
+    const { children, width, height } = matrixProps;
+    const { columnWidth, rowHeight, columnCount, rowCount } = useContext(
+      GridSettings
+    );
+
+    return (
+      <Grid
+        itemData={{ NormalCellFunc: children }}
```
axelboc-esrf: If you inline the `CellRenderer` function here, directly inside `Grid` like this: `<Grid>{(...) => { ... }}</Grid>, perhaps you won't have to pass `NormalCellFunc` as `itemData` and then retrieve it in the cell renderer?

src/h5web/dataset-visualizer/vis/MatrixVis.module.css
loichuder: First two implemented. I find that having a transparent background makes the numbers appear cluttered...

src/h5web/dataset-visualizer/vis/MatrixVis.module.css
loichuder: Looks good :+1:

src/h5web/dataset-visualizer/vis/MatrixVis.module.css
axelboc-esrf: Yeah, I wasn't entirely convinced either.

src/h5web/dataset-visualizer/vis/MatrixVis.module.css
axelboc-esrf: Yes, much better!

I actually had an idea during the demo: what if the anchor cell always stayed in the corner when `sticky` is enabled? Both index bands would then scoll under it. I think it may look even better!

You could render the anchor cell inside the `InnerGrid` component directly instead of inside the index row. (The index row would then have to be aligned to the right with `justify-content: flex-end` to account for the missing cell). The anchor cell would be positioned with `position: sticky`, `top: 0` and `bottom: 0` relative to the inner grid `div` (although `position: fixed` would work as well).

What do you reckon?

loichuder: I thought about it as well but was unsure on how to do it. Your proposal looks sensible, I will try to implement that !

src/h5web/dataset-visualizer/vis/MatrixVis.tsx
loichuder: See https://github.com/silx-kit/h5web/pull/28#discussion_r397061137

src/h5web/dataset-visualizer/vis/matrix/AnchorCell.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useContext } from 'react';
+import { FiAnchor } from 'react-icons/fi';
+import { StickyContext } from './StickyContext';
+import styles from './MatrixVis.module.css';
+
+interface Props {
+  style: React.CSSProperties;
+}
+
+function AnchorCell(props: Props): JSX.Element {
```
loichuder: Naming issue here: I find that the name of the component does not describe its functionality. Two solutions for me:
- Rename it `StickyToggler` or something similar
- Rename all `stickyIndices` to `anchorIndices` for consistency

src/h5web/dataset-visualizer/vis/matrix/IndexedGrid.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+// IndexedGrid based on https://codesandbox.io/s/0mk3qwpl4l
+// Wrapper around FixedSizeGrid to add indices row and column (sticky or not).
+function IndexedGrid(props: Props): JSX.Element {
+  const {
+    children: NormalCellFunc,
+    width,
+    height,
+    gridSettings,
+    itemData,
```
loichuder: Here, I use `itemData` to pass the accessor function from the `Matrix` component down to the `ValueCell` component. `itemData` is a  `react-window` props described as a light-weight alternative to React's `Context`. But I wonder if using a `Context` would be easier here.

src/h5web/dataset-visualizer/vis/matrix/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface CellSize {
+  width: number;
+  height: number;
+}
+
+export interface GridSettings {
+  cellSize: CellSize;
+  rowCount: number;
+  columnCount: number;
+}
```
loichuder: I put additional types in a `util.ts` but I wonder if this is worthwhile as there are few.

src/h5web/dataset-visualizer/vis/matrix/StickyContext.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const { stickyIndices, toggleStickyIndices, children } = props;
+
+  return (
+    <StickyContext.Provider value={{ stickyIndices, toggleStickyIndices }}>
+      {children}
+    </StickyContext.Provider>
+  );
+}
+
+export default StickyProvider;
```
loichuder: @axelboc-esrf Please check if this is the implementation of `Context` and its use with `State` in `IndexedGrid` are correct. Thanks !

src/h5web/dataset-visualizer/vis/matrix/MatrixVis.tsx
```diff
@@ -0,0 +1,40 @@
...
+  data: HDF5Value;
+}
+
+function MatrixVis(props: Props): JSX.Element {
+  const { dims, data } = props;
+  const myGridSettings = {
+    cellSize: { width: 116, height: 32 },
+    rowCount: dims[0] + 1, // includes IndexRow
+    columnCount: (dims.length === 2 ? dims[1] : 1) + 1, // includes IndexColumn
+  };
```
loichuder: I reverted my use of `Context` for grid settings. I feel that passing it as props to the `Grid` that needs it and then, only the `cellSize` to the components needing it, is clearer.

src/h5web/dataset-visualizer/vis/matrix/IndexedGrid.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const [stickyState, setStickyState] = useState(true);
+  function toggleStickyIndices(): void {
+    setStickyState(!stickyState);
+  }
+
+  const InnerGrid = forwardRef(({ children: innerChildren, ...rest }, ref) => {
+    return (
+      <div ref={ref as Ref<HTMLDivElement>} {...rest}>
+        <IndexTrack
+          anchorCell
```
loichuder: The anchorCell boolean will disappear following our discussion (https://github.com/silx-kit/h5web/pull/28#discussion_r397267734)

src/h5web/dataset-visualizer/vis/matrix/IndexedGrid.tsx
axelboc-esrf: Yeah that means they just pass down the prop themselves throught their internal components. Putting it in the context would be more consistent and less obscure.

src/h5web/dataset-visualizer/vis/matrix/IndexedGrid.tsx
```diff
@@ -0,0 +1,91 @@
+import React, { useState, forwardRef, Ref } from 'react';
+import { FixedSizeGrid as Grid, GridChildComponentProps } from 'react-window';
+import StickyProvider from './StickyContext';
+import styles from './MatrixVis.module.css';
+import AnchorCell from './AnchorCell';
+import IndexTrack from './IndexTrack';
```
axelboc-esrf: Wow, inline lint warnings! Cool stuff, GitHub 👏

src/h5web/dataset-visualizer/vis/matrix/IndexedGrid.tsx
axelboc-esrf: Ignore this comment, don't put it into a context, there's no need. You don't even need to pass `itemData` to react-window's `Grid`, since you now declare the cell renderer is no longer a separate component. You can use the prop directly in the renderer's code:

```
<Grid ...>
  {
    ...
    return <NormalCellFunc data="{itemData} ...>...</NormalCellFunc>
  }
</Grid>
```

src/h5web/dataset-visualizer/vis/matrix/AnchorCell.tsx
loichuder: Finally kept like that to to keep the distinction between the button and the code handling the stickiness.

src/h5web/dataset-visualizer/vis/matrix/utils.ts
loichuder: `util.ts` removed

src/h5web/dataset-visualizer/vis/matrix/MatrixVis.tsx
loichuder: Put again in `Context` as it is not possible to pass props to `innerElementType`

src/h5web/dataset-visualizer/vis/matrix/IndexedGrid.tsx
loichuder: `AnchorCell` is now a children of `IndexTrack`. The boolean was removed.

PR #30, opened 2020-03-19
Implements the display of attributes values in HsdsProvider

loichuder:


src/h5web/providers/hsds/api.ts
```diff
@@ -112,13 +125,27 @@ export class HsdsApi implements ProviderAPI {
...
+    const { attributes: bareAttributes } = await this.fetchAttributes(
+      collection,
+      id
+    );
+    const valuePromises = bareAttributes.map(attr =>
+      this.fetchAttributeWithValue(collection, id, attr.name)
+    );
+    return Promise.all(valuePromises);
+  }
+
```
loichuder: I put the fetch in the `processAttributes` method to be consistent with other `process...` methods.

src/h5web/providers/hsds/models.ts
```diff
@@ -26,9 +26,11 @@ export interface HsdsDatatypeResponse {
 }
 
 export interface HsdsAttributesResponse {
-  attributes: HDF5Attribute[];
+  attributes: HsdsAttributesList;
 }
 
+export type HsdsAttributesList = Omit<HDF5Attribute, 'value'>[];
+
```
loichuder: The introduction of this new type is perhaps not necessary and/or badly named.

src/h5web/providers/hsds/api.ts
```diff
@@ -112,13 +125,27 @@ export class HsdsApi implements ProviderAPI {
 
   /* Processing methods to fetch links and attributes in addition to the entity. Also updates API members. */
 
+  private async processAttributes(
+    collection: HDF5Collection,
+    id: HDF5Id
```
axelboc: `entityId` just for clarity?

src/h5web/providers/hsds/api.ts
axelboc: Yeah, I think it makes sense!

src/h5web/providers/hsds/api.ts
```diff
@@ -102,6 +103,18 @@ export class HsdsApi implements ProviderAPI {
     return data;
   }
 
+  private async fetchAttributeWithValue(
+    collection: HDF5Collection,
+    id: HDF5Id,
+    name: string
+  ): Promise<HDF5Attribute> {
+    const { data } = await this.client.get<HDF5Attribute>(
```
axelboc: Met un petit alias `type HsdsAttributeWithValueResponse = HDF5Attribute`

src/h5web/providers/hsds/api.ts
loichuder: Good catch :fishing_pole_and_fish:

src/h5web/providers/hsds/api.ts
```diff
@@ -112,13 +125,27 @@ export class HsdsApi implements ProviderAPI {
 
   /* Processing methods to fetch links and attributes in addition to the entity. Also updates API members. */
 
+  private async processAttributes(
+    collection: HDF5Collection,
+    id: HDF5Id
+  ): Promise<HDF5Attribute[]> {
+    const { attributes: bareAttributes } = await this.fetchAttributes(
```
axelboc: Perhaps we should merge my PR first, I did a small refactoring so that the fetch methods return directly the data that is relevant (in this case, `data.attributes`, instead of just `data`)

src/h5web/providers/hsds/models.ts
axelboc: Pas besoin d'un sous-type du tout ;)

src/h5web/providers/hsds/api.ts
```diff
@@ -94,14 +95,29 @@ export class HsdsApi implements ProviderAPI {
...
+    );
+    return Promise.all(valuePromises);
+  }
+
+  private async fetchAttributeWithValue(
+    entityCollection: HDF5Collection,
+    entityId: HDF5Id,
+    attrName: string
+  ): Promise<HDF5Attribute> {
+    const { data } = await this.client.get<HsdsAttributeWithValueResponse>(
```
axelboc: Je pense que le `get` (sans `await`) peut directement aller dans la méthode `fetchAttributes`

src/h5web/providers/hsds/api.ts
loichuder: Finally, the complete fetch (attributes + values) was put in `fetchAttributes`

src/h5web/providers/hsds/api.ts
loichuder: Done.

PR #33, opened 2020-03-20
Implement Scalar visualization

axelboc:
The Raw visualization was showing double quotes around scalar values, and backslashes in stringified JSON values. This is much more readable:

image

image

### Before with Raw Vis

image

### DIFF
--- a/src/h5web/dataset-visualizer/VisSelector.tsx
+++ b/src/h5web/dataset-visualizer/VisSelector.tsx
@@ -1,11 +1,12 @@
 import React from 'react';
-import { FiCpu, FiGrid } from 'react-icons/fi';
+import { FiCpu, FiGrid, FiCode } from 'react-icons/fi';
 import { IconType } from 'react-icons';
 import { Vis } from './models';
 import styles from './VisSelector.module.css';
 
 const VIS_ICONS: Record<Vis, IconType> = {
   [Vis.Raw]: FiCpu,
+  [Vis.Scalar]: FiCode,
   [Vis.Matrix]: FiGrid,
 };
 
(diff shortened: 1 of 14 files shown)


src/h5web/dataset-visualizer/VisSelector.tsx
```diff
@@ -1,11 +1,12 @@
 import React from 'react';
-import { FiCpu, FiGrid } from 'react-icons/fi';
+import { FiCpu, FiGrid, FiDisc } from 'react-icons/fi';
 import { IconType } from 'react-icons';
 import { Vis } from './models';
 import styles from './VisSelector.module.css';
 
 const VIS_ICONS: Record<Vis, IconType> = {
   [Vis.Raw]: FiCpu,
+  [Vis.Scalar]: FiDisc,
```
loichuder: I would put `FiCircle` instead of `FiDisc` here

src/h5web/providers/utils.ts
```diff
@@ -36,6 +37,10 @@ export function isSimpleShape(shape: HDF5Shape): shape is HDF5SimpleShape {
   return shape.class === HDF5ShapeClass.Simple;
 }
 
+export function isScalarShape(shape: HDF5Shape): shape is HDF5ScalarShape {
+  return shape.class === HDF5ShapeClass.Scalar;
+}
+
```
loichuder: The naming pains me a bit here. It is not the shape that is Scalar but rather the value it describes. Anyway, I think this is still the best solution to be consistent with `ShapeClass` and the rest of the code.

src/h5web/dataset-visualizer/VisSelector.tsx
axelboc-esrf: Changed to `FiCode`

PR #34, opened 2020-03-23
Implement line vis 🎉✨🍻 

axelboc-esrf:
image

Overall, it was quite straightforward to get a simple line graph up and running with [**Recharts**](http://recharts.org/en-US/examples/SimpleLineChart).

A few observations:

- I had some issues with their `ResponsiveContainer` so I used `AutoSizer` from `react-virtualized` instead. It does the same thing (i.e. provide a width and height to a sub-component).
- The input data array (a 1D array of numbers) has to be transformed into an array of objects representing each datapoint. I have the feeling this is the case with other charting libraries as well.
- There's an example of a [zoomable graph](http://recharts.org/en-US/examples/HighlightAndZoomLineChart) in the docs, which looks relatively straightforward.
- With 100,000 points, rendering takes a bit more than a second and the app doesn't feel sluggish afterwards, which is good news. However, the hover interaction (popup with datapoint information) struggles a bit; the popup is very "jumpy".
- Reacharts' documentation is lacking. The components seem to have quite advanced APIs, but a lot of props are poorly documented or not documented at all.

### DIFF
--- a/src/h5web/dataset-visualizer/vis/LineVis.tsx
+++ b/src/h5web/dataset-visualizer/vis/LineVis.tsx
@@ -1,0 +1,46 @@
+import React, { useMemo } from 'react';
+import {
+  LineChart,
+  Line,
+  XAxis,
+  YAxis,
+  Tooltip,
+  CartesianGrid,
+} from 'recharts';
+import AutoSizer from 'react-virtualized-auto-sizer';
+
+import styles from './LineVis.module.css';
+
+interface Props {
+  data: number[];
+}
+
+function LineVis(props: Props): JSX.Element {
+  const { data } = props;
+
+  const chartData = useMemo(() => {
+    return data.map((val, index) => ({ x: index + 1, y: val }));
+  }, [data]);
+
+  return (
+    <AutoSizer>
+      {({ width, height }) => (
+        <LineChart
+          className={styles.chart}
+          data={chartData}
+          width={width}
+          height={height}
+          margin={{ top: 32, right: 32, bottom: 16 }}
+        >
+          <CartesianGrid strokeDasharray="3 3" />
+          <XAxis type="number" dataKey="x" />
+          <YAxis />
+          <Tooltip />
+          <Line dataKey="y" dot={false} isAnimationActive={false} />
+        </LineChart>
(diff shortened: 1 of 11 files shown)


src/h5web/providers/utils.ts
```diff
@@ -55,6 +56,10 @@ export function isBaseType(type: HDF5Type): type is HDF5BaseType {
   );
 }
 
+export function isNumericType(type: HDF5Type): type is HDF5NumericType {
+  return isBaseType(type) && type.class !== HDF5TypeClass.String;
```
loichuder: I would change the typeguard into `return type.class === HDF5TypeClass.Integer || type.class === HDF5TypeClass.Float` for consistency with `HDF5NumericType` and in case `BaseType` needs to be extended.

src/h5web/dataset-visualizer/vis/LineVis.tsx
```diff
@@ -0,0 +1,46 @@
...
+          height={height}
+          margin={{ top: 32, right: 32, bottom: 16 }}
+        >
+          <CartesianGrid strokeDasharray="3 3" />
+          <XAxis type="number" dataKey="x" />
+          <YAxis />
+          <Tooltip />
+          <Line dataKey="y" dot={false} isAnimationActive={false} />
+        </LineChart>
+      )}
```
loichuder: It looks indeed quite straightforward to set up !

src/h5web/providers/utils.ts
axelboc-esrf: Ah yeah I think I did that because `isBaseType` also makes sure that `type` is not a `string`. Perhaps I can change `isBaseType` to use `isNumericType`, instead? It would make more sense this way.

axelboc-esrf: Neh, doesn't work, I'll just duplicate the logic.

loichuder: I am not sure. My opinion is that it *happens* that `NumericType`is `BaseType` without `String` but it depends on what is contained in `BaseType`.

I feel that putting that `Numeric` is either `Integer` or `Float` is clearer.

PR #37, opened 2020-03-26
Move `vis` folder up one level and rename

axelboc-esrf:


src/h5web/visualizations/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default as RawVis } from './RawVis';
+export { default as MatrixVis } from './matrix/MatrixVis';
+export { default as ScalarVis } from './ScalarVis';
+export { default as LineVis } from './LineVis';
```
axelboc-esrf: I added an index file to simplify importing the visualizations in `VisDisplay`.

loichuder: Good foresight :+1:

PR #39, opened 2020-03-27
Added EmptyVisualiser to be displayed when not selecting a dataset

loichuder:


loichuder:
Done with another implementation in #40

PR #41, opened 2020-04-02
Implement Heatmap visualization

axelboc:
- Use Three.JS via react-three-fiber
- Render mesh with place geometry and basic material.
- Compute texture data in JS with `d3-scale` and map the texture to the basic material.
- Use `interpolateMagma` from `d3-scale-chromatic` for color map.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,9 @@
   "dependencies": {
     "axios": "^0.19.2",
     "classnames": "^2.2.6",
+    "d3-color": "^1.4.0",
+    "d3-scale": "^3.2.1",
+    "d3-scale-chromatic": "^1.5.0",
     "lodash-es": "^4.17.15",
     "nanoid": "^2.1.11",
     "normalize.css": "^8.0.1",
@@ -31,15 +34,20 @@
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-icons": "^3.9.0",
+    "react-three-fiber": "^4.0.23",
     "react-use": "^13.26.3",
     "react-virtualized-auto-sizer": "^1.0.2",
     "react-window": "^1.8.5",
-    "recharts": "^1.8.5"
+    "recharts": "^1.8.5",
+    "three": "^0.115.0"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.4.0",
     "@testing-library/user-event": "^7.2.1",
+    "@types/d3-color": "^1.2.2",
+    "@types/d3-scale": "^2.2.0",
+    "@types/d3-scale-chromatic": "^1.3.1",
     "@types/jest": "^24.9.1",
     "@types/lodash-es": "^4.17.3",
     "@types/nanoid": "^2.1.0",
--- a/src/h5web/visualizations/heatmap/utils.ts
+++ b/src/h5web/visualizations/heatmap/utils.ts
@@ -1,0 +1,27 @@
+import { rgb } from 'd3-color';
+import { scaleSequential } from 'd3-scale';
+import { interpolateMagma } from 'd3-scale-chromatic';
+
+function findBounds(values: number[]): [number, number] {
+  return values.reduce(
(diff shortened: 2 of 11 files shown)


.vscode/extensions.json
```diff
@@ -3,6 +3,7 @@
     "editorconfig.editorconfig",
     "esbenp.prettier-vscode",
     "dbaeumer.vscode-eslint",
-    "firefox-devtools.vscode-firefox-debug"
+    "firefox-devtools.vscode-firefox-debug",
+    "slevesque.shader"
```
axelboc, line 7: oops, plus de shader, donc je retire

package.json
```diff
@@ -52,9 +60,11 @@
     "@types/recharts": "^1.8.9",
     "@wsmd/eslint-config": "^1.2.0",
     "eslint": "^6.8.0",
+    "glslify-loader": "^2.0.0",
     "node-sass": "^4.13.1",
     "npm-run-all": "^4.1.5",
     "prettier": "^1.19.1",
+    "raw-loader": "^4.0.0",
```
axelboc, line 67: je retire aussi ces loaders pour le moment

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -0,0 +1,28 @@
+import { rgb } from 'd3-color';
+import { scaleSequential } from 'd3-scale';
+import { interpolateMagma } from 'd3-scale-chromatic';
+
+export function computeTextureData(matrix: number[][]): Uint8Array {
+  // Flatten input matrix
+  const values = matrix.flat();
+
+  // Compute domain
+  const domain: [number, number] = values.reduce(
```
loichuder, line 10: There is no dedicated function to get the min/max of an array ?

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  // Compute RGB color for each datapoint `[[<r>, <g>, <b>], [<r>, <g>, <b>], ...]`
+  const colors = values.map(val => {
+    // Invoke color map and parse the CSS string it returns (`rgb(r, g, b)`)
+    const { r, g, b } = rgb(colorMap(val));
+    return [r, g, b];
+  });
+
+  // Prepare colors array for use as texture data
+  return Uint8Array.from(colors.flat());
+}
```
loichuder, line 28: Some comments are not really useful and can be removed.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
```
loichuder, line 3: What prompted this ?

axelboc: https://threejs.org/docs/index.html#manual/en/introduction/Typescript-setup 😃

src/h5web/visualizations/heatmap/utils.ts
axelboc: Nope, but perhaps in lodash, I'll have a look.

loichuder: There is indeed [_.min](https://lodash.com/docs/4.17.15#min) and [_.max](https://lodash.com/docs/4.17.15#max) but that means traversing the array twice to get both so I think your implementation is better.

axelboc: There's `min` and `max` in lodash, but not both together. We'd get O(2n) instead of O(n). Not awful, but significant nonetheless. I found two libraries that could work instead, notably https://github.com/dy/array-bounds, but neither of them provides type declarations. 😞 

I opened an issue on the lodash repository; I thought it was something worth suggesting: https://github.com/lodash/lodash/issues/4711

src/h5web/visualizations/heatmap/utils.ts
axelboc: I reviewed the comments and extracted the min/max algorithm into its own function for self-documentation purposes.

PR #56, opened 2020-04-08
Keep mesh within bounds when panning

axelboc:


src/h5web/visualizations/heatmap/Mesh.tsx
```diff
@@ -10,12 +10,12 @@ interface Props {
...
 function Mesh(props: Props): JSX.Element {
   const { dims, textureData } = props;
-  const [rows, cols] = dims;
 
-  const { camera, size } = useThree();
-  const { width, height } = size;
+  const { size } = useThree();
+  const pointerHandlers = usePanZoom();
 
-  const pointerHandlers = usePanZoom(camera as OrthographicCamera);
```
axelboc, line 18: I retrieve the camera within the hook instead of passing it as an argument.

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -29,9 +30,29 @@ export function computeTextureData(matrix: number[][]): Uint8Array | undefined {
...
+  const { size, camera } = useThree();
+  const { width, height } = size;
+
   const startOffsetPosition = useRef<Vector3>(); // `useRef` to avoid re-renders
 
+  const moveCamera = useCallback(
+    (x: number, y: number) => {
+      const { position, zoom } = camera;
+
+      const factor = (1 - 1 / zoom) / 2;
```
axelboc, line 43: Le facteur magique ✨

loichuder: :+1:

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -66,9 +87,9 @@ export function usePanZoom(camera: OrthographicCamera): ReactThreeFiber.Events {
...
       const { x: startX, y: startY } = startOffsetPosition.current;
 
-      camera.position.set(startX - pointX, startY - pointY, camera.position.z);
+      moveCamera(startX - pointX, startY - pointY);
     },
-    [camera, startOffsetPosition]
+    [camera, moveCamera]
   );
 
   const onWheel = useCallback(
```
loichuder, line 95: Just a remark: when I pan while zoomed and then zoom out, it is possible to be out of bounds. 
A possibility to fix this would be to add `moveCamera(camera.position.x, camera.position.y)` at the end of `onWheel` to go through the clamping code. But there is perhaps a better solution...

axelboc: That's the p(l)an! ;)

axelboc: I'll deal with moving the camera during zoom in the next PR, along with centring the zoom on the cursor.

PR #57, opened 2020-04-08
Zoom to cursor

### LINKED ISSUE #45: [Heatmap] Zoom in/out

### LINKED ISSUE #46: [Heatmap] Pan when zoomed in
Panning should be constrained to the edges of the Heatmap.

axelboc:
Fix #45, fix #46

### DIFF
--- a/src/h5web/visualizations/heatmap/utils.ts
+++ b/src/h5web/visualizations/heatmap/utils.ts
@@ -36,7 +36,7 @@ export function usePanZoom(): ReactThreeFiber.Events {
 
   const startOffsetPosition = useRef<Vector3>(); // `useRef` to avoid re-renders
 
-  const moveCamera = useCallback(
+  const moveCameraTo = useCallback(
     (x: number, y: number) => {
       const { position, zoom } = camera;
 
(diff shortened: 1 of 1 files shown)


src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -102,8 +102,17 @@ export function usePanZoom(): ReactThreeFiber.Events {
...
       camera.zoom = Math.max(1, camera.zoom * factor);
       camera.updateProjectionMatrix();
+
+      const projectedPoint = camera.worldToLocal(evt.unprojectedPoint.clone());
+      const { x: pointX, y: pointY } = projectedPoint;
+      const { x: camX, y: camY } = camera.position;
+
+      moveCamera(
+        camX + pointX * (1 - 1 / factor),
+        camY + pointY * (1 - 1 / factor)
```
axelboc, line 112: Pardon my non-scientific notations...

`point - point / factor` <=> `A - B`, where:
- `A` is the point in camera space where the cursor is located before the zoom,
- `B` is where that point moves to once the zoom is applied.

axelboc: I just move the camera by that amount. `moveCamera` then takes care of the clamping.

loichuder: In addition, passing a `Vector` as arg to `moveCamera` rather than the two coordinates could avoid destructuring in both `onPointerMove` and `onWheel`.

axelboc: - `moveCameraTo` 👍 
- I used `point` for consistency with `evt.unprojectedPoint` - it refers to the position of the pointer (which could be the mouse cursor or a finger on touchscreens).
- I'd also prefer to work with vectors but:
  - I don't want to move the camera along `z`, so it doesn't really make sense to work with and pass a `Vector3`, and I can't find a simple method to convert a `Vector3` to a `Vector2`.
  - Three.JS's vector API is a mutable API, so I have to clone the vectors before performing any operations on them, which affects readability and is prone to bugs. I feel like destructuring brings more clarity overall.

loichuder: `point` in unfortunately extremely generic... `pointer` ?

Yeah, after seeing the API of `Vector`, I feel your solution is clearer.

Issue #47, opened 2020-04-02
[Heatmap] Color scheme selection


axelboc:
Allow user to pick a different colour scheme -- https://github.com/d3/d3-scale-chromatic

loichuder:
I looked into different solutions for the selector:
- [react-dropdown](https://github.com/fraserxu/react-dropdown): easy to set up but lack of customization
- [react-dropdown-select](https://github.com/sanusart/react-dropdown-select/): looks more promising, I have to test

There is also the [dropdown from React-Semantic-UI](https://react.semantic-ui.com/modules/dropdown/) that looks more complete but installing the whole package only for that...

axelboc:
This might help as well: https://npms.io/search?q=react+dropdown

loichuder:
Settled for [react-select](https://react-select.com/home) in the end

PR #58, opened 2020-04-09
Added display of color bar in HeatmapVis

loichuder:


src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -32,6 +32,29 @@
...
+}
+
+.colorBarAxis {
+  display: block;
+  overflow: visible;
+  width: min-content;
+}
+
+.gradient {
+  border: solid black 1px;
```
loichuder, line 53: The border around the gradient is needed to my opinion but it shifts the ticks by the border size. The ticks are placed according to the height from `useMeasure` that includes the border. I don't know how to fix it yet

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -41,7 +64,6 @@
 .axis[data-orientation='left'] {
   text-anchor: end;
   dominant-baseline: middle;
-  min-height: 0;
```
loichuder, line 44: Typo. Will be fixed in next push.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -12,7 +14,13 @@ interface Props {
 
 function HeatmapVis(props: Props): JSX.Element {
   const { dims, data } = props;
-  const textureData = useMemo(() => computeTextureData(data), [data]);
+  const values = data.flat();
+  const interpolator = interpolateMagma;
```
loichuder, line 18: I anticipate a bit #47

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -12,7 +14,13 @@ interface Props {
...
+  const scale = generateColorScale(values, interpolator);
+  const textureData = useMemo(() => computeTextureData(values, scale), [
+    scale,
+    values,
+  ]);
```
loichuder, line 23: The computation of the `scale` was moved in a dedicated function as it is needed by the color bar and was therefore removed from `useMemo`. It is apparently working fine but we could put it back inside.

src/h5web/visualizations/heatmap/IndexAxis.tsx
```diff
@@ -13,11 +14,6 @@ interface Props {
   numberPixels: number;
 }
 
-const adaptedNumTicks = scaleLinear()
-  .domain([300, 900])
-  .rangeRound([3, 10])
-  .clamp(true);
-
```
loichuder, line 20: `adaptedNumTicks` was put in `utils` to be used by the color bar axis as well

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <div className={className}>
+      <div className={styles.colorBar}>
+        <div
+          ref={gradientRef}
+          className={styles.gradient}
+          style={{
+            background: `linear-gradient(${gradientColors})`,
+          }}
+        />
+        <svg className={styles.colorBarAxis}>
```
axelboc, line 33: May as well put `height={gradientHeight}` here, and a static `width`. It's better to have explicit dimensions on an SVG element. You can also remove `width: min-content` in the CSS, since this actually results in the SVG having a width of `0`.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -2,9 +2,9 @@
   display: grid;
   background-color: var(--primary-light-bg);
   grid-template-areas:
-    'not-used top-area'
-    'left-area map-area';
-  grid-template-columns: 4em 1fr;
+    'not-used top-area not-used-2'
+    'left-area map-area right-area';
```
axelboc, line 6: The convention seems to be to use a `.` for empty cells.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -32,6 +32,29 @@
...
 
+.rightArea {
+  grid-area: right-area;
+}
+
+.colorBar {
+  display: flex;
+  justify-content: center;
+  width: 100%;
+  height: 95%;
```
axelboc, line 43: Use padding on the parent instead of `height` here

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -32,6 +32,29 @@
...
 }
 
+.rightArea {
+  grid-area: right-area;
+}
+
+.colorBar {
+  display: flex;
+  justify-content: center;
+  width: 100%;
```
axelboc, line 42: use flexbox if you can

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -32,6 +32,29 @@
...
+  height: 95%;
+}
+
+.colorBarAxis {
+  display: block;
+  overflow: visible;
+  width: min-content;
```
axelboc, line 49: As per one of my previous comments, this gives the axis a width of 0, which is why the colour bar is centred horizontally as if the axis wasn't there. Just put a fixed width on the `svg` element in the JSX to fix the centring.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
axelboc: You can try to use `box-sizing: content-box`. If you then set `gradientHeight` on the `svg` element, it should become 2px shorter than the colour bar. You can then use `align-self: center` on the axis to align it vertically.

If all of this doesn't work, use `padding-top: 1px` on the axis.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -32,6 +32,29 @@
...
+}
+
+.gradient {
+  border: solid black 1px;
+  min-width: 1em;
+  width: 30%;
```
axelboc, line 55: `min-width` doesn't change anything since the grid cell has a fixed width.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: I think you can even remove `useMemo` altogether. I was wondering about it, but since we already use `React.memo` on `HeatmapVis` itself, `computeTextureData` will always be called with new values.

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -1,35 +1,48 @@
...
 
-export function computeTextureData(matrix: number[][]): Uint8Array | undefined {
-  const values = matrix.flat();
-  const [min, max] = extent(values);
-
-  if (min === undefined || max === undefined) {
+export function computeTextureData(
+  values: number[],
+  scale?: ScaleSequential<string>
+): Uint8Array | undefined {
```
axelboc, line 14: Might be easier to require `scale` and call this function like so: `const textureData = scale && computeTextureData(...)`

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <AxisRight
+            scale={scaleLinear()
+              .domain(scale.domain())
+              .range([gradientHeight, 0])}
```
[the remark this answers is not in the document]
loichuder: I don't think I can do that: `scale` is a `ScaleSequential` whereas I need a `LinearScale` here

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { className, scale } = props;
+  const interpolator = scale.interpolator();
+  const gradientColors = range(0, 1.1, 0.1)
+    .map(i => `${interpolator(i)}`)
+    .reduce((acc, val) => `${acc},${val}`);
+
+  const [gradientRef, { height: gradientHeight }] = useMeasure();
+
+  return (
+    <div className={className}>
```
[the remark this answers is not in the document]
axelboc: I think the div below, with class `colorBar` can go, not the one above this comment.

src/h5web/visualizations/heatmap/ColorBar.tsx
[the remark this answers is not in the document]
axelboc: Ah right! So you don't actually need the original `scale` in this component. Perhaps it would make more sense to pass the interpolator and the domain separately?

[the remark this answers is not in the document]
axelboc: Maybe `generateColorScale` should become `findDomain` or something... In the end, you only need the `sequentialScale` inside `computeTextureData`.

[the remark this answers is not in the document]
loichuder: My point was that all I needed was inside `scale` so that it was easier to pass it around. But it might be clearer in the end to do as you told.

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -0,0 +1,48 @@
...
+interface Props {
+  className: string;
+  interpolator: (t: number) => string;
+  dataBounds: [number, number];
+}
+
+function ColorBar(props: Props): JSX.Element {
+  const { className, interpolator, dataBounds } = props;
+  const gradientColors = range(0, 1.1, 0.1)
+    .map(i => `${interpolator(i)}`)
```
axelboc, line 18: ça retourne déjà une string, donc pas besoin de la template string -- je dirais même, autant faire tout dans le `reduce`

axelboc: `.map(interpolator)` suffit, donc

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <div
+        ref={gradientRef}
+        className={styles.gradient}
+        style={{
+          backgroundImage: `linear-gradient(to top, ${gradientColors})`,
+        }}
+      />
+      <svg
+        className={styles.colorBarAxis}
+        style={{ width: '2em', height: gradientHeight }}
```
axelboc, line 34: I meant attributes, not styles: `width="2em" height={gradientHeight}`

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -32,10 +32,28 @@
...
+  grid-area: right-area;
+  display: flex;
+  padding: 1em;
+  min-height: 0;
+}
+
+.colorBarAxis {
+  display: block;
+  overflow: visible;
+  flex: 0 1 auto;
```
axelboc, line 45: c'est la valeur par défaut, donc tu peux retirer

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -12,7 +15,13 @@ interface Props {
...
 function HeatmapVis(props: Props): JSX.Element {
   const { dims, data } = props;
-  const textureData = useMemo(() => computeTextureData(data), [data]);
+  const values = data.flat();
+
+  const interpolator = interpolateMagma;
+  const domain = findDomain(values);
+  const textureData =
+    domain &&
+    computeTextureData(values, scaleSequential(interpolator).domain(domain));
```
axelboc, line 24: Tu peux peut-être remettre la création de la scale dans `computeTextureData`, du coup ? Tu lui passes juste le `domain` et l'interpolateur et si le domain est `undefined`, tu retournes `undefined`:

```
const textureData = computeTextureData(values, domain, interpolator);
```

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -0,0 +1,47 @@
...
+            background: `linear-gradient(${gradientColors})`,
+          }}
+        />
+        <svg className={styles.colorBarAxis}>
+          <AxisRight
+            scale={scaleLinear()
+              .domain(scale.domain())
+              .range([gradientHeight, 0])}
+            strokeWidth={0}
+            numTicks={adaptedNumTicks(gradientHeight)}
```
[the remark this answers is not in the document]
loichuder: `adaptedNumTicks` seems fine so I think it comes from VX's `Axis`: [VX Axis doc](https://vx-demo.now.sh/static/docs/vx-axis.html#AxisBottom__numTicks) indeed tells that `numTicks` is approximate.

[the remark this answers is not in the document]
loichuder: Nevertheless, it is still possible to smooth rounding issues with `nice` from [d3-scale](https://github.com/d3/d3-scale/blob/master/README.md#continuous_nice).

PR #59, opened 2020-04-09
Optimise rendering of DatasetVisualizer

axelboc:
### Optimise consecutive renders when navigating in the explorer

The re-renders when navigating in the explorer are caused by `App`, so the memoization we had on `HeatmapVis` can be moved higher up, to the direct descendants of `App`: `DatasetVisualizer` and `MetadataViewer`.

Now we'll notice if we ever trigger re-renders in `HeatmapVis` by mistake.

### Let React reconcile children of `DatasetVisualizer`

`key` prop was telling React to delete the entire DOM sub-tree and to recreate it when switching between two datasets, leading to the ThreeJS canvas and rendering context having to be recreated each time.

### DIFF
--- a/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
+++ b/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo } from 'react';
+import React, { useState, useMemo, useEffect } from 'react';
 import { HDF5Dataset } from '../providers/models';
 import styles from './DatasetVisualizer.module.css';
 import VisSelector from './VisSelector';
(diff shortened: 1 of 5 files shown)


src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -14,9 +14,11 @@ function DatasetVisualizer(props: Props): JSX.Element {
   const { dataset } = props;
 
   const supportedVis = useMemo(() => getSupportedVis(dataset), [dataset]);
-  const [activeVis, setActiveVis] = useState<Vis | undefined>(
-    dataset && supportedVis[supportedVis.length - 1]
-  );
+  const [activeVis, setActiveVis] = useState<Vis>();
+
+  useEffect(() => {
+    setActiveVis(supportedVis[supportedVis.length - 1]);
```
axelboc, line 20: no out-of-bounds exception in JS, so `[][-1]` returns `undefined` which is fine by us

PR #61, opened 2020-04-09
Implements selection of the color map

### LINKED ISSUE #47: [Heatmap] Color scheme selection

loichuder:
Fix #47


src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -16,9 +17,12 @@ function HeatmapVis(props: Props): JSX.Element {
   const { dims, data } = props;
   const values = data.flat();
 
-  const interpolator = interpolateMagma;
   const domain = findDomain(values);
-  const textureData = computeTextureData(values, domain, interpolator);
+  const [colorMap, setColorMap] = useState<ColorMap>({
+    name: 'Magma',
+    interpolator: interpolateMagma,
+  });
```
axelboc, line 24: You could probably simplify a few things by not having a `ColorMap` interface. That is, instead of keeping both the name of the colour map and the intepolator in a state here, you could just store an enum value, like `Interpolator.Magma`. The enum would live in `./models.ts`.

The record that you use to map colour map names to interpolators in `ColorMapSelector`, could then be moved to `utils` and you could use it directly to access the interpolator from the enum any time you need it (like below).

loichuder: I now have a single record with names in keys and interpolators as values. The `ColorMap` type models variables that are the keys of this record.

src/h5web/visualizations/heatmap/models.ts
```diff
@@ -0,0 +1,83 @@
...
+  Oranges: interpolateOranges,
+  Purples: interpolatePurples,
+  Reds: interpolateReds,
+  // Sequential (multi-hue)
+  // Turbo missing !
+  Viridis: interpolateViridis,
+  Inferno: interpolateInferno,
+  Magma: interpolateMagma,
+  Plasma: interpolatePlasma,
+  // Cividis missing !
```
loichuder, line 63: Cividis and Turbo colormaps should implemented in the version of d3-scale-chromatic that we are using ([1.5.0](https://github.com/d3/d3-scale-chromatic/releases)) but I cannot import them. Perhaps they are missing from the `@types/d3-scale-chromatic` package ?

src/h5web/visualizations/heatmap/models.ts
```diff
@@ -0,0 +1,83 @@
+import {
```
loichuder, line 1: I wonder if `models` is a good file name given it only exports color maps.

src/h5web/visualizations/heatmap/models.ts
loichuder: Indeed, `@types/d3-scale-chromatic` is for 1.3.2 : https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/d3-scale-chromatic/index.d.ts#L9

src/h5web/visualizations/heatmap/ColorMapSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <Select
+        defaultValue={{ label: currentColorMap }}
+        options={colorMapOptions}
+        onChange={selection => {
+          const colorOption = selection as { label: ColorMap };
+          changeColorMap(colorOption.label);
+        }}
+        menuPlacement="top"
+        styles={gradientStyles}
+      />
```
axelboc, line 48: FYI, when you'll refactor the design, it may be easier [to replace](https://react-select.com/components#replacing-components) react-select's default `Option` component instead of overriding the styles. Perhaps you were already planning to.

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -1,22 +1,19 @@
...
 import { useMeasure } from 'react-use';
-import { adaptedNumTicks } from './utils';
+import { adaptedNumTicks, generateCSSLinearGradient } from './utils';
 import styles from './HeatmapVis.module.css';
+import { ColorMap, COLOR_MAPPINGS } from './models';
 
 interface Props {
   className: string;
-  interpolator: (t: number) => string;
+  colorMap: ColorMap;
```
axelboc, line 11: You can keep passing the interpolator here.

src/h5web/visualizations/heatmap/models.ts
```diff
@@ -0,0 +1,83 @@
...
+  interpolateRdPu,
+  interpolateYlGnBu,
+  interpolateYlGn,
+  interpolateYlOrBr,
+  interpolateYlOrRd,
+  interpolateRainbow,
+  interpolateSinebow,
+} from 'd3-scale-chromatic';
+
+export const COLOR_MAPPINGS: Record<string, (t: number) => string> = {
```
axelboc, line 39: You could add a type for interpolator functions. You could then use it in your `ColorBar` component.

src/h5web/visualizations/heatmap/models.ts
axelboc: Yeah, perhaps `interpolators.ts` would make more sense.

src/h5web/visualizations/heatmap/ColorMapSelector.tsx
loichuder: Thanks :+1: !

PR #63, opened 2020-04-14
Fix stale `activeVis` state

axelboc:
By removing the `key` prop from `DatasetVisualizer`, I introduced a regression where `activeVis` could become stale when switching between two datasets with different sets of supported visualizations. `activeVis` is updated asynchronously with `useEffect`, so it may become stale for one render.

I now check whether `activeVis` is stale before rendering `VisDisplay`.

### DIFF
--- a/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
+++ b/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
@@ -1,9 +1,8 @@
-import React, { useState, useMemo, useEffect } from 'react';
+import React, { useMemo } from 'react';
 import { HDF5Dataset } from '../providers/models';
 import styles from './DatasetVisualizer.module.css';
 import VisSelector from './VisSelector';
-import { Vis } from './models';
-import { getSupportedVis } from './utils';
+import { getSupportedVis, useActiveVis } from './utils';
 import VisDisplay from './VisDisplay';
 
 interface Props {
@@ -14,11 +13,7 @@ function DatasetVisualizer(props: Props): JSX.Element {
   const { dataset } = props;
 
   const supportedVis = useMemo(() => getSupportedVis(dataset), [dataset]);
-  const [activeVis, setActiveVis] = useState<Vis>();
-
-  useEffect(() => {
-    setActiveVis(supportedVis[supportedVis.length - 1]);
-  }, [supportedVis]);
+  const [activeVis, setActiveVis] = useActiveVis(supportedVis);
 
   return (
     <div className={styles.visualizer}>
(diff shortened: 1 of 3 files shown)


src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -28,8 +23,8 @@ function DatasetVisualizer(props: Props): JSX.Element {
         onChange={setActiveVis}
       />
       <div className={styles.displayArea}>
-        {dataset && activeVis ? (
-          <VisDisplay vis={activeVis} dataset={dataset} />
+        {dataset ? (
+          activeVis && <VisDisplay vis={activeVis} dataset={dataset} />
```
axelboc, line 27: If `dataset` is defined but `activeVis` isn't, it means it's either the first render or a render where `activeVis` is stale. Either way, `activeVis` will be defined in the next render, so we don't need to show the "Nothing to visualize" message.

src/h5web/dataset-visualizer/utils.ts
```diff
@@ -46,3 +47,20 @@ export function getSupportedVis(dataset?: HDF5Dataset): Vis[] {
...
+  // When switching between two datasets, `activeVis` may become stale for one render
+  const isStale = !activeVis || !supportedVis.includes(activeVis);
+
+  useEffect(() => {
+    if (isStale) {
+      setActiveVis(supportedVis[supportedVis.length - 1]);
+    }
+  }, [isStale, supportedVis]);
+
+  return [isStale ? undefined : activeVis, setActiveVis];
```
axelboc, line 65: `activeVis` is now consistently `undefined` both on first render and on stale renders.

src/h5web/dataset-visualizer/utils.ts
```diff
@@ -46,3 +47,20 @@ export function getSupportedVis(dataset?: HDF5Dataset): Vis[] {
...
   return supported.length > 1 ? supported.slice(1) : supported;
 }
+
+export function useActiveVis(
+  supportedVis: Vis[]
+): [Vis | undefined, (vis: Vis) => void] {
+  const [activeVis, setActiveVis] = useState<Vis>();
+
+  // When switching between two datasets, `activeVis` may become stale for one render
+  const isStale = !activeVis || !supportedVis.includes(activeVis);
```
axelboc, line 57: To check whether `activeVis` is stale, we could compare the current and previous `dataset` passed as prop to `DatasetVisualizer`. This would match the beahaviour before the regression: `activeVis` would reset back to the last supported visualization when selecting a dataset.

However, a better behaviour is to not change the `activeVis` if the vusalization is also supported by the newly selected dataset. This is why I only mark `activeVis` as stale if it doesn't exist in the new `supportedVis` array.

loichuder: Just a thought: I find more readable to reverse the logic by checking if the visualization is valid rather than stale.  This line would become 
```suggestion
  const isValid = activeVis && supportedVis.includes(activeVis);
``` 
and the following checks below would be reversed or check `!isValid`.

axelboc: You're totally right 👍

PR #79, opened 2020-04-15
Implemented log scale for the colormap

loichuder:
image

Uses [d3.scaleSymlog](https://github.com/d3/d3-scale/#symlog-scales) to handle 0 and negative numbers.

### DIFF
--- a/src/h5web/visualizations/heatmap/ColorBar.tsx
+++ b/src/h5web/visualizations/heatmap/ColorBar.tsx
@@ -1,24 +1,29 @@
 import React from 'react';
-import { scaleLinear } from 'd3-scale';
 import { AxisRight } from '@vx/axis';
 import { useMeasure } from 'react-use';
-import { adaptedNumTicks, generateCSSLinearGradient } from './utils';
+import {
+  adaptedNumTicks,
+  generateCSSLinearGradient,
+  ColorScale,
+} from './utils';
 import styles from './HeatmapVis.module.css';
 import { D3Interpolator } from './interpolators';
 
 interface Props {
-  className: string;
   interpolator: D3Interpolator;
-  dataBounds: [number, number];
+  colorScale: ColorScale;
 }
 
 function ColorBar(props: Props): JSX.Element {
-  const { className, interpolator, dataBounds } = props;
+  const { interpolator, colorScale } = props;
 
   const [gradientRef, { height: gradientHeight }] = useMeasure();
+  const scale = colorScale
+    .copy()
+    .range([gradientHeight, 0]) as typeof colorScale;
 
   return (
-    <div className={className}>
+    <div className={styles.colorBar}>
       <div
         ref={gradientRef}
         className={styles.gradient}
@@ -28,12 +33,10 @@ function ColorBar(props: Props): JSX.Element {
       />
       <svg className={styles.colorBarAxis} height={gradientHeight} width="2em">
         <AxisRight
-          scale={scaleLinear()
-            .domain(dataBounds)
-            .range([gradientHeight, 0])
-            .nice()}
-          strokeWidth={0}
+          scale={scale}
+          hideAxisLine
           numTicks={adaptedNumTicks(gradientHeight)}
+          tickFormat={scale.tickFormat(adaptedNumTicks(gradientHeight), '.3')}
         />
       </svg>
     </div>
--- a/src/h5web/visualizations/heatmap/HeatmapVis.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.tsx
@@ -1,12 +1,13 @@
 import { Canvas } from 'react-three-fiber';
 import React, { useState } from 'react';
 import Mesh from './Mesh';
-import { computeTextureData, findDomain } from './utils';
+import { computeTextureData, findDomain, getColorScale } from './utils';
 import styles from './HeatmapVis.module.css';
 import IndexAxis from './IndexAxis';
 import ColorBar from './ColorBar';
 import ColorMapSelector from './ColorMapSelector';
 import { ColorMap, INTERPOLATORS } from './interpolators';
+import LogScaleToggler from './LogScaleToggler';
 
 interface Props {
   dims: [number, number];
@@ -19,10 +20,13 @@ function HeatmapVis(props: Props): JSX.Element {
 
   const domain = findDomain(values);
   const [colorMap, setColorMap] = useState<ColorMap>('Magma');
+  const [logScale, setLogScale] = useState<boolean>(false);
+
+  const colorScale = getColorScale(domain, logScale);
   const textureData = computeTextureData(
     values,
-    domain,
-    INTERPOLATORS[colorMap]
+    INTERPOLATORS[colorMap],
+    colorScale
   );
 
   return (
@@ -43,12 +47,14 @@ function HeatmapVis(props: Props): JSX.Element {
         orientation="bottom"
         numberPixels={dims[1]}
       />
-      {domain && (
-        <ColorBar
-          className={styles.rightArea}
-          interpolator={INTERPOLATORS[colorMap]}
-          dataBounds={domain}
-        />
+      {colorScale && (
+        <div className={styles.rightArea}>
+          <LogScaleToggler value={logScale} onChange={setLogScale} />
+          <ColorBar
+            interpolator={INTERPOLATORS[colorMap]}
+            colorScale={colorScale}
+          />
+        </div>
       )}
       <ColorMapSelector
         className={styles.bottomRightArea}
--- a/src/h5web/visualizations/heatmap/LogScaleToggler.tsx
+++ b/src/h5web/visualizations/heatmap/LogScaleToggler.tsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import { FiEyeOff, FiEye } from 'react-icons/fi';
+import styles from './HeatmapVis.module.css';
+
+interface Props {
+  value: boolean;
+  onChange: (val: boolean) => void;
+}
+
+function LogScaleToggler(props: Props): JSX.Element {
+  const { value, onChange } = props;
+
+  return (
+    <button
+      onClick={() => {
+        onChange(!value);
+      }}
+      type="button"
+      role="switch"
+      aria-checked={value}
+      className={styles.scaleTypeSelector}
+    >
+      {value ? <FiEye /> : <FiEyeOff />}
+      SymLog scale
+    </button>
+  );
+}
+
+export default LogScaleToggler;
--- a/src/h5web/visualizations/heatmap/utils.ts
+++ b/src/h5web/visualizations/heatmap/utils.ts
@@ -1,26 +1,35 @@
-import React, { useCallback, useRef, useEffect } from 'react';
 import { extent } from 'd3-array';
 import { rgb } from 'd3-color';
-import { scaleLinear, scaleSequential } from 'd3-scale';
-import { Vector3 } from 'three';
-import { ReactThreeFiber, PointerEvent, useThree } from 'react-three-fiber';
-import { clamp, range } from 'lodash-es';
+import {
+  scaleSequential,
+  ScaleLinear,
+  ScaleSymLog,
+  scaleLinear,
+  scaleSymlog,
+} from 'd3-scale';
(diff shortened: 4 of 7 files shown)


src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -28,12 +33,10 @@ function ColorBar(props: Props): JSX.Element {
       />
       <svg className={styles.colorBarAxis} height={gradientHeight} width="2em">
         <AxisRight
-          scale={scaleLinear()
-            .domain(dataBounds)
-            .range([gradientHeight, 0])
-            .nice()}
-          strokeWidth={0}
+          scale={scale}
+          hideAxisLine
```
loichuder, line 37: `strokeWidth={0}` was replaced by `hideAxisLine` as the property was exisiting...

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -28,12 +33,10 @@ function ColorBar(props: Props): JSX.Element {
...
           numTicks={adaptedNumTicks(gradientHeight)}
+          tickFormat={scale.tickFormat(adaptedNumTicks(gradientHeight), '.3')}
```
loichuder, line 39: Ticks are a bit in funny places with the symlog scale. I think this would need some dedication to get something nice.

src/h5web/visualizations/heatmap/LogScaleToggler.tsx
```diff
@@ -0,0 +1,29 @@
...
+      aria-checked={logScale}
+      className={styles.scaleTypeSelector}
+    >
+      {logScale ? <FiEye /> : <FiEyeOff />}
+      SymLog scale
+    </button>
+  );
+}
+
+export default LogScaleToggler;
```
loichuder, line 29: The logic in there is quite simple but I preferred to extract it as a component as the `HeatmapVis` starts to be a bit bloated...

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -1,26 +1,35 @@
...
-  const scale = scaleSequential(interpolator).domain(domain);
+
+  // Map colors to the output of colorScale
+  const scale = scaleSequential(interpolator).domain(
+    colorScale.range() as [number, number]
+  );
   // Compute RGB color array for each datapoint `[[<r>, <g>, <b>], [<r>, <g>, <b>], ...]`
   const colors = values.map(val => {
-    const { r, g, b } = rgb(scale(val)); // `colorMap` returns CSS RGB strings
+    const { r, g, b } = rgb(scale(colorScale(val))); // `scale` returns CSS RGB strings
```
loichuder, line 32: The scaleSequential must map the colors to the output of the colorScale (which are either a linear extrapolation of the data or a log extrapolation).
In short, we have: `data` ---`colorScale`----> `scaledData` ---`scaleSequential`---> `colors`

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -151,3 +60,14 @@ export function generateCSSLinearGradient(
...
+
+export function getColorScale(
+  domain: [number, number] | undefined,
+  logScale: boolean
+): ColorScale | undefined {
+  if (domain === undefined) {
+    return undefined;
+  }
+
+  return logScale ? scaleSymlog().domain(domain) : scaleLinear().domain(domain);
```
loichuder, line 72: The domain is set to the bounds of the data but the range needs not to be set as:
- in `computeTextureData`: the range is irrevelevant as it is mapped directly to the domain of the `scaleSequential`.
- in the axis of `ColorBarAxis`; the range is set to the height of the gradient.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -43,12 +47,14 @@ function HeatmapVis(props: Props): JSX.Element {
...
-          className={styles.rightArea}
-          interpolator={INTERPOLATORS[colorMap]}
-          dataBounds={domain}
-        />
+      {colorScale && (
+        <div className={styles.rightArea}>
+          <LogScaleToggler logScale={logScale} setLogScale={setLogScale} />
+          <ColorBar
+            interpolator={INTERPOLATORS[colorMap]}
+            colorScale={colorScale}
```
loichuder, line 55: `colorScale` and `interpolator` are always passed as props together. It might be worthwhile to make an object out of both.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -19,9 +20,12 @@ function HeatmapVis(props: Props): JSX.Element {
 
   const domain = findDomain(values);
   const [colorMap, setColorMap] = useState<ColorMap>('Magma');
+  const [logScale, setLogScale] = useState<boolean>(false);
```
loichuder, line 23: I use a second State (the first being for the name of the colormap) to track the value of the logScale boolean and trigger rendering. I wonder if this is the best way to do it.

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -28,12 +33,10 @@ function ColorBar(props: Props): JSX.Element {
       />
       <svg className={styles.colorBarAxis} height={gradientHeight} width="2em">
         <AxisRight
-          scale={scaleLinear()
-            .domain(dataBounds)
-            .range([gradientHeight, 0])
-            .nice()}
-          strokeWidth={0}
+          scale={scale}
```
loichuder, line 36: Rather than a `linearScale`, the scale of the axis is now the `colorScale` (linear or log) mapped to the height of the gradient bar.

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -1,24 +1,29 @@
...
+  ColorScale,
+} from './utils';
 import styles from './HeatmapVis.module.css';
 import { D3Interpolator } from './interpolators';
 
 interface Props {
-  className: string;
   interpolator: D3Interpolator;
-  dataBounds: [number, number];
+  colorScale: ColorScale;
```
loichuder, line 14: `colorScale` contains the `dataBounds` as its domain

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -1,24 +1,29 @@
...
 }
 
 function ColorBar(props: Props): JSX.Element {
-  const { className, interpolator, dataBounds } = props;
+  const { interpolator, colorScale } = props;
 
   const [gradientRef, { height: gradientHeight }] = useMeasure();
+  const scale = colorScale
+    .copy()
+    .range([gradientHeight, 0]) as typeof colorScale;
```
axelboc, line 23: `const scale: typeof ColorScale = ...` ne fonctionne pas ?

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: I'm working on implementing a common heatmap state. 👍

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: The common state will help with that.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -43,12 +47,14 @@ function HeatmapVis(props: Props): JSX.Element {
...
       />
-      {domain && (
-        <ColorBar
-          className={styles.rightArea}
-          interpolator={INTERPOLATORS[colorMap]}
-          dataBounds={domain}
-        />
+      {colorScale && (
+        <div className={styles.rightArea}>
+          <LogScaleToggler logScale={logScale} setLogScale={setLogScale} />
```
axelboc, line 52: `value` and `onChange` and generally used.

src/h5web/visualizations/heatmap/LogScaleToggler.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { FiEyeOff, FiEye } from 'react-icons/fi';
+import styles from './HeatmapVis.module.css';
+
+interface Props {
+  logScale: boolean;
+  setLogScale: React.Dispatch<React.SetStateAction<boolean>>;
```
axelboc, line 7: Avoid using this type in props, as it's too specific -- if we need to pass a function that does something with the boolean before setting the state, then we'll have to change the type here too.

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -1,26 +1,35 @@
...
 
-const ZOOM_FACTOR = 0.95;
+export type ColorScale =
+  | ScaleLinear<number, number>
+  | ScaleSymLog<number, number>;
 
 export function computeTextureData(
   values: number[],
-  domain: [number, number] | undefined,
+  colorScale: ColorScale | undefined,
```
axelboc, line 19: `colorScale?: ColorScale`

src/h5web/visualizations/heatmap/utils.ts
axelboc: Do you think there's a better way to name the scales ?

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -1,26 +1,35 @@
...
 ): Uint8Array | undefined {
-  if (domain === undefined) {
+  if (colorScale === undefined) {
     return undefined;
   }
-  const scale = scaleSequential(interpolator).domain(domain);
+
+  // Map colors to the output of colorScale
+  const scale = scaleSequential(interpolator).domain(
+    colorScale.range() as [number, number]
```
axelboc, line 28: It's a bit like doing `scale.copy()` in `ColorBar.tsx`. I'm not very fond of creating scales from existing scales... In this case, it forces you to cast the type of the range with `as`. I feel like each scale should be recreated from the raw building blocks (domain, range, etc.)

Don't worry, though, as my next PR will help with this.

src/h5web/visualizations/heatmap/utils.ts
axelboc: You can simplify this: 

```ts
const scaleFunction = logScale ? scaleSymlog : scaleLinear;
return scaleFunction().domain(domain);
```

src/h5web/visualizations/heatmap/utils.ts
loichuder: I thought about `preScale` but that's ugly. I like `colorScale` as it means that the data is scaled either linearly or logarithmically to the colors. It just happens that in the implementation, we need a `scaleSequential`.
This being said, I concur that it may be misleading and if you have a suggestion, go ahead !

src/h5web/visualizations/heatmap/utils.ts
loichuder: I had to add `as ColorScale` to the return value to make TS happy. There are fishy things with d3-scale types...

src/h5web/visualizations/heatmap/ColorBar.tsx
loichuder: No it doesn't... :frowning_face: 
And to be fair, the cast should not be necessary to my opinion.

src/h5web/visualizations/heatmap/utils.ts
loichuder: I would agree but in my case, I would need to pass around the type of scale, the domain and recreate the scale which seems heavy versus passing only the scale.

Wait and see :monocle_face:

src/h5web/visualizations/heatmap/LogScaleToggler.tsx
loichuder: What do you suggest instead then ?

axelboc: Ah sorry, just `(val: boolean) => void`

src/h5web/visualizations/heatmap/utils.ts
loichuder: Will be changed in the next PR

PR #80, opened 2020-04-16
Provide centralised heatmap state through context

axelboc:
I use a lib called [easy-peasy](https://easy-peasy.now.sh/), more specifically [createContextStore](https://easy-peasy.now.sh/docs/api/create-context-store.html). It's built on top of Redux while providing a simple API for defining actions, computed properties, etc. and for accessing the state with hooks.

I've also refactored `IndexAxis` a little to use VX's `AxisLeft` and `AxisRight` components instead of `Axis`, which simplifies a couple of things.

### DIFF
--- a/src/h5web/visualizations/heatmap/IndexAxis.tsx
+++ b/src/h5web/visualizations/heatmap/IndexAxis.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { useMeasure } from 'react-use';
-import { Axis } from '@vx/axis';
+import { AxisLeft, AxisBottom } from '@vx/axis';
 import { format } from 'd3-format';
 import { scaleLinear } from 'd3-scale';
 import styles from './HeatmapVis.module.css';
@@ -16,31 +16,33 @@ interface Props {
 
 function IndexAxis(props: Props): JSX.Element {
   const { orientation, numberPixels, className } = props;
+
   const [divRef, { width, height }] = useMeasure();
   const isLeftAxis = orientation === 'left';
+  const Axis = isLeftAxis ? AxisLeft : AxisBottom;
 
-  const numTicks = isLeftAxis
-    ? adaptedNumTicks(height)
-    : adaptedNumTicks(width);
+  const scale = scaleLinear()
+    .domain([-0.5, numberPixels - 0.5])
+    .range(isLeftAxis ? [height, 0] : [0, width]);
+
+  const numTicks = Math.min(
+    numberPixels,
+    adaptedNumTicks(isLeftAxis ? height : width)
+  );
 
   return (
     <div ref={divRef} className={className}>
       <svg className={styles.axis} data-orientation={orientation}>
         <Axis
-          scale={scaleLinear()
-            .domain([-0.5, numberPixels - 0.5])
-            .range(isLeftAxis ? [height, 0] : [0, width])}
+          scale={scale}
           left={isLeftAxis ? width : 0}
-          orientation={orientation}
-          strokeWidth={0}
+          numTicks={numTicks}
+          hideAxisLine
           tickFormat={format('0')}
-          numTicks={Math.min(numberPixels, numTicks)}
-          // eslint-disable-next-line @typescript-eslint/no-unused-vars
-          tickLabelProps={(value, index) =>
-            isLeftAxis
-              ? { className: styles.tick, dx: '-0.25em' }
-              : { className: styles.tick, dy: '0.25em' }
-          }
+          tickClassName={styles.tick}
+          tickComponent={({ formattedValue, ...tickProps }) => (
+            <text {...tickProps}>{formattedValue}</text>
+          )}
         />
       </svg>
     </div>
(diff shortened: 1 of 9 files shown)


src/h5web/visualizations/heatmap/IndexAxis.tsx
```diff
@@ -16,31 +16,33 @@ interface Props {
 
 function IndexAxis(props: Props): JSX.Element {
   const { orientation, numberPixels, className } = props;
+
   const [divRef, { width, height }] = useMeasure();
   const isLeftAxis = orientation === 'left';
+  const Axis = isLeftAxis ? AxisLeft : AxisBottom;
```
axelboc, line 22: This is the line that allows rendering one component or the other conditionally.

src/h5web/visualizations/heatmap/IndexAxis.tsx
```diff
@@ -16,31 +16,33 @@ interface Props {
...
-          tickLabelProps={(value, index) =>
-            isLeftAxis
-              ? { className: styles.tick, dx: '-0.25em' }
-              : { className: styles.tick, dy: '0.25em' }
-          }
+          numTicks={numTicks}
+          tickClassName={styles.tick}
+          tickComponent={({ formattedValue, ...tickProps }) => (
+            <text {...tickProps}>{formattedValue}</text>
+          )}
```
axelboc, line 45: VX rend son propre composant `Text` par défaut, qui lui même rend un élement `svg`. On se retrouve donc avec `svg > g > svg > text`, ce qui n'est pas génail. `tickComponent` permet d'obtenir `svg > g > text` comme dans l'exemple "Axis" de VX.

src/h5web/visualizations/heatmap/interpolators.ts
```diff
@@ -38,7 +38,7 @@ import {
 
 export type D3Interpolator = (t: number) => string;
 
-export const INTERPOLATORS: Record<string, D3Interpolator> = {
+export const INTERPOLATORS = {
```
axelboc, line 41: Avec ce typage, `ColorMap` se retrouvait à valoir `string` au lieu de `'BrBG' | 'PRGn' | ...`. C'est un serpent qui se mord la queue; on ne peut pas avoir un typage fort partout sans définir un enum en plus d'`INTERPOLATORS`.

src/h5web/visualizations/heatmap/store.ts
```diff
@@ -0,0 +1,23 @@
...
+  }),
+});
+
+export function useHeatmapState(): State<HeatmapState> {
+  return HeatmapStore.useStoreState(state => state);
+}
+
+export function useHeatmapActions(): Actions<HeatmapState> {
+  return HeatmapStore.useStoreActions(state => state);
+}
```
axelboc, line 23: These two hooks are just sugar. We could also use `HeatmapStore.useStoreState` and `HeatmapStore.useStoreACtions` directly in our components:

```ts
const colorMap = HeatmapStore.useStoreState(state => state.colorMap);
```

src/h5web/visualizations/heatmap/store.ts
```diff
@@ -0,0 +1,23 @@
...
+
+interface HeatmapState {
+  colorMap: ColorMap;
+  setColorMap: Action<HeatmapState, ColorMap>;
+}
+
+export const HeatmapStore = createContextStore<HeatmapState>({
+  colorMap: 'Magma',
+  setColorMap: action((state, colorMap) => {
+    state.colorMap = colorMap;
```
axelboc, line 13: We mutate the state. easy-peasy makes this operation safe, as explained here: https://easy-peasy.now.sh/docs/tutorial/using-actions-to-update-state.html -- unfortunately, ESLint doesn't like this, which is why I disable a rule at the top of the file.

src/h5web/visualizations/heatmap/store.ts
```diff
@@ -0,0 +1,23 @@
...
+import { createContextStore, Action, State, action, Actions } from 'easy-peasy';
+import { ColorMap } from './interpolators';
+
+interface HeatmapState {
+  colorMap: ColorMap;
+  setColorMap: Action<HeatmapState, ColorMap>;
+}
+
+export const HeatmapStore = createContextStore<HeatmapState>({
+  colorMap: 'Magma',
```
axelboc, line 11: Note that by removing the record type from `INTERPOLATORS`, I now get proper auto-completion for `colorMap` here.

src/h5web/visualizations/heatmap/store.ts
```diff
@@ -0,0 +1,23 @@
...
+    state.colorMap = colorMap;
+  }),
+});
+
+export function useHeatmapState(): State<HeatmapState> {
+  return HeatmapStore.useStoreState(state => state);
+}
+
+export function useHeatmapActions(): Actions<HeatmapState> {
+  return HeatmapStore.useStoreActions(state => state);
```
loichuder, line 22: Should not it be `HeatmapStore.useStoreActions(actions => actions);` for consistency ?

axelboc: yes, t'as raison, je fix dans ma nouvelle branche pour me simplifier la vie

PR #82, opened 2020-04-16
Replaced numeral by d3-format in Cell

loichuder:
I wonder if using d3-format capabilities, we might be able to have easier-to-read numbers.

Compare now (same as previously): 
image

with this for example:
image


loichuder:
After trying different things, I withdraw my comment. This is still the best compromise:
image

PR #83, opened 2020-04-16
Move domain, interpolator and scales to Heatmap state

axelboc:
The commits give a good overview of the refactor, but basically the whole domain/interpolator/scale logic now resides in a single place: the store.


axelboc:
Putting `values` in the store would work, but I didn't want to go as far in case `findDomain` becomes an asynchronous operation in the future (i.e. if it moves into a web worker). Computed properties are not meant to be asynchronous in easy-peasy. They have _thunks_ for that, so we'll probably end up with a `setValue` thunk that finds the domain asynchronously.

PR #86, opened 2020-04-17
MetadataViewer and DatasetVisualizer are now a single switchable panel

loichuder:
Here is the first version:
image

I think some improvements are to be made on the styling so I open the PR early for discussion.

The reasons for the aspect of this first version are the following: I wanted the `roleToggler` to integrate with the rest of the viewer with the same background color. While this poses no problem in `Visualize` mode, in `Inspect` mode, the roleToggle merges visually with the first header `Link Info`. I added a border then but this is not extremely elegant...

### DIFF
--- a/src/h5web/App.module.css
+++ b/src/h5web/App.module.css
@@ -8,16 +8,45 @@
   background-color: var(--primary-bg);
 }
 
-.dataVisualizer {
+.mainArea {
   display: flex;
-  scrollbar-width: thin;
+  flex-direction: column;
 }
 
-.metadataViewer {
-  overflow: scroll !important;
-  scrollbar-width: thin;
+.toolbar {
+  display: flex;
+  justify-content: center;
+  font-size: 1.2rem;
+  background-color: var(--secondary-bg);
+  padding: 0.5rem;
 }
 
-.empty {
-  padding: 1rem;
+.roleToggler {
+  display: flex;
+  font-size: 1.2rem;
+  border-radius: 10px;
+  z-index: 1;
+  overflow: hidden;
+}
+
+.btn {
+  composes: btn-clean from global;
+  display: flex;
+  align-items: center;
+  padding: 0.375rem 1rem;
+  background-color: var(--secondary-light-bg);
+}
+
+.btn:hover,
+.btn[aria-selected='true'] {
+  transition-duration: 0.2s, 0.2s;
+}
+
+.btn:hover {
+  background-color: var(--secondary-light);
+}
+
+.btn[aria-selected='true'] {
+  background-color: var(--secondary);
+  font-weight: bold;
 }
--- a/src/h5web/App.tsx
+++ b/src/h5web/App.tsx
@@ -9,9 +9,15 @@ import styles from './App.module.css';
 import { isDataset } from './providers/utils';
 import { useEntity } from './providers/hooks';
 
+enum Role {
+  Inspect,
+  Display,
+}
+
 function App(): JSX.Element {
   const [selectedLink, setSelectedLink] = useState<HDF5Link>();
   const [selectedDataset, setSelectedDataset] = useState<HDF5Dataset>();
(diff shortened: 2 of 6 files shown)


src/h5web/App.tsx
```diff
@@ -32,32 +38,40 @@ function App(): JSX.Element {
...
-              {selectedLink ? (
-                <MetadataViewer
-                  key={JSON.stringify(selectedLink)}
-                  link={selectedLink}
-                  entity={selectedEntity}
-                />
-              ) : (
-                <div className={styles.empty}>
-                  <p>No entity selected.</p>
-                </div>
```
loichuder, line 57: The logic for displaying an empty div was moved inside `MetaViewer`. This was easier to implement the role toggler and consistent with the behaviour of `DatasetVisualizer`

src/h5web/App.tsx
```diff
@@ -9,9 +9,15 @@ import styles from './App.module.css';
 import { isDataset } from './providers/utils';
 import { useEntity } from './providers/hooks';
 
+enum Role {
+  Inspect,
+  Visualize,
+}
```
loichuder, line 15: I found clearer to use an `enum` here rather than a simple boolean.

src/styles/vars.css
```diff
@@ -1,6 +1,8 @@
 :root {
   --black: #020402;
   --primary: #c0da74;
+  --primary-dark: #9aae5d;
+  --primary-darker: #6a7840;
```
loichuder, line 5: Added some colors that could help with the styling

src/h5web/App.module.css
```diff
@@ -8,16 +8,35 @@
   background-color: var(--primary-bg);
 }
 
-.dataVisualizer {
+.mainArea {
   display: flex;
-  scrollbar-width: thin;
+  flex-direction: column;
 }
```
loichuder, line 14: `dataVisualizer` styling was moved inside its CSS module

src/h5web/App.module.css
```diff
@@ -8,16 +8,35 @@
...
 
-.metadataViewer {
-  overflow: scroll !important;
-  scrollbar-width: thin;
+.roleToggler {
+  display: flex;
+  justify-content: center;
```
loichuder, line 18: `Inspect` and `Visualize` buttons are centered for now but will be put to `flex-end` when adding the breadcrumbs

src/h5web/App.tsx
```diff
@@ -32,32 +38,40 @@ function App(): JSX.Element {
...
-                  entity={selectedEntity}
-                />
-              ) : (
-                <div className={styles.empty}>
-                  <p>No entity selected.</p>
-                </div>
-              )}
-            </ReflexElement>
-          </ReflexContainer>
+              Visualize
```
loichuder, line 63: `Visualize` could be changed in `Display`

src/h5web/App.tsx
axelboc: I was wondering about this. It's quite clear and scalable with an enum, but since it's in App, I feel like the shorter the implementation, the better. Also, the fact that the display logic in the JSX checks only one of the roles and doesn't care about the other one makes me think that a boolean would be more appropriate.

That being said, if you feel strongly about an enum, I don't mind it.

src/h5web/App.tsx
axelboc: Oh why not

src/h5web/App.tsx
loichuder: I agree for the display logic but it makes more sense for the rest. I tried first with a boolean but it felt awkward.

src/h5web/App.tsx
loichuder: Changed.

loichuder:
image

New style :arrow_up:

src/h5web/App.tsx
loichuder: Done.

PR #89, opened 2020-04-20
Axes now follows correctly when panning and zooming

### LINKED ISSUE #62: [Heatmap] Update axis on zoom/pan

loichuder:
Fixes #62

### DIFF
--- a/src/h5web/visualizations/heatmap/AxisGrid.tsx
+++ b/src/h5web/visualizations/heatmap/AxisGrid.tsx
@@ -1,17 +1,58 @@
-import React from 'react';
-import { Dom, useThree } from 'react-three-fiber';
+import React, { useState } from 'react';
+import { Dom, useThree, useFrame } from 'react-three-fiber';
+import { scaleLinear } from 'd3-scale';
 import IndexAxis from './IndexAxis';
 import styles from './HeatmapVis.module.css';
+import { Domain } from './store';
+
+interface AxisDomains {
+  left: Domain;
+  bottom: Domain;
+}
 
 interface Props {
   dims: [number, number];
+  axisOffsets: [number, number];
 }
 
 function AxisGrid(props: Props): JSX.Element {
-  const { dims } = props;
+  const { dims, axisOffsets } = props;
+  const [rows, cols] = dims;
+  const [leftAxisWidth, bottomAxisHeight] = axisOffsets;
 
-  const { size } = useThree();
+  const { camera, size } = useThree();
   const { width, height } = size;
+
+  const [domains, setDomains] = useState<AxisDomains>({
+    left: [0, rows],
+    bottom: [0, cols],
+  });
+
+  // Axis bounds in R3F camera coordinates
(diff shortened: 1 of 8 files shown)


src/h5web/visualizations/heatmap/AxisGrid.tsx
```diff
@@ -0,0 +1,74 @@
...
+  useFrame(() => {
+    const { position, zoom } = camera;
+    const factor = 1 / zoom;
+
+    const leftScale = scaleLinear()
+      .domain([-height / 2 + bottomAxisHeight, height / 2])
+      .range([0, rows]);
+    const bottomScale = scaleLinear()
+      .domain([-width / 2 + leftAxisWidth, width / 2])
+      .range([0, cols]);
```
loichuder, line 39: I have done here with d3's `scale` but I can manage to implement it mathematically if we think it may improve the performance

src/h5web/visualizations/heatmap/AxisGrid.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const bottomScale = scaleLinear()
+    .domain(bottomBounds)
+    .range([0, cols]);
+
+  useFrame(() => {
+    const { position, zoom } = camera;
+
+    // Scale bounds according to zoom and shift by camera position
+    setDomains({
+      left: leftBounds.map(b => leftScale(b / zoom + position.y)) as [
```
axelboc, line 53: Maybe introduce a type `Domain` for `[number, number]` to avoid this sort of wrapping. We're using it in a lot of places, so I think it makes sense.

src/h5web/visualizations/heatmap/AxisGrid.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { camera, size } = useThree();
+  const { width, height } = size;
+
+  const [domains, setDomains] = useState<AxisDomains>({
+    left: [0, rows],
+    bottom: [0, cols],
+  });
+
+  // Axis bounds in R3F camera coordinates
+  const leftBounds: [number, number] = [
```
axelboc, line 31: `leftAxisBounds`, `leftAxisScale`, etc. ?

src/h5web/visualizations/heatmap/AxisGrid.tsx
axelboc: Et `bound` ou `val` au lieu de `b` ;)

src/h5web/visualizations/heatmap/AxisGrid.tsx
```diff
@@ -0,0 +1,82 @@
...
+    -height / 2 + bottomAxisHeight,
+    height / 2,
+  ];
```
axelboc, line 34: You can put `as const` here instead of typing the constant.

loichuder: I wish but Typescript complains when setting it as a `d3-scale` domain...

PR #92, opened 2020-04-21
Implement orthonormal option for heatmap

### LINKED ISSUE #70: [Heatmap] Orthonormal coordinates
- Allow switching to a orthonormal coordinate system where the x and y axis share the same unit of length.
- Let the heatmap expand to fit its container while maintaining the resulting aspect ratio (e.g. 1000x800 => 0.8). `object-fit: contain` might help with this.

axelboc:
Fixes #70 

I added a boolean to the state (`true` by default), but I didn't add a button to change it as this will go in the soon-to-be toolbar.

The Heatmap takes all available width or height depending on the size ratio of the dataset.

image

image

### DIFF
--- a/src/h5web/visualizations/heatmap/hooks.ts
+++ b/src/h5web/visualizations/heatmap/hooks.ts
@@ -2,8 +2,51 @@ import React, { useCallback, useRef, useEffect } from 'react';
 import { Vector3 } from 'three';
 import { ReactThreeFiber, PointerEvent, useThree } from 'react-three-fiber';
 import { clamp } from 'lodash-es';
+import { useMeasure } from 'react-use';
+import { useHeatmapState } from './store';
 
 const ZOOM_FACTOR = 0.95;
+
+export function useHeatmapSize<T>(
+  dims: [number, number],
+  axisOffset: [number, number]
+): [(elem: T) => void, { width: number; height: number } | undefined] {
+  const { keepAspectRatio } = useHeatmapState();
+  const [wrapperRef, { width, height }] = useMeasure();
+
+  if (width === 0 && height === 0) {
+    return [wrapperRef, undefined];
+  }
+
+  if (!keepAspectRatio) {
+    return [wrapperRef, { width, height }];
+  }
+
+  const [rows, cols] = dims;
+  const aspectRatio = rows / cols;
+
+  const [leftAxisWidth, bottomAxisHeight] = axisOffset;
+  const availableWidth = width - leftAxisWidth;
+  const availableHeight = height - bottomAxisHeight;
+
+  // Determine how to compute mesh size to fit available space while maintaining aspect ratio
+  const shouldAdjustWidth = availableWidth >= availableHeight * aspectRatio;
+
+  const meshWidth = shouldAdjustWidth
+    ? availableHeight * aspectRatio
+    : availableWidth;
+  const meshHeight = shouldAdjustWidth
+    ? availableHeight
+    : availableWidth / aspectRatio;
+
+  return [
+    wrapperRef,
+    {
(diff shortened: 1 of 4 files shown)


src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -1,10 +1,53 @@
...
+import { useMeasure } from 'react-use';
+import { useHeatmapState } from './store';
 
 const ZOOM_FACTOR = 0.95;
 
+export function useHeatmapSize<T>(
+  dims: [number, number],
+  axisOffset: [number, number]
+): [(elem: T) => void, { width: number; height: number } | undefined] {
+  const { hasOrthoCoords } = useHeatmapState();
```
loichuder, line 14: I find `hasOrthoCoords` is not that clear. What about `keepImageAspectRatio` or `keepAspectRatio` ?

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -1,10 +1,53 @@
...
+  if (!hasOrthoCoords) {
+    return [wrapperRef, { width, height }];
+  }
+
+  const [rows, cols] = dims;
+  const dimsRatio = rows / cols;
+
+  const [leftAxisWidth, bottomAxisHeight] = axisOffset;
+  const actualWidth = width - leftAxisWidth;
+  const actualHeight = height - bottomAxisHeight;
```
loichuder, line 30: `imageWidth` | `imageHeight` ? or `meshWidth` ?

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -1,10 +1,53 @@
...
+
+  const shouldAdjustWidth = actualWidth >= actualHeight * dimsRatio;
```
loichuder, line 32: I would add a comment here explaining the need for this boolean

src/h5web/visualizations/heatmap/hooks.ts
axelboc: Oh I like `meshWidth`! -- I'm not a fan of the term `image`, as it's too much tied to JPEGs/PNGs/``... in my head.

axelboc: Ah actually, maybe not... I initially had `availableWidth` and `availableHeight`. I think I might revert to that, as it conveys the fact that I'm getting space available for the mesh, not the size of the mesh itself. Perhaps `targetWidth` and `targetHeight` should be the ones renamed to `meshWidth` and `meshHeight`?

loichuder: I wholly agree with `availableWidth` and renaming `target` to `mesh` :+1:

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -1,10 +1,53 @@
...
 import { clamp } from 'lodash-es';
+import { useMeasure } from 'react-use';
+import { useHeatmapState } from './store';
 
 const ZOOM_FACTOR = 0.95;
 
+export function useHeatmapSize<T>(
+  dims: [number, number],
+  axisOffset: [number, number]
+): [(elem: T) => void, { width: number; height: number } | undefined] {
```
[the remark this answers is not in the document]
loichuder: :+1: It is just a shame that it is difficult to convey meaning with the hook naming convention

src/h5web/visualizations/heatmap/hooks.ts
axelboc: Phew, I struggled with this one. Hope it is clear now.

PR #97, opened 2020-04-22
Simplify Heatmap rendering and interactions with WebGL viewport

### LINKED ISSUE #96: [Heatmap] Investigate viewports in ThreeJS
Check if this could simplify some calculations around zooming and panning.

axelboc:
Fix #96 

Comments coming up...

### DIFF
--- a/src/h5web/BreadcrumbsBar.tsx
+++ b/src/h5web/BreadcrumbsBar.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import { FiSidebar, FiChevronRight } from 'react-icons/fi';
 import styles from './BreadcrumbsBar.module.css';
 import { TreeNode } from './explorer/models';
@@ -38,12 +38,10 @@ function BreadcrumbsBar(props: Props): JSX.Element {
       {selectedNode && (
         <div className={styles.breadCrumbs}>
           {selectedNode?.parents.slice(firstParentIndex).map(member => (
-            <>
-              <span className={styles.crumb} key={(member as HDF5HardLink).id}>
-                {member.title}
-              </span>
+            <Fragment key={(member as HDF5HardLink).id}>
+              <span className={styles.crumb}>{member.title}</span>
               <FiChevronRight />
-            </>
+            </Fragment>
           ))}
           <span className={styles.crumb} data-current>
             {selectedNode.data.title}
--- a/src/h5web/visualizations/heatmap/AxisGrid.tsx
+++ b/src/h5web/visualizations/heatmap/AxisGrid.tsx
@@ -18,7 +18,6 @@ interface Props {
 function AxisGrid(props: Props): JSX.Element {
   const { dims, axisOffsets } = props;
   const [rows, cols] = dims;
-  const [leftAxisWidth, bottomAxisHeight] = axisOffsets;
 
   const { camera, size } = useThree();
   const { width, height } = size;
@@ -29,8 +28,8 @@ function AxisGrid(props: Props): JSX.Element {
   });
 
   // Axis bounds in R3F camera coordinates
-  const leftAxisBounds: Domain = [-height / 2 + bottomAxisHeight, height / 2];
-  const bottomAxisBounds: Domain = [-width / 2 + leftAxisWidth, width / 2];
+  const leftAxisBounds = [-height / 2, height / 2];
+  const bottomAxisBounds = [-width / 2, width / 2];
 
   // Scales R3F camera coordinates to axis bounds
   const leftAxisScale = scaleLinear()
@@ -56,7 +55,15 @@ function AxisGrid(props: Props): JSX.Element {
 
   return (
     <Dom>
-      <div className={styles.axisGrid} style={{ width, height }}>
+      <div
+        className={styles.axisGrid}
+        style={{
+          width,
+          height,
+          gridTemplateColumns: `${axisOffsets[0]}px 1fr`,
+          gridTemplateRows: `1fr ${axisOffsets[1]}px`,
+        }}
+      >
         <IndexAxis
           className={styles.leftAxisCell}
           orientation="left"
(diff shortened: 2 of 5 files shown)


src/h5web/visualizations/heatmap/Mesh.tsx
```diff
@@ -12,28 +12,31 @@ interface Props {
...
+
+  useFrame(() => {
+    // Set viewport of WebGL renderer to account for axis
+    gl.setViewport(
+      leftAxisWidth,
+      bottomAxisHeight,
+      width - leftAxisWidth,
+      height - bottomAxisHeight
+    );
+  });
```
axelboc, line 27: This is where the magic happens. With this call, all the computations relating to the camera can now be done as if the axis didn't exist. In other words, the "bounds" of the heatmap in camera space match the size of the canvas. This simplifies:
- the clamping computations in `usePanZoom()`
- the axis domains computations in `AxisGrid`
- the positioning and sizing of the mesh in `Mesh`.

axelboc: It also means that the axis don't have to have a white background, as nothing is rendered under them. Another interesting side effect is that the axis can now be dragged to pan the heatmap.

src/h5web/visualizations/heatmap/Mesh.tsx
```diff
@@ -12,28 +12,31 @@ interface Props {
...
     });
   }, [dims, textureData]);
 
-  const [leftAxisWidth, bottomAxisHeight] = axisOffsets;
-  const pointerHandlers = usePanZoom(leftAxisWidth, bottomAxisHeight);
+  const pointerHandlers = usePanZoom();
 
   return (
-    <mesh
-      position={[leftAxisWidth / 2, bottomAxisHeight / 2, 0]}
```
axelboc, line 29: No need to position the mesh, it can be just be centred at [0,0,0].

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -145,7 +138,7 @@ export function usePanZoom(
   useEffect(() => {
     // Move camera on resize to stay within mesh bounds
     moveCameraTo(camera.position.x, camera.position.y);
-  }, [camera, moveCameraTo, size]); // `size` is key here
+  }, [camera, moveCameraTo]);
```
axelboc, line 141: `moveCameraTo` actually changes when `size` changes, so we're good with not having `size` as a dependency here.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -19,37 +19,29 @@
 /* R3F's <Dom /> element*/
 .heatmap > div > div {
   transform: none !important;
-  pointer-events: none;
+  user-select: none; /* prevent text selection of tick labels when dragging */
```
axelboc, line 22: The axis can now be dragged to pan, so text selection is more problematic. I don't think people will need to select and copy tick labels, though.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -19,37 +19,29 @@
...
 }
 
 .axisGrid {
   display: grid;
   grid-template-areas:
     'left-axis .'
     'empty bottom-axis';
-  grid-template-columns: 72px 1fr;
-  grid-template-rows: 1fr 36px;
```
axelboc, line 31: I now set these dynamically based on `AXIS_OFFSETS`. Not necessary, it just avoids potential future problems with duplicating the values.

src/h5web/visualizations/heatmap/AxisGrid.tsx
```diff
@@ -56,7 +55,15 @@ function AxisGrid(props: Props): JSX.Element {
...
   return (
     <Dom>
-      <div className={styles.axisGrid} style={{ width, height }}>
+      <div
+        className={styles.axisGrid}
+        style={{
+          width,
+          height,
+          gridTemplateColumns: `${axisOffsets[0]}px 1fr`,
+          gridTemplateRows: `1fr ${axisOffsets[1]}px`,
```
axelboc, line 64: cf. comment in `HeatmapVis.module.css`

src/h5web/BreadcrumbsBar.tsx
```diff
@@ -38,12 +38,10 @@ function BreadcrumbsBar(props: Props): JSX.Element {
...
           {selectedNode?.parents.slice(firstParentIndex).map(member => (
-            <>
-              <span className={styles.crumb} key={(member as HDF5HardLink).id}>
-                {member.title}
-              </span>
+            <Fragment key={(member as HDF5HardLink).id}>
+              <span className={styles.crumb}>{member.title}</span>
               <FiChevronRight />
-            </>
+            </Fragment>
```
axelboc, line 44: Just a quick fix for a warning in the console. The `key` has to go on the fragment.

src/h5web/visualizations/heatmap/Mesh.tsx
loichuder: :sparkles: I think the comment would be much clearer if you add the info that *all the computations relating to the camera can now be done as if the axis didn't exist.*

src/h5web/visualizations/heatmap/hooks.ts
loichuder: What makes `moveCameraTo` change on resize ?

src/h5web/visualizations/heatmap/HeatmapVis.module.css
loichuder: We will see in the future. Good to have the comment as a reminder.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
loichuder: Yes I was worried about that. Nice :+1:

src/h5web/visualizations/heatmap/hooks.ts
axelboc: It's created with `useCallback()` and `size` as one of the dependencies.

src/h5web/visualizations/heatmap/Mesh.tsx
axelboc: I've updated the comment 💯

PR #98, opened 2020-04-23
Implements visualisation toolbar with zustand

loichuder:


src/h5web/dataset-visualizer/VisDisplay.tsx
```diff
@@ -45,12 +44,10 @@ function VisDisplay(props: Props): JSX.Element {
...
-      <HeatmapStore.Provider>
-        <HeatmapVis
-          dims={(dataset.shape as HDF5SimpleShape).dims as [number, number]}
-          data={value}
-        />
-      </HeatmapStore.Provider>
+      <HeatmapVis
+        dims={(dataset.shape as HDF5SimpleShape).dims as [number, number]}
+        data={value}
+      />
```
loichuder, line 50: `zustand` state is hook-based and does not need a Provider

src/h5web/dataset-visualizer/VisToolbar.tsx
```diff
@@ -0,0 +1,18 @@
...
+function VisToolbar(props: Props): JSX.Element {
+  const { vis } = props;
+
+  if (vis === Vis.Heatmap) {
+    return <HeatmapToolbar />;
+  }
+  return <></>;
+}
+
+export default VisToolbar;
```
loichuder, line 18: This is the new component that will display the appropriate toolbar. All interactions will take place inside the toolbar of the respective visualisation.

src/h5web/dataset-visualizer/VisToolbar.tsx
```diff
@@ -0,0 +1,18 @@
...
+import HeatmapToolbar from '../visualizations/heatmap/HeatmapToolbar';
+
+interface Props {
+  vis?: Vis;
+}
+
+function VisToolbar(props: Props): JSX.Element {
+  const { vis } = props;
+
+  if (vis === Vis.Heatmap) {
```
loichuder, line 12: For now, as we have only one supported vis, I kept this simple logic. We can move to the `Record` once we add more visus.

src/h5web/visualizations/heatmap/AspectRatioToggler.tsx
```diff
@@ -0,0 +1,28 @@
...
+        toggleAspectRatio();
+      }}
+    >
+      {keepAspectRatio ? <FiToggleRight /> : <FiToggleLeft />}
+      Keep aspect ratio
+    </button>
+  );
+}
+
+export default AspectRatioToggler;
```
loichuder, line 28: I put the aspect ratio toggler in this simple component to keep it as decoupled as possible. This way, we can move it around and inline it where necessary when the aspect of the toolbar is settled.

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -3,10 +3,15 @@ import { AxisRight } from '@vx/axis';
...
+import {
+  useHeatmapStore,
+  dataScaleSelector,
+  interpolatorSelector,
+} from './store';
 
 function ColorBar(): JSX.Element {
-  const { dataScale, interpolator } = useHeatmapState();
+  const interpolator = useHeatmapStore(interpolatorSelector);
+  const dataScale = useHeatmapStore(dataScaleSelector);
```
loichuder, line 14: Properties that were `computed` in easy-peasy needs to be computed from selectors in zustand. There is also the custom hook possibility.

src/h5web/visualizations/heatmap/ColorMapSelector.tsx
```diff
@@ -22,38 +19,34 @@ const DropdownIndicator = (props: any): JSX.Element => {
...
+  const setColorMap = useHeatmapStore(state => state.setColorMap);
 
   const gradientStyles = {
     option: (
-      styles: CSSProperties,
+      selectStyles: CSSProperties,
       { data: { label } }: { data: { label: ColorMap } }
     ) => ({
-      ...styles,
+      ...selectStyles,
```
loichuder, line 31: Renamed to avoid shadowing of `styles` imported from the CSS module

src/h5web/visualizations/heatmap/HeatmapToolbar.module.css
```diff
@@ -0,0 +1,32 @@
...
+  align-items: center;
+  justify-content: left;
+}
+
+.scaleTypeToggler:hover,
+.scaleTypeToggler[aria-checked='true'],
+.aspectRatioToggler:hover,
+.aspectRatioToggler[aria-checked='true'] {
+  background-color: var(--primary-dark-bg);
+}
```
loichuder, line 28: This needs perhaps some CSS tidying-up :sweat_smile:

src/h5web/visualizations/heatmap/HeatmapToolbar.module.css
```diff
@@ -0,0 +1,32 @@
...
+
+.colorMapSelector {
+  min-width: 8rem;
+}
```
loichuder, line 32: To be addressed when changing the aspect of the colormap selector.

src/h5web/visualizations/heatmap/ColorBar.tsx
axelboc: In terms of naming, just thinking we could also go for `selectInterpolator`. I'm fine with either.

src/h5web/visualizations/heatmap/ColorMapSelector.tsx
```diff
@@ -22,38 +19,34 @@ const DropdownIndicator = (props: any): JSX.Element => {
...
 const colorMapOptions = Object.keys(INTERPOLATORS).map(label => ({ label }));
 
-function ColorMapSelector(props: Props): JSX.Element {
-  const { className } = props;
-
-  const { colorMap } = useHeatmapState();
-  const { setColorMap } = useHeatmapActions();
+function ColorMapSelector(): JSX.Element {
+  const colorMap = useHeatmapStore(state => state.colorMap);
+  const setColorMap = useHeatmapStore(state => state.setColorMap);
```
axelboc, line 24: You can write a selector that returns more than one thing: https://github.com/react-spring/zustand#selecting-multiple-state-slices

... on second thought, the syntax is pretty heavy, so maybe not. :(

src/h5web/visualizations/heatmap/HeatmapToolbar.module.css
```diff
@@ -0,0 +1,32 @@
+.toolbar {
+  display: flex;
+  padding: 0 0.5rem;
+  align-items: stretch;
```
axelboc, line 4: This is the default, so don't need it.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -17,11 +18,14 @@ function DatasetVisualizer(props: Props): JSX.Element {
 
   return (
     <div className={styles.visualizer}>
-      <VisSelector
-        activeVis={activeVis}
-        choices={supportedVis}
-        onChange={setActiveVis}
-      />
+      <div className={styles.toolbar}>
```
axelboc, line 21: Just thinking that the term you used for the name of the branch may be better suited: `visBar`. This way we can distinguish the three elements: the parent `visBar`, the `VisSelector` and the `VisToolbar`.

src/h5web/visualizations/heatmap/LogScaleToggler.tsx
```diff
@@ -1,23 +1,25 @@
...
+import styles from './HeatmapToolbar.module.css';
+import { useHeatmapStore } from './store';
 
 function LogScaleToggler(): JSX.Element {
-  const { hasLogScale } = useHeatmapState();
-  const { toggleLogScale } = useHeatmapActions();
+  const [hasLogScale, toggleLogScale] = useHeatmapStore(state => [
+    state.hasLogScale,
+    state.toggleLogScale,
+  ]);
```
axelboc, line 10: if you use this syntax, you need to pass `shallow` so the return value is compared shallowly https://github.com/react-spring/zustand#selecting-multiple-state-slices

src/h5web/visualizations/heatmap/LogScaleToggler.tsx
```diff
@@ -1,23 +1,25 @@
...
 
   return (
     <button
-      className={styles.scaleTypeSelector}
+      className={styles.scaleTypeToggler}
```
axelboc, line 14: `logScaleToggler` ?

src/h5web/visualizations/heatmap/store.ts
```diff
@@ -27,63 +19,58 @@ export type DataScale =
...
-export const HeatmapStore = createContextStore<HeatmapState>({
+export const [useHeatmapStore] = create<HeatmapState>(set => ({
   domain: undefined,
-  findDomain: action((state, values) => {
-    const [min, max] = extent(values);
-    state.domain =
-      min === undefined || max === undefined ? undefined : [min, max];
-  }),
+  findDomain: (values: number[]) =>
+    set(() => {
```
axelboc, line 39: if you don't need the state to update the domain, so you can write:

```ts
  findDomain: (values: number[]) => {
    const [min, max] = extent(values);
    set({ domain: min === undefined || max === undefined ? undefined : [min, max] });
  }
```

src/h5web/visualizations/heatmap/AspectRatioToggler.tsx
axelboc, line 28: Just thinking that it may be worth implementing a generic `Toggler` component that takes a `className`, a `label`, a `value` and an `onChange` prop. This would localise selections from the store in the toolbar component (single hook call with shallow comparison)

src/h5web/visualizations/heatmap/HeatmapToolbar.module.css
axelboc: I have to admit I'm not a huge fan of having the two togglers stacked up. Are you worried about horizontal space? Perhaps we could shorten the labels, like `Keep ratio`, and `Log scale` ?

loichuder: No, I am not worried. It was just a preliminary layout for discussion. I agree for `Log Scale` but I fear that `Keep ratio` is not as meaningful. Still, let's roll with that and see how it goes in the future.

src/h5web/visualizations/heatmap/LogScaleToggler.tsx
loichuder: :fishing_pole_and_fish:

PR #99, opened 2020-04-27
Position axis outside of canvas instead of setting viewport

axelboc:
I think it will make more sense if you look at the files from bottom to top... and if you inspect the DOM yourself.

### DIFF
--- a/src/h5web/visualizations/heatmap/AxisGrid.tsx
+++ b/src/h5web/visualizations/heatmap/AxisGrid.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { Dom, useThree, useFrame } from 'react-three-fiber';
+import { useThree, useFrame, Dom } from 'react-three-fiber';
 import { scaleLinear } from 'd3-scale';
 import IndexAxis from './IndexAxis';
 import styles from './HeatmapVis.module.css';
@@ -18,6 +18,7 @@ interface Props {
 function AxisGrid(props: Props): JSX.Element {
   const { dims, axisOffsets } = props;
   const [rows, cols] = dims;
+  const [leftAxisWidth, bottomAxisHeight] = axisOffsets;
 
   const { camera, size } = useThree();
   const { width, height } = size;
@@ -54,16 +55,18 @@ function AxisGrid(props: Props): JSX.Element {
   });
 
   return (
-    <Dom>
-      <div
-        className={styles.axisGrid}
-        style={{
-          width,
-          height,
-          gridTemplateColumns: `${axisOffsets[0]}px 1fr`,
-          gridTemplateRows: `1fr ${axisOffsets[1]}px`,
-        }}
-      >
+    <Dom
+      className={styles.axisGrid}
+      style={{
+        // Take over space reserved for axis by `useHeatmapStyles` hook
+        width: width + leftAxisWidth,
+        height: height + bottomAxisHeight,
+        left: -leftAxisWidth,
+        gridTemplateColumns: `${leftAxisWidth}px 1fr`,
+        gridTemplateRows: `1fr ${bottomAxisHeight}px`,
+      }}
+    >
+      <>
         <IndexAxis
           className={styles.leftAxisCell}
           orientation="left"
@@ -74,7 +77,7 @@ function AxisGrid(props: Props): JSX.Element {
           orientation="bottom"
           domain={domains.bottom}
         />
-      </div>
+      </>
     </Dom>
   );
 }
--- a/src/h5web/visualizations/heatmap/HeatmapVis.module.css
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.module.css
@@ -14,24 +14,29 @@
 
 .heatmap {
   flex: none;
+  overflow: hidden;
 }
 
-/* R3F's <Dom /> element*/
-.heatmap > div > div {
+/* R3F's root element */
+.canvasWrapper {
+  overflow: visible !important; /* show child axis grid, which is bigger than canvas */
+}
+
+/* R3F's <Dom /> element */
+.canvasWrapper > div {
   transform: none !important;
-  user-select: none; /* prevent text selection of tick labels when dragging */
+  /* - Prevent panning/zooming from axis when heatmap is zoomed in.
+   * - Prevent text selection of tick labels when panning with touchpad. */
+  pointer-events: none;
 }
 
 .axisGrid {
+  position: absolute;
+  top: 0;
   display: grid;
   grid-template-areas:
     'left-axis .'
-    'empty bottom-axis';
-}
-
-.axisGrid::before {
-  grid-area: empty;
-  content: '';
+    '. bottom-axis';
 }
 
 .leftAxisCell {
--- a/src/h5web/visualizations/heatmap/HeatmapVis.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.tsx
@@ -1,12 +1,12 @@
 import { Canvas } from 'react-three-fiber';
 import React, { useEffect, useMemo } from 'react';
-import Mesh from './Mesh';
 import { computeTextureData } from './utils';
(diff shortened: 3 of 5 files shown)


src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -14,24 +14,32 @@
 
 .heatmap {
   flex: none;
+  box-sizing: content-box;
```
axelboc, line 17: This avoids having to add the axis offsets to the width and height.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -42,8 +42,10 @@ export function useHeatmapSize<T>(
   return [
     wrapperRef,
     {
-      width: meshWidth + leftAxisWidth,
-      height: meshHeight + bottomAxisHeight,
+      width: meshWidth,
+      height: meshHeight,
+      paddingBottom: bottomAxisHeight,
+      paddingLeft: leftAxisWidth,
```
axelboc, line 48: This is the most important bit: the width and height of the `heatmap` element in `HeatmapVis` now match the width and height of the canvas and of the mesh. It uses padding to leave space for the axis. Not that the CSS adds `box-sizing: content-box`, so the total width of the `heatmap` element ends up being `width + leftAxisWidth`.

src/h5web/visualizations/heatmap/Mesh.tsx
```diff
@@ -1,33 +1,18 @@
...
-     * Set viewport of WebGL renderer to allow camera-related computations to be performed
-     * as if axis didn't exist and mesh filled entire canvas.
-     */
-    gl.setViewport(
-      leftAxisWidth,
-      bottomAxisHeight,
-      width - leftAxisWidth,
-      height - bottomAxisHeight
-    );
-  });
```
axelboc, line 30: No more viewport. The mesh is rendered on the entire canvas.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -38,18 +38,16 @@ function HeatmapVis(props: Props): JSX.Element {
   return (
     <div className={styles.root}>
       <div ref={mapAreaRef} className={styles.mapArea}>
-        {heatmapSize && (
-          <div className={styles.heatmap} style={heatmapSize}>
-            <Canvas orthographic invalidateFrameloop>
+        {heatmapStyles && (
+          <div className={styles.heatmap} style={heatmapStyles}>
+            <Canvas
+              className={styles.canvasWrapper}
```
axelboc, line 44: Just adding a classname here to simplify selectors in CSS (`.heatmap > div` becomes `.canvasWrapper`)

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -38,18 +38,16 @@ function HeatmapVis(props: Props): JSX.Element {
   return (
     <div className={styles.root}>
       <div ref={mapAreaRef} className={styles.mapArea}>
-        {heatmapSize && (
-          <div className={styles.heatmap} style={heatmapSize}>
-            <Canvas orthographic invalidateFrameloop>
+        {heatmapStyles && (
+          <div className={styles.heatmap} style={heatmapStyles}>
```
axelboc, line 42: Note that the styles here now include some padding for the axis.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -14,24 +14,32 @@
...
+}
+
+/* R3F's root element */
+.canvasWrapper {
+  overflow: visible !important; /* show axis */
```
axelboc, line 22: The `canvasWrapper` no longer includes the size of the axis, so we have to make overflow visible. This has a side-effect which requires setting `pointer-events: none` below.

src/h5web/visualizations/heatmap/AxisGrid.tsx
```diff
@@ -54,16 +55,17 @@ function AxisGrid(props: Props): JSX.Element {
...
-          gridTemplateColumns: `${axisOffsets[0]}px 1fr`,
-          gridTemplateRows: `1fr ${axisOffsets[1]}px`,
-        }}
-      >
+    <Dom
+      className={styles.axisGrid}
+      style={{
+        width: width + leftAxisWidth,
+        height: height + bottomAxisHeight,
+        left: -leftAxisWidth,
```
axelboc, line 63: The axis grid is now inside the canvas, which no longer includes the size of the axis, so we need to increase the grid's size and offset it to give the axis some space and position them properly.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -1,14 +1,14 @@
 import { Canvas } from 'react-three-fiber';
 import React, { useEffect, useMemo } from 'react';
-import Mesh from './Mesh';
 import { computeTextureData } from './utils';
 import styles from './HeatmapVis.module.css';
 import ColorBar from './ColorBar';
 import ColorMapSelector from './ColorMapSelector';
 import LogScaleToggler from './LogScaleToggler';
 import { useHeatmapState, useHeatmapActions } from './store';
+import { useHeatmapSize as useHeatmapStyles } from './hooks';
```
loichuder, line 9: You may want to rename `useHeatmapSize` in `useHeatmapStyles` in the `hooks` file as well

src/h5web/visualizations/heatmap/hooks.ts
axelboc: I moved the `content-box` style here even though it's static, so the code is more self-documenting.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
axelboc: I've reviewed the comment here.

src/h5web/visualizations/heatmap/AxisGrid.tsx
axelboc: I've added a comment here.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -16,22 +16,26 @@
   flex: none;
 }
 
-/* R3F's <Dom /> element*/
-.heatmap > div > div {
+/* R3F's root element */
+.canvasWrapper {
+  overflow: visible !important; /* show child axis grid, which is bigger than canvas */
```
axelboc, line 21: clearer comment here

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -29,21 +29,24 @@ export function useHeatmapSize<T>(
...
-  // Determine how to compute mesh size to fit available space while maintaining aspect ratio
+  // Determine how to compute canvas size to fit available space while maintaining aspect ratio
   const shouldAdjustWidth = availableWidth >= availableHeight * aspectRatio;
 
-  const meshWidth = shouldAdjustWidth
+  const canvasWidth = shouldAdjustWidth
     ? availableHeight * aspectRatio
     : availableWidth;
-  const meshHeight = shouldAdjustWidth
+  const canvasHeight = shouldAdjustWidth
```
axelboc, line 38: I've renamed the variables here to make it clear that the styles are computed based on the size of the canvas.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -19,31 +21,41 @@ export function useHeatmapSize<T>(
...
-    return [wrapperRef, { width, height }];
+    return [
+      wrapperRef,
+      {
+        width,
+        height,
+        paddingBottom: bottomAxisHeight,
+        paddingLeft: leftAxisWidth,
+      },
+    ];
```
axelboc, line 32: J'avais oublié de mettre du padding ici...

PR #103, opened 2020-04-28
Implement heatmap tooltip

### LINKED ISSUE #48: [Heatmap] Tooltip on hover
Tooltip should include coordinates and value.

axelboc:
Fix #48 -- The tooltip:

- shows the data coordinates and value;
- moves to the other side of the cursor when reaching the bounds of the canvas (thanks to VX's `TooltipWithBounds`);
- disappears while panning.

image
image

In order to not couple the event handling logic for the tooltip with the zoom/pan logic, I render a second mesh behind the textured mesh to which I pass tooltip-specific pointer event listeners. I may refactor the zoom/pan code in a similar fashion so it can be re-used more easily in other visualizations.

### DIFF
--- a/src/h5web/visualizations/heatmap/HeatmapVis.module.css
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.module.css
@@ -31,7 +31,7 @@
 }
 
 .axisGrid {
-  position: absolute;
+  position: relative;
   top: 0;
   display: grid;
   grid-template-areas:
@@ -89,4 +89,16 @@
   font-size: inherit;
   text-anchor: inherit;
   dominant-baseline: inherit;
+}
+
+.tooltip {
+  color: inherit !important;
+  background-color: rgba(245, 251, 239, 0.7) !important;
+}
+
+.tooltipValue {
+  display: block;
+  font-weight: 600;
+  margin-top: 0.25rem;
+  font-size: 1.125em;
 }
--- a/src/h5web/visualizations/heatmap/Tooltip.tsx
+++ b/src/h5web/visualizations/heatmap/Tooltip.tsx
@@ -1,0 +1,89 @@
+import React, { ReactElement, useCallback } from 'react';
+import { Dom, PointerEvent, useThree } from 'react-three-fiber';
+import { TooltipWithBounds, useTooltip } from '@vx/tooltip';
+import { scaleLinear } from 'd3-scale';
+import { format } from 'd3-format';
+import styles from './HeatmapVis.module.css';
+
+interface Props {
+  data: number[][];
+  dims: [number, number];
+}
+
+function Tooltip(props: Props): ReactElement {
+  const { data, dims } = props;
+  const [rows, cols] = dims;
+
+  const { camera, size, intersect } = useThree();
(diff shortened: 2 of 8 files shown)


src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -87,8 +87,6 @@ export function usePanZoom(): ReactThreeFiber.Events {
 
   const onPointerDown = useCallback(
     (evt: PointerEvent) => {
-      evt.stopPropagation();
```
axelboc, line 90: I let these events through so they can be "heard" by the tooltip mesh.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -90,3 +90,15 @@
   text-anchor: inherit;
   dominant-baseline: inherit;
 }
+
+.tooltip {
+  color: inherit !important;
+  background-color: rgba(245, 251, 239, 0.7) !important;
```
loichuder, line 96: Why are the `!important` needed ?

src/h5web/visualizations/heatmap/Tooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  // Update tooltip when pointer moves
+  // When panning, events are handled and stopped by texture mesh and do not reach this mesh (which is behind)
+  const onPointerMove = useCallback(
+    (evt: PointerEvent) => {
+      const { zoom } = camera;
+      const projectedPoint = camera.worldToLocal(evt.unprojectedPoint.clone());
+
+      const xCoord = Math.floor(xCoordScale(evt.unprojectedPoint.y));
+      const yCoord = Math.floor(yCoordScale(evt.unprojectedPoint.x));
```
loichuder, line 45:
```suggestion
      const yCoord = Math.floor(yCoordScale(evt.unprojectedPoint.y));
```
and the other way round above ;)

src/h5web/visualizations/heatmap/Tooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+        {tooltipOpen && tooltipData ? (
+          <TooltipWithBounds
+            key={Math.random()}
+            className={styles.tooltip}
+            top={tooltipTop}
+            left={tooltipLeft}
+          >
+            {`x=${tooltipData[0]}, y=${tooltipData[1]}`}
+            <span className={styles.tooltipValue}>
+              {format('.3e')(data[tooltipData[0]][tooltipData[1]])}
```
loichuder, line 78: I think the format here should be consistent with the format of the colorbar axis

src/h5web/visualizations/heatmap/HeatmapVis.module.css
axelboc: It overrides the inline styles set by VX.

src/h5web/visualizations/heatmap/Tooltip.tsx
axelboc: Arf, silly me 🤦‍♂️

PR #105, opened 2020-04-29
Added slider to set custom bounds of the colormap

loichuder:
Based on [react-slider](https://github.com/zillow/react-slider/)


src/h5web/visualizations/heatmap/DomainSlider.module.css
```diff
@@ -0,0 +1,40 @@
...
+  font-size: 0.625rem;
+  height: 1.5rem;
+  min-width: 1.5rem;
+  padding: 0.125rem;
+  color: black;
+  cursor: grab;
+  background-color: var(--primary-bg);
+  border: 2px solid var(--secondary);
+  border-radius: 1.5em;
+}
```
loichuder, line 40: `react-slider` is shipped without any CSS so I introduce these.

src/h5web/visualizations/heatmap/DomainSlider.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { format } from 'd3-format';
+import styles from './DomainSlider.module.css';
+import { useHeatmapStore, Domain } from './store';
+
+const EXTEND_PERCENTAGE = 0.2;
+const MIN_DECIMALS = 1;
+
+function round(value: number, decimals: number): number {
+  return Number(`${Math.round(Number(`${value}e${decimals}`))}e-${decimals}`);
+}
```
loichuder, line 13: An function to round accurately numbers based on https://www.jacklmoore.com/notes/rounding-in-javascript/. May be moved in `utils` or replaced by https://lodash.com/docs/4.17.15#round.

src/h5web/visualizations/heatmap/DomainSlider.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import ReactSlider from 'react-slider';
+import shallow from 'zustand/shallow';
+import { format } from 'd3-format';
+import styles from './DomainSlider.module.css';
+import { useHeatmapStore, Domain } from './store';
+
+const EXTEND_PERCENTAGE = 0.2;
+const MIN_DECIMALS = 1;
```
loichuder, line 9: The floating point number troubles are strong in this one...

Number values are rounded and formatted to `MIN_DECIMALS` to avoid these.

src/h5web/visualizations/heatmap/DomainSlider.tsx
```diff
@@ -0,0 +1,57 @@
...
+    state.setCustomDomain,
+    shallow,
+  ]);
+
+  if (dataDomain === undefined) {
+    return <></>;
+  }
+
+  const domainExtent = dataDomain[1] - dataDomain[0];
+  const domainExtension = EXTEND_PERCENTAGE * domainExtent;
```
loichuder, line 28: An extension to the "true" domain so that slider values can go above/below the true domain.

src/h5web/visualizations/heatmap/store.ts
```diff
@@ -20,8 +20,10 @@ export type DataScale =
   | ScaleSymLog<number, number>;
 
 export interface HeatmapState {
-  domain?: Domain;
-  findDomain: (values: number[]) => void;
+  dataDomain?: Domain;
+  customDomain?: Domain;
+  findDataDomain: (values: number[]) => void;
+  setCustomDomain: (customDomain: Domain) => void;
```
loichuder, line 26: I distinguish now the "true' domain of the data `dataDomain` and the `customDomain` set by the slider. When changing the dataset, the `customDomain` is set to the `dataDomain`.

src/h5web/visualizations/heatmap/store.ts
```diff
@@ -62,14 +68,12 @@ export function selectColorScale(state: HeatmapState): ColorScale {
...
-  if (state.domain === undefined) {
+  if (state.customDomain === undefined) {
     return undefined;
   }
 
   const scale = (state.hasLogScale ? scaleSymlog : scaleLinear)();
-  scale.domain(state.domain);
-  // Extend the domain to "nice" values to have nice ticks afterwards
-  scale.nice();
+  scale.domain(state.customDomain);
```
loichuder, line 76: I removed the nicing of ticks so that the values of the slider are truly the values of the dataScale.

src/h5web/visualizations/heatmap/DomainSlider.tsx
axelboc: We already have `lodash-es`, so feel free to import `{ round }` from it ;)

src/h5web/visualizations/heatmap/DomainSlider.tsx
axelboc: Interesting. Do you think it's a feature people might want?

axelboc: If so, perhaps it would be simpler to store this extended domain in the store directly.

src/h5web/visualizations/heatmap/store.ts
axelboc: Right, because the "nicing" is done in the custom domain. Makes sense.

axelboc:
Roh, the slider looks good! ✨

image

Looks like the toolbar is becoming taller and taller, though. When you switch visualization, its height changes, so it looks a bit strange. Ideally, we should set a fixed height to the `visBar`, and make sure every toolbar component fits inside it.

src/h5web/visualizations/heatmap/DomainSlider.tsx
loichuder: Yes it is helpful sometimes to alleviate some saturation effects. :+1: for the extended domain.

PR #106, opened 2020-04-29
Provide props through context and use hooks to handle logic

axelboc:
Sorry for the gigantic diff... In summary:

- `store.ts` becomes `config.ts` -- The config store is now very dumb; it doesn't store the domain doesn't compute anything.
- Props are now provided to the heatmap through context (cf. `HeatmapProvider.tsx`). Note that the provider has to be provided again inside R3F's `Canvas` - cf. https://github.com/react-spring/react-three-fiber/issues/262.
- Interpolator, scales, texture data, etc. are all accessed through custom hooks (cf. `hooks.ts`), which abstract accessing the config and/or props and memoising computations. When web workers are introduced to perform expensive computations asynchronously, this will be done in these hooks.


src/h5web/dataset-visualizer/VisDisplay.tsx
```diff
@@ -44,10 +46,12 @@ function VisDisplay(props: Props): JSX.Element {
 
   if (vis === Vis.Heatmap) {
     return (
-      <HeatmapVis
-        dims={(dataset.shape as HDF5SimpleShape).dims as [number, number]}
+      <HeatmapProvider
+        dims={(dataset.shape as HDF5SimpleShape).dims as Dims}
         data={value}
```
axelboc, line 51: `dims` and `data` are now provided through context.

src/h5web/visualizations/heatmap/AxisGrid.tsx
```diff
@@ -3,20 +3,16 @@ import { useThree, useFrame, Dom } from 'react-three-fiber';
 import { scaleLinear } from 'd3-scale';
 import IndexAxis from './IndexAxis';
 import styles from './HeatmapVis.module.css';
-import { Domain } from './store';
+import { useProps } from './hooks';
+import { Domain } from './models';
```
axelboc, line 7: Now that the store no longer contains all the computations, I've moved all shared models to `models.ts`.

src/h5web/visualizations/heatmap/AxisGrid.tsx
```diff
@@ -3,20 +3,16 @@ import { useThree, useFrame, Dom } from 'react-three-fiber';
...
 
-interface Props {
-  dims: [number, number];
-  axisOffsets: [number, number];
-}
-
-function AxisGrid(props: Props): JSX.Element {
-  const { dims, axisOffsets } = props;
+function AxisGrid(): JSX.Element {
+  const { dims, axisOffsets } = useProps();
```
axelboc, line 15: `useProps` is a custom hook that allows access to the Heatmap's core props, `dims` and `data` (and `axisOffsets` for convenience). Internally it just returns the value of the heatmap's context.

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -3,11 +3,12 @@ import { AxisRight } from '@vx/axis';
...
 import { adaptedNumTicks, generateCSSLinearGradient } from './utils';
 import styles from './HeatmapVis.module.css';
-import { useHeatmapStore, selectDataScale, selectInterpolator } from './store';
+import { useDataScale, useInterpolator } from './hooks';
 
 function ColorBar(): JSX.Element {
-  const interpolator = useHeatmapStore(selectInterpolator);
-  const dataScale = useHeatmapStore(selectDataScale);
+  const dataScale = useDataScale();
+  const interpolator = useInterpolator();
```
axelboc, line 10: As mentioned, everything is a custom hook - cf. `hooks.ts`.

src/h5web/visualizations/heatmap/HeatmapProvider.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { createContext, ReactElement, ReactNode } from 'react';
+import { Dims, AxisOffsets } from './models';
+
+export interface HeatmapProps {
+  dims: Dims;
+  data: number[][];
+  axisOffsets: AxisOffsets;
+}
```
axelboc, line 8: This is the shape of the context's value (i.e. what's returned by `useProps()`).

src/h5web/visualizations/heatmap/HeatmapProvider.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const { dims, data, children } = props;
+
+  return (
+    <HeatmapContext.Provider value={{ dims, data, axisOffsets: [72, 36] }}>
+      {children}
+    </HeatmapContext.Provider>
+  );
+}
+
+export default HeatmapProvider;
```
axelboc, line 30: This is the context provider. It just forwards the props, along with`axisOffsets` for convenience.

src/h5web/visualizations/heatmap/HeatmapToolbar.tsx
```diff
@@ -1,23 +1,16 @@
...
     keepAspectRatio,
     toggleAspectRatio,
-  ] = useHeatmapStore(state => [
-    state.hasLogScale,
-    state.toggleLogScale,
-    state.keepAspectRatio,
-    state.toggleAspectRatio,
-    shallow,
-  ]);
+  } = useHeatmapConfig();
```
axelboc, line 13: We can just get the entire config now.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -1,39 +1,16 @@
...
-
-  const values = useMemo(() => data.flat(), [data]);
-  const textureData = useMemo(
-    () => computeTextureData(values, colorScale, dataScale),
-    [dataScale, colorScale, values]
-  );
-
-  useEffect(() => {
-    findDomain(values);
-  }, [findDomain, values]);
```
axelboc, line 36: Yep, `HeatmapVis` doesn't do much anymore... 😄 Each child component takes care of retrieving what it needs by itself.

src/h5web/visualizations/heatmap/Tooltip.tsx
```diff
@@ -57,12 +55,9 @@ function Tooltip(props: Props): ReactElement {
   const onPointerOut = useCallback(hideTooltip, [hideTooltip]);
   const onPointerDown = useCallback(hideTooltip, [hideTooltip]);
 
-  // Trigger `pointermove` when user stops panning to show tooltip if pointer still intersects with mesh
-  const onPointerUp = useCallback(intersect, [intersect]);
```
axelboc, line 61: Small bug fix: when zoomed in, `intersect()` could return true even when outside the canvas so the popup would reappear when it shouldn't. Now, I just leave the tooltip hidden until the user starts to move the mouse again.

src/h5web/visualizations/heatmap/config.ts
```diff
@@ -0,0 +1,25 @@
...
+  colorMap: 'Magma',
+  setColorMap: (colorMap: ColorMap) => set({ colorMap }),
+
+  hasLogScale: false,
+  toggleLogScale: () => set(state => ({ hasLogScale: !state.hasLogScale })),
+
+  keepAspectRatio: true,
+  toggleAspectRatio: () =>
+    set(state => ({ keepAspectRatio: !state.keepAspectRatio })),
+}));
```
axelboc, line 25: The config is now dumb. I removed the domain and the selectors.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -1,19 +1,101 @@
...
+export function useProps(): HeatmapProps {
+  const props = useContext(HeatmapContext);
 
-  const keepAspectRatio = useHeatmapStore(state => state.keepAspectRatio);
+  if (!props) {
+    throw new Error('Missing Heatmap provider.');
+  }
+
+  return props;
+}
```
axelboc, line 31: This file is where things actually happen. Here, is the `useProps` hook that provides a convenient way to access the props (`dims`, `data`, `axisOffsets`) from the context.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -1,19 +1,101 @@
...
+
+export function useValues(): number[] {
+  const { data } = useProps();
+  return useMemo(() => data.flat(), [data]);
+}
```
axelboc, line 36: `useValues` to access the flattened dataset.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -1,19 +1,101 @@
...
+
+function useDomain(): Domain | undefined {
+  const values = useValues();
+
+  return useMemo(() => {
+    const [min, max] = extent(values);
+    return min === undefined || max === undefined ? undefined : [min, max];
+  }, [values]);
+}
```
axelboc, line 45: `useDomain` to find the domain. Note that I use `useMemo` in most custom hooks 1) to avoid redoing heavy computations unnecessarily, and 2) to always return the same object reference (so computations further down the chain can themselves be memoised).

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -1,19 +1,101 @@
...
+
+export function useInterpolator(): D3Interpolator {
+  const colorMap = useHeatmapConfig(state => state.colorMap);
```
axelboc, line 48: Since I'm in a custom hooks, I can use the heatmap config's custom hook.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -1,19 +1,101 @@
...
+  const dataScale = useDataScale();
+  const interpolator = useInterpolator();
+  const values = useValues();
+
+  return useMemo(() => {
+    if (!dataScale) {
+      return undefined;
+    }
+
+    const colorScale = scaleSequential(interpolator);
```
axelboc, line 79: Note that the color scale doesn't need its own hook since it's only used here.

src/h5web/visualizations/heatmap/config.ts
```diff
@@ -0,0 +1,54 @@
...
+      return;
+    }
+
+    const extension = (max - min) * EXTEND_PERCENTAGE;
+
+    set({
+      extendedDomain: [min - extension, max + extension],
+      customDomain: [min, max], // init custom domain to data domain
+    });
+  },
```
loichuder, line 40: I think we need to take the "real" data domain `[min, max]` accessible somewhere, either stored as a variable or through a custom hook wired to `extent`. The reason is that I foresee an Auto-scale button for the limits that sets `customData` back to `dataDomain`.

PS: you didn't like my solution to keep the `dataDomain` as variable and supply a function to generate an extended version ?

src/h5web/visualizations/heatmap/DomainSlider.tsx
```diff
@@ -35,16 +36,16 @@ function DomainSlider(): JSX.Element {
...
       step={step}
       pearling
       onAfterChange={values => {
-        const roundedValues = (values as Domain).map(value =>
+        const roundedDomain = (values as number[]).map(value =>
           round(value, MIN_DECIMALS)
-        );
-        setCustomDomain(roundedValues as Domain);
+        ) as Domain;
+        setCustomDomain(roundedDomain);
```
loichuder, line 48: Thanks for the clean-up :wink:

axelboc: No worries. The important change is `value` instead of `defaultValue`. With `defaultValue`, the change of custom domain when changing dataset was not being processed by the slider.

src/h5web/visualizations/heatmap/HeatmapProvider.tsx
axelboc: hmm smart 💯

src/h5web/visualizations/heatmap/config.ts
axelboc: Hmm true, the `customDomain` could remain `undefined` until changed, with a fallback to the real domain. 💯 

I made the choice here to not store the real domain, which is why it made sense to store the extended domain in the config. With the change above, the extended domain computation can be moved to a custom hook. 💯

src/h5web/visualizations/heatmap/models.ts
```diff
@@ -0,0 +1,16 @@
+import { ScaleLinear, ScaleSymLog } from 'd3-scale';
+import { INTERPOLATORS } from './interpolators';
+
+export type Dims = [number, number];
```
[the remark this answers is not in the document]
axelboc: Yeah, having a separate type for it will make dealing with this a little easier, hopefully.

src/h5web/visualizations/heatmap/HeatmapToolbar.tsx
axelboc: Oh sorry, I changed this back. It's a little strange to have some components that are controlled (i.e. the togglers) and some that control themselves in the toolbar, but I didn't like any of the alternatives, so I left it as is for now.

PR #107, opened 2020-04-30
Perform texture computation on web worker

axelboc:
I used [`comlink`](https://github.com/GoogleChromeLabs/comlink), which simplifies writing and communicating with web workers, along with [`react-use-comlink`](https://github.com/pocesar/react-use-comlink).

I had to replace `useDataScale` with a utility function so I could recreate the data scale on the web worker (functions can't be sent to web workers).

The result is that the UI remains fluid while the texture is being recomputed.

ezgif-2-ff38971ec5c4

Next, I'll investigate whether there's a better way to send the values to the web worker. At the moment, the JS array is copied. I hear it's better to _transfer_ a `Uint8Array` by reference... I don't think I'll bother with moving finding the domain to a web worker, as it seems to be pretty fast at the moment, and it only occurs once when selecting a dataset (contrary to the texture, which is recomputed when the config changes).

### DIFF
--- a/src/h5web/visualizations/heatmap/hooks.ts
+++ b/src/h5web/visualizations/heatmap/hooks.ts
@@ -5,18 +5,23 @@ import React, {
   CSSProperties,
   useContext,
   useMemo,
+  useState,
 } from 'react';
 import { Vector3 } from 'three';
 import { ReactThreeFiber, PointerEvent, useThree } from 'react-three-fiber';
 import { clamp } from 'lodash-es';
+import { useComlink } from 'react-use-comlink';
 import { useMeasure } from 'react-use';
-import { scaleSymlog, scaleLinear, scaleSequential } from 'd3-scale';
-import { rgb } from 'd3-color';
 import shallow from 'zustand/shallow';
 import { useHeatmapConfig } from './config';
 import { HeatmapProps, HeatmapContext } from './HeatmapProvider';
-import { DataScale, D3Interpolator } from './models';
+import { D3Interpolator } from './models';
 import { INTERPOLATORS } from './interpolators';
+import { TextureWorker } from './worker';
+
+// eslint-disable-next-line
+// @ts-ignore
+import Worker from 'worker-loader!./worker'; // eslint-disable-line
 
 const ZOOM_FACTOR = 0.95;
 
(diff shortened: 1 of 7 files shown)


src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -40,44 +45,40 @@ export function useInterpolator(): D3Interpolator {
...
-      return [r, g, b];
-    });
+    (async () => {
+      setTextureData(
+        await proxy.computeTextureData(
+          values,
+          customDomain || dataDomain,
+          hasLogScale,
+          colorMap
+        )
```
axelboc, line 76: Thanks to `comlink`, it's just like calling an async method!

loichuder: Sweet !

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -1,12 +1,18 @@
...
 export const adaptedNumTicks = scaleLinear()
   .domain([300, 900])
   .rangeRound([3, 10])
   .clamp(true);
 
+export function getDataScale(domain: Domain, isLog: boolean): DataScale {
+  const scale = (isLog ? scaleSymlog : scaleLinear)();
+  scale.domain(domain);
+  return scale;
+}
```
loichuder, line 14: If I understand well, you had to make `useDataScale` a regular function rather than a hook to be able to call it from the worker that is not a hook ?

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -5,18 +5,23 @@ import React, {
...
 import { useHeatmapConfig } from './config';
 import { HeatmapProps, HeatmapContext } from './HeatmapProvider';
-import { DataScale, D3Interpolator } from './models';
+import { D3Interpolator } from './models';
 import { INTERPOLATORS } from './interpolators';
+import { TextureWorker } from './worker';
+
+// eslint-disable-next-line
+// @ts-ignore
+import Worker from 'worker-loader!./worker'; // eslint-disable-line
```
loichuder, line 24: `eslint` problems ?

src/h5web/visualizations/heatmap/utils.ts
axelboc: I would have liked to pass the scale returned by `useDataScale` to the web worker, but as I mentioned, you can't pass functions to a web worker. https://github.com/GoogleChromeLabs/comlink/blob/master/structured-clone-table.md

src/h5web/visualizations/heatmap/hooks.ts
axelboc: 🙄 and create-react-app, which doesn't include `worker-loader`...

axelboc: ... ad Jest, which doesn't really like webpack's inline loader syntax...

PR #109, opened 2020-04-30
Added a button to reset colormap limits to data

loichuder:
image

### DIFF
--- a/src/h5web/visualizations/heatmap/DomainSlider.tsx
+++ b/src/h5web/visualizations/heatmap/DomainSlider.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import ReactSlider from 'react-slider';
 import shallow from 'zustand/shallow';
 import { format } from 'd3-format';
-import { round } from 'lodash-es';
+import { round, debounce } from 'lodash-es';
 import { FiRotateCcw } from 'react-icons/fi';
 import styles from './DomainSlider.module.css';
 import { useHeatmapConfig } from './config';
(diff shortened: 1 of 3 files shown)


axelboc:
How about putting the button on the left of the slider and using an icon instead of text, like [`FiRotateCcw` or `FiRefreshCcw`](https://react-icons.netlify.com/#/icons/fi)?

src/h5web/visualizations/heatmap/DomainSlider.tsx
```diff
@@ -33,25 +33,34 @@ function DomainSlider(): JSX.Element {
...
+          const roundedDomain = (values as number[]).map(value =>
+            round(value, MIN_DECIMALS)
+          ) as Domain;
+          setCustomDomain(roundedDomain);
+        }}
+      />
+      <button
+        className={styles.resetButton}
+        type="button"
+        onClick={() => setCustomDomain(dataDomain)}
```
axelboc, line 59: You could disable the button if `customDomain` is `undefined`: `disabled={!customDomain}`

axelboc: Oh and it should be:

```jsx
onClick={() => setCustomDomain(undefined)}
```

When the custom domain is `undefined`, the visualization falls back to using the `dataDomain`. It allows doing stuff if no custom domain is set (like disabling this button).

loichuder: Clever :100:, I will do that !

PR #110, opened 2020-04-30
Implement texture loading effect

axelboc:
✨ 🎇 🌟 

It looks better live, but here's a taster:

ezgif-2-31b9933225d9

### DIFF
--- a/src/h5web/visualizations/heatmap/HeatmapVis.module.css
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.module.css
@@ -30,6 +30,32 @@
   pointer-events: none;
 }
 
+.textureLoader {
+  background-color: rgba(0, 0, 0, 0.4);
+  background-image: linear-gradient(
+    to right,
+    transparent,
+    rgba(255, 255, 255, 0.5) 60%,
+    transparent
+  );
+  background-position-x: -50%;
+  background-size: 35% 100%; /* allows animating position in percent */
+  background-repeat: no-repeat;
+  transition: opacity 0.2s ease-out;
+  opacity: 0;
+}
+
+.textureLoader[data-visible] {
+  opacity: 1;
+  animation: shiftBg 1s ease-in-out infinite;
+}
+
+@keyframes shiftBg {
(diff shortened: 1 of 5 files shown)


src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -30,6 +30,29 @@
...
+}
+
+.textureLoader[data-visible] {
+  opacity: 1;
+  animation: loading 1s ease-in-out infinite;
+}
+
+@keyframes loading {
+  to {
+    background-position-x: 150%;
```
axelboc, line 52: The gradient, which takes 35% of the width of the canvas, moves from -50% to 150% in a loop.

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -30,6 +30,29 @@
...
+      to right,
+      transparent,
+      rgba(255, 255, 255, 0.5) 60%,
+      transparent
+    ) -50% top / 35% 100% no-repeat;
+  transition: opacity 0.2s ease-out;
+  opacity: 0;
+}
+
+.textureLoader[data-visible] {
```
axelboc, line 45: I use an attribute to show/hide the element instead of just removing it from the DOM so that it can fade out.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -57,28 +61,38 @@ export function useTextureData(): Uint8Array | undefined {
   );
 
   const values = useValues();
-
   const { proxy } = useComlink<TextureWorker>(() => new Worker(), []);
-  const [textureData, setTextureData] = useState<Uint8Array>();
+
+  const [state, mergeState] = useSetState<TextureDataState>({});
```
axelboc, line 66: `useSetState` provides a function to merge the state instead of just overriding it completely. It simplifies setting `loading` to `true` while keeping the existing texture data (to avoid a flash).

PR #113, opened 2020-05-04
Transfer buffers to/from worker to avoid cloning

axelboc:
This should improve the performance of texture computations.

### DIFF
--- a/src/h5web/visualizations/heatmap/hooks.ts
+++ b/src/h5web/visualizations/heatmap/hooks.ts
@@ -1,4 +1,5 @@
 import { CSSProperties, useContext, useMemo, useEffect } from 'react';
+import { transfer } from 'comlink';
 import { useComlink } from 'react-use-comlink';
 import { useMeasure, useSetState } from 'react-use';
 import shallow from 'zustand/shallow';
@@ -58,12 +59,17 @@ export function useTextureData(): TextureDataState {
       return;
     }
 
+    // Keep existing texture data, if any
+    mergeState({ loading: true });
+
+    // Prepare transferable buffer to avoid cloning values array
+    const typedValues = Float64Array.from(values);
+
     (async () => {
-      mergeState({ loading: true }); // keep existing texture data, if any
       mergeState({
         loading: false,
         textureData: await proxy.computeTextureData(
-          values,
+          transfer(typedValues, [typedValues.buffer]),
           customDomain || dataDomain,
           hasLogScale,
           colorMap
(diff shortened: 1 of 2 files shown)


axelboc:
From a completely unscientific comparison of the loading animations in prod and in the deploy preview, it looks like it takes about **a third of the time** to compute the texture with the optimisation than it does without!

axelboc:
Something strange, though, which also occured without the optimisation:

- when you open a dataset and change the domain, every change takes about 1/3 loading animation;
- but when you change to another color map, every subsequent change of domain takes about 4/3 loading animations (one full horizontal "scan" + 1/3)

axelboc:
Never mind, some color scales just take a lot more computation time than others.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -58,12 +59,17 @@ export function useTextureData(): TextureDataState {
       return;
     }
 
+    // Keep existing texture data, if any
+    mergeState({ loading: true });
+
+    // Prepare transferable buffer to avoid cloning values array
+    const typedValues = Float64Array.from(values);
```
axelboc, line 66: Float 64 is the type for numbers in JS. Eventually, we may have to parse the JSON (or whatever format) returned by the server and create the appropriate typed array right away to optimise things further.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -58,12 +59,17 @@ export function useTextureData(): TextureDataState {
...
+
     (async () => {
-      mergeState({ loading: true }); // keep existing texture data, if any
       mergeState({
         loading: false,
         textureData: await proxy.computeTextureData(
-          values,
+          transfer(typedValues, [typedValues.buffer]),
```
axelboc, line 72: That's the `postMessage` syntax to transfer a buffer https://devdocs.io/dom/worker/postmessage

src/h5web/visualizations/heatmap/worker.ts
```diff
@@ -16,12 +16,15 @@ function computeTextureData(
   const colorScale = scaleSequential(INTERPOLATORS[colorMap]);
 
   // Compute RGB color array for each datapoint `[[<r>, <g>, <b>], [<r>, <g>, <b>], ...]`
-  const colors = values.map(val => {
+  const colors = values.reduce<number[]>((acc, val) => {
```
axelboc, line 19: I replace `map` with `reduce` to avoid having to flatten the array afterwards.

PR #115, opened 2020-05-05
Refactor wrapping of Canvas

loichuder:


src/h5web/visualizations/shared/VisCanvas.tsx
```diff
@@ -0,0 +1,57 @@
...
+            <ambientLight />
+            {children}
+          </Canvas>
+        </div>
+      )}
+    </div>
+  );
+}
+
+export default VisCanvas;
```
loichuder, line 57: Other possible names: `VisWrapper`, `VisChart`...

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -26,31 +27,20 @@ function HeatmapVis(): JSX.Element {
...
-      </div>
+      <VisCanvas axisOffsets={axisOffsets} aspectRatio={aspectRatio}>
+        {/* Provide context again - https://github.com/react-spring/react-three-fiber/issues/262 */}
+        <HeatmapProvider {...props}>
+          <AxisGrid
+            // -0.5 to have ticks at the center of pixels
+            abscissaDomain={[-0.5, cols - 0.5]}
+            ordinateDomain={[-0.5, rows - 0.5]}
+            axisOffsets={axisOffsets}
+          />
```
loichuder, line 38: `axisOffsets` are passed as props for `VisCanvas` and `AxisGrid`. You thought about a provider but it might not be worth it for only two components ?

src/h5web/visualizations/shared/utils.ts
```diff
@@ -4,3 +4,16 @@ export const adaptedNumTicks = scaleLinear()
...
+  availableHeight: number,
+  aspectRatio: number
+): { width: number; height: number } {
+  // Determine how to compute canvas size to fit available space while maintaining aspect ratio
+  const shouldAdjustWidth = availableWidth >= availableHeight * aspectRatio;
+
+  return shouldAdjustWidth
+    ? { width: availableHeight * aspectRatio, height: availableHeight }
+    : { width: availableWidth, height: availableWidth * aspectRatio };
+}
```
loichuder, line 19: This is the stripped version of the previous hook `useHeatmapStyles`. All the setting of `padding` and `boxSizing` was moved out so that it only computes and returns the width and height.

src/h5web/visualizations/shared/VisCanvas.tsx
axelboc: I think `VisCanvas` is fine 👍

src/h5web/visualizations/shared/utils.ts
axelboc: Cool, but why not keep the hook itself with `useMeasure` and the early returns?

axelboc: ```const visSize = useHeatmapSize(...)``` -- like we had before but without the padding styles

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: I think you had convinced me that `AxisGrid` could be rendered by `VisCanvas` since it doesn't need to be inside the `HeatmapProvider` anymore. 😅 I wasn't a fan of forwarding the domain props, but you're right, it's not as bad as passing `axisOffsets` in two places.

src/h5web/visualizations/shared/models.ts
```diff
@@ -1,3 +1,8 @@
 export type Domain = [number, number];
 
-export type AxisOffsets = [number, number];
+export type AxisOffsets = { left: number; bottom: number };
```
loichuder, line 3: I changed AxisOffsets signature to be consistent with AxisDomains. It also removes the need to destructure the array several times.

It might be worth merging AxisOffsets and AxisDomains once we agree on the axis configuration options.

src/h5web/visualizations/shared/utils.ts
loichuder: I moved back everything in the function except the padding and the `useMeasure`.

src/h5web/visualizations/shared/models.ts
axelboc: Good call

src/h5web/visualizations/shared/utils.ts
```diff
@@ -1,6 +1,35 @@
...
+  const availableWidth = width - axisOffsets.left;
+  const availableHeight = height - axisOffsets.bottom;
+
+  // Determine how to compute canvas size to fit available space while maintaining aspect ratio
+  const shouldAdjustWidth = availableWidth >= availableHeight * aspectRatio;
+  const canvasSize = shouldAdjustWidth
+    ? { width: availableHeight * aspectRatio, height: availableHeight }
+    : { width: availableWidth, height: availableWidth * aspectRatio };
+
+  return { ...canvasSize, boxSizing: 'content-box' };
```
axelboc, line 34: Just realised that now that the padding lives in the parent, you could move `content-box` there as well. It can be applied regardless of the aspect ratio config.

PR #116, opened 2020-05-05
Persist heatmap config

### LINKED ISSUE #72: Persist toolbar state to local storage

axelboc:
Fix #72 

I added quite a few comments in the code, since it's not trivial. Let me know if anything needs clarifying.

### DIFF
--- a/src/h5web/storage-utils.ts
+++ b/src/h5web/storage-utils.ts
@@ -1,0 +1,55 @@
+import create, { State, StateCreator, StoreApi, UseStore } from 'zustand';
+
+export interface StorageConfig {
+  storageId: string;
+  itemsToPersist: string[];
+}
+
+function fromStorage<T>(id: string): T | undefined {
+  try {
+    const value = window.localStorage.getItem(id);
+    return value !== undefined && value !== null
+      ? JSON.parse(value)
+      : undefined;
+  } catch {
+    // Access to local storage may not be authorized or available,
+    // or the data stored may be corrupted.
+    return undefined;
+  }
+}
+
+export function createPersistableState<S extends State>(
+  config: StorageConfig,
+  createState: StateCreator<S>
+): [UseStore<S>, StoreApi<S>] {
+  const { storageId, itemsToPersist } = config;
+
+  return create<S>((set, get, api) => {
+    // Retrieve persisted state, if any
+    const persistedState = fromStorage<Partial<S>>(storageId);
+
+    // Invoke provided state creator to retrieve initial state
+    const initialState = createState(
+      // Monkey patch `set` to persist state on every call
+      args => {
+        set(args);
+
+        // Retrieve new state to persist, keeping only the relevant keys
+        const newStateStr = JSON.stringify(get(), itemsToPersist);
+
+        try {
+          window.localStorage.setItem(storageId, newStateStr);
+        } catch {
+          // Do nothing (local storage may be unavailable or unauthorized)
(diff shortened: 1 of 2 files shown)


src/h5web/storage-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
+
+export function createPersistableState<S extends State>(
+  config: StorageConfig,
+  createState: StateCreator<S>
+): [UseStore<S>, StoreApi<S>] {
+  const { key, whitelist } = config;
+
+  return create<S>((set, get, api) => {
```
axelboc, line 27: I decided to wrap zustand's `create` method instead of creating a "classic" middleware. We can change that later if we need more flexibility.

src/h5web/storage-utils.ts
```diff
@@ -0,0 +1,54 @@
+import create, { State, StateCreator, StoreApi, UseStore } from 'zustand';
+
+export interface StorageConfig {
+  key: string;
+  whitelist: string[];
```
loichuder, line 5: I get why you called it `whitelist` but I find the name could be more meaningful.

`itemsToPersist`?, `persistentItems`?, `itemsToStore` ?

src/h5web/storage-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+    const value = window.localStorage.getItem(key);
+    return value !== undefined && value !== null
+      ? JSON.parse(value)
+      : undefined;
+  } catch {
+    // Access to local storage may not be authorized or available,
+    // or the data stored may be corrupted.
+    return undefined;
+  }
+}
```
loichuder, line 19: The naming puzzles me a bit here. I have the feeling that we get a single value from its key (ex: `colorMap`).

Perhaps I am simply not used to this so if it is the convention, leave it as it is.

src/h5web/storage-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    // Invoke provided state creator to retrieve initial state
+    const initialState = createState(
+      // Monkey patch `set` to persist state on every call
+      args => {
+        set(args);
+
+        try {
+          // Retrieve and persist new state
+          const stateStr = JSON.stringify(get(), whitelist); // filter keys to be persisted with whitelist
```
loichuder, line 39: Should this line be out of the `try` block to avoid catching errors not linked to the local storage ?

src/h5web/storage-utils.ts
axelboc: You can only store strings in local storage, but the string can be any valid JSON (a string, a number, a stringified object, etc.), so this function is meant as a generic utility to retrieve and parse any storage value. Would `readFromStorage` be any clearer?

src/h5web/storage-utils.ts
axelboc: totally 🎣

src/h5web/storage-utils.ts
loichuder: The function name is fine. What I mean is I am a bit confused as I see fields in the configuration as `(key, value)` pairs while here the key is the identifier of the config and the value the config itself.
In other words, one cannot do `fromStorage(colorMap)` to get the colorMap value. You have to get the whole config from its key/identifier, parse it and get the colorMap from it.

But that is not a problem, the naming stems from the fact that you want this function to be generic so you can leave it.

PR #117, opened 2020-05-06
Replaces Recharts' LineVis with a vis based on react-three-fibers

### LINKED ISSUE #42: [Line] Switch from Recharts to react-three-fiber
After experimenting with `react-three-fiber` for the Heatmap visualization, I think it'd be a much better solution than Recharts for the Line visualization. It's easy to use and stable, and it performs much better than all of the charting libraries that render to SVG.

Before we make the switch, though, we should first implement axis in the Heatmap visualization.

loichuder:
Fix #42 

image
The visualisation is still a little bare-bones but I think adding more features and tuning the aspect should be part of other PRs.

Note that pan and zooming are working thanks to the `PanZoomMesh` component !


src/h5web/visualizations/line/config.ts
```diff
@@ -0,0 +1,26 @@
...
+  STORAGE_CONFIG,
+  set => ({
+    abscissaDomain: undefined,
+    setAbscissaDomain: (abscissaDomain: Domain | undefined) =>
+      set({ abscissaDomain }),
+    ordinateDomain: undefined,
+    setOrdinateDomain: (ordinateDomain: Domain | undefined) =>
+      set({ ordinateDomain }),
+  })
+);
```
loichuder, line 26: This was adapted from the `HeatmapConfig` to directly have to possiblity to use persistent storage. Even if it is not used yet...

src/h5web/visualizations/heatmap/DomainSlider.tsx
```diff
@@ -7,8 +7,9 @@ import { FiRotateCcw } from 'react-icons/fi';
 import styles from './DomainSlider.module.css';
 import { useHeatmapConfig } from './config';
 import { Domain } from '../shared/models';
+import { extendDomain } from '../shared/utils';
 
-const EXTEND_PERCENTAGE = 0.2;
+const EXTEND_FRACTION = 0.2;
```
axelboc, line 12: `EXTEND_FACTOR` ?

src/h5web/visualizations/line/DataLine.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const points = useMemo(() => {
+    return data.map((val, index) => new Vector3(xScale(index), yScale(val), 0));
+  }, [data, xScale, yScale]);
+
+  return (
+    <line>
+      <bufferGeometry
+        attach="geometry"
+        onUpdate={useCallback(self => self.setFromPoints(points), [points])}
```
loichuder, line 39: I cannot put `points` as a props so I have to use this syntax to call the `setFromPoints` 
 method from Three.js of `bufferGeometry`

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -0,0 +1,54 @@
...
+    state => [state.setAbscissaDomain, state.setOrdinateDomain],
+    shallow
+  );
+
+  const xDomain = extendDomain([0, data.length - 1], 0.1);
+  const extentD3 = extent(data);
+  const yDomain: Domain | undefined =
+    extentD3[0] !== undefined && extentD3[1] !== undefined
+      ? extendDomain(extentD3, 0.1)
+      : undefined;
```
loichuder, line 29: I extend both x and y domains by 10% to avoid clipping of the curve ends by the axes. This can be tuned/made more clever later.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <div className={styles.root}>
+      {yDomain && (
+        <VisCanvas
+          axisOffsets={axisOffsets}
+          axisDomains={{
+            left: yDomain,
+            bottom: xDomain,
+          }}
+        >
+          <PanZoomMesh />
```
loichuder, line 46: Enabling pan and zoom is a simple as adding this line !

src/h5web/visualizations/line/DataLine.tsx
```diff
@@ -0,0 +1,46 @@
...
+    state => [state.abscissaDomain, state.ordinateDomain],
+    shallow
+  );
+
+  const xScale = scaleLinear()
+    .domain(abscissaDomain as Domain)
+    .range([-width / 2, width / 2]);
+  const yScale = scaleLinear()
+    .domain(ordinateDomain as Domain)
+    .range([-height / 2, height / 2]);
```
loichuder, line 29: The guard clause in `LineVis` ensures that `abscissaDomain` and `ordinateDomain` are always defined so `as Domain` is only to make TS happy.

axelboc:
It looks sooo good! I'm actually quite liking the zoom on both axes. Not sure about the domain extension, but as you said, we can look at that after.

src/h5web/visualizations/line/config.ts
axelboc: I don't get why you put the axis domains in there, though.

- The abscissa domain is definitely not a configuration option.
- The ordinate domain is like the data domain in the heatmap, I guess, but it's in the heatmap config store because of `customDomain`, and we don't have a configurable custom domain here. Seeing this, I'm actually reconsidering removing the data domain from the heatmap config and just using a memoised hook instead.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -0,0 +1,54 @@
...
+import VisCanvas from '../shared/VisCanvas';
+import { Domain, AxisOffsets } from '../shared/models';
+import PanZoomMesh from '../shared/PanZoomMesh';
+import { extendDomain } from '../shared/utils';
+import { useLineVisConfig } from './config';
+
+interface Props {
+  data: number[];
+  axisOffsets: AxisOffsets;
+}
```
axelboc, line 15: You don't want to replicate the provider architecture we have in the heatmap?

axelboc: Maybe it's not as useful here for now...

src/h5web/visualizations/line/LineVis.tsx
axelboc, line 29: The undefined checks for min/max could perhaps be moved into a function and re-used for the heatmap. I would also put a `useMemo` around it here, as it can be an expensive operation.

src/h5web/visualizations/line/config.ts
loichuder: Following your comment https://github.com/silx-kit/h5web/pull/117#discussion_r420615672, I will replicate the provider architecture

src/h5web/visualizations/line/DataLine.tsx
axelboc: This feels a bit yuk, because it subtly couples this component to `{ yDomain && ... }` in `LineVis`.

It would be good if the cavans could be rendered without an ordinate domain and if this component could handle an `undefined` data/ordinate domain like the mesh in the heatmap. Do get there you can:

- make `axisDomain` optional in `VisCanvas` and render `AxisGrid` conditionally in the JSX;
- move the `points` computation below to a custom hook and return `undefined` if the ordinate domain is `undefined`, then render `<bufferGeometry />`  conditionally in the JSX. To go further and replicate what we do in the heatmap's `Mesh` (for performance), you could even memoise the whole material.

src/h5web/visualizations/line/hooks.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export function useDomains(): [Domain, Domain | undefined] {
+  const { data } = useProps();
+  const dataDomain = useMemo(() => findDomain(data), [data]);
+
+  return [
+    extendDomain([0, data.length - 1], 0.05),
+    dataDomain ? extendDomain(dataDomain, 0.05) : undefined,
+  ];
```
axelboc, line 23: You could just return an `AxisDomains` object so you could pass it directly to `VisCanvas`. I think it would make sense to return `undefined` for the whole object, as there's not much we can do with just the abscissa domain. This would also be consistent with mu suggestion above to make `axisDomain` optional in `VisCanvas`.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -31,3 +32,17 @@ export function computeVisSize(
...
+
+export function extendDomain(domain: Domain, extendFactor: number): Domain {
+  const [min, max] = domain;
+  const extension = (max - min) * extendFactor;
+
+  return [min - extension, max + extension];
+}
+
+export function findDomain(data: number[]): Domain | undefined {
+  const extentD3 = extent(data);
```
axelboc, line 44: `domain` ?

loichuder: I fear `domain` is too generic and may shadow some variables as we are moving a lot of domains around...

axelboc: it's just a local variable, though

loichuder: Yes, my fear was unfounded: it is all right everywhere else in fact.

axelboc: Oh sorry for the confusion, I meant the local variable `extentD3`, not the function name! I liked `findDomain` 😅

PR #118, opened 2020-05-06
Implement screenshot button

### LINKED ISSUE #95: [Heatmap] Export to image
Note that, in Firefox at least, there's a "View image" contextual option on the heatmap's canvas, which generates an image automatically. This doesn't work for me, but it works somewhat for others:

> When I do "View image" with Firefox I see the zoomed image but then if leave go to another tab in firefox and go back to the image I see a empty (white) image. I get this URL: `blob:null/b92087e5-f865-4a88-b13c-6c7b1481b192`.

It may be worth making sure "View image" works before implementing a more advanced export feature.

axelboc:
Fix #95. The new toolbar button converts the canvas to a data URI and opens it in a new tab. This is the equivalent to <kbd>right click > View image</kbd>, which now also works consistently. The generated image can then be saved as PNG with <kbd>Ctrl+S</kbd>

Note that the generated image has the same dimension of the canvas, so you need to go full screen to export a bigger image. I think this is sufficient to meet the requirements for an image export feature, but we'll see with usage.

ezgif-2-c2e920a66f15

I've implemented the button as a shared component so it can be reused in the line vis toolbar. I also made the choice:

- to just query for the canvas in the DOM instead of trying to get a reference through React, as this would have complexified things significantly...
- to do nothing if the canvas is not in the DOM -- I doubt users will try to click on the button before actually seeing the rendered canvas;
- to assume that the DOM only contains one canvas -- we can add an ID if this ever changes;
- to use a link instead of a button in order to take advantage of `target="_blank"`, as opening a tab in JS is not trivial.

### DIFF
--- a/src/h5web/visualizations/shared/ScreenshotButton.tsx
+++ b/src/h5web/visualizations/shared/ScreenshotButton.tsx
@@ -1,0 +1,32 @@
+import React, { ReactElement } from 'react';
+import { FiCamera } from 'react-icons/fi';
+import styles from './ScreenshotButton.module.css';
+
+function ScreenshotButton(): ReactElement {
+  return (
+    <a
+      className={styles.link}
+      href="/"
+      target="_blank"
+      aria-label="Screenshot"
+      onClick={evt => {
+        const canvas = document.querySelector('canvas');
+
+        // Create data URL from canvas
+        const screnshotUrl = canvas?.toDataURL();
+
+        if (screnshotUrl) {
+          // Let link open screenshot URL in new tab/window
+          evt.currentTarget.setAttribute('href', screnshotUrl);
+        } else {
+          // Don't follow link if canvas hasn't been rendered yet
+          evt.preventDefault();
+        }
+      }}
+    >
+      <FiCamera />
+    </a>
+  );
+}
+
+export default ScreenshotButton;
--- a/src/h5web/visualizations/shared/VisCanvas.tsx
+++ b/src/h5web/visualizations/shared/VisCanvas.tsx
@@ -34,7 +34,8 @@ function VisCanvas(props: Props): JSX.Element {
           <Canvas
             className={styles.canvasWrapper}
             orthographic
-            invalidateFrameloop
+            invalidateFrameloop // disable game loop
+            gl={{ preserveDrawingBuffer: true }} // for screenshot feature
           >
             <ambientLight />
             {axisDomains && (
--- a/src/h5web/visualizations/heatmap/HeatmapToolbar.module.css
+++ b/src/h5web/visualizations/heatmap/HeatmapToolbar.module.css
@@ -1,6 +1,5 @@
 .toolbar {
   display: flex;
-  padding: 0 0.5rem;
   font-size: 0.875rem;
 }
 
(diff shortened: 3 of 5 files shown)


src/h5web/visualizations/shared/ScreenshotButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <a
+      className={styles.link}
+      href="/"
+      target="_blank"
+      aria-label="Screenshot"
+      onClick={evt => {
+        const canvas = document.querySelector('canvas');
+
+        // Create data URL from canvas
+        const screnshotUrl = canvas?.toDataURL();
```
loichuder, line 16:
```suggestion
        const screenshotUrl = canvas?.toDataURL();
```

src/h5web/visualizations/shared/VisCanvas.tsx
```diff
@@ -34,7 +34,8 @@ function VisCanvas(props: Props): JSX.Element {
           <Canvas
             className={styles.canvasWrapper}
             orthographic
-            invalidateFrameloop
+            invalidateFrameloop // disable game loop
+            gl={{ preserveDrawingBuffer: true }} // for screenshot feature
```
loichuder, line 38: Does this impact performance ?

src/h5web/visualizations/shared/ScreenshotButton.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { ReactElement } from 'react';
+import { FiCamera } from 'react-icons/fi';
+import styles from './ScreenshotButton.module.css';
+
+function ScreenshotButton(): ReactElement {
+  return (
+    <a
+      className={styles.link}
+      href="/"
+      target="_blank"
```
loichuder, line 10: What about proposing directly the download of the PNG instead of opening of a new tab ?

axelboc:
> 
> I know it would be more difficult to implement but it feels strange to export the image without its context (colorbar and name of the dataset). This being said, we can decide that we provide a way to export only the image and that is the responsability of the user to capture everything he needs.

Yeah, I think it's a reasonable choice for now. People can always take a screenshot -- especially since browsers are getting a lot better at that these days (in Firefox, you can even take a screenshot of a DOM element).

image

src/h5web/visualizations/shared/ScreenshotButton.tsx
axelboc: Hmm I don't know -- I could see people using the current behaviour to keep multiple screenshots open in their browser for reference and comparison.

src/h5web/visualizations/shared/VisCanvas.tsx
axelboc: Hehe excellent question -- it doesn't _look_ like it, but it may increase memory usage and impact performance on some devices. I'm hoping they disable it by default because it's a bad thing to enable with a game loop, but since we disable the game loop... It didn't notice a slow down in the interactions, so in my opinion we should be fine. If we ever have memory issues, then we'll know to look here.

src/h5web/visualizations/shared/ScreenshotButton.tsx
loichuder: Ok then the current behaviour makes sense.

PR #123, opened 2020-05-07
Implements togglable grid for Line and Heatmap

loichuder:
Implemented with VX's [Grid](https://vx-demo.now.sh/static/docs/vx-grid.html). I also tried to tune a bit the aspect of the axes to be more consistent with the theming:

image
image


src/h5web/visualizations/heatmap/HeatmapToolbar.module.css
```diff
@@ -4,23 +4,6 @@
...
-
-.toggler:hover,
-.toggler[aria-checked='true'] {
-  background-color: var(--primary-dark-bg);
-}
-
-.togglerLabel {
-  margin-left: 0.5rem;
-}
-
```
loichuder, line 23: Forgot to remove this when I extracted the toggler: this CSS exists now in Toggler CSS module.

src/h5web/visualizations/line/LineVisToolbar.module.css
```diff
@@ -0,0 +1,5 @@
+.toolbar {
+  display: flex;
+  padding: 0 0.5rem;
+  font-size: 0.875rem;
+}
```
loichuder, line 5: It might be worthwhile to factor out the generic toolbar styling.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const axisProps = {
+    tickStroke: '#8cdfc7', // var(--secondary)
+    hideAxisLine: true,
+    tickFormat: format('0'),
+    tickClassName: styles.tick,
+    tickComponent: ({ formattedValue, ...tickProps }: TickRendererProps) => (
+      <text {...tickProps}>{formattedValue}</text>
+    ),
+  };
```
loichuder, line 60: axis props common to left and bottom axes.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -0,0 +1,121 @@
...
+        {showGrid && (
+          <div className={styles.axisGridCell}>
+            <svg width={width} height={height}>
+              <Grid
+                xScale={xScale}
+                yScale={yScale}
+                width={width}
+                height={height}
+                numTicksColumns={xTicksNum}
+                numTicksRows={yTicksNum}
```
loichuder, line 109: Having `AxisLeft` and `AxisBottom` in `AxisSystem` allows `Grid` to have access to the needed variables `xScale`, `yScale` and `numTicks...`

src/h5web/visualizations/shared/VisCanvas.tsx
```diff
@@ -4,20 +4,25 @@ import { useMeasure } from 'react-use';
...
-  const { axisDomains, axisOffsets, aspectRatio, children } = props;
+  const { axisDomains, axisOffsets, aspectRatio, children, showGrid } = props;
   const [visAreaRef, visAreaSize] = useMeasure();
+  const availableSize = {
+    width: visAreaSize.width - axisOffsets.left,
+    height: visAreaSize.height - axisOffsets.bottom,
+  };
 
-  const visSize = computeVisSize(visAreaSize, axisOffsets, aspectRatio);
+  const visSize = computeVisSize(availableSize, aspectRatio);
```
loichuder, line 25: Unrelated to the grid but I feel it is clearer when the computation of `availableSize` exists in the same scope of the padding addition.

src/h5web/visualizations/line/LineVisToolbar.module.css
axelboc: You can move the `div.toolbar` up to the parent

src/h5web/visualizations/line/config.ts
```diff
@@ -1,13 +1,19 @@
...
-type LineVisConfig = {};
+type LineVisConfig = {
+  showGrid: boolean;
+  toggleGrid: () => void;
+};
 
 const STORAGE_CONFIG: StorageConfig = {
-  storageId: 'h5web:line',
-  itemsToPersist: [],
+  storageId: 'h5web:linevis',
```
axelboc, line 9: Would naming the components `LinePlot<...>` instead of `LineVis` be better? Just thinking that we don't use `HeatmapVis<...>`, and "plot" would make more sense for, say, the future scatter plot visualization.

src/h5web/visualizations/shared/AxisSystem.module.css
```diff
@@ -34,7 +39,7 @@
 }
 
 .tick > text {
-  fill: currentColor;
+  fill: var(--secondary-dark);
```
axelboc, line 42: I have to say I preferred the labels and the grid black/grey. 😅 I think it's too much green now, and I feel like the green grid would not come out well with some colour maps. A very transparent black is safer.

axelboc: You could try with some greyish labels too, but be mindful that #777 is the lightest you can go for accessibility reasons.

axelboc: FYI, our colours don't have sufficient colour contrast with white (as text or bg), only with black. Here is the contrast checks for the full palette for reference: https://abc.useallfive.com/?colors[]=FFFFFF,020402,1B998B,40C9A2,B3E9DA,D9F4EC,ECFAF6,9AAE5D,C0DA74,D4E09B

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -0,0 +1,121 @@
...
+  showGrid?: boolean;
+}
+
+function AxisSystem(props: Props): JSX.Element {
+  const { axisDomains, axisOffsets, showGrid } = props;
+
+  const { camera, size } = useThree();
+  const { width, height } = size;
+
+  const [domains, setDomains] = useState<AxisDomains>(axisDomains);
```
axelboc, line 24: If I'm not mistaken, the `domains` stored here are only used to compute `xScale` and `yScale`. Therefore, I think you could simplify a lot of things by storing the scales in the state and computing them inside `useFrame`.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  // Axis bounds in R3F camera coordinates
+  const leftAxisBounds = [-height / 2, height / 2];
+  const bottomAxisBounds = [-width / 2, width / 2];
+
+  // Scales R3F camera coordinates to axis bounds
+  const leftAxisScale = scaleLinear()
+    .domain(leftAxisBounds)
+    .range(axisDomains.left);
+  const bottomAxisScale = scaleLinear()
```
axelboc, line 34: Maybe this needs renaming now. These scales are used to compute the bounds/domain of the axis as opposed to the ticks inside the axis.

src/h5web/visualizations/shared/Axis.tsx
```diff
@@ -1,58 +0,0 @@
...
-}
-
-function Axis(props: Props): JSX.Element {
-  const { className, domain, orientation } = props;
-
-  const [min, max] = domain;
-  const isLeftAxis = orientation === 'left';
-  const DirectionalAxis = isLeftAxis ? AxisLeft : AxisBottom;
-
-  const [divRef, { width, height }] = useMeasure();
```
axelboc, line 25: You're actually removing two `useMeasure`, nicely done! I don't know why we didn't realise we could have used `useThree` here to get the dimensions... 🤦

src/h5web/visualizations/shared/VisCanvas.tsx
axelboc: very true 👍

src/h5web/visualizations/shared/utils.ts
```diff
@@ -1,36 +1,32 @@
...
-  visAreaSize: Size,
-  axisOffsets: AxisOffsets,
+  availableSize: Size,
   aspectRatio?: number
 ): Size | undefined {
-  const { width, height } = visAreaSize;
+  const { width, height } = availableSize;
 
-  if (width === 0 && height === 0) {
+  if (width <= 0 && height <= 0) {
```
axelboc, line 16: 🤔 don't think these can be negative ^^ - `0` is the special case for `useMeasure`'s return value.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -1,36 +1,32 @@
...
     return undefined;
   }
 
-  const availableWidth = width - axisOffsets.left;
-  const availableHeight = height - axisOffsets.bottom;
-
   if (!aspectRatio) {
-    return { width: availableWidth, height: availableHeight };
+    return availableSize;
```
axelboc, line 21: nice !

src/h5web/visualizations/shared/utils.ts
loichuder: Yes but the dimensions here are the ones with the padding subtracted. Is `useThree` always taking the padding into account ?

axelboc: Ahh well, then that's why computing `availableWidth` inside the function, after this first return, made sense.

axelboc: Otherwise, you risk not rendering anything if the rendering area is less than or equal to the padding.

src/h5web/visualizations/line/config.ts
loichuder: I used `Line...` to be consistent with the naming of `Heatmap` components

src/h5web/visualizations/shared/AxisSystem.tsx
loichuder: :fishing_pole_and_fish:

src/h5web/visualizations/shared/utils.ts
loichuder: It is fine for me :upside_down_face:

src/h5web/visualizations/shared/AxisSystem.module.css
loichuder: Labels are now in darkslategrey and grid/ticks in grey

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -0,0 +1,137 @@
...
+function AxisSystem(props: Props): JSX.Element {
+  const { axisDomains, axisOffsets, showGrid } = props;
+
+  const { camera, size } = useThree();
+  const { width, height } = size;
+
+  const [axisScales, setAxisScales] = useState<AxisScales>({
+    x: scaleLinear(),
+    y: scaleLinear(),
+  });
```
axelboc, line 32: Argh, désolé, je n'avais pas anticipé ce problème : les scales ne sont plus valides au premier render, seulement après la première frame 😞. Du coup, il faudrait soit introduire un utilitaire du style `computeScales` et l'appeler ici et dans `useFrame()`, soit revert et remettre les bounds dans le state. Bon, le silver lining c'est que le code a bien gagné en lisibilité grâce à tes renommages.

src/h5web/visualizations/shared/utils.ts
axelboc: You're right, I got confused somewhere -- if there's no space available for the canvas, no point in rendering it... 🤦

Issue #104, opened 2020-04-28
[Heatmap] Axis labels are now cropped to image dimensions 


loichuder:
![image](https://user-images.githubusercontent.com/42204205/80472915-b1fb3300-8945-11ea-96e5-d84552e33df4.png)

[This line](https://github.com/silx-kit/h5web/blob/5897aafd9cdbf28c848bc0e30a2fd5b4d678d11b/src/h5web/visualizations/heatmap/HeatmapVis.module.css#L17) introduced in #99 is the responsible. @axelboc : is the `overflow: hidden` necessary ?

axelboc:
We can propably move it to the `mapArea`. It's to avoid a scroll overflow that occurs when switching from aspect ratio disabled to enabled.

axelboc:
I pushed the fix to `master` but the labels are still cropped when `keepAspectRatio` is disabled. If I move the overflow up again in the DOM, the heatmap overflows above the color bar when `keepAspectRatio` becomes enabled, which is really ugly.

Not sure if this can be fixed in a reasonable manner. We could provide render props to VX's `Axis` components to customise the way the ticks are rendered (i.e. so as to not render the last one if it's outside the bounds) -- https://vx-demo.now.sh/axis -- but this is way over-complicated.

loichuder:
As this happens only at the right, we can perhaps add a small `padding-right` so that the axis label does not overflow.
It would imply slight changes to the aspect ratio part I think.

axelboc:
Yeah, a bit of padding-right on `visArea` would work. We could keep the canvas sizing computations as they are now by adding an extra `div` inside `visArea`.

axelboc:
The problem occurs at the top of the left axis as well, but it's less noticeable because the heatmap is flush against the tab bar.

loichuder:
I am working on a surprise PR for this ;)

PR #128, opened 2020-05-07
Add top and right padding to visualisation area

loichuder:
This has several advantages:
- Fixing #104 
- Generalizing the axis placement: putting xAxis on the top or yAxis on the right can be done more easily.
- Adding some breathing room between the canvas and the toolbar:
    * Previously:
image
    * Now:
image


loichuder:
Sorry, I accidentally pushed this on master without squashing...

PR #130, opened 2020-05-07
Add top and right padding to visualisation area

loichuder:
Reopening of #128 due to my mistaken push.

This has several advantages:
- Fixing #104 
- Generalizing the axis placement: putting xAxis on the top or yAxis on the right can be done more easily.
- Adding some breathing room between the canvas and the toolbar:
    * Previously:
image
    * Now:
image

### DIFF
--- a/src/h5web/visualizations/shared/models.ts
+++ b/src/h5web/visualizations/shared/models.ts
@@ -4,11 +4,18 @@ export type Domain = [number, number];
 
 export type Size = { width: number; height: number };
 
-export type AxisOffsets = { left: number; bottom: number };
+export type AxisOffsets = {
+  left: number;
+  bottom: number;
+  right: number;
+  top: number;
+};
 
 export interface AxisDomains {
   left: Domain;
   bottom: Domain;
+  right?: number;
+  top?: number;
 }
 
 export interface TwoDimScale {
--- a/src/h5web/dataset-visualizer/VisDisplay.tsx
+++ b/src/h5web/dataset-visualizer/VisDisplay.tsx
@@ -10,11 +10,8 @@ import {
 } from '../visualizations';
 import { useValue } from '../providers/hooks';
 import HeatmapProvider from '../visualizations/heatmap/HeatmapProvider';
-import { AxisOffsets } from '../visualizations/shared/models';
 import { Dims } from '../visualizations/heatmap/models';
 import LineProvider from '../visualizations/line/LineProvider';
-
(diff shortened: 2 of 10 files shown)


loichuder:
In fact, I see the top and right padding as potential place holders for axes. The fact that it fixes #104 is a bonus.
Then, it was making sense to me to treat all axis offsets consistently and to separate `top` from `bottom` and `left` from `right`

axelboc:
Alright, I think I know where my concern comes from:

Now that we have abstracted `VisCanvas`, I think the offsets should no longer be provided in `HeatmapProvider` and `LineMap`. To me, they are an implementation detail of `VisCanvas` and should be moved back into a constant in that component - i.e. `const AXIS_OFFSETS = { horizontal: 36, vertical: 72 };`.

Thinking ahead about supporting top/right axis, I think a better API for `VisCanvas` would be to rely on the domains declared in `axisDomains` -- if `top` is provided, then `VisCanvas` takes care of adding the top axis and appropriate padding.

Now for #104 specifically, your solution works, but I would just compute the final `axisOffset` object dynamically inside `VisCanvas` based on:
- `axisDomains`,
- the `AXIS_OFFSETS` constant above, 
- a fallback offset to use when an axis is not present (e.g. `const AXIS_FALLBACK_OFFSET = 10;`)

So something like (probably improvable):

```ts
const axisOffsets = {
  left: axisDomains.left ? AXIS_OFFSETS.vertical : AXIS_FALLBACK_OFFSET,
  right: axisDomains.right ? AXIS_OFFSETS.vertical : AXIS_FALLBACK_OFFSET,
  top: axisDomains.top ? AXIS_OFFSETS.horizontal : AXIS_FALLBACK_OFFSET,
  bottom: axisDomains.bottom ? AXIS_OFFSETS.horizontal : AXIS_FALLBACK_OFFSET,
}
```

`axisOffsets` can then be used to compute `availableSize` and passed down to `AxisSystem` as it is now.

src/h5web/visualizations/shared/models.ts
```diff
@@ -4,11 +4,18 @@ export type Domain = [number, number];
...
+export type AxisOffsets = {
+  left: number;
+  bottom: number;
+  right: number;
+  top: number;
+};
 
 export interface AxisDomains {
   left: Domain;
   bottom: Domain;
```
axelboc, line 16: ils sont tous optionnels, du coup

Issue #119, opened 2020-05-06
[Line] Add a tooltip constrained to data


loichuder:

loichuder:
Partial implementation in #135

Issue #74, opened 2020-04-14
[Line] Log scale for axis


axelboc:

axelboc:
Perhaps a stupid question: is this relevant at all to the dimensional axis (i.e. the `x` axis)?

loichuder:
Not a stupid question: it really depends what you are displaying on the axis in question. This has to be solved with #73 but I think the possibility should exist on both axes.

loichuder:
#133 implemented a [SymLog](https://www.researchgate.net/publication/233967063_A_bi-symmetric_log_transformation_for_wide-range_data) scale as it was simpler to implement for data that contain negative or null values. 

We need to decide how to implement the regular log scale that only works for strictly positive values. Possibilities include:
1. Deactivate the log scale button when the data contain negative or null values.
2. Display only the positive values on which the log transformation was able to be done.

axelboc:
3. Choose dynamically which log scale to apply and make that clear to the user.

But as discussed, perhaps the symlog scale is a satisfactory solution for the viewer, as users can export the data and run their own analysis.

PR #131, opened 2020-05-11
Refactored sceneToAxis scales computation

loichuder:
This PR addresses the problem that the scales linking the scene coordinates (R3F) to axis domains were computed separately for the `LineVis` and `AxisSystem`. I foresaw problems due to this for future `LineVis` developments (#119, #74).

In short:
* The computation of scaleToAxis scales was moved in a shared function in `visualisations/shared/utils`.
* `LineVis` now uses the output of this function to do axisToScale computations thanks to `invert`.
* The state of `AxisSystem` was revert to axisDomains rather than scales as `computeAxisScales` was not shared anyway.

### DIFF
--- a/src/h5web/visualizations/line/hooks.ts
+++ b/src/h5web/visualizations/line/hooks.ts
@@ -1,10 +1,9 @@
 import { useContext, useMemo } from 'react';
 import { Vector3 } from 'three';
-import { scaleLinear } from 'd3-scale';
 import { useThree } from 'react-three-fiber';
 import { LineProps, LineContext } from './LineProvider';
 import { AxisDomains } from '../shared/models';
-import { extendDomain, findDomain } from '../shared/utils';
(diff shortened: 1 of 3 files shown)


src/h5web/visualizations/line/hooks.ts
```diff
@@ -31,22 +30,19 @@ export function useAxisDomains(): AxisDomains | undefined {
...
     }
 
-    const xScale = scaleLinear()
-      .domain(axisDomains.bottom)
-      .range([-width / 2, width / 2]);
-    const yScale = scaleLinear()
-      .domain(axisDomains.left)
-      .range([-height / 2, height / 2]);
+    // sceneToAxis is used to get axisToScene through invert
+    const sceneToAxis = sceneToAxisScales(size, axisDomains);
```
axelboc, line 41: the comment needs a rewrite, I think, since `axisToScene` no longer exists

PR #132, opened 2020-05-11
Fix default path processing

axelboc:
The default path set with `REACT_APP_DEFAULT_PATH` no longer prevents navigating the explorer tree.

### DIFF
--- a/src/h5web/explorer/Explorer.tsx
+++ b/src/h5web/explorer/Explorer.tsx
@@ -1,11 +1,12 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { FiFileText } from 'react-icons/fi';
 import { HDF5Link } from '../providers/models';
-import { TreeNode } from './models';
+import { TreeNode, ExpandedNodes } from './models';
 import TreeView from './TreeView';
 import styles from './Explorer.module.css';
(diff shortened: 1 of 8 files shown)


src/h5web/explorer/Explorer.tsx
```diff
@@ -22,11 +23,37 @@ function Explorer(props: Props): JSX.Element {
   const domain = useDomain();
   const tree = useMetadataTree();
 
+  const [expandedNodes, setExpandedNodes] = useState<ExpandedNodes>({});
```
axelboc, line 26: I moved the `expandedNodes` state, which used to be in `TreeView`, up in the `Explorer`. So it now manages the expanded state of every node in the tree. This allows nodes on the default path to be expanded all at once, and only once, as soon as the tree has been retrieved (cf. `useEffect` below).

src/h5web/explorer/Explorer.tsx
```diff
@@ -58,10 +88,11 @@ function Explorer(props: Props): JSX.Element {
 
       {tree.children && (
         <TreeView
+          level={0}
```
axelboc, line 91: Remnant from an initial attempt, which I kept because it now seems cleaner: the nesting level is now managed in `TreeView` instead of in the nodes (`TreeNode`).

PR #133, opened 2020-05-11
Implemented y SymLog scale for LineVis

loichuder:
image
x log scale can be implemented along the same lines

### DIFF
--- a/src/h5web/visualizations/line/DataLine.tsx
+++ b/src/h5web/visualizations/line/DataLine.tsx
@@ -1,15 +1,22 @@
-import React, { useCallback } from 'react';
+import React from 'react';
+import { useUpdate } from 'react-three-fiber';
+import { BufferGeometry } from 'three';
 import { useDataPoints } from './hooks';
 
 function DataLine(): JSX.Element {
   const points = useDataPoints();
-  const setFromPoints = useCallback(self => self.setFromPoints(points), [
-    points,
-  ]);
+  const ref = useUpdate(
+    (geometry: BufferGeometry) => {
+      if (points) {
+        geometry.setFromPoints(points);
+      }
+    },
+    [points]
+  );
 
   return (
     <line>
-      {points && <bufferGeometry attach="geometry" onUpdate={setFromPoints} />}
+      <bufferGeometry attach="geometry" ref={ref} />
       <lineBasicMaterial attach="material" color="#1b998b" linewidth={2} />
     </line>
   );
--- a/src/h5web/visualizations/line/LineToolbar.tsx
+++ b/src/h5web/visualizations/line/LineToolbar.tsx
@@ -4,14 +4,24 @@ import Toggler from '../shared/Toggler';
 import { useLineConfig } from './config';
 
 function LineToolbar(): JSX.Element {
-  const [showGrid, toggleGrid] = useLineConfig(
-    state => [state.showGrid, state.toggleGrid],
+  const [showGrid, toggleGrid, hasYLogScale, toggleYLogScale] = useLineConfig(
+    state => [
+      state.showGrid,
+      state.toggleGrid,
+      state.hasYLogScale,
+      state.toggleYLogScale,
+    ],
     shallow
   );
 
   return (
     <>
       <Toggler label="Show grid" value={showGrid} onChange={toggleGrid} />
+      <Toggler
+        label="Y SymLog scale"
+        value={hasYLogScale}
+        onChange={toggleYLogScale}
+      />
     </>
   );
 }
--- a/src/h5web/visualizations/line/LineVis.tsx
+++ b/src/h5web/visualizations/line/LineVis.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import shallow from 'zustand/shallow';
 import styles from './LineVis.module.css';
 import DataLine from './DataLine';
 import VisCanvas from '../shared/VisCanvas';
@@ -9,13 +10,20 @@ import { useLineConfig } from './config';
 
 function LineVis(): JSX.Element {
   const props = useProps();
-  const showGrid = useLineConfig(state => state.showGrid);
+  const [showGrid, hasYLogScale] = useLineConfig(
+    state => [state.showGrid, state.hasYLogScale],
+    shallow
+  );
 
   const axisDomains = useAxisDomains();
 
   return (
     <div className={styles.root}>
-      <VisCanvas axisDomains={axisDomains} showGrid={showGrid}>
+      <VisCanvas
+        axisDomains={axisDomains}
+        showGrid={showGrid}
+        hasYLogScale={hasYLogScale}
+      >
         {/* Provide context again - https://github.com/react-spring/react-three-fiber/issues/262 */}
         <LineProvider {...props}>
           <PanZoomMesh />
--- a/src/h5web/visualizations/line/hooks.ts
+++ b/src/h5web/visualizations/line/hooks.ts
@@ -1,9 +1,13 @@
 import { useContext, useMemo } from 'react';
 import { Vector3 } from 'three';
-import { useThree } from 'react-three-fiber';
 import { LineProps, LineContext } from './LineProvider';
 import { AxisDomains } from '../shared/models';
-import { extendDomain, findDomain, sceneToAxisScales } from '../shared/utils';
+import {
+  extendDomain,
+  findDomain,
+  useAbscissaScale,
+  useOrdinateScale,
+} from '../shared/utils';
 
 export function useProps(): LineProps {
   const props = useContext(LineContext);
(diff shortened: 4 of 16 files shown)


src/h5web/visualizations/line/DataLine.tsx
```diff
@@ -1,15 +1,22 @@
...
-    points,
-  ]);
+  const ref = useUpdate(
+    (geometry: BufferGeometry) => {
+      if (points) {
+        geometry.setFromPoints(points);
+      }
+    },
+    [points]
+  );
```
loichuder, line 15: This was changed as the line plot was not updating whereas `points` was changed.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -9,13 +10,20 @@ import { useLineConfig } from './config';
...
 
   const axisDomains = useAxisDomains();
 
   return (
     <div className={styles.root}>
-      <VisCanvas axisDomains={axisDomains} showGrid={showGrid}>
+      <VisCanvas
+        axisDomains={axisDomains}
+        showGrid={showGrid}
+        yLogScale={yLogScale}
```
loichuder, line 25: `yLogScale` is passed to `VisCanvas` that passes it to `AxisSystem`

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -61,6 +57,14 @@ function AxisSystem(props: Props): JSX.Element {
     ),
   };
 
+  const xAxisScale = axisToCanvas.x.copy();
+  xAxisScale.domain(visibleDomains.bottom);
+  xAxisScale.range([0, width]);
+
+  const yAxisScale = axisToCanvas.y.copy();
+  yAxisScale.domain(visibleDomains.left);
+  yAxisScale.range([height, 0]);
```
loichuder, line 66: Visible axis scales are copied from axisToCanvas scales so that it uses the appropriate linear or log scale.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -43,19 +44,13 @@ export function findDomain(data: number[]): Domain | undefined {
...
-  };
+export function axisToDimsScale(
+  axisDomain: Domain,
+  dimsRange: Domain,
+  log?: boolean
+): DataScale {
+  const scale = log ? scaleSymlog() : scaleLinear();
+  scale.domain(axisDomain);
+  scale.range(dimsRange);
+  return scale;
```
loichuder, line 55: A simple wrapper to inline the scale creation

src/h5web/visualizations/line/hooks.ts
```diff
@@ -30,19 +31,23 @@ export function useAxisDomains(): AxisDomains | undefined {
 export function useDataPoints(): Vector3[] | undefined {
   const { data } = useProps();
   const axisDomains = useAxisDomains();
-  const { size } = useThree();
+  const {
+    size: { width, height },
+  } = useThree();
```
axelboc, line 36: Not sure what you think, but I tend to keep this kind of nested destructuring on separate lines because the way Prettier does it is quite ugly :(

```
const { size } = useThree();
```

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -4,49 +4,49 @@ import { AxisLeft, AxisBottom } from '@vx/axis';
...
-  const { axisDomains, axisOffsets, showGrid } = props;
+  const { axisDomains, axisOffsets, showGrid, hasYLogScale } = props;
 
   const { camera, size } = useThree();
   const { position, zoom } = camera;
   const { width, height } = size;
 
-  const [domains, setDomains] = useState<AxisDomains>({
+  // axisDomains are the complete domains. visibleDomains change with the camera
+  const [visibleDomains, setVisibleDomains] = useState<AxisDomains>({
```
axelboc, line 26: great renaming 💯

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -4,49 +4,49 @@ import { AxisLeft, AxisBottom } from '@vx/axis';
...
-  const axisScales = {
-    x: scaleLinear()
-      .domain(domains.bottom)
-      .range([0, width]),
-    y: scaleLinear()
-      .domain(domains.left)
-      .range([height, 0]),
+  const axisToCanvas = {
+    x: axisToDimsScale(axisDomains.bottom, [-width / 2, width / 2]),
+    y: axisToDimsScale(
```
axelboc, line 33: oh just realised that we use the term `dims` for the dimensions of the dataset -- perhaps `axisToSize` or `axisToScene`?

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -4,49 +4,49 @@ import { AxisLeft, AxisBottom } from '@vx/axis';
...
-  const sceneToAxis = sceneToAxisScales(size, axisDomains);
-
-  const axisScales = {
-    x: scaleLinear()
-      .domain(domains.bottom)
-      .range([0, width]),
-    y: scaleLinear()
-      .domain(domains.left)
-      .range([height, 0]),
+  const axisToCanvas = {
```
axelboc, line 31: What do you think of `axisBoundsScales` or even splitting this in two (`xAxisBoundsScale` and `yAxisBoundsScale`) ... or even just `xBoundsScale` and then `xScale` further down, for consistency with the `useDataPoints` hook?

src/h5web/visualizations/line/hooks.ts
```diff
@@ -30,19 +34,14 @@ export function useAxisDomains(): AxisDomains | undefined {
 export function useDataPoints(): Vector3[] | undefined {
   const { data } = useProps();
   const axisDomains = useAxisDomains();
-  const { size } = useThree();
+  const xScale = useAbscissaScale(axisDomains ? axisDomains.bottom : undefined);
```
axelboc, line 37: `axisDomains && axisDomains.bottom`

src/h5web/visualizations/line/hooks.ts
```diff
@@ -30,19 +34,14 @@ export function useAxisDomains(): AxisDomains | undefined {
...
+  const yScale = useOrdinateScale(axisDomains ? axisDomains.left : undefined);
 
   return useMemo(() => {
-    if (axisDomains === undefined) {
+    if (xScale === undefined || yScale === undefined) {
```
axelboc, line 41: `!xScale || !yScale`

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -61,6 +58,14 @@ function AxisSystem(props: Props): JSX.Element {
     ),
   };
 
+  const xAxisScale = axisToCanvas.x.copy();
```
axelboc, line 61: To avoid copying the scale, you could add a hook called `useAbscissaScaleFn` and use it here and inside `useAbscissaScale`.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -1,52 +1,49 @@
...
-  const axisScales = {
-    x: scaleLinear()
-      .domain(domains.bottom)
-      .range([0, width]),
-    y: scaleLinear()
-      .domain(domains.left)
-      .range([height, 0]),
+  const axisToCanvas = {
+    x: useAbscissaScale(axisDomains.bottom) as DataScale,
+    y: useOrdinateScale(axisDomains.left) as DataScale,
```
axelboc, line 34: Why not merge the two hooks into one and return and return an object with `x` and `y` properties for consistency?

axelboc: Maybe not, I don't know. I just feel like we need to pick whether we use objects with `x` and `y` or individual variables and do so everywhere.

src/h5web/visualizations/shared/VisCanvas.tsx
```diff
@@ -66,14 +76,19 @@ function VisCanvas(props: Props): JSX.Element {
...
-                axisDomains={axisDomains}
-                axisOffsets={axisOffsets}
-                showGrid={showGrid}
-              />
-            )}
-            {children}
+            <AxisSystemProvider
+              showGrid={showGrid}
+              hasXLogScale={hasXLogScale}
+              hasYLogScale={hasYLogScale}
```
axelboc, line 82: We could pass the domains as well so the `useAbscissa` and `useOrdinate` hooks can retrieve them themselves. This might be useful for the tooltip and other features as well.

src/h5web/visualizations/line/LineToolbar.tsx
```diff
@@ -4,14 +4,38 @@ import Toggler from '../shared/Toggler';
...
   );
 
   return (
     <>
       <Toggler label="Show grid" value={showGrid} onChange={toggleGrid} />
+      <Toggler
+        label="X log scale"
+        value={hasXLogScale}
+        onChange={toggleXLogScale}
+      />
```
axelboc, line 33: A log scale on the x axis makes no sense at the moment, not until we implement axis-to-dimension mapping, so I would remove it from the toolbar. We'll have to show/hide the buttons in a smart way so users don't try to log an index-based axis.

loichuder:
We ended up refactoring the scale handling once more (#131). The scales are SymLog (which is NOT the same as regular log) and this is now clear in the label of the togglers (Heatmap included).

PR #134, opened 2020-05-12
Add some tests and implement mock data provider

axelboc:
- Add some utilities tests.
- Implement mock provider to avoid making real requests. This also means we can now test `App` instead of `DemoApp`. I took a snapshot of the `metadata` and `values` properties stored by the `SilxApi` and removed the big datasets to avoid slowing down the tests.
- Improve our default `App` test to look for a few elements on the page.


src/h5web/App.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+    <MockProvider domain={DOMAIN}>
+      <App />
+    </MockProvider>
+  );
+
+  const noVisText = await findByText('Nothing to visualize');
+  expect(noVisText).toBeInTheDocument();
+
+  const title = await findByRole('heading', { name: DOMAIN });
+  expect(title).toBeInTheDocument();
```
axelboc, line 19: I've changed the breadcrumbs element to be a heading so I can query it in the tests.

axelboc: What's you're reasoning? Semantically, a `header` element would have to contain a heading element (i.e. `h1`) anyway, so I went for the simplest.

loichuder: My reasoning was the change from the regular `div` to a `header` instead of `h1` to avoid the overwritting of the styling of `h1` in the CSS. But it is fine like you did.

PR #136, opened 2020-05-13
Add application tests

axelboc:
Well, this was not trivial. I ran into versioning issues with the testing library, a bug when toggling the explorer panel, and plenty of uncertainties around when to use the testing library's async utilities...

I think the result takes us in the right direction with regard to testing, though.

Some of the documentation pages that helped me through this:

- https://testing-library.com/docs/guide-which-query - a god overview of all the queries avaialble
- https://testing-library.com/docs/dom-testing-library/api-queries - which query function throws, which is async, how to query multiple elements, etc.
- https://testing-library.com/docs/dom-testing-library/api-queries#byrole - the handiest query IMO
- https://github.com/A11yance/aria-query#elements-to-roles - if an element doesn't have an explicit `role` attribute, you may still be able to query it `ByRole` if it has one of the implicit roles in the list.
- https://testing-library.com/docs/react-testing-library/cheatsheet - explains that you should use the query functions returned by `render` instead of the global ones, as they are conveniently bound to the React container inside which you've just rendered your app
- https://testing-library.com/docs/dom-testing-library/api-events - `fireEvent.click()` and the like
- https://testing-library.com/docs/dom-testing-library/api-async - `waitFor*` functions to deal with async stuff other than an element appearing (use `findBy*` for that)
- https://testing-library.com/docs/guide-disappearance - explains how to deal with elements that appear/disappear
- https://github.com/testing-library/jest-dom - all the matchers you can use on top of the ones provided by Jest (e.g. `toBeInTheDocument()`, `toBeVisible()`, etc.)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,6 +19,7 @@
     ],
     "import/no-unresolved": "off",
     "import/no-webpack-loader-syntax": "off",
+    "jest/no-focused-tests": "warn",
     "react/jsx-props-no-spreading": "off",
     "prettier/prettier": "off"
   },
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
+    "test": "react-scripts test --env=jsdom-fourteen",
     "lint": "run-p lint:*",
     "lint:eslint": "eslint \"**/*.{ts,tsx,js,jsx}\"",
     "lint:tsc": "tsc --noEmit",
@@ -49,9 +49,9 @@
     "zustand": "^2.2.3"
   },
   "devDependencies": {
-    "@testing-library/jest-dom": "^4.2.4",
-    "@testing-library/react": "^9.5.0",
-    "@testing-library/user-event": "^7.2.1",
+    "@testing-library/jest-dom": "^5.7.0-beta.1",
+    "@testing-library/react": "^10.0.4",
+    "@testing-library/user-event": "^10.2.0",
     "@types/d3-array": "^2.0.0",
     "@types/d3-color": "^1.2.2",
     "@types/d3-format": "^1.3.1",
@@ -73,7 +73,7 @@
     "npm-run-all": "^4.1.5",
     "prettier": "^1.19.1",
     "react-reflex": "^3.0.22",
-    "react-scripts": "3.4.0",
+    "react-scripts": "^3.4.1",
     "typescript": "^3.8.2",
     "worker-loader": "^2.0.0"
   },
--- a/src/h5web/App.test.tsx
+++ b/src/h5web/App.test.tsx
@@ -1,23 +1,134 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import {
+  render,
+  RenderResult,
+  fireEvent,
+  waitFor,
+} from '@testing-library/react';
 import App from './App';
 import MockProvider from './providers/mock/MockProvider';
 
 const DOMAIN = 'bsa_002_000-integrate-sub';
 
-test('renders with root group selected', async () => {
-  const { findByText, findByRole } = render(
+function renderApp(): RenderResult {
+  return render(
     <MockProvider domain={DOMAIN}>
       <App />
     </MockProvider>
   );
+}
+
+test('renders with root group selected', async () => {
+  const { findByText, getByRole } = renderApp();
 
   const noVisText = await findByText('Nothing to visualize');
-  expect(noVisText).toBeInTheDocument();
+  expect(noVisText).toBeVisible();
 
-  const title = await findByRole('heading', { name: DOMAIN });
-  expect(title).toBeInTheDocument();
+  const title = getByRole('heading', { name: DOMAIN });
+  expect(title).toBeVisible();
+
+  const domainBtn = getByRole('treeitem', { name: DOMAIN });
+  expect(domainBtn).toBeVisible();
+  expect(domainBtn).toHaveAttribute('aria-selected', 'true');
+});
+
+test('switch between "display" and "inspect" modes', async () => {
+  const {
+    findByRole,
+    findByText,
+    getByRole,
+    queryByRole,
+    queryByText,
+  } = renderApp();
+
+  const inspectBtn = getByRole('tab', { name: 'Inspect' });
+  const displayBtn = getByRole('tab', { name: 'Display' });
+
+  // Switch to "inspect" mode
+  fireEvent.click(inspectBtn);
+
+  const noVisText1 = queryByText('Nothing to visualize');
+  expect(noVisText1).not.toBeInTheDocument();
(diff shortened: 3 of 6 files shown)


.eslintrc.json
```diff
@@ -19,6 +19,7 @@
     ],
     "import/no-unresolved": "off",
     "import/no-webpack-loader-syntax": "off",
+    "jest/no-focused-tests": "warn",
```
axelboc, line 22: So we can write `test.only()` when developing.

package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
+    "test": "react-scripts test --env=jsdom-fourteen",
```
axelboc, line 18: Will be removed when `react-scripts` v4 is released, as it will include Jest v26, which includes a newer version of JSDOM, which includes a shim of `MutationObserver`, which is required by `testing-library`.........

src/h5web/App.test.tsx
```diff
@@ -1,23 +1,134 @@
...
     <MockProvider domain={DOMAIN}>
       <App />
     </MockProvider>
   );
+}
+
+test('renders with root group selected', async () => {
+  const { findByText, getByRole } = renderApp();
 
   const noVisText = await findByText('Nothing to visualize');
```
axelboc, line 24: To get an element, try one of the `getBy*` methods first. If it doesn't work, it means the element appears asynchronously, so use `await findBy*` instead.

An element can appear asynchronously when the app re-renders because of an asynchronous state change, which can happen because of a `useEffect` or an HTTP request, notably.

src/h5web/App.test.tsx
```diff
@@ -1,23 +1,134 @@
...
+  const sidebarBtn = getByRole('button', {
+    name: 'Toggle explorer sidebar',
+  });
+
+  expect(domainBtn).toBeVisible();
+  expect(sidebarBtn).toHaveAttribute('aria-pressed', 'true');
+
+  fireEvent.click(sidebarBtn);
+  expect(domainBtn).not.toBeVisible();
+  expect(sidebarBtn).toHaveAttribute('aria-pressed', 'false');
```
axelboc, line 79: The bug I fixed was identified by this line. The main area was being unmounted and remounted, which lead to variable `sidebarBtn` referencing an obsolete element. I could have re-queried the button after each click, but fixing the problem in the code was more logical, of course.

src/h5web/App.test.tsx
```diff
@@ -1,23 +1,134 @@
...
+  expect(groupBtn).toHaveAttribute('aria-expanded', 'true'); // remains expanded as it wasn't previously selected
+  expect(childGroupBtn).toHaveAttribute('aria-selected', 'false');
+
+  // Collapse `entry_0000` group
+  fireEvent.click(groupBtn);
+
+  expect(queryByRole('treeitem', { name: '1_cormap' })).not.toBeInTheDocument();
+
+  // `waitFor` avoids warning due to async work occuring after completion of test
+  await waitFor(() => {
```
axelboc, line 130: If something asynchronous happens after performing an action (with `fireEvent.click()` for instance), you have three options:

- if you know an element will be added to the DOM, use `await findBy*(...)`;
- if you know an element will be removed from the DOM, use `await waitForElementToBeRemoved(() => queryBy*(...))`
- in all other cases, use `await waitFor(() => { /* something that throws until the async operation has completed, like `expect or `getBy` */ })`

In this particular case, as the comment states, `waitFor` is only required because some async work that I don't really care about occurs.

src/h5web/App.test.tsx
```diff
@@ -1,23 +1,134 @@
...
     </MockProvider>
   );
+}
+
+test('renders with root group selected', async () => {
+  const { findByText, getByRole } = renderApp();
 
   const noVisText = await findByText('Nothing to visualize');
-  expect(noVisText).toBeInTheDocument();
+  expect(noVisText).toBeVisible();
```
axelboc, line 25: `toBeVisible` is better than `toBeInTheDocument` as it checks that the element isn't hidden with CSS or by other means. That being said, it's also more computationally heavy as it goes through all the parents to make sure none of them are hidden, so I wouldn't recommend using it every time.

src/h5web/App.test.tsx
```diff
@@ -1,23 +1,134 @@
...
 
-  const title = await findByRole('heading', { name: DOMAIN });
-  expect(title).toBeInTheDocument();
+  const title = getByRole('heading', { name: DOMAIN });
```
axelboc, line 27: If the element you're looking for has a specific semantic, then it's better to look for it with `getByRole` (or `findByRole` or `queryByRole`).

src/h5web/App.test.tsx
```diff
@@ -1,23 +1,134 @@
...
+  } = renderApp();
+
+  const inspectBtn = getByRole('tab', { name: 'Inspect' });
+  const displayBtn = getByRole('tab', { name: 'Display' });
+
+  // Switch to "inspect" mode
+  fireEvent.click(inspectBtn);
+
+  const noVisText1 = queryByText('Nothing to visualize');
+  expect(noVisText1).not.toBeInTheDocument();
```
axelboc, line 51: `queryByText` doesn't throw if the element isn't in the DOM, which is useful to expect this very fact.

axelboc: if the element disappears asynchronously, see my comment further down

src/h5web/App.tsx
```diff
@@ -41,7 +41,9 @@ function App(): JSX.Element {
           <Explorer selectedNode={selectedNode} onSelect={setSelectedNode} />
         </ReflexElement>
 
-        {isExplorerOpen && <ReflexSplitter />}
```
axelboc, line 44: This line was triggering a re-mount of everything below... 🤷‍♂️

src/h5web/App.test.tsx
```diff
@@ -1,23 +1,134 @@
...
+  fireEvent.click(sidebarBtn);
+  expect(domainBtn).not.toBeVisible();
+  expect(sidebarBtn).toHaveAttribute('aria-pressed', 'false');
+
+  fireEvent.click(sidebarBtn);
+  expect(domainBtn).toBeVisible();
+  expect(sidebarBtn).toHaveAttribute('aria-pressed', 'true');
+});
+
+test('navigate groups in explorer', async () => {
```
loichuder, line 86: Would it make sense/be possible to put this test in a file `Explorer.test.tsx` as it tests only `Explorer` features ?

axelboc: Yeah, I did consider it, but it would require rendering the `Explorer` component instead of the whole app. Conceptually, I feel like rendering the whole app and testing bits of it is safer and more valuable for now, at least until we start exporting the components.

loichuder: :ok_hand:

PR #137, opened 2020-05-14
Added dataset tests and some utils tests

loichuder:


src/h5web/App.test.tsx
```diff
@@ -126,9 +126,115 @@ test('navigate groups in explorer', async () => {
...
 
-  // `waitFor` avoids warning due to async work occuring after completion of test
+  // `waitFor` avoids warning due to async work occurring after completion of test
   await waitFor(() => {
     expect(groupBtn).toHaveAttribute('aria-selected', 'true');
     expect(groupBtn).toHaveAttribute('aria-expanded', 'false');
   });
 });
+
+test('select a scalar dataset and display the value', async () => {
```
axelboc, line 136: I was trying to write these from the point of view of the user -- trying to identify their purpose when using the app -- so `visualize scalar dataset` would make more sense, I think.

src/h5web/App.test.tsx
```diff
@@ -126,9 +126,115 @@ test('navigate groups in explorer', async () => {
...
+  const { findByRole, getByRole, getByText } = renderApp();
+
+  // Select the scalar dataset entry_0000/1_cormap/version
+  fireEvent.click(await findByRole('treeitem', { name: 'entry_0000' }));
+  fireEvent.click(getByRole('treeitem', { name: '1_cormap' }));
+  const datasetBtn = getByRole('treeitem', { name: 'version' });
+  expect(datasetBtn).toBeVisible();
+  fireEvent.click(datasetBtn);
```
axelboc, line 144: Some blank lines would make things more readable -- around (groups of) simulated events especially.

axelboc: Actually, it might make sense to abstract this into a utility function, what do you think? Something like `navigateTo('entry_0000/1_cormap/version')` - or perhaps with an array.

src/h5web/App.test.tsx
```diff
@@ -126,9 +126,115 @@ test('navigate groups in explorer', async () => {
...
+
+  // Select the 1D dataset entry_0000/1_cormap/to_merge
+  fireEvent.click(await findByRole('treeitem', { name: 'entry_0000' }));
+  fireEvent.click(getByRole('treeitem', { name: '1_cormap' }));
+  const datasetBtn = getByRole('treeitem', { name: 'to_merge' });
+  expect(datasetBtn).toBeVisible();
+  fireEvent.click(datasetBtn);
+
+  const visSelector = await findByRole('tablist', { name: 'Visualization' });
+  expect(visSelector).toBeVisible();
```
axelboc, line 167: I don't know that these two lines are worth having. You're already testing the tabs below.

src/h5web/App.test.tsx
```diff
@@ -126,9 +126,115 @@ test('navigate groups in explorer', async () => {
...
+  expect(matrixTab).toBeVisible();
+  expect(matrixTab).toHaveAttribute('aria-selected', 'false');
+
+  // Switching visualisations
+  fireEvent.click(matrixTab);
+  expect(matrixTab).toHaveAttribute('aria-selected', 'true');
+  expect(lineTab).toHaveAttribute('aria-selected', 'false');
+});
+
+test('switch visualisations for a 2D dataset', async () => {
```
axelboc, line 183: Not sure this one is very valuable since you've already tested that the vis selector works and we have utility tests that check that supported visualizations are computed correctly. How about testing the matrix vis, perhaps? You could check the number of rows/cols for instance.

Then, the previous test could be renamed simply `switch visualization`, and you could add another test to check the matrix vis with a 1D dataset.

src/h5web/App.test.tsx
```diff
@@ -126,9 +126,115 @@ test('navigate groups in explorer', async () => {
...
+  expect(shapeRow1).toBeVisible();
+
+  // Switch to "display" mode
+  fireEvent.click(displayBtn);
+
+  const shapeRow2 = queryByRole('row', { name: /Shape/ });
+  expect(shapeRow2).not.toBeInTheDocument();
+
+  const visSelector2 = await findByRole('tablist', { name: 'Visualization' });
+  expect(visSelector2).toBeVisible();
```
axelboc, line 239: No need to switch back I think, that's already tested.

src/h5web/visualizations/heatmap/utils.test.ts
```diff
@@ -0,0 +1,20 @@
...
+      .replace(')', '\\)');
+    const lastColor = interpolateRdBu(1)
+      .replace('(', '\\(')
+      .replace(')', '\\)');
+    const regexp = RegExp(
+      `linear-gradient\\(to top,${firstColor},.*${lastColor}\\)`
+    );
+    expect(regexp.test(generateCSSLinearGradient(interpolateRdBu, 'top'))).toBe(
+      true
+    );
```
axelboc, line 18: https://jestjs.io/docs/en/expect#expectstringmatchingstring--regexp

```
expect(result).toBe(expect.stringMatching(regex))
```

axelboc: That being said, to avoid the complex regex creation, you could just use a simple interpolator and hard code the gradient string returned with it in your test.

PR #138, opened 2020-05-14
Implements curve shown as points in LineVis

loichuder:
image

Uses `pointsMaterial` that are rendered as squares with any further tuning. 

To implement other glyphs, we might use  `pointsMaterial` with textures/sprites as shown in https://threejs.org/examples/#webgl_points_sprites and https://threejs.org/examples/#webgl_points_billboards. Another possibility are shaders: https://threejs.org/examples/#webgl_points_waves.


src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import shallow from 'zustand/shallow';
 import { format } from 'd3-format';
 import styles from './LineVis.module.css';
-import DataLine from './DataLine';
+import DataCurve from './DataCurve';
```
loichuder, line 5: Renamed to avoid confusion between the `Line` glyph and `DataLine`

src/h5web/visualizations/line/models.ts
```diff
@@ -0,0 +1,4 @@
+export enum Glyph {
+  Line = 'Line',
+  Square = 'Square',
+}
```
loichuder, line 4: Named `Glyph` to follow [VX's naming](https://vx-demo.now.sh/glyphs). Other possible names include: `Marker`, `DataSymbol`, `Symbol`...

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -0,0 +1,42 @@
...
+  if (glyph === Glyph.Line) {
+    return (
+      <line>
+        <bufferGeometry attach="geometry" ref={ref} />
+        <lineBasicMaterial attach="material" color={color} linewidth={2} />
+      </line>
+    );
+  }
+
+  if (glyph === Glyph.Square) {
```
loichuder, line 30: Was named `Square` as `pointsMaterial` is rendered as such and we might expand the possiblities to `Circle`, `Star`...

loichuder:
I think we should also implement a `Combined` option with the line and the points but I would do this on another PR.

PR #139, opened 2020-05-15
Added tooltip guides

loichuder:
image32

Vertical for `LineVis`. Both directions for `HeatmatVis`. Uses `Line` from `@vx/shape`

More fancy things such as constraining the guide to the data and adding a marker must wait until axes domains/scales are refactored.
As a consequence, it is not necessary to merge this right now.

### DIFF
--- a/src/h5web/visualizations/shared/TooltipMesh.tsx
+++ b/src/h5web/visualizations/shared/TooltipMesh.tsx
@@ -1,19 +1,22 @@
 import React, { ReactElement, useCallback } from 'react';
 import { Dom, PointerEvent, useThree } from 'react-three-fiber';
 import { TooltipWithBounds, useTooltip } from '@vx/tooltip';
+import { Line } from '@vx/shape';
 
 import styles from './TooltipMesh.module.css';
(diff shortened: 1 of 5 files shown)


src/h5web/visualizations/shared/TooltipMesh.tsx
```diff
@@ -1,19 +1,22 @@
...
 import styles from './TooltipMesh.module.css';
 import { useAbscissaScale, useOrdinateScale } from './utils';
 
 type Coords = [number, number];
+type Direction = 'horizontal' | 'vertical';
 
 interface Props {
   formatIndex: (t: Coords) => string;
   formatValue: (t: Coords) => string | undefined;
+  guides?: Direction[];
```
axelboc, line 15: This array allows you to write `['horizontal', horizontal']` or `['vertical', 'horizontal']`, so it's a bit too unclear I think. It'd be easier to have a single value: 'horizontal' | 'vertical' | 'both' or an object with `{ horizontal: true, vertical: true }, perhaps.

axelboc:
I think this can be merged before the scale refactor 😉

PR #140, opened 2020-05-18
Refactor VisCanvas API

axelboc:
The new API allows passing a config for each axis. It gets us one step closer to the axis-to-dimension mapping feature.

An axis config can declare either an "index" axis (i.e. ordinal) or a "data" axis (i.e. non-ordinal) -- this behaviour is based on the property used to pass the domain: `indexDomain` or `dataDomain`.

- A data axis config can have `isLog: true`, but not a index axis config.
- Both configs can have `showGrid: true`, but I haven't changed the current grid for now, so it's only visible if both axis configs have `showGrid: true`.
- There are no additional configuration options for now, although the goal will be to add `centerTicks` for index axis, and perhaps a `logType` enum for data axis, an `oppositeSide` boolean to move the axis to the top/right side, etc.

### DIFF
--- a/src/h5web/visualizations/line/LineVis.tsx
+++ b/src/h5web/visualizations/line/LineVis.tsx
@@ -5,27 +5,38 @@ import styles from './LineVis.module.css';
 import DataCurve from './DataCurve';
 import VisCanvas from '../shared/VisCanvas';
 import PanZoomMesh from '../shared/PanZoomMesh';
-import { useProps, useAxisDomains } from './hooks';
+import { useProps, useDataDomain } from './hooks';
 import LineProvider from './LineProvider';
 import { useLineConfig } from './config';
 import TooltipMesh from '../shared/TooltipMesh';
+import { extendDomain } from '../shared/utils';
 
 function LineVis(): JSX.Element {
   const props = useProps();
   const { data } = props;
+
   const [showGrid, hasYLogScale] = useLineConfig(
     state => [state.showGrid, state.hasYLogScale],
     shallow
   );
 
-  const axisDomains = useAxisDomains();
+  const dataDomain = useDataDomain();
+  if (!dataDomain) {
+    return <></>;
+  }
 
   return (
     <div className={styles.root}>
       <VisCanvas
-        axisDomains={axisDomains}
-        showGrid={showGrid}
-        hasYLogScale={hasYLogScale}
+        abscissaConfig={{
+          indexDomain: extendDomain([0, data.length - 1], 0.01),
+          showGrid,
+        }}
+        ordinateConfig={{
+          dataDomain: extendDomain(dataDomain, 0.01),
+          showGrid,
+          isLog: hasYLogScale,
+        }}
       >
         {/* Provide context again - https://github.com/react-spring/react-three-fiber/issues/262 */}
         <LineProvider {...props}>
--- a/src/h5web/visualizations/shared/AxisSystem.tsx
+++ b/src/h5web/visualizations/shared/AxisSystem.tsx
@@ -5,7 +5,7 @@ import { format } from 'd3-format';
 import { TickRendererProps } from '@vx/axis/lib/types';
 import { Grid } from '@vx/grid';
 import styles from './AxisSystem.module.css';
-import { AxisOffsets, AxisDomains } from './models';
+import { AxisOffsets, Domain } from './models';
 import { adaptedNumTicks, useAbscissaScale, useOrdinateScale } from './utils';
 import { AxisSystemContext } from './AxisSystemProvider';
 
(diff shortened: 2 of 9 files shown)


src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -18,14 +19,24 @@ function LineVis(): JSX.Element {
     shallow
   );
 
-  const axisDomains = useAxisDomains();
+  const dataDomain = useDataDomain();
```
axelboc, line 22: I refactored this hook to return only the data domain. The axis domains are now extended passed inline in the JSX. I think this is more readable with the new config, since we no longer have a prop that takes an `AxisDomains`.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -19,28 +19,28 @@ function AxisSystem(props: Props): JSX.Element {
...
-  const [visibleDomains, setVisibleDomains] = useState<AxisDomains>(
-    axisDomains
-  );
-
-  // Finds the visible domains from the camera FOV (zoom and position).
-  const { abscissaScale, abscissaScaleFn } = useAbscissaScale();
-  const { ordinateScale, ordinateScaleFn } = useOrdinateScale();
+  const [visibleDomains, setVisibleDomains] = useState<AxisDomains>({
+    x: abscissa.domain,
+    y: ordinate.domain,
```
axelboc, line 29: The hook now returns an object of type `AxisScale`, which includes the domain. The hook already retrieves the domain from the config appropriately, so it seemed smarter to just pass it on than to duplicate the logic here.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -100,7 +100,7 @@ function AxisSystem(props: Props): JSX.Element {
             />
           </svg>
         </div>
-        {showGrid && (
+        {abscissaConfig.showGrid && ordinateConfig.showGrid && (
```
axelboc, line 103: We still need `abscissaConfig` and `ordinateConfig` in this component because of the grid. We could put the grid in its own component at some point.

src/h5web/visualizations/shared/AxisSystemProvider.tsx
```diff
@@ -1,28 +1,24 @@
 import React, { createContext, ReactElement, ReactNode } from 'react';
-import { AxisDomains } from './models';
+import { AxisConfig } from './models';
 
 export interface DisplayAxisProps {
-  axisDomains: AxisDomains;
-  showGrid?: boolean;
-  hasXLogScale?: boolean;
-  hasYLogScale?: boolean;
```
axelboc, line 8: The log setting is no longer bound to x and y anymore. Each axis can be logged by providing a data axis config for it.

src/h5web/visualizations/shared/AxisSystemProvider.tsx
```diff
@@ -1,28 +1,24 @@
...
+  abscissaConfig: AxisConfig;
+  ordinateConfig: AxisConfig;
 }
 
-export const AxisSystemContext = createContext<DisplayAxisProps>({
-  axisDomains: { x: [0, 1], y: [0, 1] },
-});
+export const AxisSystemContext = createContext<DisplayAxisProps>(
+  {} as DisplayAxisProps
```
axelboc, line 10: I prefer the default state to be blatantly invalid and throw if used than to be valid and give us unexpected results silently.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -19,30 +19,30 @@ function AxisSystem(props: Props): JSX.Element {
...
 
   // axisDomains are the complete domains. visibleDomains change with the camera
-  const [visibleDomains, setVisibleDomains] = useState<AxisDomains>(
-    axisDomains
-  );
-
-  // Finds the visible domains from the camera FOV (zoom and position).
-  const { abscissaScale, abscissaScaleFn } = useAbscissaScale();
-  const { ordinateScale, ordinateScaleFn } = useOrdinateScale();
+  const [visibleDomains, setVisibleDomains] = useState<[Domain, Domain]>([
```
axelboc, line 27: I simplified the type of the state here since `AxisDomains` was no longer used anywhere else.

src/h5web/visualizations/shared/models.ts
```diff
@@ -1,8 +1,28 @@
...
 
+export type Size = { width: number; height: number };
+
 export type Domain = [number, number];
 
-export type Size = { width: number; height: number };
+export interface IndexAxisConfig {
+  indexDomain: Domain;
+  showGrid?: boolean;
+  isLog?: never; // invalid
```
axelboc, line 10: Because of how TypeScript works, `IndexAxisConfig | DataAxisConfig` doesn't prevent `isLog` from being declared in an index axis config. This line forces TypeScript to error, which helps users declare the config correctly.

If we add more properties to one config interface, we'll have to add it in this way to the other interface.

src/h5web/visualizations/shared/models.ts
```diff
@@ -1,8 +1,28 @@
...
+  isLog?: boolean;
+}
+
+export type AxisConfig = IndexAxisConfig | DataAxisConfig;
+
+export interface AxisScale {
+  scale: DataScale;
+  scaleFn: DataScaleFn;
+  domain: Domain;
+}
```
axelboc, line 25: This is the shape of the objects returned by the scale hooks. As explained, it now includes the domain for convenience.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -48,40 +55,48 @@ export function findDomain(data: number[]): Domain | undefined {
     : undefined;
 }
 
+export function isIndexAxisConfig(
+  config: AxisConfig
+): config is IndexAxisConfig {
+  return 'indexDomain' in config;
+}
```
axelboc, line 62: Type guard used to check whether an axis config declares an index or data axis.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -48,40 +55,48 @@ export function findDomain(data: number[]): Domain | undefined {
...
+  const { dataDomain, isLog = false } = config;
+  const scaleFn = getDataScaleFn(isLog);
 
-  const abscissaScaleFn = getDataScaleFn(!!hasXLogScale);
-  const abscissaScale = abscissaScaleFn();
-  abscissaScale.domain(axisDomains.x);
-  abscissaScale.range([-width / 2, width / 2]);
+  const scale = scaleFn();
+  scale.domain(dataDomain);
+  scale.range([-rangeSize / 2, rangeSize / 2]);
```
axelboc, line 85: The scale creation code looks a bit duplicated for index/data axis, but I kept it separate because it will probably diverge in the future, when we add `centerTicks`, notably.

src/h5web/visualizations/shared/models.ts
loichuder: `AxisScale` is becoming redundant: we have `scaleFn` and `domain` which can both derive from `scale`... In fact, we could imagine including `range` instead of `scale` and put the scale creation in the code using `AxisScale`: `scaleFn.domain(domain).range(range)`...

src/h5web/visualizations/shared/AxisSystem.tsx
loichuder: And use the context directly in the grid component. I agree.

src/h5web/visualizations/shared/models.ts
axelboc: Yeah, I did wonder about that, but in practice it's not amazing... TS would probably ask us to keep the domain/range calls on separate lines, we'd run into naming conflicts if destructuring, etc.

I think this will evolve again in the future, but I'm not sure how yet.

PR #143, opened 2020-05-19
Added sprite-based glyphs to LineVis

loichuder:
image

PR #145, opened 2020-05-19
Implements combined LineVis: Line and glyphs

loichuder:
image

* Removes the dropdrown to choose glyphs (#143) to avoid cluttering the interface. The glyph is forced to a cross.
* Adds a button group to choose between Line, Points and Line+Points. Implementation and design to be discussed.

PS: Tests are failing cause we have two tabs named `Line`. This might change with the implementation of `ButtonGroup`.


src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -1,40 +1,35 @@
...
-  );
-  const glyph = useLineConfig(state => state.glyph);
-  const color = '#1b998b';
+  const dataGeometry = useMemo(() => {
+    const geometry = new BufferGeometry();
+    if (points) {
+      geometry.setFromPoints(points);
+    }
+    return geometry;
+  }, [points]);
```
loichuder, line 17: This allows to compute the geometry only once and use it for both the line and point materials.

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -1,40 +1,35 @@
...
 
   return (
     <Suspense fallback={<></>}>
-      <points>
-        <bufferGeometry attach="geometry" ref={ref} />
-        <DataCurveMaterial glyph={glyph} color={color} />
-      </points>
+      {showGlyphs && <DataGlyphs geometry={dataGeometry} color={color} />}
+      {showLine && (
+        <Line geometry={dataGeometry}>
```
loichuder, line 29: I need to use `Line` from 'react-three-fiber/components' instead of the usual `line` otherwise TypeScript thinks it is a SVG line and refuses the `geometry` prop.

src/h5web/visualizations/shared/ButtonGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+  onClick: () => void;
+}
+
+interface Props {
+  buttons: Button[];
+  className?: string;
+  buttonClassName?: string;
+}
+
+function ButtonGroup(props: Props): JSX.Element {
```
loichuder, line 15: A tentative implementation of `ButtonGroup`: kept extremely dumb as styles and button functionalities are coming from the props.

src/h5web/visualizations/line/DataCurve.tsx
axelboc: So we recreate the geometry when the points change. Alright, not a big deal since it only happens when switching to a log scale. 👍

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -1,40 +1,35 @@
...
-        <bufferGeometry attach="geometry" ref={ref} />
-        <lineBasicMaterial attach="material" color={color} linewidth={2} />
-      </line>
-    );
-  }
+  const curveType = useLineConfig(state => state.curveType);
+  const showLine = curveType !== CurveType.OnlyGlyphs;
+  const showGlyphs = curveType !== CurveType.OnlyLine;
+
+  const color = '#1b998b';
```
axelboc, line 23: You can move this out of the function into a `COLOR` constant.

src/h5web/visualizations/line/DataGlyphs.tsx
```diff
@@ -0,0 +1,37 @@
...
+};
+
+interface Props {
+  geometry: BufferGeometry;
+  color: string;
+}
+
+function DataGlyphs(props: Props): JSX.Element {
+  const { geometry, color } = props;
+  const sprite = useLoader(TextureLoader, GLYPH_URLS[Glyph.Cross]);
```
axelboc, line 22: Thinking ahead, I think the glyph URL should be a prop, which defaults to `GLYPH_URLS[Glyph.Cross]`. This way, people can reuse it and pass their own sprite URL.

The enum for the type of glyph does not make much sense anymore. Perhaps we should change it to `GlyphUrls` directly (instead of having the URLs in a separate object). The way I see, we'd then export the enum so people can use it to pass a built-in sprite URL to `DataGlyphs` (or to a parent component that would forward the prop).

src/h5web/visualizations/shared/ButtonGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const { buttons, className, buttonClassName } = props;
+
+  if (buttons.length === 0) {
+    return <></>;
+  }
+
+  return (
+    <div role="tablist" className={className}>
```
axelboc, line 23: This is not quite a `tablist`, but more a `radiogroup`. See the text alignment options in this toolbar example: https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html

This will fix the tests.

src/h5web/visualizations/shared/ButtonGroup.tsx
axelboc: The toolbar as a whole needs a rethink, so it's perfect for now 💯

src/h5web/visualizations/line/DataGlyphs.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface Props {
+  geometry: BufferGeometry;
+  color: string;
+  glyphURL?: string;
+}
+
+function DataGlyphs(props: Props): JSX.Element {
+  const { geometry, color, glyphURL } = props;
+  const sprite = useLoader(TextureLoader, glyphURL || GLYPH_URLS.Cross);
```
axelboc, line 14: Pour info, tu peux aussi mettre `glyphURL = GLYPH_URLS.Cross` dans le destructuring directement -- mais c'est pareil.

src/h5web/visualizations/line/DataGlyphs.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { useLoader } from 'react-three-fiber';
+import { TextureLoader, BufferGeometry } from 'three';
+import { GLYPH_URLS } from './models';
```
axelboc, line 4: Vu que c'est un enum, maintenant, il faudrait repasser en PascalCase.

src/h5web/visualizations/line/models.ts
```diff
@@ -1,8 +1,15 @@
...
+import squareURL from './sprites/square.png';
+
+export enum CurveType {
+  OnlyLine = 'OnlyLine',
+  OnlyGlyphs = 'OnlyGlyphs',
+  LineAndGlyphs = 'LineAndGlyphs',
 }
 
-export type SpriteGlyph = Exclude<Glyph, Glyph.None>;
+export const GLYPH_URLS = {
```
axelboc, line 11: Ah, c'est pas un enum -- tu peux en faire un enum je pense.

PR #152, opened 2020-05-22
Show only integer ticks on index axes

### LINKED ISSUE #144: [Heatmap] Stop centering ticks
As per our discussion this morning, the ESRF's convention seems to be to keep the tick to the left of each "data pixel" and to not show a tick on the right of the last "data pixel" - i.e. `[0, cols[`.

For now we can just remove the `0.5` offset on the domain, but it may be worth making this configurable in the future, as other facilities may have different conventions.

### LINKED ISSUE #146: Too many ticks on index axis
Index axes currently show ticks at non-integer values when zoomed in or with small datasets. This is caused by the linear scale that is used, which is linked to the way VX `Axis` works.

Previously, we used to pass `numTicks` through `Math.min()` to get around the problem, but this is a bit hacky, since the numbers are a bit magic.

The current linear scale has another downside linked to #144: if we stop centering ticks on heatmaps, we'll end up with ticks on the right/top of the last "data pixels". What we want is more something like `scaleBand` or `scalePoint`, but with a linear input domain.

My recommendation would be to implement our own scale, which I think would be relatively easy to do.

axelboc:
Fixes #146. The main fix is to ask for a number of ticks that matches the number of integer values in the visible domain. This works very well in most cases.

Unfortunately, [`d3.ticks()`](https://github.com/d3/d3-array#ticks) (used internally by d3-scale) doesn't always respect the requested number of ticks. For instance, `d3.ticks(0, 4, 3)` outputs 5 ticks instead of the requested 3. This is easy to check on https://observablehq.com/@d3/d3-ticks. 

This weird behaviour can lead to decimal tick values in some cases. For instance, `d3.ticks(0, 2, 3)`  returns `[0, 0.5, 1, 1.5, 2]`.

The problem comes from `d3.tickIncrement` (or its equivalent, `d3.tickSteps`), which is used internally by `d3.ticks()`. To work around the issue, I re-implemented a simplified version of `d3.ticks()` that calls `d3.tickSteps()`, but then clamps the returned step so it's never lower than 1.

I've done my best to document this in `utils.test.ts`.

---

The PR also includes the following changes which were coupled to the problem at hand:

- The ticks are no longer centered on the heatmap, which fixes #144.
- The grid columns and rows are now controlled independently based on the `showGrid` flag of their respective axis configs.

### DIFF
--- a/src/h5web/visualizations/shared/AxisSystem.tsx
+++ b/src/h5web/visualizations/shared/AxisSystem.tsx
@@ -3,11 +3,21 @@ import { useThree, useFrame, Dom } from 'react-three-fiber';
 import { AxisLeft, AxisBottom } from '@vx/axis';
 import { format } from 'd3-format';
 import { TickRendererProps } from '@vx/axis/lib/types';
-import { Grid } from '@vx/grid';
+import { GridColumns, GridRows } from '@vx/grid';
 import styles from './AxisSystem.module.css';
 import { AxisOffsets, Domain } from './models';
-import { adaptedNumTicks, useAbscissaScale, useOrdinateScale } from './utils';
+import { useAbscissaScale, useOrdinateScale, getTicksProp } from './utils';
 import { AxisSystemContext } from './AxisSystemProvider';
+
+const SHARED_PROPS = {
+  tickStroke: 'grey',
+  hideAxisLine: true,
+  tickFormat: format('0'),
+  tickClassName: styles.tick,
+  tickComponent: ({ formattedValue, ...tickProps }: TickRendererProps) => (
+    <text {...tickProps}>{formattedValue}</text>
+  ),
+};
 
 interface Props {
   axisOffsets: AxisOffsets;
@@ -45,27 +55,17 @@ function AxisSystem(props: Props): JSX.Element {
     ]);
   });
 
-  const sharedAxisProps = {
-    tickStroke: 'grey',
-    hideAxisLine: true,
-    tickFormat: format('0'),
-    tickClassName: styles.tick,
-    tickComponent: ({ formattedValue, ...tickProps }: TickRendererProps) => (
-      <text {...tickProps}>{formattedValue}</text>
-    ),
-  };
+  // Restrain ticks scales to visible domains
+  const xTicksScale = abscissa.scaleFn();
+  xTicksScale.domain(visibleDomains[0]);
+  xTicksScale.range([0, width]);
 
-  // Restrain axis scales to visible domains
-  const xAxisScale = abscissa.scaleFn();
-  xAxisScale.domain(visibleDomains[0]);
-  xAxisScale.range([0, width]);
+  const yTicksScale = ordinate.scaleFn();
+  yTicksScale.domain(visibleDomains[1]);
+  yTicksScale.range([height, 0]);
 
-  const yAxisScale = ordinate.scaleFn();
-  yAxisScale.domain(visibleDomains[1]);
-  yAxisScale.range([height, 0]);
-
-  const abscissaTicksNumber = adaptedNumTicks(width);
-  const ordinateTicksNumber = adaptedNumTicks(height);
+  const xTicksProp = getTicksProp(abscissaConfig, visibleDomains[0], width);
+  const yTicksProp = getTicksProp(ordinateConfig, visibleDomains[1], height);
 
   return (
     <Dom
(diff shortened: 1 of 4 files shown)


src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -45,27 +55,17 @@ function AxisSystem(props: Props): JSX.Element {
...
-      <text {...tickProps}>{formattedValue}</text>
-    ),
-  };
-
-  // Restrain axis scales to visible domains
-  const xAxisScale = abscissa.scaleFn();
-  xAxisScale.domain(visibleDomains[0]);
-  xAxisScale.range([0, width]);
+  // Restrain ticks scales to visible domains
+  const xTicksScale = abscissa.scaleFn();
```
axelboc, line 59: I renamed those to avoid confusing them with the axis scales.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -45,27 +55,17 @@ function AxisSystem(props: Props): JSX.Element {
...
-  yAxisScale.domain(visibleDomains[1]);
-  yAxisScale.range([height, 0]);
+  const yTicksScale = ordinate.scaleFn();
+  yTicksScale.domain(visibleDomains[1]);
+  yTicksScale.range([height, 0]);
 
-  const abscissaTicksNumber = adaptedNumTicks(width);
-  const ordinateTicksNumber = adaptedNumTicks(height);
+  const xTicksProp = getTicksProp(abscissaConfig, visibleDomains[0], width);
+  const yTicksProp = getTicksProp(ordinateConfig, visibleDomains[1], height);
```
axelboc, line 68: `getTicksProp` is just a helper function that returns an object with either `numTicks` (data axis) or `tickValues` (index axis). The object is then destructured on both the axis and grid components.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -65,7 +65,7 @@ export function getDataScaleFn(isLog: boolean): DataScaleFn {
   return isLog ? scaleSymlog : scaleLinear;
 }
 
-function useAxisScale(config: AxisConfig, rangeSize: number): AxisScale {
+function getAxisScale(config: AxisConfig, rangeSize: number): AxisScale {
```
axelboc, line 68: This function doesn't need to be a hook.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -90,13 +90,59 @@ function useAxisScale(config: AxisConfig, rangeSize: number): AxisScale {
...
+  if (optimalCount === 0) {
+    return [];
+  }
+
+  // Make sure we always use a step that is greater than or equal to 1
+  const step = Math.max(tickStep(intMin, intMax, optimalCount), 1);
+
+  const start = Math.ceil(min / step);
+  const stop = Math.floor(max / step);
+  const len = Math.ceil(stop - start + 1);
```
loichuder, line 126: Is `ceil` necessary here as `start` and `stop` are already integer ? Btw, I would rename `len` in `nTicks`

src/h5web/visualizations/shared/utils.ts
```diff
@@ -90,13 +90,59 @@ function useAxisScale(config: AxisConfig, rangeSize: number): AxisScale {
...
+
+  const ticks = new Array(len);
+  for (let i = 0; i < len; i++) {
+    ticks[i] = (start + i) * step;
```
loichuder, line 130: <s>Should not it be `min + i*step` ? As `start = ceil(min / step)` it should be more or less the same but it feels strange to divide by `step` and multiply again at each step of the loop.</s>

EDIT : Nvm `start` is integer. `min` is not.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -90,13 +90,59 @@ function useAxisScale(config: AxisConfig, rangeSize: number): AxisScale {
...
+export function getTicksProp(
+  config: AxisConfig,
+  visibleDomain: Domain,
+  availableSize: number
+): { tickValues: number[] } | { numTicks: number } {
+  const numTicks = adaptedNumTicks(availableSize);
+
+  return isIndexAxisConfig(config)
+    ? { tickValues: getIntegerTicks(visibleDomain, numTicks) }
+    : { numTicks };
```
loichuder, line 145: For the last tick values, you could put a boolean `hideLastTickValue` in `config` and check here if you need the slice the result from `getIntegerTicks(visibleDomain, numTicks)` :shrug:

src/h5web/visualizations/shared/utils.ts
axelboc: Yeah, I wondered that as well. This part comes straight from D3's code, so I assumed there was a reason... but I don't see it either. I'll remove it 👍

src/h5web/visualizations/shared/utils.ts
axelboc: As discussed, see #144

PR #154, opened 2020-05-25
Pass axes info through axis system context instead of raw config

axelboc:
The main goal of this refactoring is to simplify the code that relies on the axis configs passed to `VisCanvas`. From now on, `AxisSystemProvider` converts the two configs into two `AxisInfo` objects. These objects have two jobs:

- to represent the configs in a format common to both index and data axes (e.g. `domain` instead of `indexDomain` and `dataDomain`, `isLog` always present but always `false` for index axes, etc.);
- to store any information that can be deduced from the configs so the child components don't have to deduce them themselves (e.g. `isIndexAxis`, `scaleFn`).

This implementation allows `VisCanvas` to keep a simple, type-safe, public API, while also reducing a lot of the complexity in our internal components.

### DIFF
--- a/src/h5web/visualizations/heatmap/utils.ts
+++ b/src/h5web/visualizations/heatmap/utils.ts
@@ -1,5 +1,11 @@
 import { range } from 'lodash-es';
-import { D3Interpolator, ColorMapOption, ColorMap } from './models';
+import { scaleSymlog, scaleLinear } from 'd3-scale';
+import {
+  D3Interpolator,
+  ColorMapOption,
+  ColorMap,
+  DataScaleFn,
+} from './models';
 
 export function generateCSSLinearGradient(
   interpolator: D3Interpolator,
@@ -10,6 +16,10 @@ export function generateCSSLinearGradient(
     .reduce((acc, val) => `${acc},${val}`);
 
   return `linear-gradient(to ${direction},${gradientColors})`;
+}
+
+export function getDataScaleFn(isLog: boolean): DataScaleFn {
+  return isLog ? scaleSymlog : scaleLinear;
 }
 
 export function convertToOptions(
--- a/src/h5web/visualizations/line/hooks.ts
+++ b/src/h5web/visualizations/line/hooks.ts
@@ -1,12 +1,10 @@
 import { useContext, useMemo } from 'react';
 import { Vector3 } from 'three';
+import { useThree } from 'react-three-fiber';
 import { LineProps, LineContext } from './LineProvider';
 import { Domain } from '../shared/models';
-import {
-  findDomain,
-  useAbscissaScale,
-  useOrdinateScale,
-} from '../shared/utils';
+import { findDomain, getAxisScale } from '../shared/utils';
+import { AxisSystemContext } from '../shared/AxisSystemProvider';
 
 export function useProps(): LineProps {
   const props = useContext(LineContext);
@@ -25,8 +23,13 @@ export function useDataDomain(): Domain | undefined {
 
 export function useDataPoints(): Vector3[] | undefined {
   const { data } = useProps();
-  const { scale: abscissaScale } = useAbscissaScale();
-  const { scale: ordinateScale } = useOrdinateScale();
+
+  const { abscissaInfo, ordinateInfo } = useContext(AxisSystemContext);
+  const { size } = useThree();
+  const { width, height } = size;
+
+  const abscissaScale = getAxisScale(abscissaInfo, width);
+  const ordinateScale = getAxisScale(ordinateInfo, height);
 
   return useMemo(() => {
     return data.map(
--- a/src/h5web/visualizations/shared/AxisSystem.tsx
+++ b/src/h5web/visualizations/shared/AxisSystem.tsx
@@ -6,10 +6,10 @@ import { TickRendererProps } from '@vx/axis/lib/types';
 import { GridColumns, GridRows } from '@vx/grid';
 import styles from './AxisSystem.module.css';
 import { AxisOffsets, Domain } from './models';
-import { useAbscissaScale, useOrdinateScale, getTicksProp } from './utils';
+import { getTicksProp, getAxisScale } from './utils';
 import { AxisSystemContext } from './AxisSystemProvider';
 
-const SHARED_PROPS = {
+const AXIS_PROPS = {
   tickStroke: 'grey',
   hideAxisLine: true,
   tickFormat: format('0'),
@@ -26,17 +26,17 @@ interface Props {
 function AxisSystem(props: Props): JSX.Element {
   const { axisOffsets } = props;
 
+  const { abscissaInfo, ordinateInfo } = useContext(AxisSystemContext);
   const { camera, size } = useThree();
   const { width, height } = size;
 
-  const { abscissaConfig, ordinateConfig } = useContext(AxisSystemContext);
-  const abscissa = useAbscissaScale();
-  const ordinate = useOrdinateScale();
+  const abscissaScale = getAxisScale(abscissaInfo, width);
+  const ordinateScale = getAxisScale(ordinateInfo, height);
 
   // axisDomains are the complete domains. visibleDomains change with the camera
   const [visibleDomains, setVisibleDomains] = useState<[Domain, Domain]>([
-    abscissa.domain,
-    ordinate.domain,
+    abscissaInfo.domain,
+    ordinateInfo.domain,
   ]);
 
   useFrame(() => {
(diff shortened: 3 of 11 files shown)


src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -12,6 +18,10 @@ export function generateCSSLinearGradient(
   return `linear-gradient(to ${direction},${gradientColors})`;
 }
 
+export function getDataScaleFn(isLog: boolean): DataScaleFn {
+  return isLog ? scaleSymlog : scaleLinear;
+}
```
axelboc, line 23: I decided to move `getDataScaleFn` here because I don't think it makes sense to use the same logic for both data axis scales and heatmap data scales. The logic to choose the log scale function for a data axis may become more complex because of log/symlog, while the heatmap data scale function may remain simple.

I removed the original `getDataScaleFn` for now, as you'll see, as it was no longer used in multiple places. It may make sense to re-extract it once we implement proper log scale support, but we'll see about it then.

src/h5web/visualizations/shared/AxisSystemContext.ts
```diff
@@ -0,0 +1,10 @@
+import { createContext } from 'react';
+import { AxisInfo } from './models';
+
+export interface AxisConfigs {
+  abscissa: AxisInfo;
+  ordinate: AxisInfo;
+}
+
+const AxisSystemContext = createContext<AxisConfigs>({} as AxisConfigs);
+export default AxisSystemContext;
```
axelboc, line 10: I had to move the context into a separate file to avoid a cyclic dependency with `utils`... Just realised I'll be able to bring it back by moving the hooks in `utils` into a file called `hooks`. I'll do it once you've reviewed the PR to avoid cluttering the diff.

src/h5web/visualizations/line/hooks.ts
```diff
@@ -25,8 +23,13 @@ export function useDataDomain(): Domain | undefined {
...
   const { data } = useProps();
-  const { scale: abscissaScale } = useAbscissaScale();
-  const { scale: ordinateScale } = useOrdinateScale();
+
+  const { abscissa, ordinate } = useContext(AxisSystemContext);
+  const { size } = useThree();
+  const { width, height } = size;
+
+  const abscissaScale = getAxisScale(abscissa, width);
+  const ordinateScale = getAxisScale(ordinate, height);
```
axelboc, line 32: I decided to remove the `useAbscissaScale` and `useOrdinateScale` now that the interface of `getAxisScale` is a lot simpler. I no longer think that the convenience is worth the added complexity.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -26,12 +26,12 @@ interface Props {
...
 
+  const { abscissa, ordinate } = useContext(AxisSystemContext);
   const { camera, size } = useThree();
   const { width, height } = size;
 
-  const { abscissaConfig, ordinateConfig } = useContext(AxisSystemContext);
-  const abscissa = useAbscissaScale();
-  const ordinate = useOrdinateScale();
+  const abscissaScale = getAxisScale(abscissa, width);
+  const ordinateScale = getAxisScale(ordinate, height);
```
axelboc, line 34: Note that this hook-less implementation is now consistent with the way we call `getTicksProp` below: we pass the axis info object as first parameter.

src/h5web/visualizations/shared/AxisSystemProvider.tsx
```diff
@@ -1,24 +1,45 @@
...
-);
+  const { dataDomain, isLog = false, showGrid = false } = config;
+  return {
+    isIndexAxis: false,
+    scaleFn: isLog ? scaleSymlog : scaleLinear,
+    domain: dataDomain,
+    isLog,
+    showGrid,
+  };
+}
```
axelboc, line 27: Messy diff, but this `getAxisInfo` function is the core of this refactoring: it takes an `AxisConfig` and returns an `AxisInfo`. This is where I moved the `getDataScaleFn` logic for now, by the way.

src/h5web/visualizations/shared/TooltipMesh.tsx
```diff
@@ -1,10 +1,17 @@
...
 import styles from './TooltipMesh.module.css';
-import { useAbscissaScale, useOrdinateScale } from './utils';
+import { getAxisScale } from './utils';
+import AxisSystemContext from './AxisSystemContext';
+
+const GUIDE_PROPS = {
+  stroke: 'gray',
+  strokeWidth: 1.5,
+  strokeOpacity: 0.5,
+};
```
axelboc, line 14: I just moved this static object out of the component and renamed it for consistency.

src/h5web/visualizations/shared/TooltipMesh.tsx
```diff
@@ -18,9 +25,14 @@ interface Props {
 function TooltipMesh(props: Props): ReactElement {
   const { formatIndex, formatValue, guides } = props;
 
+  const { abscissa, ordinate } = useContext(AxisSystemContext);
   const { camera, size } = useThree();
   const { width, height } = size;
 
+  // Scales to compute data coordinates from unprojected mesh coordinates
+  const abscissaScale = getAxisScale(abscissa, width);
+  const ordinateScale = getAxisScale(ordinate, height);
```
axelboc, line 34: The removal of the hooks requires retrieving the axis info objects. I think it actually makes things clearer. Also, we were already retrieving the size of the canvas anyway.

src/h5web/visualizations/shared/TooltipMesh.tsx
```diff
@@ -86,18 +86,18 @@ function TooltipMesh(props: Props): ReactElement {
             </TooltipWithBounds>
             {guides && (
               <svg width={width} height={height}>
-                {verticalGuide && (
+                {guides !== 'horizontal' && (
```
axelboc, line 89: Just a small simplification of the logic.

src/h5web/visualizations/shared/models.ts
```diff
@@ -18,10 +18,18 @@ export interface DataAxisConfig {
...
+export type AxisScale =
+  | ScaleLinear<number, number>
+  | ScaleSymLog<number, number>;
+
+export interface AxisInfo {
+  isIndexAxis: boolean;
+  scaleFn: AxisScaleFn;
   domain: Domain;
+  isLog: boolean;
+  showGrid: boolean;
```
axelboc, line 32: The new `AxisInfo` interface.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -61,42 +59,14 @@ export function isIndexAxisConfig(
   return 'indexDomain' in config;
 }
 
-export function getDataScaleFn(isLog: boolean): DataScaleFn {
-  return isLog ? scaleSymlog : scaleLinear;
-}
-
-function getAxisScale(config: AxisConfig, rangeSize: number): AxisScale {
```
axelboc, line 68: This function drops in complexity because we've already computed the scale function, and the domain can be accessed from the same `domain` key.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -134,13 +103,13 @@ function getIntegerTicks(domain: Domain, count: number): number[] {
...
 export function getTicksProp(
-  config: AxisConfig,
+  info: AxisInfo,
   visibleDomain: Domain,
   availableSize: number
 ): { tickValues: number[] } | { numTicks: number } {
   const numTicks = adaptedNumTicks(availableSize);
 
-  return isIndexAxisConfig(config)
+  return info.isIndexAxis
```
axelboc, line 112: No need to call the guard here anymore.

src/h5web/visualizations/shared/models.ts
loichuder: I wonder if `isLog` is still necessary in `AxisInfo` given that its purpose was to choose which scale function to pick and it is already done as this point.
That being said, I find it clearer to keep it..

src/h5web/visualizations/heatmap/utils.ts
loichuder: >  I don't think it makes sense to use the same logic for both data axis scales and heatmap data scales

I don't fully agree with this but I get your point. As you said, because proper log scale might come, it can stay like this.

src/h5web/visualizations/shared/models.ts
axelboc: Ha true!

>That being said, I find it clearer to keep it..

As you wish 👍

PR #155, opened 2020-05-25
Replaced Line and Heatmap providers by a common VisProvider

loichuder:


src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -5,13 +5,15 @@ import VisSelector from './VisSelector';
...
 import VisBar from './VisBar';
+import VisProvider from './VisProvider';
 
 interface Props {
   dataset?: HDF5Dataset;
 }
 
 function DatasetVisualizer(props: Props): JSX.Element {
   const { dataset } = props;
+  const dimensionMapping = { x: 0, y: 1 };
```
loichuder, line 16: This is only dummy for now.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -31,7 +33,13 @@ function DatasetVisualizer(props: Props): JSX.Element {
       <div className={styles.displayArea}>
         {dataset ? (
           activeVis && (
-            <VisDisplay key={dataset.id} vis={activeVis} dataset={dataset} />
+            <VisProvider
+              key={dataset.id}
```
loichuder, line 37: I wonder if the `key` is needed for the provider... I am not versed in these things so I would be happy to have your input.

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -0,0 +1,51 @@
...
+  props: VisProps & { children: ReactNode }
+): ReactElement {
+  const { mapping, dims, data, children } = props;
+
+  return (
+    <VisContext.Provider value={{ data, dims, mapping }}>
+      {children}
+    </VisContext.Provider>
+  );
+}
```
loichuder, line 33: R3F's `Canvas` require that we provide the context once again. To avoid triggering `useValue` once more and problems due to the mismatch of input/output props, we need a special version of the provider that only forwards the props.

src/h5web/dataset-visualizer/hooks.ts
```diff
@@ -0,0 +1,35 @@
...
+  dims: [number, number];
+} {
+  const { data, dims } = useProps();
+
+  if (dims.length === 2) {
+    return { data, dims: dims as [number, number] };
+  }
+
+  throw new Error('Data not supported by HeatmapVis');
+}
```
loichuder, line 35: This is where the mapping will come into play. For now, it is only providing the needed props to the corresponding vis.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -40,7 +41,7 @@ function HeatmapVis(): JSX.Element {
         aspectRatio={aspectRatio}
       >
         {/* Provide context again - https://github.com/react-spring/react-three-fiber/issues/262 */}
-        <HeatmapProvider {...props}>
+        <VisProviderForCanvas {...visProps}>
```
loichuder, line 44: As the provider does no longer depend on the vis, we could put `VisProviderForCanvas` inside the `VisCanvas` which would be quite convenient but could also render `VisCanvas` less independent.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
axelboc: Remove it, open a 2D dataset, select another 2D dataset, observe if something behaves inappropriately while things update. You're looking for a component that still uses the data, dimension, domain, whatever... of the first dataset you selected while the second dataset is loading or its texture is being computed. For instance, make sure that: the domain in the toolbar changes instantly; the texture is not recomputed twice; etc. Try also with 1D datasets.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -31,7 +33,13 @@ function DatasetVisualizer(props: Props): JSX.Element {
...
+              mapping={dimensionMapping}
+              dataset={dataset}
+            >
+              <VisDisplay key={dataset.id} vis={activeVis} dataset={dataset} />
```
axelboc, line 41: No need for a `key` here either way.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
axelboc: I think it's still needed because there's still a render when the dimensions passed to the visualization (known right away from `dataset`) doesn't match the value retrieved with `useValue` (which is asynchronous). A fix might be to return a distinct `dataset` object from `useValue` asynchronously (i.e. by storing it with the value in the hook's local state). We can do this later.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
axelboc: I don't think you need to pass `dataset` here either any more.

src/h5web/dataset-visualizer/VisProvider.tsx
axelboc: Looking at the code, I don't think you need a component for this. Just use `<VisContext.Provider {...providerProps}>` in the visualizations directly.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -3,17 +3,18 @@ import shallow from 'zustand/shallow';
...
 import { useHeatmapConfig } from './config';
 import PanZoomMesh from '../shared/PanZoomMesh';
 import VisCanvas from '../shared/VisCanvas';
+import { useProps, useHeatmapValue } from '../../dataset-visualizer/hooks';
+import { VisProviderForCanvas } from '../../dataset-visualizer/VisProvider';
 
 function HeatmapVis(): JSX.Element {
-  const props = useProps();
-  const { dims, data } = props;
+  const visProps = useProps();
```
axelboc, line 16: Might be worth naming the hook `useVisProps` as well.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
loichuder: Agreed. I kept the key on `VisProvider` only.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
loichuder: Done !

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
loichuder: In the end, I took care of the remaining visualizations and removed `VisDisplay` altogether.

loichuder:
For now, each vis has dedicated hooks to get the sliced data/dims (`useData` and `useDims`) while the provider holds the rawData/rawDims.
We can change this in the future if we feel it is needed.

PR #156, opened 2020-05-26
Added DimensionMapper for 2D LineVis

loichuder:
A first try to implement dimension mapping in the LineVis for 2D datasets.

This is rough around the edges but I think it was worth discussing the path it takes before going too far in the refinement.

image

### DIFF
--- a/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
+++ b/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
@@ -4,23 +4,7 @@ import styles from './DatasetVisualizer.module.css';
 import VisSelector from './VisSelector';
 import { getSupportedVis, useActiveVis } from './utils';
 import VisBar from './VisBar';
-import VisProvider from './VisProvider';
-import { Vis } from './models';
-import {
-  RawVis,
-  ScalarVis,
-  MatrixVis,
-  LineVis,
-  HeatmapVis,
-} from '../visualizations';
-
-const VIS_COMPONENTS = {
-  [Vis.Raw]: RawVis,
-  [Vis.Scalar]: ScalarVis,
-  [Vis.Matrix]: MatrixVis,
-  [Vis.Line]: LineVis,
-  [Vis.Heatmap]: HeatmapVis,
-};
+import VisDisplay from './VisDisplay';
 
 interface Props {
   dataset?: HDF5Dataset;
@@ -28,11 +12,9 @@ interface Props {
 
 function DatasetVisualizer(props: Props): JSX.Element {
   const { dataset } = props;
-  const dimensionMapping = { x: 0, y: 1 };
 
   const supportedVis = useMemo(() => getSupportedVis(dataset), [dataset]);
   const [activeVis, setActiveVis] = useActiveVis(supportedVis);
-  const ActiveVis = activeVis && VIS_COMPONENTS[activeVis];
 
   return (
     <div className={styles.visualizer}>
@@ -48,14 +30,12 @@ function DatasetVisualizer(props: Props): JSX.Element {
       </div>
       <div className={styles.displayArea}>
         {dataset ? (
-          ActiveVis && (
-            <VisProvider
+          activeVis && (
+            <VisDisplay
               key={dataset.id}
-              mapping={dimensionMapping}
+              activeVis={activeVis}
               dataset={dataset}
-            >
-              <ActiveVis />
-            </VisProvider>
+            />
           )
         ) : (
           <p className={styles.noVis}>Nothing to visualize</p>
--- a/src/h5web/dataset-visualizer/DimensionMapper.module.css
+++ b/src/h5web/dataset-visualizer/DimensionMapper.module.css
@@ -1,0 +1,87 @@
+.mapper {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  background-color: var(--primary-light-bg);
+}
+
+.buttonGroupWrapper {
+  display: flex;
+  align-items: center;
+  justify-content: space-evenly;
+  margin: 0.25rem;
+}
+
+.buttonGroup {
+  display: flex;
+}
+
+.buttonGroupLabel {
+  margin-right: 0.5rem;
+}
+
+.btn {
+  composes: btn-clean from global;
+  display: flex;
+  align-items: center;
+  padding: 0.25rem 0.75rem 0.375rem;
+  background-color: var(--secondary);
+  transition: background-color 0s ease-out;
+  color: white;
+  border-radius: 10%;
+}
+
+.btn:hover {
+  background-color: var(--secondary-dark);
+}
+
+.btn[aria-checked='true'],
+.btn[aria-pressed='true'] {
+  background-color: var(--secondary-dark);
+  font-weight: bold;
+}
+
+.sliderWrapper {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  margin: 1rem 0;
+}
+
+.slider {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  align-self: center;
+  margin: 0.5rem 0.125rem;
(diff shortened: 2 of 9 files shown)


src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -28,7 +29,10 @@ interface Props {
 
 function DatasetVisualizer(props: Props): JSX.Element {
   const { dataset } = props;
-  const dimensionMapping = { x: 0, y: 1 };
+  const [dimensionMapping, setMapping] = useState<DimensionMapping>({
+    x: 0,
+    slicingIndices: { 1: 0 },
+  });
```
loichuder, line 35: This should be replaced by something more clever that depends on the dataset and the activeVis.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -0,0 +1,92 @@
...
+interface Props {
+  activeVis: Vis;
+  setMapping: (d: DimensionMapping) => void;
+}
+
+function DimensionMapper(props: Props): JSX.Element {
+  const { activeVis, setMapping } = props;
+  const { rawDims, mapping } = useVisProps();
+
+  if (activeVis === Vis.Line && rawDims.length === 2) {
```
loichuder, line 17: For now, the apparition of the DimensionMapper is only triggered for LineVis of 2D datasets.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -0,0 +1,92 @@
...
+    return (
+      <div className={styles.mapper}>
+        <div className={styles.buttonGroupWrapper}>
+          <span className={styles.buttonGroupLabel}>X</span>
+          <ButtonGroup
+            className={styles.buttonGroup}
+            buttonClassName={styles.btn}
+            buttons={Object.keys(rawDims).map(dimKey => {
```
loichuder, line 25: Even though there are only 2 dims, I tried to keep a syntax that can be extensible to higher dimensions.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -0,0 +1,92 @@
...
+              return {
+                label: dimKey,
+                isSelected: mapping.x === dimIndex,
+                onClick: () => {
+                  if (mapping.x !== dimIndex) {
+                    const newMapping: DimensionMapping = {
+                      x: dimIndex,
+                      slicingIndices: { [mapping.x]: 0 },
+                    };
+                    setMapping(newMapping);
```
loichuder, line 36: Well except here... the `newMapping` set the new `dimIndex` as `x` and the slicingIndex of the old `x` (in `mapping.x`) back to 0.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -0,0 +1,92 @@
...
+                      return;
+                    }
+                    const newMapping: DimensionMapping = {
+                      x: mapping.x,
+                      slicingIndices: {
+                        ...mapping.slicingIndices,
+                        [Number(dimNumber)]: values as number,
+                      },
+                    };
+                    setMapping(newMapping);
```
loichuder, line 74: The slider only needs to update the appropirate entry of `slicingIndices`. We can then imagine replacing this part by some kind of merge state.

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -50,4 +51,10 @@ export function useVisProps(): VisProps {
   return props;
 }
 
+export function useFlatValues(): number[] {
+  const { rawValues } = useVisProps();
+
+  return useMemo(() => rawValues.flat(Infinity), [rawValues]);
+}
+
```
loichuder, line 59: A new hook to get the values *completely* flattened to slice deep dimensions. Otherwise, `rawValues.flat()` only flattens the first level.

WE might want to replace the `useValues` in `HeatmapVis` by this.

src/h5web/dataset-visualizer/models.ts
```diff
@@ -8,5 +8,6 @@ export enum Vis {
 
 export interface DimensionMapping {
   x: number;
-  y: number;
+  y?: number;
+  slicingIndices: Record<number, number>;
 }
```
loichuder, line 13: Ok, this is where it is a bit tricky and where we may want to think really hard: the structure of the mapping object really dictates how components are created, how it should be updated and how to slice the data afterwards.

I have chosen the above structure (**which is greatly open for discussion**) where the index of the dimension that will be used for the visualisation (let's call it the data dimension) is stored as the field `x`. The other dimensions need then to be sliced and it is the job of `slicingIndices` to keep the indices used for the slicing. It is a record where the keys are all dimensions indices **except the data dimension** and the values are the indices along which the slicing should be made.

To give an example, for a 2D dataset in a `LineVis`: we can have
```
{ 
  x: 0
  slicingIndices: {1: slicingIndexForDim1}
}
```
or
```
{ 
  x: 1
  slicingIndices: {0: slicingIndexForDim0}
}
```

If we go to 3D datasets, we will have the following examples:
```
{ 
  x: 1
  slicingIndices: {0: slicingIndexForDim0, 2: slicingIndexForDim2}
}
```
or
```
{ 
  x: 2
  slicingIndices: {0: slicingIndexForDim0, 1: slicingIndexForDim1}
}
```
etc.

src/h5web/visualizations/line/hooks.ts
```diff
@@ -4,15 +4,29 @@ import { useThree } from 'react-three-fiber';
...
+  const values = useFlatValues();
 
   if (rawDims.length === 1) {
     return rawValues;
   }
 
+  if (rawDims.length === 2) {
+    if (mapping.x === 1) {
+      return rawValues[mapping.slicingIndices[0]];
+    }
```
loichuder, line 23: If the dataDimension is the deepest one, we don't have to think really hard on how to access the data: it is only `data[slicingIndexForDim0]`

src/h5web/visualizations/line/hooks.ts
```diff
@@ -4,15 +4,29 @@ import { useThree } from 'react-three-fiber';
...
+
+    return values.filter(
+      (v, i) => i % rawDims[1] === mapping.slicingIndices[1]
+    );
```
loichuder, line 27: If we want to slice the deepest dimension, things are trickier: we have to filter the flattened array from the indices to select the appropriate values.

Unfortunately, this is not very efficient and the manipulation of the slider feels sluggish when slicing along dimension 1.
A way out would be to reorder the data at each change of the data dimension so that the slicing stays efficient.

src/h5web/dataset-visualizer/DimensionMapper.module.css
```diff
@@ -0,0 +1,87 @@
...
+  font-size: 0.625rem;
+  height: 1.5rem;
+  min-width: 1.5rem;
+  cursor: grab;
+}
+
+.sliderThumb:global(.active) {
+  cursor: none;
+  border-color: var(--secondary-dark);
+}
```
loichuder, line 87: The CSS was fetched from other components modules and adapted. I hope to do something cleaner once the toolbar refactor is finished.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -54,6 +58,10 @@ function DatasetVisualizer(props: Props): JSX.Element {
               mapping={dimensionMapping}
               dataset={dataset}
             >
+              <DimensionMapper
+                activeVis={activeVis as Vis}
+                setMapping={setMapping}
```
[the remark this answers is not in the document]
loichuder: This was fixed by introducting a new (old?) component `VisDisplay` that takes a `Vis` `activeVis` as props

PR #157, opened 2020-05-27
Improve design of toggler buttons + misc

axelboc:
I squeezed a couple extra things in there (cf. commits).


src/h5web/visualizations/heatmap/HeatmapToolbar.tsx
```diff
@@ -1,46 +1,50 @@
...
-  ] = useHeatmapConfig(state => [
-    state.hasLogScale,
-    state.toggleLogScale,
-    state.keepAspectRatio,
-    state.toggleAspectRatio,
-    state.showGrid,
-    state.toggleGrid,
-    shallow,
-  ]);
+  } = useHeatmapConfig();
```
axelboc, line 19: Technically, this simplification means that the toolbar is re-rendered when the domain slider and color map selector values change, but I'll make those components controlled as well. It feels wrong to retrieve half the config in the toolbar and the rest in a few child components. Ideally it should be retrieved only in child components, but that makes it impossible to create a re-usable toggler component. So I think it makes more sense to retrieve the whole config in the toolbar and re-render everything on change, and to optimise the heavy components as needed with `React.memo`.

src/h5web/visualizations/shared/Separator.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { ReactElement } from 'react';
+
+import styles from './Separator.module.css';
+
+function Separator(): ReactElement {
+  return <span className={styles.sep} />;
```
axelboc, line 6: Just a dumb component to make it easy to group buttons visually on the toolbar. I think it's important that toggler buttons and button groups are separated from non-toggleable buttons like the screenshot button, to make it clearer that they don't behave the same way. I did make sure they looked slightly different (e.g. no inner shadow on "normal" buttons), but I don't think it's quite enough.

src/h5web/visualizations/shared/Toggler.tsx
```diff
@@ -18,8 +20,10 @@ function Toggler(props: Props): JSX.Element {
       aria-checked={value}
       onClick={onChange}
     >
-      {value ? <FiToggleRight /> : <FiToggleLeft />}
-      <span className={styles.togglerLabel}>{label}</span>
+      <span className={styles.box}>
```
axelboc, line 23: The `box` is the element that actually looks like a button.

axelboc: I plan on sharing the `box` code between normal buttons and togglers, eventually.

src/h5web/visualizations/shared/Toolbar.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import styles from './Toolbar.module.css';
+
+interface Props {
+  children: ReactNode;
+}
+
+function Toolbar(props: Props): ReactElement {
+  const { children } = props;
+  return <div className={styles.toolbar}>{children}</div>;
```
axelboc, line 11: Dumb for now, but that's where the responsive logic will take place.

src/h5web/visualizations/shared/TooltipMesh.tsx
```diff
@@ -65,11 +65,22 @@ function TooltipMesh(props: Props): ReactElement {
...
+  const onPointerUp = useCallback(
+    (evt: PointerEvent) => {
+      const { offsetX: x, offsetY: y } = evt.nativeEvent;
+      if (x >= 0 && x <= width && y >= 0 && y <= height) {
+        onPointerMove(evt);
+      }
+    },
+    [height, onPointerMove, width]
+  );
+
```
axelboc, line 78: Just a quick side fix as explained in the comment. At the moment, you have to move the mouse again after dragging to see the tooltip.

src/h5web/visualizations/shared/Toggler.tsx
```diff
@@ -18,8 +20,10 @@ function Toggler(props: Props): JSX.Element {
       aria-checked={value}
```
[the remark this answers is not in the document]
axelboc: Hehe but it's not a button, it has role `switch`: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role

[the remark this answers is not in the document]
axelboc: From what I read, there's very little difference... Right down the button of this page: https://scottaohara.github.io/a11y_styled_form_controls/src/toggle-button-switch/ they do recommend only using `role="switch"` when the button looks like a switch. In our case, it's really more of a pressed button, so I'll make the change.

PR #158, opened 2020-05-27
Refactor update of active visualization

axelboc:
Previously, we had:

1. Dataset changes
1. `DatasetVisualizer` renders
    - `activeVis` is identified as invalid, so `undefined` is used instead;
    - `useEffect` schedules reset of `activeVis` state on next render.
1. Children of `DatasetVisualizer` render, but `VisDisplay` is removed from DOM because `activeVis` is `undefined`.
1. `DatasetVisualizer` renders again on next tick (because of change of `activeVis` state in `useEffect`) 
1. Children of `DatasetVisualizer` render, and `VisDisplay` is re-added to the DOM now that `activeVis` is valid again.

Now, we have:

1. Dataset changes
1. `DatasetVisualizer` renders
    - Change of dataset is detected, so `setActiveVis` is called right away (instead of being scheduled in `useEffect`) -- this cancels rendering of `DatasetVisualizer`'s children.
    - Value of `activeVis` still corresponds to previous dataset, but it doesn't matter since rendering of children is cancelled.
1. `DatasetVisualizer` renders again because of the call to `setActiveVis`.
1. Children of `DatasetVisualizer` render with `activeVis` which now corresponds to the current dataset.

Overall, it's a very minor optimisation: `VisDisplay` was previously removed from the DOM then re-added; now it's simply replaced. The code, however, is less hacky and more in line with React's recommendations on how to deal with derived state: https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

### DIFF
--- a/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
+++ b/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
@@ -1,17 +1,11 @@
-import React, { ElementType, useState, ReactElement, useMemo } from 'react';
+import React, { useState, ReactElement, useMemo } from 'react';
 import { HDF5Dataset } from '../providers/models';
 import styles from './DatasetVisualizer.module.css';
 import VisSelector from './VisSelector';
 import { getSupportedVis } from './utils';
 import VisDisplay from './VisDisplay';
 import { Vis } from './models';
-import LineToolbar from '../visualizations/line/LineToolbar';
-import HeatmapToolbar from '../visualizations/heatmap/HeatmapToolbar';
-
-const VIS_TOOLBARS: Record<string, ElementType> = {
-  [Vis.Line]: LineToolbar,
-  [Vis.Heatmap]: HeatmapToolbar,
-};
+import { VIS_DEFS } from '../visualizations';
 
 interface Props {
   dataset?: HDF5Dataset;
@@ -35,7 +29,7 @@ function DatasetVisualizer(props: Props): ReactElement {
     );
   }
 
-  const VisToolbar = activeVis && VIS_TOOLBARS[activeVis];
+  const VisToolbar = activeVis && VIS_DEFS[activeVis].Toolbar;
 
   return (
     <div className={styles.visualizer}>
--- a/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
+++ b/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
@@ -1,8 +1,8 @@
-import React, { ElementType } from 'react';
+import React, { ElementType, useState, ReactElement, useMemo } from 'react';
 import { HDF5Dataset } from '../providers/models';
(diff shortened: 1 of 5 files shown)


src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -17,11 +17,23 @@ interface Props {
   dataset?: HDF5Dataset;
 }
 
-function DatasetVisualizer(props: Props): JSX.Element {
+function DatasetVisualizer(props: Props): ReactElement {
   const { dataset } = props;
 
-  const supportedVis = getSupportedVis(dataset);
-  const [activeVis, setActiveVis] = useActiveVis(supportedVis);
+  const supportedVis = useMemo(() => getSupportedVis(dataset), [dataset]);
```
axelboc, line 23: I re-added `useMemo` here after all so the method doesn't get called twice on every change of dataset (because of `setActiveVis`).

axelboc:
I pushed an extra commit in there. I combined all `Record<Vis, ...>` into one, as discussed. I ended up calling it `VIS_DEF`, though. It felt better than just `VIS`, especially since I needed a TS interface to go with it (`VisDef`).

axelboc:
Also, I put the code in `visualization/index.ts` instead of creating a new file called `visualization.ts` in `dataset-visualizer`. I think it's a good place for it: it declares all the visualizations available and exposes their components. We no longer need to export the actual vis components from it any more (for now).

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -36,14 +42,12 @@ function DatasetVisualizer(props: Props): JSX.Element {
...
       <div className={styles.displayArea}>
-        {dataset ? (
-          activeVis && (
-            <VisDisplay
-              key={dataset.id}
-              activeVis={activeVis}
-              dataset={dataset}
-            />
-          )
+        {dataset && activeVis ? (
```
loichuder, line 45: Much easier to read :+1:

src/h5web/dataset-visualizer/VisDisplay.tsx
```diff
@@ -42,7 +28,7 @@ function VisDisplay(props: Props): JSX.Element {
   return (
     <VisProvider mapping={dimensionMapping} dataset={dataset}>
       <DimensionMapper activeVis={activeVis} onChange={setMapping} />
-      <ActiveVis />
+      <Component />
```
loichuder, line 31: Minor complaint here: I would rename it `VisComponent` but it can be kept as `Component` in the `VIS_DEFS`. This way, we can understand what the component `VisDisplay` returns without searching what `Component` refers to in the code.

In other words, the destructuring above would be `const { Component:  VisComponent } = VIS_DEFS[activeVis];`

PR #159, opened 2020-05-28
Improvements to mapping of LineVis for 2d

loichuder:
- Mapping is now an array of indices or dimensions.
- Values are transposed so that the slicing always operate on shallow dimensions for performance
- Some CSS styling

All in all, I hope this leads to clearer and better typed code.

### DIFF
--- a/src/h5web/dataset-visualizer/DimensionMapper.module.css
+++ b/src/h5web/dataset-visualizer/DimensionMapper.module.css
@@ -16,12 +16,12 @@
   display: flex;
 }
 
-.buttonGroupLabel {
+.dimensionName {
+  font-family: serif;
   margin-right: 0.5rem;
 }
 
-.btn {
-  composes: btn-clean from global;
+.btnAspect {
   display: flex;
   align-items: center;
   padding: 0.25rem 0.75rem 0.375rem;
@@ -29,6 +29,11 @@
   transition: background-color 0s ease-out;
   color: white;
   border-radius: 10%;
+}
+
+.btn {
+  composes: btn-clean from global;
+  composes: btnAspect;
 }
 
 .btn:hover {
@@ -46,6 +51,12 @@
   flex-direction: column;
   align-items: center;
   margin: 1rem 0;
+}
+
+.sliderLabel {
+  composes: btnAspect;
+  background-color: var(--secondary-dark);
+  font-weight: bold;
 }
 
 .slider {
--- a/src/h5web/dataset-visualizer/DimensionMapper.tsx
+++ b/src/h5web/dataset-visualizer/DimensionMapper.tsx
@@ -1,24 +1,64 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import ReactSlider from 'react-slider';
-import { Vis, DimensionMapping } from './models';
-import { useVisProps } from './VisProvider';
+import { isNumber } from 'lodash-es';
+import { Vis, DimensionMapping, MappingType } from './models';
 import ButtonGroup from '../visualizations/shared/ButtonGroup';
 import styles from './DimensionMapper.module.css';
 
 interface Props {
   activeVis: Vis;
+  rawDims: number[];
+  mapperState: DimensionMapping;
   onChange: (d: DimensionMapping) => void;
 }
 
 function DimensionMapper(props: Props): JSX.Element {
-  const { activeVis, onChange } = props;
-  const { rawDims, mapping } = useVisProps();
+  const { activeVis, rawDims, mapperState, onChange } = props;
+
+  function renderSlicingSlider(
+    slicingIndex: MappingType,
+    dimension: number
+  ): JSX.Element | undefined {
+    // Do not show a slider for 'x' or 'y'
+    if (!isNumber(slicingIndex)) {
+      return undefined;
+    }
+
+    return (
+      // eslint-disable-next-line react/no-array-index-key
+      <Fragment key={dimension}>
+        <span className={styles.sliderLabel}>D{dimension}</span>
+        <ReactSlider
+          className={styles.slider}
+          trackClassName={styles.sliderTrack}
+          thumbClassName={styles.sliderThumb}
+          renderThumb={(thumbProps, state) => (
+            <div {...thumbProps}>{state.valueNow}</div>
+          )}
+          defaultValue={slicingIndex}
+          onChange={value => {
+            if (!isNumber(value)) {
+              return;
+            }
+            const newMapperState = mapperState.slice();
+            newMapperState[dimension] = value;
+            onChange(newMapperState);
+          }}
+          min={0}
+          max={rawDims[dimension] - 1}
+          step={1}
+          orientation="vertical"
+          invert
+        />
+      </Fragment>
+    );
+  }
 
   if (activeVis === Vis.Line && rawDims.length === 2) {
     return (
       <div className={styles.mapper}>
         <div className={styles.buttonGroupWrapper}>
-          <span className={styles.buttonGroupLabel}>X</span>
+          <span className={styles.dimensionName}>X</span>
           <ButtonGroup
             ariaLabel="Dimension as X"
             className={styles.buttonGroup}
@@ -26,15 +66,15 @@ function DimensionMapper(props: Props): JSX.Element {
             buttons={Object.keys(rawDims).map(dimKey => {
               const dimIndex = Number(dimKey);
               return {
-                label: dimKey,
-                isSelected: mapping.x === dimIndex,
+                label: `D${dimIndex}`,
+                isSelected: mapperState[dimIndex] === 'x',
                 onClick: () => {
-                  if (mapping.x !== dimIndex) {
-                    const newMapping: DimensionMapping = {
-                      x: dimIndex,
-                      slicingIndices: { [mapping.x]: 0 },
-                    };
-                    onChange(newMapping);
+                  const prevX = mapperState.indexOf('x');
+                  if (prevX !== dimIndex) {
+                    const newMapperState = mapperState.slice();
+                    newMapperState[prevX] = 0;
+                    newMapperState[dimIndex] = 'x';
+                    onChange(newMapperState);
                   }
                 },
               };
@@ -42,47 +82,7 @@ function DimensionMapper(props: Props): JSX.Element {
           />
         </div>
         <div className={styles.sliderWrapper}>
-          {Object.entries(mapping.slicingIndices).map(
-            ([dimNumber, slicingIndex]) => (
-              <>
-                <button
-                  type="button"
-                  className={styles.btn}
-                  aria-pressed="true"
-                >
-                  {dimNumber}
-                </button>
-                <ReactSlider
-                  key={dimNumber}
-                  className={styles.slider}
-                  trackClassName={styles.sliderTrack}
-                  thumbClassName={styles.sliderThumb}
-                  renderThumb={(thumbProps, state) => (
-                    <div {...thumbProps}>{state.valueNow}</div>
-                  )}
-                  defaultValue={slicingIndex}
-                  onChange={values => {
-                    if (values === undefined) {
-                      return;
-                    }
-                    const newMapping: DimensionMapping = {
-                      x: mapping.x,
-                      slicingIndices: {
-                        ...mapping.slicingIndices,
-                        [Number(dimNumber)]: values as number,
-                      },
-                    };
-                    onChange(newMapping);
-                  }}
-                  min={0}
-                  max={rawDims[Number(dimNumber)] - 1}
-                  step={1}
-                  orientation="vertical"
-                  invert
-                />
-              </>
-            )
-          )}
+          {mapperState.map(renderSlicingSlider)}
         </div>
       </div>
     );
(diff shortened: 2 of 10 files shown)


src/h5web/dataset-visualizer/DimensionMapper.module.css
```diff
@@ -16,12 +16,12 @@
...
 
-.buttonGroupLabel {
+.dimensionName {
+  font-family: serif;
   margin-right: 0.5rem;
 }
 
-.btn {
-  composes: btn-clean from global;
+.btnAspect {
```
loichuder, line 24: A CSS class to give the button aspect without being a button !

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -1,86 +1,78 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import ReactSlider from 'react-slider';
+import { isNumber } from 'lodash-es';
```
loichuder, line 3: Type-guard to check if `value is number`. I could implement my own with `typeof value === 'number'` but it is already there in `lodash` :shrug:

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -1,86 +1,78 @@
...
   activeVis: Vis;
+  rawDims: number[];
+  mapperState: DimensionMapping;
   onChange: (d: DimensionMapping) => void;
 }
 
 function DimensionMapper(props: Props): JSX.Element {
-  const { activeVis, onChange } = props;
-  const { rawDims, mapping } = useVisProps();
+  const { activeVis, rawDims, mapperState, onChange } = props;
```
loichuder, line 16: As discussed, everything is fetched by props rather than with the hook. This is enforced as `DimensionMapper` has no longer access to the `VisContext`.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -1,86 +1,78 @@
...
           <ButtonGroup
             ariaLabel="Dimension as X"
             className={styles.buttonGroup}
             buttonClassName={styles.btn}
             buttons={Object.keys(rawDims).map(dimKey => {
               const dimIndex = Number(dimKey);
               return {
-                label: dimKey,
-                isSelected: mapping.x === dimIndex,
+                label: `D${dimIndex}`,
```
loichuder, line 30: Added a capital `D` for dimension to labels.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -1,86 +1,78 @@
...
-                      x: dimIndex,
-                      slicingIndices: { [mapping.x]: 0 },
-                    };
-                    onChange(newMapping);
+                  const prevX = mapperState.indexOf('x');
+                  if (prevX !== dimIndex) {
+                    const newMapperState = mapperState.slice();
+                    newMapperState[prevX] = 0;
+                    newMapperState[dimIndex] = 'x';
+                    onChange(newMapperState);
```
loichuder, line 38: With the new mapping `mapperState`, the change in the `x` dimension resulting in switching the position of `x` with the previous one and setting the slicing index of the previous dimension to 0.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -1,86 +1,78 @@
...
                   trackClassName={styles.sliderTrack}
                   thumbClassName={styles.sliderThumb}
                   renderThumb={(thumbProps, state) => (
                     <div {...thumbProps}>{state.valueNow}</div>
                   )}
                   defaultValue={slicingIndex}
-                  onChange={values => {
-                    if (values === undefined) {
+                  onChange={value => {
+                    if (!isNumber(value)) {
```
loichuder, line 60: The stronger guard clause allows to enforce the fact that `value` is defined AND is a number.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -1,86 +1,78 @@
...
-                  type="button"
-                  className={styles.btn}
-                  aria-pressed="true"
-                >
-                  {dimNumber}
-                </button>
+          {mapperState.map((slicingIndex, dim) =>
+            typeof slicingIndex === 'number' ? (
+              // eslint-disable-next-line react/no-array-index-key
+              <Fragment key={dim}>
```
loichuder, line 49: This might bite us back once we start implementing multiple sliders so I keep the explicit disable.

src/h5web/dataset-visualizer/VisDisplay.tsx
```diff
@@ -18,18 +18,40 @@ function VisDisplay(props: Props): JSX.Element {
...
+  // Update mapping when vis changes
+  // https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops
+  if (activeVis !== prevVis) {
+    setPrevVis(activeVis);
+    setMapperState(
+      activeVis === Vis.Line
+        ? [...range(0, datasetDims.length - 1).fill(0), 'x']
+        : undefined
+    );
+  }
```
loichuder, line 35: This might look familiar to you :wink: !

I think your approach for the `activeVis` in `DataVisualizer` was well suited to update the `mapperState` here.

src/h5web/dataset-visualizer/VisDisplay.tsx
```diff
@@ -18,18 +18,40 @@ function VisDisplay(props: Props): JSX.Element {
...
-      <VisComponent />
-    </VisProvider>
+    <>
+      {mapperState && (
+        <DimensionMapper
+          activeVis={activeVis}
+          rawDims={datasetDims}
+          mapperState={mapperState}
+          onChange={setMapperState}
+        />
```
loichuder, line 45: The apparition of the `DimensionMapper` depends now on whether there is a mapping defined.

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -5,36 +5,59 @@ import React, {
...
 import { HDF5Dataset, HDF5Value } from '../providers/models';
 import { useValue } from '../providers/hooks';
 import { isSimpleShape } from '../providers/utils';
 
 export interface VisProps {
-  mapping: DimensionMapping;
-  rawValues: HDF5Value;
+  visValues: HDF5Value;
   rawDims: number[];
+  slicingIndices?: number[];
```
loichuder, line 18: `slicingIndices` is now optional as we could imagine having only one dimension on `x`.

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -5,36 +5,59 @@ import React, {
...
 };
 
 function VisProvider(props: Props): ReactElement {
-  const { dataset, mapping, children } = props;
+  const { dataset, mapperState, children } = props;
   const rawValues = useValue(dataset.id);
   const rawDims = isSimpleShape(dataset.shape) ? dataset.shape.dims : [];
 
-  if (rawValues === undefined) {
+  const xDim = mapperState ? mapperState.indexOf('x') : -1;
```
loichuder, line 35: If there is no mapperState, setting `xDim` to `-1` (or `null`) sets the `mappedValues` to the `rawValues`. It is a bit awkward right now but it will be improved when more Vis/dims are supported.

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -5,36 +5,59 @@ import React, {
...
+
+  const mappedValues = useMemo(() => {
+    if (!rawValues) {
+      return undefined;
+    }
+
+    if (rawDims.length === 2 && xDim === 0) {
+      return transpose(rawValues);
+    }
```
loichuder, line 44: If the sliding dimension is the deepest one, the matrix is transposed so that we slice always in the shallowest dimension.

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -5,36 +5,59 @@ import React, {
...
     return <></>;
   }
 
   return (
-    <VisContext.Provider value={{ rawValues, rawDims, mapping }}>
+    <VisContext.Provider
+      value={{
+        visValues: mappedValues,
+        rawDims,
+        slicingIndices: mapperState && mapperState.filter(isNumber),
```
loichuder, line 58: `slicingIndices` is basically the `mapperState` without the `x` field

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -5,36 +5,59 @@ import React, {
...
+
+  if (mappedValues === undefined) {
     return <></>;
   }
 
   return (
-    <VisContext.Provider value={{ rawValues, rawDims, mapping }}>
+    <VisContext.Provider
+      value={{
+        visValues: mappedValues,
```
loichuder, line 56: I prefered `visValues` as these are the values dedicated to the vis components and are not always "mapped".

src/h5web/visualizations/line/hooks.ts
```diff
@@ -4,30 +4,16 @@ import { useThree } from 'react-three-fiber';
...
 export function useData(): number[] {
-  const { rawValues, rawDims, mapping } = useVisProps();
-  const values = useFlatValues();
+  const { visValues, slicingIndices } = useVisProps();
 
-  if (rawDims.length === 1) {
-    return rawValues;
+  if (slicingIndices) {
+    return (visValues as number[][])[slicingIndices[0]];
   }
```
loichuder, line 14: If we need to slice, it is always in the shallowest dimension as we arrange `visValues` as such. If not, `slicingIndices` are undefined and we return the raw values.

src/h5web/dataset-visualizer/DimensionMapper.module.css
axelboc: Ooh nice 👍

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -1,86 +1,78 @@
...
-                  max={rawDims[Number(dimNumber)] - 1}
+                  max={rawDims[dim] - 1}
                   step={1}
                   orientation="vertical"
                   invert
                 />
-              </>
+              </Fragment>
+            ) : (
+              undefined
```
axelboc, line 75: Just for lisibility, it might be nicer to have a proper function body for the `map` arrow function: `=> { ... }` with an early return inside.

src/h5web/dataset-visualizer/VisDisplay.tsx
axelboc: Perfect! Later, perhaps we can add a `getInitialMapperState` method to each vis definition in `VIS_DEF`.

src/h5web/dataset-visualizer/VisProvider.tsx
axelboc: What about just `values`? 🤷

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -14,10 +14,10 @@ import { TextureWorker } from './worker';
 import { useVisProps } from '../../dataset-visualizer/VisProvider';
 
 export function useData(): number[][] {
```
axelboc, line 16: I wonder if this hook is still needed. If the code is correct in the provider, then the check and throw are superfluous. If `rawDims` doesn't have a length of 2, the vis will break anyway.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -14,10 +14,10 @@ import { TextureWorker } from './worker';
...
-  const { rawValues, rawDims } = useVisProps();
+  const { visValues: mappedValues, rawDims } = useVisProps();
 
   if (rawDims.length === 2) {
-    return rawValues;
+    return mappedValues;
   }
 
   throw new Error('Data not supported by HeatmapVis');
```
axelboc, line 23: I think you can just return `visValues` as is for now, without any check/throw. You did remove the throw in the line vis' hook and I think it makes sense.

axelboc:
Great refactoring, btw! 💯 I think the code is a lot more comprehensible and scalable to complex mappings, now.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -1,88 +1,88 @@
...
+  mapperState: DimensionMapping;
   onChange: (d: DimensionMapping) => void;
 }
 
 function DimensionMapper(props: Props): JSX.Element {
-  const { activeVis, onChange } = props;
-  const { rawDims, mapping } = useVisProps();
+  const { activeVis, rawDims, mapperState, onChange } = props;
+
+  function showSlicingSlider(
```
axelboc, line 18: this type of function is usually named `renderSomething`

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -1,88 +1,88 @@
...
+    slicingIndex: MappingType,
+    dimension: number
+  ): JSX.Element | undefined {
```
axelboc, line 21: `ReactElement` ;)

PR #160, opened 2020-05-29
Refactoring to use `ndarray` to handle vis data

loichuder:
After having a go at the mapping, we agreed on using a dedicated library to store the data used in the visualisations. 

This PR shows the use of [ndarray](https://github.com/scijs/ndarray) that I have chosen for the following reasons:
- It basically uses a 1D array with special indexing so that we have the feeling to access to a nD array. This means that performance is good whatever the dimensions chosen.
- It handles slicing quite nicely !
- Personally, it reminds me of NumPy arrays which I am used to use.
- It is used as the backbone for [numjs](https://github.com/nicolaspanel/numjs), a library well maintained for numerical calculations in JS.

Other contenders were:
- `numjs`: I prefered to stay "low-level" as I didn't need the extensive features it implements.
- `TensorFlow.js`: extremely well maintained but I felt the slicing was awkward while it is the main feature we need...

All in all, I feel the use of `ndarray` simplifies *SO MUCH* the code that I could go on refactoring for hours but I felt I had to stop at some point :wink:. I didn't find any performance loss but we might want to check thoroughly. 
Of course, there are some downsides:
- Our code is less strongly typed as I had to use the `ndarray` type instead of array types where the dimensionnality was clear e.g. `number[]` for 1D and `number[][]` for 2D.
- Speaking of typing, we will need to provide the typing package for `ndarray-unpack`. I have a working definition file on my own (not included) but it is a bit dirty...
- We have to be careful as `ndarray` slicing returns *views* or the data which means the data is not copied and may not trigger `useMemo`s as expected.


src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -5,17 +5,16 @@ import React, {
...
-import { HDF5Dataset, HDF5Value } from '../providers/models';
+import { HDF5Dataset } from '../providers/models';
 import { useValue } from '../providers/hooks';
 import { isSimpleShape } from '../providers/utils';
 
 export interface VisProps {
-  values: HDF5Value;
-  rawDims: number[];
-  slicingIndices?: number[];
+  values: ndarray<number>;
```
loichuder, line 17: `VisProps` now only exposes an `ndarray` that is already sliced which means `slicingIndices` are no longer necessary. Also the dimensions are contained in `values.shape` meaning that `rawDims` is also superfluous.

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -32,19 +31,29 @@ function VisProvider(props: Props): ReactElement {
...
-
   const mappedValues = useMemo(() => {
     if (!rawValues) {
       return undefined;
     }
 
-    if (rawDims.length === 2 && xDim === 0) {
-      return transpose(rawValues);
+    if (rawDims.length !== 0) {
+      return ndarray(rawValues.flat(), rawDims);
```
loichuder, line 40: We only convert to a `ndarray` if it IS a array.

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -32,19 +31,29 @@ function VisProvider(props: Props): ReactElement {
...
+    if (mapperState === undefined) {
+      return mappedValues;
+    }
+    const slicingIndices = mapperState.map(val => (isNumber(val) ? val : null));
+    const slicedDims = rawDims.filter((d, i) => !isNumber(mapperState[i]));
+    return ndarray(
+      unpack(mappedValues.pick(...slicingIndices)).flat(),
+      slicedDims
+    );
+  }, [mappedValues, mapperState, rawDims]);
```
loichuder, line 56: Slicing is now done directly in the `VisProvider` so that it provides the sliced values. The `useMemo`s could probably be merged.

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -32,19 +31,29 @@ function VisProvider(props: Props): ReactElement {
...
 
     return rawValues;
-  }, [rawValues, rawDims.length, xDim]);
+  }, [rawValues, rawDims]);
+
+  const values = useMemo(() => {
+    if (mapperState === undefined) {
+      return mappedValues;
+    }
+    const slicingIndices = mapperState.map(val => (isNumber(val) ? val : null));
```
loichuder, line 50: `slicingIndices` are `null` for 'x' and 'y' so that the full dimension is taken (see https://github.com/scijs/ndarray#arraypickp0-p1-)

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -32,19 +31,29 @@ function VisProvider(props: Props): ReactElement {
...
+    const slicedDims = rawDims.filter((d, i) => !isNumber(mapperState[i]));
```
loichuder, line 51: The dimensions untouched by the slicing are only the ones of 'x' and 'y'

src/h5web/dataset-visualizer/VisProvider.tsx
```diff
@@ -32,19 +31,29 @@ function VisProvider(props: Props): ReactElement {
...
+    return ndarray(
+      unpack(mappedValues.pick(...slicingIndices)).flat(),
```
loichuder, line 53: - `mappedValues.pick(...slicingIndices)` is a view of the sliced array.
- The view is unpacked to get the values as JS array to copy them
- The JS array is flattened to generate a *new* ndarray with the sliced values inside and the sliced dimensions.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -3,17 +3,17 @@ import shallow from 'zustand/shallow';
...
 import PanZoomMesh from '../shared/PanZoomMesh';
 import VisCanvas from '../shared/VisCanvas';
+import { useVisProps } from '../../dataset-visualizer/VisProvider';
 
 function HeatmapVis(): JSX.Element {
-  const data = useData();
-  const dims = useDims();
-  const [rows, cols] = dims;
+  const { values: dataArray } = useVisProps();
+  const [rows, cols] = dataArray.shape;
```
loichuder, line 16: The only hook we need now is `useVisProps()` ! :tada:

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -42,7 +42,7 @@ function HeatmapVis(): JSX.Element {
         <TooltipMesh
           formatIndex={([x, y]) => `x=${Math.floor(x)}, y=${Math.floor(y)}`}
           formatValue={([x, y]) =>
-            format('.3')(data[Math.floor(y)][Math.floor(x)])
+            format('.3')(dataArray.get(Math.floor(y), Math.floor(x)))
```
loichuder, line 45: AS we have a `ndarray` now, we need its method `get` to get the value

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -27,7 +28,7 @@ function LineVis(): JSX.Element {
     <div className={styles.root}>
       <VisCanvas
         abscissaConfig={{
-          indexDomain: extendDomain([0, data.length - 1], 0.01),
+          indexDomain: extendDomain([0, values.size - 1], 0.01),
```
loichuder, line 31: As `values` is 1D, `size` is equal to the length of the flattened array but shorter to write.

src/h5web/visualizations/line/hooks.ts
```diff
@@ -6,23 +6,14 @@ import { findDomain, getAxisScale } from '../shared/utils';
...
-  }
-
-  return values as number[];
-}
-
 export function useDataDomain(): Domain | undefined {
-  const data = useData();
-  return useMemo(() => findDomain(data), [data]);
+  const { values } = useVisProps();
+  return useMemo(() => findDomain(values.data as number[]), [values.data]);
```
loichuder, line 11: `values.data` is the underlying 1D JS array used by `ndarray` to store the values. I will use it everything we need the values as `number[]`

src/h5web/visualizations/line/hooks.ts
```diff
@@ -6,23 +6,14 @@ import { findDomain, getAxisScale } from '../shared/utils';
...
 }
 
 export function useDataPoints(): Vector3[] | undefined {
-  const data = useData();
+  const { values } = useVisProps();
+  const data = values.data as number[];
```
loichuder, line 16: I have to dig into the typing of `values.data` to see if the casting here can be removed/replaced

src/h5web/providers/hsds/api.ts
```diff
@@ -35,12 +36,13 @@ export class HsdsApi implements ProviderAPI {
   private datatypes: Record<HDF5Id, HDF5Datatype> = {};
   private values: Record<HDF5Id, HDF5Value> = {};
 
-  constructor(private readonly domain: string) {
+  constructor(username: string, password: string, filepath: string) {
```
axelboc, line 39: Need `private readonly` in front of every param.

src/h5web/visualizations/line/hooks.ts
axelboc: I thin you can just specify `values` instead of `values.data` in the dependency array.

axelboc:
I've cherry picked your first two commits onto `master` and rebased this PR to keep only the third commit. I needed the updated code of the `DimensionMapper` for my work on the toggle button groups.

src/h5web/providers/hsds/api.ts
loichuder: Do I really need to if I only want to keep the config as a member of the class and not the individual parameters ?

src/demo-app/DemoApp.tsx
```diff
@@ -5,6 +5,10 @@ import SilxProvider from '../h5web/providers/silx/SilxProvider';
 import MockProvider from '../h5web/providers/mock/MockProvider';
 import HsdsProvider from '../h5web/providers/hsds/HsdsProvider';
 
+const HSDS_USERNAME = process.env.REACT_APP_HSDS_USERNAME || '';
+const HSDS_PASSWORD = process.env.REACT_APP_HSDS_PASSWORD || '';
+const HSDS_FILEPATH = process.env.REACT_APP_HSDS_FILEPATH || '';
```
axelboc, line 10: Since `.env` will always have empty values for these vars, they should already default to empty strings here.

loichuder: Actually, TS gives the type of `process.env.REACT_APP_HSDS_USERNAME` as `string | undefined`...

PR #161, opened 2020-06-02
Fix dim mapper not appearing on initial dataset selection

axelboc:
When the first dataset I open is a 1D dataset (line vis by default), the dimension mapper doesn't appear. I have to switch to the matrix vis and come back to the line vis for the mapper to appear.

This commit fixes this issue, and applies the fix to the other use of the derived state pattern. It also fixes an issue with the line vis when `slicingIndices` is an empty array, which occurs when there are no indices to slice (e.g. line vis with 1D dataset).

### DIFF
--- a/src/h5web/dataset-visualizer/VisDisplay.tsx
+++ b/src/h5web/dataset-visualizer/VisDisplay.tsx
@@ -15,13 +15,10 @@ interface Props {
 
 function VisDisplay(props: Props): JSX.Element {
   const { activeVis, dataset } = props;
-  const { Component: VisComponent } = VIS_DEFS[activeVis];
-
   const datasetDims = isSimpleShape(dataset.shape) ? dataset.shape.dims : [];
-  const [mapperState, setMapperState] = useState<DimensionMapping | undefined>(
-    undefined
-  );
-  const [prevVis, setPrevVis] = useState(activeVis);
+
+  const [prevVis, setPrevVis] = useState<Vis>();
+  const [mapperState, setMapperState] = useState<DimensionMapping>();
 
   // Update mapping when vis changes
   // https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops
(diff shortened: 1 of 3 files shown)


src/h5web/dataset-visualizer/VisDisplay.tsx
```diff
@@ -15,13 +15,10 @@ interface Props {
...
   const { activeVis, dataset } = props;
-  const { Component: VisComponent } = VIS_DEFS[activeVis];
-
   const datasetDims = isSimpleShape(dataset.shape) ? dataset.shape.dims : [];
-  const [mapperState, setMapperState] = useState<DimensionMapping | undefined>(
-    undefined
-  );
-  const [prevVis, setPrevVis] = useState(activeVis);
+
+  const [prevVis, setPrevVis] = useState<Vis>();
```
axelboc, line 20: By initialising `prevState` to nothing (i.e. `undefined`), we go through the derived state `if` block below on first render, which in turn initialises `mapperState` appropriately.  Before, we weren't going through the `if` block, so the `mapperState` was initialised to `undefined`.

In `DatasetVisualizer` above, I had worked around the issue by duplicating the logic in the initial state. I think going through the derived state `if` block is a cleaner solution (especially since rendering of the children is canceled).

src/h5web/dataset-visualizer/VisDisplay.tsx
```diff
@@ -15,13 +15,10 @@ interface Props {
 
 function VisDisplay(props: Props): JSX.Element {
   const { activeVis, dataset } = props;
-  const { Component: VisComponent } = VIS_DEFS[activeVis];
-
   const datasetDims = isSimpleShape(dataset.shape) ? dataset.shape.dims : [];
-  const [mapperState, setMapperState] = useState<DimensionMapping | undefined>(
-    undefined
-  );
```
axelboc, line 23: `useState` uses `undefined` by default, and when you write `useState<DimensionMapping>()`, the implied type is `DimensionMapping | undefined`.

PR #162, opened 2020-06-02
Implement toggle button group

axelboc:
I've used the new component, `ToogleGroup` everywhere (display/inspect toggle, dimension mapper, and line/points/both toggle in the line vis toolbar)

image

I have to use a background colour for unselected buttons to clearly outline that they belong to button groups.


src/h5web/BreadcrumbsBar.tsx
```diff
@@ -26,15 +28,13 @@ function BreadcrumbsBar(props: Props): JSX.Element {
...
-        aria-label="Toggle explorer sidebar"
-        aria-pressed={isExplorerOpen}
-        onClick={onToggleExplorer}
-      >
-        <FiSidebar />
-      </button>
+      <ToggleBtn
+        label="Toggle explorer sidebar"
+        icon={FiSidebar}
+        hideLabel
```
axelboc, line 34: Prop to hide the label, for the sidebar button.

src/h5web/BreadcrumbsBar.tsx
```diff
@@ -48,26 +48,17 @@ function BreadcrumbsBar(props: Props): JSX.Element {
...
-          role="tab"
-          className={styles.btn}
-          aria-selected={isInspecting}
-          onClick={() => onSetInspecting(true)}
-        >
-          Inspect
-        </button>
-      </div>
+      <ToggleGroup
+        role="tablist"
```
axelboc, line 52: The display/inspect toggle group is a `tablist`, not a `radiogroup` like the others. I've decided to make the `role` a mandatory prop instead of choosing a default, as I think it provides valuable information.

src/h5web/BreadcrumbsBar.tsx
```diff
@@ -48,26 +48,17 @@ function BreadcrumbsBar(props: Props): JSX.Element {
...
+        ariaLabel="Viewer mode"
+        value={String(isInspecting)}
+        onChange={val => {
+          onChangeInspecting(val === 'true' || false);
```
axelboc, line 56: The state is a boolean, but the values can only be string. I was unable to add a generic type to `ToggleGroup` because one cannot create a context with a generic type in React - cf. https://stackoverflow.com/questions/51448291/how-to-create-a-generic-react-component-with-a-typed-context-provider.

I decided it was cleaner to force `value` to be a string rather than to use a union type `string | boolean` and to have to always force-cast the value in `onChange`.

src/h5web/BreadcrumbsBar.tsx
```diff
@@ -48,26 +48,17 @@ function BreadcrumbsBar(props: Props): JSX.Element {
...
+        }}
+      >
+        <ToggleGroup.Btn label="Display" value="false" />
+        <ToggleGroup.Btn label="Inspect" value="true" />
```
axelboc, line 60: `ToggleGroup.Btn` is very similar to `ToggleBtn`, except that it reads the current `string` value of its parent toggle group from context (instead of reading its own `boolean` value from a prop). I will probably move their styles into a shared global utility class later on.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -54,40 +54,43 @@ function DimensionMapper(props: Props): JSX.Element {
     );
   }
 
-  function renderDimensionButton(dimension: 'x' | 'y'): JSX.Element {
+  function renderAxisMapper(axis: 'x' | 'y'): JSX.Element {
```
axelboc, line 57: I renamed a couple of things here (function and class names).

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -54,40 +54,43 @@ function DimensionMapper(props: Props): JSX.Element {
...
+    const selectedDim = mapperState.indexOf('x');
     return (
-      <div className={styles.dimensionButtonWrapper}>
-        <span className={styles.dimensionName}>{dimension}</span>
-        <ButtonGroup
-          ariaLabel={`Dimension as ${dimension}`}
-          className={styles.buttonGroup}
-          buttonClassName={styles.btn}
-          buttons={Object.keys(rawDims).map(dimKey => {
```
axelboc, line 65: I've used the compound components pattern: https://kentcdodds.com/blog/compound-components-with-react-hooks

`ToggleGroup` provides the currently selected value of the group through a local context, which its child `ToggleGroup.Btn` components read in order to determine whether they are the selected button. The group also provides the `onChange` callback function through context so the buttons can call it when they are clicked.

src/h5web/dataset-visualizer/DimensionMapper.tsx
```diff
@@ -54,40 +54,36 @@ function DimensionMapper(props: Props): JSX.Element {
...
+              newMapperState[selectedDim] = 0; // reset slicing index of previously selected dimension
+              newMapperState[newDim] = axis; // assign axis to newly selected dimension
+              onChange(newMapperState);
+            }
+          }}
+        >
+          {Object.keys(rawDims).map(dimKey => (
+            <ToggleGroup.Btn key={dimKey} label={`D${dimKey}`} value={dimKey} />
+          ))}
+        </ToggleGroup>
```
axelboc, line 79: I've used the compound components pattern: https://kentcdodds.com/blog/compound-components-with-react-hooks

`ToggleGroup` provides the currently selected value of the group through a local context, which its child `ToggleGroup.Btn` components read in order to determine whether they are the selected button. The group also provides the `onChange` callback function through context so the buttons can call it when they are clicked.

src/h5web/visualizations/line/LineToolbar.tsx
```diff
@@ -19,35 +19,26 @@ function LineToolbar(): JSX.Element {
...
-        ]}
-      />
-      <Toggler
+        value={curveType}
+        onChange={setCurveType}
+      >
+        <ToggleGroup.Btn label="Line" value={CurveType.LineOnly} />
+        <ToggleGroup.Btn label="Points" value={CurveType.GlyphsOnly} />
+        <ToggleGroup.Btn label="Both" value={CurveType.LineAndGlyphs} />
+      </ToggleGroup>
```
axelboc, line 31: This is the simplest example of the new `ToggleGroup` component.

src/h5web/visualizations/shared/ScreenshotButton.module.css
```diff
@@ -5,7 +5,7 @@
   color: inherit;
 }
 
-.box {
+.btnLike {
```
axelboc, line 8: I took inspiration from your `btnAspect` class but I also remembered having seen the naming `fooLike` being used before for this very purpose - i.e. for elements that "look like" other elements. It's quite common for headings, notably (e.g. `h2Like`)

src/h5web/visualizations/line/LineToolbar.tsx
```diff
@@ -19,35 +19,26 @@ function LineToolbar(): JSX.Element {
...
+      <Separator />
+
+      <ToggleBtn
         label="Symlog"
         icon={MdLinearScale}
         value={hasYLogScale}
         onChange={toggleYLogScale}
       />
-      <Toggler
+      <ToggleBtn
```
axelboc, line 41: Renaming this for consistency with `ToggleGroup.Btn`.

src/h5web/visualizations/shared/ToggleGroup.tsx
```diff
@@ -0,0 +1,71 @@
...
+    <ToggleGroupContext.Provider value={{ role, value, onChange }}>
+      <div className={styles.toggleGroup} role={role} aria-label={ariaLabel}>
+        {children}
+      </div>
+    </ToggleGroupContext.Provider>
+  );
+}
+
+ToggleGroup.Btn = Btn;
+export default ToggleGroup;
```
axelboc, line 71: Since `Btn` and `ToggleGroupContext` are not meant to be used without `ToggleGroup`, I decided to put everything in the same file and only expose `ToggleGroup` and `ToggleGroup.Btn`.

axelboc:
Merging this now, but we can come back to it next week.

src/h5web/visualizations/shared/ToggleBtn.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+function ToggleBtn(props: Props): JSX.Element {
+  const { label, icon: Icon, hideLabel, value, onChange } = props;
+
+  return (
+    <button
+      className={styles.btn}
+      type="button"
+      aria-label={hideLabel ? label : undefined}
```
axelboc, line 20: Putting the label in an `aria-label` attribute when it's hidden.

PR #163, opened 2020-06-03
Replace drop-down library and improve design of colour map selector

axelboc:
I finally found a promising drop-down library: https://github.com/davidtheclark/react-aria-menubutton. It provides low-level, unstyled components and it's up to us to combine them and style them the way we like.

The API is really simple and doesn't get in the way like `react-select`'s API -- we can pass our own classes, specify which HTML tags to use, etc.

image

Only feature missing and not trivial to bring back is adjusting the height of the menu to the available viewport height.


axelboc:
Alright, I've re-implemented the dynamic height feature. The menu can no longer overflow the viewport:

image

package.json
```diff
@@ -64,18 +64,15 @@
...
-    "@types/react-select": "^3.0.11",
     "@types/react-slider": "^1.0.0",
-    "@types/react-virtualized-auto-sizer": "^1.0.0",
     "@types/react-window": "^1.8.1",
     "@wsmd/eslint-config": "^1.2.0",
     "eslint": "^6.8.0",
-    "node-sass": "^4.13.1",
     "npm-run-all": "^4.1.5",
     "prettier": "^1.19.1",
-    "react-reflex": "^3.0.22",
```
axelboc, line 78: I've cleaned up some of the dependencies.

src/h5web/visualizations/heatmap/ColorMapSelector.tsx
```diff
@@ -1,137 +1,83 @@
...
-  },
-  {
-    label: 'Cyclical',
-    options: convertToOptions(CYCLICAL),
-  },
-  {
-    label: 'Diverging',
-    options: convertToOptions(DIVERGING),
-  },
-];
```
axelboc, line 45: No need to pre-process the color maps with `convertToOptions` since we no longer need to pass an `options` object -- we just render the options (`MenuItem`) ourselves instead.

src/h5web/visualizations/heatmap/ColorMapSelector.tsx
```diff
@@ -1,137 +1,83 @@
...
+import styles from './ColorMapSelector.module.css';
+import { INTERPOLATOR_GROUPS, INTERPOLATORS } from './interpolators';
+import { ColorMap, D3Interpolator } from './models';
+import { generateCSSLinearGradient } from './utils';
 
-function ColorMapLabel(props: ColorMapOption): JSX.Element {
-  const { label, value } = props;
+const MENU_IDEAL_HEIGHT = 320; // 20rem
+const MENU_TOP = 87; // HACK: height of breadcrumbs bar + height of toolbar
+const MENU_BOTTOM = 16; // offset from bottom of viewport
```
axelboc, line 13: These constants are for the dynamic height feature. We need to know the top position of the menu on the viewport, but `useMeasure` returns parent-relative positions... I decided to hard code `MENU_TOP` for now, hoping that `react-use` will improve its hook in the future.

src/h5web/visualizations/heatmap/HeatmapToolbar.tsx
```diff
@@ -22,7 +24,8 @@ function HeatmapToolbar(): ReactElement {
     <Toolbar>
       <DomainSlider />
       <Separator />
-      <ColorMapSelector />
+
+      <ColorMapSelector value={colorMap} onChange={setColorMap} />
```
axelboc, line 28: The selector is now controlled for consistency with the other toolbar components (the domain slider will be next).

src/h5web/visualizations/heatmap/interpolators.ts
```diff
@@ -101,3 +101,11 @@ export const INTERPOLATORS = {
...
   ...DIVERGING,
 };
+
+export const INTERPOLATOR_GROUPS = {
+  Common: MOST_USED,
+  'Single hue': SINGLE_HUE,
+  'Multi hue': MULTI_HUE,
+  Cyclical: CYCLICAL,
+  Diverging: DIVERGING,
+};
```
axelboc, line 111: This is what's left from the now-removed `colorMapOptions ` object.

axelboc:
Always nice to see 😄 

image

PR #165, opened 2020-06-04
Update dependencies (minor only)

axelboc:
- R3F's `<Dom />` element moved to a separate library called `drei` and was renamed [`<HTML />`](https://github.com/react-spring/drei#%EF%B8%8F-html-)
- R3F's documentation has been moved out of their README and [into separate files](https://github.com/react-spring/react-three-fiber#api)
- `react-icons`'s website has moved to https://react-icons.github.io/react-icons/#/ and includes a couple more icons sets.
- [TypeScript 3.9 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/) -- notable changes for us:
  - [// @ts-expect-error Comments](https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#ts-expect-error-comments) - could be useful to test our typings
  - [Quick Fixes for Missing Return Expressions](https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#missing-returns) - let VSCode either add a `return` inside an arrow function with a `{ }` block, or replace the `{ }` block with parens.


.vscode/settings.json
```diff
@@ -7,5 +7,7 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
   },
-  "typescript.tsdk": "node_modules/typescript/lib"
+
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "typescript.enablePromptUseWorkspaceTsdk": true
```
axelboc, line 12: This should prompt VSCode users when the version of TS they're using is not the same one as the project's TS version.

PR #167, opened 2020-06-04
Upgrade dependencies (major)

axelboc:
`npm outdated` now shows:

image

I couldn't upgrade:

- `react-use` because of a typing bug in `useMeasure` -- I've opened a PR to fix it: https://github.com/streamich/react-use/pull/1267
- `eslint` because `@wsmd/eslint-config` doesn't yet support ESLint v7 (and is likely to not support it for a while)
- `@types/jest` and `@types/node` - those versions are specified by CRA


src/h5web/providers/utils.ts
```diff
@@ -1,4 +1,4 @@
-import nanoid from 'nanoid';
+import { nanoid } from 'nanoid';
```
axelboc, line 1: Breaking change in `nanoid`.

axelboc:
Arf the tests fails because of `nanoid`, or rather because of CRA: https://github.com/ai/nanoid/issues/205 -- I thought CRA had released the fix, but it seems they're pushing everything onto the upcoming v4.0.... 😢  Hopefully they release it soon.

I'll close the PR until then, but I'll keep the branch -- I'll re-open it once CRA v4 is out.

PR #168, opened 2020-06-08
Improve domain slider design

axelboc:
image
image
image

### DIFF
--- a/src/h5web/visualizations/heatmap/HeatmapToolbar.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapToolbar.tsx
@@ -10,6 +10,9 @@ import ColorMapSelector from './ColorMapSelector';
 
 function HeatmapToolbar(): ReactElement {
   const {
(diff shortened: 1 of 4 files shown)


src/h5web/visualizations/heatmap/HeatmapToolbar.tsx
```diff
@@ -22,7 +25,13 @@ function HeatmapToolbar(): ReactElement {
...
   return (
     <Toolbar>
-      <DomainSlider />
+      {dataDomain && (
+        <DomainSlider
+          dataDomain={dataDomain}
+          value={customDomain}
+          onChange={setCustomDomain}
+        />
+      )}
```
axelboc, line 34: I also turn the component into a controlled component.

axelboc:
That's the last toolbar component that needed refactoring! Now I need to share the styles -- perhaps create a generic button or button-like component -- or just some shared CSS classes...

PR #172, opened 2020-06-10
Compose toolbar button styles

### LINKED ISSUE #147: Refactor toolbar components
Now that the toolbar has a few different components, it's time to give it some love.

### Design/UX

The toolbar components should :

- look more like toolbar components;
- be more consistent with one another visually;
- take less space.

Here's a mockup of what I'm thinking:

toolbar

I think the visual elements shouldn't span the height of the toolbar like they do now. However, I think the clickable area of each component should, to avoid frustrations. The clickable areas as I see them are drawn in red.

Note that the burger icon on the right is the overflow menu as per #126.

### Code

The code should be refactored :

- to increase re-usability of components and avoid duplicating logic and styles (e.g. use the same button component for single toggle buttons and in toggle button groups) -- it's particularly important for components not to be coupled to the visualisation configs/states;
- to reduce boilerplate when adding new components -- it'd be good not to have to create a new component each time we add something to the toolbar;
- to prepare for #126:
  - perhaps each component could access a `condensed` context prop provided by the responsive toolbar, that would tell it to render itself in a condensed format to save space;
  - we could also imagine an optional `priority` prop on every toolbar component to give the responsive toolbar a hint as to which components should be kept visible in priority, and which can be moved to the overflow menu when space is lacking.

axelboc:
I moved the core toolbar button styles to `Toolbar.module.css` and I compose the classes from the components that need them. I initially tried creating a `ToolbarBtn` component, then a `BtnLike` component, before I remembered that it's possible to compose classes from other CSS modules (I didn't want to make the button styles global).

This fixes #147.


src/h5web/visualizations/heatmap/DomainSlider.module.css
```diff
@@ -26,50 +26,35 @@
...
-  display: flex;
-  align-items: center;
+  composes: btn from '../shared/Toolbar.module.css';
   padding: 0 0.125rem;
 }
 
+.resetBtn:disabled {
+  opacity: 0.3;
+  pointer-events: none;
+}
```
loichuder, line 48: Just wondering if this should be put in the `btn` class.

PR #176, opened 2020-06-11
Implemented LogScale for LineVis (disabled if the values are not supported)

loichuder:
image

Following our discussion on #74, I choose the following implementation:
> 1.Deactivate the log scale button when the data contain negative or null values.

This PR achieves that by implementing the following:
- Refactor to implement multiple scaleTypes in AxisConfig
- Add the possibility to disable buttons of ToggleGroup
- Disable Log button if the domain does not allow it.

The last item was a bit tricky as it meant that we had to propagate information coming from the data to the `LineToolbar` that was until now completely decoupled from the data. 
The propagation was done by adding a boolean to the `LineConfig` that tells if the log button should be disabled. This boolean is reset at each data change in `LineVis`.
I felt this was the best solution as `LineConfig` is the only entry point of `LineToolbar` but I am happy to get feedback.

### DIFF
--- a/src/h5web/visualizations/line/LineVis.tsx
+++ b/src/h5web/visualizations/line/LineVis.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useMemo } from 'react';
+import React, { ReactElement } from 'react';
 import shallow from 'zustand/shallow';
 import { format } from 'd3-format';
 import type ndarray from 'ndarray';
@@ -8,7 +8,8 @@ import VisCanvas from '../shared/VisCanvas';
 import PanZoomMesh from '../shared/PanZoomMesh';
 import { useLineConfig } from './config';
 import TooltipMesh from '../shared/TooltipMesh';
-import { extendDomain, findDomain } from '../shared/utils';
+import { extendDomain } from '../shared/utils';
+import { useAxisDomain } from './hooks';
 
 interface Props {
   dataArray: ndarray<number>;
@@ -17,14 +18,12 @@ interface Props {
 function LineVis(props: Props): ReactElement {
   const { dataArray } = props;
 
-  const [showGrid, hasYLogScale] = useLineConfig(
-    (state) => [state.showGrid, state.hasYLogScale],
+  const [showGrid, scaleType] = useLineConfig(
+    (state) => [state.showGrid, state.scaleType],
     shallow
   );
 
-  const dataDomain = useMemo(() => {
-    return findDomain(dataArray.data as number[]);
-  }, [dataArray]);
+  const dataDomain = useAxisDomain(dataArray, scaleType);
 
   if (!dataDomain) {
     return <></>;
@@ -38,9 +37,9 @@ function LineVis(props: Props): ReactElement {
           showGrid,
         }}
         ordinateConfig={{
-          dataDomain: extendDomain(dataDomain, 0.01),
+          dataDomain,
           showGrid,
-          isLog: hasYLogScale,
+          scaleType,
         }}
       >
         <TooltipMesh
--- a/src/h5web/visualizations/line/config.ts
+++ b/src/h5web/visualizations/line/config.ts
@@ -1,18 +1,21 @@
 import { StorageConfig, createPersistableState } from '../../storage-utils';
 import { CurveType } from './models';
+import { ScaleType } from '../shared/models';
 
 type LineConfig = {
   curveType: CurveType;
   setCurveType: (type: CurveType) => void;
+
   showGrid: boolean;
   toggleGrid: () => void;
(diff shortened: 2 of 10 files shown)


src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -24,6 +37,14 @@ function LineVis(): JSX.Element {
     return <></>;
   }
 
+  const yDomain = extendDomain(dataDomain, 0.01);
+  disableLog(yDomain[0] * yDomain[1] <= 0);
```
loichuder, line 41: As the domain of the log scale is the extended domain, the check on domain boundaries must be done on this very domain. It can indeed happen that `dataDomain` spans strictly positive values but `yDomain` does not as it is extended.
We might need to change something here as it means that the log might be disabled in this case whereas the data spans strictly positive values...

Oh, and `scaleLog` works without problem if the domain spans strictly negative values. So we only need to check that the domain does not span positive AND negative (or null) values.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -24,6 +37,14 @@ function LineVis(): JSX.Element {
...
+
+  // Fallback on SymLog if Log is disabled
+  if (scaleType === ScaleType.Log && isLogDisabled) {
+    setScaleType(ScaleType.SymLog);
+  }
```
loichuder, line 46: This is needed as `scaleType` is persistent between datasets. Meaning me might switch from a dataset where the log is supported to one where it is one. In this case, the `scaleType` must not stay `Log`.

src/h5web/visualizations/line/config.ts
```diff
@@ -1,18 +1,21 @@
...
 
 type LineConfig = {
   curveType: CurveType;
   setCurveType: (type: CurveType) => void;
   showGrid: boolean;
   toggleGrid: () => void;
-  hasYLogScale: boolean;
-  toggleYLogScale: () => void;
+  scaleType: ScaleType;
+  setScaleType: (type: ScaleType) => void;
```
loichuder, line 11: The boolean is replaced by an enum member `scaleType`

src/h5web/visualizations/line/config.ts
```diff
@@ -22,8 +25,9 @@ export const [useLineConfig] = createPersistableState<LineConfig>(
     setCurveType: (type: CurveType) => set({ curveType: type }),
     showGrid: true,
     toggleGrid: () => set((state) => ({ showGrid: !state.showGrid })),
-    hasYLogScale: false,
-    toggleYLogScale: () =>
-      set((state) => ({ hasYLogScale: !state.hasYLogScale })),
+    scaleType: ScaleType.Linear,
+    setScaleType: (type: ScaleType) => set({ scaleType: type }),
+    isLogDisabled: false,
+    disableLog: (b: boolean) => set({ isLogDisabled: b }),
```
loichuder, line 31: As explained in the PR intro, this tells if the log button should be disabled in the `LineToolbar` and is reset at each change of dataset in `LineVis`.

src/h5web/visualizations/shared/ToggleGroup.module.css
```diff
@@ -8,6 +8,10 @@
   padding: 1px;
 }
 
+.btn:disabled {
+  opacity: 0.5;
+}
```
axelboc, line 13: should we just increase the opactiy of the toolbar's `btn` class?

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -32,9 +40,13 @@ function LineVis(): JSX.Element {
           showGrid,
         }}
         ordinateConfig={{
-          dataDomain: extendDomain(dataDomain, 0.01),
+          dataDomain: extendDomain(
+            dataDomain,
+            0.05,
+            scaleType === ScaleType.Log
+          ),
```
loichuder, line 47: I wonder if the domain extension could be put in the `setDataDomain` from the store as well...

src/h5web/visualizations/shared/utils.ts
```diff
@@ -37,10 +37,18 @@ export function computeVisSize(
...
+): Domain {
   const [min, max] = bareDomain;
-  const extension = (max - min) * extendFactor;
 
+  if (log) {
+    return [min / (1 + extendFactor), max * (1 + extendFactor)];
+  }
+
+  const extension = (max - min) * extendFactor;
   return [min - extension, max + extension];
```
loichuder, line 52: The `extendDomain` now accepts a log bool to extend the domain appropriately for log scales. This way, domain bounds never change sign.

This works fine for now but I plan to review this in the future so that we get consistent domain extensions (even for linear scale).

PR #177, opened 2020-06-11
Reset texture when changing dimensions

### LINKED ISSUE #174: [Heatmap] Swapping x/y dimensions breaks texture while loading
The axis system uses the new shape while the texture is being recomputed.

image

axelboc:
Fix #174

### DIFF
--- a/src/h5web/visualizations/heatmap/HeatmapVis.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.tsx
@@ -3,7 +3,7 @@ import shallow from 'zustand/shallow';
 import { format } from 'd3-format';
 import styles from './HeatmapVis.module.css';
 import ColorBar from './ColorBar';
-import { useValues } from './hooks';
+import { useValues, useDims } from './hooks';
 import Mesh from './Mesh';
 import TooltipMesh from '../shared/TooltipMesh';
 import { useHeatmapConfig } from './config';
@@ -13,14 +13,13 @@ import { useDataArray } from '../../dataset-visualizer/VisProvider';
 
 function HeatmapVis(): JSX.Element {
   const dataArray = useDataArray();
-  const [rows, cols] = dataArray.shape;
-
   const [keepAspectRatio, showGrid] = useHeatmapConfig(
     (state) => [state.keepAspectRatio, state.showGrid],
     shallow
   );
 
   // width / height <=> cols / rows
+  const { rows, cols } = useDims();
   const aspectRatio = keepAspectRatio ? cols / rows : undefined;
 
   const values = useValues();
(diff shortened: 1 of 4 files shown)


src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -81,5 +81,14 @@ export function useTextureData(): TextureDataState {
...
   ]);
 
+  const [rows, cols] = useDims();
+  const dims = `${rows}x${cols}`;
+  const [prevDims, setPrevDims] = useState<string>();
+
+  if (prevDims !== dims) {
+    setPrevDims(dims);
+    mergeState({ textureData: undefined });
+  }
```
axelboc, line 91: Another great use case for derived state ^^ -- and it's all self-contained within the existing custom hook.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -13,14 +13,13 @@ import { useDataArray } from '../../dataset-visualizer/VisProvider';
...
   const dataArray = useDataArray();
-  const [rows, cols] = dataArray.shape;
-
   const [keepAspectRatio, showGrid] = useHeatmapConfig(
     (state) => [state.keepAspectRatio, state.showGrid],
     shallow
   );
 
   // width / height <=> cols / rows
+  const [rows, cols] = useDims();
```
axelboc, line 22: Just making use of the Heatmap's `useDims` hook

loichuder: I wonder if this hook is still relevant as you can get the same result from `useDataArray` and `dataArray.shape`. 

On the other hand, we could give it meaning by returning an object of type `{cols, rows}` so that it avoids switching the two when getting them through the hook.

src/h5web/visualizations/heatmap/hooks.ts
loichuder: I have a minor complain about this: the textureData is not reset when the dimensions of the new dataset are the ones of the old dataset (ex: swapping dimensions of a square matrix). This poses no display problem as the texture stays the same as the dimensions are the same but still, we get a different behaviour from when the dimensions are different as the texture is set to gray in the latter case. Perhaps we could attach some `id` to the `dataArray` and reset the texture on this ?

This being said, your implementation covers 99% of the cases and for the remaining 1%, often the rendering is so quick that we don't notice anything...

axelboc: Yeah, it's on purpose. The texture loader gives a good-enough indication that something is about to change, so the fact that the texture underneath is stale is not a problem IMO. It's actually consistent with the change of color map/domain. Also, it saves a render.

If I had wanted to the clear the texture every time, I would have checked `dataArray.shape` directly as we get a new reference each time the mapping changes.

loichuder: Makes sense.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: Yeah, it's still good for typing, I think. Should we also change the type `Dims`, then?

Issue #170, opened 2020-06-10
Optimise uses of ndarray


axelboc:
- [x] Split `useMemo` in two in `VisProvider` (one for ndarray creation, one for slicing).
- [x] Simplify `unpack(mappedView).flat()` by unpacking a flat array directly.
- [ ] ~Find domain with cwise.~
- [ ] ~Create data points for line vis with cwise.~
- [ ] ~Create typed array for heatmap texture without calling `flat` (pick => shape 1D => unpack), or even without calling `Float64Array.from` (cwise + push to buffer).~ - no longer calling `flat`

loichuder:
WIP at #178. 
Waiting for `cwise` to support recent ES versions to go further (https://github.com/scijs/cwise/pull/25).

axelboc:
I think #198 brings us to a good state. The `cwise` optimisations no longer seem relevant anyway. There are still a few places where the code could be optimised (e.g. `findDomain` called twice with log scale and negative/positive domain, two uses of `findDomain(values.filter())` which amounts to two loops, etc.), but I don't think the gain would be worthwhile when compared to the increase in code complexity.

loichuder:
> There are still a few places where the code could be optimised (e.g. `findDomain` called twice with log scale and negative/positive domain, two uses of `findDomain(values.filter())` which amounts to two loops, etc.), but I don't think the gain would be worthwhile when compared to the increase in code complexity.

Yes, I think the performance is not an issue right now: things are running quite smoothly. Let's wait until we get more complains on that :smile:

PR #178, opened 2020-06-15
Optimisation of ndarray uses

loichuder:
See #170.


loichuder:
This is way too outdated to build on it. I will open a new one with the two first points of #170 (`useMemo` splitting and `pack`)

Issue #175, opened 2020-06-11
[Heatmap] Flicker after dimension swap


axelboc:
When swapping x/y dimensions (not sure if this happens when slicing), the visualization flickers after the texture has been recomputed - i.e. there seems to be an extra render happening. This is most visible with large datasets.

PR #179, opened 2020-06-15
Refactor Dataset Visualizer and improve typings throughout

axelboc:
Okay, so this was a bit insane...

My initial goal was to see what happened when I removed `VisProvider`. Turns out it didn't fix the flicker of #175, but it did make things a lot better, in my opinion:

- the custom hooks we had were quite tricky to follow;
- we had to propagate the provider inside R3F;
- people using the visualization components on their own would have had to wrap them with a provider (i.e. not the most straightfoward API);
- typing was strongly limited by the fact that the provider could not know which visualization it provided data for (i.e. no way to improve typing of the public API);
- the ndarray mapping code in the provider, meant that people wanting to use the scalar vis would have had to use a bundle that included ndarray, which would have been a huge waste;
- and more...

So, here's roughly how I achieved this goal:

1. First, I **changed the props of every visualization component**. Each component now receives one strongly typed prop: the value or array to display. This makes for a very clean and clear public API. People do have to convert whatever array they have to an ndarray, but that's it. Visualization components do not know anything about dimension mapping and datasets.
1. Then, I removed all the custom hooks that took care of retrieving the provider's value (`useDataArray`, `useDims` and so on). I passed props to child components as required, which led to more readable code, in my opinion. I also removed the provider forwarded to R3F.
1. Then I went up into the `DatasetVisualizer` and worked on removing the provider and passing actual props with proper types:
  1. In doing so, I realised that we could no longer retrieve and render the visualizations through `VIS_DEFS[activeVis].Component`, since the visualization components no longer shared the same public API (different prop names and types). I didn't want to go back to the consecutive `if (activeVis === Vis.Matrix)` blocks that we had initially in `VisDisplay` -- each visualization had to decide how to render itself. That's why I decided to had a `render` function to each vis definition in `VIS_DEFS`.
  1. Since each visualization is now free to decide how to render itself based on the input data, the dimension mapping process can now be performed as part of this render process, and only by visualizations that need it. That's what led me to create `MappedVis`, a component that takes care of converting the input data to a `dataArray` and then rendering a given visualization component with this `dataArray`. This component basically contains the code that used to be in `VisProvider`.
  1. At this point, I noticed that it didn't matter if the input data (i.e. the dataset value) was a number or whatever, because `HDF5Value` was set to `any`. I couldn't tell if my code was correct because TypeScript would let me do anything I wanted... That's when I realised that we needed to replace `any` with `unknown`, which basically tells TypeScript to not care about the type of a variable until you actually try to use it or pass it somewhere where a stronger type is needed. This changed forced me to type a couple of things better throughout the app, and to use assertion functions before passing the input data to the visualization components or to `MappedVis`.
1. At this point, I ran into state management issues in `DatasetVisualizer` and `VisDisplay`. The way I solved it was to move the `mapperState` and the dataset value state up into `DatasetVisualizer`, next to `activeVis`. This allowed me to considerably simplify the state reset logic when changing dataset, and to remove the `key` on `VisDisplay`. I kept `VisDisplay` even though it's quite dumb now, because it's convenient to not have to check whether `activeVis` is `undefined`.
1. The move of `mapperState` into `DatasetVisualizer` required me to change type `DimensionMapping`. I added `| undefined` to the type, which meant that we no longer had to make `mapperState` props optional with `?`. This led to some typing issues in `DimensionMapper`, which led me to refactor it by extracting the `SlicingSlider` and `AxisMapper` into their own components.

... and with all this, the viewer works just as well as before... 😄


src/h5web/dataset-visualizer/mapper/SlicingSlider.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+function SlicingSlider(props: Props): ReactElement {
+  const { dimension, slicingIndex, rawDims, mapperState, onChange } = props;
+
+  return (
+    <div key={dimension} className={styles.sliderWrapper}>
+      <span className={styles.sliderLabel}>D{dimension}</span>
+      <ReactSlider
+        key={`${mapperState.includes('y')}`}
```
loichuder, line 21: Now that this is in a separate component, a comment would be needed to explain the *raison d'être* of this key

src/h5web/dataset-visualizer/mapper/SlicingSlider.tsx
```diff
@@ -0,0 +1,44 @@
...
+        className={styles.slider}
+        trackClassName={styles.sliderTrack}
+        thumbClassName={styles.sliderThumb}
+        renderThumb={(thumbProps, state) => (
+          <div {...thumbProps}>{state.valueNow}</div>
+        )}
+        value={slicingIndex}
+        onChange={(value) => {
+          const newMapperState = mapperState.slice();
+          newMapperState[dimension] = Number(value);
```
loichuder, line 31: Is the `Number(value)` necessary here (rather than `value`) ?

src/h5web/metadata-viewer/AttributesInfo.tsx
```diff
@@ -25,7 +25,7 @@ function AttributesInfo(props: Props): JSX.Element {
           ({ name, value }: HDF5Attribute): JSX.Element => (
             <tr key={name}>
               <th>{name}</th>
-              <td>{value.toString()}</td>
+              <td>{JSON.stringify(value)}</td>
```
loichuder, line 28: Good catch :fishing_pole_and_fish: ! It is true that attribute values can be arrays.

src/h5web/visualizations/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+    Icon: FiGrid,
+    supportsDataset: (dataset) => {
+      const { type, shape } = dataset;
+      return isBaseType(type) && isSimpleShape(shape) && shape.dims.length >= 1;
+    },
+    defaultMappingState: (dataset) => {
+      const { dims } = dataset.shape as HDF5SimpleShape;
+      return dims.length === 1
+        ? ['x']
+        : [...range(dims.length - 2).fill(0), 'y', 'x'];
```
loichuder, line 70: Clever :100:

src/h5web/visualizations/shared/utils.ts
```diff
@@ -112,3 +112,15 @@ export function getTicksProp(
     ? { tickValues: getIntegerTicks(visibleDomain, numTicks) }
     : { numTicks };
 }
+
+export function assertNumOrStr(val: unknown): asserts val is number | string {
+  if (typeof val !== 'number' && typeof val !== 'string') {
+    throw new Error('Expected number or string');
+  }
+}
```
loichuder, line 120: Why not use the `ScalarValue` model here ?

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -1,22 +1,38 @@
...
+import { AxisSystemContext } from '../shared/AxisSystemProvider';
+import { getAxisScale } from '../shared/utils';
 
 const COLOR = '#1b998b';
 
-function DataCurve(): JSX.Element {
-  const points = useDataPoints();
+interface Props {
+  values: number[];
+}
```
loichuder, line 15: Just a minor thing: now that `DataCurve` accepts props I would put the `color` in it as an optional prop (with a fallback to `#12998b`)

src/h5web/visualizations/matrix/Cell.tsx
```diff
@@ -13,13 +13,14 @@ function Cell(props: GridChildComponentProps): JSX.Element {
...
   const dataValue = valueAccessor(rowIndex - 1, columnIndex - 1);
+
   return (
     <div
       className={styles.cell}
       style={style}
       data-bg={(rowIndex + columnIndex) % 2 === 1 ? '' : undefined}
     >
-      {format('.3e')(dataValue)}
+      {typeof dataValue === 'number' ? format('.3e')(dataValue) : dataValue}
```
loichuder, line 23: What prompted this ?

src/h5web/visualizations/shared/utils.ts
```diff
@@ -112,3 +112,15 @@ export function getTicksProp(
...
+
+export function assertArray<T>(val: unknown): asserts val is T[] {
+  if (!Array.isArray(val)) {
+    throw new Error('Expected array');
+  }
+}
```
loichuder, line 126: I wonder if we can go further than that by writing custom types for 1D and 2D ndarrays and a typeguard that will check the dimensionnality of the ndarray.
What do you think ? Perhaps in another PR ?

src/h5web/dataset-visualizer/mapper/SlicingSlider.tsx
axelboc: You're right, it's more correct to use `as number`. The value is already a number but TS doesn't know it.

src/h5web/visualizations/shared/utils.ts
axelboc: Just wanted it to be generic. To be perfectly honest, I don't think `ScalarValue` and `DataArray` are worth keeping. `MatrixVis` accepts a prop of type `ndarray<number | string>`, so I feel like sharing prop types between visualizations no longer makes much sense. Are you happy for me to remove them?

src/h5web/visualizations/matrix/Cell.tsx
axelboc: `MatrixVis` has a prop of type `ndarray<number | string>`. Why? Because of its support predicate:

```ts
isBaseType(type) && isSimpleShape(shape) && shape.dims.length >= 1
```

`isBaseType` checks for `[HDF5TypeClass.Integer, HDF5TypeClass.Float, HDF5TypeClass.String]`

axelboc: So we now support visualizing arrays of strings with `MatrixVis`.

src/h5web/visualizations/shared/utils.ts
axelboc: Hmm, I did wonder a bit about this, but decided it wasn't worth it. `MappedVis` doesn't really need stronger typing on its `value` prop since we just flatten the array right at the start. What matters here is telling TypeScript that the `unknown` value is just an `array` of some sort.

If the value array's dimensions don't match the dataset's metadata dimensions, it's an error in the file, so we don't really care what happens. If the vis breaks, good; if it doesn't, good. Throwing ourselves isn't worth it, since there could be other problems, like values in the array that are not numbers.

src/h5web/visualizations/shared/utils.ts
loichuder: I am all right with removing `ScalarValue` but I would keep `DataArray` for now as the single source of truth for the typing. I indeed foresee that the type `ndarray<number>` will change with improvements of #178.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -66,29 +59,18 @@ export function useTextureData(): TextureDataState {
...
+  // Reset texture when dimensions change to avoid rendering glitch while computing new texture
   const dims = `${rows}x${cols}`;
-  const [prevDims, setPrevDims] = useState<string>(dims);
-
-  if (prevDims !== dims) {
-    setPrevDims(dims);
-    mergeState({ textureData: undefined });
+  if (dims !== state.prevDims) {
+    mergeState({ textureData: undefined, prevDims: dims });
   }
```
axelboc, line 74: I had meant to do this but forgot: the previous dimensions are now stored in the same state as the texture. This avoids an extra `useState`. I've also added a comment.

src/h5web/visualizations/matrix/Cell.tsx
loichuder: All right :+1:

PR #181, opened 2020-06-15
Implemented LogScale for LineVis (hide points if the values are not supported)

loichuder:
Another go at the log scale but the approach is different than the one in #176. Here I choose the following implementation (discussed in #74):
> 2. Display only the positive values on which the log transformation was able to be done.

There are things already discussed in #176 but I feel it was cleaner to open a new PR.

### DIFF
--- a/src/h5web/visualizations/line/DataCurve.tsx
+++ b/src/h5web/visualizations/line/DataCurve.tsx
@@ -19,21 +19,29 @@ function DataCurve(props: Props): JSX.Element {
   const { values, color = DEFAULT_COLOR } = props;
 
   const { abscissaInfo, ordinateInfo } = useContext(AxisSystemContext);
-  const { size } = useThree();
+  const { camera, size } = useThree();
   const { width, height } = size;
 
   const dataGeometry = useMemo(() => {
     const abscissaScale = getAxisScale(abscissaInfo, width);
     const ordinateScale = getAxisScale(ordinateInfo, height);
 
-    const points = values.map(
-      (val, index) => new Vector3(abscissaScale(index), ordinateScale(val), 0)
-    );
+    const points = values.map((val, index) => {
+      const ordinate = ordinateScale(val);
+      return new Vector3(
+        abscissaScale(index),
+        // This is to avoid a three.js warning when ordinateScale(val) is Infinity
+        Number.isFinite(ordinate) ? ordinate : 0,
+        // Move NaN/Infinity out of the camera FOV (negative val for logScale).
+        // This allows to have only curve segments for the positive values
+        Number.isFinite(ordinate) ? 0 : camera.far
+      );
+    });
 
     const geometry = new BufferGeometry();
     geometry.setFromPoints(points);
     return geometry;
-  }, [abscissaInfo, height, ordinateInfo, values, width]);
+  }, [abscissaInfo, camera, height, ordinateInfo, values, width]);
 
   const curveType = useLineConfig((state) => state.curveType);
   const showLine = curveType !== CurveType.GlyphsOnly;
--- a/src/h5web/visualizations/line/LineVis.tsx
+++ b/src/h5web/visualizations/line/LineVis.tsx
@@ -9,6 +9,7 @@ import PanZoomMesh from '../shared/PanZoomMesh';
 import { useLineConfig } from './config';
 import TooltipMesh from '../shared/TooltipMesh';
 import { extendDomain, findDomain } from '../shared/utils';
+import { ScaleType } from '../shared/models';
 
 interface Props {
   dataArray: ndarray<number>;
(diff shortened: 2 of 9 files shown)


src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -19,21 +19,29 @@ function DataCurve(props: Props): JSX.Element {
...
     const points = values.map(
-      (val, index) => new Vector3(abscissaScale(index), ordinateScale(val), 0)
+      (val, index) =>
+        new Vector3(
+          abscissaScale(index),
+          // This is to avoid a three.js warning when ordinateScale(val) is Infinity
+          Number.isFinite(ordinateScale(val)) ? ordinateScale(val) : 0,
+          // Move NaN/Infinity out of the camera FOV (negative val for logScale).
+          // This allows to have only curve segments for the positive values
+          Number.isFinite(ordinateScale(val)) ? 0 : camera.far
```
axelboc, line 37: You can compute the scale values once beforehand to avoid calling the ordinate scale three times.

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -19,21 +19,29 @@ function DataCurve(props: Props): JSX.Element {
...
+        )
     );
 
     const geometry = new BufferGeometry();
     geometry.setFromPoints(points);
     return geometry;
-  }, [abscissaInfo, height, ordinateInfo, values, width]);
+  }, [abscissaInfo, camera.far, height, ordinateInfo, values, width]);
```
axelboc, line 44: Just `camera` here. `far` is probably mutated anyway, so it wouldn't trigger a re-render if it changed.

src/h5web/visualizations/line/hooks.ts
```diff
@@ -0,0 +1,18 @@
...
+): Domain | undefined {
+  return useMemo(() => {
+    const isLogScale = scaleType === ScaleType.Log;
+    const rawDomain = isLogScale
+      ? findDomain((dataArray.data as number[]).filter((x) => x > 0))
+      : findDomain(dataArray.data as number[]);
+
+    return rawDomain && extendDomain(rawDomain, 0.05, isLogScale);
+  }, [dataArray.data, scaleType]);
+}
```
loichuder, line 18: This is where it happens: the domain is only computed from positive values if we use a logscale. 
The hiding of negative values is handled when building the geometry.

src/h5web/visualizations/line/hooks.ts
```diff
@@ -0,0 +1,18 @@
...
+  scaleType: ScaleType
+): Domain | undefined {
+  return useMemo(() => {
+    const isLogScale = scaleType === ScaleType.Log;
+    const rawDomain = isLogScale
+      ? findDomain((dataArray.data as number[]).filter((x) => x > 0))
+      : findDomain(dataArray.data as number[]);
+
+    return rawDomain && extendDomain(rawDomain, 0.05, isLogScale);
+  }, [dataArray.data, scaleType]);
```
axelboc, line 17: `dataArray` here instead of `dataArray.data`. If `dataArray.data` changes, `dataArray` changes too, so no point in being overly precise.

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -19,21 +19,29 @@ function DataCurve(props: Props): JSX.Element {
...
-      (val, index) => new Vector3(abscissaScale(index), ordinateScale(val), 0)
+      (val, index) =>
+        new Vector3(
+          abscissaScale(index),
+          // This is to avoid a three.js warning when ordinateScale(val) is Infinity
+          Number.isFinite(ordinateScale(val)) ? ordinateScale(val) : 0,
+          // Move NaN/Infinity out of the camera FOV (negative val for logScale).
+          // This allows to have only curve segments for the positive values
+          Number.isFinite(ordinateScale(val)) ? 0 : camera.far
+        )
```
loichuder, line 38: While a bit clunky, this allows to move out of the FOV the data points corresponding to negative values **and** the lines connecting to them when displaying the lines.
Simply removing negative values from the data would create a curve where all points are connected which is not wanted.

src/h5web/visualizations/line/hooks.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function useAxisDomain(
+  dataArray: ndarray,
+  scaleType: ScaleType
+): Domain | undefined {
+  return useMemo(() => {
+    const isLogScale = scaleType === ScaleType.Log;
+    const rawDomain = isLogScale
+      ? findDomain((dataArray.data as number[]).filter((x) => x > 0))
+      : findDomain(dataArray.data as number[]);
```
axelboc, line 14: For readability, I would store `dataArray.data` in a constant and move the ternary inside `findDomain`.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -17,14 +18,12 @@ interface Props {
...
-    (state) => [state.showGrid, state.hasYLogScale],
+  const [showGrid, scaleType] = useLineConfig(
+    (state) => [state.showGrid, state.scaleType],
     shallow
   );
 
-  const dataDomain = useMemo(() => {
-    return findDomain(dataArray.data as number[]);
-  }, [dataArray]);
+  const dataDomain = useAxisDomain(dataArray, scaleType);
```
loichuder, line 26: This was made a single hook rather than put in the `LineConfig` as I had an issue where the domain of the `LineConfig` was out-of-phase with the data displayed. This led to at best, an additional render and at worst, a WebGL crash...

src/h5web/visualizations/shared/utils.ts
```diff
@@ -37,10 +37,18 @@ export function computeVisSize(
     : { width, height: width / aspectRatio };
 }
 
-export function extendDomain(bareDomain: Domain, extendFactor: number): Domain {
+export function extendDomain(
+  bareDomain: Domain,
+  extendFactor: number,
+  log?: boolean
```
axelboc, line 43: `isLog` ?

src/h5web/visualizations/line/DataCurve.tsx
axelboc: Very clever

src/h5web/visualizations/line/DataCurve.tsx
loichuder: :fishing_pole_and_fish:

axelboc:
So, should we add a fallback domain somewhere so it looks a bit less weird when slicing?

PR #186, opened 2020-09-10
Install Storybook and write AxisSystem stories

axelboc:
image

### DIFF
--- a/src/h5web/visualizations/shared/AxisSystem.tsx
+++ b/src/h5web/visualizations/shared/AxisSystem.tsx
@@ -5,6 +5,7 @@ import { format } from 'd3-format';
 import type { TickRendererProps } from '@vx/axis/lib/types';
 import { GridColumns, GridRows } from '@vx/grid';
 import { HTML } from 'drei';
+import { useUpdateEffect } from 'react-use';
 import styles from './AxisSystem.module.css';
 import type { AxisOffsets, Domain } from './models';
 import { getTicksProp, getAxisScale } from './utils';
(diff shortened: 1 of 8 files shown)


src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -34,12 +35,16 @@ function AxisSystem(props: Props): JSX.Element {
...
+  // `axisDomains` are the complete domains; `visibleDomains` change with the camera
   const [visibleDomains, setVisibleDomains] = useState<[Domain, Domain]>([
     abscissaInfo.domain,
     ordinateInfo.domain,
   ]);
 
+  useUpdateEffect(() => {
+    setVisibleDomains([abscissaInfo.domain, ordinateInfo.domain]);
+  }, [abscissaInfo.domain, ordinateInfo.domain]);
+
```
axelboc, line 47: This block allows changing domains dynamically in Storybook's "Controls" panel.

axelboc:
I need to add a section to the README.

src/h5web/visualizations/shared/AxisSystem.tsx
axelboc: After some manual testing, it did not seem to affect the app in any way. `useUpdateEffect` means the effect does not run on mount, which would be unnecessary, since on mount, the visible domain's state is already initialised with the axis domains.

loichuder: Why not use a regular `useEffect` ?

src/h5web/visualizations/shared/VisCanvas.tsx
```diff
@@ -9,7 +9,7 @@ import AxisSystemProvider from './AxisSystemProvider';
 
 const AXIS_OFFSETS = { vertical: 72, horizontal: 36, fallback: 16 };
 
-interface Props {
+export interface Props {
```
loichuder, line 12: Should we rename `Props` to `VisCanvasProps` avoid exporting an interface with a too generic name ?

loichuder (approved):
:tada: Long time no update :tada:

src/h5web/visualizations/shared/VisCanvas.tsx
axelboc: You're right, I put 

```js
export type { Props as VisCanvasProps };
```

at the bottom of the file instead, just before the default export.

axelboc: This allows us to keep the simple name `Props` inside the file.

PR #189, opened 2020-09-10
Improve AxisSystem stories

### LINKED ISSUE #187: [storybook] Minor complains on AxisSystem
- Why the AxisSystem does not take all available space ? Is it on purpose ?
image

- For `showGrid`, I would put `showGrid: false` in the config instead of removing it completely.

axelboc:
Fix #187 

- The canvas now fills the available space in every story.
- I've also added a story to demonstrate how to style the tick labels.

image

### DIFF
--- a/src/stories/AxisSystem.stories.tsx
+++ b/src/stories/AxisSystem.stories.tsx
@@ -44,18 +44,26 @@ export const AspectRatio = Template.bind({});
 
 AspectRatio.args = {
   abscissaConfig: { indexDomain: [0, 10], showGrid: true },
-  ordinateConfig: { indexDomain: [0, 1], showGrid: true },
-  aspectRatio: 10,
+  ordinateConfig: { indexDomain: [0, 2], showGrid: true },
+  aspectRatio: 10 / 2,
 };
 
 export const NoGrid = Template.bind({});
 
 NoGrid.args = {
-  abscissaConfig: { indexDomain: [-5, 20] },
-  ordinateConfig: { dataDomain: [0, 1] },
+  abscissaConfig: { indexDomain: [-5, 20], showGrid: false },
+  ordinateConfig: { dataDomain: [0, 2], showGrid: false },
 };
 
 export default {
   title: 'VisCanvas/AxisSystem',
   component: VisCanvas,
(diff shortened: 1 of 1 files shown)


src/stories/AxisSystem.stories.tsx
```diff
@@ -44,18 +44,49 @@ export const AspectRatio = Template.bind({});
...
+        display: 'flex',
+        fontFamily: 'monospace',
+        fontWeight: 'bold',
+        fontSize: '1.125rem',
+      }}
+    >
+      <VisCanvasStory />
+    </div>
+  ),
+];
```
loichuder, line 79: I don't understand the point of this story :thinking:

loichuder: Oh, it is for styling ticks. Still it is a shame to not be able to change the styling dynamically in storybook :slightly_frowning_face:

src/stories/AxisSystem.stories.tsx
```diff
@@ -44,18 +44,49 @@ export const AspectRatio = Template.bind({});
...
+  aspectRatio: 10 / 2,
 };
 
 export const NoGrid = Template.bind({});
 
 NoGrid.args = {
-  abscissaConfig: { indexDomain: [-5, 20] },
-  ordinateConfig: { dataDomain: [0, 1] },
+  abscissaConfig: { indexDomain: [-5, 20], showGrid: false },
+  ordinateConfig: { dataDomain: [0, 2], showGrid: false },
```
loichuder, line 55: Shame we can't use [Controls](https://storybook.js.org/docs/react/essentials/controls) with nested args...

loichuder (approved):
:+1:

src/stories/AxisSystem.stories.tsx
axelboc: You can define custom controls, but there's an ongoing issue on the add-on for supporting nested objects automatically, so I didn't bother.

For now, my goal is more to make sure components work in isolation. We can improve the documentation aspect later on.

loichuder: Let me rephrase that: "Shame we can't use Controls easily because of the nested args" ^^

Issue #184, opened 2020-09-08
[Heatmap] Add log scale


loichuder:
Symlog scale for Heatmap was added in #79 but there is no "true" log scale yet.

As in #74, we have to decide on the behaviour of unsupported values (i.e negative values). I see two options for now:
- deactivate completely the log option in the presence of such values
- allow the log to be selected and set the pixels of unsupported values to a neutral colour

Note that, contrary to the LineVis, the domain of the scale can be selected by the user. Therefore, the behaviour must be able to change dynamically.

loichuder:
I put some thoughts in this:
> - deactivate completely the log option in the presence of such values

Then, we would need a fallback behaviour when switching datasets/changing to a domain of unsupported values. Should it fallback to Symlog or Linear ? 

This might prove inconvenient for users as the scale type may end up switching from Log to other values when exploring datasets.

> - allow the log to be selected and set the pixels of unsupported values to a neutral colour

Tough to say what would be the neutral color as we support multiple colormaps of different hues. A bright red `(255, 0, 0)` fits most cases but is too eye-catching in my opinion.

Also, it will be tough to integrate in our `HeatmapVis` as it is. To allow this behaviour, we would indeed need to compute a domain of supported values (let's call it `supportedDomain`) as done in https://github.com/silx-kit/h5web/blob/master/src/h5web/visualizations/line/LineVis.tsx#L27 and I see several problems:
- Computing the domain must be done at two places (`ColorBar` and the hook computing `TextureData`) but `ColorBar` has no access to the `values` array. It would need then to be passed as props or through the store but neither feels right. The ideal would be to compute `supportedDomain` in the HOC `HeatmapVis`. We then run in the next problem
- The `dataDomain` and the `customDomain` are handled through the store. To integrate seamlessly the `supportedDomain`, it would need to be in the store as well. But we end up with three different domains in the store which hints me that we are doing something wrong
- Also, the interplay with `customDomain` is not clear. Should `customDomain` overwrite `supportedDomain` ? Then, what would happen of the DomainSlider: should it display the "true" `supportedDomain` or leave the user choose its own domain ? Or something in-between where we clamp the limits of the domain to the supported values ? On a related note, would it be possible to add some complexity to the `setCustomDomain` method to handle this interplay ?

@axelboc Your thoughts ?

axelboc:
How about showing "Log scale not supported for this dataset"? 😄

axelboc:
[...]

PR #192, opened 2020-09-14
Heatmap is working with LogScale with invalid values (set to white)

loichuder:
image

After our discussion in #184, I finally settled for the following behaviour:
- Pixels with invalid values are set to white color*
- The store is untouched: the `supportedDomain` is computed in the `HeatmapVis` as the single source of truth and passed to the components needing it.
- As a consequence, `customDomain` and its associated slider do **not** reflect the changes in `supportedDomain`

_* This is the behaviour in matplotlib so I expect users to be used to it:_
image

### DIFF
--- a/src/h5web/visualizations/heatmap/ColorBar.tsx
+++ b/src/h5web/visualizations/heatmap/ColorBar.tsx
@@ -1,29 +1,30 @@
 import React from 'react';
 import { AxisRight } from '@vx/axis';
 import { useMeasure } from 'react-use';
-import shallow from 'zustand/shallow';
 import { adaptedNumTicks } from '../shared/utils';
-import { useInterpolator } from './hooks';
-import { useHeatmapConfig } from './config';
 import styles from './HeatmapVis.module.css';
 import { generateCSSLinearGradient } from './utils';
-import { SCALE_FUNCTIONS } from '../shared/models';
+import { SCALE_FUNCTIONS, ScaleType, Domain } from '../shared/models';
+import type { ColorMap } from './models';
+import { INTERPOLATORS } from './interpolators';
 
-function ColorBar(): JSX.Element {
-  const [dataDomain, customDomain, scaleType] = useHeatmapConfig(
-    (state) => [state.dataDomain, state.customDomain, state.scaleType],
-    shallow
-  );
+interface Props {
+  domain?: Domain;
+  scaleType: ScaleType;
+  colorMap: ColorMap;
+}
 
-  const interpolator = useInterpolator();
+function ColorBar(props: Props): JSX.Element {
+  const { domain, scaleType, colorMap } = props;
+  const interpolator = INTERPOLATORS[colorMap];
   const [gradientRef, { height: gradientHeight }] = useMeasure();
 
-  if (!dataDomain) {
+  if (!domain) {
     return <></>;
   }
 
   const axisScale = SCALE_FUNCTIONS[scaleType]();
-  axisScale.domain(customDomain || dataDomain);
+  axisScale.domain(domain);
   axisScale.range([gradientHeight, 0]);
 
   return (
--- a/src/h5web/visualizations/heatmap/HeatmapVis.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect } from 'react';
-import shallow from 'zustand/shallow';
 import { format } from 'd3-format';
 import type ndarray from 'ndarray';
 import styles from './HeatmapVis.module.css';
@@ -9,7 +8,7 @@ import TooltipMesh from '../shared/TooltipMesh';
 import { useHeatmapConfig } from './config';
 import PanZoomMesh from '../shared/PanZoomMesh';
 import VisCanvas from '../shared/VisCanvas';
-import { getDims } from './utils';
+import { getDims, useMemoColorScaleDomain } from './utils';
 
 interface Props {
   dataArray: ndarray<number>;
@@ -21,19 +20,29 @@ function HeatmapVis(props: Props): JSX.Element {
   const { rows, cols } = getDims(dataArray);
   const values = dataArray.data as number[];
 
-  const [keepAspectRatio, showGrid] = useHeatmapConfig(
-    (state) => [state.keepAspectRatio, state.showGrid],
-    shallow
-  );
+  const {
+    dataDomain,
+    customDomain,
+    scaleType,
+    colorMap,
+    keepAspectRatio,
+    showGrid,
+    initDataDomain,
+  } = useHeatmapConfig();
 
   // width / height <=> cols / rows
   const aspectRatio = keepAspectRatio ? cols / rows : undefined;
 
-  const initDataDomain = useHeatmapConfig((state) => state.initDataDomain);
-
   useEffect(() => {
     initDataDomain(values);
   }, [initDataDomain, values]);
+
+  const domain = useMemoColorScaleDomain(
+    scaleType,
+    values,
+    dataDomain,
+    customDomain
+  );
 
   return (
     <div className={styles.root}>
@@ -52,9 +61,16 @@ function HeatmapVis(props: Props): JSX.Element {
           guides="both"
         />
         <PanZoomMesh />
-        <Mesh rows={rows} cols={cols} values={values} />
+        <Mesh
+          rows={rows}
+          cols={cols}
+          values={values}
+          domain={domain}
+          scaleType={scaleType}
+          colorMap={colorMap}
+        />
       </VisCanvas>
-      <ColorBar />
+      <ColorBar domain={domain} scaleType={scaleType} colorMap={colorMap} />
     </div>
   );
 }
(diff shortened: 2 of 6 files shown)


src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -26,20 +27,14 @@ export interface TextureDataState {
...
-      state.customDomain,
-      state.scaleType,
-      state.colorMap,
-    ],
+  const [scaleType, colorMap] = useHeatmapConfig(
+    (state) => [state.scaleType, state.colorMap],
     shallow
   );
 
-  const domain = customDomain || dataDomain;
```
loichuder, line 41: `domain` comes from the props now

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -17,3 +19,28 @@ export function generateCSSLinearGradient(
...
+  }
+
+  const supportedDomain = findDomain(values.filter((x) => x > 0));
+  if (supportedDomain && customDomain) {
+    // Clamp custom domain minimum to the first positive value
+    return [supportedDomain[0], customDomain[1]] as Domain;
+  }
+
+  return supportedDomain;
+}
```
loichuder, line 46: This is the main point of the MR: adding a function to compute once and for all components needing it, the domain to be used when mapping to colours.

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -1,29 +1,28 @@
...
-  );
+import { SCALE_FUNCTIONS, ScaleType, Domain } from '../shared/models';
 
+function ColorBar({
+  domain,
+  scaleType,
+}: {
+  domain: Domain | undefined;
+  scaleType: ScaleType;
+}): JSX.Element {
```
axelboc, line 16: Formatting can be fixed with:

- `props` + destructuring inside the function
- `Props` interface

src/h5web/visualizations/heatmap/ColorBar.tsx
```diff
@@ -1,29 +1,28 @@
...
-import { useHeatmapConfig } from './config';
 import styles from './HeatmapVis.module.css';
 import { generateCSSLinearGradient } from './utils';
-import { SCALE_FUNCTIONS } from '../shared/models';
-
-function ColorBar(): JSX.Element {
-  const [dataDomain, customDomain, scaleType] = useHeatmapConfig(
-    (state) => [state.dataDomain, state.customDomain, state.scaleType],
-    shallow
-  );
```
axelboc, line 16: I like the idea of using less of `useHeatmapConfig` so we can pass the config as a prop to `HeatmapVis`. That being said, this will trigger more re-renders of `HeatmapVis` itself, so we may have to add `React.memo()` here and there, like on `VisCanvas` to avoid re-rendering too many children when the user changes the color map or whatever.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -21,8 +21,20 @@ function HeatmapVis(props: Props): JSX.Element {
...
+    keepAspectRatio,
+    showGrid,
+  ] = useHeatmapConfig(
+    (state) => [
+      state.dataDomain,
+      state.customDomain,
+      state.scaleType,
+      state.keepAspectRatio,
+      state.showGrid,
+    ],
```
axelboc, line 37: Isn't this basically the whole config now? `const { ... } = useHeatmapConfig()` would be a lot cleaner, I reckon.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -26,20 +27,14 @@ export interface TextureDataState {
...
 ): TextureDataState {
-  const [dataDomain, customDomain, scaleType, colorMap] = useHeatmapConfig(
-    (state) => [
-      state.dataDomain,
-      state.customDomain,
-      state.scaleType,
-      state.colorMap,
-    ],
+  const [scaleType, colorMap] = useHeatmapConfig(
+    (state) => [state.scaleType, state.colorMap],
```
axelboc, line 34: Might as well pass everything through the props now, don't you think?

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -17,3 +19,28 @@ export function generateCSSLinearGradient(
...
+  if (!domain) {
+    return undefined;
+  }
+
+  // If the scale is not log or if the domain does not cross zero, the domain does not need to be adjusted
+  if (scaleType !== ScaleType.Log || domain[0] * domain[1] > 0) {
+    return domain;
+  }
+
+  const supportedDomain = findDomain(values.filter((x) => x > 0));
```
axelboc, line 39: This line filters positive values, but the condition above supports datasets with only negative values as well. The `filter` needs a ternary inside it that chooses which function to apply based on whether `domain[0] * domain[1]` is greater or lower than 0. Perhaps `d3-array` has something to simplify all this logic? I haven't looked.

axelboc (changes requested):
Nice and clean 👍  -- just one issue and a couple of minor suggestions.

src/h5web/visualizations/heatmap/ColorBar.tsx
axelboc: I've opened an issue to remember.

axelboc:
Oh and I agree with your decisions:
- to show white data points (I believe that's what is done in silx view as well, if I recall)
- to pass the domain through props and reduce the use of `useHeatmapConfig`, as I've mentioned in a comment.

src/h5web/visualizations/heatmap/ColorBar.tsx
loichuder: Oops, I wrote that for quick prototyping and let this slide :sweat:

src/h5web/visualizations/heatmap/ColorBar.tsx
loichuder: All right. I did not notice any performance drop but it is better to keep it in mind

src/h5web/visualizations/heatmap/HeatmapVis.tsx
loichuder: Well, there are still the setters inside the config, but we might as well cross the bridge

src/h5web/visualizations/heatmap/hooks.ts
loichuder: I think therefore I will do so :slightly_smiling_face:

src/h5web/visualizations/heatmap/utils.ts
loichuder: I see your point, but I think this necessitates some thought. Example: if you have a dataset with positive and negative values, you may argue that you want the supported domain to be computed from negative values. Then, how do you choose between the two possibilities ?

From my point of view, I expect the logScale to be used only for positive values, which is why I used this condition (and the same in `LineVis`). That being said, `d3-scale` also accepts a domain of negative values so I allow the log scale to take advantage of that for datasets containing only negatives values, with the condition above.

TL;DR: it raises interesting questions that should be addressed in another MR, ideally conjointly with `LineVis`.

loichuder:
> Oh and I fully agree with your decisions:
> 
>     * to show white data points (I believe that's what is done in silx view as well, if I recall)

`silx` uses `matplotlib` so the behaviour should be indeed the same.
> 
>     * to pass the domain through props and reduce the use of `useHeatmapConfig`, as I've mentioned in a comment.
With the removal of the `useInterpolator` hook, the `useHeatmapConfig` is only used in `HeatmapVis` now :+1:

src/h5web/visualizations/heatmap/utils.ts
axelboc: When you get to line 39, you're certain that the domain does not span both negative and positive values, it's always either one or the other. If you don't use `x < 0` when the domain is entirely negative, then you'll just see a white heatmap.

axelboc: > based on whether domain[0] * domain[1] is greater or lower than 0

This was wrong, sorry. You can just check the sign of the min.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: They can't change so they can't trigger re-renders.

src/h5web/visualizations/heatmap/utils.ts
loichuder: > When you get to line 39, you're certain that the domain does not span both negative and positive values, it's always either one or the other. 
Hm I don't think I get you there. On the contrary, the early return if `domain[0] * domain[1] > 0` (that is min and max have the same sign) ensures that line 39 is executed *if and only if* the domain spans negative and positive values, which is why the domain needs adjustment.

That being said, the discussion is fruitful as thanks to your comments, I discovered an edge case where the domain is `[-X, 0]` which will not trigger the early return and lead indeed to a white heatmap.

axelboc: Oooh right, sorry :sweat: thanks for the clarification - nice catch on the edge case :fishing_pole_and_fish:

axelboc (approved):

PR #195, opened 2020-09-14
Allow profiling visualisations

axelboc:
- Add env var to enable profiling in development (use of React's Profiler are automatically removed by React when building for production).
- Create component to wrap a tree with React's `Profiler` based on the new env var.
- Wrap the visualizations render call in `VisDisplay` with a profiler to profile every visualization.

image

### DIFF
--- a/.env
+++ b/.env
@@ -15,3 +15,6 @@ REACT_APP_HSDS_URL=
 REACT_APP_HSDS_USERNAME=
 REACT_APP_HSDS_PASSWORD=
 REACT_APP_HSDS_FILEPATH=
+
+# Rendering performance profiling (local development only)
+REACT_APP_PROFILING_ENABLED=false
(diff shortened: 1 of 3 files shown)


axelboc:
The `worst-case` duration shown in the console is the [`baseDuration`](https://reactjs.org/docs/profiler.html):

> **baseDuration: number** - Duration of the most recent render time for each individual component within the Profiler tree. This value estimates a worst-case cost of rendering (e.g. the initial mount or a tree with no memoization).

Not sure why sometimes it's less than the actual duration... It's a bit vague.

The goals are basically:
- to not have unnecessary `update` renders (by avoiding unnecessary state changes - e.g. currently, opening the colour map drop-down triggers a re-render);
- to reduce the `duration` of the initial `render` (by optimising computations, not blocking rendering while performing heavy computations, etc.);
- to drastically reduce the `duration` of subsequent `update` renders (notably with `useMemo` and `React.memo()`).

loichuder (approved):
Excellent stuff ! 

What are the unit of the durations ? That can't be seconds.

axelboc:
Neh, it's not specified in the docs, but I assume it's milliseconds.

PR #197, opened 2020-09-15
Fix custom domain resetting

axelboc:
The texture flickering issue is gone for good!

- Compute data domain in `HeatmapVis` instead of config (in `useMemo` to avoid unnecessary recomputations).
- Make `HeatmapVis` the source of truth for the custom domain whenever `values` prop changes (because for one render, the config may have a stale custom domain).
- Use effect hook to update data domain and reset custom domain in config when `values` prop changes.

Since the domain is no longer computed asynchronously, we can also:

- remove the workaround that used to avoid consecutive texture recomputations;
- render an error message and return early when the domain can be computed from the current `values`.

### DIFF
--- a/src/h5web/visualizations/heatmap/HeatmapVis.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.tsx
@@ -1,6 +1,7 @@
-import React, { useEffect } from 'react';
+import React, { useMemo, useEffect } from 'react';
 import { format } from 'd3-format';
 import type ndarray from 'ndarray';
+import { usePrevious } from 'react-use';
 import styles from './HeatmapVis.module.css';
 import ColorBar from './ColorBar';
 import Mesh from './Mesh';
@@ -8,7 +9,9 @@ import TooltipMesh from '../shared/TooltipMesh';
 import { useHeatmapConfig } from './config';
 import PanZoomMesh from '../shared/PanZoomMesh';
 import VisCanvas from '../shared/VisCanvas';
-import { getDims, useMemoColorScaleDomain } from './utils';
+import { getDims } from './utils';
+import { findDomain } from '../shared/utils';
+import { useMemoColorScaleDomain } from './hooks';
 
 interface Props {
   dataArray: ndarray<number>;
@@ -21,21 +24,27 @@ function HeatmapVis(props: Props): JSX.Element {
   const values = dataArray.data as number[];
 
   const {
-    dataDomain,
-    customDomain,
+    requestedDomain,
     scaleType,
     colorMap,
     keepAspectRatio,
     showGrid,
-    initDataDomain,
+    resetDomains,
   } = useHeatmapConfig();
 
-  // width / height <=> cols / rows
-  const aspectRatio = keepAspectRatio ? cols / rows : undefined;
+  const dataDomain = useMemo(() => findDomain(values), [values]);
+  const prevDataDomain = usePrevious(dataDomain);
+
+  // If `dataDomain` just changed, use `undefined` custom domain for this render until config's `requestedDomain` is updated through `useEffect` below
+  const customDomain =
+    dataDomain !== prevDataDomain ? undefined : requestedDomain;
 
   useEffect(() => {
-    initDataDomain(values);
-  }, [initDataDomain, values]);
+    // When `dataDomain` changes:
+    // - set data domain in config so toolbar can render domain slider with correct bounds;
+    // - reset requested domain in config so toolbar can reset value of domain slider, and so Heatmap can use in its next render.
+    resetDomains(dataDomain);
+  }, [dataDomain, resetDomains]);
 
   const domain = useMemoColorScaleDomain(
     scaleType,
@@ -43,6 +52,9 @@ function HeatmapVis(props: Props): JSX.Element {
     dataDomain,
     customDomain
   );
+
+  // width / height <=> cols / rows
+  const aspectRatio = keepAspectRatio ? cols / rows : undefined;
 
   return (
     <div className={styles.root}>
--- a/src/h5web/visualizations/heatmap/HeatmapVis.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.tsx
@@ -56,6 +56,10 @@ function HeatmapVis(props: Props): JSX.Element {
   // width / height <=> cols / rows
   const aspectRatio = keepAspectRatio ? cols / rows : undefined;
 
+  if (!domain) {
+    return <p className={styles.domainError}>Unable to compute domain</p>;
+  }
+
   return (
     <div className={styles.root}>
       <VisCanvas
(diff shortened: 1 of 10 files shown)


src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -47,5 +46,3 @@ function getColorScaleDomain(
 
   return supportedDomain;
 }
-
-export const useMemoColorScaleDomain = createMemo(getColorScaleDomain);
```
axelboc, line 51: I moved this to the `hooks` file.

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -57,7 +46,7 @@ export function useTextureData(
         ),
       });
     })();
-  }, deps); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [colorMap, domain, scaleType, proxy, mergeState, values]);
```
axelboc, line 49: No more workaround!

src/h5web/visualizations/heatmap/config.ts
```diff
@@ -1,11 +1,10 @@
 import type { ColorMap } from './models';
 import { Domain, ScaleType } from '../shared/models';
 import { StorageConfig, createPersistableState } from '../../storage-utils';
-import { findDomain } from '../shared/utils';
 
 interface HeatmapConfig {
   dataDomain: Domain | undefined;
-  initDataDomain: (values: number[]) => void;
+  resetDomains: (dataDomain: Domain | undefined) => void;
```
axelboc, line 7: Wasn't sure what to call it. I hope this works.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -21,29 +24,42 @@ function HeatmapVis(props: Props): JSX.Element {
...
     colorMap,
     keepAspectRatio,
     showGrid,
-    initDataDomain,
+    resetDomains,
   } = useHeatmapConfig();
 
-  // width / height <=> cols / rows
-  const aspectRatio = keepAspectRatio ? cols / rows : undefined;
+  const dataDomain = useMemo(() => findDomain(values), [values]);
```
axelboc, line 35: The data domain is now computed here instead of in the config. This means that it is never stale during the render, even when slicing through dimensions.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -21,29 +24,42 @@ function HeatmapVis(props: Props): JSX.Element {
...
+  const prevDataDomain = usePrevious(dataDomain);
+
+  // If `dataDomain` just changed, use `undefined` custom domain for this render until config is updated through `useEffect` below
+  const freshCustomDomain =
+    dataDomain !== prevDataDomain ? undefined : customDomain;
```
axelboc, line 40: The second part of the solution: we disregard the custom domain from the config if it is potentially stale.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -21,29 +24,42 @@ function HeatmapVis(props: Props): JSX.Element {
...
 
   useEffect(() => {
-    initDataDomain(values);
-  }, [initDataDomain, values]);
+    // When `dataDomain` changes:
+    // - set data domain in config so toolbar can render domain slider with correct bounds;
+    // - reset custom domain in config so toolbar can reset value of domain slider, and so Heatmap can re-render with correct, `undefined` custom domain.
+    resetDomains(dataDomain);
+  }, [dataDomain, resetDomains]);
```
axelboc, line 47: And the third part: we update the config, which update the domain slider and the toolbar and re-renders the Heatmap vis. During this second render, `dataDomain !== prevDataDomain` is `false`, so we use the `customDomain` from the config... but it's fine because `customDomain` is now `undefined` in the config and therefore no longer stale.

src/h5web/visualizations/heatmap/utils.ts
loichuder: Ah, good call :+1:

src/h5web/visualizations/heatmap/config.ts
loichuder: `initDomains` ? `resetDomains` is fine to me tho.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
loichuder: I feel we can find a better name than `freshCustomDomain`... But I cannot :sweat_smile: 

Perhaps it is `customDomain` that should be renamed to `sliderDomain` or `toolbarDomain` :shrug: ?

src/h5web/visualizations/heatmap/hooks.ts
```diff
@@ -67,3 +56,5 @@ export function useTextureData(
 
   return state;
 }
+
+export const useColorScaleDomain = createMemo(getColorScaleDomain);
```
loichuder, line 60: You did not want to keep the `Memo` in the hook name ? I found it handy to know that it was in fact a disguised `useMemo` hook

loichuder (approved):
:+1:

src/h5web/visualizations/heatmap/config.ts
axelboc: Yeah, I thought `initDomains` didn't convey the fact that the domains could change (it's not just called on page load or mount or whatever).

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: I had `effectiveCustomDomain` and `safeCustomDomain` at some point...

src/h5web/visualizations/heatmap/hooks.ts
axelboc: Yeah, on second thought, you're right, it's clearer. It just felt weird at first :smile:

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: Not a fan of chaning the name in the config, since it's only during one render that we skip using it.

loichuder: Thinking about it, I stand my point: I think it would make sense to rename `customDomain` to `sliderDomain` or something similar and `freshCustomDomain` to `customDomain`.

Then, the ternary would convey the meaning: 
*If the data domain is updated, set the custom domain to `undefined` else use the customDomain from the store*

axelboc: I could rename it in the destructuring expression, but then if we pass the config as prop... Not sure... :D

axelboc: Yeah okay (sorry GitHub had not updated)

PR #198, opened 2020-09-15
Optimize ndarray uses in MappedVis

### LINKED ISSUE #170: Optimise uses of ndarray

loichuder:
A revised version of #178 to fix #170.

> Split useMemo in two in VisProvider (one for ndarray creation, one for slicing).

The `useMemo` now lives in `MappedVis` that replaced `VisProvider`. But it was done.
> Simplify unpack(mappedView).flat() by unpacking a flat array directly.

Done with `ndarray-ops` (https://github.com/scijs/ndarray/issues/32#issuecomment-253972844)

Other tasks were not considered as they depend on `cwise` which does not seem actively maintained [as I pointed out](https://github.com/silx-kit/h5web/issues/170#issuecomment-643970544).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16471,6 +16471,14 @@
         "is-buffer": "^1.0.2"
       }
     },
+    "ndarray-ops": {
+      "version": "1.2.2",
+      "resolved": "https://registry.npmjs.org/ndarray-ops/-/ndarray-ops-1.2.2.tgz",
+      "integrity": "sha1-WeiNLDKn7ryxvGkPrhQVeVV6YU4=",
+      "requires": {
+        "cwise-compiler": "^1.0.0"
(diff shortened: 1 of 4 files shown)


axelboc (approved):
Very well found indeed 👏

PR #199, opened 2020-09-16
Adding story for the ColorBar

loichuder:
image

To get used to Storybook, I had some fun to change the domain args into two sliders. But it may obfuscate the true `domain` arg so I can revert to a regular `domain` control.


axelboc (approved):
Nice ! ✨

PR #200, opened 2020-09-16
Added more ColorBar stories

loichuder:
image


axelboc (approved):
Awesome! Gosh, I really like what they've done with Storybook.

PR #202, opened 2020-09-16
Remove AxisInfo from getTickProps

loichuder:
<s>
## Motivation
Getting back to the integer ticks problems, I felt that the current implementation was sometimes hard to follow:
- `AxisInfo` and `AxisConfig` are quite similar but used for different purposes
- The distinction between `IndexAxisConfig` and `DataAxisConfig` leads to additional code (different types, a type guard) as both are again very similar. Also, it is not clear from first approach that `IndexAxisConfig` forces the axis to contain only integers.

Reading the code, I found myself sometimes misdirected by all these different but similar types.

## Refactor
To address these points, I did the following:
- I removed `AxisInfo`. `AxisConfig` is used instead with a new boolean member `onlyIntegers`.
- The boolean `onlyIntegers` makes the distinction between `IndexAxisConfig` and `DataAxisConfig` superfluous. Both are merged in `AxisConfig` by changing `dataDomain` and `indexDomain` to `domain`.
- Rather than passing the whole config to `getTicksProp`, only `onlyIntegers` is passed as arg.

## Aftermath
- I hope it makes the code clearer as we don't have anymore to compare `AxisInfo` and `AxisConfig` to show what does what. The same object is used from start to finish.
- The introduction of the boolean `onlyIntegers` should also render the config more readable and easily usable.
- See other comments below
</s>

In the end, the refactor didn't feel right:
 - To keep the API safe (that is forcing linear scale for index/integer axis), I ended up needing to mutate `AxisConfig`, which conceptually, was the purpose of `AxisInfo`... 
- Also, for the same reason, `IndexAxisConfig` and `DataAxisConfig` typings could not be easily merged as `scaleType` serves no purpose when `isIndexAxis` is `true`.

 In the end, I just changed the arg in `getTickProps` :smile:

### DIFF
--- a/src/h5web/visualizations/shared/AxisSystem.tsx
+++ b/src/h5web/visualizations/shared/AxisSystem.tsx
@@ -70,8 +70,16 @@ function AxisSystem(props: Props): JSX.Element {
   yTicksScale.domain(visibleDomains[1]);
   yTicksScale.range([height, 0]);
 
-  const xTicksProp = getTicksProp(abscissaInfo, visibleDomains[0], width);
-  const yTicksProp = getTicksProp(ordinateInfo, visibleDomains[1], height);
+  const xTicksProp = getTicksProp(
+    visibleDomains[0],
+    width,
+    abscissaInfo.isIndexAxis
+  );
+  const yTicksProp = getTicksProp(
+    visibleDomains[1],
+    height,
+    ordinateInfo.isIndexAxis
+  );
 
   return (
     <HTML
--- a/src/h5web/visualizations/shared/utils.test.ts
+++ b/src/h5web/visualizations/shared/utils.test.ts
@@ -1,6 +1,5 @@
 import { tickStep } from 'd3-array';
 import { computeVisSize, findDomain, getTicksProp } from './utils';
-import type { AxisInfo } from './models';
 
 describe('Shared visualization utilities', () => {
   describe('computeVisSize', () => {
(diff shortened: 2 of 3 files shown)


src/h5web/visualizations/shared/models.ts
```diff
@@ -23,35 +23,20 @@ export type Size = { width: number; height: number };
 
 export type Domain = [number, number];
 
-export interface IndexAxisConfig {
-  indexDomain: Domain;
-  showGrid?: boolean;
-  scaleType?: never; // invalid
-}
```
loichuder, line 30: One could argue that it is now possible to have integer ticks with a scale that is not linear (and therefore not representative of an IndexAxis).

I think it is fine ! It adds more flexiblity and if the caller wants an index axis, it is its responsibility to set it properly.

src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -62,16 +68,28 @@ function AxisSystem(props: Props): JSX.Element {
...
+  const xTicksScale = SCALE_FUNCTIONS[
+    abscissaConfig.scaleType || ScaleType.Linear
+  ]();
   xTicksScale.domain(visibleDomains[0]);
   xTicksScale.range([0, width]);
 
-  const yTicksScale = ordinateInfo.scaleFn();
+  const yTicksScale = SCALE_FUNCTIONS[
+    ordinateConfig.scaleType || ScaleType.Linear
+  ]();
```
loichuder, line 79: It is the only strong drawback that I encountered: as `scaleType` can be undefined in `AxisConfig`, I have to supply a fallback it here.

In other places (`Tooltip`, `LineVis`...), the fallback is handled in `getAxisScale`.

loichuder: `scaleType` is now always in `AxisConfig`

src/h5web/visualizations/shared/models.ts
loichuder: Actually, this conflits with the heatmap: the heatmap cannot accept X/Y axis that are not linear. So I will review this.

src/h5web/visualizations/shared/AxisSystemProvider.tsx
```diff
@@ -1,51 +1,34 @@
...
 
-  const abscissaInfo = getAxisInfo(abscissaConfig);
-  const ordinateInfo = getAxisInfo(ordinateConfig);
-
   return (
-    <AxisSystemContext.Provider value={{ abscissaInfo, ordinateInfo }}>
+    <AxisSystemContext.Provider
+      value={{
+        abscissaConfig: { ...abscissaConfig, scaleType: ScaleType.Linear },
+        ordinateConfig: { ...ordinateConfig, scaleType: ScaleType.Linear },
```
axelboc, line 29: Hmmm, you're always overriding it like this. You need to destructure the config after the defaults. Perhaps add a `DEFAULT_AXIS_CONFIG` at the top of the file as well, with `scaleType` and perhaps also `isIndexAxis: false` for clarity?

src/h5web/visualizations/shared/models.ts
```diff
@@ -23,19 +23,24 @@ export type Size = { width: number; height: number };
 
 export type Domain = [number, number];
 
-export interface IndexAxisConfig {
-  indexDomain: Domain;
+export type IndexAxisConfig = {
+  domain: Domain;
   showGrid?: boolean;
-  scaleType?: never; // invalid
```
axelboc, line 29: you want to keep this here (TS is too permissive by default)

src/h5web/visualizations/shared/models.ts
```diff
@@ -23,19 +23,24 @@ export type Size = { width: number; height: number };
...
   scaleType?: ScaleType;
-}
+};
 
-export type AxisConfig = IndexAxisConfig | DataAxisConfig;
+export type AxisConfig = {
+  domain: Domain;
+  showGrid?: boolean;
+  scaleType: ScaleType;
+  isIndexAxis?: boolean;
```
axelboc, line 42: You could make this required and add `isIndexAxis?: false` in `DataAxisConfig.

axelboc: It may also make it easier to add `scaleFn` in this interface, and compute it in `AxisSystemProvider`, before injecting the final config.

axelboc (changes requested):

axelboc (approved):
😄 better than nothing

PR #203, opened 2020-09-16
First round of dependency upgrades

### LINKED ISSUE #194: Major release of D3
We should upgrade our D3 libraries to the latest. I'm hoping it will fix #183 and allow us to remove some code introduced in #152 to fix #146. See [changelog for D3 v6](https://github.com/d3/d3/blob/master/CHANGES.md).

axelboc:
Fix #194


src/h5web/visualizations/shared/PanZoomMesh.tsx
```diff
@@ -32,7 +32,7 @@ function PanZoomMesh(): JSX.Element {
 
   const onPointerDown = useCallback(
     (evt: PointerEvent) => {
-      const { currentTarget, pointerId } = evt as React.PointerEvent;
+      const { currentTarget, pointerId } = evt;
```
axelboc, line 35: Some minor gains here and there... 🙂

src/setupTests.js
```diff
@@ -9,3 +9,12 @@ import '@testing-library/jest-dom';
...
 });
+
+// Silence web workers warning from `drei`
+const { log } = console;
+// eslint-disable-next-line no-console
+console.log = (...args) => {
+  if (args.length === 0 || !/web workers not allowed/.test(args[0])) {
+    log(...args);
+  }
+};
```
loichuder, line 20: What does the warning says ?

loichuder (approved):

src/setupTests.js
axelboc: I managed to find a better solution: I just mock the problematic package (`troika`), which doesn't get used anyway.

src/h5web/App.tsx
```diff
@@ -45,11 +45,7 @@ function App(): JSX.Element {
           style={{ display: isExplorerOpen ? undefined : 'none' }}
         />
 
-        <ReflexElement
-          className={styles.mainArea}
-          flex={isExplorerOpen ? 0.75 : 1}
-          minSize={500}
-        >
+        <ReflexElement className={styles.mainArea} flex={75} minSize={500}>
```
loichuder, line 48: We had a ternary before `flex={isExplorerOpen ? 0.75 : 1}` that is now missing. Not sure if that's on purpose.

loichuder (approved):

src/h5web/App.tsx
axelboc: Yeah it works without it now. Since it's the only flex item when the explorer is hidden, it grows to fit the space. It wasn't working before for some reason I never understood.

PR #206, opened 2020-09-17
ColorBar can now be horizontal

loichuder:
image

Perhaps a bit early, but I foresee this will be needed for future integrations.

### DIFF
--- a/src/h5web/visualizations/heatmap/ColorBar.module.css
+++ b/src/h5web/visualizations/heatmap/ColorBar.module.css
@@ -1,0 +1,26 @@
+.colorBar {
+  display: flex;
+  padding: 1em;
+}
+
+.colorBar:not([data-horizontal]) {
+  width: 6em;
+  min-height: 0;
+}
+
+.colorBar[data-horizontal] {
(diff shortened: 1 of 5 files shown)


axelboc (approved):
Hehe nice!

src/h5web/visualizations/heatmap/ColorBar.module.css
```diff
@@ -0,0 +1,26 @@
...
+  display: flex;
+  padding: 1em;
+}
+
+.colorBar[data-horizontal='false'] {
+  width: 6em;
+  min-height: 0;
+}
+
+.colorBar[data-horizontal='true'] {
```
axelboc, line 11: You can use a ternary with `undefined` to avoid having to specify `true`/`false` here (and instead use `[data-horizontal]` and `:not([data-horizontal])`. I find it a little nicer than to check a boolean string.

loichuder: Nice, I wanted to do so but didn't know the `:not` syntax :+1:

PR #208, opened 2020-09-17
Turn MappedVis into a hook

axelboc:
Next step:

1. Pass vis configs as props.

### DIFF
--- a/src/h5web/visualizations/MappedMatrixVis.tsx
+++ b/src/h5web/visualizations/MappedMatrixVis.tsx
@@ -1,9 +1,9 @@
 import React, { ReactElement } from 'react';
-import MappedVis from './shared/MappedVis';
 import type { HDF5Dataset, HDF5Value } from '../providers/models';
 import type { DimensionMapping } from '../dataset-visualizer/models';
 import MatrixVis from './matrix/MatrixVis';
 import { assertArray } from './shared/utils';
+import { useMappedArray } from './shared/hooks';
 
 interface Props {
   value: HDF5Value;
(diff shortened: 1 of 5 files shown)


src/react-app-env.d.ts
```diff
@@ -10,7 +10,5 @@ declare module 'ndarray-unpack' {
 }
 
 declare module 'ndarray-ops' {
-  function assign(a: ndarray<T>, b: ndarray<T>);
-
-  export = { assign };
+  export function assign(a: ndarray<T>, b: ndarray<T>);
```
loichuder, line 13: It does not change the export from `{ assign }` to `assign` ?

src/h5web/visualizations/MappedMatrixVis.tsx
```diff
@@ -15,14 +15,8 @@ function MappedMatrixVis(props: Props): ReactElement {
...
-  return (
-    <MappedVis
-      component={MatrixVis}
-      dataset={dataset}
-      value={value}
-      mapperState={mapperState}
-    />
-  );
+  const dataArray = useMappedArray(dataset, value, mapperState);
+  return <MatrixVis dataArray={dataArray} />;
```
loichuder, line 19: You could even inline the `dataArray`. As you wish.

loichuder (approved):

src/react-app-env.d.ts
axelboc: Before you were declaring a default export equal to an object with an `assign` property -- now, `assign` is a named export. This allows the  use of `import { assign }` instead of `import ops` and then `ops.assign`.

src/h5web/visualizations/MappedMatrixVis.tsx
axelboc: Meh, I quite like it like this, especially with what's coming next 😉

src/react-app-env.d.ts
loichuder: All right, understood

PR #209, opened 2020-09-17
Pass vis configs as props

axelboc:
Much better!

### DIFF
--- a/src/h5web/visualizations/MappedHeatmapVis.tsx
+++ b/src/h5web/visualizations/MappedHeatmapVis.tsx
@@ -1,9 +1,11 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useEffect } from 'react';
+import { usePrevious } from 'react-use';
 import type { HDF5Dataset, HDF5Value } from '../providers/models';
 import type { DimensionMapping } from '../dataset-visualizer/models';
 import HeatmapVis from './heatmap/HeatmapVis';
 import { assertArray } from './shared/utils';
-import { useMappedArray } from './shared/hooks';
+import { useMappedArray, useDataDomain } from './shared/hooks';
+import { useHeatmapConfig } from './heatmap/config';
 
 interface Props {
   value: HDF5Value;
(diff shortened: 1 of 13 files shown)


src/h5web/visualizations/MappedHeatmapVis.tsx
```diff
@@ -16,7 +18,19 @@ function MappedHeatmapVis(props: Props): ReactElement {
   assertArray<number>(value);
 
   const dataArray = useMappedArray(dataset, value, mapperState);
-  return <HeatmapVis dataArray={dataArray} />;
+
+  const { resetDomains, ...config } = useHeatmapConfig(
+    ({ dataDomain, ...rest }) => rest, // eslint-disable-line @typescript-eslint/no-unused-vars
+    shallow
+  );
```
loichuder, line 25: I am afraid I don't understand this piece of code...

axelboc: Sorry, I was about to write a comment. I'm excluding `dataDomain` from the state selection to save a render. (Technically, it saves a render only if the user hadn't selected a custom domain.)

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -6,31 +6,32 @@ import styles from './HeatmapVis.module.css';
...
 import { getDims } from './utils';
 import { findDomain } from '../shared/utils';
 import { useMemoColorScaleDomain } from './hooks';
+import type { HeatmapConfig } from './config';
+import type { Domain } from '../shared/models';
 
 interface Props {
   dataArray: ndarray<number>;
+  config: HeatmapConfig;
+  onDataDomainChange: (dataDomain: Domain | undefined) => void;
```
axelboc, line 20: I don't like this, because it makes users responsible for resetting the `requestedDomain` in `onDataDomainChange`, which they could do incorrectly. That and the API is not self-documenting.

loichuder (approved):
Nice !

axelboc (commented):
For what it's worth, posting my original comments. Better later than never! :D I'll respond more precisely to your review in a  bit.

src/h5web/visualizations/MappedLineVis.tsx
```diff
@@ -15,8 +16,20 @@ function MappedLineVis(props: Props): ReactElement {
...
+  const { scaleType, curveType, showGrid } = useLineConfig();
+
   const dataArray = useMappedArray(dataset, value, mapperState);
-  return <LineVis dataArray={dataArray} />;
+  const dataDomain = useDataDomain(dataArray.data as number[]);
+
+  return (
+    <LineVis
+      dataArray={dataArray}
+      domain={dataDomain}
```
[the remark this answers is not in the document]
loichuder: `domain` is fine to me. It doesn't have to be the domain of the data.

PR #212, opened 2020-09-22
Allow hiding loader on Heatmap and add story for live data

axelboc:


src/stories/HeatmapVis.stories.tsx
```diff
@@ -75,6 +76,28 @@ NoGrid.args = {
...
+      .slice(0)
+      .sort(() => 0.5 - Math.random());
+
+    setShuffledArray(
+      ndarray<number>(shuffledValues, dataset.shape.dims).transpose(1, 0)
+    );
+  }, 5000);
+
+  return <HeatmapVis {...args} dataArray={shuffledArray} />;
+};
```
axelboc, line 93: Every 5000ms, I shuffle the array.

src/stories/HeatmapVis.stories.tsx
```diff
@@ -75,6 +76,28 @@ NoGrid.args = {
   showGrid: false,
 };
 
+export const LiveData: Story<HeatmapVisProps> = (args): ReactElement => {
```
loichuder, line 79: It would be nice to mention the loader in the story name. `LiveDataWithoutLoader` ?

loichuder (approved):
:+1:

src/stories/HeatmapVis.stories.tsx
axelboc: nice one 👍

PR #213, opened 2020-09-22
Fix log axis showing too many ticks

### LINKED ISSUE #146: Too many ticks on index axis

### LINKED ISSUE #183: [Line] Too many ticks with log scale
image

axelboc:
Fix #183 

image

### DIFF
--- a/src/h5web/visualizations/shared/AxisSystem.tsx
+++ b/src/h5web/visualizations/shared/AxisSystem.tsx
@@ -1,19 +1,17 @@
 import React, { useState, useContext } from 'react';
 import { useThree, useFrame } from 'react-three-fiber';
 import { AxisLeft, AxisBottom, TickRendererProps } from '@vx/axis';
-import { format } from 'd3-format';
 import { GridColumns, GridRows } from '@vx/grid';
 import { HTML } from 'drei';
 import { useUpdateEffect } from 'react-use';
 import styles from './AxisSystem.module.css';
 import type { AxisOffsets, Domain } from './models';
-import { getTicksProp, getAxisScale } from './utils';
+import { adaptedNumTicks, getTicksProp, getAxisScale } from './utils';
 import { AxisSystemContext } from './AxisSystemProvider';
 
 const AXIS_PROPS = {
   tickStroke: 'grey',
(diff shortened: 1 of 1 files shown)


loichuder (changes requested):
I am sorry to raise an issue while it seemed so elegantly handled: if you display only labels for exact powers of 10, it means you can have **only** ticks without labels when zooming (see below).

I would prefer that labels continue to appear in this case as in the linear case :slightly_smiling_face: 

image

axelboc:
No worries at all, very good point. Perhaps I could take the minimum of the number of ticks requested and the number of ticks actually generated by D3?

loichuder:
Actually, the axe of the colorbar has the behaviour you want to implement, without any tuning from us. 

Perhaps we are missing an `Axis` option that would also to have the same behaviour for the `AxisSystem` without any custom implementation ?

image

axelboc:
Good call! 

```js
tickFormat={yTicksScale.tickFormat(adaptedNumTicks(height), '.3')}
```

vasole:
You should not try to reinvent the wheel. Just take a look at how it is done in silx for the OpenGL backend

For reference, you can take a look at the Graphics Gems book chapter Nice Numbers for Graph Labels from Paul S. Heckbert.

axelboc:
Still losing ticks at some zoom levels with `scale.tickFormat(...)`...

image

axelboc:
Closing as the solution is not satisfactory. I've updated the issue to document the attempts: https://github.com/silx-kit/h5web/issues/183#issuecomment-696727626

PR #214, opened 2020-09-23
Renamed drei's HTML into Html

loichuder:
Fixes the following warning in the console as `<HTML />` component was renamed to `<Html />` (https://github.com/pmndrs/drei#html-):
image


axelboc (approved):
🎣

PR #215, opened 2020-09-25
Set up NPM packages

axelboc:
@loichuder, this follows on your work to export the app for use in Jupyter Lab.

- I set up two packages for publishing: `@h5web/lib` and `@h5web/app` -- see README for more info.
- I add a homepage to the storybook documentation site (i.e. the documentation site for package `@h5web/lib`).
- I've already published the two packages to NPM: https://www.npmjs.com/org/h5web


src/stories/Home.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+      npm install --save-dev @types/ndarray
+      ```
+
+   1. Pass an `ndarray` and domain to the visualization:
+
+      ```tsx
+      import ndarray from 'ndarray';
+      import { HeatmapVis, findDomain } from '@h5web/lib';
+
+      // Source nD array
```
loichuder, line 48:
```suggestion
      // Source 2D array
```

`HeatmapVis` only accepts 2D arrays

src/stories/Home.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+      const values = [
+        [0, 1, 2],
+        [3, 4, 5],
+      ];
+
+      // Flatten array
+      const flatArray: number[] = values.flat(Infinity);
+
+      // Convert to ndarray and find domain
+      const dataArray = ndarray<number>(values, [2, 3]);
```
loichuder, line 58: Should be `flatArray` here instead of `values`.

src/stories/Home.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+      import { HeatmapVis, findDomain } from '@h5web/lib';
+
+      // Source nD array
+      const values = [
+        [0, 1, 2],
+        [3, 4, 5],
+      ];
+
+      // Flatten array
+      const flatArray: number[] = values.flat(Infinity);
```
loichuder, line 55: `flatValues` ? To avoid confusion with the `dataArray` that is an `ndarray`

src/stories/Home.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+      ```
+
+1. Configure the visualization:
+
+   ```tsx
+   <HeatmapVis
+     dataArray={dataArray}
+     domain={domain}
+     colorMap="Inferno"
+     showLoader={false}
```
loichuder, line 73: We might as well put every possible argument: `scaleType`, `showGrid`, `keepAspectRatio`

src/stories/Home.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+const values: number[] = dataset.value.flat(Infinity);
+const dataArray = ndarray<number>(values, dataset.shape.dims);
+```
+
+## Utilities
+
+The library exposes a number of utility functions, which are documented below, as well as all the enums and types used by these utility functions.
+If you use only the top-level visualization components, the only utilities you need are `findDomain` and/or `useDataDomain`.
+
+- **`findDomain(data: number[]): Domain | undefined`** - Find the min and max values contained in a flat array of numbers. Uses [`extent`](https://github.com/d3/d3-array#extent) from `d3-array` under the hood.
```
loichuder, line 95:
```suggestion
- **`findDomain(data: number[]): Domain | undefined`** - Find the min and max values contained in a flat array of numbers. Uses [`extent`](https://github.com/d3/d3-array#extent) from `d3-array`.
```

packages/package-lib.json
```diff
@@ -0,0 +1,20 @@
...
+    "url": "https://github.com/silx-kit/h5web"
+  },
+  "peerDependencies": {
+    "react": ">=16",
+    "react-dom": ">=16"
+  },
+  "optionalDependencies": {
+    "ndarray": "^1.x"
+  }
+}
```
loichuder, line 20: So this will be copied by webpack as `package.json` for the lib ? And same for `package-app.json` for the app ?

loichuder (changes requested):

packages/package-lib.json
axelboc: Indeed

src/stories/Home.stories.mdx
axelboc: Hmm, the idea here is just to give an example, not to document the Heatmap, which is done by the Heatmap's stories ("Doc" tab).

src/stories/Home.stories.mdx
axelboc: 🎣

src/stories/Home.stories.mdx
loichuder: Yes I get it but I think it is better to give a complete example from the get-go. It doesn't add much lines and will save someone a few seconds :slightly_smiling_face:

src/h5web/visualizations/heatmap/HeatmapVis.module.css
```diff
@@ -29,8 +29,3 @@
     background-position-x: 150%;
   }
 }
-
-.domainError {
-  padding: 1em;
-  color: var(--danger);
-}
```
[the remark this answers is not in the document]
loichuder: :fishing_pole_and_fish:

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -8,7 +8,7 @@ import TooltipMesh from '../shared/TooltipMesh';
 import PanZoomMesh from '../shared/PanZoomMesh';
 import VisCanvas from '../shared/VisCanvas';
 import { getDims } from './utils';
-import { useSupportedDomain } from './hooks';
+import { useSupportedDomain } from '../shared/hooks';
```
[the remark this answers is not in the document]
loichuder: :fishing_pole_and_fish:

src/packages/lib.ts
```diff
@@ -0,0 +1,48 @@
+// Visualizations
+export { default as ScalarVis } from '../h5web/visualizations/ScalarVis';
```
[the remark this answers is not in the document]
loichuder: Makes sense :+1:

src/stories/AxisSystem.stories.tsx
```diff
@@ -80,7 +80,7 @@ InheritedStyles.decorators = [
 ];
 
 export default {
-  title: 'Canvas/AxisSystem',
+  title: 'Building Blocks/AxisSystem',
```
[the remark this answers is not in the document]
loichuder: Good call :+1:

src/h5web/visualizations/heatmap/models.ts
```diff
@@ -8,8 +8,3 @@ export type Dims = {
 export type D3Interpolator = (t: number) => string;
 
 export type ColorMap = keyof typeof INTERPOLATORS;
-
-export type ColorMapOption = {
-  label: ColorMap;
-  value: ColorMap;
-};
```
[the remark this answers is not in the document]
loichuder: :fish:

.storybook/preview.js
```diff
@@ -1,3 +1,6 @@
 export const parameters = {
   actions: { argTypesRegex: '^on[A-Z].*' },
+  options: {
+    storySort: { order: ['Home', 'Visualizations'] },
```
loichuder, line 4: Is it missing `Building blocks` ?

packages/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "target": "es5",
```
loichuder, line 4: I didn't check but it might be unnecessary to change the `target` from `es6` to `es5`

src/stories/Home.stories.mdx
axelboc: I don't know, I don't see the relevance of passing the props you mention, since they have sensible defaults -- for instance, passing `keepAspectRatio` would be redundant, and `keepAspectRatio={false}` would not make for a good first example. Moreover, if users have TypeScript, they'll get autocompletion on props, anyway.

axelboc: Maybe just `scaleType` to show how to use the `ScaleType` enum?

.storybook/preview.js
axelboc: Stuff that aren't listed just get added to the end.

loichuder: Great :+1:

packages/tsconfig.json
axelboc: Very true!

axelboc: We can even go up to ES2017 I reckon https://kangax.github.io/compat-table/es2016plus/

src/stories/Home.stories.mdx
loichuder: I disagree: I always find frustrating to have an example with partial arguments as I have to go check in the full doc/signature. My point of view is either, give me only the required arguments or give me all of them. 

Sure, autocompletion helps but I can imagine some situations where it will not work.

Your call but at least adding `scaleType` would be a nice touch :+1:

packages/tsconfig.json
loichuder: Then, it should be changed in the original `tsconfig` ?

loichuder (approved):

src/stories/Home.stories.mdx
axelboc: I'll add `scaleType`, then 👍

axelboc:
Just realised we weren't including any polyfills. I've added them.

PR #217, opened 2020-09-28
Fix vis not reducing in height on resize

axelboc:


loichuder (changes requested):
I tried doing something similar when I implemented the horizontal colorbar. 

But fixing the display by adding a decorator only to the `Horizontal` story means it does not work when switching from vertical to horizontal in other ColorBar stories :slightly_frowning_face:

.storybook/decorators/FillHeight.tsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
 import type { Story } from '@storybook/react';
 
-function FillHeight(VisCanvasStory: Story) {
+function FillHeight(WrappedStory: Story) {
```
loichuder, line 4: `aStory` ?

loichuder (approved):

.storybook/decorators/FillHeight.tsx
axelboc: Needs to have a capital letter because it's a component.

loichuder: `MyStory` ? It is just that I don't like `WrappedStory` (wrapped by whom ? by what ?)

axelboc: It's what they use for the HOC pattern: https://reactjs.org/docs/higher-order-components.html, but I hesitated with `MyStory`, so I'm fine with that too 👍

PR #220, opened 2020-09-28
Implements Auto-scaling toggle

### LINKED ISSUE #182: [Line] Auto-scaling toggle
When slicing through a dataset on the Line visualization, provide the option to use the domain of the whole dataset rather than that of the current slice - that is, to turn off auto-scaling. Aut-scaling should be off by default.

loichuder:
Fix #182 

It bothers me a bit though: this toggle serves no purpose when the array is not mapped and I did not find any easy way to disable it in this case. 

Also I find that the label `Auto scale` does not convey the true purpose of this button.

### DIFF
--- a/src/h5web/visualizations/MappedHeatmapVis.tsx
+++ b/src/h5web/visualizations/MappedHeatmapVis.tsx
@@ -4,7 +4,7 @@ import type { HDF5Dataset, HDF5Value } from '../providers/models';
 import type { DimensionMapping } from '../dataset-visualizer/models';
 import HeatmapVis from './heatmap/HeatmapVis';
 import { assertArray } from './shared/utils';
-import { useMappedArray, useDataDomain } from './shared/hooks';
+import { useMappedArray, useDataDomain, useBaseArray } from './shared/hooks';
 import { useHeatmapConfig } from './heatmap/config';
 
 interface Props {
@@ -26,7 +26,8 @@ function MappedHeatmapVis(props: Props): ReactElement {
     resetDomains,
   } = useHeatmapConfig();
 
-  const dataArray = useMappedArray(dataset, value, mapperState);
+  const baseArray = useBaseArray(dataset, value);
+  const dataArray = useMappedArray(baseArray, mapperState);
   const dataDomain = useDataDomain(dataArray.data as number[]);
   const prevDataDomain = usePrevious(dataDomain);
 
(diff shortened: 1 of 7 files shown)


axelboc:
With regards to your implementation, I find nothing to say! It works amazingly well. ✨ Also, the slicing is super-fast, since the domain is looked up only once.

Your concern makes sense, the toggle is only needed for 2D+ datasets. Perhaps the solution would be to make the toggle part of the dimension mapper? After all, the `autoScale` flag is not used by the `LineVis` but by its parent, `MappedLineVis`. That being said, the toolbar just feels like a better place for it...

We could do the same as with the `dataDomain`: add a flag in the config state to disable the auto-scale button for 1D datasets, and set it from `MappedLineVis` with `useEffect`. It doesn't matter if it is enabled for only one render.

loichuder:
> With regards to your implementation, I find nothing to say! It works amazingly well. sparkles Also, the slicing is super-fast, since the domain is looked up only once.
> 
> Your concern makes sense, the toggle is only needed for 2D+ datasets. Perhaps the solution would be to make the toggle part of the dimension mapper? After all, the `autoScale` flag is not used by the `LineVis` but by its parent, `MappedLineVis`. That being said, the toolbar just feels like a better place for it...
I thought about it but we end up reversing the problem: in `DimensionMapper`, we know the mapping but we wouldn't know which is the active Vis and if it supports the auto-scale. But yeah, in terms of UI, I find it fits better in the toolbar.
> 
> We could do the same as with the `dataDomain`: add a flag in the config state to disable the auto-scale button for 1D datasets, and set it from `MappedLineVis` with `useEffect`. It doesn't matter if it is wrongly enabled or disabled for only one render.
Good call. It ends up a bit too complex for my taste but there are not many ways around it.
> 
> On the matter of the term "auto-scale", I find it quite clear, actually. I even like the icon you chose smile -- I would just add a hyphen to the button label: "Auto-scale" instead of "Auto scale".
:+1:

axelboc:
By the way, what do you think of moving the "Auto-scale" toggle to the left of the "Grid" toggle?

axelboc (approved):

loichuder:
I foresee that we will add the auto-scale button to the Heatmap as well. As it will interact with the domain, I think it would make sense to have the auto-scale to the left in this case.

As a consequence, for consistency, I would keep to Auto-scale to the left in the LineVis as well :smile:

PR #222, opened 2020-09-29
Reduce lib bundle size by tree-shaking ThreeJS

### LINKED ISSUE #216: Tree-shaking ThreeJS
`@h5web/lib` is huge (over 1 MB) because of ThreeJS. We should investigate creating a custom build to only include the features we use.

axelboc:
Fix #216 -- see https://github.com/silx-kit/h5web/issues/216#issuecomment-700613125 for details (which I've updated with the solution implemented in this PR).

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,15 +61,15 @@ from `react-icons/fi`.
 
 The `src/packages` folder contains entry points for packages published to NPM.
 
-To publish the packages, bump their version numbers in their respective `packages/package-<name>.json` files, then
+To publish the packages, bump their version numbers in their respective `packages/<name>/package.json` files, then
 follow the steps below:
 
 ```bash
 cd packages
 npm install
 npm run build
-cd dist-lib && npm publish
-cd ../dist-app && npm publish
+cd lib/dist && npm publish
+cd ../../app/dist && npm publish
 ```
 
 > To test the packages locally, run `npm link` instead of `npm publish`, and then `npm link @h5web/<package>` in your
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -61,15 +61,15 @@ from `react-icons/fi`.
 
 The `src/packages` folder contains entry points for packages published to NPM.
 
-To publish the packages, bump their version numbers in their respective `packages/package-<name>.json` files, then
+To publish the packages, bump their version numbers in their respective `packages/<name>/package.json` files, then
```
loichuder, line 64: :fishing_pole_and_fish:

packages/webpack.config.js
```diff
@@ -2,9 +2,11 @@ const fs = require('fs');
 const path = require('path');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const ThreeMinifierPlugin = require('@yushijinhun/three-minifier-webpack');
```
loichuder, line 5: I thought this was not satisfactory (https://github.com/silx-kit/h5web/issues/216#issuecomment-700613125) ?

loichuder (approved):

packages/webpack.config.js
axelboc: It wasn't satisfactory in the sense that not much was being removed with tree-shaking (because of `drei`). It still makes the tree-shaking work for ThreeJS.

PR #225, opened 2020-09-30
Added Auto-scale for Heatmap

### LINKED ISSUE #221: [Heatmap] Add auto-scaling toggle
Following #182, we could add the auto-scaling toggle to the Heatmap when browsing 3D+ datasets. That would be useful to keep the contrast consistent when exploring an image stack.

Note that we should think about how it will interact with the custom domain.

loichuder:
Fix #221 

In terms of interaction with the custom domain:

- When autoscale is off, the custom domain is kept when moving through slices.
- When autoscale is on, the custom domain is set to the `dataDomain` (that is the domain of the slice) at each slice change. However, it can still be changed when looking at a single slice (same as before).
- When toggling the autoscale, the custom domain is reset

### DIFF
--- a/src/h5web/visualizations/MappedHeatmapVis.tsx
+++ b/src/h5web/visualizations/MappedHeatmapVis.tsx
@@ -24,12 +24,21 @@ function MappedHeatmapVis(props: Props): ReactElement {
     keepAspectRatio,
     showGrid,
     resetDomains,
+    autoScale,
+    disableAutoScale,
   } = useHeatmapConfig();
 
   const baseArray = useBaseArray(dataset, value);
   const dataArray = useMappedArray(baseArray, mapperState);
-  const dataDomain = useDataDomain(dataArray.data as number[]);
+  const dataDomain = useDataDomain(
+    (autoScale ? dataArray.data : baseArray.data) as number[]
+  );
   const prevDataDomain = usePrevious(dataDomain);
+
+  // Disable `autoScale` for 2D datasets (baseArray and dataArray span the same values)
+  useEffect(() => {
+    disableAutoScale(!baseArray.shape || baseArray.shape.length <= 2);
+  }, [baseArray.shape, disableAutoScale]);
 
   // Use `customDomain` if any, unless `dataDomain` just changed (in which case it is stale and needs to be reset - cf. `useEffect` below)
   const domain =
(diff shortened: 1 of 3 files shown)


src/h5web/visualizations/MappedHeatmapVis.tsx
```diff
@@ -24,13 +24,22 @@ function MappedHeatmapVis(props: Props): ReactElement {
...
+    autoScale,
+    disableAutoScale,
   } = useHeatmapConfig();
 
   const baseArray = useBaseArray(dataset, value);
   const dataArray = useMappedArray(baseArray, mapperState);
-  const dataDomain = useDataDomain(dataArray.data as number[]);
+  const dataDomain = useDataDomain(
+    (autoScale ? dataArray.data : baseArray.data) as number[]
+  );
```
loichuder, line 35: I feel we might do something better here to use the fact that `dataDomain` needs not to be recomputed if `autoScale` is `false`

axelboc: What do you not like about this? Not sure I understand.

src/h5web/visualizations/MappedHeatmapVis.tsx
```diff
@@ -24,13 +24,22 @@ function MappedHeatmapVis(props: Props): ReactElement {
...
   const prevDataDomain = usePrevious(dataDomain);
 
+  // Disable `autoScale` for 2D datasets (baseArray and dataArray are the same)
```
axelboc, line 38: I wouldn't say the two arrays are the same, since they may be remapped (D0 <-> D1) -- so I would remove the part in parens.

axelboc (approved):

src/h5web/visualizations/MappedHeatmapVis.tsx
loichuder: Nevermind, I had the false impression that `dataDomain` was recomputed as each slice change when `autoScale` is false.

It is fine !

src/h5web/visualizations/MappedHeatmapVis.tsx
loichuder: :fishing_pole_and_fish:

PR #227, opened 2020-10-02
Make breadcrumbs responsive

### LINKED ISSUE #127: Responsive breadcrumbs
image

We could just [hide the parent crumbs](https://codepen.io/jyoakum-swx/pen/meRYjd) for now, or [crop them with ellipsis](https://codepen.io/JEETPAL/pen/qNKWRy). If we decide to make them clickable we'd have to turn them into a sort of [drop-down menu](https://codepen.io/mejiaj/pen/Kdwma) or make them [expandable](https://codepen.io/fraddski/pen/azjwev).

axelboc:
Fix #127 

Because of the resizable explorer panel, we can't rely on a simple CSS media query to hide crumbs. So to avoid measuring things with JS, I just went with `text-overflow: ellipsis;`.

Thanks to flexbox, the crumbs all shrink proportionately, so it actually looks great. I tell the last crumb to not shrink, though.

image

If there's not enough space even for the last crumb, then a last-resort `overflow: hidden` comes into play:

image

### DIFF
--- a/src/h5web/BreadcrumbsBar.module.css
+++ b/src/h5web/BreadcrumbsBar.module.css
@@ -7,19 +7,32 @@
 }
 
 .breadCrumbs {
+  flex: 1 1 0%;
   display: flex;
   align-items: center;
+  justify-content: center;
+  overflow: hidden;
+  min-width: 0;
   margin-bottom: 0;
+  padding: 0 0.25rem;
   font-size: inherit;
   font-weight: inherit;
+  white-space: nowrap;
 }
 
 .crumb {
   margin: 0 0.125rem 0.125rem;
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 
 .crumb[data-current] {
+  flex: none; /* never shrink current breadcrumb */
   font-weight: 600;
+}
+
+.separator {
+  min-width: 0.75rem;
 }
 
 .modeToggler {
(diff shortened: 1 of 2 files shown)


loichuder (approved):
I like how you handled solely with CSS ! 

We can merge it as it is but what do you think of replacing the middle part by ellipses instead ? Like having `entry_0000 > ... > intensity_normed` in your example.

axelboc:
> We can merge it as it is but what do you think of replacing the middle part by ellipses instead ? Like having `entry_0000 > ... > intensity_normed` in your example.

Hehe, what you're suggesting requires hiding chevrons and applying an ellipsis across multiple crumbs. Unfortunately, you can't have an ellipsis across multiple elements unless they're all inline and you put `text-overflow: ellipsis` on the parent. Here we use flexbox to align the crumbs and chevrons so the ellipsis can't be set on the flex container -- it has to be set on the flex items, and each flex item shows the ellipsis based on its own size -- it doesn't know about the size of its neighbours... All in all, I don't think it's worth trying to replace Flexbox or to use JS for this.

loichuder:
Yeah, I figured that it would be move involved and probably need the same kind of tricks you used in #226. Moreover, the first group is perhaps not the most relevant.

Well, it is just an idea worth keeping in mind.

PR #228, opened 2020-10-02
Show loader while fetching datasets values

### LINKED ISSUE #223: Silx provider loading state
We should show some sort of loader while datasets are being fetched.

axelboc:
Fix #223 

To retrieve the value of a dataset, providers expose a function called `getValue` that returns a promise. The value is therefore retrieved asynchronously whether or not the provider already has access to it in memory. One challenge was to not flash the loader when this is the case.

I used react-use's `useAsyncFn` hook for this. Hopefully this is clear in the code, but the goal is to set the `loading` flag to `true` only if `getValue` takes more than 50ms to complete.

ezgif-5-174754273f54

### DIFF
--- a/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
+++ b/src/h5web/dataset-visualizer/DatasetVisualizer.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement, useEffect, useContext } from 'react';
-import { useSetState } from 'react-use';
+import { useSetState, useTimeoutFn } from 'react-use';
 import type { HDF5Dataset, HDF5Value } from '../providers/models';
 import styles from './DatasetVisualizer.module.css';
 import VisSelector from './VisSelector';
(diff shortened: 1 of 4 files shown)


src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -44,15 +45,24 @@ function DatasetVisualizer(props: Props): ReactElement {
     });
   }
 
-  // Fetch dataset value asynchronously
   const { getValue } = useContext(ProviderContext);
+  const [scheduleLoadingFlag, cancelLoadingFlag] = useTimeoutFn(() => {
+    mergeState({ loading: true });
+  }, 50);
```
loichuder, line 51: I do not get why you need to use a function with a timeout. Is it to skip the loading part if it goes fast enough ?

loichuder (approved):
Dat loader :heart:

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
loichuder: Ok my bad. Next time, I will read the PR description :see_no_evil:

PR #229, opened 2020-10-02
Improve display of log ticks

### LINKED ISSUE #183: [Line] Too many ticks with log scale

axelboc:
Fix #183

I just had another crack at the issue, and it's looking good... 😤 

1. I log the min and max of the axis' visible domain and compute the difference.
1. I compare the result with a threshold, which I compute with a linear scale based on the space available to the axis (similarly to `adaptedNumTicks`).
1. If the logged extent of the domain spans less than the threshold, then it means that all ticks can be displayed without any overlap and I use the default tick formatter (`format('0')`).
1. Otherwise, I use a formatter that displays only exact powers of 10.

The only downside I see is that I don't check whether the visible ticks are far apart or close together; the threshold is computed on the assumption that the ticks are close together. As a result, there may be cases where zero or one tick is visible when there would be space to show more (e.g. on a narrow screen, a little above a power of 10). That being said, I don't think it matters and I believe matplotlib has the same downside (though this needs to be verified).

### Height of 465px

image

image

image

### Height of 230px

image

image

### Case where more ticks could be shown

image

### DIFF
--- a/src/h5web/visualizations/shared/AxisSystem.tsx
+++ b/src/h5web/visualizations/shared/AxisSystem.tsx
@@ -1,19 +1,17 @@
 import React, { useState, useContext } from 'react';
 import { useThree, useFrame } from 'react-three-fiber';
 import { AxisLeft, AxisBottom, TickRendererProps } from '@vx/axis';
-import { format } from 'd3-format';
 import { GridColumns, GridRows } from '@vx/grid';
 import { useUpdateEffect } from 'react-use';
 import Html from './Html';
 import styles from './AxisSystem.module.css';
 import type { AxisOffsets, Domain } from './models';
-import { getTicksProp, getAxisScale } from './utils';
+import { getTicksProp, getAxisScale, getTickFormatter } from './utils';
 import { AxisSystemContext } from './AxisSystemProvider';
 
 const AXIS_PROPS = {
   tickStroke: 'grey',
   hideAxisLine: true,
-  tickFormat: format('0'),
   tickClassName: styles.tick,
   tickComponent: ({ formattedValue, ...tickProps }: TickRendererProps) => (
     <text {...tickProps}>{formattedValue}</text>
@@ -80,6 +78,17 @@ function AxisSystem(props: Props): JSX.Element {
     ordinateInfo.isIndexAxis
   );
 
+  const xTickFormat = getTickFormatter(
+    visibleDomains[0],
+    width,
+    abscissaInfo.scaleType
+  );
+  const yTickFormat = getTickFormatter(
+    visibleDomains[1],
+    height,
+    ordinateInfo.scaleType
+  );
+
   return (
     <Html
       className={styles.axisSystem}
@@ -95,7 +104,12 @@ function AxisSystem(props: Props): JSX.Element {
     >
       <div className={styles.bottomAxisCell}>
         <svg className={styles.axis} data-orientation="bottom">
-          <AxisBottom scale={xTicksScale} {...xTicksProp} {...AXIS_PROPS} />
+          <AxisBottom
+            scale={xTicksScale}
+            tickFormat={xTickFormat}
+            {...xTicksProp}
+            {...AXIS_PROPS}
+          />
         </svg>
       </div>
       <div className={styles.leftAxisCell}>
@@ -103,6 +117,7 @@ function AxisSystem(props: Props): JSX.Element {
           <AxisLeft
             scale={yTicksScale}
             left={axisOffsets.left}
+            tickFormat={yTickFormat}
             {...yTicksProp}
             {...AXIS_PROPS}
           />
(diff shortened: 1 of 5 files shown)


src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -80,6 +78,17 @@ function AxisSystem(props: Props): JSX.Element {
...
+  const xTickFormat = getTickFormatter(
+    visibleDomains[0],
+    width,
+    abscissaInfo.scaleType
+  );
+  const yTickFormat = getTickFormatter(
+    visibleDomains[1],
+    height,
+    ordinateInfo.scaleType
+  );
```
axelboc, line 90: `AxisSystem` is becoming a bit too complex for my taste, especially considering that the signature of `getTickFormatter` is very similar to that of `getTickProps`. There's clearly a need for a refactor, but I'll work on it if/once this is merged.

loichuder: Related to that, I find that `getTickProps` is a bad function in itself as the return value changes completely depending on a flag.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -10,11 +11,17 @@ import {
...
+const TICK_FORMAT = format('0');
+
 export const adaptedNumTicks = scaleLinear()
   .domain([300, 900])
   .rangeRound([3, 10])
   .clamp(true);
 
+const adaptedLogTicksThreshold = scaleLinear()
+  .domain([300, 500])
+  .range([0.8, 1.4]);
```
loichuder, line 23: How did you determine the threshold values ?

src/h5web/visualizations/shared/utils.ts
```diff
@@ -144,6 +151,29 @@ export function getTicksProp(
...
+  availableSize: number,
+  scaleType: ScaleType
+): (val: number | { valueOf(): number }) => string {
+  if (scaleType !== ScaleType.Log) {
+    return TICK_FORMAT;
+  }
+
+  // If available size allows for all log ticks to be rendered without overlap, use default formatter
+  const threshold = adaptedLogTicksThreshold(availableSize);
+  if (Math.log10(domain[1]) - Math.log10(domain[0]) < threshold) {
```
loichuder, line 165: `domain` may be composed of two negative values making the issue of this uncertain.

loichuder (approved):
I would need some info on how you determined the threshold to give an enlightened review on that :smile_cat:

src/h5web/visualizations/shared/AxisSystem.tsx
axelboc: Hmm can't think of a refactor that'd be worthwhile. If I simplify things somewhere, I complexify them somewhere else... My initial idea was to have something like this to replace lines 61 to 90:

```tsx
const [xAxisProps, xGridProps] = getAxisProps(visibleDomain[0], width, abscissaInfo);
const [yAxisProps, yGridProps] = getAxisProps(visibleDomain[1], height, ordinateInfo);
// ... where `<x|y>AxisProps` includes `scale`, `tickFormat` and (`numTicks` or `tickValues`)
// ... and `<x|y>GridProps` includes `scale` and (`numTicks` or `tickValues`)

// ... which I could all destructure easily in the JSX
<AxisBottom
  {...xAxisProps}
  {...AXIS_PROPS}
/>
```

But the range of the x-axis scale is `[0, width]`, while the range of the y-axis scale is `[height, 0]`, so it's not as easy... Assuming I keep the `scale` computation in `AxisSystem`, I could probably get rid of `getTicksProp` and move the logic for this in `getAxisProp`, but I don't think I could get rid of `getTickFormatter`... So I wonder whether it's worth the extra level of abstraction...

src/h5web/visualizations/shared/utils.ts
axelboc: Same way we determined the values for `adaptedNumTicks`: I picked a height and found the threshold that worked for it, then picked another height and did the same. Granted, the thresholds may not work as well for the horizontal axis... but I figured we'd deal with this in due time. One problem at a time! 😄

src/h5web/visualizations/shared/utils.ts
axelboc: Good catch, should work by comparing the absolute of the difference, right? `Math.abs(Math.log10(domain[1]) - Math.log10(domain[0])) < threshold`

axelboc: Ah no, because the values are reversed...

axelboc: arf...

src/h5web/visualizations/shared/AxisSystem.tsx
loichuder: To be discussed elsewhere more in depth ^^

loichuder (approved):

PR #230, opened 2020-10-05
Move toolbar components and configs to `src/h5web/toolbar` folder

axelboc:
The `visualizations` folder was becoming a bit packed... I've grouped all the toolbar-related components ~~and configs~~ in a new, higher-level folder: `src/h5web/toolbar`. This restructuration seems to make sense now that the core visualization components are "dumber", so to speak.


src/h5web/toolbar/ToggleGroup.module.css
```diff
@@ -4,12 +4,12 @@
 }
 
 .btn {
-  composes: btn from '../shared/Toolbar.module.css';
+  composes: btn from './Toolbar.module.css';
```
axelboc, line 7: These imports could have been simplified before.

src/h5web/toolbar/heatmap/ColorMapSelector.tsx
```diff
@@ -4,9 +4,15 @@ import { Wrapper, Button, Menu, MenuItem } from 'react-aria-menubutton';
...
-import { generateCSSLinearGradient } from './utils';
+import {
+  INTERPOLATOR_GROUPS,
+  INTERPOLATORS,
+} from '../../visualizations/heatmap/interpolators';
+import type {
+  ColorMap,
+  D3Interpolator,
+} from '../../visualizations/heatmap/models';
+import { generateCSSLinearGradient } from '../../visualizations/heatmap/utils';
```
axelboc, line 15: This is the main downside: importing utilities and models from toolbar components is a bit more tedious.

src/h5web/toolbar/ToggleGroup.module.css
loichuder: :fishing_pole_and_fish:

loichuder (approved):
Most of it makes sense to me.

However, I would have left the vis-specific components (and configs) in their folder to avoid the creation of folders `toolbar/heatmap` and `toolbar/line` which make some imports tedious. What do you think ?

axelboc:
> However, I would have left the vis-specific components (and configs) in their folder to avoid the creation of folders `toolbar/heatmap` and `toolbar/line` which make some imports tedious. What do you think ?

Hmm tough call. I feel like all the controls at least should be in the same place, even `ColorMapSelector` and such. We may need a domain slider or a color map selector for another type of visualization one day.

However, you're right, I don't have as good a feeling about `HeatmapToolbar`, `LineToolbar` and the configs. How about I moved them at the root of the `visualizations` folder, next to `MappedHeatmapVis` and the like?

axelboc:
With `HeatmapToolbar`, `LineToolbar` and the configs out of the way I could just remove the `toolbar/heatmap` and `toolbar/line` subfolders.

loichuder:
> However, you're right, I don't have as good a feeling about HeatmapToolbar, LineToolbar and the configs. How about I moved them at the root of the visualizations folder, next to MappedHeatmapVis and the like?

Well, I found them quite at home in the `visualisations/line` and `visualisations/heatmap` folders.

> With HeatmapToolbar, LineToolbar and the configs out of the way I could just remove the toolbar/heatmap and toolbar/line subfolders.

Agreed !

axelboc:
> Well, I found them quite at home in the visualisations/line and visualisations/heatmap folders.

Yeah true, there isn't much left in those folders anyway. Should I move `MappedHeatmapVis` and `MappedLineVis` in there as well?

loichuder:
> Yeah true, there isn't much left in those folders anyway. Should I move MappedHeatmapVis and MappedLineVis in there as well?

Your call. I don't mind.

axelboc:
### BEFORE

```
+ visualizations
  + heatmap
    - ColorMapSelector
    - HeatmapVis
    - HeatmapToolbar
    - config
  + shared
    - Toolbar
    - ToggleGroup
    - OverflowMenu
    ...
  - MappedHeatmapVis
  ...
```

### AFTER

```
+ toolbar
  + controls
    - ToggleGroup
    - ColorMapSelector
    ...
  - HeatmapToolbar
  - LineToolbar
  - Toolbar
  - OverflowMenu
+ visualizations
  + heatmap
    - config
    - MappedHeatmapVis
    - HeatmapVis
    ...
```

PR #239, opened 2020-10-07
Refactor axis system and fix re-rendering bug

axelboc:
I've split the fix/refactor in two commits:

1. In the first commit, I fix the bug where the axes were being temporarily reset on any change, like the color map.

    The bug occurs because of the `useUpdateEffect` in `AxisSystem`, which resets the visible domains to the initial axis domains any time the latter changed ... which happens on every render of `HeatmapVis` and `LineVis` since we pass inline arrays as domains - e.g. `[0, rows]`.

    I could have fixed this by computing the visible domains in the `useUpdateEffect` (in addition to computing them in `useFrame`), but this led to an incoherent state for one render when the newly calculated scale would no longer be coherent with the yet-to-be-calculated visible domains.

    I then realised that storing the visible domains in the state was effectively useless and that a better solution would be to just compute them on every render and to force the `AxisSystem` to re-render on every R3F frame directly. We still re-render as many times as we did with `useFrame` and `useUpdateEffect`; the difference is just that we compute the visible domains synchronously during the render rather than asynchronously inside an effect. It's a lot simpler, really.

2. In the second commit, I refactor `AxisSystem` by ntroducing a new `Axis` component that takes care of rendering the VX axis and grid components. The computation of the scales and visible domains stays in `AxisSystem`, as each axis has specificities that are difficult to abstract out.

### DIFF
--- a/src/h5web/visualizations/shared/AxisSystem.module.css
+++ b/src/h5web/visualizations/shared/AxisSystem.module.css
@@ -2,7 +2,7 @@
   display: grid;
   grid-template-areas:
     '. top-axis .'
-    'left-axis axis-grid right-axis'
+    '. axis-grid right-axis'
     '. bottom-axis .';
 
   /* - Prevent panning/zooming from axis when zoomed in.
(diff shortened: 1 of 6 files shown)


src/h5web/visualizations/shared/AxisSystem.module.css
```diff
@@ -2,45 +2,32 @@
...
    * - Prevent text selection of tick labels when panning with touchpad. */
   pointer-events: none;
 }
 
-.leftAxisCell {
-  grid-area: left-axis;
-  display: flex;
-  flex-direction: column;
-  min-height: 0;
-}
```
axelboc, line 18: I removed the `div`'s that wrapped the `svg`'s -- they just complicated things.

src/h5web/visualizations/shared/hooks.ts
```diff
@@ -43,3 +44,11 @@ export function useMappedArray<T>(
...
 
 export const useSupportedDomain = createMemo(getSupportedDomain);
+
+export function useFrameRendering(): void {
+  const [, setNum] = useState();
+
+  useFrame(() => {
+    setNum(Math.random());
+  });
+}
```
axelboc, line 54: A bit of trickery to trigger a re-render in React on each R3F frame. This is what we used to do before, so there's no change in performance. Another way to see it is that this hook allows us to use the camera position and zoom during renders.

src/h5web/visualizations/shared/AxisSystem.module.css
```diff
@@ -2,45 +2,32 @@
...
 }
 
-.axis[data-orientation='left'] {
+.axis[data-type='abscissa'] {
+  grid-area: bottom-axis;
+  text-anchor: middle;
+}
+
+.axis[data-type='ordinate'] {
+  grid-area: axis-grid; /* `AxisLeft` aligns ticks to the outer left of its own bounding box  */
```
loichuder, line 25: Odd.

loichuder (approved):
Good to me :+1:

PR #240, opened 2020-10-07
Updated mock data with nexus entry

loichuder:
Note that regenerating the mock data changes the datasets'  `id`. We then need a more robust way to access mocked datasets in Stories.

I choose to `find` datasets from their `alias` with an explicit cast to avoid TS errors.


src/stories/Home.stories.mdx
```diff
@@ -90,7 +90,7 @@ The library exposes a mock data object for testing purposes:
 ```ts
 import { mockData } from '@h5web/lib';
 
-const dataset = mockData.datasets['c8f60c29-aae2-11ea-84a9-b94ddd2ec9e8'];
+const dataset = mockData.datasets['c8f60c29-aae2-11ea-84a9-b94ddd2ec9e8']; // THIS WONT WORK
```
loichuder, line 93: This needs to be changed by something more robust. I am reluctant to put the method I used in the stories as it is a bit heavy.

axelboc: The way you find mock datasets is fine I think. Could put it in a utility function somewhere, perhaps in the mock data provider's folder and expose it in the library so it can be used in other stories and in the `Home` documentation?

src/stories/HeatmapVis.stories.tsx
```diff
@@ -12,7 +12,9 @@ import { ScaleType } from '../h5web/visualizations/shared/models';
 import { INTERPOLATORS } from '../h5web/visualizations/heatmap/interpolators';
 
 // A 2D dataset
-const dataset = mockData.datasets['c8f60c29-aae2-11ea-84a9-b94ddd2ec9e8'];
+const dataset = Object.values(mockData.datasets).find(
+  (d) => d.alias[0] === '/nD/twoD'
+) as { value: number[][]; shape: { dims: number[] } };
```
axelboc, line 17: Could you maybe re-use an existing provider interface for the casting?

axelboc (approved):

src/stories/HeatmapVis.stories.tsx
loichuder: If I use `HDF5...` types, I run into all sorts of TS errors: the JSON data appears to not fit exactly that and then, I still need to cast afterwards as the `HDF5...` are too generic.

src/h5web/providers/mock/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  value: number[][];
+  shape: { dims: number[] };
+};
+
+export function getMocked2dDataset(): TwoDDataset {
+  const dataset = Object.values(mockData.datasets).find(
+    (d) => d.alias[0] === '/nD/twoD'
+  ) as TwoDDataset;
+  return dataset;
+}
```
axelboc, line 13:
```ts
interface MockedDataset<T extends number[] | number[][] | number[][][]> = {
  value: T;
  shape: { dims: number[] };
};

export function getMockedDataset<T>(): MockedDataset<T> {
  const dataset = Object.values(mockData.datasets).find(
    (d) => d.alias[0] === '/nD/twoD'
  ) as MockedDataset<T>;
  return dataset;
}
```

PR #241, opened 2020-10-07
Implement NeXus visualizer for NXdata groups

axelboc:


src/h5web/App.tsx
```diff
@@ -60,7 +50,7 @@ function App(): JSX.Element {
               entity={selectedEntity}
             />
           ) : (
-            <DatasetVisualizer dataset={selectedDataset} />
+            <Visualizer entity={selectedEntity} />
```
axelboc, line 53: I introduce a new top-level component called `Visualizer` that decides which "visualizer" to show, if any.

src/h5web/Visualizer.tsx
```diff
@@ -0,0 +1,30 @@
...
+  if (entity && isDataset(entity)) {
+    return <DatasetVisualizer dataset={entity} />;
+  }
+
+  if (
+    entity &&
+    isGroup(entity) &&
+    entity.attributes?.find(({ value }) => value === 'NXdata')
+  ) {
+    return <NexusVisualizer group={entity} />;
```
axelboc, line 24: The new `NexusVisualizer`. We may end up moving the `NXdata` detection code inside the visualizer... For now, it can live here.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -1,20 +1,20 @@
...
 import VisDisplay from './VisDisplay';
-import type { Vis, DimensionMapping } from './models';
+import { DimensionMapping, Vis } from './models';
 import { VIS_DEFS } from '../visualizations';
 import { ProviderContext } from '../providers/context';
 import { useSupportedVis } from './hooks';
 
 interface Props {
-  dataset?: HDF5Dataset;
+  dataset: HDF5Dataset;
```
axelboc, line 13: The `dataset` can no longer be `undefined`. This brings back some simpler code that we had at some point.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -1,20 +1,20 @@
...
 }
 
 interface State {
-  activeVis?: Vis;
+  activeVis: Vis;
```
axelboc, line 17: I initialise the local state with a default `Vis`, which allows me to remove another bunch of checks.

src/h5web/nexus-visualizer/NexusVisualizer.tsx
```diff
@@ -0,0 +1,19 @@
...
+interface Props {
+  group: HDF5Group;
+}
+
+function NexusVisualizer(props: Props): ReactElement {
+  const { group } = props;
+  return (
+    <div className={styles.visualizer}>
+      <div className={styles.visBar} />
+      <p className={styles.fallback}>NeXus group {group.id}</p>
```
axelboc, line 14: Just showing the ID of the group for now.

This component may evolve to become a lot similar to `DatasetVisualizer` (a `VisSelector`, a toolbar, a `VIS_DEFS` object, etc.), but until we know more, I think it's better to duplicate things as needed instead of adding complexity to the existing code that may lead us to a dead end.

src/h5web/Visualizer.tsx
loichuder: Yes, it is fine for now.

Just an idea: perhaps, we will need stronger typings for `NXData` for future features (dedicated type + corresponding typeguard).

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -25,22 +25,20 @@ function DatasetVisualizer(props: Props): ReactElement {
   const { dataset } = props;
 
   const supportedVis = useSupportedVis(dataset);
-  const [state, mergeState] = useSetState<State>({});
+  const [state, mergeState] = useSetState<State>({ activeVis: Vis.Raw });
```
loichuder, line 28: I would put a comment here to tell that you are "allowed" to do that because `Vis.Raw` supports all datasets (see `visualisation/index.ts`)

src/h5web/nexus-visualizer/NexusVisualizer.tsx
loichuder: Good call :+1:

loichuder (approved):
I like how you manage to simplify `DatasetVisualizer` through the addition of the `NexusVisualizer` :+1: 

For the name, perhaps `NexusDataVisualizer` would be more appropriate but it feels a bit heavy.

axelboc:
> For the name, perhaps NexusDataVisualizer would be more appropriate but it feels a bit heavy.

Yeah I did think of it, but I wasn't sure whether this component would end up dealing with the `default` attribute and other Nexus features... We'll see how things evolves.

loichuder:
> > For the name, perhaps NexusDataVisualizer would be more appropriate but it feels a bit heavy.
> 
> Yeah I did think of it, but I wasn't sure whether this component would end up dealing with the `default` attribute and other Nexus features... We'll see how things evolves.

Good point ! Agreed :+1:

PR #242, opened 2020-10-09
Move fetching of dataset value down into VisDisplay

axelboc:


src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -66,30 +29,20 @@ function DatasetVisualizer(props: Props): ReactElement {
...
-              mapperState: VIS_DEFS[vis].defaultMappingState(dataset),
-            });
-          }}
+          onChange={setActiveVis}
         />
         {VisToolbar && <VisToolbar />}
       </div>
       <div className={styles.displayArea}>
         <VisDisplay
+          key={`${dataset.id}_${activeVis}`}
```
axelboc, line 38: This forces a reset of the state of `VisDisplay` (dataset value + mapper state) whenever the visualization and/or dataset changes.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
```diff
@@ -1,63 +1,26 @@
...
 interface Props {
   dataset: HDF5Dataset;
 }
 
-interface State {
-  activeVis: Vis;
-  value?: HDF5Value;
-  loading?: boolean;
-  mapperState?: DimensionMapping;
-  prevDataset?: HDF5Dataset;
```
axelboc, line 21: The derived state condition below is much simpler, and we no longer need to store the `prevDataset`.

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
loichuder: I would rather see this in comment in the code :wink;

loichuder (approved):
:rocket:

src/h5web/dataset-visualizer/DatasetVisualizer.tsx
axelboc: Oh sorry, I did mean to 😓

Issue #234, opened 2020-10-06
[Nexus] Heatmap vis for 3D+ @signal


axelboc:
Part of epic #196

loichuder:
Fixed by #243

Issue #232, opened 2020-10-06
[Nexus] Heatmap vis with 2D @signal (no @axes)


axelboc:
Part of epic #196

loichuder:
Fixed by #243

PR #243, opened 2020-10-09
NexusVisualizer now provides a vis for NXdata group with valid signal attributes

### LINKED ISSUE #231: [Nexus] Line vis for 1D @signal (no @axes)
Part of epic #196 

### First-issue prerequisites

- [x] Mock Nexus data
- [ ] Nexus support detection + Nexus visualization
  - Pass groups to `DatasetVisualizer`
  - Add vis component and vis def
  - Show vis when group has `NX_class="NXdata"` attribute

loichuder:
Fix #231, #232 and #234 :shrug:

### DIFF
--- a/src/h5web/dataset-visualizer/VisSelector.tsx
+++ b/src/h5web/dataset-visualizer/VisSelector.tsx
@@ -6,7 +6,7 @@ import { VIS_DEFS } from '../visualizations';
 interface Props {
   activeVis: Vis;
   choices: Vis[];
-  onChange: (vis: Vis) => void;
+  onChange?: (vis: Vis) => void;
 }
 
 function VisSelector(props: Props): JSX.Element {
@@ -16,6 +16,7 @@ function VisSelector(props: Props): JSX.Element {
     <div className={styles.selector} role="tablist" aria-label="Visualization">
       {choices.map((vis) => {
         const { Icon } = VIS_DEFS[vis];
+        const onClick = onChange ? () => onChange(vis) : undefined;
         return (
           <button
             key={vis}
@@ -23,9 +24,7 @@ function VisSelector(props: Props): JSX.Element {
             type="button"
             role="tab"
             aria-selected={vis === activeVis}
-            onClick={() => {
-              onChange(vis);
-            }}
+            onClick={onClick}
           >
             <Icon className={styles.icon} />
             {vis}
--- a/src/h5web/nexus-visualizer/NexusVisualizer.tsx
+++ b/src/h5web/nexus-visualizer/NexusVisualizer.tsx
@@ -1,6 +1,14 @@
 import React, { ReactElement } from 'react';
 import type { HDF5Group } from '../providers/models';
+import { getAttributeValue, getInterpretationVis } from './utils';
 import styles from '../Visualizer.module.css';
+import { useEntity } from '../providers/hooks';
+import { isDataset } from '../providers/utils';
+import VisSelector from '../dataset-visualizer/VisSelector';
+import VisDisplay from '../dataset-visualizer/VisDisplay';
+import { VIS_DEFS } from '../visualizations';
+import { getSupportedVis } from '../dataset-visualizer/utils';
+import { NXDataAttribute } from './models';
 
 interface Props {
   group: HDF5Group;
@@ -8,10 +16,58 @@ interface Props {
 
 function NexusVisualizer(props: Props): ReactElement {
   const { group } = props;
+
+  const signalDatasetTitle = getAttributeValue(group, NXDataAttribute.Signal);
+  const signalLink =
+    signalDatasetTitle &&
+    group.links?.find((l) => l.title === signalDatasetTitle);
+  const signalDataset = useEntity(signalLink);
+
+  if (signalDatasetTitle === undefined) {
+    return (
+      <p className={styles.fallback}>NXdata group has no attribute signal.</p>
+    );
+  }
+
+  if (!signalLink || !signalDataset) {
+    return (
+      <p className={styles.fallback}>
+        NXdata signal {signalDatasetTitle} points to an non-existing dataset.
+      </p>
+    );
+  }
+
+  if (!signalDataset || !isDataset(signalDataset)) {
+    return (
+      <p className={styles.fallback}>
+        NXdata signal {signalDatasetTitle} is not a valid dataset.
+      </p>
+    );
+  }
+
+  const supportedVis = getSupportedVis(signalDataset);
+  const interpretationVis = getInterpretationVis(group);
+  // Prefer vis deduced from interpretation. Else take the first supported vis.
+  const nexusVis =
+    interpretationVis && supportedVis.includes(interpretationVis)
+      ? interpretationVis
+      : supportedVis[supportedVis.length - 1];
+
+  const VisToolbar = VIS_DEFS[nexusVis].Toolbar;
+
   return (
     <div className={styles.visualizer}>
-      <div className={styles.visBar} />
-      <p className={styles.fallback}>NeXus group {group.id}</p>
+      <div className={styles.visBar}>
+        <VisSelector activeVis={nexusVis} choices={[nexusVis]} />
+        {VisToolbar && <VisToolbar />}
+      </div>
+      <div className={styles.displayArea}>
+        <VisDisplay
+          key={signalDataset.id} // reset local state when dataset changes
+          activeVis={nexusVis}
+          dataset={signalDataset}
+        />
+      </div>
     </div>
   );
 }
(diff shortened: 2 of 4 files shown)


src/h5web/nexus-visualizer/NexusVisualizer.tsx
```diff
@@ -1,17 +1,70 @@
...
   return (
     <div className={styles.visualizer}>
-      <div className={styles.visBar} />
-      <p className={styles.fallback}>NeXus group {group.id}</p>
+      <div className={styles.visBar}>
+        <VisSelector
+          activeVis={nexusVis}
+          choices={[nexusVis]}
+          onChange={() => {}}
+        />
```
loichuder, line 58: This could be enhanced as we don't need to support vis selection.

src/h5web/nexus-visualizer/NexusVisualizer.tsx
```diff
@@ -1,17 +1,70 @@
...
+        {VisToolbar && <VisToolbar />}
+      </div>
+      <div className={styles.displayArea}>
+        <VisDisplay
+          key={`${signalDataset.id}`} // reset local state when active vis or dataset changes
+          activeVis={nexusVis}
+          dataset={signalDataset}
+        />
+      </div>
```
loichuder, line 67: Copy-paste from `DatasetVisualizer` for now

src/h5web/nexus-visualizer/NexusVisualizer.tsx
```diff
@@ -1,17 +1,70 @@
...
+  const signalLink = getSignalLink(group);
+  const signalDataset = useEntity(signalLink);
+
+  if (!signalLink) {
+    return (
+      <div className={styles.visualizer}>
+        <div className={styles.visBar} />
+        <p className={styles.fallback}>
+          NXdata group has no attribute signal or it points to an non-existing
+          dataset.
```
loichuder, line 28: We might want to break down `getSignalLink` to be able to give the real cause of the "error". See my other comment on `signalLink`.

src/h5web/nexus-visualizer/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function getSignalLink(group: HDF5Group): HDF5Link | undefined {
+  const signalAttribute = group.attributes?.find((a) => a.name === 'signal');
+
+  if (!signalAttribute) {
+    return undefined;
+  }
+
+  return group.links?.find((l) => l.title === signalAttribute.value);
+}
```
loichuder, line 12: As `getSignalLink` is written, it returns `undefined` if the `signal` attribute is not found (guard clause) **or** if the link does not exist (`group.links?.find`). This means both causes are undistinguishable for the consumer but I think it is not a big issue.

src/h5web/nexus-visualizer/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Vis } from '../dataset-visualizer/models';
+
+export function getSignalLink(group: HDF5Group): HDF5Link | undefined {
+  const signalAttribute = group.attributes?.find((a) => a.name === 'signal');
+
+  if (!signalAttribute) {
+    return undefined;
+  }
+
+  return group.links?.find((l) => l.title === signalAttribute.value);
```
axelboc, line 11: `return signalAttribute && ...`

src/h5web/nexus-visualizer/NexusVisualizer.tsx
axelboc: Ah yes, in this case you may want to inline the code of `getSignalLink`.

I don't hink you need all those `div` wrappers and the vis bar. It should be fine to just return the fallback text.

axelboc: perhaps a little `getAttributeValue()` would help keep things clean

src/h5web/nexus-visualizer/NexusVisualizer.tsx
```diff
@@ -1,17 +1,70 @@
...
+        <p className={styles.fallback}>
+          NXdata signal {signalLink.title} is not a valid dataset.
+        </p>
+      </div>
+    );
+  }
+
+  // Prefer vis deduced from interpretation. Else take the first supported vis.
+  const nexusVis =
+    getInterpretationVis(group) ?? getSupportedVis(signalDataset)[0];
```
axelboc, line 47: Jumping the gun a bit here 😄 - oh well, it works. We'll have to remember to add some sort of error recovery code, if the interpretation doesn't match the dimensions of the signal dataset for instance.

axelboc: `getSupportedVis(signalDataset)[0]` will always return `Vis.Matrix` for nD datasets. You want to get the last one in the array, I think.

src/h5web/nexus-visualizer/NexusVisualizer.tsx
axelboc: Just make `onChange` optional! 🤷

src/h5web/nexus-visualizer/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+    return undefined;
+  }
+
+  if (signalInterpretation.value === 'spectrum') {
+    return Vis.Line;
+  }
+
+  if (signalInterpretation.value === 'image') {
+    return Vis.Heatmap;
+  }
```
axelboc, line 29: A `Record` constant may be more appropriate 😉

axelboc (changes requested):
One fell swoop, nicely done! I've suggested a few changes.

src/h5web/dataset-visualizer/VisSelector.tsx
```diff
@@ -16,16 +16,15 @@ function VisSelector(props: Props): JSX.Element {
     <div className={styles.selector} role="tablist" aria-label="Visualization">
       {choices.map((vis) => {
         const { Icon } = VIS_DEFS[vis];
+        const onClick = onChange ? () => onChange(vis) : undefined;
```
axelboc, line 19: Doesn't matter, but you might be able to inline it nicely with ` &&` maybe:

```
onClick={ onChange && () => {
  onChange(vis);
}}
```

src/h5web/nexus-visualizer/NexusVisualizer.tsx
```diff
@@ -1,17 +1,73 @@
...
 function NexusVisualizer(props: Props): ReactElement {
   const { group } = props;
+
+  const signalDatasetTitle = getAttributeValue(group, NXDataAttribute.Signal);
+  const signalLink =
+    signalDatasetTitle &&
+    group.links?.find((l) => l.title === signalDatasetTitle);
+  const signalDataset = useEntity(signalLink);
+
+  if (signalDatasetTitle === undefined) {
```
axelboc, line 26: If `signalDatasetTitle` is `""`, the error is still relevant, isn't it? So `!signalDatasetTitle ` might do.

src/h5web/dataset-visualizer/VisSelector.tsx
loichuder: I tried but TS didn't let me :cry:

src/h5web/nexus-visualizer/NexusVisualizer.tsx
loichuder: No as in this case, `signalDatasetTitle` exists ! So it should be caught by the second guard clause.

src/h5web/nexus-visualizer/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function getAttributeValue(
+  entity: HDF5Dataset | HDF5Group,
+  attributeName: NXDataAttribute
+): HDF5Value {
+  if (!entity.attributes) {
+    return undefined;
+  }
+
+  return entity.attributes.find((attr) => attr.name === attributeName)?.value;
```
axelboc, line 17: You didn't like `entity.attributes?.find(...)` ? Meh, doesn't matter 👍

axelboc (approved):

src/h5web/dataset-visualizer/VisSelector.tsx
axelboc: oh maybe it's `onChange?`, does it work with `| undefined` instead ? Again, no biggy either way.

src/h5web/nexus-visualizer/utils.ts
loichuder: As much as I find `?.` practical when coding, I feel it can be hard to catch when reading the code. So I prefer to have something explicit instead whenever I can :wink:

PR #245, opened 2020-10-09
Only dims are now passed to Vis rather than the full dataset

loichuder:
A small refactoring as Vis components do not need the full dataset object but only the dimensions.

In fact, I think we could even make a single object/interface out of `value` and `rawDims` to be passed to vis components.

Your thoughts @axelboc ?


axelboc:
Hmm not sure, I feel like passing `rawDims` to `VisComponent` is too specific. Some visualizations do not care about the dimensions, like `RawVis` or `ScalarVis`. TS didn't warn you because extra props are allowed on React components:

```ts
function Test<P extends {val: number}> (props: P) {}
Test({ val: 0, foo: 'test' }); // no error
```

loichuder:
True. I did not think about `RawVis` and `ScalarVis`...

PR #247, opened 2020-10-12
Fix no auto-scale behaviour with log scale and domain that crosses zero

### LINKED ISSUE #246: [Line] Disabling auto-scale with log scale still updates the Y axis domain
This happens for datasets with unsupported values for the log scale. 

See slices of `entry_0000/2_buffer_substraction/results/intensity_normed` to reproduce the bug.

axelboc:
Fix #246 

`getSupportedDomain` merges into `getDataDomain` so that `useDataDomain` now returns a supported domain right away. In other words, the visualizations no longer compute the supported domain themselves; they must now receive a supported domain as prop.

Passing an unsupported domain to a visualization fails silently (but not invisibly); it does not throw an error and break the app. This can be observed in the Storybook:

- with `LineVis`, the y axis and the curve are not rendered;
- with `HeatmapVis`, the textured mesh and the color bar's axis are not rendered.

I believe this behaviour to be adequate and sufficient.

### DIFF
--- a/src/h5web/visualizations/shared/utils.ts
+++ b/src/h5web/visualizations/shared/utils.ts
@@ -1,5 +1,5 @@
 import { scaleLinear } from 'd3-scale';
-import { extent, tickStep } from 'd3-array';
+import { tickStep } from 'd3-array';
 import { format } from 'd3-format';
 import {
   Size,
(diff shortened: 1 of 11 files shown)


src/h5web/visualizations/shared/utils.ts
```diff
@@ -45,33 +45,37 @@ export function computeVisSize(
...
+      return [
+        accMin > val ? val : accMin,
+        accMax < val ? val : accMax,
+        val > 0 && (accPositiveMin === undefined || accPositiveMin > val)
+          ? val
+          : accPositiveMin,
+      ];
+    },
+    [values[0], values[0], undefined]
+  );
```
axelboc, line 70: This algorithm requires only one loop over the array, even when the domain has to be clamped. Before, we used to call `findDomain`, and therefore `d3.extent`, twice.

loichuder: Nice !

loichuder (approved):
Funny how fixing this bug ends up simplifying things so much.

axelboc:
> Funny how fixing this bug ends up simplifying things so much.

I know right!?

PR #249, opened 2020-10-13
Fetch metadata in provider before rendering app

axelboc:
This allows the metadata to be provided synchronously to components, which simplifies a bunch of code.

`useEntity` no longer contains an async state, which means that in no longer triggers an extra rendering of `App`, which makes the use of `React.memo` in `DatasetVisualizer` obsolete. I was tempted to turn `useEntity(link)` into `getEntity(metadata, link)` to be able to call the function more freely, but I'll leave that until I actually need the flexbiity (e.g. to find the `default` NeXus group).

### DIFF
--- a/src/h5web/dataset-visualizer/VisDisplay.tsx
+++ b/src/h5web/dataset-visualizer/VisDisplay.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement, useContext, useEffect } from 'react';
-import { useSetState, useTimeoutFn } from 'react-use';
+import { useSetState, useTimeoutFn, usePromise } from 'react-use';
 import type { Vis, DimensionMapping } from './models';
 import type {
   HDF5Dataset,
@@ -31,6 +31,7 @@ function VisDisplay(props: Props): ReactElement {
   });
 
   const { getValue } = useContext(ProviderContext);
+  const ifMounted = usePromise();
   const [scheduleLoadingFlag, cancelLoadingFlag] = useTimeoutFn(() => {
     mergeState({ loading: true });
   }, 50);
@@ -39,11 +40,18 @@ function VisDisplay(props: Props): ReactElement {
   useEffect(() => {
     scheduleLoadingFlag(); // in case retrieving value takes too long (e.g. initial fetch of `SilxProvider`)
 
-    getValue(dataset.id).then((val) => {
+    ifMounted(getValue(dataset.id)).then((val) => {
       cancelLoadingFlag();
       mergeState({ value: val, loading: false });
     });
-  }, [cancelLoadingFlag, dataset, getValue, mergeState, scheduleLoadingFlag]);
+  }, [
+    cancelLoadingFlag,
+    dataset,
+    getValue,
+    mergeState,
+    ifMounted,
+    scheduleLoadingFlag,
+  ]);
 
   const { Component: VisComponent } = VIS_DEFS[activeVis];
 
--- a/src/h5web/providers/Provider.tsx
+++ b/src/h5web/providers/Provider.tsx
@@ -1,4 +1,5 @@
 import React, { ReactNode } from 'react';
+import { useAsync } from 'react-use';
 import { ProviderAPI, ProviderContext } from './context';
 
 interface Props {
@@ -9,15 +10,18 @@ interface Props {
 function Provider(props: Props): JSX.Element {
   const { api, children } = props;
 
-  if (!api) {
+  // Wait until metadata is fetched before rendering app
+  const { value: metadata } = useAsync(async () => api?.getMetadata(), [api]);
+
+  if (!api || !metadata) {
     return <></>;
   }
 
   return (
     <ProviderContext.Provider
       value={{
-        getDomain: api.getDomain.bind(api),
-        getMetadata: api.getMetadata.bind(api),
+        domain: api.domain,
+        metadata,
         getValue: api.getValue.bind(api),
       }}
     >
--- a/src/h5web/providers/context.ts
+++ b/src/h5web/providers/context.ts
@@ -2,19 +2,13 @@ import { createContext } from 'react';
 import type { HDF5Id, HDF5Value, HDF5Metadata } from './models';
 
 export abstract class ProviderAPI {
-  abstract getDomain: () => string;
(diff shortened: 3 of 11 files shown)


src/h5web/providers/context.ts
```diff
@@ -2,19 +2,13 @@ import { createContext } from 'react';
...
-export const ProviderContext = createContext<ProviderAPI>({
-  getDomain: missing as any, // eslint-disable-line @typescript-eslint/no-explicit-any
-  getMetadata: missing as any, // eslint-disable-line @typescript-eslint/no-explicit-any
-  getValue: missing as any, // eslint-disable-line @typescript-eslint/no-explicit-any
-});
+export const ProviderContext = createContext<{
+  domain: string;
+  metadata: HDF5Metadata;
+  getValue: ProviderAPI['getValue'];
+}>({} as any); // eslint-disable-line @typescript-eslint/no-explicit-any
```
axelboc, line 14: The app doesn't get rendered without metadata anymore, so providing an initial state in case users don't wrap the `App` in a provider seems less relevant.

src/h5web/providers/hsds/api.ts
```diff
@@ -159,25 +176,6 @@ export class HsdsApi implements ProviderAPI {
...
-    await this.processGroup(rootId);
-
-    this.metadata = {
-      root: rootId,
-      groups: this.groups,
-      datasets: this.datasets,
-      datatypes: this.datatypes,
-    };
-
-    return this.metadata;
```
axelboc, line 178: We no longer need to store the metadata in the class, since `getMetadata` is only called once when the app renders (or when the provider's API changes, but same thing) and is not exposed to the app itself.

src/h5web/dataset-visualizer/VisDisplay.tsx
```diff
@@ -39,11 +40,18 @@ function VisDisplay(props: Props): ReactElement {
   useEffect(() => {
     scheduleLoadingFlag(); // in case retrieving value takes too long (e.g. initial fetch of `SilxProvider`)
 
-    getValue(dataset.id).then((val) => {
+    whileMounted(getValue(dataset.id)).then((val) => {
       cancelLoadingFlag();
       mergeState({ value: val, loading: false });
```
axelboc, line 45: In case this isn't clear, thanks to `whileMounted`, the `then` callback isn't called if the component has been unmounted. I had a warning in the tests.

loichuder: Thanks for the clarification. Mind putting this in comment in the code ? :sweat_smile:

src/h5web/providers/Provider.tsx
```diff
@@ -9,15 +10,18 @@ interface Props {
 function Provider(props: Props): JSX.Element {
   const { api, children } = props;
 
-  if (!api) {
+  // Wait until metadata is fetched before rendering app
+  const { value: metadata } = useAsync(async () => api?.getMetadata(), [api]);
```
loichuder, line 14: The `useAsync` is to trigger a render at the promise resolution ?

src/h5web/providers/context.ts
```diff
@@ -2,19 +2,13 @@ import { createContext } from 'react';
...
-
-export const ProviderContext = createContext<ProviderAPI>({
-  getDomain: missing as any, // eslint-disable-line @typescript-eslint/no-explicit-any
-  getMetadata: missing as any, // eslint-disable-line @typescript-eslint/no-explicit-any
-  getValue: missing as any, // eslint-disable-line @typescript-eslint/no-explicit-any
-});
+export const ProviderContext = createContext<{
+  domain: string;
+  metadata: HDF5Metadata;
+  getValue: ProviderAPI['getValue'];
```
loichuder, line 13: Simply `ProviderAPI.getValue` doesn't work ?

loichuder (commented):
Could you explain why tests needed to be changed by adding `await` at several places ?

src/h5web/providers/Provider.tsx
axelboc: It's like a `useState` combined with a `useEffect` for making requests and storing the result. It also manages loading/error state.

src/h5web/providers/context.ts
axelboc: No, doesn't work like that with a type apparently. 🤷‍♂️

src/h5web/dataset-visualizer/VisDisplay.tsx
axelboc: Doesn't feel right adding a comment for this technical detail. I was hoping that `whileMounted` would provide enough information. If anything I would rather switch to a more imperative solution like:

```ts
getValue(dataset.id)).then((val) => {
    if (isUnmounted) {
      return;
    }
    
    cancelLoadingFlag();
    mergeState({ value: val, loading: false });
    }
  });
```

Can we think of a better name for `whileMounted` maybe? `ifMounted`, `ifStillMounted`, `unlessUnmounted`, ...

loichuder: `ifMounted` seems good :+1:

axelboc:
> Could you explain why tests needed to be changed by adding `await` at several places ?

The fact that the app is now rendered entirely asynchronously broke only one test: `switch between "display" and "inspect" modes`. This test was expecting the `Inpsect` tab to be rendered right away. Replacing `getByRole` with `await findByRole` fixed it.

The other changes I made to the tests were to fix warnings. I think those may have appeared back when we moved `mergeState` into `VisDisplay`. I looked at each test one by one with `.only` to find the culprits. One warning was caused by a state change after unmount in `VisDisplay`, which I fixed with `usePromise`. The other warnings read "An update to VisDisplay inside a test was not wrapped in act(...).", which meant that the promise callback in `VisDisplay` was being called after a test had completed -- I fixed those with async utilities like `waitFor` and `findBy`.

loichuder:
Ok, that was I thought: the other changes were not related to the PR.

loichuder (approved):

axelboc:
I've just pushed some code in `setupTests.js` to fail tests that log and error or warning to the browser console. Hopefully we won't miss any such regressions in the tests any more.

PR #251, opened 2020-10-16
 LineVis and HeatmapVis now accept arrays of values for x and y

loichuder:
... and falls back to indices when arrays are not supplied (previous behaviour)

<s>We may want to discuss a change that may be seen as a regression: the tooltip on the LineVis now shows the X/Y values of the closest point to the left and not of the closest point.
Ex: Previously, moving the mouse to X=28.75 showed the X/Y values for X=29. Now it shows the X/Y values for X=28.

The reason is that the previous behaviour can not be easily reproduced when X represents arbitrary values and not indices.</s>

**EDIT**: Actually, I found a way to do this ! See the end of https://github.com/silx-kit/h5web/pull/251#discussion_r506201340

I will fix the `AxisSystem` story after the first round of discussions.

### DIFF
--- a/src/h5web/visualizations/heatmap/HeatmapVis.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { format } from 'd3-format';
 import type ndarray from 'ndarray';
+import { range } from 'd3-array';
 import styles from './HeatmapVis.module.css';
 import ColorBar from './ColorBar';
 import Mesh from './Mesh';
@@ -10,6 +11,7 @@ import VisCanvas from '../shared/VisCanvas';
 import { getDims } from './utils';
 import { Domain, ScaleType } from '../shared/models';
 import type { ColorMap } from './models';
+import { getDomain, getValueToIndexScale } from '../shared/utils';
 
 interface Props {
   dataArray: ndarray<number>;
@@ -19,6 +21,8 @@ interface Props {
   keepAspectRatio?: boolean;
   showGrid?: boolean;
   showLoader?: boolean;
+  abscissas?: number[];
+  ordinates?: number[];
 }
 
 function HeatmapVis(props: Props): JSX.Element {
@@ -35,19 +39,66 @@ function HeatmapVis(props: Props): JSX.Element {
   const { rows, cols } = getDims(dataArray);
   const aspectRatio = keepAspectRatio ? cols / rows : undefined; // width / height <=> cols / rows
 
+  const { abscissas = range(cols + 1), ordinates = range(rows + 1) } = props;
+
+  if (abscissas.length !== cols + 1) {
+    throw new Error(
+      `Abscissas size (${
+        abscissas.length
+      }) does not match data length along X (${cols + 1})`
+    );
+  }
+  if (ordinates.length !== rows + 1) {
+    throw new Error(
+      `Ordinate size (${
+        ordinates.length
+      }) does not match data length along Y (${rows + 1})`
+    );
+  }
+
+  const abscissaToIndex = getValueToIndexScale(abscissas);
+
+  const abscissaDomain = useMemo(() => {
+    return getDomain(abscissas);
+  }, [abscissas]);
+  if (abscissaDomain === undefined) {
+    throw new Error(`Abscissas (${abscissas}) have an empty domain`);
+  }
+
+  const ordinateToIndex = getValueToIndexScale(ordinates);
+
+  const ordinateDomain = useMemo(() => {
+    return getDomain(ordinates);
+  }, [ordinates]);
+  if (ordinateDomain === undefined) {
+    throw new Error(`Ordinates (${ordinates}) have an empty domain`);
+  }
+
   return (
     <div className={styles.root}>
       <VisCanvas
-        abscissaConfig={{ indexDomain: [0, cols], showGrid }}
-        ordinateConfig={{ indexDomain: [0, rows], showGrid }}
+        abscissaConfig={{
+          domain: abscissaDomain,
+          showGrid,
+          isIndexAxis: !abscissas,
+        }}
+        ordinateConfig={{
+          domain: ordinateDomain,
+          showGrid,
+          isIndexAxis: !ordinates,
+        }}
         aspectRatio={aspectRatio}
       >
         <TooltipMesh
-          formatIndex={([x, y]) => `x=${Math.floor(x)}, y=${Math.floor(y)}`}
+          formatIndex={([x, y]) => {
+            return `x=${abscissas[abscissaToIndex(x)]}, y=${
+              ordinates[ordinateToIndex(y)]
+            }`;
+          }}
           formatValue={([x, y]) => {
-            return x < cols && y < rows
-              ? format('.3')(dataArray.get(Math.floor(y), Math.floor(x)))
-              : undefined;
+            return format('.3')(
+              dataArray.get(ordinateToIndex(y), abscissaToIndex(x))
+            );
           }}
           guides="both"
         />
--- a/src/h5web/visualizations/line/DataCurve.tsx
+++ b/src/h5web/visualizations/line/DataCurve.tsx
@@ -10,14 +10,16 @@ import { getAxisScale } from '../shared/utils';
 const DEFAULT_COLOR = '#1b998b';
 
 interface Props {
-  values: number[];
+  abscissas: number[];
+  ordinates: number[];
   color?: string;
   curveType?: CurveType;
 }
 
 function DataCurve(props: Props): JSX.Element {
   const {
-    values,
+    abscissas,
+    ordinates,
     color = DEFAULT_COLOR,
     curveType = CurveType.LineOnly,
   } = props;
@@ -30,10 +32,10 @@ function DataCurve(props: Props): JSX.Element {
     const abscissaScale = getAxisScale(abscissaInfo, width);
     const ordinateScale = getAxisScale(ordinateInfo, height);
 
-    const points = values.map((val, index) => {
+    const points = ordinates.map((val, index) => {
       const ordinate = ordinateScale(val);
       return new Vector3(
-        abscissaScale(index),
+        abscissaScale(abscissas[index]),
         // This is to avoid a three.js warning when ordinateScale(val) is Infinity
         Number.isFinite(ordinate) ? ordinate : 0,
         // Move NaN/Infinity out of the camera FOV (negative val for logScale).
@@ -45,7 +47,7 @@ function DataCurve(props: Props): JSX.Element {
     const geometry = new BufferGeometry();
     geometry.setFromPoints(points);
     return geometry;
-  }, [abscissaInfo, camera, height, ordinateInfo, values, width]);
+  }, [abscissaInfo, abscissas, camera, height, ordinateInfo, ordinates, width]);
 
   const showLine = curveType !== CurveType.GlyphsOnly;
   const showGlyphs = curveType !== CurveType.LineOnly;
--- a/src/h5web/visualizations/line/LineVis.tsx
+++ b/src/h5web/visualizations/line/LineVis.tsx
@@ -1,23 +1,25 @@
 import React, { ReactElement, useMemo } from 'react';
 import { format } from 'd3-format';
 import type ndarray from 'ndarray';
+import { range } from 'd3-array';
 import styles from './LineVis.module.css';
 import DataCurve from './DataCurve';
 import VisCanvas from '../shared/VisCanvas';
 import PanZoomMesh from '../shared/PanZoomMesh';
 import TooltipMesh from '../shared/TooltipMesh';
-import { extendDomain } from '../shared/utils';
 import { ScaleType, Domain } from '../shared/models';
 import { CurveType } from './models';
+import { getValueToIndexScale, getDomain, extendDomain } from '../shared/utils';
 
 const DEFAULT_DOMAIN: Domain = [0.1, 1];
 
 interface Props {
   dataArray: ndarray<number>;
-  domain: Domain | undefined;
+  domain?: Domain;
   scaleType?: ScaleType;
   curveType?: CurveType;
   showGrid?: boolean;
+  abscissas?: number[];
 }
 
 function LineVis(props: Props): ReactElement {
@@ -27,9 +29,26 @@ function LineVis(props: Props): ReactElement {
     curveType = CurveType.LineOnly,
     showGrid = true,
     scaleType = ScaleType.Linear,
+    abscissas = range(dataArray.size),
   } = props;
 
-  const indexDomain = extendDomain([0, dataArray.size - 1], 0.01);
+  if (abscissas.length !== dataArray.size) {
+    throw new Error(
+      `Abscissas size (${abscissas.length}) does not match data length (${dataArray.size})`
+    );
+  }
+
+  const abscissaToIndex = getValueToIndexScale(abscissas, true);
+
+  const abscissaDomain = useMemo(() => {
+    const rawDomain = getDomain(abscissas);
+    return rawDomain && extendDomain(rawDomain, 0.01);
+  }, [abscissas]);
+
+  if (abscissaDomain === undefined) {
+    throw new Error(`Abscissas (${abscissas}) have an empty domain`);
+  }
+
   const dataDomain = useMemo(() => {
     return domain
       ? extendDomain(domain, 0.05, scaleType === ScaleType.Log)
@@ -39,24 +58,33 @@ function LineVis(props: Props): ReactElement {
   return (
     <div className={styles.root}>
       <VisCanvas
-        abscissaConfig={{ indexDomain, showGrid }}
-        ordinateConfig={{ dataDomain, showGrid, scaleType }}
+        abscissaConfig={{
+          domain: abscissaDomain,
+          showGrid,
+          isIndexAxis: !abscissas,
+        }}
+        ordinateConfig={{
+          domain: dataDomain,
+          showGrid,
+          scaleType,
+        }}
       >
         <TooltipMesh
-          formatIndex={([x]) => `x=${Math.round(x)}`}
+          formatIndex={([x]) =>
+            `x=${format('0')(abscissas[abscissaToIndex(x)])}`
+          }
           formatValue={([x]) => {
-            const value = dataArray.get(Math.round(x));
+            const value = dataArray.get(abscissaToIndex(x));
             return value !== undefined ? format('.3f')(value) : undefined;
           }}
           guides="vertical"
         />
         <PanZoomMesh />
-        {dataDomain && (
-          <DataCurve
-            curveType={curveType}
-            values={dataArray.data as number[]}
-          />
-        )}
+        <DataCurve
+          curveType={curveType}
+          abscissas={abscissas}
+          ordinates={dataArray.data as number[]}
+        />
       </VisCanvas>
     </div>
   );
--- a/src/h5web/visualizations/shared/AxisSystemProvider.tsx
+++ b/src/h5web/visualizations/shared/AxisSystemProvider.tsx
@@ -1,7 +1,6 @@
 import React, { ReactElement, ReactNode, createContext } from 'react';
 import { scaleLinear } from 'd3-scale';
 import { AxisConfig, AxisInfo, ScaleType, SCALE_FUNCTIONS } from './models';
-import { isIndexAxisConfig } from './utils';
 
 export interface AxisInfos {
   abscissaInfo: AxisInfo;
@@ -11,24 +10,19 @@ export interface AxisInfos {
 export const AxisSystemContext = createContext<AxisInfos>({} as AxisInfos);
 
 function getAxisInfo(config: AxisConfig): AxisInfo {
-  if (isIndexAxisConfig(config)) {
-    const { indexDomain, showGrid = false } = config;
-    return {
-      isIndexAxis: true,
-      scaleFn: scaleLinear,
-      domain: indexDomain,
-      scaleType: ScaleType.Linear,
-      showGrid,
-    };
-  }
+  const {
+    domain,
+    scaleType = ScaleType.Linear,
+    isIndexAxis,
+    showGrid = false,
+  } = config;
 
-  const { dataDomain, scaleType = ScaleType.Linear, showGrid = false } = config;
   return {
-    isIndexAxis: false,
-    scaleFn: SCALE_FUNCTIONS[scaleType],
-    domain: dataDomain,
-    scaleType,
+    onlyIntegers: isIndexAxis,
+    scaleFn: isIndexAxis ? scaleLinear : SCALE_FUNCTIONS[scaleType],
+    scaleType: isIndexAxis ? ScaleType.Linear : scaleType,
     showGrid,
+    domain,
   };
 }
 
--- a/src/h5web/visualizations/shared/models.ts
+++ b/src/h5web/visualizations/shared/models.ts
@@ -23,19 +23,12 @@ export type Size = { width: number; height: number };
 
 export type Domain = [number, number];
 
-export interface IndexAxisConfig {
-  indexDomain: Domain;
-  showGrid?: boolean;
-  scaleType?: never; // invalid
-}
-
-export interface DataAxisConfig {
-  dataDomain: Domain;
+export interface AxisConfig {
+  isIndexAxis?: boolean;
+  domain: Domain;
   showGrid?: boolean;
   scaleType?: ScaleType;
 }
-
-export type AxisConfig = IndexAxisConfig | DataAxisConfig;
 
 export type ScaleFn = typeof scaleLinear | typeof scaleLog | typeof scaleSymlog;
 
@@ -45,7 +38,7 @@ export type AxisScale =
   | ScaleSymLog<number, number>;
 
 export interface AxisInfo {
-  isIndexAxis: boolean;
+  onlyIntegers?: boolean;
   scaleFn: ScaleFn;
   domain: Domain;
   scaleType: ScaleType;
--- a/src/h5web/visualizations/shared/utils.ts
+++ b/src/h5web/visualizations/shared/utils.ts
@@ -1,15 +1,7 @@
-import { scaleLinear } from 'd3-scale';
-import { tickStep } from 'd3-array';
+import { scaleLinear, scaleThreshold, ScaleThreshold } from 'd3-scale';
+import { tickStep, range } from 'd3-array';
 import { format } from 'd3-format';
-import {
-  Size,
-  Domain,
-  AxisScale,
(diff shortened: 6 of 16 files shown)


src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -30,24 +36,66 @@ function HeatmapVis(props: Props): JSX.Element {
...
+          formatIndex={([x, y]) => {
+            return `x=${abscissaToIndex(x)}, y=${ordinateToIndex(y)}`;
+          }}
           formatValue={([x, y]) => {
-            return x < cols && y < rows
-              ? format('.3')(dataArray.get(Math.floor(y), Math.floor(x)))
-              : undefined;
+            return format('.3')(
+              dataArray.get(ordinateToIndex(y), abscissaToIndex(x))
+            );
```
loichuder, line 98: Now that domains of axis can be arbitrary values, the tooltip coordinates are no longer indices. To retrieve the tooltip value, I need a way to get the index from the value and this is done with a threshold scale that maps values to indices.

See the comment below for more details on the scale construction.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -30,24 +36,66 @@ function HeatmapVis(props: Props): JSX.Element {
...
+    );
+  }
+
+  if (ordinateValues.length !== rows + 1) {
+    return (
+      <p>{`Ordinate size (${
+        ordinateValues.length
+      }) does not match data length along Y (${rows + 1})`}</p>
+    );
+  }
```
loichuder, line 74: This check needs to be below all computations due to the `useMemo`.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -30,24 +36,66 @@ function HeatmapVis(props: Props): JSX.Element {
...
+  const abscissaDomain = useMemo(() => {
+    return getDataDomain(abscissaValues) ?? DEFAULT_DOMAIN;
+  }, [abscissaValues]);
+
+  const ordinateIndices = range(rows + 1);
+  const ordinateValues = ordinates || ordinateIndices;
+  const ordinateToIndex = getValueToIndexScale(ordinateValues);
+  const ordinateDomain = useMemo(() => {
+    return getDataDomain(ordinateValues) ?? DEFAULT_DOMAIN;
+  }, [ordinateValues]);
```
loichuder, line 58: Several similar operations that could be factorized in future steps.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -30,24 +36,66 @@ function HeatmapVis(props: Props): JSX.Element {
...
+    );
+  }
+
   return (
     <div className={styles.root}>
       <VisCanvas
-        abscissaConfig={{ indexDomain: [0, cols], showGrid }}
-        ordinateConfig={{ indexDomain: [0, rows], showGrid }}
+        abscissaConfig={{
+          domain: abscissaDomain,
```
loichuder, line 80: The `domain` can now be something else than indices. That means that the scales built by `getAxisScale` that map axis coordinates to positions in the ThreeJS Scene now deal with arbitrary abscissa/ordinates.

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -30,10 +32,10 @@ function DataCurve(props: Props): JSX.Element {
     const abscissaScale = getAxisScale(abscissaInfo, width);
     const ordinateScale = getAxisScale(ordinateInfo, height);
 
-    const points = values.map((val, index) => {
+    const points = ordinates.map((val, index) => {
       const ordinate = ordinateScale(val);
       return new Vector3(
-        abscissaScale(index),
+        abscissaScale(abscissas[index]),
```
loichuder, line 38: As explained in https://github.com/silx-kit/h5web/pull/251/files#r506183326, the axis coordinates are now `abscissas` and not always indices. This means the curve construction must be matched.

src/h5web/visualizations/shared/models.ts
```diff
@@ -23,20 +23,13 @@ export type Size = { width: number; height: number };
...
-}
-
-export interface DataAxisConfig {
-  dataDomain: Domain;
+export interface AxisConfig {
+  isIndexAxis?: boolean;
+  domain: Domain;
   showGrid?: boolean;
   scaleType?: ScaleType;
 }
```
loichuder, line 31: `IndexAxisConfig` and `DataAxisConfig` were merged in a single general interface. I only kept a boolean `isIndexAxis` that will force the axis to display only integers (see the updated `AxisInfo`).

src/h5web/visualizations/shared/models.ts
```diff
@@ -45,7 +38,7 @@ export type AxisScale =
   | ScaleSymLog<number, number>;
 
 export interface AxisInfo {
-  isIndexAxis: boolean;
+  onlyIntegers?: boolean;
```
loichuder, line 41: `isIndexAxis` is only used to check whether ticks should only display integers so I think this renaming made sense.

src/h5web/visualizations/shared/AxisSystemProvider.tsx
```diff
@@ -11,24 +10,19 @@ export interface AxisInfos {
...
-    isIndexAxis: false,
-    scaleFn: SCALE_FUNCTIONS[scaleType],
-    domain: dataDomain,
-    scaleType,
+    onlyIntegers: isIndexAxis,
+    scaleFn: isIndexAxis ? scaleLinear : SCALE_FUNCTIONS[scaleType],
+    scaleType: isIndexAxis ? ScaleType.Linear : scaleType,
     showGrid,
+    domain,
   };
```
loichuder, line 26: Note that `AxisInfo` and `AxisConfig` are now extremely similar: there is only `scaleFn` that can be removed as it depends on `scaleType`.

I already pointed out the fact that it was misleading to have two interfaces `Info` and `Config` so I think we can merge them in the near future.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -93,10 +85,16 @@ export function extendDomain(
...
+export function getValueToIndexScale(
+  values: number[]
+): ScaleThreshold<number, number> {
+  const indices = range(values.length).map((i) => i);
+
+  const valueToIndexScale = scaleThreshold();
+  valueToIndexScale.domain(values.slice(1)); // First threshold (going to 0 to 1) should be for the second value. Scaling the first value should return at 0.
+  valueToIndexScale.range(indices);
+
+  return valueToIndexScale;
```
loichuder, line 97: A [scaleThreshold](https://github.com/d3/d3-scale#threshold-scales) separates a `range` into arbitrary subsets that are given by the `domain`.

About the `slice`: imagine that we have 5 abscissas `[-5, -2.5, 0, 2.5, 5]` that should be mapped to `[0, 1, 2, 3, 4]`. Using naively these arrays will lead to
image

which is not what we want as `scale(x)` with `x` in `[-5, -2.5]` will return `1`. This is to be expected as the domain gives the thresholds: the first threshold being `-5`, the switch from `0` to `1` happens at this value`.

We want instead the switch from `0` to `1` to happen at the **second value** `-2.5`  hence the `slice`:
image

We might be able to tune this to change the tooltip behaviour in the `LineVis` but we would need to make assumptions on the abscissas array (namely it is regularly spaced and increasing) while the `scaleThreshold` stay generic. This is a point to discuss as one could argue that abscissas **should always** be regularly spaced and increasing !

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -27,36 +33,62 @@ function LineVis(props: Props): ReactElement {
...
       <VisCanvas
-        abscissaConfig={{ indexDomain, showGrid }}
-        ordinateConfig={{ dataDomain, showGrid, scaleType }}
+        abscissaConfig={{
+          domain: abscissaDomain,
+          showGrid,
+          isIndexAxis: !abscissas,
+        }}
+        ordinateConfig={{
+          domain: domain || ordinateDomain,
```
loichuder, line 70: Note that `LineVis` still accepts a `domain` prop that will take precedence over the `ordinateDomain`. This is useful for the `autoScale` behaviour where the `domain` is inferred from the whole dataset rather than from the slice.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -30,24 +36,66 @@ function HeatmapVis(props: Props): JSX.Element {
...
+  }, [abscissaValues]);
+
+  const ordinateIndices = range(rows + 1);
+  const ordinateValues = ordinates || ordinateIndices;
+  const ordinateToIndex = getValueToIndexScale(ordinateValues);
+  const ordinateDomain = useMemo(() => {
+    return getDataDomain(ordinateValues) ?? DEFAULT_DOMAIN;
+  }, [ordinateValues]);
+
+  if (abscissaValues.length !== cols + 1) {
```
loichuder, line 60: `abscissaValues` and `ordinateValues` give the positions of the pixels' edges. For `N` pixels, we have `N+1` edges.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -27,36 +33,66 @@ function LineVis(props: Props): ReactElement {
...
-  const dataDomain = useMemo(() => {
-    return domain
-      ? extendDomain(domain, 0.05, scaleType === ScaleType.Log)
+  const abscissaValues = abscissas || range(dataArray.size);
+  const abscissaToIndex = getValueToIndexScale(
+    abscissaValues.map(
+      (x, i) =>
+        abscissaValues[i - 1] + (abscissaValues[i] - abscissaValues[i - 1]) / 2
+    )
+  ); // Shifting index change threshold to midpoints
```
loichuder, line 45: Explained in greater details at https://github.com/silx-kit/h5web/pull/251#discussion_r506201340

The `map` operation gives `NaN` for `i=0` but this is irrelevant as the first value of the array is discarded when constructing the `ScaleThreshold`.

axelboc: Great solution, but seeing how it works for the Heatmap without any changes, I wonder... could we do something like this?

```js
const abscissaToIndex = getValueToIndexScale([...abscissaValues, NaN]);
```

Adding a fake value at the end of the array would shift all the thresholds like on the heatmap, wouldn't it? The last number would not appear on the axis since it's not part of `abscissaValues`, and it wouldn't appear in the tooltip since I think the last pixel of the plot would match the last real abscissa value.

src/h5web/visualizations/line/LineVis.tsx
axelboc: Is this change required for now? I'm not sure I agree with it, so I'd like to discuss it further. Note that, as is, it changes the behaviour, since the domain that is passed when auto-scale is off is no longer extended.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -27,36 +33,66 @@ function LineVis(props: Props): ReactElement {
     curveType = CurveType.LineOnly,
     showGrid = true,
     scaleType = ScaleType.Linear,
+    abscissas,
```
axelboc, line 36: You could assign a default value to this argument instead of declaring a new `abscissaValues` constant.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -30,24 +36,66 @@ function HeatmapVis(props: Props): JSX.Element {
...
+    ordinates,
   } = props;
 
   const { rows, cols } = getDims(dataArray);
   const aspectRatio = keepAspectRatio ? cols / rows : undefined; // width / height <=> cols / rows
 
+  const abscissaValues = abscissas || range(cols + 1);
+  const abscissaToIndex = getValueToIndexScale(abscissaValues);
+  const abscissaDomain = useMemo(() => {
+    return getDataDomain(abscissaValues) ?? DEFAULT_DOMAIN;
```
axelboc, line 49: Is the `DEFAULT_DOMAIN` required? If the domain is `undefined`, then the axis won't appear, which gives a good indication to the user that something isn't right. I think the default domain is only useful for the color map, when auto-scale is on.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -27,36 +33,66 @@ function LineVis(props: Props): ReactElement {
...
         <TooltipMesh
-          formatIndex={([x]) => `x=${Math.round(x)}`}
+          formatIndex={([x]) =>
+            `x=${format('0')(abscissaValues[abscissaToIndex(x)])}`
+          }
           formatValue={([x]) => {
-            const value = dataArray.get(Math.round(x));
+            const index = abscissaToIndex(x);
+            const value = ordinateValues[index];
             return value !== undefined ? format('.3f')(value) : undefined;
```
axelboc, line 86: I wouldn't worry with the `index` constant here. Also, can the `value` still be `undefined`?

src/h5web/visualizations/shared/models.ts
axelboc: Hehe, random thought, but in the end, showing only integers on the heatmap axes is equivalent to not showing interpolated ticks when passing an array of abscissa or ordinate values.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -93,10 +85,16 @@ export function extendDomain(
   return [min - extension, max + extension];
 }
 
-export function isIndexAxisConfig(
-  config: AxisConfig
-): config is IndexAxisConfig {
-  return 'indexDomain' in config;
+export function getValueToIndexScale(
```
axelboc, line 88: Nice! I would just rename this `getValuesToIndexScale`.

src/stories/HeatmapVis.stories.tsx
```diff
@@ -103,6 +103,19 @@ LiveDataWithoutLoader.args = {
...
+
+CustomCoordinates.args = {
+  dataArray,
+  domain,
+  abscissas: Array(dataArray.shape[1] + 1)
+    .fill(0)
+    .map((x, i) => 100 + 10 * i),
+  ordinates: Array(dataArray.shape[0] + 1)
+    .fill(0)
+    .map((x, i) => -5 + 0.5 * i),
```
axelboc, line 116: `lodash`'s `range` function accepts a step https://lodash.com/docs/4.17.15#range

src/stories/LineVis.stories.tsx
```diff
@@ -69,18 +62,23 @@ export const SymLogScale = Template.bind({});
...
   dataArray,
-  domain,
   showGrid: false,
 };
 
+export const CustomAbscissas = Template.bind({});
+
+CustomAbscissas.args = {
+  dataArray,
+  abscissas: dataArray.data as number[], // plotting x = y
```
axelboc, line 79: Might be worth passing a custom range to avoid any confusion.

axelboc (commented):
Nicely done! Just a couple of discussion points.

src/h5web/visualizations/shared/models.ts
loichuder: Well spotted ! We should indeed keep this in mind.

src/h5web/visualizations/line/LineVis.tsx
loichuder: The change is not **required** per se but not making it breaks the Autoscale feature as the domain will be always inferred from the data passed to the LineVis.

src/h5web/visualizations/line/LineVis.tsx
loichuder: This won't work and I explain below.

The argument of `getValueToIndexScale` is directly the domain of the `scaleThreshold`. This domain contains the values of the threshold, period. There is no shift, the scale separates the `range` at the values given in `domain` (that can be arbitrary).

It works without tricks in the `Heatmap` as we want the tooltip there to switch between two values when crossing the edges that are the abscissa values.

In the `LineVis` however, we want to trigger the switch at the midpoint between two neighbouring abscissas. As `scaleThreshold` takes directly the values where the switch should appear, there is no way around creating the array of midpoints. 

To have the behaviour you suggest, we could use related scales `scaleQuantize` and `scaleQuantile`. I tried for some time but could not really get what I wanted (or it would involve circumvolutions). I settled in the end for `scaleThreshold` where I set the thresholds explicitly that I preferred over twisting around the other scales.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
loichuder: I would agree but this means that to satisfy TS, we would need to accept `undefined` for `domain` all the way down to `Axis`...

src/h5web/visualizations/line/LineVis.tsx
axelboc: Right, sorry I was picturing this working with `scaleQuantile`, but what you're saying makes sense. I've just played with the scales a bit to try to understand them a bit better and `scaleThreshold` does make the most sense.

src/h5web/visualizations/line/LineVis.tsx
loichuder: :+1: for `index`

As for `value`, given that we extend the domain in `x`, we could imagine having a value under the mouse that doesn't correspond to an index, leading to `undefined` for `value`.

src/stories/HeatmapVis.stories.tsx
loichuder: True, but that means I have to compute the `end`  to have the number of elements I need :smile:

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: Right, and we didn't use to because we were always initializing the domain to something... Would it make sense to return an error, then? We could check both domains at once to save time.

src/h5web/visualizations/line/LineVis.tsx
loichuder: In fact, this is a fix to an unneeded change I made that puts the computation of the domain in the LineVis. I will revert all of this.

src/h5web/visualizations/line/LineVis.tsx
axelboc, line 45: Note that you could use `x` instead of `abscissaValues[i]`.

One thing that I find a bit obscure is the fact that the shifting is done in the caller, but the slicing is done in `getValueToIndexScale`. The function does a bit of the work, but not all of it, creating an implicit coupling with the caller.

I feel like the shifting should be done inside `getValueToIndexScale`, perhaps with a boolean argument like `shiftThresholds` and a comment explaining when to use it.

src/h5web/visualizations/line/LineVis.tsx
axelboc: Ah right, forgot about the domain extension.

src/stories/HeatmapVis.stories.tsx
axelboc: Well, you do anyway when you initialize the array, don't you? :D

axelboc: Ah no, because the end has to be computed with the step... Well that sucks.

loichuder: Shame there is no equivalent to [linspace](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html)

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -30,24 +36,66 @@ function HeatmapVis(props: Props): JSX.Element {
...
+    );
+  }
+
+  if (ordinateValues.length !== rows + 1) {
+    return (
+      <p>{`Ordinate size (${
+        ordinateValues.length
+      }) does not match data length along Y (${rows + 1})`}</p>
+    );
+  }
```
axelboc, line 72: Just another thought: would it be better to throw an error? This is basically an assertion to make sure things don't break later on, so to me, we'd be better off breaking hard.... especially since I'm currently adding an error boundary to handle errors in visualization components! If we don't throw, then the component tree has no way of knowing that an error ocurred.

loichuder: Sure !

src/h5web/visualizations/line/LineVis.tsx
loichuder: > Note that you could use `x` instead of `abscissaValues[i]`.

I found the syntax `abscissaValues[i] - abscissaValues[i - 1]` clearer than `x - abscissaValues[i - 1]`

> One thing that I find a bit obscure is the fact that the shifting is done in the caller, but the slicing is done in `getValueToIndexScale`. The function does a bit of the work, but not all of it, creating an implicit coupling with the caller.
> 
> I feel like the shifting should be done inside `getValueToIndexScale`, perhaps with a boolean argument like `shiftThresholds` and a comment explaining when to use it.

Yes I see your point, I will try to implement this

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -30,24 +34,74 @@ function HeatmapVis(props: Props): JSX.Element {
...
+  const abscissaValues = abscissas || range(cols + 1);
+  const abscissaToIndex = getValueToIndexScale(abscissaValues);
+  const abscissaDomain = useMemo(() => {
+    return getDataDomain(abscissaValues);
+  }, [abscissaValues]);
+
+  const ordinateValues = ordinates || range(rows + 1);
+  const ordinateToIndex = getValueToIndexScale(ordinateValues);
+  const ordinateDomain = useMemo(() => {
+    return getDataDomain(ordinateValues);
```
axelboc, line 53: Now I realise that this leads to a slight loss of performance. When `ordinates` (or `abscissas`) is not provided, we loop through the array to find the domain even though we know for a fact it's going to be `[0, rows + 1]`. Should we keep a ternary to avoid calling `getDataDomain`?

Also, should we rename `getDataDomain` and `useDataDomain` to just `getDomain` and `useDomain`?

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -30,24 +34,74 @@ function HeatmapVis(props: Props): JSX.Element {
     keepAspectRatio = true,
     showGrid = false,
     showLoader = true,
+    abscissas,
+    ordinates,
```
axelboc, line 38: You can move calls to `range` here like in `LineVis`.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -27,36 +33,62 @@ function LineVis(props: Props): ReactElement {
...
     return domain
       ? extendDomain(domain, 0.05, scaleType === ScaleType.Log)
       : DEFAULT_DOMAIN;
   }, [scaleType, domain]);
 
+  if (abscissas.length !== dataArray.size) {
+    throw new Error(
+      `Abscissas size (${abscissas.length}) does not match data length (${dataArray.size})`
+    );
+  }
```
axelboc, line 56: A hook can't come after an early return, but I think it can come after a throw. I've just come across this use case in my refactoring and I'm not seeing any linting errors so far. When a component "instance" throws, React probably unmounts it and resets its hook calls history. So you can move this check higher up in the component.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -27,36 +33,62 @@ function LineVis(props: Props): ReactElement {
...
     scaleType = ScaleType.Linear,
+    abscissas = range(dataArray.size),
   } = props;
 
-  const indexDomain = extendDomain([0, dataArray.size - 1], 0.01);
+  const abscissaToIndex = getValueToIndexScale(abscissas, true);
+
+  const abscissaDomain = useMemo(() => {
+    const rawDomain = getDataDomain(abscissas);
+    return rawDomain ? extendDomain(rawDomain, 0.01) : undefined;
```
axelboc, line 43: Since we're checking `abscissaDomain` for `undefined` now, it may make sense for `extendDomain` to support an `undefined` input/output domain: `return extendDomain(getDataDomain(abscissas, 0.01))`

src/h5web/visualizations/shared/utils.ts
```diff
@@ -93,10 +85,21 @@ export function extendDomain(
...
+  switchAtMidpoints?: boolean
+): ScaleThreshold<number, number> {
+  const indices = range(values.length);
+
+  const thresholds = switchAtMidpoints
+    ? values.map((x, i) => values[i - 1] + (values[i] - values[i - 1]) / 2) // Shift the thresholds for the switch from i-1 to i to happen between values[i-1] and values[i]
+    : values; // Else, the switch from i-1 to i will happen at values[i]
+
+  const valueToIndexScale = scaleThreshold();
+  valueToIndexScale.domain(thresholds.slice(1)); // First threshold (going to 0 to 1) should be for the second value. Scaling the first value should return at 0.
```
axelboc, line 99: `going from 0 to 1`

axelboc (approved):
Awesome! 🍺

Only minor details left, feel free to merge whenever.

src/h5web/visualizations/line/LineVis.tsx
loichuder: A good idea on the paper but is it really in practice ? We will need then to check for `undefined` on every use of `extendDomain`...

src/h5web/visualizations/heatmap/HeatmapVis.tsx
loichuder: Can I really ? I need `rows` and `cols` for that, that are returned from `getDims`

src/h5web/visualizations/heatmap/HeatmapVis.tsx
loichuder: > Now I realise that this leads to a slight loss of performance. When `ordinates` (or `abscissas`) is not provided, we loop through the array to find the domain even though we know for a fact it's going to be `[0, rows + 1]`. Should we keep a ternary to avoid calling `getDataDomain`?

I was a bit concerned by this as well. I will look into it.

> Also, should we rename `getDataDomain` and `useDataDomain` to just `getDomain` and `useDomain`?

It would make more sense to me

src/h5web/visualizations/line/LineVis.tsx
axelboc: Ah yeah, forgot it's used in `DomainSlider`... Oh well, you can at least simplify with `rawDomain && extendDomain(...)`

src/h5web/visualizations/heatmap/HeatmapVis.tsx
axelboc: Hmm, you can have a second destructuring of `props` after the call to `getDims`. I think it's still nicer than to have an extra `abscissaValues` constant.

PR #252, opened 2020-10-19
Refactor visualizer and vis defs, and implement vis containers

axelboc:
Where to start... So basically:

1. `DatasetVisualizer` becomes the new root `Visualizer` (`activeVis` state, `VisSelector`, toolbar, etc.)
1. `NexusVisualizer` is removed and NeXus visualizations are added as new visualizations in the `Vis` enum and in `VIS_DEFS`.
1. The `Visualizer` no longer checks the type of the selected entity; groups and datasets are treated the same way -- it is up to each visualization to check the type of the entity and to say whether it supports it or not. As a result, the visualization definitions now have a `supportsEntity` function instead of `supportsDataset`. 
1. The visualization definitions now point to "container" components that receive a single prop: an entity.
1. `VisDisplay` is removed. It is now up to each vis container to:
  - make sure it's receiving the right kind of entity (dataset vs group) -- basically, for type checking purposes and to make sure the visualization's `supportsEntity` function works properly (or that the container is used properly on its own);
  - retrieve the value of the entity asynchronously (with [`use-async-resource`](https://github.com/andreiduca/use-async-resource));
  - display the `DimensionMapper` and manage the `mapperState` if appropriate.
1. `Visualizer` has a loading and error boundary (`use-async-resource` supports Suspense), so any error that occurs in a vis container or component is cleanly caught and displayed, and the loader appears while fetching (after a min of 50ms).

Once this PR is merged, I'll do a big clean up. The `dataset-visualizer` should probably be renamed `visualizer` and the `Visualizer` moved into it. Also, I'll need to review the location of a lot of the utilities, hooks and models.


src/h5web/visualizations/index.tsx
```diff
@@ -1,101 +1,152 @@
...
+  [Vis.NxSpectrum]: {
+    Icon: FiActivity,
+    Toolbar: LineToolbar,
+    Component: NxSpectrumContainer,
+    supportsEntity: (entity, metadata) => {
+      if (!isGroup(entity) || !isNxDataGroup(entity)) {
+        return false;
+      }
+
+      const dataset = getSignalDataset(entity, metadata);
```
axelboc, line 108: I need the `metadata` object to find the signal dataset, which is why the signature of `supportsEntity` now has a `metadata` argument.

src/h5web/visualizations/nexus/NxImageContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+function NxImageContainer(props: VisContainerProps): ReactElement {
+  const { entity } = props;
+  assertGroup(entity);
+
+  const { metadata } = useContext(ProviderContext);
+  const dataset = getSignalDataset(entity, metadata);
+
+  if (!dataset) {
+    throw new Error('Signal dataset not found');
+  }
```
axelboc, line 22: I look for the signal dataset again here (I already did once in `supportsEntity`), but repeating the call to `getSignalDataset` means we don't have any coupling between the code that checks support and the code that actually prepares the visualization. If an error is thrown here it is caught by the new error boundary.

src/h5web/visualizations/shared/hooks.ts
```diff
@@ -1,12 +1,25 @@
...
+} from '../../providers/models';
 import type { DimensionMapping } from '../../dataset-visualizer/models';
 import { getDomain } from './utils';
+import { ProviderContext } from '../../providers/context';
+
+export function useDatasetValue(id: HDF5Id): HDF5Value | undefined {
+  const { getValue } = useContext(ProviderContext);
+  const [valueReader] = useAsyncResource(getValue, id);
+  return valueReader();
+}
```
axelboc, line 22: The hook that does all the fetching. Much more concise than what used to be in `DatasetVisualizer`!

tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "allowSyntheticDefaultImports": true,
     "strict": true,
     "forceConsistentCasingInFileNames": true,
-    "importsNotUsedAsValues": "error",
+    // "importsNotUsedAsValues": "error",
```
axelboc, line 11: I'm getting an` import type` TS error in `node_modules`, in `use-async-resource`, even though `skipLibCheck` is enabled. I think this is a TypeScript bug, so I've disabled the check for now.

src/h5web/nexus-visualizer/models.ts
```diff
@@ -1,16 +0,0 @@
-import { Vis } from '../dataset-visualizer/models';
```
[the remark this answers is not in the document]
loichuder: > Moved into `visualizations/nexus`.
> 
> I did remove `NXDataAttribute` as I felt it was an unnecessary abstraction for now -- `getAttributeValue(dataset, 'signal')` is just easier to read in my opinion.

:crying_cat_face: 

I get your point but I think it is better to have a single place where we define Nexus keywords. Moreover, `axes` and `default` will be added to the list in next PRs.
> 
> I also removed `INTERPRETATION_VIS` because I no longer need the mapping. I use the interpretation in the `supportsEntity` functions of the two NeXus visualizations, so I can just check `interpretation === Vis.NxImage` for instance.

:+1: for the rest

src/h5web/visualizations/index.tsx
```diff
@@ -1,101 +1,152 @@
...
+      if (!dataset) {
+        return false;
+      }
+
+      const interpretation = getAttributeValue(dataset, 'interpretation');
```
loichuder, line 140: `interpretation` should be checked on `entity` not on `dataset`. Same for `NXSpectrum`

src/h5web/visualizations/index.tsx
```diff
@@ -1,101 +1,152 @@
...
+    Component: NxImageContainer,
+    supportsEntity: (entity, metadata) => {
+      if (!isGroup(entity) || !isNxDataGroup(entity)) {
+        return false;
+      }
+
+      const dataset = getSignalDataset(entity, metadata);
+      if (!dataset) {
+        return false;
+      }
```
loichuder, line 138: Again, I am not sure about this as this early return will lead to a `Nothing to visualize` message that is not that helpful. 

Would it be possible to throw an error be caught by the `ErrorBoundary` ?

src/h5web/visualizations/nexus/NxImageContainer.tsx
loichuder: The error will never be thrown as an entity pointing to a non-valid `dataset` will be considered as not supported by the `NxImage`.

See my comment above in the `supportsEntity` function of `NxImage`

src/h5web/providers/hooks.ts
```diff
@@ -1,16 +1,9 @@
...
   const { metadata } = useContext(ProviderContext);
-
-  if (!link || !isReachable(link)) {
-    return undefined;
-  }
-
-  const { collection, id } = link;
-  const dict = metadata[collection];
-  return dict && dict[id];
+  return getEntity(link, metadata);
```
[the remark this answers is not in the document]
loichuder: Is the `useEntity` hook really useful then ?

src/h5web/dataset-visualizer/hooks.ts
```diff
@@ -1,4 +1,25 @@
-import { createMemo } from 'react-use';
-import { getSupportedVis } from './utils';
+import { useContext } from 'react';
+import { HDF5Entity } from '../providers/models';
+import { Vis } from './models';
+import { VIS_DEFS } from '../visualizations';
+import { ProviderContext } from '../providers/context';
 
-export const useSupportedVis = createMemo(getSupportedVis);
```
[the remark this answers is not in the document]
loichuder: You could do as in `useEntity`: have a `getSupportedVis` function with `metadata` in arg and a hook `useSupportedVis` that will call the function.

loichuder (changes requested):

src/h5web/visualizations/line/LineVisContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  return (
+    <>
+      <DimensionMapper
+        rawDims={dims}
+        mapperState={mapperState}
+        onChange={setMapperState}
+      />
+      <MappedLineVis value={value} dataset={entity} mapperState={mapperState} />
```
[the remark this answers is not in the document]
axelboc: It will definitely need to be investigated, but the mapped components are not trivial and inlining them in the containers would duplicate a lot more code in the NeXuS visualizations.

src/h5web/dataset-visualizer/hooks.ts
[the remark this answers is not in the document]
axelboc: Yeah, though in this case, I don't need the `getSupportedVis` function on its own -- for now at least.

src/h5web/visualizations/nexus/NxImageContainer.tsx
axelboc: Yep, yep, but the checks in the containers are important for two reasons:

- If the container ends up being used outside of h5web.
- For type checking purposes (we don't have to worry about the dataset being `undefined` in the rest of the container).

src/h5web/visualizations/heatmap/HeatmapVisContainer.tsx
```diff
@@ -0,0 +1,44 @@
...
+import MappedHeatmapVis from './MappedHeatmapVis';
+import { VisContainerProps } from '../shared/models';
+
+function HeatmapVisContainer(props: VisContainerProps): ReactElement {
+  const { entity } = props;
+  assertDataset(entity);
+
+  const value = useDatasetValue(entity.id);
+
+  const { dims } = entity.shape as HDF5SimpleShape;
```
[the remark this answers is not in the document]
axelboc: Hmm, yeah but it wouldn't hurt. We don't need the check for typechecking purposes, but it would be useful if the container were used on its own, outside of h5web. What do you think?

src/h5web/nexus-visualizer/models.ts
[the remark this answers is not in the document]
axelboc: How about typing `getAttributeValue` so it only accepts a union of keywords, i.e. `type NxAttribute = 'signal' | 'interpretation' | 'axes' | ...` ?

src/h5web/providers/hooks.ts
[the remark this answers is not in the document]
axelboc: Probably not -- it's only called in `App` and making it clear that we retrieve the metadata from the context in `App` isn't a bad thing. 👍

src/h5web/visualizations/line/LineVisContainer.tsx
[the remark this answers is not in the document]
loichuder: Ok let's leave this for later then.

src/h5web/visualizations/shared/hooks.ts
axelboc: I'll try to replace it with `use-suspender`, as discussed.

tsconfig.json
axelboc: Hopefully `use-suspender` won't trigger this error...

src/h5web/visualizations/heatmap/HeatmapVisContainer.tsx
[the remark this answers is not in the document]
loichuder: We don't export containers yet so it should not need a check. On the other end, we might forget it if we export them in several weeks/months.

I would add an `assert`. I think it fits well the situation here where we know from the context that the dataset has a particular shape but still, if something goes wrong, it should hard-break.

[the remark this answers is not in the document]
axelboc: Agreed, it also strengthens the idea of keeping the checks and the containers completely decoupled.

src/h5web/nexus-visualizer/models.ts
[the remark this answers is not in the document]
loichuder: I find the `enum` handy as autocompletion ensures that no typos are made when calling the function.

But your proposition is fine as well. I just want to have a structure that defines the keywords.

src/h5web/visualizations/nexus/NxImageContainer.tsx
loichuder: Makes sense :+1:

src/h5web/providers/hooks.ts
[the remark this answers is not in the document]
axelboc: Boom, one fewer hook utility file 👍

src/h5web/visualizations/index.tsx
```diff
@@ -1,101 +1,152 @@
...
+
+      const interpretation = getAttributeValue(dataset, 'interpretation');
+      if (interpretation) {
+        return interpretation === NxInterpretation.Spectrum;
+      }
```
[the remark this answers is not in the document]
axelboc: Hmm problem... The error boundary is located in `Visualizer`, and so is the call to `useSupportedVis` which calls `supportsEntity`. If I throw in a `supportsEntity` function, it won't be caught by the boundary.

We could also throw if the NX data group doesn't have an `signal` attribute, or if the attribute points to a non-existing dataset... The HDF5 file could misuse NeXuS in many ways... To me, this is a different class of errors than the assertions that we have in the containers, which are programmer errors and have to break things (i.e. stop rendering the vis component and show an error).

The need here is for the creator of a file (or a person involved in the file's existence) to be able to debug what's wrong with it. Not everyone looking at a file would have this debugging need, so displaying an error in the same way as we display programmer and data fetching errrors might worsen the UX for these people.

What we could do is simply log errors to the console. Alternatively, we could throw errors from `supportsEntity`, catch them in `useSupportedVis` and return the list of errors that got caught to `Visualizer` so it can display them in a smart way (e.g. in a tooltip or an accordion).

Should we put this aside for this PR and talk about it in an issue?

src/h5web/nexus-visualizer/models.ts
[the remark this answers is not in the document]
axelboc: Cool, but note that the union type would also be auto-completed.

src/h5web/visualizations/index.tsx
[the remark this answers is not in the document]
loichuder: > Hmm problem... The error boundary is located in Visualizer, and so is the call to useSupportedVis which calls supportsEntity. If I throw in a supportsEntity function, it won't be caught by the boundary.

I feared as such.

> Should we put this aside for this PR and talk about it in an issue?

Yes, sounds good :+1:

src/h5web/visualizations/index.tsx
axelboc: The `entity` is the NxData group. According to the spec (and to the water file), the `interpretation` attribute is not defined on this group (https://manual.nexusformat.org/classes/base_classes/NXdata.html?highlight=nxdata) but on the `signal` _field_ (i.e. the `dataset`) that the NXdata group references: https://manual.nexusformat.org/nxdl_desc.html?highlight=interpretation#interpretation

loichuder: I will never get these right :joy_cat:

src/h5web/visualizations/index.tsx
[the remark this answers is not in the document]
axelboc: See #254

src/h5web/visualizations/index.tsx
axelboc: See #254

src/h5web/visualizations/shared/hooks.ts
axelboc: Unlike `use-async-resource`'s `valueReader`, the hook generated with `use-suspender` calls the async function (i.e. `getValue`) every time, which suspends the rendering of the vis containers and leads to render flashes when slicing. We're better off staying with `use-async-resource`.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+import { getEntity, isDataset } from '../../providers/utils';
+import { NxAttribute } from './models';
+
+export function isNxDataGroup(group: HDF5Group): boolean {
+  return !!group.attributes?.find(({ value }) => value === 'NXdata');
+}
+
+export function getAttributeValue(
+  entity: HDF5Dataset | HDF5Group,
+  attributeName: NxAttribute
```
loichuder, line 16: The typing means `getAttributeValue` cannot be used outside a NeXuS context.

Not a huge deal for now as we don't use it elsewhere but worth keeping in mind.

src/h5web/visualizations/index.tsx
```diff
@@ -1,101 +1,152 @@
...
-    dataset: HDF5Dataset;
-    mapperState: DimensionMapping;
-  }>;
-  supportsDataset(dataset: HDF5Dataset): boolean;
-  defaultMappingState(dataset: HDF5Dataset): DimensionMapping;
+  Container: ElementType<VisContainerProps>;
+  supportsEntity(entity: HDF5Entity, metadata: HDF5Metadata): boolean;
 }
 
+// @ts-ignore
```
loichuder, line 38: ?

loichuder (approved):

src/h5web/visualizations/index.tsx
axelboc: 🎣

src/h5web/visualizations/nexus/utils.ts
axelboc: That's kind of why I had removed the enum and left a simple `string`. I figured that we were just as likely to make a typo in the call to `getAttributeValue` as we were to make a typo in the enum or the union type. Either way, you would notice the typo quite quickly since the visualization would just not work 😄. Anyway, we can see what happens later on. We can rename the union type to include other HDF5 attributes, rename the function `getNxAttributeValue`, go back to a simple `string`, etc. plenty of options.

PR #255, opened 2020-10-21
Move files, models and utilities following visualizer refactoring

axelboc:


src/h5web/visualizations/index.tsx
```diff
@@ -13,21 +12,31 @@ import {
...
-import LineVisContainer from './line/LineVisContainer';
-import HeatmapVisContainer from './heatmap/HeatmapVisContainer';
-import NxSpectrumContainer from './nexus/NxSpectrumContainer';
+import RawViscontainer from './containers/RawVisContainer';
+import type { VisContainerProps } from './containers/models';
+import ScalarVisContainer from './containers/ScalarVisContainer';
+import MatrixVisContainer from './containers/MatrixVisContainer';
+import LineVisContainer from './containers/LineVisContainer';
+import HeatmapVisContainer from './containers/HeatmapVisContainer';
+import NxSpectrumContainer from './containers/NxSpectrumContainer';
```
loichuder, line 21: We could add an `index.ts` to `containers` to import all `Containers` from `containers`

loichuder (approved):

src/h5web/visualizations/index.tsx
axelboc: Ah great idea

Issue #254, opened 2020-10-20
Error management for malformed NeXuS structure


axelboc:
In #252 we [discussed](https://github.com/silx-kit/h5web/pull/252#discussion_r507823537) throwing errors inside the `supportsEntity` functions of the NeXuS visualizations when we determine that the NeXuS structure is invalid (e.g. missing `signal` dataset, unknown `interpretation`, etc.)

> [...] To me, this is a different class of errors than the assertions that we have in the containers, which are programmer errors and have to break things (i.e. stop rendering the vis component and show an error).
>
> The need here is for the creator of a file (or a person involved in the file's existence) to be able to debug what's wrong with it. Not everyone looking at a file would have this debugging need, so displaying an error in the same way as we display programmer and data fetching errrors might worsen the UX for these people.
>
> What we could do is simply log errors to the console. Alternatively, we could throw errors from `supportsEntity`, catch them in `useSupportedVis` and return the list of errors that got caught to `Visualizer` so it can display them in a smart way (e.g. in a tooltip or an accordion).

PR #256, opened 2020-10-21
NXSpectrum now uses the axes attribute to fetch the abscissa dataset to be used in the vis.

### LINKED ISSUE #236: [Nexus] Line vis for 1D @signal with @axes
Part of epic #196

loichuder:
Fix #236

Even with the heavy refactoring of #251, this ended up more complex than expected.

`NXSpectrum` uses now the `MappedLineVis` and not the `LineVis` which means I had to implement a way to determine the correct dataset as abscissas to use given its position in the `axes` list.

### DIFF
--- a/src/h5web/providers/Provider.tsx
+++ b/src/h5web/providers/Provider.tsx
@@ -17,12 +17,20 @@ function Provider(props: Props): JSX.Element {
     return <></>;
   }
 
+  const getValue = api.getValue.bind(api);
+
   return (
     <ProviderContext.Provider
       value={{
         domain: api.domain,
         metadata,
-        getValue: api.getValue.bind(api),
+        getValue,
+        getValues: (idsRecord) =>
+          Promise.all(
+            Object.entries(idsRecord).map(([name, id]) =>
+              getValue(id).then((val) => [name, val])
+            )
+          ).then(Object.fromEntries),
       }}
     >
       {children}
--- a/src/h5web/providers/context.ts
+++ b/src/h5web/providers/context.ts
@@ -11,4 +11,7 @@ export const ProviderContext = createContext<{
   domain: string;
   metadata: HDF5Metadata;
   getValue: ProviderAPI['getValue'];
+  getValues: (
+    idsRecord: Record<string, HDF5Id>
+  ) => Promise<Record<string, HDF5Value>>;
 }>({} as any); // eslint-disable-line @typescript-eslint/no-explicit-any
--- a/src/h5web/visualizations/containers/NxSpectrumContainer.tsx
+++ b/src/h5web/visualizations/containers/NxSpectrumContainer.tsx
@@ -1,13 +1,17 @@
 import React, { ReactElement, useState, useContext } from 'react';
 import { range } from 'lodash-es';
-import { HDF5SimpleShape } from '../../providers/models';
-import { useDatasetValue } from './hooks';
-import { assertGroup } from '../../providers/utils';
+import { HDF5SimpleShape, HDF5Id } from '../../providers/models';
+import { useDatasetValues } from './hooks';
+import { assertGroup, isDataset } from '../../providers/utils';
 import DimensionMapper from '../../dimension-mapper/DimensionMapper';
 import { DimensionMapping } from '../../dimension-mapper/models';
 import MappedLineVis from '../line/MappedLineVis';
 import { ProviderContext } from '../../providers/context';
-import { getSignalDataset } from '../nexus/utils';
+import {
+  getSignalDataset,
+  getLinkedEntity,
+  getAxesLabels,
+} from '../nexus/utils';
 import { VisContainerProps } from './models';
 
 function NxSpectrumContainer(props: VisContainerProps): ReactElement {
@@ -15,25 +19,39 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
   assertGroup(entity);
 
   const { metadata } = useContext(ProviderContext);
-  const dataset = getSignalDataset(entity, metadata);
+  const signalDataset = getSignalDataset(entity, metadata);
 
-  if (!dataset) {
+  if (!signalDataset) {
     throw new Error('Signal dataset not found');
   }
 
-  const { dims } = dataset.shape as HDF5SimpleShape;
+  const { dims } = signalDataset.shape as HDF5SimpleShape;
   if (dims.length < 1) {
     throw new Error('Expected dataset with at least one dimension');
   }
+
+  const axesLabels = getAxesLabels(entity);
+
+  const axesIds = axesLabels.reduce<Record<string, HDF5Id>>((acc, axis) => {
+    if (!axis) {
+      return acc;
+    }
+
+    const dataset = getLinkedEntity(entity, metadata, axis);
+    if (dataset && isDataset(dataset)) {
+      acc[axis] = dataset.id;
+    }
+    return acc;
+  }, {});
 
   const [mapperState, setMapperState] = useState<DimensionMapping>([
     ...range(dims.length - 1).fill(0),
     'x',
   ]);
 
-  const value = useDatasetValue(dataset.id);
+  const values = useDatasetValues({ signal: signalDataset.id, ...axesIds });
 
-  if (!value) {
+  if (!values || !values.signal) {
     return <></>;
   }
 
@@ -45,8 +63,10 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
         onChange={setMapperState}
       />
       <MappedLineVis
-        value={value}
-        dataset={dataset}
+        value={values.signal}
+        axesLabels={axesLabels}
+        axesValues={values}
+        dims={dims}
         mapperState={mapperState}
       />
     </>
--- a/src/h5web/visualizations/line/MappedLineVis.tsx
+++ b/src/h5web/visualizations/line/MappedLineVis.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement, useEffect } from 'react';
-import type { HDF5Dataset, HDF5Value } from '../../providers/models';
+import type { HDF5Value } from '../../providers/models';
 import type { DimensionMapping } from '../../dimension-mapper/models';
 import LineVis from './LineVis';
 import { assertArray } from '../shared/utils';
@@ -8,12 +8,14 @@ import { useLineConfig } from './config';
 
 interface Props {
   value: HDF5Value;
-  dataset: HDF5Dataset;
+  dims: number[];
   mapperState: DimensionMapping;
+  axesLabels?: (string | undefined)[];
+  axesValues?: Record<string, HDF5Value>;
 }
 
 function MappedLineVis(props: Props): ReactElement {
-  const { value, dataset, mapperState } = props;
+  const { value, axesLabels = [], axesValues = {}, dims, mapperState } = props;
   assertArray<number>(value);
 
   const {
@@ -24,7 +26,7 @@ function MappedLineVis(props: Props): ReactElement {
     disableAutoScale,
   } = useLineConfig();
 
-  const baseArray = useBaseArray(dataset, value);
+  const baseArray = useBaseArray(value, dims);
   const dataArray = useMappedArray(baseArray, mapperState);
 
   // Disable `autoScale` for 1D datasets (baseArray and dataArray are the same)
@@ -37,9 +39,16 @@ function MappedLineVis(props: Props): ReactElement {
     scaleType
   );
 
+  const abscissaLabel = mapperState && axesLabels[mapperState.indexOf('x')];
+  const abscissas = abscissaLabel && axesValues[abscissaLabel];
+  if (abscissas) {
+    assertArray<number>(abscissas);
+  }
+
   return (
     <LineVis
       dataArray={dataArray}
+      abscissas={abscissas as number[]}
       domain={dataDomain}
       scaleType={scaleType}
       curveType={curveType}
--- a/src/h5web/visualizations/nexus/utils.ts
+++ b/src/h5web/visualizations/nexus/utils.ts
@@ -3,9 +3,11 @@ import type {
   HDF5Metadata,
   HDF5Dataset,
   HDF5Value,
+  HDF5Entity,
 } from '../../providers/models';
 import { getEntity, isDataset } from '../../providers/utils';
 import { NxAttribute, NX_INTERPRETATIONS } from './models';
+import { assertArray, assertStr } from '../shared/utils';
 
 export function isNxDataGroup(group: HDF5Group): boolean {
   return !!group.attributes?.find(({ value }) => value === 'NXdata');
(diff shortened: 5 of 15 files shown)


src/h5web/visualizations/line/MappedLineVis.tsx
```diff
@@ -37,9 +38,16 @@ function MappedLineVis(props: Props): ReactElement {
     scaleType
   );
 
+  const abscissas =
+    mapperState && axesValues && axesValues[mapperState.indexOf('x')];
```
loichuder, line 42: This is the line from which everything stems. Because the only way to determine the dataset to use is its position in the array `axesValues`, I have to keep the number of elements (and its ordering) from the parsing of the `axes` attribute to here.

This means that, at several places, I have to keep `undefined` values in the arrays as will be pointed out in other comments.

src/h5web/providers/Provider.tsx
```diff
@@ -23,6 +23,12 @@ function Provider(props: Props): JSX.Element {
         domain: api.domain,
         metadata,
         getValue: api.getValue.bind(api),
+        getValues: (ids) =>
+          Promise.all(
+            ids.map((id) =>
+              id ? api.getValue(id) : Promise.resolve(undefined)
```
loichuder, line 29: To keep the number of elements: https://github.com/silx-kit/h5web/pull/256/files#r509287208

src/h5web/providers/context.ts
```diff
@@ -11,4 +11,7 @@ export const ProviderContext = createContext<{
   domain: string;
   metadata: HDF5Metadata;
   getValue: ProviderAPI['getValue'];
+  getValues: (
+    ids: (HDF5Id | undefined)[]
+  ) => Promise<(HDF5Value | undefined)[]>;
```
loichuder, line 16: `getValues` is then not consistent with `getValue` that does not accept `undefined` id but I found no other alternatives: https://github.com/silx-kit/h5web/pull/256/files#r509287208

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
```diff
@@ -31,9 +32,12 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
     'x',
   ]);
 
-  const value = useDatasetValue(dataset.id);
+  const values = useDatasetValues([
+    dataset.id,
+    ...axesDatasets.map((d) => d && d.id),
```
loichuder, line 37: If the dataset was not found, return an `undefined` id to keep the numbers of elements in the array (https://github.com/silx-kit/h5web/pull/256/files#r509287208). This means `useDatasetValues` and its Promise-based version `getValues` must accept `undefined` arguments (https://github.com/silx-kit/h5web/pull/256/files#r509288853)

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -30,13 +31,52 @@ export function getAttributeValue(
...
+  metadata: HDF5Metadata
+): (HDF5Dataset | undefined)[] | undefined {
+  const axesList = getAttributeValue(group, 'axes') as
+    | string
+    | Array<string>
+    | undefined;
+  const axes =
+    typeof axesList === 'string'
+      ? (axesList as string).match(/\w+|\./g)
+      : axesList;
```
loichuder, line 69: Parse the `axes` attributes. 

I consider two usecases for now:
- The attribute is an array of strings (ex: `['x', 'y']). In this case, no parsing is needed.
- The attribute is a string with axis names separated by a separator character (ex: `"x y"`) as shown in [NeXuS docs](https://manual.nexusformat.org/rules.html?highlight=axes). To convert in an array, I match with a simple reg exp. If no match, this will be `null`. The particular case `.` (no dataset) is matched by the regex as well which will lead to an `undefined` dataset below.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -30,13 +31,52 @@ export function getAttributeValue(
   return entity.attributes.find((attr) => attr.name === attributeName)?.value;
 }
 
+export function getChildEntityByName(
```
axelboc, line 34: `getLinkedEntity` ?

src/h5web/visualizations/nexus/utils.ts
axelboc: Alright, so this can all be simplified as instructed by Armando and Thomas.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -30,13 +31,52 @@ export function getAttributeValue(
...
 }
+
+export function getAxesDatasets(
+  group: HDF5Group,
+  metadata: HDF5Metadata
+): (HDF5Dataset | undefined)[] | undefined {
+  const axesList = getAttributeValue(group, 'axes') as
+    | string
+    | Array<string>
+    | undefined;
```
axelboc, line 65: It would be better to use `assertArray<string>` here now, instead of casting.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -30,13 +31,52 @@ export function getAttributeValue(
   return entity.attributes.find((attr) => attr.name === attributeName)?.value;
 }
 
+export function getChildEntityByName(
+  group: HDF5Group,
+  metadata: HDF5Metadata,
+  entityName: string | undefined
```
axelboc, line 37: This should probably be `string` -- just make sure you don't call it with an undefined `entityName`.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -30,13 +31,52 @@ export function getAttributeValue(
...
+  return getEntity(childLink, metadata);
+}
+
 export function getSignalDataset(
   group: HDF5Group,
   metadata: HDF5Metadata
 ): HDF5Dataset | undefined {
   const signal = getAttributeValue(group, 'signal');
-  const signalLink = signal && group.links?.find((l) => l.title === signal);
+  const signalDataset = getChildEntityByName(group, metadata, signal as string);
```
axelboc, line 53: Maybe assert the type of `signal` instead of casting it?

src/h5web/visualizations/line/MappedLineVis.tsx
axelboc: I feel like this should be moved to `LineVisContainer`. You should simply propagate an `abscissas` prop`.

axelboc: There's something I don't understand, though: how do you know that the index of `x` in `mapperState` matches a dataset (and not a dot) in the `axes` attribute? Doesn't `mapperState` need to be initialized based on `axes`?

axelboc (changes requested):
I'm going to need help to understand the relationship between `axes` and `mapperState`. It's unclear to me, and I feel like things could be simplified by managing this relationship better.

src/h5web/visualizations/line/MappedLineVis.tsx
```diff
@@ -10,10 +10,11 @@ interface Props {
   value: HDF5Value;
   dataset: HDF5Dataset;
```
axelboc, line 11: By the way, we can probably pass the dimensions instead of the dataset now.

src/h5web/visualizations/line/MappedLineVis.tsx
axelboc: `axesValues` cannot be `undefined` since you use `slice()`, which always returns an array.

src/h5web/visualizations/nexus/utils.ts
loichuder: True. The NeXuS page I linked is in fact outdated: https://github.com/nexusformat/NIAC/issues/76 

 I consider now the following possibilities:

- The attribute is an array of strings (ex: `['x', 'y']`). 
- The attribute is a string. In this case, I consider the string as a single axis value (ex: `"time"` is interpreted as `["time"]`)

axelboc (commented):
I have an idea to avoid having an array of `HDF5Value | undefined`. How about this:

1. Get the value of the `axes` attribute and transform it to always be an array and to replace `"."` elements with `undefined` (i.e. leaving the dataset names alone) -- let's call the resulting array `axesArr`.
2. Reduce `axesArr` to transform it into an object with dataset names as keys and dataset entities as values (skipping `undefined` elements) - e.g. `["Y", undefined, "X"]` would for instance give `{ X: xDataset, Y: yDataset }` -- let's call the resulting constant `axesDatasets`.
3. Call `useDatasetValues` with `axesDatasets` combined with the signal dataset: `const values = useDatasetValues({ signal: signalDataset, ...axesDatasets })` - the returned `values` is an object mapping dataset names to dataset values.
4. You can now access this `values` object to pass the `value` prop to the mapped vis: `value={values.signal}`.
5. You can also determine the abscissa values as follows:

```ts
const abscissaDatasetName = axes && mapperState && axes[mapperState.indexOf('x')];
...
abscissas={abscissaDatasetName && values[abscissaDatasetName]}
```

Let me know if you think this is any clearer.

src/h5web/providers/Provider.tsx
```diff
@@ -17,12 +18,26 @@ function Provider(props: Props): JSX.Element {
...
+        getValues: (datasetsRecord) => {
+          const dsetNames = Object.keys(datasetsRecord);
+          return Promise.all(
+            dsetNames.map((n) => getValue(datasetsRecord[n].id))
+          ).then((values) => {
+            const valuesRecord: Record<string, HDF5Value> = {};
+            dsetNames.forEach((n, i) => {
+              valuesRecord[n] = values[i];
+            });
+            return valuesRecord;
```
axelboc, line 38: You can probably simplify all this with `Object.entities` and by providing a record of dataset IDs rather than datasets -- something like this:

```ts
return Promise.all(
  Object.entities(datasetsRecord).map(([name, id]) => getValue(id).then(val => [name, val])
).then(Object.fromEntries);
```

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
```diff
@@ -1,39 +1,58 @@
...
+      `Number of axes ${axesLabels.length} does not correspond to the dataset dimensions (${dims})`
+    );
+  }
+  const axesDatasets: Record<string, HDF5Dataset> = {};
+  axesLabels.forEach((a) => {
+    const dataset = a && a !== '.' && getLinkedEntity(entity, metadata, a);
+    if (dataset && isDataset(dataset)) {
+      axesDatasets[a] = dataset;
+    }
+  });
```
axelboc, line 46: You can do this with `axesLabels.reduce<Record<string, HDF5Dataset>>(..., {});`

axelboc: and you can store the datasets IDs directly in the record.

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
```diff
@@ -1,39 +1,58 @@
...
     throw new Error('Expected dataset with at least one dimension');
   }
 
+  const axesLabels =
+    getAxesLabels(entity) || Array(dims.length).fill(undefined);
+  if (dims.length !== axesLabels.length) {
+    throw new Error(
+      `Number of axes ${axesLabels.length} does not correspond to the dataset dimensions (${dims})`
+    );
+  }
```
axelboc, line 39: If I'm not mistaken, the rest of the code still works if `axesLabels` has fewer elements than the number of dimensions. So we could just support this use case and assume that `['X']` for a 2D dataset is equivalent to `['X', '.']`. This would allow the fallback of `axesLabels` to just be `[]` instead of an array of `undefined` values.

src/h5web/visualizations/line/MappedLineVis.tsx
```diff
@@ -37,9 +39,17 @@ function MappedLineVis(props: Props): ReactElement {
     scaleType
   );
 
+  const abscissaLabel =
+    mapperState && axesLabels && axesLabels[mapperState.indexOf('x')];
+  const abscissas = abscissaLabel && axesValues[abscissaLabel];
+  if (abscissas) {
+    assertArray<number>(abscissas);
```
axelboc, line 46: this

src/h5web/visualizations/line/MappedLineVis.tsx
```diff
@@ -8,12 +8,14 @@ import { useLineConfig } from './config';
...
-  dataset: HDF5Dataset;
+  dims: number[];
   mapperState: DimensionMapping;
+  axesLabels?: (string | undefined)[];
+  axesValues?: Record<string, HDF5Value>;
 }
 
 function MappedLineVis(props: Props): ReactElement {
-  const { value, dataset, mapperState } = props;
+  const { value, axesLabels, axesValues = {}, dims, mapperState } = props;
```
axelboc, line 18: With a default empty array for `axesLabels`, you can remove `axesLabels &&` below.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -30,13 +32,40 @@ export function getAttributeValue(
...
 
-  const signalDataset = getEntity(signalLink, metadata);
   return signalDataset && isDataset(signalDataset) ? signalDataset : undefined;
 }
+
+export function getAxesLabels(group: HDF5Group): Array<string> | undefined {
+  const axesList = getAttributeValue(group, 'axes');
+
+  if (!axesList) {
+    return undefined;
```
axelboc, line 64: Maybe this is where you could return an empty array.

axelboc (approved):
Cool cool cool

src/h5web/providers/Provider.tsx
loichuder: Oooh clever :100:

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
loichuder: > You can do this with `axesLabels.reduce<Record<string, HDF5Dataset>>(..., {});`

I tried this first but TS wouldn't let me. I will it give another go.

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
loichuder: The code should work but the NeXuS specification does not allow it (https://manual.nexusformat.org/classes/base_classes/NXdata.html). Should I do as you say and leave the verification to #254 ?

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
loichuder: It works ! I was forgetting the template in `reduce`.

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
axelboc: Well, the goal of #254 is not to validate the file but to help users debug invalid syntax that h5web does not support. In this case it is easier to support the invalid syntax than to validate it, so I would argue that being permissive makes sense.

loichuder: I don't know. 

It seems dangerous to me to assume that `['X']` should be `['X', '.']` (why not `['.', 'X']`) ? We may end up attributing the wrong dataset to the wrong dimensions leading to error in `LineVis` while the source is related to NeXuS.

axelboc (approved):
Nothing more to say! 💯

Issue #253, opened 2020-10-20
Mock data generation


axelboc:
As discussed last week, I propose that we use JS to generate mock data for testing purposes. The current workflow of cleaning up and adding to an existing JSON export of an HDF5 file with Python, and then updating the mock source files in the repo is quite tedious.

If the mock data was generated dynamically (but deterministically) with JS at runtime, we would be able to update it more easily to cover more use cases and support more tests. We could generate the data in the best possible format to avoid any transformations in the `MockDataProvider`. We could provide a number of utility functions to allow for the generation code to remain concise and readable.

Note that it is important that the mock data be generated only when visiting the `/mock` endpoint. This means lazy-loading the mock data module.

axelboc:
Note that the mock data utility functions will allow us to write more concise tests, notably in `visualizations/index.test.ts`.

PR #257, opened 2020-10-21
Unit test utility functions

axelboc:


src/h5web/visualizer/Visualizer.tsx
```diff
@@ -1,20 +1,22 @@
...
 interface Props {
   entity?: HDF5Entity;
 }
 
 function Visualizer(props: Props): ReactElement {
   const { entity } = props;
+  const { metadata } = useContext(ProviderContext);
 
-  const supportedVis = useSupportedVis(entity);
+  const supportedVis = entity ? getSupportedVis(entity, metadata) : [];
```
axelboc, line 19: We talked about it I think, I'm turning the hook into a simple function and I move a bit of the responsibility out of it and into `Visualizer`. This makes it more testable and just easier to read I think.

src/h5web/visualizations/index.test.ts
```diff
@@ -287,8 +306,253 @@ describe('Visualization definitions', () => {
...
+        const metadata = {
+          datasets: { [baseDataset.id]: signal },
+        } as HDF5Metadata;
+
+        const supported = nxSpectrumDef.supportsEntity(nxDataGroup, metadata);
+        expect(supported).toBe(true);
+      });
+    });
+
+    it('should not support NXdata group referencing signal with non-numeric type, non-simple shape, less than two dimensions or with known, non-spectrum interpretation', () => {
```
loichuder, line 362: less than **one** dimension

src/h5web/visualizations/index.test.ts
```diff
@@ -287,8 +306,253 @@ describe('Visualization definitions', () => {
...
+          attributes: [], // no interpretation
+        },
+        {
+          ...baseDataset,
+          type: { class: HDF5TypeClass.Float, base: 'H5T_IEEE_F32BE' },
+          shape: { class: HDF5ShapeClass.Simple, dims: [1] },
+          attributes: [
+            { name: 'interpretation', value: 'unknown' } as HDF5Attribute, // unknown interpretation
+          ],
+        },
```
loichuder, line 349: Given how interpretations are handled for now, I would also add a 2D dataset with unknown interpretation to the supported datasets.

src/h5web/visualizations/index.test.ts
```diff
@@ -287,8 +306,253 @@ describe('Visualization definitions', () => {
...
+            ],
+          },
+        },
+      } as HDF5Metadata;
+
+      const supported = nxSpectrumDef.supportsEntity(nxDataGroup, metadata);
+      expect(supported).toBe(true);
+    });
+
+    it('should support NXdata group referencing signal with numeric type, simple shape, one dimension and no or unknown interpreation', () => {
```
loichuder, line 334: interpre**t**ation. To be corrected here and in all the sentences below :smile:

src/h5web/visualizations/index.test.ts
```diff
@@ -287,8 +306,253 @@ describe('Visualization definitions', () => {
...
+        {} as HDF5Metadata
+      );
+
+      expect(supported).toBe(false);
+    });
+
+    it('should not support dataset, datatype and non-NXdata group', () => {
+      const unsupportedEntities: HDF5Entity[] = [
+        { ...baseDataset, type: '', shape: { class: HDF5ShapeClass.Null } },
+        { ...baseDatatype, type: '' },
```
loichuder, line 550: What happens if you put a valid `NXdata` attribute to a dataset or a datatype ?

loichuder (approved):
Good !

src/h5web/visualizations/index.test.ts
axelboc: 🎣

src/h5web/visualizations/index.test.ts
axelboc: If the interpretation is unknown, it falls back to the dimension check. Since a 2D dataset with an unknown interpretation is supported by `Vis.NxImage`, then it doesn't need to be supported by `Vis.NxSpectrum`. That's why the check is different than for the line vis: `Vis.NxSpectrum` only supports 1D datasets. We had discussed this with Thomas, if I recall.

src/h5web/visualizations/index.test.ts
axelboc: I got lazy on this one -- you're right, it's important to check.

src/h5web/visualizations/index.test.ts
loichuder: > Vis.NxSpectrum only supports 1D datasets, not nD datasets.

src/h5web/visualizations/index.test.ts
axelboc: Hmm, if you don't mind, I'm going to work on mock data generation first (#253), because this is becoming quite unwieldy. Once I have a set of utility functions I can use to generate all kinds of HDF5 models, it will clean up this test file significantly.

PR #259, opened 2020-10-22
Lazy-load MockProvider for code-splitting purposes

axelboc:
With #253, I'm planning to provide a number of mock data generation utilities, and perhaps generate the mock `metadata` object right away, so as discussed, lazy-loading the provider to split it from the main bundle makes sense.


loichuder (approved):
:sloth:

Issue #244, opened 2020-10-09
[Nexus] Display the @signal title as plot title for plots of NXdata


loichuder:
Part of #196

PR #260, opened 2020-10-22
NxSpectrum now shows the axis label for x and y

loichuder:
image

Updates also the tooltip.

I also added some stories showing this for `LineVis` and `AxisSystem`

### DIFF
--- a/src/h5web/visualizations/containers/NxSpectrumContainer.tsx
+++ b/src/h5web/visualizations/containers/NxSpectrumContainer.tsx
@@ -8,20 +8,23 @@ import { DimensionMapping } from '../../dimension-mapper/models';
 import MappedLineVis from '../line/MappedLineVis';
 import { ProviderContext } from '../../providers/context';
 import {
-  getSignalDataset,
+  getAttributeValue,
   getLinkedEntity,
   getAxesLabels,
 } from '../nexus/utils';
 import { VisContainerProps } from './models';
+import { assertStr } from '../shared/utils';
 
 function NxSpectrumContainer(props: VisContainerProps): ReactElement {
   const { entity } = props;
   assertGroup(entity);
 
   const { metadata } = useContext(ProviderContext);
-  const signalDataset = getSignalDataset(entity, metadata);
+  const signal = getAttributeValue(entity, 'signal');
+  assertStr(signal);
+  const signalDataset = getLinkedEntity(entity, metadata, signal);
 
-  if (!signalDataset) {
+  if (!signalDataset || !isDataset(signalDataset)) {
     throw new Error('Signal dataset not found');
   }
 
@@ -64,6 +67,7 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
       />
       <MappedLineVis
         value={values.signal}
+        valueLabel={signal}
         axesLabels={axesLabels}
         axesValues={values}
         dims={dims}
--- a/src/h5web/visualizations/line/LineVis.tsx
+++ b/src/h5web/visualizations/line/LineVis.tsx
@@ -7,7 +7,7 @@ import DataCurve from './DataCurve';
 import VisCanvas from '../shared/VisCanvas';
 import PanZoomMesh from '../shared/PanZoomMesh';
 import TooltipMesh from '../shared/TooltipMesh';
-import { ScaleType, Domain } from '../shared/models';
+import { ScaleType, Domain, AxisParams } from '../shared/models';
 import { CurveType } from './models';
 import { getValueToIndexScale, getDomain, extendDomain } from '../shared/utils';
 
@@ -19,7 +19,8 @@ interface Props {
   scaleType?: ScaleType;
   curveType?: CurveType;
   showGrid?: boolean;
-  abscissas?: number[];
+  abscissaParams?: AxisParams;
+  ordinateLabel?: string;
 }
 
 function LineVis(props: Props): ReactElement {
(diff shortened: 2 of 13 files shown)


src/h5web/visualizations/shared/VisCanvas.tsx
```diff
@@ -7,7 +7,7 @@ import { computeVisSize } from './utils';
 import AxisSystem from './AxisSystem';
 import AxisSystemProvider from './AxisSystemProvider';
 
-const AXIS_OFFSETS = { vertical: 72, horizontal: 36, fallback: 16 };
+const AXIS_OFFSETS = { vertical: 72, horizontal: 54, fallback: 16 };
```
loichuder, line 10: I had to make a bit of space for the axis label. Otherwise, it was superimposed with ticks.

src/h5web/visualizations/shared/Axis.tsx
```diff
@@ -9,6 +9,9 @@ const AXIS_PROPS = {
   tickStroke: 'grey',
   hideAxisLine: true,
   tickClassName: styles.tick,
+  labelClassName: styles.label,
+  labelProps: {}, // To avoid any styling props on the parent svg
```
loichuder, line 13: Otherwise the parent `svg` has `font-size=10` which makes a tiny tiny label due to `font-size: inherit` from `styles.label`.

src/h5web/visualizations/shared/AxisSystem.module.css
```diff
@@ -37,3 +37,10 @@
   text-anchor: inherit;
   dominant-baseline: inherit;
 }
+
+.label {
+  fill: darkslategrey;
+  font-family: inherit;
+  font-size: inherit;
+  font-weight: 600;
```
loichuder, line 45: This was to make the axis label stand out from the tick labels.

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
```diff
@@ -8,20 +8,26 @@ import { DimensionMapping } from '../../dimension-mapper/models';
...
   const { entity } = props;
   assertGroup(entity);
 
   const { metadata } = useContext(ProviderContext);
-  const signalDataset = getSignalDataset(entity, metadata);
+  const signal = getAttributeValue(entity, 'signal');
+  if (!signal) {
+    throw new Error('Entity has no signal attribute');
+  }
+  assertStr(signal);
```
axelboc, line 27: `assertStr` is enough to check to `undefined`, so I don't think we need the throw. If the string is empty, `getLinkedEntity` will return `undefined` anyway.

src/h5web/visualizations/line/LineVis.tsx
```diff
@@ -30,6 +32,8 @@ function LineVis(props: Props): ReactElement {
     showGrid = true,
     scaleType = ScaleType.Linear,
     abscissas = range(dataArray.size),
+    abscissaLabel,
+    ordinateLabel,
```
axelboc, line 36: What do you think of what I mentioned in the chat: combining `abscissas` and `abscissaLabel` in a single prop called `abscissa` with type `{ label: string; values: number[] }`?

src/h5web/visualizations/shared/VisCanvas.tsx
axelboc: Any way of making this number dynamic based on the absence/presence of a label?

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
axelboc: This makes me realise that we shouldn't throw inside `getSignalDataset` since it's now only called from the `supportsEntity` functions, and these functions are not equipped to catch errors. 

```ts
  if (!signal) {
    return undefined;
  }

  assertStr(signal);
```

... should be replaced with:

```ts
  if (typeof signal !== 'string') {
    return undefined;
  }
```

Perhaps we should get rid of `getSignalDataset` altogether and inline the code?

axelboc (changes requested):
Looks good! Just a couple of suggestions and questions.

src/h5web/visualizations/line/LineVis.tsx
loichuder: I think it is a nice idea :smile_cat:

src/h5web/visualizations/shared/VisCanvas.tsx
loichuder: I thought about it but then I asked myself the following question: it is better to make space when we need to or to keep a consistent layout ?

axelboc: Excellent question... Hmm, so the layout will change when switching from an NXdata group to a classic dataset. I don't think it's a big deal. I feel like we should take more space if we can, especially since Nexus visualizations will also have a title... It will end up being quite a bit of unused whitespace for classic visualizations.

axelboc: Happy to deal with this later either way.

loichuder: Yeah perhaps we should leave it for once we tackled #244 to make everything in one go.

loichuder: Well, I looked into it and in the end, it was quite simple to implement : https://github.com/silx-kit/h5web/pull/260/files#diff-1b754f8c9c4a57ec8444d387642a4309b20ca9327a9621a7c56a33c731f7d1fbR26

axelboc (approved):
Awesome! 🏆

PR #261, opened 2020-10-23
Use mock data generation utilities in tests

axelboc:
Start of #253 

I add a file called `providers/mocks/data.ts` with mock HDF5 objects and generation utilities, which I then used to rewrite the tests of the visualizations' `supportsEntity` functions.

In the next PR, I will use these objects and utilities to generate mock data for the `/mock` endpoint.

### DIFF
--- a/src/h5web/providers/mock/data.ts
+++ b/src/h5web/providers/mock/data.ts
@@ -1,0 +1,180 @@
+import {
+  HDF5Id,
+  HDF5LinkClass,
+  HDF5Collection,
+  HDF5Attribute,
+  HDF5Dataset,
+  HDF5Metadata,
+  HDF5Values,
+  HDF5TypeClass,
+  HDF5IntegerType,
+  HDF5FloatType,
+  HDF5StringType,
+  HDF5AdvancedType,
+  HDF5SimpleShape,
+  HDF5Dims,
+  HDF5ShapeClass,
+  HDF5ScalarShape,
+  HDF5Type,
+  HDF5Shape,
+  HDF5Group,
+  HDF5Link,
+  HDF5Datatype,
+  HDF5Value,
+  HDF5HardLink,
+} from '../models';
+
+/* ----------------- */
+/* ----- TYPES ----- */
+
+export const intType: HDF5IntegerType = {
+  class: HDF5TypeClass.Integer,
+  base: 'H5T_STD_I32LE',
+};
+
+export const floatType: HDF5FloatType = {
+  class: HDF5TypeClass.Float,
+  base: 'H5T_IEEE_F64LE',
+};
+
+export const stringType: HDF5StringType = {
+  class: HDF5TypeClass.String,
+  charSet: 'H5T_CSET_ASCII',
+  strPad: 'H5T_STR_NULLPAD',
+  length: 1,
+};
+
+export const advancedType: HDF5AdvancedType = {
+  class: HDF5TypeClass.Compound,
+  fields: [],
+};
+
+/* ------------------ */
+/* ----- SHAPES ----- */
+
+export const scalarShape: HDF5ScalarShape = { class: HDF5ShapeClass.Scalar };
+
+export function makeSimpleShape(dims: HDF5Dims): HDF5SimpleShape {
+  return { class: HDF5ShapeClass.Simple, dims };
+}
+
+/* ---------------------- */
+/* ----- ATTRIBUTES ----- */
+
+export function makeStrAttr(name: string, value: HDF5Value): HDF5Attribute {
+  return { name, value, type: stringType, shape: scalarShape };
+}
+
+/* -------------------- */
+/* ----- ENTITIES ----- */
+
+export function makeDataset(
+  id: HDF5Id,
+  type: HDF5Type,
+  shape: HDF5Shape,
+  attributes?: HDF5Attribute[]
+): HDF5Dataset {
+  return { id, collection: HDF5Collection.Datasets, type, shape, attributes };
+}
+
+export function makeSimpleDataset(
+  id: HDF5Id,
+  type: HDF5Type,
+  dims: HDF5Dims,
+  attributes?: HDF5Attribute[]
+): HDF5Dataset {
+  return makeDataset(id, type, makeSimpleShape(dims), attributes);
+}
+
+export function makeGroup(
+  id: HDF5Id,
+  attributes?: HDF5Attribute[],
+  links?: HDF5Link[]
+): HDF5Group {
+  return { id, collection: HDF5Collection.Groups, attributes, links };
+}
+
+export function makeDatatype(id: HDF5Id, type: HDF5Type): HDF5Datatype {
+  return { id, collection: HDF5Collection.Datatypes, type };
+}
+
+export function withAttributes<T extends HDF5Dataset | HDF5Group>(
+  datasetOrGroup: T,
+  attributes: HDF5Attribute[]
+): T {
+  return {
+    ...datasetOrGroup,
+    attributes: [...(datasetOrGroup.attributes || []), ...attributes],
+  };
+}
+
+/* ----------------- */
+/* ----- LINKS ----- */
+
+export function makeDatasetHardLink(title: string, id: string): HDF5HardLink {
+  return {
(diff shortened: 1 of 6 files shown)


src/h5web/providers/mock/data.ts
```diff
@@ -0,0 +1,194 @@
...
+
+/* ----------------- */
+/* ----- NEXUS ----- */
+
+export function makeNxDataGroup(
+  id: string,
+  opts: {
+    signal: string;
+    axes?: string | string[];
+    ids?: Record<string, HDF5Id>;
```
axelboc, line 151: This signature gives me more flexibility (i.e. I can pass `ids` without `axes`) and clarity (i.e. the signal and axes params are identified as such in the caller).

src/h5web/providers/mock/data.ts
```diff
@@ -0,0 +1,194 @@
...
+
+/* -------------------- */
+/* ----- METADATA ----- */
+
+export function makeMetadata(
+  opts: {
+    root?: HDF5Id;
+    datasets?: HDF5Dataset | HDF5Dataset[];
+    groups?: HDF5Group | HDF5Group[];
+    datatypes?: HDF5Datatype | HDF5Datatype[];
```
axelboc, line 175: Same, an `opts` object allows me to pass only some of these parameters in any order.

src/h5web/visualizations/index.test.ts
```diff
@@ -1,559 +1,408 @@
...
-      const unsupportedEntities: HDF5Entity[] = [
-        { ...baseGroup },
-        { ...baseDatatype, type: '' },
-      ];
-
-      unsupportedEntities.forEach((entity) => {
-        expect(supportsEntity(Vis.Raw, entity)).toBe(false);
-      });
+      expect(supportsEntity(groupEmpty)).toBe(false);
+      expect(supportsEntity(datatypeAdvanced)).toBe(false);
```
axelboc, line 50: The new test syntax is a lot clear:

- no type casts (i.e. `as ...`) -- the generated mock objects are all valid HDF5 objects;
- no loop, instead each expect is written explicitly.

src/h5web/visualizations/index.test.ts
```diff
@@ -1,559 +1,408 @@
...
-        },
-      ];
-
-      unsupportedDatasets.forEach((dataset) => {
-        expect(supportsEntity(Vis.Scalar, dataset)).toBe(false);
-      });
+    it('should not support dataset with complex type', () => {
+      const dataset = makeDataset('foo', advancedType, scalarShape);
+      expect(supportsEntity(dataset)).toBe(false);
     });
```
axelboc, line 65: I split the "should not support" tests for more clarity.

src/h5web/visualizations/index.test.ts
```diff
@@ -1,559 +1,408 @@
...
+  stringType,
+  makeStrAttr,
+  makeMetadata,
+  withAttributes,
+  makeDatasetHardLink,
+  makeNxDataGroup,
+} from '../providers/mock/data';
+
+const datasetIntScalar = makeDataset('dataset_int', intType, scalarShape);
+const datasetFltScalar = makeDataset('dataset_flt', intType, scalarShape);
```
loichuder, line 22: `datasetFltScalar` was created with `intType`.

Anyway, there is no distinction (that I know of) between a `datasetIntScalar` and `datasetIntScalar` so perhaps one is sufficient.

src/h5web/visualizations/index.test.ts
```diff
@@ -1,559 +1,408 @@
...
+const datasetInt0D = makeSimpleDataset('dataset_int_0d', intType, []);
+const datasetInt1D = makeSimpleDataset('dataset_int_1d', intType, [5]);
+const datasetInt2D = makeSimpleDataset('dataset_int_2d', intType, [5, 3]);
+const datasetStr1D = makeSimpleDataset('dataset_str_1d', stringType, [5]);
+const datasetStr2D = makeSimpleDataset('dataset_str_2d', stringType, [5, 3]);
+const datasetFlt3D = makeSimpleDataset('dataset_flt_3d', intType, [5, 3, 1]);
+
+const groupEmpty = makeGroup('group_empty');
```
loichuder, line 30: Nitpicking but I would find the variables easier to read if the adjectives were put before: `emptyGroup`, `advancedDatatype`.

The dataset names above are fine to me though.

loichuder (approved):
Nice :slightly_smiling_face:

src/h5web/providers/mock/data.ts
```diff
@@ -0,0 +1,194 @@
...
+  return makeDataset(id, type, makeSimpleShape(dims), attributes);
+}
+
+export function makeGroup(
+  id: HDF5Id,
+  attributes: HDF5Attribute | HDF5Attribute[] = [],
+  links: HDF5Link | HDF5Link[] = []
+): HDF5Group {
+  const attributesArr = Array.isArray(attributes) ? attributes : [attributes];
+  const linksArr = Array.isArray(links) ? links : [links];
```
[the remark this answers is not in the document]
axelboc: You think I should always ask for an array?

src/h5web/visualizations/index.test.ts
```diff
@@ -1,559 +1,408 @@
...
+    it('should not support dataset with non-scalar shape', () => {
+      expect(supportsEntity(datasetInt1D)).toBe(false);
+    });
 
-      unsupportedEntities.forEach((entity) => {
-        expect(supportsEntity(Vis.Scalar, entity)).toBe(false);
-      });
+    it('should not support group and datatype', () => {
+      expect(supportsEntity(groupEmpty)).toBe(false);
+      expect(supportsEntity(datatypeAdvanced)).toBe(false);
```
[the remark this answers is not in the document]
axelboc: You mean a base type, like `intType`? Yeah makes sense.

src/h5web/providers/mock/data.ts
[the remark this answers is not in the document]
loichuder: Yeah I think so. But your call.

src/h5web/visualizations/index.test.ts
```diff
@@ -1,559 +1,408 @@
...
-      supportedSignals.forEach((signal) => {
-        const metadata = {
-          datasets: { [baseDataset.id]: signal },
-        } as HDF5Metadata;
-
-        const supported = nxSpectrumDef.supportsEntity(nxDataGroup, metadata);
-        expect(supported).toBe(true);
+      const group = makeNxDataGroup('foo', {
+        signal: 'my_signal',
+        ids: { my_signal: datasetIntScalar.id }, // dataset type and shape doesn't matter
```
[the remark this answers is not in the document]
axelboc: Yep... and our philosophy so far for malformed NeXuS has been to avoid errors, and will be to allow users to debug them somehow as per #254. So yeah, we should add a check in `supportsEntity`. I'll think about it.

src/h5web/visualizations/index.test.ts
axelboc: Good catch. `isBaseType` actually checks for int, float or string, so it makes sense to have a test for it in scalar vis. I'm adding a check with a `datasetStrScalar` for completeness.

src/h5web/visualizations/index.test.ts
axelboc: Yeah, I hesitated. I felt like having the most important aspect on the left was actually easier to read where the variables are being initialised. It also fits with the order of the function call and parameters: dataset then type then shape.

src/h5web/providers/mock/data.ts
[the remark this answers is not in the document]
axelboc: Yeah, it adds a few formatting lines in the tests, but nothing crazy.

src/h5web/visualizations/index.test.ts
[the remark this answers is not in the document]
axelboc: Ok, so I've updated the logic in `supportsEntity` and I've added an extra test for each Nx vis. So we now have two tests with the correct interpretation: one with a dataset with a supported type and shape and one with an unsupported dataset.

loichuder (approved):
Looks good :eyes:

PR #263, opened 2020-10-23
NxImage now uses the axes attribute values and labels

### LINKED ISSUE #237: [Nexus] Heatmap vis with 2D @signal with @axes
Part of epic #196

loichuder:
Fix #237 

image

### DIFF
--- a/src/h5web/visualizations/containers/NxImageContainer.tsx
+++ b/src/h5web/visualizations/containers/NxImageContainer.tsx
@@ -1,12 +1,12 @@
 import React, { ReactElement, useState, useContext } from 'react';
 import { range } from 'lodash-es';
 import { HDF5SimpleShape } from '../../providers/models';
-import { useDatasetValue } from './hooks';
+import { useDatasetValues } from './hooks';
 import { assertGroup, isDataset } from '../../providers/utils';
 import DimensionMapper from '../../dimension-mapper/DimensionMapper';
 import { DimensionMapping } from '../../dimension-mapper/models';
 import { ProviderContext } from '../../providers/context';
-import { getAttributeValue, getLinkedEntity } from '../nexus/utils';
+import { getAttributeValue, getLinkedEntity, getNxAxes } from '../nexus/utils';
 import { VisContainerProps } from './models';
 import MappedHeatmapVis from '../heatmap/MappedHeatmapVis';
 import { assertStr } from '../shared/utils';
@@ -35,9 +35,11 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
     'x',
   ]);
 
-  const value = useDatasetValue(signalDataset.id);
+  const axes = getNxAxes(entity, metadata);
 
-  if (!value) {
+  const values = useDatasetValues({ signal: signalDataset.id, ...axes.ids });
+
+  if (!values || !values.signal) {
     return <></>;
   }
 
@@ -49,10 +51,12 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
         onChange={setMapperState}
       />
       <MappedHeatmapVis
-        value={value}
+        value={values.signal}
         valueLabel={signal}
         dims={dims}
         mapperState={mapperState}
+        axesLabels={axes.labels}
+        axesValues={values}
       />
     </>
   );
--- a/src/h5web/visualizations/heatmap/HeatmapVis.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.tsx
@@ -1,15 +1,14 @@
 import React, { useMemo } from 'react';
 import { format } from 'd3-format';
 import type ndarray from 'ndarray';
-import { range } from 'd3-array';
 import styles from './HeatmapVis.module.css';
 import ColorBar from './ColorBar';
 import Mesh from './Mesh';
 import TooltipMesh from '../shared/TooltipMesh';
 import PanZoomMesh from '../shared/PanZoomMesh';
 import VisCanvas from '../shared/VisCanvas';
-import { getDims } from './utils';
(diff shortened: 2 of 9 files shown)


src/stories/HeatmapVis.stories.tsx
```diff
@@ -108,12 +108,25 @@ export const CustomCoordinates = Template.bind({});
...
   dataArray,
   domain,
-  abscissas: Array(dataArray.shape[1] + 1)
-    .fill(0)
-    .map((x, i) => 100 + 10 * i),
-  ordinates: Array(dataArray.shape[0] + 1)
-    .fill(0)
-    .map((x, i) => -5 + 0.5 * i),
+  abscissaParams: {
+    values: Array(dataArray.shape[1])
```
loichuder, line 112: I now pass datasets of size `n` (and not `n+1`) to test the extension of custom coordinates.

src/h5web/visualizations/heatmap/utils.ts
```diff
@@ -17,3 +17,23 @@ export function generateCSSLinearGradient(
...
+  if (pixelCoordinates.length === nPixels) {
+    // Add the last edge assuming the pixelCoordinates are regularly spaced
+    const deltaCoord = pixelCoordinates[1] - pixelCoordinates[0];
+    return [
+      ...pixelCoordinates,
+      pixelCoordinates[pixelCoordinates.length - 1] + deltaCoord,
+    ];
+  }
+
+  return pixelCoordinates;
```
loichuder, line 38: I could move the `throw` here but I prefer to return `pixelCoordinates` and let the caller deal with it.

src/h5web/visualizations/containers/NxImageContainer.tsx
```diff
@@ -35,9 +39,22 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
...
+      return acc;
+    }
 
-  if (!value) {
+    const dataset = getLinkedEntity(entity, metadata, axis);
+    if (dataset && isDataset(dataset)) {
+      acc[axis] = dataset.id;
+    }
+    return acc;
+  }, {});
```
axelboc, line 53: Probably worth adding one or more utility functions to reduce duplication between the two containers, since this is pretty much the exact same code.

src/h5web/visualizations/heatmap/MappedHeatmapVis.tsx
```diff
@@ -57,6 +70,8 @@ function MappedHeatmapVis(props: Props): ReactElement {
       scaleType={scaleType}
       keepAspectRatio={keepAspectRatio}
       showGrid={showGrid}
+      abscissaParams={{ label: abscissaLabel, values: abscissas as number[] }}
+      ordinateParams={{ label: ordinateLabel, values: ordinates as number[] }}
```
axelboc, line 74: These `as number[]` have been bothering me, but I assume TS is being annoying... It's strange that it can't infer the type from the conditional assertion... Perhaps `assertOptionalArray<number>` would solve it?

src/h5web/visualizations/heatmap/utils.ts
axelboc: Hmm I feel like putting the throw here would be clearer, since it's the unsupported case that follows our three supported cases: `!pixelCoordinates`, `pixelCoordinates.length ===  numPixels + 1` and `pixelCoordinates.length ===  numPixels`.

I would actually add an if/return between the two blocks you already have, to identify clearly the second case I list above, `pixelCoordinates.length ===  numPixels + 1`. The final line can then become the `throw`, which should explain that the input pixel coordinates have an unsupported length.

src/stories/HeatmapVis.stories.tsx
axelboc: Should we test both?

axelboc (changes requested):

src/stories/HeatmapVis.stories.tsx
loichuder: Yes, I think so !

src/h5web/visualizations/heatmap/MappedHeatmapVis.tsx
loichuder: > These `as number[]` have been bothering me, but I assume TS is being annoying... It's strange that it can't infer the type from the conditional assertion...

That is exactly that.
> Perhaps `assertOptionalArray<number>` would solve it?

Good idea, I will try that

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -36,12 +35,15 @@ function HeatmapVis(props: Props): JSX.Element {
...
 
   const { rows, cols } = getDims(dataArray);
   const aspectRatio = keepAspectRatio ? cols / rows : undefined; // width / height <=> cols / rows
 
-  const { abscissas = range(cols + 1), ordinates = range(rows + 1) } = props;
+  const abscissas = getPixelEdges(abscissaParams.values, cols);
+  const ordinates = getPixelEdges(ordinateParams.values, rows);
 
   if (abscissas.length !== cols + 1) {
     throw new Error(
```
axelboc, line 49: These throws can be removed now, right? If `getPixelEdges` throws, then we don't know which axis is concerned, but it's not a big deal. The dimension of the axis is already a good hint.

axelboc (approved):
Niiice ✨

src/h5web/visualizations/heatmap/HeatmapVis.tsx
loichuder: :fishing_pole_and_fish:

PR #264, opened 2020-10-26
NxImage and NxSpectrum now supports NX groups with default attribute

### LINKED ISSUE #233: [Nexus] Default plot
Part of epic #196

- Update Nexus vis support function to look for `default` attribute (ignoring `NX_class`).
- Add vis wrapper similar to mapped vis components that follows `default` attributes until `NXdata` is found.

loichuder:
Fix #233 

image

The default attribute can:
- be a relative path in the file
- be an absolute path in the file
- points to another entity with a default attribute

### DIFF
--- a/src/h5web/App.test.tsx
+++ b/src/h5web/App.test.tsx
@@ -35,22 +35,15 @@ test('renders with root group selected', async () => {
 });
 
 test('switch between "display" and "inspect" modes', async () => {
-  const {
-    findByRole,
-    findByText,
-    getByRole,
-    queryByRole,
-    queryByText,
-  } = renderApp();
+  const { findByRole, getByRole, queryByRole } = renderApp();
 
   const inspectBtn = await findByRole('tab', { name: 'Inspect' });
   const displayBtn = getByRole('tab', { name: 'Display' });
 
   // Switch to "inspect" mode
   fireEvent.click(inspectBtn);
-
-  const noVisText1 = queryByText('Nothing to visualize');
-  expect(noVisText1).not.toBeInTheDocument();
+  const visSelector1 = queryByRole('tablist', { name: 'Visualization' });
+  expect(visSelector1).not.toBeInTheDocument();
 
   const groupIdRow1 = await findByRole('row', { name: /Entity ID/ });
   expect(groupIdRow1).toBeVisible();
@@ -61,8 +54,8 @@ test('switch between "display" and "inspect" modes', async () => {
   const groupIdRow2 = queryByRole('row', { name: /Entity ID/ });
   expect(groupIdRow2).not.toBeInTheDocument();
 
-  const noVisText2 = await findByText('Nothing to visualize');
-  expect(noVisText2).toBeVisible();
+  const visSelector2 = await findByRole('tablist', { name: 'Visualization' });
+  expect(visSelector2).toBeVisible();
 });
 
 test('toggle explorer sidebar', async () => {
--- a/src/h5web/providers/mock/data.ts
+++ b/src/h5web/providers/mock/data.ts
@@ -132,7 +132,7 @@ export function makeGroupHardLink(title: string, id: string): HDF5HardLink {
 /* ----------------- */
 /* ----- NEXUS ----- */
 
-export function makeNxDataGroup(
+export function makeNxData(
   id: string,
   opts: {
     signal: string;
@@ -150,6 +150,38 @@ export function makeNxDataGroup(
       ...(axes ? [makeStrAttr('axes', axes)] : []),
     ],
     Object.entries(ids).map(([...args]) => makeDatasetHardLink(...args))
+  );
+}
+
+export function makeNxEntry(
+  id: string,
+  opts: {
+    defaultPath?: string;
+    ids?: Record<string, HDF5Id>;
+  }
+): HDF5Group {
+  const { ids = {}, defaultPath } = opts;
+
+  return makeGroup(
+    id,
+    [makeStrAttr('NX_class', 'NXentry'), makeStrAttr('default', defaultPath)],
+    Object.entries(ids).map(([...args]) => makeGroupHardLink(...args))
+  );
+}
+
+export function makeNxRoot(
+  id: string,
+  opts: {
+    defaultPath?: string;
+    ids?: Record<string, HDF5Id>;
+  }
+): HDF5Group {
+  const { ids = {}, defaultPath } = opts;
+
+  return makeGroup(
+    id,
+    [makeStrAttr('NX_class', 'NXroot'), makeStrAttr('default', defaultPath)],
+    Object.entries(ids).map(([...args]) => makeGroupHardLink(...args))
   );
 }
 
--- a/src/h5web/visualizations/containers/NxImageContainer.tsx
+++ b/src/h5web/visualizations/containers/NxImageContainer.tsx
@@ -6,7 +6,12 @@ import { assertGroup, isDataset } from '../../providers/utils';
 import DimensionMapper from '../../dimension-mapper/DimensionMapper';
 import { DimensionMapping } from '../../dimension-mapper/models';
 import { ProviderContext } from '../../providers/context';
-import { getAttributeValue, getLinkedEntity, getNxAxes } from '../nexus/utils';
+import {
+  getAttributeValue,
+  getLinkedEntity,
+  getNxAxes,
+  getNxDataGroup,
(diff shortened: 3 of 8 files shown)


vasole:
The default attribute can only point to an NXdata group within the same entry except in the case of the top level (NXroot) where it has to point to an NXentry.

It cannot point to another group with a default attribute.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -71,3 +71,42 @@ export function getNxAxes(
...
+  if (!defaultEntity || !isGroup(defaultEntity)) {
+    return undefined;
+  }
+
+  const nextDefaultPath = getAttributeValue(defaultEntity, 'default');
+  if (typeof nextDefaultPath !== 'string') {
+    return defaultEntity;
+  }
+
+  return getDefaultEntity(defaultEntity, metadata);
```
axelboc, line 111: I don't quite get what this function does and how the recursion algorithm. I feel like it might make more sense if it were called `getNxDataGroup` and if its job was to return either an `NXdata` group or `undefined`. Here is what the algorithm would look like in my mind:

1. If input entity is `undefined` or no a group, return `undefined`.
1. If entity is `NXdata` group, return entity.
1. If entity has no `default` attribute, return `undefined`; else recurse with target default entity.

axelboc (changes requested):

src/h5web/visualizations/nexus/utils.ts
loichuder: Given @vasole comment, I can simplify this function.

src/h5web/App.test.tsx
```diff
@@ -35,22 +35,15 @@ test('renders with root group selected', async () => {
...
   const inspectBtn = await findByRole('tab', { name: 'Inspect' });
   const displayBtn = getByRole('tab', { name: 'Display' });
 
   // Switch to "inspect" mode
   fireEvent.click(inspectBtn);
-
-  const noVisText1 = queryByText('Nothing to visualize');
-  expect(noVisText1).not.toBeInTheDocument();
+  const visSelector1 = queryByRole('tablist', { name: 'Visualization' });
+  expect(visSelector1).not.toBeInTheDocument();
```
loichuder, line 46: Funnily, I had to change this as the root node has now a visualization :slightly_smiling_face:

loichuder:
In the end, the algorithm is more permissive than the specification (no checks to see if the default attribute points to an NXentry for a NXroot or if we are within the same entry) as it allows for lighter code.

We support then the specification but do not enforce it for now as discussed in #254 and previous PRs.

src/h5web/providers/mock/data.ts
```diff
@@ -153,6 +153,38 @@ export function makeNxDataGroup(
   );
 }
 
+export function makeNxEntry(
```
axelboc, line 156: I like this naming convention. Mind renaming `makeNxDataGroup` to `makeNxData` ?

axelboc: Alternatively, we could also have a generic `makeNxGroup` that accepts a class and a record of IDs, which would be called by `makeNxData` and perhaps `makeNxGroupWithDefault`. I'm fine with the current implementation, though ; just a thought.

loichuder: Meh, not sure the abstraction is worth it

src/h5web/visualizations/containers/NxImageContainer.tsx
```diff
@@ -16,9 +21,16 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
   assertGroup(entity);
 
   const { metadata } = useContext(ProviderContext);
-  const signal = getAttributeValue(entity, 'signal');
+
+  const NxData = getNxDataGroup(entity, metadata);
+  if (!NxData) {
```
axelboc, line 26: These variables (here and in the other container) need to start with a lowercase.

src/h5web/visualizations/index.test.ts
```diff
@@ -29,6 +32,19 @@ const datasetStr1D = makeSimpleDataset('dataset_str_1d', stringType, [5]);
 const datasetStr2D = makeSimpleDataset('dataset_str_2d', stringType, [5, 3]);
 const datasetFlt3D = makeSimpleDataset('dataset_flt_3d', intType, [5, 3, 1]);
 
+const NxImageDataset = makeSimpleDataset(
```
axelboc, line 35: Idem, first letter should be lower case.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -71,3 +71,41 @@ export function getNxAxes(
...
+    : defaultPath;
+
+  const defaultEntity = relativePath.split('/').reduce<HDF5Entity | undefined>(
+    (previousEntity, currentComponent) => {
+      if (!previousEntity) {
+        return undefined;
+      }
+      if (!isGroup(previousEntity)) {
+        return previousEntity;
+      }
```
axelboc, line 104: Don't you end up in a loop if the last entity in the path is a dataset? Shouldn't you return `undefined` in this case instead?

src/h5web/visualizations/index.test.ts
```diff
@@ -304,6 +320,189 @@ describe('Visualization definitions', () => {
...
+        ids: { my_signal: NxSpectrumDataset.id },
+      });
+
+      const entry = makeNxEntry('entry', {
+        ids: { spectrum: group.id }, // missing default attr
+      });
+
+      const root = makeNxRoot('root', {
+        ids: { entry: entry.id },
+      });
```
axelboc, line 379: The root group isn't needed for this test.

src/h5web/visualizations/index.test.ts
```diff
@@ -304,6 +320,189 @@ describe('Visualization definitions', () => {
...
+      const metadata = makeMetadata({
+        root: root.id,
+        datasets: [NxSpectrumDataset],
+        groups: [root, group, entry],
+      });
+
+      expect(supportsEntity(entry, metadata)).toBe(false);
+    });
+
+    it('should support a NXroot with a default attribute pointing to a valid NXentity with a valid default (relative paths)', () => {
```
axelboc, line 415: move above the `not support` tests above?

src/h5web/visualizations/index.test.ts
```diff
@@ -304,6 +320,189 @@ describe('Visualization definitions', () => {
...
+      const group = makeNxDataGroup('spectrum', {
+        signal: 'my_signal',
+        ids: { my_signal: NxSpectrumDataset.id },
+      });
+
+      const intermediateGroup = makeGroup(
+        'intermediate',
+        [],
+        [makeGroupHardLink('spectrum', group.id)]
+      );
```
axelboc, line 487: Perhaps the intermediate group is not needed for this test.

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
```diff
@@ -16,9 +21,16 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
   assertGroup(entity);
 
   const { metadata } = useContext(ProviderContext);
-  const signal = getAttributeValue(entity, 'signal');
+
+  const NxData = getNxDataGroup(entity, metadata);
+  if (!NxData) {
+    throw new Error('NXdata group not found');
+  }
+  assertGroup(NxData);
```
axelboc, line 29: This assertion is not needed since the function is already typed to return a group.

axelboc (changes requested):
A couple of tweaks and a potential bug.

src/h5web/visualizations/nexus/utils.ts
loichuder: Not that I know of. Let me break it down:
- the `reduce` will end up with `defaultEntity` being a dataset
- It will then call `getNxDataGroup(defaultEntity) 
- The guard clause at the beginning of the function `if(!entity | !isGroup(entity)` will make it return `undefined`. End of the story.

If I return `undefined`, the behaviour will be exactly the same except that `defaultEntity` will be `undefined` instead of a dataset.

src/h5web/visualizations/index.test.ts
loichuder: True. I made a dedicated tests for "deep" paths.

axelboc (approved):
Nice use of `reduce` by the way ;)

PR #266, opened 2020-10-27
Fix non-integer ticks when abscissa/ordinate values were not supplied

loichuder:
I thought a refactoring would be needed to fix this but I found a workaround.

That being said, I am not extremely happy about the current state of things. To be discussed together when we have time.

### DIFF
--- a/src/h5web/visualizations/heatmap/HeatmapVis.tsx
+++ b/src/h5web/visualizations/heatmap/HeatmapVis.tsx
@@ -69,13 +69,13 @@ function HeatmapVis(props: Props): JSX.Element {
         abscissaConfig={{
           domain: abscissaDomain,
           showGrid,
-          isIndexAxis: !abscissas,
+          isIndexAxis: !abscissaParams.values,
           label: abscissaParams.label,
         }}
         ordinateConfig={{
           domain: ordinateDomain,
           showGrid,
-          isIndexAxis: !ordinates,
+          isIndexAxis: !ordinateParams.values,
           label: ordinateParams.label,
         }}
         aspectRatio={aspectRatio}
--- a/src/h5web/visualizations/line/LineVis.tsx
+++ b/src/h5web/visualizations/line/LineVis.tsx
@@ -68,7 +68,7 @@ function LineVis(props: Props): ReactElement {
         abscissaConfig={{
           domain: abscissaDomain,
           showGrid,
-          isIndexAxis: !abscissas,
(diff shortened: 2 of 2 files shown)


src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -69,13 +69,13 @@ function HeatmapVis(props: Props): JSX.Element {
         abscissaConfig={{
           domain: abscissaDomain,
           showGrid,
-          isIndexAxis: !abscissas,
+          isIndexAxis: !abscissaParams.values,
```
loichuder, line 72: `abscissas` is never `undefined` as it is set to `range(dataArray.size)` when destructuring `abscissaParams.values`.

Changing the condition to `abscissaParams.values` restores the check validity.

axelboc: Hmm yeah I get you. You could store this condition in a constant at the top of the component and extract the `range` call out of `getPixelEdges`. Is this what you initially had in mind as a refactoring?

```ts
const isAbscissaIndexAxis = !abscissaParams.values;
const abscissas = isAbscissaIndexAxis ? range(cols + 1) : getPixelEdges(abscissaParams.values, cols);
```

It's still not great, is it? We could also pass `isIndexAxis` as part of each axis' params object prop, but this would complexify the public API...

I like your current solution the best for now. Unless you had something else in mind?

axelboc (approved):

src/h5web/visualizations/heatmap/HeatmapVis.tsx
loichuder: > Hmm yeah I get you. You could store this condition in a constant at the top of the component and extract the `range` call out of `getPixelEdges`. Is this what you initially had in mind as a refactoring?
> 
> ```ts
> const isAbscissaIndexAxis = !abscissaParams.values;
> const abscissas = isAbscissaIndexAxis ? range(cols + 1) : getPixelEdges(abscissaParams.values, cols);
> ```
> 
> It's still not great, is it? 

I followed the exact same line of thought ! I feel we can do better but did not manage to find a satisfying solution on my own.

> We could also pass `isIndexAxis` as part of each axis' params object prop, but this would complexify the public API...
> 
> I like your current solution the best for now. Unless you had something else in mind?

I tried a few things here and there but in the end, I think this fix is the most straightforward solution

PR #267, opened 2020-10-27
Removed AxisInfo (replaced by AxisConfig and fallbacks)

loichuder:
Also changed `Axis`


src/h5web/visualizations/shared/AxisSystem.tsx
```diff
@@ -36,15 +36,6 @@ function AxisSystem(props: Props): JSX.Element {
...
 
-  // Restrain ticks scales to visible domains
-  const xTicksScale = abscissaInfo.scaleFn();
-  xTicksScale.domain(xVisibleDomain);
-  xTicksScale.range([0, width]);
-
-  const yTicksScale = ordinateInfo.scaleFn();
-  yTicksScale.domain(yVisibleDomain);
-  yTicksScale.range([height, 0]);
-
```
loichuder, line 47: Scale construction was moved inside `Axis` to:
- Avoid a fallback of `scaleType` in `AxisSystem` **and** `Axis`
- Avoid passing `scale` to `Axis` as it already depends on `domain` and `canvasSize`

I had to introduce a new flag `flipAxis` to deal with the inverted range of yAxis but I think it might be handy later.

axelboc: Nice! I like it 💯

axelboc (approved):
Looks great! It cuts down a lot of complexity. I like the renaming of `getCanvasScale` too.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -102,12 +109,15 @@ export function getValueToIndexScale(
   return valueToIndexScale;
 }
 
-export function getAxisScale(info: AxisInfo, rangeSize: number): AxisScale {
-  const { scaleFn, domain } = info;
+export function getCanvasScale(
```
loichuder, line 112: I felt this renaming was needed: the scale computed by this function relates the data values to the **canvas** coordinates.

The scale of `Axis` are different as they map the data values to the axis length,

PR #268, opened 2020-10-27
Generate mock data in JS

### LINKED ISSUE #253: Mock data generation

axelboc:
Fix #253 

I use some of the utilities we created for testing the `supportsEntity` functions (plus some new ones), to generate a mock `HDF5Metadata` object and mock dataset values.

This allows me to remove `data.json` and simplify a number of things in the Provider. For instance, the `MockAPI` became so simple that I removed it completely and inlined it as a simple object in the provider.

image

The new mock metadata tree looks like this:

image


src/h5web/providers/mock/utils.ts
```diff
@@ -1,15 +1,13 @@
-import mockData from './data.json';
-import type { HDF5SimpleShape } from '../models';
+import { mockMetadata } from './data';
+import { assertSimpleShape } from '../utils';
 
-const datasets = Object.values(mockData.datasets);
+export function getMockDatasetDims(name: string): number[] {
```
axelboc, line 4: Retrieving a dataset's value in the stories is now trivial - e.g. `mockValues.twoD`, so I removed `getMockedDataset`. However, I had to replace with `getMockDatasetDims` to help retrieve a dataset's dimensions based on its name.

src/demo-app/DemoApp.tsx
```diff
@@ -18,7 +17,7 @@ function DemoApp(): JSX.Element {
     <Router>
       <Switch>
         <Route path="/mock">
-          <Suspense fallback={<Loader />}>
+          <Suspense fallback={<></>}>
```
axelboc, line 20: Showing a loader while fetching the metadata looked weird. Given how fast the fetch is, I think a blank page is cleaner.

src/h5web/App.test.tsx
```diff
@@ -81,79 +81,95 @@ test('toggle explorer sidebar', async () => {
...
-  expect(datasetBtn).not.toHaveAttribute('aria-expanded'); // A dataset cannot be expanded
+  expect(datasetBtn).not.toHaveAttribute('aria-expanded'); // a dataset cannot be expanded
 
   const scalarTab = await findByRole('tab', { name: 'Scalar' });
   expect(scalarTab).toBeVisible();
   expect(scalarTab).toHaveAttribute('aria-selected', 'true');
+  expect(getByText('0')).toBeVisible();
+
+  fireEvent.click(getByRole('treeitem', { name: 'scalar_str' }));
+  expect(await findByText('foo')).toBeVisible();
```
axelboc, line 146: I check both an int and a string scalar.

src/h5web/App.test.tsx
```diff
@@ -81,79 +81,95 @@ test('toggle explorer sidebar', async () => {
...
+});
+
+test('inspect scalar dataset', async () => {
```
axelboc, line 149: Just moving this test up and "simplifying" it by inspecting a scalar dataset rather than a 1D dataset.

src/h5web/BreadcrumbsBar.module.css
```diff
@@ -17,6 +17,7 @@
   padding: 0 0.25rem;
   font-size: inherit;
   font-weight: inherit;
+  line-height: 1.3; /* fix cropping of glyphs */
```
axelboc, line 20: image

src/h5web/providers/context.ts
```diff
@@ -4,7 +4,7 @@ import type { HDF5Id, HDF5Value, HDF5Metadata } from './models';
 export abstract class ProviderAPI {
   abstract domain: string;
   abstract async getMetadata(): Promise<HDF5Metadata>;
-  abstract async getValue(id: HDF5Id): Promise<HDF5Value>;
+  abstract async getValue(id: HDF5Id): Promise<HDF5Value | undefined>;
```
axelboc, line 7: Making the API's `getValue` signature a bit more accurate. It doesn't change anything in practice.

src/h5web/providers/mock/MockProvider.tsx
```diff
@@ -9,13 +10,18 @@ interface Props {
...
-    setApi(new MockApi(domain));
-  }, [domain]);
-
-  return <Provider api={api}>{children}</Provider>;
+  return (
+    <Provider
+      api={{
+        domain,
+        getMetadata: async () => mockMetadata,
+        getValue: async (id: HDF5Id) => (mockValues as HDF5Values)[id],
```
axelboc, line 19: The API is now so simple that I can just inline it here! Note that I cast `mockValues` here instead of typing `mockValues` in `data.ts`, so that I can just write `mockValues.twoD` in the stories and get a `number[][]`.

src/h5web/providers/mock/api.ts
```diff
@@ -1,58 +0,0 @@
...
-import mockData from './data.json';
-
-function addMissingProperties<T extends HDF5Entity>(
-  collection: HDF5Collection
-) {
-  return ([id, entity]: [string, Partial<T> & { value?: HDF5Value }]) => [
-    id,
-    { id, collection, ...entity },
-  ];
-}
```
axelboc, line 27: The mock metadata now includes these missing properties.

src/h5web/providers/utils.ts
```diff
@@ -75,6 +75,14 @@ export function assertGroup(entity: HDF5Entity): asserts entity is HDF5Group {
   }
 }
 
+export function assertSimpleShape(
+  shape: HDF5Shape
+): asserts shape is HDF5SimpleShape {
+  if (shape.class !== HDF5ShapeClass.Simple) {
+    throw new Error('Expected simple shape');
+  }
+}
```
axelboc, line 84: Used in `getMockDatasetDims`. I thought it best to place it here, next to some of the other assertion function.

src/h5web/visualizations/containers/RawVisContainer.tsx
```diff
@@ -9,7 +9,7 @@ function RawViscontainer(props: VisContainerProps): ReactElement {
   assertDataset(entity);
 
   const value = useDatasetValue(entity.id);
-  if (!value) {
+  if (value === undefined) {
```
axelboc, line 12: This bloody regression was back again... 😄 The scalar (and raw) vis was not display datasets with falsy values like `0` or `""`. That's why I created a mocked scalar entity with a value of `0`, and updated a feature test to look for it.

src/stories/HeatmapVis.stories.tsx
```diff
@@ -2,24 +2,18 @@ import React, { ReactElement, useState } from 'react';
...
-const dataset = getMockedDataset<number[][]>('/nD/twoD');
-const values = dataset.value.flat(Infinity) as number[];
-
-const transposedArray = ndarray<number>(values, dataset.dims).transpose(1, 0); // makes for a nicer-looking heatmap
-// Work with the real array and not the transposed view
-const dataArray = ndarray<number>([], transposedArray.shape);
-assign(dataArray, transposedArray);
+// Prepare 2D data array
+const values = mockValues.twoD.flat(Infinity) as number[];
+const dataArray = ndarray<number>(values, getMockDatasetDims('twoD'));
```
axelboc, line 16: We no longer need to transpose the dataset, since I made sure it had a horizontal aspect ratio.

loichuder (approved):
Nice !

PR #270, opened 2020-10-28
Upgrade Create React App to v4

axelboc:
Here is a summary of what has changed:

## [CRA 4](https://github.com/facebook/create-react-app/releases/tag/v4.0.0)

- drops Node 8
- Jest's `resetMocks` option is enabled by default
- `EXTEND_ESLINT` env var is no longer needed

It also brings a lot of major dependency upgrades:

### [ESLint 7](https://eslint.org/blog/2020/05/eslint-v7.0.0-released)

- drops Node 8
- ignores nested `node_modules` by default
- allows additional comments in disable comments - e.g. `// eslint-disable-line no-console - this is why`

Our config, `@wsmd/eslint-config` hasn't upgraded to ESLint v7 and seems unmaintained... We're going to have to replace it, perhaps with https://github.com/ljosberinn/eslint-config-galex. In the meantime, I've removed it completely and I now extend CRA's built-in config. I've also cleaned up some of the overrides that we had for `@wsmd/eslint-config`.

### [Jest 25](https://jestjs.io/blog/2020/01/21/jest-25)

- drops Node 6! 😱 
- faster and better console outputs
- JSDOM from 11 to 15

### [Jest 26](https://jestjs.io/blog/2020/05/05/jest-26)

- smaller to install
- JSDOM 16
- mocked `Date` object based on fake timer clock
- `--detect-leaks`

### [JSDOM 15/16](https://github.com/jsdom/jsdom/blob/master/Changelog.md)

- so far we were actually using JSDOM 14 thanks to `--env=jsdom-fourteen`
- drops Node 8
- more up-to-date with latest DOM standard

### [Typescript 4](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/)

- variadic tuples, labeled tuples, etc.
- faster project startup with partial semantic mode
- smarter auto-imports for new dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test --env=jsdom-fourteen",
+    "test": "react-scripts test",
     "lint": "run-p lint:*",
     "lint:eslint": "eslint \"**/*.{ts,tsx,js,jsx}\" --max-warnings=0",
     "lint:tsc": "tsc --noEmit",
@@ -59,7 +59,7 @@
     "zustand": "^3.1.2"
   },
   "devDependencies": {
-    "@babel/core": "^7.11.6",
+    "@babel/core": "^7.12.3",
     "@storybook/addon-actions": "^6.0.21",
     "@storybook/addon-essentials": "^6.0.21",
     "@storybook/addon-links": "^6.0.21",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -74,37 +74,34 @@
...
     "worker-loader": "^3.0.3"
   },
   "jest": {
-    "resetMocks": true,
     "transformIgnorePatterns": [
       "/node_modules/(?!lodash-es|react-use-comlink)/.*"
     ],
     "moduleNameMapper": {
-      "worker-loader*": "<rootDir>/src/moduleMock.js",
-      "troika": "<rootDir>/src/moduleMock.js"
```
axelboc, line 107: Forgot to remove this when we removed `drei`.

package.json
```diff
@@ -116,8 +113,7 @@
     ],
     "development": [
       "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
```
axelboc, line 120: We're not using on Safari in dev, so I've removed it.

src/h5web/visualizations/shared/hooks.ts
```diff
@@ -41,7 +41,7 @@ export function useMappedArray<T>(
 export const useDomain = createMemo(getDomain);
 
 export function useFrameRendering(): void {
-  const [, setNum] = useState();
+  const [, setNum] = useState(0);
```
axelboc, line 44: Random error, probably from the patch upgrade of `@types/react`.

axelboc:
I'll upgrade React to v17 separately.

loichuder:
Did you try to export the `lib` with the update ?

axelboc:
Ah good catch, I'm getting a typing error that I'm not seeing when building the app. Upgrading TS and a few other dependencies in `packages/package.json` doesn't fix it. I'll investigate.

axelboc:
Very strange... Oh well, I added an explicit type and it fixed it. 🤷‍♂️

loichuder:
What was the problem exactly ?

axelboc:
`err` has implicit `any` type on the `errorMessage` prop of `AsyncResourceContent`, in `Visualizer.tsx`. But no error when building the app itself or in VSCode; only when building the `lib` package.

loichuder (approved):

PR #274, opened 2020-10-29
Nx Axis labels are now determined from long_name and units attributes

### LINKED ISSUE #265: [Nexus] Support units for NXdata with `axes`
Part of #196 

Units should be displayed in the axis labels

loichuder:
Fix #265 

image

I added the support of `long_name` as well as it was the same logic.

Also updated the mock data with some `long_name` and `units` attributes.

### DIFF
--- a/src/h5web/providers/mock/data.ts
+++ b/src/h5web/providers/mock/data.ts
@@ -308,15 +308,30 @@ export const mockMetadata = makeMetadata({
     ),
   ],
   datasets: [
-    makeSimpleDataset('oneD', intType, [41]),
+    makeSimpleDataset(
+      'oneD',
+      intType,
+      [41],
+      [makeStrAttr('units', 'arb. units')]
+    ),
     makeSimpleDataset('twoD', intType, [20, 41]),
     makeSimpleDataset('threeD', intType, [9, 20, 41]),
-    makeSimpleDataset('fourD', intType, [3, 9, 20, 41]),
+    makeSimpleDataset(
+      'fourD',
+      intType,
+      [3, 9, 20, 41],
+      [makeStrAttr('long_name', 'Interference fringes')]
+    ),
     makeDataset('raw', compoundType, scalarShape),
     makeDataset('scalar_int', intType, scalarShape),
     makeDataset('scalar_str', stringType, scalarShape),
-    makeSimpleDataset('X', intType, [41]),
-    makeSimpleDataset('Y', intType, [20]),
+    makeSimpleDataset('X', intType, [41], [makeStrAttr('units', 'nm')]),
+    makeSimpleDataset(
+      'Y',
+      intType,
+      [20],
+      [makeStrAttr('units', 'deg'), makeStrAttr('long_name', 'Angle (degrees)')]
+    ),
   ],
   datatypes: [makeDatatype('datatype', compoundType)],
 });
--- a/src/h5web/visualizations/containers/NxImageContainer.tsx
+++ b/src/h5web/visualizations/containers/NxImageContainer.tsx
@@ -1,7 +1,6 @@
 import React, { ReactElement, useState, useContext } from 'react';
 import { range } from 'lodash-es';
 import { HDF5SimpleShape } from '../../providers/models';
-import { useDatasetValues } from './hooks';
 import { assertGroup, isDataset } from '../../providers/utils';
 import DimensionMapper from '../../dimension-mapper/DimensionMapper';
 import { DimensionMapping } from '../../dimension-mapper/models';
@@ -9,12 +8,16 @@ import { ProviderContext } from '../../providers/context';
 import {
   getAttributeValue,
   getLinkedEntity,
-  getNxAxes,
   getNxDataGroup,
+  getNxAxisMapping,
+  getLinkedDatasets,
+  getNxAxesParams,
+  getDatasetLabel,
 } from '../nexus/utils';
 import { VisContainerProps } from './models';
 import MappedHeatmapVis from '../heatmap/MappedHeatmapVis';
 import { assertStr } from '../shared/utils';
+import { useDatasetValues } from './hooks';
 
 function NxImageContainer(props: VisContainerProps): ReactElement {
   const { entity } = props;
@@ -27,9 +30,9 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
     throw new Error('NXdata group not found');
   }
 
-  const signal = getAttributeValue(nxDataGroup, 'signal');
-  assertStr(signal);
-  const signalDataset = getLinkedEntity(nxDataGroup, metadata, signal);
+  const signalName = getAttributeValue(nxDataGroup, 'signal');
+  assertStr(signalName);
+  const signalDataset = getLinkedEntity(signalName, nxDataGroup, metadata);
 
   if (!signalDataset || !isDataset(signalDataset)) {
     throw new Error('Signal dataset not found');
@@ -40,17 +43,27 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
     throw new Error('Expected signal dataset with at least two dimensions');
   }
 
-  const [mapperState, setMapperState] = useState<DimensionMapping>([
+  const [dimensionMapping, setDimensionMapping] = useState<DimensionMapping>([
     ...range(dims.length - 2).fill(0),
     'y',
     'x',
   ]);
 
-  const axes = getNxAxes(nxDataGroup, metadata);
+  const nxAxisMapping = getNxAxisMapping(nxDataGroup);
+  const axisDatasets = getLinkedDatasets(
+    nxAxisMapping.filter((v): v is string => !!v),
+    nxDataGroup,
+    metadata
+  );
+  const datasetValues = useDatasetValues({
+    [signalName]: signalDataset.id,
+    ...Object.fromEntries(
+      Object.entries(axisDatasets).map(([axis, dataset]) => [axis, dataset.id])
+    ),
+  });
+  const nxAxesParams = getNxAxesParams(axisDatasets, datasetValues);
 
-  const values = useDatasetValues({ signal: signalDataset.id, ...axes.ids });
-
-  if (!values || !values.signal) {
+  if (!datasetValues || !datasetValues[signalName]) {
     return <></>;
   }
 
@@ -58,16 +71,16 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
     <>
       <DimensionMapper
         rawDims={dims}
-        mapperState={mapperState}
-        onChange={setMapperState}
+        mapperState={dimensionMapping}
+        onChange={setDimensionMapping}
       />
       <MappedHeatmapVis
-        value={values.signal}
-        valueLabel={signal}
+        value={datasetValues[signalName]}
+        title={getDatasetLabel(signalDataset, signalName)}
         dims={dims}
-        mapperState={mapperState}
-        axesLabels={axes.labels}
-        axesValues={values}
+        dimensionMapping={dimensionMapping}
+        axisMapping={nxAxisMapping}
+        axesParams={nxAxesParams}
       />
     </>
   );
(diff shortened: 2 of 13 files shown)


src/h5web/providers/mock/data.ts
```diff
@@ -308,15 +308,30 @@ export const mockMetadata = makeMetadata({
...
     makeDataset('scalar_int', intType, scalarShape),
     makeDataset('scalar_str', stringType, scalarShape),
-    makeSimpleDataset('X', intType, [41]),
-    makeSimpleDataset('Y', intType, [20]),
+    makeSimpleDataset('X', intType, [41], [makeStrAttr('units', 'nm')]),
+    makeSimpleDataset(
+      'Y',
+      intType,
+      [20],
+      [makeStrAttr('units', 'deg'), makeStrAttr('long_name', 'Angle (degrees)')]
```
loichuder, line 333: This is to verify that `long_name` takes precedence over `units`

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -46,30 +46,48 @@ export function getLinkedEntity(
...
+    labels: Object.fromEntries(
+      axesDatasets.map(([axis, dataset]) => [
+        axis,
+        getDatasetLabel(dataset, axis),
+      ])
+    ),
+    ids: Object.fromEntries(
+      axesDatasets.map(([axis, dataset]) => [axis, dataset.id])
+    ),
+  };
```
loichuder, line 90: `axesNames` are the dataset names as stored in `@axes` (previously called `axesLabels`)

`axesLabels` are now an object with dataset names as keys and labels (which can now be different) as values.

src/h5web/visualizations/shared/AxisSystem.module.css
```diff
@@ -48,6 +48,7 @@
   font-family: inherit;
   font-size: inherit;
   font-weight: 600;
+  text-anchor: middle;
```
loichuder, line 51: To center axis labels with respect to axes

src/h5web/visualizations/heatmap/MappedHeatmapVis.tsx
```diff
@@ -12,14 +12,16 @@ interface Props {
...
+  axesLabels?: Record<string, string>;
   axesValues?: Record<string, HDF5Value>;
 }
 
 function MappedHeatmapVis(props: Props): ReactElement {
   const {
     value,
-    axesLabels = [],
+    axesNames = [],
+    axesLabels = {},
```
axelboc, line 24: Maybe we could pass `axesParams={{ some_axis_name: { label: ..., values: ...}, some_other_axis_name: {...} }}` and pick the right params object below? This would replace three props with one.

src/h5web/visualizations/nexus/utils.ts
axelboc: I feel like we're starting to have a lot of variables objects/arrays that are coupled to one another. I think we need a better data structure. Maybe this function would return the `axesParams` object I mention above: 

```ts
interface AxisParams {
  name: string;
  label: string;
  values: HDF5Value | undefined;
}

type AxesParams = Record<id, AxisParams>;
```

Not sure exactly, I'd have to look at it more, but do you get my point?

axelboc: I could also imagine the `useDatasetValues` hook being refactored to have a signature like:

```ts
interface VisParams {
  title?: string;
  values: HDF5Value | undefined;
  axesParams: AxesParams;
}

function useVisParams(signalId: string, axesParams: AxesParams): VisParams | undefined {...}
```

The mapped vis would then pick the correct axis params object for the abscissa/ordinate based on `mapperState` as they do now.

Anyway, just some thoughts. If you prefer to do this later, let me know, but I think this PR is a good opportunity.

src/h5web/visualizations/containers/hooks.ts
```diff
@@ -17,3 +27,47 @@ export function useDatasetValues(
...
+  return Object.entries(nxDatasets).reduce<Record<string, AxisParams>>(
+    (acc, [axis, dataset]) => {
+      const values = dsetValues[axis];
+      assertArray<number>(values);
+      acc[axis] = { values, label: getDatasetLabel(dataset, axis) };
+      return acc;
+    },
+    {}
+  );
+}
```
loichuder, line 73: The new hook that fetches values and labels for axes and the signal altogether. The aim is to get `AxisParams` all the way down to the lower-level Vis.

src/h5web/visualizations/line/MappedLineVis.tsx
```diff
@@ -47,9 +48,7 @@ function MappedLineVis(props: Props): ReactElement {
     scaleType
   );
 
-  const abscissaLabel = mapperState && axesLabels[mapperState.indexOf('x')];
-  const abscissas = abscissaLabel && axesValues[abscissaLabel];
-  assertOptionalArray<number>(abscissas);
```
loichuder, line 52: The `assert` is no longer needed as it was done beforehand in the hook `useNxAxisParams`.

loichuder (commented):
Happy to get feedback on naming.

axelboc (commented):
Ok, I think the idea of creating `AxisParams` objects is working, but now I feel that `useNxAxisParams` should be split into three separate functions/hooks:

1. a function to generate a record of datasets from the axes names: `datasets: Record<string, HDF5Dataset>`);
2. a hook to fetch the values of the signal and axes datasets into a `Record`: `datasetValues: Record<string, HDF5Value | undefined>`;
3. a function to put together the `AxesParams` object from the results of 1. and 2.

Dealing with the signal in 1. and 3. is too confusing I think. I would keep it separate, even if it means calling `getLinkedEntry` and `getDatasetLabel` again.

If you do this, for 1., ignore my comment about creating an `entries` array instead of a `Record`. You could still use `Object.fromEntries(blah.map(...).filter(...))` instead of `reduce`, though.

src/h5web/visualizations/containers/NxImageContainer.tsx
```diff
@@ -1,20 +1,20 @@
...
 import { assertGroup, isDataset } from '../../providers/utils';
 import DimensionMapper from '../../dimension-mapper/DimensionMapper';
 import { DimensionMapping } from '../../dimension-mapper/models';
 import { ProviderContext } from '../../providers/context';
 import {
   getAttributeValue,
   getLinkedEntity,
-  getNxAxes,
   getNxDataGroup,
+  getNxAxesNames,
```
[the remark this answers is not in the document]
loichuder: Excellent idea !

src/h5web/visualizations/containers/NxImageContainer.tsx
```diff
@@ -62,12 +68,12 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
         onChange={setMapperState}
       />
       <MappedHeatmapVis
-        value={values.signal}
-        valueLabel={signal}
+        value={signal.values}
+        valueLabel={signal.label}
```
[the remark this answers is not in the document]
loichuder: I am biased as it is called `label` in matplotlib so I find that it is the most appropriate word...

src/h5web/visualizations/heatmap/MappedHeatmapVis.tsx
```diff
@@ -76,8 +72,8 @@ function MappedHeatmapVis(props: Props): ReactElement {
       scaleType={scaleType}
       keepAspectRatio={keepAspectRatio}
       showGrid={showGrid}
-      abscissaParams={{ label: abscissaLabel, values: abscissas }}
-      ordinateParams={{ label: ordinateLabel, values: ordinates }}
+      abscissaParams={abscissaName ? axesParams[abscissaName] : undefined}
+      ordinateParams={ordinateName ? axesParams[ordinateName] : undefined}
```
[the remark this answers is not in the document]
loichuder: Neh: `ordinateName && axesParams[ordinateName]` returns the string `""` when `ordinateName` is `""`...

src/h5web/visualizations/containers/hooks.ts
```diff
@@ -17,3 +27,47 @@ export function useDatasetValues(
...
+
+  return Object.entries(nxDatasets).reduce<Record<string, AxisParams>>(
+    (acc, [axis, dataset]) => {
+      const values = dsetValues[axis];
+      assertArray<number>(values);
+      acc[axis] = { values, label: getDatasetLabel(dataset, axis) };
+      return acc;
+    },
+    {}
+  );
```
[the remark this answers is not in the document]
loichuder: I am not so sure: that means using a `map` first on a `nxNames` yielding an array of `[keys, values]` of the same size. But the resulting object can have less entries than the elements of `nxNames` (a entry should not be created if the corresponding axis is `undefined` or links to an entity that is not a dataset).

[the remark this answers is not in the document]
loichuder: Oh but then I can do the `filter` **after** the `map`... Ok understood :+1:

src/h5web/visualizations/heatmap/MappedHeatmapVis.tsx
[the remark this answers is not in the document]
axelboc: ah right

src/h5web/visualizations/containers/NxImageContainer.tsx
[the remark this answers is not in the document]
axelboc: Ah hmm weird. The good thing about `title` is that it can stand on its own (as opposed to `label`/`valueLabel`) -- it's clear that it's the plot's title and not an axis label. I don't know, I feel like `title` is clearer, but I trust you.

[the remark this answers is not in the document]
loichuder: Sorry, I thought we were talking about the `LineVis`. You are right, `title` makes much more sense here for the `HeatmapVis`

loichuder:
Updated. Not entirely convinced but let me know your opinion ^^

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -106,3 +89,56 @@ export function getNxDataGroup(
...
+    return `${datasetName} (${units})`;
+  }
+
+  return datasetName;
+}
+
+export function getDatasetsFromMapping(
+  nxDataGroup: HDF5Group,
+  metadata: HDF5Metadata,
+  nxAxisMapping: (string | undefined)[]
```
axelboc, line 113: I would put the mapping as the first argument. Also, I think the function should be named either:

- `getDatasetsFromAxisMapping` (with a parameter named, simply, `mapping`), or
- `getLinkedDatasets` (with a parameter named `names` which would only contain strings, and would therefore be called with `getLinkedDatasets(nxAxisMapping.filter(val => !!val), ...)`.

I reckon the second option might be better, because it keeps the utility function generic. I know it adds a bit of complexity back into the containers, but it remains readable.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -106,3 +89,56 @@ export function getNxDataGroup(
...
+    .map((axis) => {
+      if (!axis) {
+        return undefined;
+      }
+
+      const dataset = getLinkedEntity(nxDataGroup, metadata, axis);
+      if (!dataset || !isDataset(dataset)) {
+        return undefined;
+      }
+      return [axis, dataset];
```
axelboc, line 125: You can shorten this with a ternary.

src/h5web/visualizations/containers/NxImageContainer.tsx
```diff
@@ -40,34 +43,44 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
...
+    metadata,
+    nxAxisMapping
+  );
+  const datasetValues = useDatasetValues({
+    [signalName]: signalDataset.id,
+    ...Object.fromEntries(
+      Object.entries(axisDatasets).map(([axis, dataset]) => [axis, dataset.id])
+    ),
+  });
+  const nxAxisParams = getNxAxisParams(axisDatasets, datasetValues);
```
axelboc, line 64: I would just rename this `nxAxesParams` and `getNxAxesParams` for consistency with `axesParams`.

My reasoning is that `axisParams`, although technically correct, could be misinterpreted. Note that I'm happy with `axisMapping` and not `axesMapping`, though, because it's consistent with `dimensionMapping`.

axelboc (approved):
I feel good about this 💪  -- now that the containers code is clearer, I feel like it will be easier to find further refactoring opportunities.

src/h5web/visualizations/nexus/utils.ts
loichuder: Settled for `getLinkedDatasets`. I also changed the signature of `getLinkedEntity` so that the `name` is the first argument as well.

PR #275, opened 2020-10-29
Update some dependencies and replace remaining occrurrences of `JSX.Element` with` ReactElement`

axelboc:


loichuder (approved):
So we will encounter https://github.com/DefinitelyTyped/DefinitelyTyped/issues/48299 if we update `@types/d3-scale` to `>=2.2.4` ?

axelboc:
Yes, I've subscribed to the Pull Request to know when we can upgrade 😉

PR #276, opened 2020-10-30
Replace unmaintained shared linting config

axelboc:
CRA v4 requires ESLint v7, but the shared linting config that we use, [@wsmd/eslint-config](https://github.com/wsmd/eslint-config) doesn't support it and is no longer maintained.

I considered switching to Airbnb's config, but it has too many downsides:

- it includes formatting rules, which requires installing a package to disable them;
- it requires installing (and maintaining) a lot of plugins manually
- it doesn't cover our testing toolchain (jest, Testing Library)
- it's missing some other interesting plugins (e.g. sonarjs, unicorn, etc.)

Instead, I found [eslint-config-galex](https://github.com/ljosberinn/eslint-config-galex), which fits out toolchain a lot better and is actively maintained. We'll see if it lasts in the long run, but it's better than to configure ESLint and its tens of plugins ourselves...

I've disabled a number of rules for various reasons, and enabled a few others that were enabled in our previous config. I've done my best to justify my decisions in comments in `.eslintrc.js`.

The new config is stricter. I hope you'll like the changes in the code.


.eslintrc.js
```diff
@@ -0,0 +1,74 @@
...
+const {
+  files: reactFiles,
+} = require('eslint-config-galex/src/overrides/react');
+const {
+  files: tsFiles,
+} = require('eslint-config-galex/src/overrides/typescript');
+
+module.exports = createConfig({
+  rules: {
+    'import/order': 'off',
```
axelboc, line 12: I'm planning to enable this in a separate PR as it will create a huge diff.

packages/webpack.config.js
```diff
@@ -33,7 +33,7 @@ function getConfig(pkg) {
     module: {
       rules: [
         {
-          test: /\.tsx?$/,
+          test: /\.tsx?$/u,
```
axelboc, line 36: `u` is for unicode. Using it on every regex is recommended because it enforces a stricter regex syntax.

src/demo-app/DemoApp.tsx
```diff
@@ -5,7 +5,12 @@ import SilxProvider from '../h5web/providers/silx/SilxProvider';
 import HsdsProvider from '../h5web/providers/hsds/HsdsProvider';
 
 // Split mock data generation code out of main bundle
-const MockProvider = lazy(() => import('../h5web/providers/mock/MockProvider'));
+const MockProvider = lazy(
+  () =>
+    import(
+      /* webpackChunkName: "mock" */ '../h5web/providers/mock/MockProvider'
```
axelboc, line 11: Providing a chunk name is recommended for long-term caching.

src/h5web/App.test.tsx
```diff
@@ -21,48 +21,52 @@ function renderApp(): RenderResult {
 }
 
 test('renders with root group selected', async () => {
-  const { findByText, getByRole } = renderApp();
+  renderApp();
 
-  const noVisText = await findByText('Nothing to visualize');
+  const noVisText = await screen.findByText('Nothing to visualize');
```
axelboc, line 26: `screen` makes tests easier to write thanks to auto-completion. Adding each method to the destructuring assignment takes a lot longer.

src/h5web/App.test.tsx
```diff
@@ -190,32 +196,27 @@ test('switch visualisations', async () => {
...
   // Select the line vis
-  fireEvent.click(await findByRole('tab', { name: 'Line' }));
+  fireEvent.click(await screen.findByRole('tab', { name: 'Line' }));
 
   // Ensure that the dimension mapper is only visible for X (and not for Y)
-  const xRadioGroup = getByLabelText('Dimension as x axis');
+  const xRadioGroup = screen.getByLabelText('Dimension as x axis');
   expect(xRadioGroup).toBeVisible();
-  const xDimsButtons = getAllByRoleInElement(xRadioGroup, 'radio');
+  const xDimsButtons = within(xRadioGroup).getAllByRole('radio');
```
axelboc, line 211: Newer syntax.

src/h5web/dimension-mapper/AxisMapper.tsx
```diff
@@ -8,7 +8,7 @@ interface Props {
   axis: Axis;
   rawDims: number[];
   mapperState: NonNullable<DimensionMapping>;
-  onChange(mapperState: DimensionMapping): void;
+  onChange: (mapperState: DimensionMapping) => void;
```
axelboc, line 11: With `onChange` declared as a method, ESLint wants us to call it with `props.onChange(...)` instead of just `onChange(...)` to ensure that `this` is properly bound. Of course we don't care about `this` here, so a simple function is more adequate.

src/h5web/metadata-viewer/EntityInfo.tsx
```diff
@@ -14,8 +14,7 @@ function EntityInfo(props: Props): ReactElement {
   const { collection } = entity;
 
   const entityKind =
-    collection.charAt(0).toUpperCase() +
-    collection.substr(1, collection.length - 2);
+    collection.charAt(0).toUpperCase() + collection.slice(1, -1);
```
axelboc, line 17: Prefer `slice` over `substr` for its more powerful syntax.

src/h5web/metadata-viewer/MetadataViewer.tsx
```diff
@@ -6,7 +6,6 @@ import EntityInfo from './EntityInfo';
 import styles from './MetadataViewer.module.css';
 
 interface Props {
-  key: string;
```
axelboc, line 9: Unused prop. Also, we no longer need a `key` on `MetadataViewer` since `selectedEntity` is no longer retrieved asynchronously; it is always in sync with `selectedLink`.

src/h5web/App.tsx
```diff
@@ -45,11 +45,7 @@ function App(): ReactElement {
             selectedNode={selectedNode}
           />
           {isInspecting ? (
-            <MetadataViewer
-              key={JSON.stringify(selectedLink)}
-              link={selectedLink}
-              entity={selectedEntity}
-            />
+            <MetadataViewer link={selectedLink} entity={selectedEntity} />
```
axelboc, line 48: Not an ESLint error. I just noticed that we no longer need a key on MetadataViewer since selectedEntity is no longer retrieved asynchronously; it is always in sync with selectedLink.

src/h5web/providers/Provider.tsx
```diff
@@ -25,12 +25,15 @@ function Provider(props: Props): ReactElement {
...
+        getValues: async (idsRecord) =>
+          Object.fromEntries(
+            await Promise.all(
+              Object.entries(idsRecord).map(async ([name, id]) => {
+                const val = await getValue(id);
+                return [name, val];
+              })
             )
-          ).then(Object.fromEntries),
+          ),
```
axelboc, line 36: `async`/`await` instead of `then`.

src/h5web/toolbar/LineToolbar.tsx
```diff
@@ -27,7 +27,9 @@ function LineToolbar(): ReactElement {
         role="radiogroup"
         ariaLabel="Curve type"
         value={curveType}
-        onChange={setCurveType}
+        onChange={(val) => {
+          setCurveType(val as CurveType);
+        }}
```
axelboc, line 32: Type error (`val` is `string` and it's too difficult to type the `ToggleGroup` context with a generic). Not sure why it hadn't been detected before.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -13,7 +13,7 @@ import type { ColorMap } from './models';
 import { getDomain, getValueToIndexScale } from '../shared/utils';
 
 interface Props {
-  dataArray: ndarray<number>;
+  dataArray: ndarray;
```
axelboc, line 16: `number` is the default.

src/h5web/visualizations/shared/Html.tsx
```diff
@@ -3,7 +3,7 @@ import ReactDOM from 'react-dom';
 import { useThree } from 'react-three-fiber';
 
 // Simplified version of `drei`'s `<Html>` component
-// https://github.com/pmndrs/drei/blob/master/src/misc/Html.tsx
+// https://github.com/pmndrs/drei/blob/v2.2.3/src/Html.tsx
```
axelboc, line 6: non-inclusive language... but also broken link

loichuder (approved):
Thanks for the hard work :1st_place_medal:

PR #277, opened 2020-11-02
Split features tests into multiple files

axelboc:
Even though we're still always rendering the whole app, I think splitting the tests based on the component being tested makes sense. `App.test.tsx` file was becoming too unwieldy.


.eslintrc.js
```diff
@@ -52,8 +52,6 @@ module.exports = createConfig({
     {
       files: jestFiles,
       rules: {
-        'jest/require-top-level-describe': 'off', // until App.test.tsx is split into multiple files
-
```
axelboc, line 56: I've added a `describe` block in each test file.

src/h5web/explorer/Explorer.test.tsx
```diff
@@ -0,0 +1,86 @@
+import { fireEvent, screen } from '@testing-library/react';
+import { DOMAIN, renderApp } from '../test-utils';
+
+describe('Explorer', () => {
+  test('select root group by default', async () => {
+    renderApp();
```
loichuder, line 6: Is the app created without any mock data here ?

axelboc: Nothing's changed, it renders the `App` with the `MockProvider`, so it renders exactly the same data as the `/mock` route.

loichuder: Read until the end: it has mock data.

Then, perhaps the check on `Nothing to visualize` is a bit tricky as the root node might have some visualisations with NeXuS.

axelboc: Hmm good point... Seems to be passing when it shouldn't...

loichuder (approved):

PR #284, opened 2020-11-04
Upgraded storybook

loichuder:
To fix `Invalid hook call` error in the Storybook.


loichuder:
@axelboc Could you try on your end to see if it fixes the error ?

axelboc (approved):

PR #285, opened 2020-11-04
Test vis selector

axelboc:
Some higher-level feature tests for `supportsEntity` and the vis selector. I only test ideal scenarios with existing mock data. I think the tests in `visualizations/index.test.ts` are worth keeping, though.

### DIFF
--- a/src/h5web/visualizations/Visualizations.test.tsx
+++ b/src/h5web/visualizations/Visualizations.test.tsx
@@ -1,0 +1,24 @@
+import { fireEvent, screen } from '@testing-library/react';
+import { renderApp } from '../test-utils';
+
+describe('Visualizer', () => {
+  test('visualise scalar dataset', async () => {
+    renderApp();
+
+    fireEvent.click(await screen.findByRole('treeitem', { name: 'entities' }));
+    const datasetBtn = screen.getByRole('treeitem', { name: 'scalar_int' });
+    expect(datasetBtn).toBeVisible();
+
+    fireEvent.click(datasetBtn);
+    expect(datasetBtn).toHaveAttribute('aria-selected', 'true');
+    expect(datasetBtn).not.toHaveAttribute('aria-expanded'); // a dataset cannot be expanded
+
+    const scalarTab = await screen.findByRole('tab', { name: 'Scalar' });
+    expect(scalarTab).toBeVisible();
+    expect(scalarTab).toHaveAttribute('aria-selected', 'true');
+    expect(screen.getByText('0')).toBeVisible();
+
+    fireEvent.click(screen.getByRole('treeitem', { name: 'scalar_str' }));
+    expect(await screen.findByText('foo')).toBeVisible();
+  });
+});
--- a/src/h5web/visualizer/Visualizer.test.tsx
+++ b/src/h5web/visualizer/Visualizer.test.tsx
@@ -1,32 +1,140 @@
 import React from 'react';
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import {
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+  within,
+} from '@testing-library/react';
 import { mockValues } from '../../packages/lib';
 import App from '../App';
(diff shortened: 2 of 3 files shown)


src/h5web/visualizations/Visualizations.test.tsx
```diff
@@ -0,0 +1,24 @@
+import { fireEvent, screen } from '@testing-library/react';
+import { renderApp } from '../test-utils';
```
axelboc, line 2: Just moving this test out of `Visualizer.test.tsx`.

src/h5web/visualizer/Visualizer.test.tsx
```diff
@@ -1,32 +1,138 @@
...
+
+async function findVisSelectorTabs() {
+  return within(
+    await screen.findByRole('tablist', { name: 'Visualization' })
+  ).getAllByRole('tab');
+}
 
 describe('Visualizer', () => {
-  test('visualise scalar dataset', async () => {
+  test('show "Raw" tab for dataset with complex type', async () => {
```
loichuder, line 22:
```suggestion
  test('show only "Raw" tab for dataset with complex type', async () => {
```

src/h5web/visualizer/Visualizer.test.tsx
```diff
@@ -1,32 +1,138 @@
...
-    const datasetBtn = screen.getByRole('treeitem', { name: 'scalar_int' });
-    expect(datasetBtn).toBeVisible();
+    fireEvent.click(screen.getByRole('treeitem', { name: 'raw' }));
 
-    fireEvent.click(datasetBtn);
-    expect(datasetBtn).toHaveAttribute('aria-selected', 'true');
-    expect(datasetBtn).not.toHaveAttribute('aria-expanded'); // a dataset cannot be expanded
+    const tabs = await findVisSelectorTabs();
+    expect(tabs).toHaveLength(1);
+    expect(tabs[0]).toHaveTextContent(Vis.Raw);
```
loichuder, line 30: You could also test the selection ? :shrug:

src/h5web/visualizer/Visualizer.test.tsx
```diff
@@ -1,32 +1,138 @@
...
+  });
+
+  test('show "NX Image" tab for NXdata group with 2D signal', async () => {
+    renderApp();
+
+    fireEvent.click(
+      await screen.findByRole('treeitem', { name: 'nexus_entry' })
+    );
+    fireEvent.click(screen.getByRole('treeitem', { name: 'nx_process' }));
+    fireEvent.click(screen.getByRole('treeitem', { name: 'nx_data' }));
```
loichuder, line 108: Didn't we talk about adding a helper to select nested entities more easily ?

loichuder (approved):
:octopus:

src/h5web/visualizer/Visualizer.test.tsx
axelboc: Yeah, I was planning on doing this soon.

src/h5web/visualizer/Visualizer.test.tsx
axelboc: Hmm if I specify it there, I'd have to specify it in the other tests as well. To me, it's not the main part of the test, a bit like the `aria-selected` check, so it doesn't have to be in the test's name.

loichuder: I would agree but the `Raw` vis is a bit 

> Hmm if I specify it there, I'd have to specify it in the other tests as well. 
Not really. The `Raw` vis is special in the way that it is shown only when there are not other vis so it is a bit of a special case.

Anyhow, it is not that important :smile_cat:

axelboc: Yeah, so you could argue that the other tests should also state that _only_ the tabs listed should be shown (i.e. that `Raw` is not shown).

loichuder: True. Let's keep it as it is, then !

PR #286, opened 2020-11-04
Line and Heatmap now show the entity name as title

### LINKED ISSUE #278: Plot titles for datasets visualizations
Non-NeXuS visualizations should show the dataset name as plot title:

image

loichuder:
Fix #278 

<s>I had to make a few changes to have access to the entity name. Indeed, Vis containers only recieve `entity` in props which did not contain the name of the entity.

I therefore changed the `HDF5Entity` names and the API creating them so that the name is now included in `entity`.</s> 

See https://github.com/silx-kit/h5web/pull/286#issuecomment-721728077

### DIFF
--- a/src/h5web/providers/mock/data.ts
+++ b/src/h5web/providers/mock/data.ts
@@ -252,6 +252,7 @@ export const mockMetadata = makeMetadata({
       makeDatasetHardLink('scalar_str'),
     ]),
     makeGroup('nD_datasets', undefined, [
+      makeDatasetHardLink('oneD_linear'),
       makeDatasetHardLink('oneD'),
       makeDatasetHardLink('twoD'),
       makeDatasetHardLink('threeD'),
@@ -308,6 +309,7 @@ export const mockMetadata = makeMetadata({
     ),
   ],
   datasets: [
+    makeSimpleDataset('oneD_linear', intType, [41]),
     makeSimpleDataset(
       'oneD',
       intType,
@@ -359,6 +361,7 @@ export const mockValues = {
   raw: { int: 42 },
   scalar_int: 0,
   scalar_str: 'foo',
+  oneD_linear: arr1,
   oneD,
   twoD,
   threeD,
--- a/src/h5web/App.tsx
+++ b/src/h5web/App.tsx
@@ -47,7 +47,10 @@ function App(): ReactElement {
           {isInspecting ? (
             <MetadataViewer link={selectedLink} entity={selectedEntity} />
           ) : (
-            <Visualizer entity={selectedEntity} />
+            <Visualizer
+              entity={selectedEntity}
+              entityName={selectedLink && selectedLink.title}
+            />
           )}
         </ReflexElement>
       </ReflexContainer>
--- a/src/h5web/visualizations/containers/HeatmapVisContainer.tsx
+++ b/src/h5web/visualizations/containers/HeatmapVisContainer.tsx
@@ -9,7 +9,7 @@ import MappedHeatmapVis from '../heatmap/MappedHeatmapVis';
 import { VisContainerProps } from './models';
 
 function HeatmapVisContainer(props: VisContainerProps): ReactElement {
-  const { entity } = props;
+  const { entity, entityName } = props;
   assertDataset(entity);
 
   const value = useDatasetValue(entity.id);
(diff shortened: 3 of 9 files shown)


src/h5web/providers/hsds/api.ts
```diff
@@ -163,31 +164,33 @@ export class HsdsApi implements ProviderAPI {
...
-      return this.processGroup(id);
+      return this.processGroup(id, title);
     }
     if (collection === HDF5Collection.Datasets) {
-      return this.processDataset(id);
+      return this.processDataset(id, title);
     }
     if (collection === HDF5Collection.Datatypes) {
-      return this.processDatatype(id);
+      return this.processDatatype(id, title);
```
loichuder, line 193: In HSDS, the `name` of the entity is gotten from the title of the link pointing to the entity.

src/h5web/providers/mock/data.ts
```diff
@@ -99,13 +107,18 @@ export function makeSimpleDataset(
...
-  return { id, collection: HDF5Collection.Groups, attributes, links };
+  return { id, collection: HDF5Collection.Groups, attributes, links, name };
 }
 
-export function makeDatatype(id: HDF5Id, type: HDF5Type): HDF5Datatype {
-  return { id, collection: HDF5Collection.Datatypes, type };
+export function makeDatatype(
+  id: HDF5Id,
+  type: HDF5Type,
+  name: string = id
```
loichuder, line 119: In mock data, the `name` can be given or falls back to the `id`

src/h5web/providers/mock/data.ts
```diff
@@ -252,6 +265,7 @@ export const mockMetadata = makeMetadata({
       makeDatasetHardLink('scalar_str'),
     ]),
     makeGroup('nD_datasets', undefined, [
+      makeDatasetHardLink('oneD_simple'),
```
loichuder, line 268: I added a new simple 1D dataset `y=x` for the LineVis story. The dataset `oneD` is very nice for other stories but not for the log scale stories of the LineVis.

src/h5web/providers/silx/api.ts
```diff
@@ -8,7 +8,17 @@ import {
...
+
+function getName(entity: SilxEntity): string {
+  const { alias } = entity;
+  if (alias.length === 0) {
+    return '';
+  }
+  const pathMembers = alias[0].split('/');
+
+  return pathMembers.length > 0 ? pathMembers[pathMembers.length - 1] : '';
+}
```
loichuder, line 21: In SilxAPI, I don't have an easy access to the link titles. 

I make use of the `alias` that is an array of the paths leading to the entity. The last part of the path is therefore the entity name.

src/h5web/providers/silx/models.ts
```diff
@@ -1,17 +1,35 @@
...
+  type: HDF5Type;
+  alias: string[];
+}
+
+interface SilxDatatype {
+  type: HDF5Type;
+  alias: string[];
+}
+
+export type SilxEntity = SilxGroup | SilxDataset | SilxDatatype;
```
loichuder, line 28: Added clearer dedicated types for `Silx`

axelboc:
I don't get why you need all of this. In `App`, `selectedNode` already includes the node's `label` which is the `title` of the corresponding `HDF5Link`. Can't we just pass a prop called `entityName` to the containers?

src/h5web/providers/mock/data.ts
axelboc: Is this the plot you don't like?

image

Right, so you want `y=x + a` basically. Makes sense. Not sure about the term `simple`, though, since we use it to refer to HDF5 simple shapes. Perhaps `oneD_linear`?

loichuder: This one and the `Negative Log Scale` that doesn't show anything given that there are only positive values. :+1: for `oneD_linear`

loichuder:
> I don't get why you need all of this. In `App`, `selectedNode` already includes the node's `label` which is the `title` of the corresponding `HDF5Link`. Can't we just pass a prop called `linkName` or `entityName` or whatever through the `Visualizer` and down to the containers?

Sure we can do it like this as well. I just thought it might be cleaner to have an entity with a name field from the get-go, in case we need it lower down in the hierarchy. 

I was also concerned that if we do that, we might end up putting a lot of arg to the `Visualizer` each time we would need something in the Vis. And I liked the fact that it only needed an `entity` prop.

axelboc (approved):
Nickel ✨

PR #287, opened 2020-11-04
Streamline feature tests with helper functions

axelboc:


src/h5web/dimension-mapper/DimensionMapper.test.tsx
```diff
@@ -31,26 +25,16 @@ describe('DimensionMapper', () => {
...
-    await waitFor(() => {
-      expect(xDimsButtons[0]).toBeChecked();
-      expect(xDimsButtons[1]).not.toBeChecked();
-      const D1Slider = screen.getByRole('slider');
-      expect(D1Slider).toHaveAttribute('aria-valueNow', '0');
-    });
+    expect(xDimsButtons[0]).toBeChecked();
+    expect(xDimsButtons[1]).not.toBeChecked();
+    const D1Slider = screen.getByRole('slider');
+    expect(D1Slider).toHaveAttribute('aria-valueNow', '0');
```
axelboc, line 31: We no longer seem to need these `waitFor`. Probably thanks to `Suspense`.

src/h5web/visualizations/Visualizations.test.tsx
```diff
@@ -1,24 +1,21 @@
-import { fireEvent, screen } from '@testing-library/react';
-import { renderApp } from '../test-utils';
+import { screen } from '@testing-library/react';
+import { renderApp, selectExplorerNode } from '../test-utils';
 
 describe('Visualizer', () => {
-  test('visualise scalar dataset', async () => {
+  test('visualise raw dataset', async () => {
```
axelboc, line 5: New test to test the raw vis.

src/h5web/visualizations/Visualizations.test.tsx
```diff
@@ -1,24 +1,21 @@
...
     renderApp();
+    await selectExplorerNode('entities/raw');
 
-    fireEvent.click(await screen.findByRole('treeitem', { name: 'entities' }));
-    const datasetBtn = screen.getByRole('treeitem', { name: 'scalar_int' });
-    expect(datasetBtn).toBeVisible();
+    expect(await screen.findByText(/"int": 42/u)).toBeVisible();
```
loichuder, line 9: Would it be possible to have a more robust way to get the value of the dataset ?

loichuder (approved):
Looks good :eyes:

src/h5web/visualizations/Visualizations.test.tsx
axelboc: I could do `JSON.stringify(mockValues.raw)` but it would be re-implementing the raw vis 😅 -- you can see it as a snapshot test. If it fails even though we didn't change the mock data, then there's a problem.

PR #288, opened 2020-11-04
Improve accessibility of Matrix vis table

axelboc:
The Matrix vis table was just made of `div`s with no semantic whatsoever. I noticed it when trying to feature test the Matrix vis (which I ended up giving up on, because of `useMeasure` not working in JSDOM).


loichuder (approved):
Nice :sparkles:

PR #291, opened 2020-11-04
Now using NXdata dataset title for plot titles when available

loichuder:
image

### DIFF
--- a/src/h5web/providers/mock/data.ts
+++ b/src/h5web/providers/mock/data.ts
@@ -281,7 +281,7 @@ export const mockMetadata = makeMetadata({
     makeGroup(
       'nx_data',
       [makeStrAttr('NX_class', 'NXdata'), makeStrAttr('signal', 'twoD')],
-      [makeDatasetHardLink('twoD')]
+      [makeDatasetHardLink('twoD'), makeDatasetHardLink('title', 'title_twoD')]
     ),
     makeGroup(
       'spectrum',
@@ -334,6 +334,7 @@ export const mockMetadata = makeMetadata({
       [20],
       [makeStrAttr('units', 'deg'), makeStrAttr('long_name', 'Angle (degrees)')]
     ),
+    makeDataset('title_twoD', stringType, scalarShape),
   ],
   datatypes: [makeDatatype('datatype', compoundType)],
 });
@@ -368,4 +369,5 @@ export const mockValues = {
   fourD,
   X: arr1,
   Y: arr2,
+  title_twoD: 'NeXus 2D',
 };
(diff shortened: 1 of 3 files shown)


axelboc:
Sorry, I think you missed the outcome of the discussion we had with Thomas on the chat. The NeXus specification does not define a `title` attribute on the `NXdata`, but a `title` **dataset** _inside_ the `NXdata` group (a scalar dataset of type string to be exact).

src/h5web/providers/mock/data.ts
```diff
@@ -368,4 +369,5 @@ export const mockValues = {
   fourD,
   X: arr1,
   Y: arr2,
+  title: 'NeXus 2D',
```
axelboc, line 372: Maybe use a more distinctive ID, like `title_twoD` and then call `makeDatasetHardLink` with a second argument?

src/h5web/visualizations/containers/NxImageContainer.tsx
```diff
@@ -55,12 +55,18 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
...
     [signalName]: signalDataset.id,
     ...Object.fromEntries(
       Object.entries(axisDatasets).map(([axis, dataset]) => [axis, dataset.id])
     ),
-  });
+  };
+  const titleDataset = getLinkedEntity('title', nxDataGroup, metadata);
+  if (titleDataset && isDataset(titleDataset)) {
+    idsToFetch.title = titleDataset.id;
+  }
```
axelboc, line 68: You can do something like:
```ts
const datasetValues = useDatasetValues({
  [signalName]: signalDataset.id,
  ...(titleDataset && isDataset(titleDataset) ? { title: titleDataset.id } : {}),
  ...Object.fromEntries(
    Object.entries(axisDatasets).map(([axis, dataset]) => [axis, dataset.id])
  ),
});
```

src/h5web/visualizations/containers/NxImageContainer.tsx
```diff
@@ -76,7 +82,11 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
       />
       <MappedHeatmapVis
         value={datasetValues[signalName]}
-        title={getDatasetLabel(signalDataset, signalName)}
+        title={
+          datasetValues.title && typeof datasetValues.title === 'string'
```
axelboc, line 86: You don't need the first part of the condition, since `typeof undefined === "undefined"`

axelboc (approved):
I think the `useDatasetValues` version is best 👍

axelboc (approved):

PR #297, opened 2020-11-06
Now using custom hook to parse nxData entities

loichuder:
Another tentative refactoring. Following our discussion, I:
- added a dedicated `NxData` type
- implemented the fetching of **all** children datasets at the beginning at it is much more easier then to build the NxData object.

Let me know what you think.

### DIFF
--- a/cypress/integration/app.spec.ts
+++ b/cypress/integration/app.spec.ts
@@ -98,6 +98,7 @@ describe('App', () => {
 
       cy.findByRole('heading', { name: 'nexus_entry / image' }).should('exist');
       cy.findByRole('tab', { name: 'NX Image' }).should('exist');
+      cy.wait(500);
     });
 
     it('use signal name and units to compute title', () => {
--- a/src/h5web/visualizations/containers/NxImageContainer.tsx
+++ b/src/h5web/visualizations/containers/NxImageContainer.tsx
@@ -1,23 +1,13 @@
 import React, { ReactElement, useState, useContext } from 'react';
 import { range } from 'lodash-es';
-import { HDF5SimpleShape } from '../../providers/models';
-import { assertGroup, isDataset } from '../../providers/utils';
+import { assertGroup } from '../../providers/utils';
 import DimensionMapper from '../../dimension-mapper/DimensionMapper';
 import { DimensionMapping } from '../../dimension-mapper/models';
 import { ProviderContext } from '../../providers/context';
-import {
-  getAttributeValue,
-  getLinkedEntity,
-  getNxDataGroup,
-  getNxAxisMapping,
-  getLinkedDatasets,
-  getNxAxesParams,
-  getDatasetLabel,
-} from '../nexus/utils';
+import { getNxDataGroup } from '../nexus/utils';
 import { VisContainerProps } from './models';
 import MappedHeatmapVis from '../heatmap/MappedHeatmapVis';
-import { assertStr } from '../shared/utils';
-import { useDatasetValues } from './hooks';
+import { useNxData } from '../nexus/hooks';
 
 function NxImageContainer(props: VisContainerProps): ReactElement {
   const { entity } = props;
@@ -30,15 +20,9 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
     throw new Error('NXdata group not found');
   }
 
-  const signalName = getAttributeValue(nxDataGroup, 'signal');
-  assertStr(signalName);
-  const signalDataset = getLinkedEntity(signalName, nxDataGroup, metadata);
+  const nxData = useNxData(nxDataGroup, metadata);
 
-  if (!signalDataset || !isDataset(signalDataset)) {
-    throw new Error('Signal dataset not found');
-  }
-
-  const { dims } = signalDataset.shape as HDF5SimpleShape;
+  const { dims } = nxData.signal;
   if (dims.length < 2) {
     throw new Error('Expected signal dataset with at least two dimensions');
   }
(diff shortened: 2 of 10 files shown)


src/h5web/visualizations/containers/NxImageContainer.tsx
```diff
@@ -30,15 +19,13 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
     throw new Error('NXdata group not found');
   }
 
-  const signalName = getAttributeValue(nxDataGroup, 'signal');
-  assertStr(signalName);
-  const signalDataset = getLinkedEntity(signalName, nxDataGroup, metadata);
+  const nxData = useParsedNxData(nxDataGroup, metadata);
```
axelboc, line 22: I think just `useNxData` would be fine.

src/h5web/visualizations/containers/hooks.ts
```diff
@@ -17,3 +24,27 @@ export function useDatasetValues(
...
+
+  const datasetEntries = datasetLinks
+    .map((link) => {
+      const entity = getEntity(link, metadata);
+
+      if (!entity || !isDataset(entity)) {
+        return undefined;
+      }
+
+      return [link.title, entity.id];
```
axelboc, line 45: You can maybe use a ternary here?

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -124,15 +125,49 @@ export function getLinkedDatasets(
...
 
-export function getNxAxesParams(
-  nxDatasets: Record<string, HDF5Dataset>,
-  datasetValues: Record<string, HDF5Value> | undefined
-): Record<string, AxisParams> {
-  const paramsEntries = Object.entries(nxDatasets).map(([axis, dataset]) => {
-    const values = datasetValues && datasetValues[axis];
-    assertOptionalArray<number>(values);
-    return [axis, { values, label: getDatasetLabel(dataset, axis) }];
+export function useParsedNxData(
```
axelboc, line 128: This should probably go in a file called `hooks.ts`.

src/h5web/visualizations/containers/hooks.ts
```diff
@@ -17,3 +24,27 @@ export function useDatasetValues(
 
   return valuesReader();
 }
+
+export function useChildDatasetValues(
```
axelboc, line 28: You could argue that this utility function is specific to NeXus and should live in `nexus/utils`.

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
```diff
@@ -30,16 +19,13 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
...
 
-  const signalName = getAttributeValue(nxDataGroup, 'signal');
-  assertStr(signalName);
+  const nxData = useParsedNxData(nxDataGroup, metadata);
 
-  const signalDataset = getLinkedEntity(signalName, nxDataGroup, metadata);
-
-  if (!signalDataset || !isDataset(signalDataset)) {
+  if (!nxData) {
     throw new Error('Signal dataset not found');
```
axelboc, line 25: Hmm if `nxData` is `undefined`, it means the values are being fetched, doesn't it? So you should return `<></>` rather than trow, maybe?

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -124,15 +125,49 @@ export function getLinkedDatasets(
...
+    return undefined;
+  }
+
+  const signalName = getAttributeValue(group, 'signal');
+  assertStr(signalName);
+  const signalDataset = getLinkedEntity(signalName, group, metadata);
+  const signalValue = datasetValues[signalName];
+
+  if (!signalDataset || !isDataset(signalDataset) || !signalValue) {
+    return undefined;
```
axelboc, line 143: I think this is where you should throw `'Signal dataset not found'`.

axelboc: You can also remove `!isDataset()` from the check and use `assertDataset()` instead.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -124,15 +125,49 @@ export function getLinkedDatasets(
...
+  }
+  assertArray<number>(signalValue);
+
+  const signal = {
+    label: getDatasetLabel(signalDataset, signalName),
+    value: signalValue,
+    dims: (signalDataset.shape as HDF5SimpleShape).dims,
+  };
```
axelboc, line 151: I would just move this object as a nested object in the `return` statement and wouldn't bother with a `signal` constant.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -124,15 +125,49 @@ export function getLinkedDatasets(
...
+    axisMapping.filter((v): v is string => !!v),
+    group,
+    metadata
+  );
+
+  const axesEntries = Object.entries(axisDatasets).map(([name, dataset]) => {
+    const value = datasetValues[name];
+    assertOptionalArray<number>(value);
+    return [name, { value, label: getDatasetLabel(dataset, name) }];
   });
```
axelboc, line 164: When you end up refactoring `axisMapping`+`axesParams` into a single `axisParamsMapping`(?) array, this bit will become much simpler: you'll just have to map through `axisMapping`, and for each value, call `getLinkedEntity`, assert the entity and the value, and return an `AxisParams` object.

axelboc (changes requested):
This is awesome, great job! 💪 Just a couple of assertions to move around and the CI to fix (I'll take a look).

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
loichuder: Good point but apparently, the rendering waits for the fetching ? :shrug: 

Also, I cannot return here as there is the `useState` hook below whose initial value depends on `nxData.signal.dims`...

axelboc: I think we don't see any error because `Suspense` kind of hides it. The function finishes executing, so throwing does not seem like a good idea. Hmm this is quite a problem, in my opinion.

axelboc: Perhaps `useNxData` should return what it can the first time around just so the `useState` can be properly initialized (and the `return` moved below the `useState`.

cypress/integration/app.spec.ts
```diff
@@ -98,6 +98,7 @@ describe('App', () => {
 
       cy.findByRole('heading', { name: 'nexus_entry / image' }).should('exist');
       cy.findByRole('tab', { name: 'NX Image' }).should('exist');
+      cy.wait(500);
```
loichuder, line 101: :roll_eyes: 

Sometimes this test makes the following test pending. I suspect it is because the texture computation in the worker didn't finish. Adding this wait makes it consistently passing

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (!signalDataset) {
+    throw new Error('Signal dataset not found');
+  }
+  assertDataset(signalDataset);
+
+  const signalValue = datasetValues && datasetValues[signalName];
+  if (!signalValue) {
+    return {
+      signal: {
+        label: getDatasetLabel(signalDataset, signalName),
```
axelboc, line 60: I think you can do without returning `label` and `value`.

axelboc (approved):

Issue #235, opened 2020-10-06
[NeXus] Error bars


axelboc:

loichuder:
- [ ] Create a Vis that shows error bars (extend `LineVis` ?)
- [ ] Add support for `errors` dataset in NxData groups

PR #298, opened 2020-11-09
Added support for error bars in LineVis

loichuder:
image

First step of #235.

A few assumptions/limitations:
- Errors for a dataset of `N` points are supplied as an array `[err0, err1, ..., errN]`. The errors are therefore considered symmetric (i.e the error bar is between `y0 - err0` and `y0 + err0`).
- Errors for abscissas are not implemented for now
- Error bars are not accounted for where computing the domain. This is because `LineVis` is "dumb" with respect of `domain`.
- Error bars are simple lines for now. We can imagine putting caps in the future.


src/h5web/visualizations/line/ErrorBarCurve.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    // [x, y + yerr+]
+    errorAbscissas.push(abscissas[index]);
+    errorOrdinates.push(ordinates[index] + errors[index]);
+  }
+
+  const errorGeometry = useDataGeometry(errorAbscissas, errorOrdinates);
+
+  return (
+    <Suspense fallback={<></>}>
```
axelboc, line 32: Might as well put `null` as fallback.

src/h5web/visualizations/line/ErrorBarCurve.tsx
```diff
@@ -0,0 +1,40 @@
...
+  errors: number[];
+  color?: string;
+}
+
+function ErrorBarCurve(props: Props): ReactElement {
+  const { abscissas, ordinates, errors, color = DEFAULT_COLOR } = props;
+
+  const errorAbscissas = [];
+  const errorOrdinates = [];
+  for (let index = 0; index < abscissas.length; index++) {
```
axelboc, line 19: With `for of` it's more modern ^^ https://stackoverflow.com/a/34349073

axelboc: Isn't `errorAbscissas` just `abscissas` 🤔 If you remove it from the loop, you can use a `map` instead of the ordinates.

axelboc (approved):
Impressive 👏

src/h5web/visualizations/line/ErrorBarCurve.tsx
loichuder: Actually, I think I can remove the `Suspense` completely as there is no loading of external resources (contrary to the sprites in `DataCurve`)

PR #299, opened 2020-11-09
Adding caps to error bars

loichuder:
image

Part of #235 

Could have been part of #298, sorry ! :pray:

### DIFF
--- a/src/h5web/visualizations/line/ErrorBarCurve.tsx
+++ b/src/h5web/visualizations/line/ErrorBarCurve.tsx
@@ -14,16 +14,34 @@ interface Props {
 function ErrorBarCurve(props: Props): ReactElement {
   const { abscissas, ordinates, errors, color = DEFAULT_COLOR } = props;
 
+  const capLength = 0.1 * (abscissas[1] - abscissas[0]);
+
   const errorAbscissas = [];
   const errorOrdinates = [];
   for (const [index, x] of abscissas.entries()) {
     const y = ordinates[index];
     const yerr = errors[index];
 
-    // Error bar segment: [x, y - yerr] to [x, y + yerr]
+    if (yerr === 0) {
+      continue;
+    }
+
+    // First segment: bottom cap
+    errorAbscissas.push(x - capLength);
+    errorOrdinates.push(y - yerr);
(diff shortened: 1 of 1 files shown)


src/h5web/visualizations/line/ErrorBarCurve.tsx
```diff
@@ -14,17 +14,35 @@ interface Props {
 function ErrorBarCurve(props: Props): ReactElement {
   const { abscissas, ordinates, errors, color = DEFAULT_COLOR } = props;
 
+  const capLength = 0.1 * (abscissas[1] - abscissas[0]);
```
loichuder, line 17: Taking 10% of the spacing between two points seems to give a good aspect ratio.

axelboc (approved):
Great, just wondering though if we should still be working with two arrays. We could rethink the `useDataGeometry` hook so that it accepts a single array: `[x1, y1, x2, y2, ...]`. It might make things more straightforward.

- In the hook you could loop through the array with `map` and use `index % 2` to decide which scale to apply.
- In the curve components, you could `reduce` abscissas and build the array of points:

```ts
points.push(x - capLength, y - yerr, x + capLength, y - yerr); // bottom cap
points.push(x, y - yerr, x, y + yerr); // error bar
...
```

loichuder:
> Great, just wondering though if we should still be working with two arrays. We could rethink the `useDataGeometry` hook so that it accepts a single array: `[x1, y1, x2, y2, ...]`. It might make things more straightforward.
> 
>     * In the hook you could loop through the array with `map` and use `index % 2` to decide which scale to apply.
> 
>     * In the curve components, you could `reduce` abscissas and build the array of points:
> 
> 
> ```ts
> points.push(x - capLength, y - yerr, x + capLength, y - yerr); // bottom cap
> points.push(x, y - yerr, x, y + yerr); // error bar
> ...
> ```

While handy for `ErrorBarCurve`, it will be a bit more convoluted in `DataCurve` as we would need to intertwine `abscissas` and `ordinates`.

Can we keep this for a future PR, perhaps ?

axelboc:
Oh yeah, for sure 👍

PR #301, opened 2020-11-10
Fixed NeXus vis that did not use `axes` values

### LINKED ISSUE #300: [Nexus] Vis no longer uses axes values
But they do use labels...

Probably since #297

loichuder:
Fix #300 and adds the relevant Cypress test to avoid unnoticed regression.

The source was the fact that `AxisParams` uses a member `values` while in the refacto #297 I did use `value` instead. Not sure why it was not caught by TS...


axelboc (approved):

loichuder:
300th issue/PR processed :partying_face:

axelboc:
🎉 😄

Issue #281, opened 2020-11-03
[NeXus] Add support for `NXdata@SILX_style` Nexus extension for defining signal scale


t20100:
The extra `NXData@SILX_style` attribute allows to define the default normalization to use for the Nexus signal and axes.
`@SILX_style` contains a json string with a mapping defining extra parameters:
- `"signal_scale_type"`: string (either `"log"` or `"linear"`)
- `"axes_scale_type"`: list of strings (also either `"log"` or `"linear"`)

The idea is to first support the `"signal_scale_type"` parameter which allows to define the default normalization of the colormap for image and the Y axis scale for curve with:

`@SILX_style = {"signal_scale_type": "log"}`

More exemples: https://github.com/silx-kit/silx/issues/2928#issuecomment-639572971

Related issue https://github.com/silx-kit/silx/issues/2928 and PR https://github.com/silx-kit/silx/pull/3092 with discussions.

loichuder:
- [ ] Support `axes_scale_type` for `LineVis`
- [ ] Support `signal_scale_type` for `LineVis`
- [ ] Support `signal_scale_type` for `HeatmapVis`

`axes_scale_type` for `HeatmapVis` seems out of scope. Also, it would imply reviewing completely how the texture is computed.

t20100:
Yes, there is no need for `axes_scale_type` for images.

axelboc:
Note that for `signal_scale_type`, the toolbar will need to be updated with the specified scale type.

Also, I suggest we limit the scope of this issue to supporting only linear/log/symlog for `signal_scale_type` on `HeatmapVis`.

loichuder:
[All milestones closed](https://github.com/silx-kit/h5web/issues/281#issuecomment-724811613). Closing this.

PR #303, opened 2020-11-12
Added support for silx axes_scale_style in LineVis

loichuder:
image

First step of #281

### DIFF
--- a/src/h5web/visualizations/nexus/hooks.ts
+++ b/src/h5web/visualizations/nexus/hooks.ts
@@ -19,6 +19,7 @@ import {
   getLinkedEntity,
   getDatasetLabel,
   getNxAxisNames,
+  parseSilxStyleAttribute,
 } from './utils';
 
 export function useLinkedDatasetValues(
@@ -65,7 +66,10 @@ export function useNxData(group: HDF5Group, metadata: HDF5Metadata): NxData {
 
   const axesNames = getNxAxisNames(group);
 
-  const axisMapping = axesNames.map((name) => {
+  const silxStyle = parseSilxStyleAttribute(group);
+  const { axes_scale_type } = silxStyle;
+
+  const axisMapping = axesNames.map((name, i) => {
     if (!name) {
       return undefined;
     }
@@ -79,7 +83,11 @@ export function useNxData(group: HDF5Group, metadata: HDF5Metadata): NxData {
     const value = datasetValues && datasetValues[name];
     assertOptionalArray<number>(value);
 
-    return { value, label: getDatasetLabel(dataset, name) };
+    return {
+      value,
+      label: getDatasetLabel(dataset, name),
+      scaleType: axes_scale_type && axes_scale_type[i],
+    };
   });
 
   const errors = datasetValues && datasetValues.errors;
--- a/src/h5web/visualizations/nexus/utils.ts
+++ b/src/h5web/visualizations/nexus/utils.ts
@@ -6,8 +6,13 @@ import {
   HDF5Entity,
 } from '../../providers/models';
 import { getEntity, isGroup } from '../../providers/utils';
-import { NxAttribute, NX_INTERPRETATIONS } from './models';
-import { assertArray } from '../shared/utils';
+import {
+  NxAttribute,
+  NX_INTERPRETATIONS,
+  RawSilxStyle,
+  SilxStyle,
+} from './models';
+import { assertArray, isScaleType } from '../shared/utils';
 
 export function isNxDataGroup(group: HDF5Group): boolean {
   return !!group.attributes?.find(({ value }) => value === 'NXdata');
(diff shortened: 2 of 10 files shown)


src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -65,7 +66,12 @@ export function useNxData(group: HDF5Group, metadata: HDF5Metadata): NxData {
 
   const axesNames = getNxAxisNames(group);
 
-  const axisMapping = axesNames.map((name) => {
+  const silxStyle = getAttributeValue(group, 'SILX_style');
+  const scaleStyles: SilxStyle | undefined =
+    typeof silxStyle === 'string' ? JSON.parse(silxStyle) : undefined;
```
axelboc, line 71: You may want to add ` && silxStyle` to the condition, as `JSON.parse("")` throws.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -105,3 +106,13 @@ export function getDatasetLabel(
...
 }
+
+export function normalizeSilxStyle(
+  silxScaleType: 'linear' | 'log' | undefined
+): ScaleType | undefined {
+  if (!silxScaleType) {
+    return undefined;
+  }
+
+  return silxScaleType === 'log' ? ScaleType.Log : ScaleType.Linear;
```
axelboc, line 117: I believe we could lowercase the values of the `ScaleType` enum to avoid having to resort to this utility function.

axelboc: We just need to check what SILX uses for `SymLog`, whether it's `symlog` or something else.

src/stories/LineVis.stories.tsx
```diff
@@ -90,6 +94,23 @@ CustomAbscissas.args = {
   },
 };
 
+export const SemilogX = Template.bind({});
+
+SemilogX.args = {
```
axelboc, line 99: Some naming/capitalisation inconsistencies between stories (with or without `ScaleFor`, SemiLog` with upper or lower case `L`).

axelboc (approved):
Roh, so good! 💯

src/h5web/visualizations/nexus/hooks.ts
loichuder: Ha :fishing_pole_and_fish: ! 

Thanks, not used to `JSON.parse`

src/stories/LineVis.stories.tsx
loichuder: In fact, I wrote `Semilog` (which is [a whole word](https://en.wiktionary.org/wiki/semilog)) to describe the plot. While `SymLog` (for Symmetric Log) is the scale used for the plot.

loichuder: Should I rather use `[Log|SymLog]ScaleFor[X/Y]` in every story name ?

src/h5web/visualizations/nexus/utils.ts
loichuder: According to @t20100, we can expect `symlog` as a possible scale type.

src/stories/LineVis.stories.tsx
loichuder: Instead, I use another level of hierarchy to put all scale-related stories in their own folder

loichuder:
@axelboc Could you have another look ? I notably added a dedicated parsing function for `SilxStyle` to palliate the blind typing from `any` to `SilxStyle`

src/h5web/visualizations/shared/utils.ts
```diff
@@ -204,3 +204,11 @@ export function assertOptionalArray<T>(
     assertArray<T>(val);
   }
 }
+
+export function isScaleType(val: unknown): val is ScaleType {
+  if (typeof val !== 'string') {
+    return false;
+  }
+
+  return (Object.values(ScaleType) as string[]).includes(val);
```
axelboc, line 213: `Object.values<string>()` should do it 😉

src/stories/LineVisDisplay.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+WithTitle.args = {
+  dataArray,
+  domain,
+  title: 'A simple curve',
+};
+
+export default {
+  title: 'Visualizations/LineVis/Display',
+  ...LineVisStoriesDefault,
```
axelboc, line 72: You can leave the defaults in `LineVis.stories.tsx` and then just import them from there:

```ts
import { default as LineVisStoriesConfig } from './LineVis.stories';

export default {
  ...LineVisStoriesConfig,
  title: '...'
}
```

axelboc (approved):

src/stories/LineVisDisplay.stories.tsx
loichuder: Clever :+1:

PR #304, opened 2020-11-12
Added support of signal_scale_type for LineVis

loichuder:
Once we agree on this, I will use the same process for the HeatmapVis

### DIFF
--- a/src/h5web/visualizations/containers/NxSpectrumContainer.tsx
+++ b/src/h5web/visualizations/containers/NxSpectrumContainer.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useState, useContext } from 'react';
+import React, { ReactElement, useState, useContext, useEffect } from 'react';
 import { range } from 'lodash-es';
 import { assertGroup } from '../../providers/utils';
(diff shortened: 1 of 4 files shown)


src/h5web/visualizations/containers/NxSpectrumContainer.tsx
```diff
@@ -34,6 +35,14 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
 
   const { signal, title, errors, axisMapping } = nxData;
 
+  const { setScaleType } = useLineConfig();
+
+  useEffect(() => {
+    if (signal.scaleType) {
+      setScaleType(signal.scaleType);
+    }
+  }, [setScaleType, signal.scaleType]);
```
loichuder, line 44: Tentative implementation

I also consider passing `signal.scaleType` as a `scaleType` prop of `MappedLineVis` and putting the `useEffect` in the `MappedLineVis`.

Let me know what you think.

axelboc (approved):
Hmm no it's excellent as is. It's a NeXus feature, so it makes sense to leave the `useEffect` it in the container, at least for now. 👍

PR #305, opened 2020-11-12
Added support of signal_scale_type for HeatmapVis

loichuder:
Same, same.

Final step of #281 

Love to put these green checkmarks everywhere :heavy_check_mark:


axelboc (approved):
On a roll 😄 ✔️

PR #306, opened 2020-11-13
Set up visual regression testing

### LINKED ISSUE #290: Visual regression tests
- Set up [BackstopJS](https://github.com/garris/BackstopJS) (or investigate Cypress plugins).
  - Must run in Docker container to avoid GPU rendering inconsistencies.
  - Must be able to run locally and in CI (against deploy previews).
- Test all mock data visualizations at `/mock` in desktop viewport.
- Test toolbar features.

axelboc:
Alright!! 😤 Fix #290.

Here is how it all works:

1. I push a branch and open a PR.
1. The `cypress` workflow starts and runs all the Cypress tests with `CYPRESS_TAKE_SNAPSHOTS=true`, which takes visual snapshots in the relevant tests.
1. Three outcomes are possible:

### All snapshots match

If all the snapshots match the version-controlled references snapshots, the workflow succeeds.

image

### At least one snapshot does not match

If a snapshot does not match its reference snapshot, the workflow fails.

1. The snapshot's diff is uploaded as a workflow artifact, so I can download and review it.

    image
    
    heatmap_2D diff

1. If the diff is intentional/expected, I go back to the PR and post a comment with the following text: `/cypress updateSnapshots=true`. This command triggers a run of the `cypress` workflow, in which Cypress is told to update the reference screenshots.

    image

1. The updated reference snapshots are automatically committed to a new `update-snapshots` branch and a PR is opened to merge this branch into my working branch.

    image

1. I then go to this new PR, review the updated snapshots, and merge.
    
    image
    
1. The `cypress` workflow runs again on my working branch. It now succeeds since all the snapshots match their reference snapshots.

### First-time snapshot

If a snapshot is taken for the first time and all the other snapshots match their reference snapshots, the workflow succeeds.

1. The new snapshot is automatically committed to a new `update-snapshots` branch and a PR is opened to merge this branch into my working branch.
1. I then go to this new PR, review the new snapshot, and merge.
1. The `cypress` workflow runs and succeeds again on my working branch. This time, the new snapshot has a reference snapshot, so nothing else happens.


cypress/specs/app.spec.ts
```diff
@@ -90,46 +115,34 @@ describe('App', () => {
...
       cy.findByRole('heading', { name: 'nexus_entry / image' }).should('exist');
       cy.findByRole('tab', { name: 'NX Image' }).should('exist');
-      cy.wait(500); // Monkey-patch: sometimes this test makes the following test hanging if the texture computation didn't end.
-    });
-
-    it('use signal name and units to compute title', () => {
-      cy.findByRole('treeitem', { name: 'nexus_entry' }).click();
-      cy.findByRole('treeitem', { name: 'spectrum' }).click();
-
-      cy.findByRole('figure', { name: 'oneD (arb. units)' }).should('exist');
```
axelboc, line 108: I've combined some of these tests for simplicity.

cypress.json
```diff
@@ -1,5 +1,9 @@
 {
   "baseUrl": "http://localhost:3000",
+  "integrationFolder": "cypress/specs",
+  "pluginsFile": "cypress/plugins.js",
+  "supportFile": "cypress/support.js",
+  "screenshotsFolder": "cypress/debug",
   "fixturesFolder": false,
-  "pluginsFile": false
+  "video": false
```
axelboc, line 8: Recording a video is time-consuming and a bit over the top for us. I think the debug screenshots are enough.

Note that I also rename a couple of folders, and I remove the unnecessary `plugins` and `support` folders by pointing directly at files at the root of the `cypress` folder.

src/h5web/visualizations/shared/AxisSystem.module.css
```diff
@@ -11,14 +11,13 @@
...
-
-.grid,
-.axis {
   display: block;
   overflow: visible;
+
+  /* Avoid potential overflow caused by default intrinsic size of `svg` elements. */
+  /* Chrome/Edge require values greater than 0. */
+  width: 1px;
+  height: 1px;
```
axelboc, line 20: The axes had disappeared on Chrome/Edge because of this.

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "baseUrl": ".",
```
axelboc, line 3: Avoids a warning when running Cypress.

axelboc:
I had to push the workflow that implements the `/cypress` command to `master` in order to test it: https://github.com/silx-kit/h5web/commit/dd708ec3b91c8b49602a1bff3181ab79a00fadf3

README.md
```diff
@@ -21,6 +21,25 @@ npm start
 - `npm run storybook` - manually test components in isolation in
   [Storybook](https://storybook.js.org/docs/react/get-started/introduction), at http://localhost:6006
 
+#### Cypress
+
+[Cypress](https://docs.cypress.io/guides/overview/why-cypress.html) is used to perform end-to-end and visual regression
+testing. To avoid rendering inconsistencies between platforms, Cypress must always be run inside a docker container.
```
axelboc, line 27: This is outdated. I'll update if you're happy with the new workflow.

axelboc:
Note that I have two ideas to improve all this:

1. Link the `update-snapshots` PR to the source PR with a `#123` reference, so we don't miss it (notably when the `cypress` workflow succeeds but includes first-time snapshots).
2. On failure, post a comment to the PR with the snapshot diffs so we don't have to go download them.

loichuder (approved):
Excellent !

I would have simply put a dedicated command `/updateScreenshots` rather than `/cypress ...` as we will not use this command to do other Cypress thingies.

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Serve ⚡️
+        run: npx serve -s build -l 3000 & npx wait-on http://localhost:3000
+
+      - name: Run Cypress 🌳
+        run: npm run cypress:run -- --env updateSnapshots=${{ github.event.inputs.updateSnapshots }}
+        env:
+          CYPRESS_TAKE_SNAPSHOTS: true
+
+      - name: Upload debug screenshots and diffs on failure 🎁
```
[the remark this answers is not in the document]
axelboc: Haha, fine by me. I'll use 🎁 for the PR step instead 😄

axelboc:
> I would have simply put a dedicated command /updateScreenshots rather than /cypress ... as we will not use this command to do other Cypress thingies.

Unfortunately, that would mean duplicating `cypress.yml` into `updateSnapshot.yml`, since updating the snapshots follows the same workflow as running the end-to-end tests (install/build/server/run Cypress).

What's annoying is that the slash-command action I found doesn't provide a way to alias a command -- I can't have a command called `/updateSnapshot` thats run `cypress.yml`. :/

tsconfig.json
axelboc: Something in the line of "no baseUrl defined, so TypeScript paths won't be supported" ... but we don't have `"paths"` in `tsconfig`, so it's meaningless.

loichuder:
Oh I see. I didn't realize that `commands: cypress` was both for the `/<command>` and for the name of the corresponding .yml. It is fine then !

loichuder (approved):
Some suggestions.

At some point, we should move the dev-related stuff out of the README (in CONTRIBUTING ?). For now, it is alright as the README is not that long.

README.md
```diff
@@ -17,10 +17,54 @@ npm start
...
+#### Visual regression
+
+Cypress is used for end-to-end testing, but also for visual regression testing. The idea is to take screenshots
+(referred to as "snapshots") of the app in various situations and compare them with previously approved "reference
+snapshots". If any pixel has changed, the test fails and a diff image that highlights the differences is created.
+
+Cypress is unable to take consistent screenshots across platforms because the exact rendering depends on the GPU (even
+when Docker is involved). For this reason, visual regression tests must always run on the same platform, that is, on the
+CI. Environment variable `CYPRESS_TAKE_SCREENSHOTS` is used to enable visual regression testing in the CI. This variable
+is set automatically in the `cypress.yml` GitHub workflow, which runs on push to any branch.
```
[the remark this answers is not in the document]
axelboc: It's important to check for regressions before merging to master. In my opinion, it should remain part of a PR review process.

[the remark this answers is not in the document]
axelboc: I would rather find ways to accelerate the CI workflow, with caching or by building the app without production optimisations for instance.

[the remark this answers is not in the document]
loichuder: > It's important to check for regressions before merging to master. In my opinion, it should remain part of a PR review process.

Yeah but you could imagine run the job only when there is a PR for the branch rather than for any branch.

README.md
```diff
@@ -17,10 +17,54 @@ npm start
...
   [Storybook](https://storybook.js.org/docs/react/get-started/introduction), at http://localhost:6006
 
+#### Visual regression
+
+Cypress is used for end-to-end testing, but also for visual regression testing. The idea is to take screenshots
+(referred to as "snapshots") of the app in various situations and compare them with previously approved "reference
+snapshots". If any pixel has changed, the test fails and a diff image that highlights the differences is created.
+
+Cypress is unable to take consistent screenshots across platforms because the exact rendering depends on the GPU (even
+when running in Docker). For this reason, visual regression tests are only run on the CI CI. Environment variable
```
loichuder, line 33: Typo:
```suggestion
when running in Docker). For this reason, visual regression tests are only run on the CI. Environment variable
```

PR #309, opened 2020-11-16
Refactored error bar curve computation

### LINKED ISSUE #308: [Line] Review error bars display
I have noticed a few issues:

- Plots with close points have extremely small caps:
image

This is because the cap length depends on the abscissas spacing. It might be better to set a fixed length instead.

- Incomplete bars when log scale is on:
image

Not sure about this one. The data point is at `x=1` and the errors are `+/- 1` so the point should be visible. But more generally, errors bars should only be displayed if the data point is visible.

loichuder:
Fix #308 

I had to review a bit how to compute the error bar segments so having a shared `useDataGeometry` was no longer relevant.


src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -21,7 +23,26 @@ function DataCurve(props: Props): ReactElement {
...
+        // Move NaN/Infinity out of the camera FOV (negative val for logScale).
+        // This allows to have only curve segments for the positive values
+        Number.isFinite(ordinate) ? 0 : camera.far
+      );
+    });
+
+    const geometry = new BufferGeometry();
+    geometry.setFromPoints(points);
+    return geometry;
+  }, [abscissaScale, abscissas, camera, ordinateScale, ordinates]);
```
loichuder, line 45: Inlined `useDataGeometry`

src/h5web/visualizations/line/ErrorBarCurve.tsx
```diff
@@ -1,51 +1,73 @@
...
+
+  const { abscissaScale, ordinateScale } = useCanvasScales();
+
+  const errorGeometry = useMemo(() => {
+    const points = ordinates.flatMap((val, index) => {
+      const [x, y] = [abscissaScale(abscissas[index]), ordinateScale(val)];
+
+      if (!Number.isFinite(y)) {
+        return [];
+      }
```
loichuder, line 36: Do not show error bar if the point has no finite coordinates.

src/h5web/visualizations/line/ErrorBarCurve.tsx
```diff
@@ -1,51 +1,73 @@
...
+
+      const error = errors[index];
+      if (error === 0) {
+        return [];
+      }
```
loichuder, line 41: Do not show error bars if there is no errors

src/h5web/visualizations/line/ErrorBarCurve.tsx
```diff
@@ -1,51 +1,73 @@
...
+      const yerr_min = ordinateScale(val - error);
+      const yerr_max = ordinateScale(val + error);
+
+      return [
+        // First segment: bottom cap
+        ...lineSegment(
+          x - capLength / 2,
```
loichuder, line 48: `capLength` is now a fixed length in canvas coordinates.

src/h5web/visualizations/line/ErrorBarCurve.tsx
```diff
@@ -1,51 +1,73 @@
...
+          yerr_min,
+          x + capLength / 2,
+          yerr_min
+        ),
+        // Second segment: bottom error bar
+        ...lineSegment(x, yerr_min, x, y),
+        // Third segment: top error bar
+        ...lineSegment(x, y, x, yerr_max),
```
loichuder, line 56: The error bar is now in two segments to be able to show the bottom/top part even though the other is not shown.

src/h5web/visualizations/line/MappedLineVis.tsx
```diff
@@ -52,7 +52,11 @@ function MappedLineVis(props: Props): ReactElement {
       return data;
     }
 
-    return data.flatMap((value, i) => [value - errors[i], value + errors[i]]);
+    return data.flatMap((value, i) => [
+      value - errors[i],
+      value,
```
loichuder, line 57: Adding `value` to the array used to compute the domain to handle boundary cases such the one shown as second point of #308.

src/h5web/visualizations/shared/hooks.ts
```diff
@@ -47,3 +49,17 @@ export function useFrameRendering(): void {
     setNum(Math.random());
   });
 }
+
+export function useCanvasScales(): {
```
loichuder, line 53: Introduced a new handy hook.

loichuder (commented):
To be merged after #306

src/h5web/visualizations/shared/utils.ts
```diff
@@ -212,3 +213,14 @@ export function isScaleType(val: unknown): val is ScaleType {
...
 }
+
+export function lineSegment(
+  startX: number,
+  startY: number,
+  endX: number,
+  endY: number
+): [Vector3, Vector3] | [] {
+  return [startX, startY, endX, endY].every(isFinite)
+    ? [new Vector3(startX, startY, 0), new Vector3(endX, endY, 0)]
```
axelboc, line 224: It's a shame that `DataCurve` doesn't benefit from this `every(isFinite)`. Perhaps what's needed is more a helper to make a `Vector3` from two values? This would avoid having to destructure the result of `lineSegment`... but then you'd have to filter for `undefined` afterwards. Hmm... Not sure.

axelboc (approved):
Looks good! Happy to see the `for` loop go. 😄

Maybe you could still have a hook to abstract out `useMemo` + `new BufferGeometry()` + `setFromPoints`?

I have a small performance concern about the use of `flatMap` (over `reduce`, probably). But it's always best not to optimise too early. 👍

src/h5web/visualizations/shared/utils.ts
loichuder: The aim of this function is to provide segments for `LineSegment` geometry` while `DataCurve` uses the `Line` geometry`. The difference is that `Line` draws segments between successive points while `LineSegments` draws segments between couples of points.

To use this in `DataCurve`, I would have to switch to `LineSegment`. It is possible but I would have to put explicitly the segments between successive points (i.e. duplicating points).

axelboc: I was more suggesting replacing `lineSegment` with `makePoint` or similar (i.e. to return a single `Vector`) so that you could use it in `DataCurve` without changing the geometry. In `ErrorBarCurve`, you would then call the functiion twice per line. But I'm not convinced this simplies things, so nevermind 😉

loichuder: Yeah and also, it changes the behaviour I wanted: if one of the point of the segment has not finite coordinates, I want to skip[ to skip the whole segment and not only the point in question.

PR #320, opened 2020-11-19
Show error when encountering malformed NeXus metadata

### LINKED ISSUE #254: Error management for malformed NeXuS structure

axelboc:
Fix #254 

The `supportsEntity`  functions now throw if they encounter malformed NeXus metadata, for instance if they're passed an entity with a `default` attribute that cannot be resolved to an `NXdata` group.

If a `default` attribute resolves to an entity with an `NX_class` attribute equal to `nxdata` instead of `NXData`, `supportsEntity` does not throw. In other words, we don't _actively_ look for malformed metadata; we just perform basic sanity checks (like in the vis containers).

I'm planning a bunch of follow-up refactors, which I'll mention in the comments.


cypress/specs/app.spec.ts
```diff
@@ -172,5 +172,12 @@ describe('App', () => {
         cy.matchImageSnapshot('logspectrum');
       }
     });
+
+    it('visualize dataset with malformed NeXus metadata', () => {
+      cy.findByRole('treeitem', { name: 'invalid' }).click();
+      cy.findByRole('treeitem', { name: 'default_not_found' }).click();
+
+      cy.findByText('Expected to find NXdata group at path "/test"');
+    });
```
axelboc, line 181: Adding one E2E test to make sure an error thrown in a `supportsEntity` function is indeed displayed.

src/h5web/providers/mock/data.ts
```diff
@@ -271,6 +272,15 @@ export const mockMetadata = makeMetadata({
...
       ]
     ),
+    makeGroup('invalid', undefined, [
+      makeGroupHardLink('default_not_string'),
+      makeGroupHardLink('default_empty'),
+      makeGroupHardLink('default_not_found'),
+      makeGroupHardLink('no_signal'),
+      makeGroupHardLink('signal_not_string'),
+      makeGroupHardLink('signal_dataset_not_numeric'),
+      makeGroupHardLink('signal_dataset_not_simple'),
```
axelboc, line 282: The only error not covered here is the dimension check in `NXSpectrum`'s `supportsEntity`. I wasn't sure how to create a dataset with zero dimension. A value of `[]` would already indicate one dimension, so it wouldn't make much sense.

src/h5web/providers/mock/data.ts
```diff
@@ -271,6 +272,18 @@ export const mockMetadata = makeMetadata({
...
+      makeGroupHardLink('default_not_string'),
+      makeGroupHardLink('default_empty'),
+      makeGroupHardLink('default_not_found'),
+      makeGroupHardLink('no_signal'),
+      makeGroupHardLink('signal_not_string'),
+      makeGroupHardLink('signal_not_found'),
+      makeGroupHardLink('signal_not_dataset'),
+      makeGroupHardLink('signal_dataset_not_numeric'),
+      makeGroupHardLink('signal_dataset_not_simple'),
+      makeGroupHardLink('signal_dataset_zero_dim'),
```
axelboc, line 285: Every error thrown by `supportsEntity` is covered. This will allow me to replace some of the `supportsEntity` unit tests with feature tests (with `renderApp()`). I'll do this in another PR.

src/h5web/providers/utils.ts
```diff
@@ -62,10 +62,11 @@ export function isNumericType(type: HDF5Type): type is HDF5NumericType {
 }
 
 export function assertDataset(
-  entity: HDF5Entity
+  entity: HDF5Entity,
+  message = 'Expected dataset'
 ): asserts entity is HDF5Dataset {
   if (!isDataset(entity)) {
-    throw new Error('Expected dataset');
+    throw new Error(message);
```
axelboc, line 69: Allows providing more explicit assertion error messages.

src/h5web/visualizations/index.tsx
```diff
@@ -112,36 +116,38 @@ export const VIS_DEFS: Record<Vis, VisDef> = {
...
       const interpretation = getAttributeValue(dataset, 'interpretation');
-      return (
-        (!isNxInterpretation(interpretation) && dimsCount === 1) || // NxImage already supports datasets with 2+ dimensions
-        interpretation === NxInterpretation.Spectrum
-      );
+      if (isNxInterpretation(interpretation)) {
+        return interpretation === NxInterpretation.Spectrum;
+      }
+
+      return dimsCount === 1; // NxImage already supports datasets with 2+ dimensions
```
axelboc, line 150: I don't like this coupling between the two nexus `supportsEntity` functions. I'm thinking of refactoring this to do something similar to `RawVis`: if the two nexus are supported, then `getSupportedVis` should return only one.

src/h5web/visualizations/index.tsx
```diff
@@ -151,36 +157,38 @@ export const VIS_DEFS: Record<Vis, VisDef> = {
...
+        `Expected "${signal}" dataset to have numeric type`
+      );
+      assertSimpleShape(
+        dataset.shape,
+        `Expected "${signal}" dataset to have simple shape`
+      );
 
       const dimsCount = dataset.shape.dims.length;
       if (dimsCount < 2) {
         return false;
```
axelboc, line 183: I don't throw if `dimsCount` equals `0` like `NXSpectrum`'s supportsEntity` does. Again, not a fan of doing things differently here. I'd like the two functions to be exactly the same apart from the `interpretation` equality. Again, I'll refactor this later to not clutter this PR.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -67,31 +62,35 @@ export function getNxDataGroup(
     return undefined;
   }
 
-  if (isNxDataGroup(entity)) {
-    return entity;
+  if (getAttributeValue(entity, 'NX_class') === 'NXdata') {
+    return entity; // `NXdata` group found
```
axelboc, line 66: I now look for the `NX_class` attribute directly instead of looking through all the attributes for a value equal to `NXdata`.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -67,31 +62,35 @@ export function getNxDataGroup(
...
-      return getLinkedEntity(currentComponent, previousEntity, metadata);
+  const defaultEntity = pathSegments.reduce<HDF5Entity | undefined>(
+    (parentEntity, currSegment) => {
+      return parentEntity && isGroup(parentEntity)
+        ? getLinkedEntity(currSegment, parentEntity, metadata)
+        : undefined;
     },
-    defaultPath.startsWith('/') ? metadata.groups[metadata.root] : entity
+    isAbsolutePath ? metadata.groups[metadata.root] : entity
   );
```
axelboc, line 86: I had trouble going through this algorithm again, so I renamed and refactored a few things.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -67,31 +62,35 @@ export function getNxDataGroup(
...
 
-  return getNxDataGroup(defaultEntity, metadata);
+  const nxDataGroup = getNxDataGroup(defaultEntity, metadata);
+  if (!nxDataGroup) {
+    throw new Error(`Expected to find NXdata group at path "${defaultPath}"`);
+  }
```
axelboc, line 91: When we first call `getNxDataGroup` from `supportsEntity`, we don't want it to throw if the entity is not a group for instance. However, when we call it after following a `default` path, we _do_ want it to throw. This is the simplest way I found to achieve this behaviour.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -180,9 +180,21 @@ export function getTickFormatter(
...
 
-export function assertStr(val: unknown): asserts val is string {
+export function assertDefined<T>(
+  val: T,
+  message = 'Expected some value'
+): asserts val is NonNullable<T> {
+  if (val === undefined) {
+    throw new TypeError(message);
+  }
+}
```
axelboc, line 190: For info, when calling `assertDefined` with a value of type `unknown`, the return will still be `unknown`.

src/h5web/visualizer/utils.ts
```diff
@@ -2,19 +2,30 @@ import type { HDF5Entity, HDF5Metadata } from '../providers/models';
...
+    // Remove Raw vis if any other vis is supported
+    const supportedVis =
+      supported.length > 1
+        ? supported.filter((vis) => vis !== Vis.Raw)
+        : supported;
+
+    return { supportedVis };
+  } catch (error) {
+    return { supportedVis: [], error };
+  }
```
axelboc, line 30: The `try` wraps the `reduce`, so the function returns on the first error thrown by any of the `supportsEntity` functions.

src/h5web/providers/mock/data.ts
axelboc: Done in this PR after all.

loichuder (approved):
Nice :)

src/h5web/providers/mock/data.ts
```diff
@@ -388,6 +437,7 @@ const fourD = arr4.map((divider) =>
 );
 
 export const mockValues = {
+  null: null,
```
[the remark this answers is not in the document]
axelboc: yeah yeah, in ES2015, it's all good 👍

src/h5web/visualizer/utils.ts
axelboc: This was my first idea, but it breaks the rule of hooks.

loichuder: For which hook ?

axelboc: `const [activeVis, setActiveVis] = useState<Vis>();`

axelboc: As discussed, I can move the hook up but the `try/catch` has to wrap the rest of the body of the `Visualizer` function, so it's pretty ugly.

PR #323, opened 2020-11-19
Added toggler for error bars in NxSpectrum vis

### LINKED ISSUE #319: [NeXus] Toggler for error bars
It might be a good idea to allow users to hide the error bars. I reckon the toolbar control should appear on NeXus visualizations and be disabled if the NXdata group doesn't container an `errors` or `DATASETNAME_errors` dataset.

loichuder:
Fix #319 

image

Also made use of `visible` to toggle the visibility of Three.js objects (https://discoverthreejs.com/tips-and-tricks/#disposing-of-things). Did not see a great improvement but seems a bit more fluid.


src/h5web/toolbar/NxSpectrumToolbar.tsx
```diff
@@ -0,0 +1,23 @@
...
+import LineToolbar from './LineToolbar';
+
+function NxSpectrumToolbar(): ReactElement {
+  const { showErrors, toggleErrors, areErrorsDisabled } = useNxSpectrumConfig();
+
+  return (
+    <LineToolbar>
+      <ToggleBtn
+        label="Error bars"
+        icon={(props) => <MdFitnessCenter transform="rotate(45)" {...props} />}
```
loichuder, line 14: :neutral_face:  ...

Taking proposals for the icon :rofl:

src/h5web/visualizations/containers/models.ts
```diff
@@ -1,6 +1,11 @@
+import { ReactElement } from 'react';
 import type { HDF5Entity } from '../../providers/models';
 
 export interface VisContainerProps {
   entity: HDF5Entity;
   entityName?: string;
 }
+
+export interface ToolbarProps {
+  children?: ReactElement;
```
loichuder, line 10: Issue with the typing here. I cannot put `ReactElement[]` as I get typing errors in `LineToolbar` (probably due to the `Toolbar` component).

src/h5web/visualizations/nexus/spectrumConfig.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export const useNxSpectrumConfig = create<NxSpectrumConfig>(
+  (set): NxSpectrumConfig => ({
+    showErrors: true,
+    areErrorsDisabled: false,
+    toggleErrors: () => set((state) => ({ showErrors: !state.showErrors })),
+    disableErrors: (areErrorsDisabled: boolean) => set({ areErrorsDisabled }),
+  })
+);
```
loichuder, line 17: I felt that a "classic" (without persistence) zustand store was enough.

axelboc (approved):
> Issue with the typing here. I cannot put ReactElement[] as I get typing errors in LineToolbar (probably due to the Toolbar component).

This issue is mind-boggling. I haven't found a way to fix, but since it makes sense for `LineVisToolbar` to accept the same props as `Toolbar`, I think the cleanest is to resort to `as any`. I pushed a commit.

src/h5web/toolbar/NxSpectrumToolbar.tsx
axelboc: What about:

image

axelboc: I think the label of the toggler could also just be `Errors`

PR #324, opened 2020-11-19
Remove coupling between `supportsEntity` functions of NeXus vis

axelboc:
The `supportsEntity` function of `NxSpectrum` no longer worries about the behaviour of `NxImage`. The spectrum vis supports nD datasets, not just 1D datasets, so that's what it checks for. The `NxSpectrum` is removed from the `supportedVis` array later, if `NxImage` is also supported. This matches the behaviour of the `Raw` vis.

### DIFF
--- a/src/h5web/visualizations/index.tsx
+++ b/src/h5web/visualizations/index.tsx
@@ -149,7 +149,7 @@ export const VIS_DEFS: Record<Vis, VisDef> = {
         return interpretation === NxInterpretation.Spectrum;
       }
 
-      return dimsCount === 1; // NxImage already supports datasets with 2+ dimensions
+      return true;
     },
   },
 
@@ -181,6 +181,12 @@ export const VIS_DEFS: Record<Vis, VisDef> = {
       );
 
       const dimsCount = dataset.shape.dims.length;
+      if (dimsCount === 0) {
+        throw new Error(
+          `Expected "${signal}" dataset to have at least one dimension`
+        );
+      }
+
       if (dimsCount < 2) {
         return false;
       }
--- a/src/h5web/visualizer/utils.test.ts
+++ b/src/h5web/visualizer/utils.test.ts
@@ -1,46 +1,50 @@
 import { getSupportedVis } from './utils';
 import { Vis } from '../visualizations';
-import {
-  makeDataset,
-  makeSimpleDataset,
-  makeAttr,
-  makeStrAttr,
-  intType,
-  compoundType,
-  scalarShape,
-  makeGroup,
-  makeMetadata,
-} from '../providers/mock/data';
-
-const metadata = makeMetadata();
+import { mockMetadata } from '../providers/mock/data';
 
 describe('Visualizer utilities', () => {
   describe('getSupportedVis', () => {
-    it('should identify dataset as supporting raw vis', () => {
-      const dataset = makeDataset('foo', compoundType, scalarShape);
-      const supportedVis = getSupportedVis(dataset, metadata);
+    it('should return supported visualizations', () => {
+      const supportedVis = getSupportedVis(
+        mockMetadata.datasets?.raw,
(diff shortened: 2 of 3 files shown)


src/h5web/visualizations/index.tsx
```diff
@@ -181,6 +181,12 @@ export const VIS_DEFS: Record<Vis, VisDef> = {
       );
 
       const dimsCount = dataset.shape.dims.length;
+      if (dimsCount === 0) {
+        throw new Error(
+          `Expected "${signal}" dataset to have at least one dimension`
+        );
+      }
```
axelboc, line 188: This check makes sense for `NxImage` as well, since a signal dataset with no dimension at all is wrong. The two `supportsEntity` function are now exactly identical except for the interpretation checks and for the dimension check with `NxImage`.

src/h5web/visualizer/utils.test.ts
```diff
@@ -1,46 +1,50 @@
...
 describe('Visualizer utilities', () => {
   describe('getSupportedVis', () => {
-    it('should identify dataset as supporting raw vis', () => {
-      const dataset = makeDataset('foo', compoundType, scalarShape);
-      const supportedVis = getSupportedVis(dataset, metadata);
+    it('should return supported visualizations', () => {
+      const supportedVis = getSupportedVis(
+        mockMetadata.datasets?.raw,
+        mockMetadata
+      );
```
axelboc, line 11: I realised that I could just use the mock datasets and groups directly...

src/h5web/visualizer/utils.ts
```diff
@@ -1,6 +1,18 @@
...
+    // Remove `Raw` or `NxSpectrum` vis if any other vis is supported
+    (visArr[0] === Vis.Raw || visArr[0] === Vis.NxSpectrum) &&
+    visArr.length > 1
+  ) {
+    return visArr.slice(1);
+  }
+
+  return visArr;
+}
+
```
axelboc, line 15: This is where we remove `NxSpectrum` from the array of supported visualizations when `NxImage` is also supported.

loichuder: How can we be sure that `Raw` or `NxSpectrum` will always be in first position ?

src/h5web/visualizer/utils.ts
```diff
@@ -1,6 +1,18 @@
 import type { HDF5Entity, HDF5Metadata } from '../providers/models';
 import { VIS_DEFS, Vis } from '../visualizations';
 
+function applySpecialVisRules(visArr: Vis[]): Vis[] {
```
loichuder, line 4: `removeUnneededVis` ? 

I wonder if this should not be kept inlined (with a ternary) to avoid the naming problem.

loichuder (changes requested):

src/h5web/visualizer/utils.ts
axelboc: Because the vis definitions are ordered. I considered using a more complex algorithm, but thought it might over-engineering. We have tests to check that `Raw` and `NxSpectrum` are indeed removed, so if we re-order the vis definitions, it will catch the regression.

src/h5web/visualizer/utils.ts
axelboc: 👍 

I didn't want to have too much code inside the `try`/`catch`, and I thought a function could be a bit more self-documenting.

src/h5web/visualizer/utils.ts
loichuder: Not a fan of relying on this ordering but as long as we can catch the regression...

axelboc: Yeah... Though the ordering is critical visually but also for the default vis selection, so this function is not the only thing that's coupled to it.

PR #325, opened 2020-11-20
Rethink checks and assertions when looking for NXdata group

axelboc:


src/h5web/visualizations/index.tsx
```diff
@@ -115,7 +116,11 @@ export const VIS_DEFS: Record<Vis, VisDef> = {
     Toolbar: LineToolbar,
     Container: NxSpectrumContainer,
     supportsEntity: (entity, metadata) => {
-      const group = getNxDataGroup(entity, metadata);
+      if (!isGroup(entity)) {
+        return false;
+      }
+
+      const group = findNxDataGroup(entity, metadata);
```
axelboc, line 123: Checking that the entity is a group with `isGroup` before looking for the `NXdata` group is consistent with:

- the `isDataset` calls in the `supportsEntity` functions of the other visualizations;
- the `assertGroup` calls in the spectrum and image vis containers.

Changing the type of the first argument of `getNxDataGroup` also brings more simplification within this function (see below).

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -54,19 +59,15 @@ export function getNxAxisNames(group: HDF5Group): (string | undefined)[] {
...
 }
 
-export function getNxDataGroup(
-  entity: HDF5Entity | undefined,
+export function findNxDataGroup(
+  group: HDF5Group,
   metadata: HDF5Metadata
 ): HDF5Group | undefined {
-  if (!entity || !isGroup(entity)) {
-    return undefined;
```
axelboc, line 62: Since we already know that the first argument is a group (because this has been checked/asserted before calling `findNxDataGroup`), we can remove this check.

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -82,15 +83,13 @@ export function getNxDataGroup(
...
-    isAbsolutePath ? metadata.groups[metadata.root] : entity
+    isAbsolutePath ? metadata.groups[metadata.root] : group
   );
 
-  const nxDataGroup = getNxDataGroup(defaultEntity, metadata);
-  if (!nxDataGroup) {
-    throw new Error(`Expected to find NXdata group at path "${defaultPath}"`);
-  }
+  assertDefined(defaultEntity, `Expected entity at path "${defaultPath}"`);
+  assertGroup(defaultEntity, `Expected group at path "${defaultPath}"`);
```
axelboc, line 90: Now we can just assert that the entity found by the `reduce` is defined and is a group before recursing.

src/h5web/visualizations/index.tsx
axelboc: Note that I have renamed `getNxDataGroup` to `findNxDataGroup` as an attempt to make functions that use assertions stand out a bit more from those that don't (like `getLinkedEntity`).

`find*` functions live at a higher level: they call `get*` functions and assert on the results. My plan is to introduce more `find*` functions, like `findSignalDataset`.

loichuder (approved):

PR #326, opened 2020-11-20
Add generic types to HDF5Dataset and improve assertions

axelboc:
I just had a spark on how we can better type `HDF5Dataset` so that our assertions can accept `HDF5Dataset` objects rather than `HDF5Shape` or `HDF5Type`. This should open up some refactoring opportunities. 🎇 

Apart from that, I use single-line assertions in a few more places to make the code more readable.

### DIFF
--- a/.env
+++ b/.env
@@ -16,3 +16,7 @@ REACT_APP_HSDS_DOMAIN=
 
 # Rendering performance profiling (local development only)
 REACT_APP_PROFILING_ENABLED=false
+
+# Logging limit of React Testing Library
+# https://testing-library.com/docs/dom-testing-library/api-helpers/#debugging
+DEBUG_PRINT_LIMIT=
--- a/src/h5web/providers/utils.ts
+++ b/src/h5web/providers/utils.ts
@@ -79,21 +79,25 @@ export function assertGroup(
   }
 }
 
-export function assertNumericType(
-  type: HDF5Type,
-  message = 'Expected numeric type'
-): asserts type is HDF5NumericType {
-  if (!isNumericType(type)) {
-    throw new Error(message);
+export function assertNumericType<S extends HDF5Shape>(
+  dataset: HDF5Dataset<S>
+): asserts dataset is HDF5Dataset<S, HDF5NumericType> {
+  if (!isNumericType(dataset.type)) {
+    throw new Error('Expected dataset to have numeric type');
   }
 }
 
-export function assertSimpleShape(
-  shape: HDF5Shape,
-  message = 'Expected simple shape'
-): asserts shape is HDF5SimpleShape {
+export function assertSimpleShape<T extends HDF5Type>(
+  dataset: HDF5Dataset<HDF5Shape, T>
+): asserts dataset is HDF5Dataset<HDF5SimpleShape, T> {
+  const { shape } = dataset;
+
   if (!isSimpleShape(shape)) {
-    throw new Error(message);
+    throw new Error('Expected dataset to have simple shape');
+  }
+
+  if (shape.dims.length === 0) {
+    throw new Error('Expected dataset with simple shape to have dimensions');
   }
 }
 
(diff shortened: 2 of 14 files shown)


.env
```diff
@@ -16,3 +16,7 @@ REACT_APP_HSDS_DOMAIN=
 
 # Rendering performance profiling (local development only)
 REACT_APP_PROFILING_ENABLED=false
+
+# Logging limit of React Testing Library
+# https://testing-library.com/docs/dom-testing-library/api-helpers/#debugging
+DEBUG_PRINT_LIMIT=
```
axelboc, line 22: When Testing Library doesn't find an element, it logs a whole bunch of DOM elements, which makes it difficult to see which test has failed. Set `DEBUG_PRINT_LIMIT=0` in `.env.local` to remove these logs.

src/h5web/providers/utils.ts
```diff
@@ -79,21 +79,25 @@ export function assertGroup(
...
+  const { shape } = dataset;
+
   if (!isSimpleShape(shape)) {
-    throw new Error(message);
+    throw new Error('Expected dataset to have simple shape');
+  }
+
+  if (shape.dims.length === 0) {
+    throw new Error('Expected dataset with simple shape to have dimensions');
   }
```
axelboc, line 101: A simple-shaped dataset with a dimension array equal to `[]` makes no sense. Since it is quite a niche error, I move the assertion here to not clutter the important code.

src/h5web/visualizations/containers/HeatmapVisContainer.tsx
```diff
@@ -11,10 +14,10 @@ import { VisContainerProps } from './models';
 function HeatmapVisContainer(props: VisContainerProps): ReactElement {
   const { entity, entityName } = props;
   assertDataset(entity);
+  assertSimpleShape(entity);
+  assertNumericType(entity);
```
axelboc, line 18: Adding assertions here for consistency with `supportsEntity`, and to avoid the type cast below.

src/h5web/visualizations/heatmap/HeatmapVis.tsx
```diff
@@ -43,21 +47,14 @@ function HeatmapVis(props: Props): ReactElement {
...
   const abscissas = getPixelEdges(abscissaParams.value, cols);
-  const ordinates = getPixelEdges(ordinateParams.value, rows);
-
   const abscissaToIndex = getValueToIndexScale(abscissas);
-
   const abscissaDomain = useMemo(() => getDomain(abscissas), [abscissas]);
-  if (!abscissaDomain) {
-    throw new Error(`Abscissas have undefined domain`);
-  }
+  assertDefined(abscissaDomain, 'Abscissas have undefined domain');
```
axelboc, line 52: Using `assertDefined` in a few more places.

src/h5web/visualizations/index.tsx
```diff
@@ -132,22 +132,8 @@ export const VIS_DEFS: Record<Vis, VisDef> = {
...
-      );
-
-      const dimsCount = dataset.shape.dims.length;
-      if (dimsCount === 0) {
-        throw new Error(
-          `Expected "${signal}" dataset to have at least one dimension`
-        );
-      }
+      assertNumericType(dataset);
+      assertSimpleShape(dataset);
```
axelboc, line 136: Note that I'm no longer logging the signal's name. I feel like the debugging benefit is not sufficient to justify the clutter in the code. In the future, the assertion functions could take care of this, though (through a hypotetic `dataset.name` property -- on this note, I wonder whether we could set this property in `getLinkedEntity` and return an `HDF5NamedEntity` of sorts...)

loichuder: You could the same trick you used on other assertion functions: using the error message as parameter to log the signal name.

Anyhow, I think it is fine like this

loichuder (approved):
Nice idea !

PR #328, opened 2020-11-23
Use generic types in HDF5 type guards

axelboc:
Follow up to #326

### DIFF
--- a/src/h5web/metadata-viewer/EntityInfo.tsx
+++ b/src/h5web/metadata-viewer/EntityInfo.tsx
@@ -1,7 +1,7 @@
 import React, { ReactElement } from 'react';
 import type { HDF5Entity } from '../providers/models';
 import styles from './MetadataViewer.module.css';
-import { isBaseType, isSimpleShape } from '../providers/utils';
+import { isDataset, isDatatype, hasSimpleShape } from '../providers/utils';
 import { renderShapeDims } from './utils';
 import RawInspector from './RawInspector';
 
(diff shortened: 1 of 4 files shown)


src/h5web/metadata-viewer/EntityInfo.tsx
```diff
@@ -24,17 +24,21 @@ function EntityInfo(props: Props): ReactElement {
...
-        {'type' in entity && (
+        {(isDataset(entity) || isDatatype(entity)) && (
           <tr>
             <th scope="row">Type</th>
-            <td>{isBaseType(entity.type) ? entity.type.class : 'Complex'}</td>
+            <td>
+              {typeof entity.type === 'string'
+                ? entity.type
+                : entity.type.class}
+            </td>
```
axelboc, line 34: The term `Complex` was confusing and the condition was hiding useful information.

src/h5web/metadata-viewer/EntityInfo.tsx
```diff
@@ -24,17 +24,21 @@ function EntityInfo(props: Props): ReactElement {
         </tr>
       </thead>
       <tbody>
-        {'type' in entity && (
+        {(isDataset(entity) || isDatatype(entity)) && (
           <tr>
             <th scope="row">Type</th>
-            <td>{isBaseType(entity.type) ? entity.type.class : 'Complex'}</td>
+            <td>
+              {typeof entity.type === 'string'
```
loichuder, line 31: This condition puzzles me.

`entity.type` can be `string` or something that has a `class` ?

loichuder (approved):
Looks nicer !

src/h5web/metadata-viewer/EntityInfo.tsx
axelboc: Yeah, it can be the ID of a datatype entity.

PR #330, opened 2020-11-24
MappedLineVis now expect nD errors 

loichuder:
(should have the same dimensionality as value)

- Using now `ndarray` for errors to slice through appropriate dimensions
- Changed `getDomain` to find the extent of values with errors


src/h5web/visualizations/shared/hooks.ts
```diff
@@ -40,6 +40,18 @@ export function useMappedArray<T>(
...
+  value: T[],
+  rawDims: number[],
+  dimensionMapping: DimensionMapping
+) {
+  const baseArray = useBaseArray(value, rawDims) as ndarray<T>;
+  return {
+    baseArray,
+    dataArray: useMappedArray(baseArray, dimensionMapping) as ndarray<T>,
+  };
+}
```
loichuder, line 53: I had to introduce `undefined` in `use...Array` hooks but I cannot make TS infer that the array will be defined if the value in input is defined.

To fix this, I added this hook that makes the necessary assertions when `value` is defined to get arrays that are always defined. I then use this hook in all `Mapped...` component.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -45,37 +45,52 @@ export function computeVisSize(
...
+export function getExtent(
+  values: number[],
+  initialExtent?: [number, number, number]
+): [number, number, number] {
+  return values.reduce((acc, val) => {
+    const [accMin, accMax, accPositiveMin] = acc;
+    return [
+      Math.min(val, accMin),
+      Math.max(val, accMax),
+      val > 0 ? Math.min(val, accPositiveMin) : accPositiveMin,
```
loichuder, line 57: I tried to lighten a bit the code here by using `Math.min`, `Math.max` and setting initially `accPositiveMin` to `Infinity`.

src/h5web/visualizations/shared/utils.ts
```diff
@@ -45,37 +45,52 @@ export function computeVisSize(
...
+
+  const [min, max, positiveMin] = errors
+    ? values.reduce(
+        (acc, val, i) => {
+          const err = errors[i];
+          return getExtent([val - err, val, val + err], acc);
+        },
+        [values[0], values[0], Infinity]
+      )
+    : getExtent(values);
```
loichuder, line 85: I am not extremely happy about how this turned out. Happy to get some ideas.

src/h5web/providers/mock/data.ts
```diff
@@ -397,6 +409,12 @@ export const mockMetadata = makeMetadata({
       [makeStrAttr('units', 'arb. units')]
     ),
     makeSimpleDataset('twoD', intType, [20, 41]),
+    makeSimpleDataset(
+      'twoD_spectrum',
+      intType,
+      [20, 41],
+      [makeStrAttr('interpretation', 'spectrum')]
+    ),
```
loichuder, line 417: A new nD dataset to test `errors`

src/h5web/providers/mock/data.ts
```diff
@@ -415,10 +403,28 @@ export const mockMetadata = makeMetadata({
...
+      ]
+    ),
+    makeSimpleDataset('errors_twoD', floatType, [20, 41]),
+    makeSimpleDataset(
+      'NxFourD',
+      intType,
+      [3, 9, 20, 41],
+      [
+        makeStrAttr('long_name', 'Interference fringes'),
+        makeStrAttr('interpretation', 'image'),
```
loichuder, line 425: This fixes the misplaced `interpretation` attribute: it should be on the signal dataset and not on the NXdata group

src/h5web/visualizations/shared/hooks.ts
```diff
@@ -9,18 +9,18 @@ import { getCanvasScale, getDomain } from './utils';
...
 
 export function useMappedArray<T>(
-  baseArray: ndarray<T>,
+  baseArray: ndarray<T> | undefined,
   mapperState: DimensionMapping
-): ndarray<T> {
+): ndarray<T> | undefined {
   return useMemo(() => {
-    if (mapperState === undefined) {
+    if (!baseArray || mapperState === undefined) {
```
axelboc, line 23: `|| !mapperState`

src/h5web/visualizations/shared/hooks.ts
axelboc: Right, that will do for now. Might be good to add a comment, though, since it's not very self-documenting. I'm tempted to ask a question on StackOverflow -- it's such a simple problem that I feel there must be a solution.

src/h5web/visualizations/shared/utils.ts
axelboc: Haha nice, not sure why I didn't think of that. 🤦‍♂️

src/h5web/visualizations/shared/utils.ts
```diff
@@ -45,37 +45,52 @@ export function computeVisSize(
...
-        val > 0 && (accPositiveMin === undefined || accPositiveMin > val)
-          ? val
-          : accPositiveMin,
-      ];
-    },
-    [values[0], values[0], undefined]
-  );
+  if (errors && errors.length !== values.length) {
+    throw new Error(
+      'Errors when computing domain should have the same size as values'
```
axelboc, line 73: I would word this `Expected 'errors' array to have same length as 'values' array` for consistency.

`when computing domain` isn't very useful to the user, and for us, we can easily find the error in the code (or in the stack trace).

src/h5web/visualizations/shared/utils.ts
axelboc: Hmm, took me a minute to understand, but I think it's quite elegant. Another solution would be for `getExtent` to only be the inner function of the `reduce`, and to call it once for the value, and two more times for the value with errors:

```
values.reduce((acc, val, i) => {
  const res = getExtent(acc, value);
  return !errors ? res : getExtent(getExtent(res, value - err), value + err);
});
```

Not as elegant, but perhaps simpler.

axelboc (approved):

src/h5web/visualizations/shared/hooks.ts
```diff
@@ -9,18 +9,18 @@ import { getCanvasScale, getDomain } from './utils';
 import AxisSystemContext from './AxisSystemContext';
 import { AxisScale } from './models';
 
-export function useBaseArray<T>(value: T[], rawDims: number[]): ndarray<T> {
+export function useBaseArray<T>(value: T[] | undefined, rawDims: number[]) {
```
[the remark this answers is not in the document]
loichuder: It was lost in the wake of my TS experiments :smile:

src/h5web/visualizations/line/MappedLineVis.tsx
```diff
@@ -44,29 +49,24 @@ function MappedLineVis(props: Props): ReactElement {
...
-      value - errors[i],
-      value,
-      value + errors[i],
-    ]);
-  }, [autoScale, baseArray.data, dataArray.data, errors]);
-
-  const dataDomain = useDomain(dataWithErrors, yScaleType);
+  const errorValues = autoScale
+    ? errorArray && (errorArray.data as number[])
+    : baseErrors && (baseErrors.data as number[]);
```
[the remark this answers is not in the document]
loichuder: > I wonder whether errorValues should come before yScaleType in the signature of useDomain.

But then, we would have to spectify `errors` in the `Heatmap`'s call of `useDomain`:
```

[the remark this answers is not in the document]
axelboc: arf, right 👍

axelboc (approved):
Nice and clean with the new type `Bounds` 👍

Issue #318, opened 2020-11-19
[NeXus] `DATASETNAME_errors` support


axelboc:
The `errors` dataset can also be named `DATASETNAME_errors` - https://github.com/silx-kit/h5web/issues/316#issuecomment-729819269

loichuder:
Fixed by #337

PR #332, opened 2020-11-24
Simplify signature of `useDatasetValues` hook

axelboc:
I'm going step by step to keep things clear. I'll refactor `findSignalDataset` next.

### DIFF
--- a/src/h5web/providers/Provider.tsx
+++ b/src/h5web/providers/Provider.tsx
@@ -32,14 +32,9 @@ function Provider(props: Props): ReactElement {
         domain: api.domain,
         metadata,
         getValue,
-        getValues: async (idsRecord) =>
+        getValues: async (ids) =>
           Object.fromEntries(
-            await Promise.all(
-              Object.entries(idsRecord).map(async ([name, id]) => {
-                const val = await getValue(id);
-                return [name, val];
-              })
-            )
+            await Promise.all(ids.map(async (id) => [id, await getValue(id)]))
           ),
       }}
     >
--- a/src/h5web/providers/utils.ts
+++ b/src/h5web/providers/utils.ts
@@ -112,6 +112,10 @@ export function assertSimpleShape<T extends HDF5Type>(
   }
 }
 
+export function getLink(name: string, group: HDF5Group): HDF5Link | undefined {
+  return group.links?.find((l) => l.title === name);
+}
+
 export function getEntity(
   link: HDF5Link | undefined,
   metadata: HDF5Metadata
@@ -130,6 +134,6 @@ export function getLinkedEntity(
   group: HDF5Group,
   metadata: HDF5Metadata
 ): HDF5Entity | undefined {
-  const childLink = group.links?.find((l) => l.title === entityName);
-  return getEntity(childLink, metadata);
+  const link = getLink(entityName, group);
+  return getEntity(link, metadata);
 }
--- a/src/h5web/visualizations/containers/hooks.ts
+++ b/src/h5web/visualizations/containers/hooks.ts
@@ -10,10 +10,10 @@ export function useDatasetValue(id: HDF5Id): HDF5Value | undefined {
 }
 
 export function useDatasetValues(
-  datasetsRecord: Record<string, HDF5Id>
-): Record<string, HDF5Value> | undefined {
+  ids: HDF5Id[]
+): Record<HDF5Id, HDF5Value | undefined> {
   const { getValues } = useContext(ProviderContext);
-  const [valuesReader] = useAsyncResource(getValues, datasetsRecord);
+  const [valuesReader] = useAsyncResource(getValues, ids);
 
-  return valuesReader();
+  return valuesReader() || {};
 }
--- a/src/h5web/visualizations/nexus/hooks.ts
+++ b/src/h5web/visualizations/nexus/hooks.ts
@@ -3,16 +3,20 @@ import {
   HDF5Group,
   HDF5Id,
   HDF5Metadata,
+  HDF5Value,
 } from '../../providers/models';
 import {
   assertDataset,
-  getEntity,
+  getLink,
   getLinkedEntity,
-  isDataset,
   isReachable,
 } from '../../providers/utils';
 import { useDatasetValues } from '../containers/hooks';
-import { assertArray, assertOptionalArray } from '../shared/utils';
+import {
+  assertArray,
+  assertDefined,
+  assertOptionalArray,
+} from '../shared/utils';
 import { NxData } from './models';
 import {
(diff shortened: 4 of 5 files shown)


src/h5web/providers/Provider.tsx
```diff
@@ -32,14 +32,9 @@ function Provider(props: Props): ReactElement {
...
-        getValues: async (idsRecord) =>
+        getValues: async (ids) =>
           Object.fromEntries(
-            await Promise.all(
-              Object.entries(idsRecord).map(async ([name, id]) => {
-                const val = await getValue(id);
-                return [name, val];
-              })
-            )
+            await Promise.all(ids.map(async (id) => [id, await getValue(id)]))
```
axelboc, line 37: We now expect an array of IDs and return a `Record<HDF5Id, HDF5Value>` (cf. the context for the typing).

src/h5web/providers/utils.ts
```diff
@@ -112,6 +112,10 @@ export function assertSimpleShape<T extends HDF5Type>(
   }
 }
 
+export function getLink(name: string, group: HDF5Group): HDF5Link | undefined {
+  return group.links?.find((l) => l.title === name);
+}
```
axelboc, line 117: I need this to get to the ID of the `"errors"` dataset. We don't need to retrieve the whole dataset to get the ID. I may remove this in my next PR, though, since I may end up retrieving the errors dataset to assert on its shape/type.

src/h5web/visualizations/containers/hooks.ts
```diff
@@ -10,10 +10,10 @@ export function useDatasetValue(id: HDF5Id): HDF5Value | undefined {
...
-  datasetsRecord: Record<string, HDF5Id>
-): Record<string, HDF5Value> | undefined {
+  ids: HDF5Id[]
+): Record<HDF5Id, HDF5Value | undefined> {
   const { getValues } = useContext(ProviderContext);
-  const [valuesReader] = useAsyncResource(getValues, datasetsRecord);
+  const [valuesReader] = useAsyncResource(getValues, ids);
 
-  return valuesReader();
+  return valuesReader() || {};
```
axelboc, line 18: I always return an object to avoid having to do `datasetValues && datasetValues[id]`.

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -23,32 +27,23 @@ import {
...
-
-      return entity && isDataset(entity) ? [link.title, entity.id] : undefined;
-    })
-    .filter((entry): entry is [string, HDF5Id] => !!entry);
+  group: HDF5Group
+): Record<HDF5Id, HDF5Value | undefined> {
+  const datasetIds = group.links
+    ?.filter(isReachable)
+    .filter((link) => link.collection === HDF5Collection.Datasets)
+    .map((link) => link.id);
```
axelboc, line 35: It simplifies things considerably.

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -59,29 +54,29 @@ export function useNxData(group: HDF5Group, metadata: HDF5Metadata): NxData {
...
       return undefined;
     }
 
-    const dataset = getLinkedEntity(name, group, metadata);
-    if (!dataset) {
-      return undefined;
-    }
-    assertDataset(dataset);
+    const axisDataset = getLinkedEntity(axisName, group, metadata);
+    assertDefined(axisDataset);
```
axelboc, line 63: Now enforcing the presence of the axis dataset if it's listed in the `axes` attribute.

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -59,29 +54,29 @@ export function useNxData(group: HDF5Group, metadata: HDF5Metadata): NxData {
...
       scaleType: axes_scale_type && axes_scale_type[i],
     };
   });
 
-  const errors = datasetValues && datasetValues.errors;
-  assertOptionalArray<number>(errors);
+  const errorsLink = getLink('errors', group);
+  const errorsValue =
+    errorsLink && isReachable(errorsLink) && values[errorsLink.id];
+  assertOptionalArray<number>(errorsValue);
```
axelboc, line 79: Here is where I have to use `getLink`. This code will likely change again, as mentioned (and also because of #318).

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -90,11 +85,8 @@ export function useNxData(group: HDF5Group, metadata: HDF5Metadata): NxData {
       dims: signalDataset.shape.dims,
       scaleType: signal_scale_type,
     },
-    errors,
-    title:
-      datasetValues && typeof datasetValues.title === 'string'
-        ? datasetValues.title
-        : undefined,
+    errors: errorsValue,
+    title: typeof values.title === 'string' ? values.title : undefined,
```
loichuder, line 89: I don't see how this is still working as the keys from `values` are now `HDF5Id`s.

Shouldn't you use the same trick as for `errors`: get the `title` link to get its `id` ?

src/h5web/providers/utils.ts
loichuder: You may need it for `title` as well

loichuder (changes requested):

src/h5web/visualizations/nexus/hooks.ts
loichuder: I see now why it isn't caught by the test on mock data:
- We use the name of the dataset as id in mock data
- The name of the title dataset is ... `title` sso that the id stays `title` :rofl:

src/h5web/test-utils.tsx
```diff
@@ -45,7 +45,10 @@ export async function selectVisTab(vis: Vis): Promise<void> {
  * Call returned `resetConsole` function to restore original method.
  */
 /* eslint-disable no-console */
-export function prepareForConsoleError() {
+export function prepareForConsoleError(): {
+  consoleErrorMock: jest.Mock;
+  resetConsole: () => void;
+} {
```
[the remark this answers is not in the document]
axelboc: I wanted to enforce return types on functions like in our previous ESLint config, but it's just too annoying sometimes.

src/h5web/providers/utils.ts
axelboc: 🎣 I made the same temporary fix.

loichuder (approved):
Perhaps you could also change the `id` of the `title` dataset so that the error is caught in the future ?

axelboc:
I double checked and it actually is different already.

image

I remember that one of the E2E tests had correctly caught the regression.

PR #335, opened 2020-11-25
Update to Cypress 6 and also other minor updates

### LINKED ISSUE #331: Update to Cypress 6.0.0
Relevant links:
- https://docs.cypress.io/guides/references/changelog.html#6-0-0
- https://docs.cypress.io/guides/references/migration-guide.html#Migrating-to-Cypress-6-0

The update should be smooth at first glance

loichuder:
Fix #331 

I did not update Typescript to 4.1 as I get the following error:

```
.../h5web/node_modules/react-scripts/scripts/utils/verifyTypeScriptSetup.js:239
      appTsConfig.compilerOptions[option] = value;
```

It should probably be updated altogether with the update of `react-scripts` to v17


axelboc (commented):
`react-slider` has a new `marks` prop to show ticks: https://github.com/zillow/react-slider/blob/master/CHANGELOG.md#110-2020-11-03 -- should we give it a try?

axelboc:
How is it looking for an upgrade to React 17? Still too many dependency warnings?

loichuder:
> How is it looking for an upgrade to React 17? Still too many dependency warnings?

Yeah it is a bit all over the place. Also, we would need to change the `@vx` packages to `@visx`. I prefer to keep that for later.

> `react-slider` has a new `marks` prop to show ticks: https://github.com/zillow/react-slider/blob/master/CHANGELOG.md#110-2020-11-03 -- should we give it a try?

We can, it looks interesting !

axelboc (approved):

PR #342, opened 2020-11-25
Move mapper state to mapped vis components

axelboc:


src/h5web/visualizations/heatmap/MappedHeatmapVis.tsx
```diff
@@ -31,6 +32,12 @@ function MappedHeatmapVis(props: Props): ReactElement {
     disableAutoScale,
   } = useHeatmapConfig();
 
+  const [dimensionMapping, setDimensionMapping] = useState<DimensionMapping>([
+    ...range(dims.length - 2).fill(0),
```
loichuder, line 36: Just realized that we can replace this by `...Array(dims.length - 2).fill(0)` to avoid an external call to `lodash`

loichuder (approved):

src/h5web/visualizations/heatmap/MappedHeatmapVis.tsx
axelboc: Toutafé

PR #343, opened 2020-11-26
Replace `useDataArray` hook with overrides and more

axelboc:
Four commits:

- **Pass errors as `ndarray` to Line vis** instead of as `number[]` for consistency with data array and with errors in Heatmap vis.
- **Make `DimensionMapping` type non-nullable**. Ever since the big refactoring of `Visualizer`, we only use the `DimensionMapper` on visualizations that need it, so we never have a `dimensionMapper` state of `undefined`.
- **Move Matrix vis mapper state from container to mapped vis**. I forgot about the Matrix vis in #342...
- **Replace `useDataArray` hook with overrides**. The overrides add a bit of bloat, but as discussed, they're a little bit cleaner than alternative solutions.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,5 +11,6 @@
   },
 
   "typescript.tsdk": "node_modules/typescript/lib",
-  "typescript.enablePromptUseWorkspaceTsdk": true
+  "typescript.enablePromptUseWorkspaceTsdk": true,
+  "typescript.preferences.importModuleSpecifier": "relative"
 }
--- a/src/h5web/visualizations/shared/hooks.ts
+++ b/src/h5web/visualizations/shared/hooks.ts
@@ -9,6 +9,12 @@ import { getCanvasScale, getDomain } from './utils';
 import AxisSystemContext from './AxisSystemContext';
 import { AxisScale } from './models';
 
+export function useBaseArray<T>(value: T[], rawDims: number[]): ndarray<T>;
+export function useBaseArray<T>(
+  value: T[] | undefined,
+  rawDims: number[]
+): ndarray<T> | undefined;
+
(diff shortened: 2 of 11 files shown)


.vscode/settings.json
```diff
@@ -11,5 +11,6 @@
   },
 
   "typescript.tsdk": "node_modules/typescript/lib",
-  "typescript.enablePromptUseWorkspaceTsdk": true
+  "typescript.enablePromptUseWorkspaceTsdk": true,
+  "typescript.preferences.importModuleSpecifier": "relative"
```
axelboc, line 15: For some reason, VSCode started to generate absolute imports... So I'm forcing TypeScript to generate relative imports instead.

src/h5web/visualizations/shared/hooks.ts
```diff
@@ -18,20 +24,28 @@ export function useBaseArray<T>(
...
 export function useMappedArray<T>(
   baseArray: ndarray<T> | undefined,
-  mapperState: DimensionMapping
+  mapping: DimensionMapping
 ): ndarray<T> | undefined {
   return useMemo(() => {
-    if (!baseArray || !mapperState) {
+    if (!baseArray || !mapping) {
       return baseArray;
     }
```
loichuder, line 43:
```js
    if (!baseArray) {
      return undefined;
    }
```
`mapping` cannot be `undefined` anymore, right ?

loichuder (approved):
Mice :mouse: !

src/h5web/visualizations/shared/hooks.ts
axelboc: Indeed, because `DimensionMapping` no longer includes `undefined`.

PR #345, opened 2020-11-30
Update @vx to @visx

### LINKED ISSUE #340: Update @vx packages to @visx

loichuder:
Fix #340


loichuder:
No breaking changes when upgrading `grid`, `tooltip` and `shape` to their `@vis` counterpart :tada: 

Upgrading `axis` gives me TS errors that I am digging into. This was already attempted by the way (https://github.com/silx-kit/h5web/commit/78c62e80caf66bcd946766fb42b745cf72d9d359).

loichuder:
I fixed the TS errors by using `@visx/scale` to generate the scales. This allows to have a consistent typing of scales throughout the `@visx` components and our app.

We could remove `d3-scale` and `@types/d3-scale` except for `scaleSequential`. It is the only component left that is not exposed by `@visx/scale`. I have opened an issue on `@visx` repo: https://github.com/airbnb/visx/issues/953.

axelboc (approved):
Right, I get it, so because the overrides of `createScale` only accept a single type of scale, TS can't fit our union of three scales in any of them ; and since it can't combine the types of multiple overrides, it falls back to the generic override that accepts all the scale types.

I fell into another TypeScript rabbit hole and found a way to "augment" the type of `@visx/scale` with an override. However, I don't like the result, as the module declaration has to live in a `ts` file and the `createAxis` function has to be re-exported:

image

So let's keep your solution, which adds runtime code, but is much more readable.

PR #347, opened 2020-12-02
Introduce new nested HDF5 models to replace Explorer's `TreeNode`

axelboc:
**Part 1 of The Big Refactoring** ™️  that aims to remove the complexity of having to follow links to find entities by providing a nested HDF5 structure to the entire app.

Since the Explorer already uses a nested structure (`TreeNode`), I thought it would be a good place to start. So this first refactoring consists of replacing the generic `TreeNode<T>` model with a specific, nested HDF5 structure, and to refactor the `buildTree` function to return an object with this new structure.

To keep things manageable and limit the scope of the refactoring, I decided:

- to keep all the existing HDF5 models for now;
- to temporarily prefix the new models with `My`;
- to keep references to the raw links and entities in the new HDF5 structure, so I can pass them to `Visualizer` and `MetadataViewer`;
- to keep injecting a `metadata` object of type `HDF5Metadata` through the `ProviderContext`.

### DIFF
--- a/src/h5web/explorer/Explorer.tsx
+++ b/src/h5web/explorer/Explorer.tsx
@@ -6,60 +6,55 @@ import React, {
   ReactElement,
 } from 'react';
 import { FiFileText } from 'react-icons/fi';
-import type { HDF5Link } from '../providers/models';
-import type { TreeNode, ExpandedNodes } from './models';
-import TreeView from './TreeView';
+import type { MyHDF5Entity } from '../providers/models';
+import TreeView, { ExpandedGroups } from './TreeView';
 import styles from './Explorer.module.css';
-import Icon from './Icon';
-import { buildTree, getNodesOnPath } from './utils';
+import { buildTree, getEntityAtPath } from './utils';
 import { ProviderContext } from '../providers/context';
+import { isMyGroup } from '../providers/utils';
 
 const DEFAULT_PATH: number[] = JSON.parse(
   process.env.REACT_APP_DEFAULT_PATH || '[]'
 );
 
 interface Props {
-  onSelect: (node: TreeNode<HDF5Link>) => void;
-  selectedNode?: TreeNode<HDF5Link>;
+  onSelect: (entity: MyHDF5Entity) => void;
+  selectedEntity?: MyHDF5Entity;
 }
 
 function Explorer(props: Props): ReactElement {
-  const { onSelect, selectedNode } = props;
+  const { onSelect, selectedEntity } = props;
 
   const { domain, metadata } = useContext(ProviderContext);
-  const tree = useMemo(() => buildTree(metadata, domain), [domain, metadata]);
+  const root = useMemo(() => buildTree(metadata, domain), [domain, metadata]);
 
-  const [expandedNodes, setExpandedNodes] = useState<ExpandedNodes>({});
+  const [expandedGroups, setExpandedGroups] = useState<ExpandedGroups>({});
 
-  function handleNodeSelect(node: TreeNode<HDF5Link>): void {
-    const isBranch = !!node.children;
-    const isExpanded = !!expandedNodes[node.uid];
-    const isSelected = node === selectedNode;
+  function handleSelect(entity: MyHDF5Entity): void {
+    const isExpanded = !!expandedGroups[entity.uid];
+    const isSelected = entity === selectedEntity;
+    onSelect(entity);
 
-    onSelect(node);
-
-    if (isBranch && (!isExpanded || isSelected)) {
-      setExpandedNodes({
-        ...expandedNodes,
-        [node.uid]: !isExpanded,
+    if (isMyGroup(entity) && (!isExpanded || isSelected)) {
+      setExpandedGroups({
+        ...expandedGroups,
+        [entity.uid]: !isExpanded,
       });
     }
   }
 
   useEffect(() => {
-    // Get nodes on default path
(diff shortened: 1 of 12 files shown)


src/h5web/providers/models.ts
```diff
@@ -48,6 +48,47 @@ export interface HDF5Datatype<T = HDF5Type> {
...
+export interface MyHDF5Entity {
+  uid: string;
+  id?: HDF5Id;
+  name: string;
+  kind: MyHDF5EntityKind;
+  parents: MyHDF5Group[];
+  attributes: HDF5Attribute[];
+  rawLink: HDF5Link;
+  rawEntity?: HDF5Entity;
+}
```
axelboc, line 67: The new models (with temporary names). The old models may eventually be replaced or moved to `SilxProvider`.

Note that apart from being nested, a big change is that `MyHDF5Entity` is no longer a union but a base class that other interfaces extend. I read this tip somewhere recently on an article about optimising the performance of TypeScript: with this object-oriented pattern, when we use the base type `MyHDF5Entity`, the engine knows right away which keys are allowed; it doesn't have to look for common keys across all the members of the union. I'll share the article on #webdev. I think this pattern is also more common in the wild, and it feels more descriptive than a simple union (the common keys are clearly identified).

src/h5web/providers/models.ts
```diff
@@ -48,6 +48,47 @@ export interface HDF5Datatype<T = HDF5Type> {
...
+
+export interface MyHDF5Datatype<T = HDF5Type> extends MyHDF5Entity {
+  kind: MyHDF5EntityKind.Datatype;
+  type: T;
+}
+
+export interface MyHDF5Link extends MyHDF5Entity {
+  kind: MyHDF5EntityKind.Link;
+}
```
axelboc, line 90: This type is not used yet, but we'll most likely end up removing `rawLink` from` MyHDF5Entity` and exposing some properties of the link here.

src/h5web/explorer/Explorer.tsx
```diff
@@ -6,104 +6,78 @@ import React, {
...
+  selectedEntity?: MyHDF5Entity;
 }
 
 function Explorer(props: Props): ReactElement {
-  const { onSelect, selectedNode } = props;
+  const { onSelect, selectedEntity } = props;
 
   const { domain, metadata } = useContext(ProviderContext);
-  const tree = useMemo(() => buildTree(metadata, domain), [domain, metadata]);
+  const root = useMemo(() => buildTree(metadata, domain), [domain, metadata]);
```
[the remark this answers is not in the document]
loichuder: `buildTree` should be renamed in a future refacto

src/h5web/explorer/Explorer.tsx
```diff
@@ -6,104 +6,78 @@ import React, {
...
-      </div>
+    const entityToSelect = getEntityAtPath(root, DEFAULT_PATH);
+    onSelect(entityToSelect);
+
+    // Expand groups on default path and selected entity if group
+    setExpandedGroups(
+      entityToSelect.parents.reduce(
+        (acc, node) => ({ ...acc, [node.uid]: true }),
+        isMyGroup(entityToSelect) ? { [entityToSelect.uid]: true } : {}
+      )
```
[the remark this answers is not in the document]
loichuder: Can't the argument of `setExpandedGroups` be a simple array of `uid` ?

src/h5web/explorer/Explorer.tsx
```diff
@@ -6,104 +6,78 @@ import React, {
...
-          expandedNodes={expandedNodes}
-          onSelect={handleNodeSelect}
-          renderIcon={(data, isBranch, isExpanded) => (
-            <Icon data={data} isBranch={isBranch} isExpanded={isExpanded} />
-          )}
-        />
-      )}
+      <TreeView
+        level={0}
+        entities={root.children}
```
loichuder, line 76: I don't have any replacement but I find the name `entities` not extremely clear...

src/h5web/explorer/utils.ts
```diff
@@ -1,57 +1,126 @@
...
 
+function prepareEntity(
+  link: HDF5HardLink | HDF5RootLink,
+  parents: MyHDF5Group[]
+): Omit<MyHDF5Entity, 'kind'> {
   return {
     uid: nanoid(),
-    label: link.title,
-    data: link,
+    id: link.id,
```
loichuder, line 22: Would we be able to merge `id` and `uid` later ?

loichuder (approved):
Great work !!

There are a few things to iron out but this can be done later once everything is settled. I would advise you to merge this as it is, keep the comments for later and continue with the refactoring.

src/h5web/explorer/Explorer.tsx
axelboc: Maybe it would make more sense if the component was called `EntityList`?

loichuder: Probably yes :slightly_smiling_face:

src/h5web/explorer/utils.ts
axelboc: Well, it's the usual conundrum of whether the same entity can be referenced from multiple links, but yes, we can definitely expose a single ID in the structure and let the providers either:

- use the IDs returned by the server if they are guaranteed to be unique; or
- generate unique IDs dynamically.

If a provider ends up generating unique IDs, it will have to keep a mapping of the generated IDs to the raw IDs for the fetching of dataset values to keep working.

PR #349, opened 2020-12-02
Domain is computed without errors if they are disabled in MappedLineVis

### LINKED ISSUE #341: [NeXus] Toggling errors does not change the domain
The domain should be updated by not taking error bars into account when toggling off

loichuder:
Fix #341


axelboc (approved):
🎉

PR #351, opened 2020-12-03
Simplify TS overrides of `useBaseArray` and `useMappedArray`

axelboc:
See https://stackoverflow.com/questions/65020025/reduce-duplication-when-overloading-functions-in-typescript


loichuder (approved):
It also strengthens the typing as previously we had:
`res = useBaseArray(undefined)` > `res is number[] | undefined`
and now
`res = useBaseArray(undefined)` > `res is undefined`

axelboc:
Indeed! 🎉

PR #361, opened 2020-12-07
Clean up obsolete utilities

axelboc:


src/h5web/metadata-viewer/MetadataViewer.tsx
```diff
@@ -21,8 +21,8 @@ function MetadataViewer(props: Props): ReactElement {
 
   return (
     <div className={styles.metadataViewer}>
-      <EntityInfo entity={entity} />
-      <AttributesInfo attributes={entity.attributes} />
+      <EntityTable entity={entity} />
+      <AttributesTable attributes={entity.attributes} />
```
axelboc, line 25: Just renaming these two components to distinguish them from the `LinkInfo` components.

src/h5web/providers/models.ts
```diff
@@ -57,32 +57,32 @@ export enum MyHDF5EntityKind {
...
   kind: MyHDF5EntityKind;
   parents: MyHDF5Group[];
   attributes: HDF5Attribute[];
   rawLink?: HDF5Link;
 }
 
-export interface MyHDF5Group extends MyHDF5Entity {
+export interface MyHDF5ResolvedEntity extends MyHDF5Entity {
   id: HDF5Id;
+}
```
axelboc, line 69: Introducing the concept of a "resolved" entity (as opposed to unresolved links). This allows me to add a guard called `isResolved`, useful when I need to access an entity's `id`.

loichuder (changes requested):

src/h5web/providers/utils.ts
```diff
@@ -11,17 +9,16 @@ import {
...
   HDF5ScalarShape,
   HDF5NumericType,
-  HDF5Group,
   MyHDF5Entity,
   MyHDF5EntityKind,
   MyHDF5Group,
   MyHDF5Datatype,
   MyHDF5Dataset,
   MyHDF5Link,
+  MyHDF5ResolvedEntity,
```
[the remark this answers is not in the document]
axelboc: I'll wait a bit longer on this one, until I refactor the providers, just in case I need to adjust a few things.

src/h5web/visualizations/index.tsx
```diff
@@ -86,8 +84,8 @@ export const VIS_DEFS: Record<Vis, VisDef> = {
     Container: LineVisContainer,
     supportsEntity: (entity) => {
       return (
-        isMyDataset(entity) &&
-        hasMyNumericType(entity) &&
+        isDataset(entity) &&
+        hasNumericType(entity) &&
         hasMySimpleShape(entity) &&
```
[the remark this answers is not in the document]
axelboc: It's only used exactly like this twice -- the Matrix vis calls `hasBaseType` instead of `hasNumericType`, so not sure it's worth it. Unless you feel strongly about it? Wouldn't know what to call it anyway 😂

src/h5web/providers/models.ts
axelboc: Should I keep the changes as is for now or would you prefer I revert?

src/h5web/visualizations/index.tsx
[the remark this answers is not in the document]
loichuder: Ah yes, these are assertions in the containers... Nevermind

src/h5web/providers/models.ts
loichuder: You can merge as it is. I will make a thorough review of all of this when I have time.

Issue #358, opened 2020-12-07
Release workflow


axelboc:
Set up a semi-automated release process for the packages. Requirements:

- Publish packages in CI
- Keep storybook in sync with latest package versions.
- Production Storybook should use published packages.
- Keep deploying demo from `master`.

Suggested set-up:

- Configure a webpack alias to import the published package `@h5web/lib` from Storybook when building for production (instead of importing `h5web` components with relative imports). See `react-three-fiber` for a similar set-up.
- Add CI workflow triggered on push of a tag, that:
    1. Reads the version number from the tag.
    1. Bumps the packages versions, commits the changes to `master` and update the tag if possible?
    1. Publishes the packages to NPM.
    1. Builds and deploys the Storybook for production.

axelboc:
After much thinking, here is a workflow I think I can set up:

- Version number is managed in a single place: in the root `package.json` of the project. The `package.json` files in the `packages` folder have a placeholder version field set to `0.0.1`.
- To release a new version:
    1. Check out `master`.
    2. Make sure the working tree is clean (no uncommitted changes, no commits to be pushed).
    3. Make sure the latest commit on `master` has passed on the CI.
    4. Run `npm version [ patch | minor | major]`.
- `npm version` automatically bumps the version number in `package.json` and `package-lock.json`, commits the change, creates a tag, and runs the `postversion` script, which pushes the new commit and the new tag.
- In the CI, a job is triggered by the push of the new tag. This job:
    1. Installs the dependencies in the `packages` folder.
    2. Builds the packages.
    3. Reads the new version from the tag name.
    4. Sets the new version in the `package.json` located in the `dist` folder of the packages.
    5. Publishes the packages to NPM.
    6. Runs the `Deploy Storybook` workflow to build and deploy the Storybook site in a way that it uses the latest version of the published packages.
- To finalize the release:
    - Make sure the release and storybook deployment workflows on the CI have passed.
    - Make sure the new package versions are available on NPM and that the live Storybook site still works as expected.
    - Upgrade and test the packages in apps and code sandboxes, as required.
    - Write and publish the release notes on GitHub.

axelboc:
The "Slach command" workflow seems to be failing randomly. Testing...

axelboc:
- [x] Implement release workflow to deploy packages to NPM when pushing a versioned tag.
- [ ] Build and deploy Storybook site for production with latest NPM packages once they're published.
- [ ] Document process in README/CONTRIBUTING.

PR #365, opened 2020-12-08
Generate nested mock metadata

axelboc:
I use the new metadata in `MockProvider` with a temporary change to the types of the providers' context. This will allow us to migrate the providers one at a time.

### DIFF
--- a/src/h5web/explorer/Explorer.test.tsx
+++ b/src/h5web/explorer/Explorer.test.tsx
@@ -1,15 +1,15 @@
 import { fireEvent, screen } from '@testing-library/react';
-import { MOCK_DOMAIN } from '../providers/mock/MockProvider';
+import { mockDomain } from '../providers/mock/data';
 import { renderApp } from '../test-utils';
 
 describe('Explorer', () => {
   test('select root group by default', async () => {
     renderApp();
 
-    const title = await screen.findByRole('heading', { name: MOCK_DOMAIN });
+    const title = await screen.findByRole('heading', { name: mockDomain });
     expect(title).toBeVisible();
 
-    const domainBtn = screen.getByRole('treeitem', { name: MOCK_DOMAIN });
+    const domainBtn = screen.getByRole('treeitem', { name: mockDomain });
     expect(domainBtn).toBeVisible();
     expect(domainBtn).toHaveAttribute('aria-selected', 'true');
   });
@@ -18,7 +18,7 @@ describe('Explorer', () => {
     renderApp();
 
     const domainBtn = await screen.findByRole('treeitem', {
-      name: MOCK_DOMAIN,
+      name: mockDomain,
     });
     const sidebarBtn = screen.getByRole('button', {
       name: 'Toggle explorer sidebar',
--- a/src/h5web/explorer/Explorer.tsx
+++ b/src/h5web/explorer/Explorer.tsx
@@ -5,7 +5,7 @@ import EntityList from './EntityList';
 import styles from './Explorer.module.css';
 import { buildTree, getEntityAtPath, getParents } from './utils';
 import { ProviderContext } from '../providers/context';
-import { isGroup } from '../providers/utils';
+import { isGroup, isMetadataMyGroup } from '../providers/utils';
 import { useSet } from 'react-use';
 
 const DEFAULT_PATH: number[] = JSON.parse(
@@ -19,9 +19,11 @@ interface Props {
 
 function Explorer(props: Props): ReactElement {
   const { onSelect, selectedEntity } = props;
-
   const { domain, metadata } = useContext(ProviderContext);
-  const root = useMemo(() => buildTree(metadata, domain), [domain, metadata]);
+
+  const root = useMemo(() => {
+    return isMetadataMyGroup(metadata) ? metadata : buildTree(metadata, domain);
+  }, [domain, metadata]);
 
   const [
     expandedGroups,
--- a/src/h5web/explorer/Explorer.tsx
+++ b/src/h5web/explorer/Explorer.tsx
@@ -5,7 +5,7 @@ import EntityList from './EntityList';
 import styles from './Explorer.module.css';
 import { buildTree, getEntityAtPath, getParents } from './utils';
 import { ProviderContext } from '../providers/context';
-import { isGroup, isMetadataMyGroup } from '../providers/utils';
+import { isGroup, isMyMetadata } from '../providers/utils';
 import { useSet } from 'react-use';
 
 const DEFAULT_PATH: number[] = JSON.parse(
@@ -22,7 +22,7 @@ function Explorer(props: Props): ReactElement {
   const { domain, metadata } = useContext(ProviderContext);
 
   const root = useMemo(() => {
-    return isMetadataMyGroup(metadata) ? metadata : buildTree(metadata, domain);
+    return isMyMetadata(metadata) ? metadata : buildTree(metadata, domain);
   }, [domain, metadata]);
 
   const [
--- a/src/h5web/providers/context.ts
+++ b/src/h5web/providers/context.ts
@@ -1,15 +1,15 @@
 import { createContext } from 'react';
-import type { HDF5Id, HDF5Value, HDF5Metadata } from './models';
+import type { HDF5Id, HDF5Value, HDF5Metadata, MyHDF5Group } from './models';
 
 export abstract class ProviderAPI {
   abstract domain: string;
-  abstract getMetadata(): Promise<HDF5Metadata>;
+  abstract getMetadata(): Promise<HDF5Metadata | MyHDF5Group>;
   abstract getValue(id: HDF5Id): Promise<HDF5Value | undefined>;
 }
 
 export const ProviderContext = createContext<{
   domain: string;
-  metadata: HDF5Metadata;
+  metadata: HDF5Metadata | MyHDF5Group;
   getValue: ProviderAPI['getValue'];
   getValues: (ids: HDF5Id[]) => Promise<Record<HDF5Id, HDF5Value | undefined>>;
 }>({} as any); // eslint-disable-line @typescript-eslint/no-explicit-any
--- a/src/h5web/providers/context.ts
+++ b/src/h5web/providers/context.ts
@@ -1,15 +1,15 @@
 import { createContext } from 'react';
-import type { HDF5Id, HDF5Value, HDF5Metadata, MyHDF5Group } from './models';
+import type { HDF5Id, HDF5Value, HDF5Metadata, MyHDF5Metadata } from './models';
 
 export abstract class ProviderAPI {
   abstract domain: string;
-  abstract getMetadata(): Promise<HDF5Metadata | MyHDF5Group>;
+  abstract getMetadata(): Promise<HDF5Metadata | MyHDF5Metadata>;
   abstract getValue(id: HDF5Id): Promise<HDF5Value | undefined>;
 }
 
 export const ProviderContext = createContext<{
   domain: string;
-  metadata: HDF5Metadata | MyHDF5Group;
+  metadata: HDF5Metadata | MyHDF5Metadata;
   getValue: ProviderAPI['getValue'];
   getValues: (ids: HDF5Id[]) => Promise<Record<HDF5Id, HDF5Value | undefined>>;
 }>({} as any); // eslint-disable-line @typescript-eslint/no-explicit-any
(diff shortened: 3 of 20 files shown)


src/h5web/explorer/Explorer.test.tsx
```diff
@@ -1,15 +1,15 @@
 import { fireEvent, screen } from '@testing-library/react';
-import { MOCK_DOMAIN } from '../providers/mock/MockProvider';
+import { mockDomain } from '../providers/mock/data';
```
axelboc, line 2: I need the mock domain when generating the nested mock data (it's the `name` of the root group), so I moved it to `data.ts`.

src/h5web/explorer/Explorer.tsx
```diff
@@ -19,9 +19,11 @@ interface Props {
 
 function Explorer(props: Props): ReactElement {
   const { onSelect, selectedEntity } = props;
-
   const { domain, metadata } = useContext(ProviderContext);
-  const root = useMemo(() => buildTree(metadata, domain), [domain, metadata]);
+
+  const root = useMemo(() => {
+    return isMetadataMyGroup(metadata) ? metadata : buildTree(metadata, domain);
```
axelboc, line 25: I'm temporarily changing the type of `metadata` in the context to `HDF5Metadata | MyHDF5Group` so we can refactor the providers one by one. As a result, I need to use a temporary guard here to know

src/h5web/providers/mock/data.ts
```diff
@@ -87,7 +98,7 @@ export function makeDataset(
   return { id, collection: HDF5Collection.Datasets, type, shape, attributes };
 }
 
-function makeSimpleDataset(
+export function makeSimpleDataset(
```
axelboc, line 101: Some of the previous mock data functions are still used in the tests of `buildTree`. I thought it unwise to remove the ones that are no longer used anywhere for the time being, in case we want to write more tests or something. Either way, it also didn't want to clutter the diff, so I'm just exporting them to avoid the ESLint warning.

src/h5web/providers/mock/utils.ts
```diff
@@ -1,10 +1,34 @@
...
+  const pathSegments = path.slice(1).split('/');
+
+  const dataset = pathSegments.reduce<MyHDF5Entity | undefined>(
+    (parentEntity, currSegment) => {
+      return parentEntity && isGroup(parentEntity)
+        ? getChildEntity(parentEntity, currSegment)
+        : undefined;
+    },
+    mockMetadata
+  );
```
axelboc, line 23: This is the function we discussed, the one that is used in the stories. Here, the algorithm to find an entity from a path is taken from `useNxData`. In a future PR, I'll abstract this into a utility function, which I was thinking could even be thje existing `getEntityAtPath` (which currently accepts an array of indices).

src/h5web/providers/my-utils.ts
```diff
@@ -2,25 +2,32 @@ import { nanoid } from 'nanoid';
...
+  MyHDF5Link,
 } from '../providers/models';
 import { makeSimpleShape, makeStrAttr } from '../providers/mock/data';
-import { NxInterpretation } from '../visualizations/nexus/models';
+import { NxInterpretation, SilxStyle } from '../visualizations/nexus/models';
 
-type Opts = Partial<Pick<MyHDF5Dataset, 'id' | 'attributes'>>;
+type EntityOpts = Partial<Pick<MyHDF5Dataset, 'id' | 'attributes'>>;
+type GroupOpts = EntityOpts & { children?: MyHDF5Entity[] };
+type AllOrNone<T> = T | { [K in keyof T]?: never };
```
axelboc, line 25: I use this to type `makeMyNxDataGroup` to enforce that both or neither of `axes` and `axesAttr` be provided.

src/h5web/providers/my-utils.ts
```diff
@@ -104,46 +133,55 @@ export function withMyInterpretation<
   ]);
 }
 
-export function makeMyNxDataGroup(
```
axelboc, line 107: I'm merging the two `makeMyNxDataGroup` functions together.

src/h5web/visualizations/index.test.ts
```diff
@@ -176,35 +176,34 @@ describe('Visualization definitions', () => {
     const supportsEntity = makeSupportFn(Vis.NxSpectrum);
 
     it('should support NXdata group referencing signal with spectrum interpretation', () => {
-      const group = makeMyNxDataGroup('foo', spectrumDatasetInt1D);
+      const group = makeMyNxDataGroup('foo', { signal: spectrumDatasetInt1D });
```
axelboc, line 179: just a minor change of signature

src/h5web/visualizer/Visualizer.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { fireEvent, render, screen } from '@testing-library/react';
-import { mockValues } from '../../packages/lib';
+import { mockValues } from '../providers/mock/data';
```
axelboc, line 3: VSCode has been annoying me by auto-importing from `/packages` a lot of the time. 💢

src/stories/HeatmapVis.stories.tsx
```diff
@@ -8,14 +8,12 @@ import HeatmapVis, {
...
 import { INTERPOLATORS } from '../h5web/visualizations/heatmap/interpolators';
-import { getDomain, mockValues } from '../packages/lib';
-import { getMockDatasetDims } from '../h5web/providers/mock/utils';
-
-// Prepare 2D data array
-const values = mockValues.twoD.flat(Infinity) as number[];
-const dataArray = ndarray(values, getMockDatasetDims('twoD'));
-const domain = getDomain(values);
-const logSafeDomain = getDomain(values, ScaleType.Log);
+import { getDomain } from '../packages/lib';
```
loichuder, line 11: Should it import from `lib` here ?

src/h5web/visualizations/index.test.ts
loichuder: Looks better :slightly_smiling_face:

src/h5web/providers/context.ts
```diff
@@ -1,15 +1,15 @@
...
   abstract domain: string;
-  abstract getMetadata(): Promise<HDF5Metadata>;
+  abstract getMetadata(): Promise<HDF5Metadata | MyHDF5Group>;
   abstract getValue(id: HDF5Id): Promise<HDF5Value | undefined>;
 }
 
 export const ProviderContext = createContext<{
   domain: string;
-  metadata: HDF5Metadata;
+  metadata: HDF5Metadata | MyHDF5Group;
```
loichuder, line 12: You could add a `MyHDF5Metadata` type to track easier the places you made the changes. Your choice.

src/h5web/providers/mock/utils.ts
loichuder: I would rename `path` to `absolutePath` to make clear it should begin by `/`

src/h5web/providers/my-utils.ts
```diff
@@ -2,25 +2,32 @@ import { nanoid } from 'nanoid';
...
-type Opts = Partial<Pick<MyHDF5Dataset, 'id' | 'attributes'>>;
+type EntityOpts = Partial<Pick<MyHDF5Dataset, 'id' | 'attributes'>>;
+type GroupOpts = EntityOpts & { children?: MyHDF5Entity[] };
+type AllOrNone<T> = T | { [K in keyof T]?: never };
 
 export function makeMyGroup(
   name: string,
   children: MyHDF5Entity[] = [],
-  opts: Opts = {}
+  opts: EntityOpts = {}
```
loichuder, line 30: A bit inconsistent to have `GroupOpts` in the NxGroup creation and `children` and `EntityOps` here...

src/h5web/providers/my-utils.ts
```diff
@@ -86,6 +92,28 @@ export function makeMyDatatype<T extends HDF5Type>(
   };
 }
 
+export function makeMyExternalLink(
+  name: string,
+  file: string,
+  h5path: string,
+  opts: Omit<EntityOpts, 'id'> = {}
```
loichuder, line 99: Isn't it just `{attributes}` ?

src/h5web/providers/my-utils.ts
```diff
@@ -104,46 +132,55 @@ export function withMyInterpretation<
...
-      ...(opts.attributes || []),
-      makeStrAttr('NX_class', 'NXdata'),
-      makeStrAttr('signal', signal.name),
-    ],
-  });
-}
-
-export function makeMyNxDataGroupWithAxes(
+export function makeMyNxDataGroup<
+  U extends Record<string, MyHDF5Dataset<HDF5SimpleShape, HDF5NumericType>>
```
loichuder, line 136: I don't understand why you need a generic here

src/h5web/providers/my-utils.ts
loichuder: Could it be replaced by supplying two overrides of `makeMyNxDataGroup` (with and, without the `axes` and `axesAttr` ?)

src/h5web/providers/my-utils.ts
```diff
@@ -104,46 +132,55 @@ export function withMyInterpretation<
...
+    axesAttr,
+    ...groupOpts
+  } = opts;
+
+  return makeMyGroup(
+    name,
+    [
+      signal,
+      ...(title ? [title] : []),
+      ...(errors ? [errors] : []),
```
loichuder, line 162: I wish there was an easier way to write this...

src/h5web/providers/models.ts
```diff
@@ -87,8 +87,10 @@ export interface MyHDF5Datatype<T = HDF5Type> extends MyHDF5ResolvedEntity {
   type: T;
 }
 
-export interface MyHDF5Link extends MyHDF5Entity {
+export interface MyHDF5Link<T extends HDF5Link = HDF5Link>
+  extends MyHDF5Entity {
   kind: MyHDF5EntityKind.Link;
+  rawLink: T;
```
loichuder, line 93: Why do you need to `rawLink` ?

loichuder (changes requested):

src/stories/HeatmapVis.stories.tsx
axelboc: Meh, doesn't matter here. We'll switch it to `@h5web/lib` later anyway for #358.

src/h5web/providers/my-utils.ts
axelboc: Yeah, but the signature is super long, so not sure it's more readable...

src/h5web/providers/my-utils.ts
axelboc: Yeah, I know, but passing the `children` in the options here led to too much nesting in `data.ts`. 😢 However, I can replace `EntityOpts` here with `Omit<GroupOpts, 'children'>` 👍

src/h5web/providers/my-utils.ts
axelboc: I have a fix coming for this in my next PR. Technically, a link doesn't have attributes, so this function can be simplified.

src/h5web/providers/my-utils.ts
axelboc: Yeah, I had to put it back because of `AllOrNone`. The trick of doing `keyof typeof opts.axes`  was no longer working. One argument for using overrides...

src/h5web/providers/my-utils.ts
axelboc: Hehehe me too, I even looked it up 😄 -- either you wrap a `push` in an `if`, or you `filter` for `undefined` afterwards...

src/h5web/providers/models.ts
axelboc: For the metadata viewer. (At least for now.)

src/h5web/providers/my-utils.ts
axelboc: I've tried a few things, but nothing else worked and/or looked clearer.

axelboc: I should clarify that it works with two overrides, but then ESLint complains that they can be combined, but when I do, the typing of `axesAttr` stops working 😢

loichuder: :shrug:

loichuder (approved):
All good :+1:

PR #366, opened 2020-12-08
Uses now sprites to display error bar caps

### LINKED ISSUE #322: [NeXus] Error bar caps grow with zoom
Should they keep the same width at all times?

image

loichuder:
Fix #322 

Before:
image

Now:
image


src/h5web/visualizations/line/DataGlyphs.tsx
```diff
@@ -6,12 +6,13 @@ import { GLYPH_URLS } from './models';
 interface Props {
   geometry: BufferGeometry;
   color: string;
+  size: number;
   glyphURL?: string;
-  visible: boolean;
+  visible?: boolean;
```
loichuder, line 11: Making `visible` optional to match the signature of `<points />` below

src/h5web/visualizations/line/ErrorBarCurve.tsx
```diff
@@ -23,58 +21,65 @@ function ErrorBarCurve(props: Props): ReactElement {
...
-      ];
-    });
+      const bottom_cap = abscissaScale(val - error);
+      const top_cap = abscissaScale(val + error);
 
-    const geometry = new BufferGeometry();
-    geometry.setFromPoints(points);
-    return geometry;
-  }, [abscissaScale, abscissas, capLength, errors, ordinateScale, ordinates]);
+      const xyVector = new Vector2(x, y);
```
loichuder, line 48: I discovered that we can work with `Vector2` as three.js will automatically convert them in `Vector3` with `z=0`

src/h5web/visualizations/line/ErrorBarCurve.tsx
```diff
@@ -23,58 +21,65 @@ function ErrorBarCurve(props: Props): ReactElement {
...
+        barSegments.push(xyVector, topVector);
+      }
+    }
+
+    return [barSegments, capPoints].map((points) => {
+      const geometry = new BufferGeometry();
+      geometry.setFromPoints(points);
+      return geometry;
+    });
+  }, [abscissaScale, abscissas, errors, ordinateScale, ordinates]);
```
loichuder, line 68: I tried to improve in so many manners but cannot find a way out.

 The issue there is that I want to be able to treat `top` and `bottom` coordinates separately (to keep one even if the other is not finite) but at the same time, I need to separate `caps` and `bars` as they are used for separate geometries...

src/h5web/visualizations/line/models.ts
```diff
@@ -1,6 +1,7 @@
 import circleURL from './sprites/circle.png';
 import crossURL from './sprites/cross.png';
 import squareURL from './sprites/square.png';
+import capURL from './sprites/cap.png';
```
axelboc, line 4: Is it on purpose that the image is square with transparent zones? Could it just be rectangular without transparency? Might save a few bytes.

loichuder: Yep. Putting a simple rectangular image leads to this:
image

axelboc: Oh right, I get it 👍

axelboc (approved):

loichuder:
After discussion, we decided to merge this and I will make a heavier refactoring in another PR to review the geometry computation of `DataCurve` and `ErrorBarCurve` altogether

PR #368, opened 2020-12-09
Store only direct parents in nested entities

axelboc:
Some isolated preparatory work for #365.

When creating mock data, it feels too complex and unnecessary to recursively add a parent to its nested children. A classic tree structure in which entities reference only their direct parent is perfectly suited to our use case. It is indeed very easy to recursively find every ancestor of an entity when needed.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
 /build/
 /storybook-static/
 /packages/**/dist/
+/package-lock.json
--- a/src/h5web/explorer/utils.ts
+++ b/src/h5web/explorer/utils.ts
@@ -15,13 +15,13 @@ import { isGroup, isReachable } from '../providers/utils';
 
 function prepareEntity(
   link: HDF5HardLink | HDF5RootLink,
-  parents: MyHDF5Group[]
+  parent?: MyHDF5Group
 ): Omit<MyHDF5Group | MyHDF5Dataset | MyHDF5Datatype, 'kind'> {
   return {
     uid: nanoid(),
     id: link.id,
     name: link.title,
-    parents,
+    parent,
     attributes: [],
     rawLink: link,
   };
@@ -30,12 +30,12 @@ function prepareEntity(
 function buildDataset(
   metadata: HDF5Metadata,
   link: HDF5HardLink | HDF5RootLink,
-  parents: MyHDF5Group[]
+  parent?: MyHDF5Group
 ): MyHDF5Dataset {
   const rawDataset = metadata.datasets![link.id]; // eslint-disable-line @typescript-eslint/no-non-null-assertion
 
   return {
-    ...prepareEntity(link, parents),
+    ...prepareEntity(link, parent),
     kind: MyHDF5EntityKind.Dataset,
     attributes: rawDataset.attributes || [],
     shape: rawDataset.shape,
@@ -46,12 +46,12 @@ function buildDataset(
 function buildDatatype(
   metadata: HDF5Metadata,
   link: HDF5HardLink | HDF5RootLink,
-  parents: MyHDF5Group[]
+  parent?: MyHDF5Group
 ): MyHDF5Datatype {
   const rawDatatype = metadata.datatypes![link.id]; // eslint-disable-line @typescript-eslint/no-non-null-assertion
 
   return {
-    ...prepareEntity(link, parents),
+    ...prepareEntity(link, parent),
     kind: MyHDF5EntityKind.Datatype,
     type: rawDatatype.type,
   };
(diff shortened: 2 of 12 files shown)


.prettierignore
```diff
@@ -1,3 +1,4 @@
 /build/
 /storybook-static/
 /packages/**/dist/
+/package-lock.json
```
axelboc, line 4: Shaving off a couple of seconds when running Prettier.

src/h5web/providers/models.ts
```diff
@@ -59,7 +59,7 @@ export interface MyHDF5Entity {
   uid: string;
   name: string;
   kind: MyHDF5EntityKind;
-  parents: MyHDF5Group[];
+  parent?: MyHDF5Group;
```
axelboc, line 62: Optional so as to allow creating mock entities without parents - i.e. to allow creating child entities before their parents.

src/h5web/test-utils.tsx
```diff
@@ -76,31 +76,26 @@ export function prepareForConsoleError() {
 }
 /* eslint-enable no-console */
 
-type Opts = Partial<Pick<MyHDF5Dataset, 'id' | 'parents' | 'attributes'>>;
+type Opts = Partial<Pick<MyHDF5Dataset, 'id' | 'attributes'>>;
```
axelboc, line 79: While I'm at it, I'm removing the ability to create a parent before its child (see in tests below).

src/h5web/visualizations/index.test.ts
```diff
@@ -211,13 +211,16 @@ describe('Visualization definitions', () => {
     });
 
     it('should support group with absolute `default` path to supported NXdata group', () => {
-      const group = makeMyNxEntityGroup('foo', 'NXentry', {
-        defaultPath: '/foo/bar',
-        children: [makeMyNxDataGroup('bar', spectrumDatasetInt1D)],
-        parents: [makeMyNxEntityGroup('root', 'NXroot')],
```
axelboc, line 217: In this "bottom-up" syntax, parents were created before children (i.e. in pseudo-code `makeChild(makeParent())`). In #365, I'm using a "top-down" in which children are created before their parents (i.e. in pseudo-code: `makeParent(makeChild())`).

The "bottom-up" syntax was only used in this file and has a simple alternative as seen below. I think it's better to stick to only one syntax, and to favour the syntax that matches the mental model of a tree (i.e. the "top-down" syntax).

src/h5web/test-utils.tsx
```diff
@@ -76,32 +76,28 @@ export function prepareForConsoleError() {
...
-    parent.children = [...parent.children, group];
-  });
-
-  group.children.forEach((child) => {
-    child.parents = [group, ...child.parents];
-  });
+  group.children = group.children.map((child) => ({
+    ...child,
+    parent: group,
+  }));
```
axelboc, line 100: I'm no longer mutating the children.

src/h5web/providers/models.ts
loichuder: And also for the root link ?

src/h5web/visualizations/index.test.ts
loichuder: I would have inverted the names "top-down" and "bottom-up" :joy: 

But yes, the new syntax feels more straightforward to me

src/h5web/explorer/utils.ts
```diff
@@ -60,38 +60,36 @@ function buildDatatype(
...
     }
 
     switch (link.collection) {
       case HDF5Collection.Groups:
-        return buildGroup(metadata, link, newParents);
+        return buildGroup(metadata, link, group);
       case HDF5Collection.Datasets:
-        return buildDataset(metadata, link, newParents);
+        return buildDataset(metadata, link, group);
       default:
```
loichuder, line 91: I caught a glimpse of this before: I think the `default` of the switch should throw an error and that a `case: HDF5Collection.Datatypes` should be added so that creation of malformed entities do not fail silently.

src/h5web/visualizations/index.test.ts
```diff
@@ -211,13 +211,16 @@ describe('Visualization definitions', () => {
...
+        children: [
+          makeMyNxEntityGroup('foo', 'NXentry', {
+            defaultPath: '/foo/bar',
+            children: [makeMyNxDataGroup('bar', spectrumDatasetInt1D)],
+          }),
+        ],
       });
 
-      expect(supportsEntity(group)).toBe(true);
+      expect(supportsEntity(group.children[0])).toBe(true);
```
loichuder, line 223: It is not as clear that the test should check that it is the `foo` group that should be supported.

Perhaps renaming `group` to `rootGroup` would help.

loichuder (approved):

src/h5web/providers/models.ts
axelboc: Yes, sorry, that's the main reason ^^ -- but I could have made it optional only on groups.

src/h5web/visualizations/index.test.ts
axelboc: Haha yeah, can go either way depending on whether we're talking imperatively or declaratively.

src/h5web/visualizations/index.test.ts
axelboc: I've also added `const [groupWithDefault] = rootGroup.children;`

PR #371, opened 2020-12-09
Refactoring error bars

loichuder:
- Now using `useUpdate` and declarative geometries
- Changed the cap sprite
- Refactored the computation of bar segments and, data and caps points.


src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -19,45 +23,82 @@ function DataCurve(props: Props): ReactElement {
...
-        color={color}
-        visible={showGlyphs}
-        size={6}
-      />
-      <Line geometry={dataGeometry} visible={showLine}>
+      <Line visible={showLine}>
         <lineBasicMaterial attach="material" color={color} linewidth={2} />
+        <bufferGeometry attach="geometry" ref={ref} />
       </Line>
+      <points visible={showGlyphs} geometry={ref.current}>
```
loichuder, line 87: As the `geometry` is set differently here and in `ErrorBars`, I moved the `<points /> out of `DataGlyphs` to keep only the sprite material (now `GlyphMaterial`)

src/h5web/visualizations/line/ErrorBars.tsx
```diff
@@ -0,0 +1,41 @@
...
+      </LineSegments>
+      <points visible={visible}>
+        <GlyphMaterial glyphURL={GLYPH_URLS.Cap} color={color} size={8} />
+        <bufferGeometry attach="geometry" ref={capsGeometry} />
+      </points>
+    </>
+  );
+}
+
+export default ErrorBars;
```
loichuder, line 41: Renamed `ErrorBarCurve` in `ErrorBars` so that we don't think that `DataCurve` and `ErrorBarCurve` are at the same level of hierarchy.

I only added the `useUpdate` (so that it takes points in props rather than the geometry) and removed the `Suspense` as it is already in the HOC

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -19,45 +22,82 @@ function DataCurve(props: Props): ReactElement {
...
-        // This is to avoid a three.js warning when ordinateScale(val) is Infinity
-        Number.isFinite(ordinate) ? ordinate : 0,
-        // Move NaN/Infinity out of the camera FOV (negative val for logScale).
-        // This allows to have only curve segments for the positive values
-        Number.isFinite(ordinate) ? 0 : camera.far
-      );
+  const points = useMemo(() => {
+    const dataPoints: Vector3[] = [];
+    const errorBarSegments: Vector3[] = [];
+    const errorCapPoints: Vector3[] = [];
```
axelboc, line 37: I was about to suggest initializing the arrays with `new Array(ordinates.length)`, and then assigning values instead of pushing them, but apparently the performance gain is highly debatable.

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -19,45 +22,82 @@ function DataCurve(props: Props): ReactElement {
...
+      const y = ordinateScale(val);
+
+      // Move NaN/Infinity out of the camera FOV (negative val for logScale).
+      // This allows to have only curve segments for the positive values
+      const finiteData = Number.isFinite(x) && Number.isFinite(y);
+      const dataVector = finiteData
+        ? // Set x,y to 0 to avoid a three.js warning when one is Infinity
+          new Vector3(x, y, 0)
+        : // Move NaN/Infinity out of the camera FOV (negative val for logScale).
+          // This allows to have only curve segments for the positive values
```
axelboc, line 50: duplicate comment

src/h5web/visualizations/line/DataCurve.tsx
```diff
@@ -19,45 +22,82 @@ function DataCurve(props: Props): ReactElement {
...
+          new Vector3(0, 0, camera.far);
+
+      dataPoints.push(dataVector);
+      if (!errors || !errors[index] || !finiteData) {
```
axelboc, line 54: Maybe add a line `const error = errors && errors[index]` and then check `!error || !isFiniteData`? This way you don't need to reaccess the `error` value with the `index` below.

src/h5web/visualizations/line/ErrorBars.tsx
```diff
@@ -0,0 +1,41 @@
...
+    (geometry) => geometry.setFromPoints(capsPoints),
+    [capsPoints]
+  );
+
+  return (
+    <>
+      <LineSegments visible={visible}>
+        <lineBasicMaterial attach="material" color={color} linewidth={2} />
+        <bufferGeometry attach="geometry" ref={barsGeometry} />
+      </LineSegments>
```
axelboc, line 32: `lineSegments` with a lower case doesn't work by any chance?

axelboc (approved):
Awesome!!

src/h5web/visualizations/line/DataCurve.tsx
loichuder: Also, only `dataPoints` is ensured to have the same length as `ordinates`.

src/h5web/visualizations/line/DataCurve.tsx
loichuder: :fishing_pole_and_fish:

src/h5web/visualizations/line/ErrorBars.tsx
loichuder: :fishing_pole_and_fish: :fishing_pole_and_fish:

src/h5web/visualizations/line/DataCurve.tsx
axelboc: Oh right, I must have been thinking this when looking at the previous version, when you were adding `undefined` values.

PR #377, opened 2020-12-10
Refactor `getEntityAtPath` to accept string path

axelboc:
`REACT_APP_DEFAULT_PATH` now accepts a string path, like `/entry_0000/2_buffer_subtraction`.

### DIFF
--- a/src/h5web/explorer/utils.ts
+++ b/src/h5web/explorer/utils.ts
@@ -17,6 +17,7 @@ import {
   makeMyLink,
 } from '../providers/my-utils';
 import { isGroup, isReachable } from '../providers/utils';
+import { getChildEntity } from '../visualizations/nexus/utils';
 
 function buildDataset(
   metadata: HDF5Metadata,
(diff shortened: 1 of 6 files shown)


src/h5web/explorer/utils.ts
```diff
@@ -87,17 +88,32 @@ export function buildTree(
   });
 }
 
+function findRoot(entity: MyHDF5Group): MyHDF5Metadata {
+  const { parent } = entity;
+  return parent ? findRoot(parent) : entity;
+}
```
axelboc, line 94: Moved up from below and simplified to only accept groups.

src/h5web/explorer/utils.ts
```diff
@@ -87,17 +88,32 @@ export function buildTree(
...
+
 export function getEntityAtPath(
-  entity: MyHDF5Entity,
-  path: number[]
-): MyHDF5Entity {
-  if (path.length === 0 || !isGroup(entity)) {
-    return entity;
+  baseGroup: MyHDF5Group,
+  path: string,
+  allowSelf = true
```
axelboc, line 99: I found it more "natural" for this function to return the entity itself when the path is `''` (or when the path is `'/'` and the entity is the root entity). However, this behaviour is problematic in `findNxData`, when following an empty default attribute (or `'/'` on the root), because it leads to infinite recursion. I chose to add this attribute to make the behaviour explicit.

src/h5web/explorer/utils.ts
loichuder: I would have kept `MyHDF5Entity` as the arg type as it might be useful to find the root from any entity.

src/h5web/explorer/utils.ts
loichuder: Looks good :+1:

loichuder (approved):
Nice !

src/h5web/explorer/utils.ts
axelboc: Arf, I remember now why I changed it. If I change the arg to `MyHDF5Entity`, I need to change the return type to `MyHDF5Entity` as well. That's because a dataset for instance may not have a parent, and the dataset itself is returned in this case. However, this doesn't seem to affect the `getEntityAtPath` function, so the change makes sense. We can always add an override with `MyHDF5Group` later on if we need to.

loichuder: I would argue that if you can't follow the parents up to the root group, it should return `undefined`. But then, it raises other problems...

axelboc: Well, it's arguable since the dataset is, in that case, the root of the structure! 😄

axelboc: That's what I had initially, but it was all a bit more pointlessly complicated, I thought.

loichuder: Ha, perhaps we don't have the same view on the signification of `root`. 

For me, it should be the root group of the file and nothing else. But as you say, it is arguable since we are dealing here with the tree structure and not the file structure itself.

PR #378, opened 2020-12-11
Fixed geometry display in DataCurve

### LINKED ISSUE #373: Problem with switching scale on fully negative curves
- http://localhost:3000/mock
- `nD_datasets/threeD`
- Switch to log on `Y` => curve disappears
- Go to other dataset and come back
- Switch to linear on `Y` => curve doesn't reappear! 🤷‍♂️

loichuder:
Fix #373 

This reverts a part of #371 by creating the shared `BufferGeometry` imperatively rather than declaratively with `useUpdate`. `useUpdate` indeed leads a `undefined` shared geometry at first render: https://github.com/pmndrs/react-three-fiber/discussions/863

### DIFF
--- a/src/h5web/visualizations/line/DataCurve.tsx
+++ b/src/h5web/visualizations/line/DataCurve.tsx
@@ -2,7 +2,7 @@ import React, { Suspense, ReactElement, useMemo } from 'react';
 import { Line } from 'react-three-fiber/components';
 import { CurveType } from './models';
 import GlyphMaterial from './GlyphMaterial';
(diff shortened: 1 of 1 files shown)


axelboc (approved):
You'll fix the error geometries in another PR? You could probably create a little utility and then wrap it in `createMemo`.

loichuder:
The issue is when **sharing** the geometry with `ref.current`. In the error geometries, I have separate geometries for the caps and the bars that I set via `ref` so it works without trouble.

axelboc:
Oh right! 💯

PR #379, opened 2020-12-11
Remove obsolete type guards and fix `axes` attribute creation

axelboc:


src/h5web/providers/my-utils.ts
```diff
@@ -149,7 +148,10 @@ export function makeMyNxDataGroup<
     errors?: MyHDF5Dataset<HDF5SimpleShape, HDF5NumericType>;
     title?: MyHDF5Dataset<HDF5ScalarShape, HDF5StringType>;
     silxStyle?: SilxStyle;
-  } & AllOrNone<{ axes: T; axesAttr: (keyof T | '.')[] }> &
+  } & (
+    | { axes: T; axesAttr: (Extract<keyof T, string> | '.')[] }
+    | { axes?: never; axesAttr?: never }
+  ) &
```
axelboc, line 154: Changing my mind on this. I think inlining is clearer after all. 🤷‍♂️ 

The change below also required a slight change in typing: `Extract<keyof T, string>` instead of `keyof T`, as `keyof T` resolves to `string | number | Symbol` as explained in https://github.com/microsoft/TypeScript/issues/23724#issuecomment-384807714 ... even though `T` is `Record<string, ...>`! This seems very odd... there's a comment about it in the linked PR, but no answer.

src/h5web/providers/my-utils.ts
```diff
@@ -176,7 +178,7 @@ export function makeMyNxDataGroup<
         ...(groupOpts.attributes || []),
         makeStrAttr('NX_class', 'NXdata'),
         makeStrAttr('signal', signal.name),
-        makeStrAttr('axes', axesAttr),
+        ...(axesAttr ? [makeNxAxesAttr(axesAttr)] : []),
```
axelboc, line 181: `makeStrAttr` was accepting `any` value, so we were creating an attribute of type/shape string/scalar but with an array as value... Since we never look at the type/shape of attributes, this had no impact on the app.

I was also creating the attribute regardless of whether `axesAttr` was provided... 🤦

src/h5web/providers/utils.ts
```diff
@@ -123,21 +116,9 @@ export function assertNumericType<S extends HDF5Shape>(
...
-  dataset: HDF5Dataset<HDF5Shape, T>
-): asserts dataset is HDF5Dataset<HDF5SimpleShape, T> {
-  if (!hasSimpleShape(dataset)) {
-    throw new Error('Expected dataset to have simple shape');
-  }
-
-  if (dataset.shape.dims.length === 0) {
-    throw new Error('Expected dataset with simple shape to have dimensions');
-  }
-}
```
axelboc, line 135: This was the last remaining "old" assertion function/type guard combination.

src/h5web/providers/my-utils.ts
loichuder: Probably `keyof` does not behave well with templates ? :shrug:

src/h5web/providers/my-utils.ts
loichuder: :fishing_pole_and_fish:

loichuder (approved):

src/h5web/providers/my-utils.ts
axelboc: Generics you mean? The guy's comment on https://github.com/microsoft/TypeScript/issues/23724#issuecomment-384807714 doesn't mention a generic so I don't think it's related.

axelboc: Link to the comment: https://github.com/microsoft/TypeScript/issues/23724#issuecomment-638840419

PR #380, opened 2020-12-11
Remove `id` and `collection` from raw HDF5 entities

axelboc:
This brings the raw models (`HDF5Dataset` and the like) in line with `hdf5tojson`'s output, and therefore with what is returned by the Silx and HSDS back-ends. Next step will be to move `buildTree` into the providers!


axelboc (commented):
I should add that we can do this refactoring because `buildTree` doesn't need `id` and `collection` on entities to generate the new nested structure. It gets these information from the links directly.

loichuder (approved):
Simplifies many things :+1: 

A dumb question then: remind me how we find the `id` to fetch dataset values, then ?

axelboc:
The `id` and the `collection` are both present on hard links, so `buildTree` uses those when creating the entities.

PR #384, opened 2020-12-14
Added fallback to RawVis when the dataset is too big

loichuder:


axelboc (approved):
Awesome, so the `stringify` is fast, good know.

loichuder:
> Awesome, so the `stringify` is fast, good know.

Yup, as you suspected, it is the display in the DOM that was slow.

PR #386, opened 2020-12-14
First round of clean up

axelboc:
- Move shared guards to `h5web/guards.ts`.
- Don't use `make*` mock data utilities in `buildTree` to avoid confusion.
- Remove `RawSilxStyle` model and rename properties of `SilxStyle` to camel case.
- Silently protect against malformed `SILX_style `attribute.

### DIFF
--- a/src/h5web/explorer/utils.ts
+++ b/src/h5web/explorer/utils.ts
@@ -9,83 +9,110 @@ import {
   MyHDF5Datatype,
   HDF5HardLink,
   MyHDF5Metadata,
+  MyHDF5EntityKind,
+  HDF5Datatype,
+  HDF5Dataset,
 } from '../providers/models';
-import {
-  makeMyDataset,
-  makeMyDatatype,
-  makeMyGroup,
-  makeMyLink,
-} from '../providers/my-utils';
 import { isGroup, isReachable } from '../guards';
 import { getChildEntity } from '../visualizations/nexus/utils';
+import { nanoid } from 'nanoid';
 
-function buildDataset(
-  metadata: HDF5Metadata,
-  link: HDF5HardLink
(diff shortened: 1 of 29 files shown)


src/h5web/explorer/utils.ts
```diff
@@ -9,83 +9,110 @@ import {
...
-import { isGroup, isReachable } from '../providers/utils';
+import { isGroup, isReachable } from '../guards';
 import { getChildEntity } from '../visualizations/nexus/utils';
+import { nanoid } from 'nanoid';
 
-function buildDataset(
-  metadata: HDF5Metadata,
-  link: HDF5HardLink
-): MyHDF5Dataset {
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
axelboc, line 26: I've managed to get rid of these ESLint comments by changing the signature of `buildGroup` to accept a metadata object of type `Required<HDF5Metadata>`. I prepare `HDF5Metadata` to fit this type in `buildTree` below.

src/h5web/explorer/utils.ts
```diff
@@ -9,83 +9,110 @@ import {
...
 
     switch (link.collection) {
       case HDF5Collection.Groups:
         return buildGroup(metadata, link);
       case HDF5Collection.Datasets:
-        return buildDataset(metadata, link);
+        return buildDataset(metadata.datasets[link.id], link);
       case HDF5Collection.Datatypes:
-        return buildDatatype(metadata, link);
+        return buildDatatype(metadata.datatypes[link.id], link);
```
axelboc, line 75: Passing the raw entities instead of the whole metadata object just simplifies the signature of `buildDataset` and `buildDatatype`.

PR #387, opened 2020-12-14
Move provider and mock data utilities

axelboc:


src/h5web/providers/mock/data.ts
```diff
@@ -203,3 +210,16 @@ export const mockValues = {
   fourD_image: fourD,
   oneD_errors: oneD.map((x) => Math.abs(x) / 10),
 };
+
+export function getMockDataArray(path: string): ndarray {
```
axelboc, line 214: I couldn't put this function in `utils`, because it needs access to `mockValues` and `mockMetadata`, which are located in this file. If I put it in `utils`, then `utils` imports `data` and `data` imports `utils`...

axelboc: It felt lost in `utils` anyway. I think it makes more sense to have it here: `utils` holds the functions that generate mock data; and `data` is for generating and accessing mock data.

loichuder (approved):
All tidied up :sparkles:

PR #389, opened 2020-12-15
Split raw HDF5 models into separate file and rename new models

axelboc:
Final step!! 🎉 

Next up will be refactoring a few things to take even more advantage of the new structure (`getDatasetLabel`, `findSignalName/Dataset`, etc.)

### DIFF
--- a/src/h5web/providers/utils.ts
+++ b/src/h5web/providers/utils.ts
@@ -1,28 +1,24 @@
+import { Group, Dataset, Datatype, Metadata, EntityKind } from './models';
 import {
   HDF5Collection,
   HDF5LinkClass,
   HDF5RootLink,
   HDF5Metadata,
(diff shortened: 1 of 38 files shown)


src/h5web/providers/utils.ts
```diff
@@ -1,46 +1,39 @@
...
     type,
     rawLink: link,
   };
 }
 
-function buildDatatype(
-  datatype: HDF5Datatype,
-  link: HDF5HardLink
-): MyHDF5Datatype {
+function buildDatatype(datatype: HDF5Datatype, link: HDF5HardLink): Datatype {
```
axelboc, line 29: Shorter names mean that a few signatures can now fit on single lines.

loichuder (approved):
:tada: 

Looks much better without the `MyHDF5` prefix

PR #391, opened 2020-12-15
Take advantage of `name` property on `Dataset` model

axelboc:
Sooo much cleaner!! Good call on having a second look at `useDatasetValues` 😉


loichuder (approved):
:+1:

PR #393, opened 2020-12-15
Bit more cleanup in `useNxData`

axelboc:
See comments.

### DIFF
--- a/src/h5web/providers/mock/data.ts
+++ b/src/h5web/providers/mock/data.ts
@@ -21,7 +21,7 @@ import {
   makeDatatype,
   makeExternalLink,
   makeGroup,
-  makeNxData,
+  makeNxDataGroup,
   makeNxDataset,
   makeNxGroup,
   makeSimpleDataset,
@@ -55,7 +55,7 @@ export const mockMetadata = makeNxGroup(mockDomain, 'NXroot', {
       children: [
         makeNxGroup('nx_process', 'NXprocess', {
           children: [
-            makeNxData('nx_data', {
+            makeNxDataGroup('nx_data', {
               signal: makeNxDataset('twoD', intType, [20, 41]),
               silxStyle: { signalScaleType: ScaleType.SymLog },
               title: makeDataset('title', scalarShape, stringType, {
@@ -64,7 +64,7 @@ export const mockMetadata = makeNxGroup(mockDomain, 'NXroot', {
             }),
           ],
         }),
-        makeNxData('spectrum', {
+        makeNxDataGroup('spectrum', {
           signal: makeNxDataset('twoD_spectrum', intType, [20, 41], {
             interpretation: 'spectrum',
             units: 'arb. units',
@@ -75,7 +75,7 @@ export const mockMetadata = makeNxGroup(mockDomain, 'NXroot', {
           axes: { X: makeNxDataset('X', intType, [41], { units: 'nm' }) },
           axesAttr: ['.', 'X'],
         }),
-        makeNxData('image', {
+        makeNxDataGroup('image', {
           signal: makeNxDataset('fourD_image', intType, [3, 9, 20, 41], {
             longName: 'Interference fringes',
             interpretation: 'image',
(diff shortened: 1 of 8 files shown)


src/h5web/providers/mock/data.ts
```diff
@@ -21,7 +21,7 @@ import {
   makeDatatype,
   makeExternalLink,
   makeGroup,
-  makeNxData,
+  makeNxDataGroup,
```
axelboc, line 24: Adding back the `group` suffix for consistency with `findNxDataGroup` (as in the group entity) and `useNxData` (as in the parsed NXdata info).

src/h5web/providers/mock/utils.ts
```diff
@@ -192,6 +192,10 @@ export function makeExternalLink(
 /* ----------------- */
 /* ----- NEXUS ----- */
 
+function makeNxAxesAttr(axes: string[]): HDF5Attribute {
+  return makeAttr('axes', makeSimpleShape([axes.length]), stringType, axes);
+}
+
```
axelboc, line 198: I'm shuffling some functions around and I'm refactoring `makeNxDataGroup` so it calls `makeNxGroup` internally instead of `makeGroup`. The former adds the `NX_class` attribute. It's not much, but it feels more logical and consistent with other mock data functions.

src/h5web/visualizations/containers/hooks.ts
```diff
@@ -16,5 +16,5 @@ export function useDatasetValues(
   const { getValues } = useContext(ProviderContext);
   const [valuesReader] = useAsyncResource(getValues, datasets);
 
-  return valuesReader() || {};
+  return valuesReader();
```
axelboc, line 19: `valuesReader()` either returns the fetched data or throws if they're still being fetched. So we don't need a fallback, and we don't need an early return in `useNxData` after all.

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -12,8 +12,8 @@ import type { NxData } from './models';
 import {
   findSignalDataset,
   getDatasetLabel,
-  getNxAxisNames,
-  parseSilxStyleAttribute,
+  getNxAxes,
+  getSilxStyle,
```
axelboc, line 16: Just renaming these two utilities for consistency.

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -22,27 +22,22 @@ export function useNxData(group: Group): NxData {
 
   const signalDataset = findSignalDataset(group);
   const signalValue = values[signalDataset.name];
-  if (!signalValue) {
-    return { signal: { dims: signalDataset.shape.dims } };
-  }
```
axelboc, line 27: No need for the early return as mentioned. If we reach this point, the array must be available.

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -22,27 +22,22 @@ export function useNxData(group: Group): NxData {
...
-  const axesNames = getNxAxisNames(group);
 
-  const silxStyle = parseSilxStyleAttribute(group);
+  const silxStyle = getSilxStyle(group);
   const { axesScaleType, signalScaleType } = silxStyle;
 
-  const axisMapping = axesNames.map((axisName, i) => {
-    if (!axisName) {
+  const axisMapping = getNxAxes(group).map((axisName, i) => {
+    if (axisName === '.') {
```
axelboc, line 31: Checking for `'.'` here directly instead of converting them to `undefined` in `getNxAxes` (formerly `getNxAxisNames`).

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -22,27 +22,22 @@ export function useNxData(group: Group): NxData {
...
     }
 
     const axisDataset = getChildEntity(group, axisName);
     assertDefined(axisDataset);
     assertDataset(axisDataset);
 
-    const axisValue = values[axisDataset.name];
-    assertOptionalArray<number>(axisValue);
+    const axisValue = values[axisName];
+    assertArray<number>(axisValue);
```
axelboc, line 40: If an axis is listed in the `axes` attribute, then it must have a value.

loichuder (approved):

PR #394, opened 2020-12-15
Basic working version of the Jupyter backend provider

loichuder:
A few loose ends left to tie:

## What we can fix:
- [x] The root response is cast to `Group`/`Metadata`

## What needs to be fixed in the backend:
- [ ] The type and shape of attributes cannot be inferred yet as the backend does not provide the necessary info. **Left unresolved for now. See  #406**
- [ ] The type of datasets is also tough to find as the `dtype` field is the NumPy type and not the HDF5 type that we use. We could infer the HDF5 type from the `dtype` on the front but I would prefer that the backend provides directly the HDF5 type. **Left unresolved for now. See  #406**
- [x] Similarly, `shape` is the NumPy shape that are what we call `dims`. <s>Complex shapes are therefore not supported yet.</s>. 
**EDIT**: I think it is fine for now: shapes can be either scalar, simple or null. Inferring from the dims seems sufficient. 
- [x] The signature of the attribute response is not the one described in the API. It necessitates an unpacking step (`attrs`).

## Other points:
- [x] The path in the file is used as `id`. The backend could return this information but it is not critical for now.
- [x] As discussed earlier, having an `attributeCount` and a `childrenCount` field in the metadata could help in avoiding unnecessary requests.

### DIFF
--- a/src/h5web/providers/jupyter/api.ts
+++ b/src/h5web/providers/jupyter/api.ts
@@ -1,0 +1,155 @@
+import axios, { AxiosInstance } from 'axios';
+import { Group, Dataset, Metadata, EntityKind, Link } from '../models';
+import type { ProviderAPI } from '../context';
+import {
+  assertGroupContent,
+  isDatasetResponse,
+  isGroupResponse,
+} from './utils';
+import type {
+  JupyterAttrsResponse,
+  JupyterContentResponse,
+  JupyterDataResponse,
+  JupyterMetaResponse,
+} from './models';
+import { nanoid } from 'nanoid';
+import { makeStrAttr, floatType } from '../mock/utils';
+import {
+  HDF5Id,
+  HDF5LinkClass,
+  HDF5ShapeClass,
+  HDF5Value,
+  HDF5SoftLink,
+} from '../hdf5-models';
+import { assertGroup } from '../../guards';
+
+export class JupyterApi implements ProviderAPI {
+  public readonly domain: string;
+  private readonly client: AxiosInstance;
+
+  private values: Record<HDF5Id, HDF5Value> = {};
+
+  public constructor(url: string, domain: string) {
+    this.domain = domain;
+    this.client = axios.create({
+      baseURL: `${url}/hdf`,
+    });
+  }
+
+  public async getMetadata(): Promise<Metadata> {
+    const rootId = '/';
+    const rootGrp = await this.processEntity(rootId);
+    assertGroup(rootGrp);
+    return rootGrp;
+  }
+
+  public async getValue(id: HDF5Id): Promise<HDF5Value | undefined> {
+    if (id in this.values) {
+      return this.values[id];
+    }
+
+    const value = await this.fetchValue(id);
+    this.values[id] = value;
+    return this.values[id];
+  }
+
+  private async fetchAttributes(path: string): Promise<JupyterAttrsResponse> {
+    const { data } = await this.client.get<JupyterAttrsResponse>(
+      `/attrs/${this.domain}?uri=${path}`
+    );
+    return data;
+  }
+
+  private async fetchMetadata(path: string): Promise<JupyterMetaResponse> {
+    const { data } = await this.client.get<JupyterMetaResponse>(
+      `/meta/${this.domain}?uri=${path}`
+    );
+    return data;
+  }
+
+  private async fetchContents(path: string): Promise<JupyterContentResponse> {
+    const { data } = await this.client.get<JupyterContentResponse>(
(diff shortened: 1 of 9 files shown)


src/h5web/providers/jupyter/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+import {
+  JupyterMetaDatasetResponse,
+  JupyterMetaGroupResponse,
+  JupyterMetaResponse,
+} from './models';
+
+export function isGroupResponse(
+  response: JupyterMetaResponse
+): response is JupyterMetaGroupResponse {
+  return response.type === EntityKind.Group;
```
axelboc, line 11: Ah, funny that it uses something similar to our new `EntityKind` enum.

src/h5web/providers/jupyter/api.ts
```diff
@@ -0,0 +1,136 @@
...
+    this.domain = domain;
+    this.client = axios.create({
+      baseURL: url,
+    });
+  }
+
+  public async getMetadata(): Promise<Metadata> {
+    const rootId = '/';
+    // TODO: To fix with generics ?
+    return (await this.processEntity(rootId)) as Group;
```
axelboc, line 37: You could just use `assertGroup`.

src/h5web/providers/jupyter/api.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    const value = await this.fetchValue(id);
+
+    this.values[id] = value;
+    return this.values[id];
+  }
+
+  private async fetchAttributes(uri: string): Promise<JupyterAttrResponse> {
+    const { data } = await this.client.get<JupyterAttrResponse>(
+      `/hdf/attrs/${this.domain}?uri=${uri}`
```
axelboc, line 53: I would put the `/hdf` prefix in the base URL config of axios.

axelboc: I would also rename `uri` to `path` here and in other places.

src/h5web/providers/jupyter/models.ts
```diff
@@ -0,0 +1,43 @@
...
+  type: EntityKind.Dataset;
+  uri: string;
+}
+
+export interface JupyterContent {
+  type: EntityKind;
+  name: string;
+  uri: string;
+  content: null;
+}
```
axelboc, line 33: Could you remind me what the term "content" refers to? I'm confused.

loichuder: I think it is not clear yet :smile:

loichuder: The *content* endpoint is to recover `name`, `type` and `uri` of children entities of a group. 

For a dataset, it returns the `name`, `type` and `uri` **of the dataset** as well as its metadata (that you could get through the *meta* endpoint) in a `content` field.

axelboc: Alright, shame that `fetchMetadata` doesn't already return the URIs of the children. It's fine for now of course 👍

src/h5web/providers/jupyter/api.ts
```diff
@@ -0,0 +1,155 @@
...
+      attributeCount > 0 ? await this.fetchAttributes(path) : {};
+    const attributes = Object.entries(attrReponse).map(([name, value]) =>
+      // TODO: fix this once I can infer the type of attributes
+      makeStrAttr(name, value)
+    );
+
+    if (isGroupResponse(response)) {
+      const { name, type, childrenCount } = response;
+      const contents = childrenCount > 0 ? await this.fetchContents(path) : [];
+      assertGroupContent(contents);
```
axelboc, line 101: I reckon you could rename `fetchContents` to `fetchGroupContents` and assume that you get a group content response. We already know we're fetching a group and we won't need it for other entities.

axelboc: If you want to keep the exact mapping between the API and the `fetch` methods, you could instead use a generic: `this.fetchContents<JupyterContentGroupResponse>()`

axelboc (approved):
Nicely done!

src/h5web/providers/jupyter/api.ts
loichuder: I know that it is double-checking but I like the explicit `assert`.

`fetchGroupContents` despite its name could be misused. The `assert` has the merit of making things really clear and without ambiguity.

For the generic, from what I understand, I cannot infer the type from the parameter which leads to a disguised type assertion: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes

PR #395, opened 2020-12-16
Remove obsolete early returns in containers

axelboc:
... + a couple of other minor changes

### DIFF
--- a/src/h5web/dimension-mapper/DimensionMapper.tsx
+++ b/src/h5web/dimension-mapper/DimensionMapper.tsx
@@ -14,10 +14,6 @@ interface Props {
 function DimensionMapper(props: Props): ReactElement {
   const { rawDims, mapperState, onChange } = props;
 
-  if (!mapperState) {
-    return <></>;
-  }
-
   return (
     <div className={styles.mapper}>
       <div className={styles.axisMapperWrapper}>
--- a/src/h5web/visualizations/containers/HeatmapVisContainer.tsx
+++ b/src/h5web/visualizations/containers/HeatmapVisContainer.tsx
@@ -20,10 +20,6 @@ function HeatmapVisContainer(props: VisContainerProps): ReactElement {
   }
 
   const value = useDatasetValue(entity.id);
-  if (!value) {
-    return <></>;
-  }
-
   return <MappedHeatmapVis value={value} dims={dims} title={entity.name} />;
 }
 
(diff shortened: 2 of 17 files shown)


src/h5web/dimension-mapper/DimensionMapper.tsx
```diff
@@ -14,10 +14,6 @@ interface Props {
 function DimensionMapper(props: Props): ReactElement {
   const { rawDims, mapperState, onChange } = props;
 
-  if (!mapperState) {
-    return <></>;
-  }
```
axelboc, line 19: This one is obsolete as well since the `mapperState` cannot be `undefined`.

src/h5web/visualizations/containers/HeatmapVisContainer.tsx
```diff
@@ -20,10 +20,6 @@ function HeatmapVisContainer(props: VisContainerProps): ReactElement {
   }
 
   const value = useDatasetValue(entity.id);
-  if (!value) {
-    return <></>;
-  }
```
axelboc, line 25: `useDatasetValue` throws if `value` is being fetched. If for some reason we receive an `undefined` value, then the assertion in the mapped vis component or in the vis component (for `ScalarVis`) will throw.

src/h5web/visualizations/containers/NxSpectrumContainer.tsx
```diff
@@ -21,10 +21,6 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
     disableErrors(!errors);
   }, [disableErrors, errors]);
 
-  if (!signal.value) {
-    return <></>;
-  }
-
   if (errors && signal.value.length !== errors.length) {
```
axelboc, line 24: I'm able to remove the early return by marking `value` as being required in `SignalParams`. This is fine because we throw if we encounter an undefined signal value in `useNxData`.

src/h5web/visualizations/heatmap/MappedHeatmapVis.tsx
```diff
@@ -13,7 +13,7 @@ interface Props {
   value: HDF5Value;
   dims: number[];
   title?: string;
-  axisMapping?: AxisMapping[];
+  axisMapping?: AxisMapping;
```
axelboc, line 16: Just moving the `[]` into `AxisMapping` itself for consistency with `DimensionMapping`.

src/h5web/visualizations/nexus/models.ts
```diff
@@ -17,13 +17,14 @@ export enum NxInterpretation {
 
 interface SignalParams extends AxisParams {
   dims: number[];
+  value: number[];
```
axelboc, line 20: `value` now required as mentioned.

loichuder (approved):
Thanks for going through this !

PR #397, opened 2020-12-17
Unit test utilities

axelboc:


src/h5web/providers/mock/data.ts
```diff
@@ -40,6 +39,7 @@ export const mockMetadata = makeNxGroup(mockDomain, 'NXroot', {
       makeExternalLink('external_link', 'my_file', 'entry_000/dataset'),
       makeDatatype('datatype', compoundType),
       makeDataset('raw', scalarShape, compoundType),
+      makeDataset('raw_large', scalarShape, compoundType),
```
axelboc, line 42: Adding a large raw mock dataset for testing purposes.

src/h5web/providers/mock/data.ts
```diff
@@ -103,70 +103,18 @@ export const mockMetadata = makeNxGroup(mockDomain, 'NXroot', {
       ],
     }),
     makeGroup('nexus_malformed', [
-      makeGroup('default_not_string', [], {
-        attributes: [makeAttr('default', scalarShape, intType, 42)],
-      }),
-      makeGroup('default_empty', [], {
-        attributes: [makeStrAttr('default', '')],
-      }),
```
axelboc, line 111: Removing a few unlikely cases, here and in the E2E tests, now that they are better covered with unit tests.

src/h5web/visualizations/RawVis.tsx
```diff
@@ -11,12 +11,12 @@ function RawVis(props: Props): ReactElement {
   const valueAsStr = JSON.stringify(value, null, 2);
 
   if (valueAsStr.length > 1000) {
-    // eslint-disable-next-line no-console
-    console.log(valueAsStr);
+    console.log(value); // eslint-disable-line no-console
```
axelboc, line 14: Logging the raw value instead of the stringified value.

loichuder: Good call :+1:

src/h5web/providers/mock/utils.ts
```diff
@@ -76,6 +76,10 @@ export function makeStrAttr(name: string, value: string): HDF5Attribute {
   return makeAttr(name, { class: HDF5ShapeClass.Scalar }, stringType, value);
 }
 
+export function makeScalarIntAttr(name: string, value: number): HDF5Attribute {
```
loichuder, line 79: Would you mind replacing `Int` by `Integer` ? I had to read it twice the function name to understand it (and I got my coffee this time !)

src/h5web/visualizations/RawVis.tsx
```diff
@@ -11,12 +11,12 @@ function RawVis(props: Props): ReactElement {
...
+
     return (
       <p className={styles.fallback}>
-        The dataset is too big to be displayed. The values were printed in the
-        console.
+        The dataset is too big to be displayed and was logged to the console
+        instead.
```
loichuder, line 19: Thanks for the rewording. I was not really happy either but did not find an improvement.

src/h5web/visualizations/nexus/utils.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    });
+
+    it('should throw if signal dataset has no dimensions', () => {
+      const dataset = makeDataset('dataset', makeSimpleShape([]), stringType);
+      const group = makeGroup('group', [dataset], {
+        attributes: [makeStrAttr('signal', 'dataset')],
+      });
+
+      expect(() => findSignalDataset(group)).toThrow(/to have dimensions/u);
+    });
```
loichuder, line 180: Actually, I am not sure this case is possible. Either a dataset has a scalar shape or it has a simple shape and should have dims in this case.

Doesn't hurt to keep it, though.

src/h5web/visualizations/nexus/utils.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    it('should combine dataset name and value of `units` attribute if available', () => {
+      const dataset = makeDataset('foo', scalarShape, intType, {
+        attributes: [makeStrAttr('units', 'nm')],
+      });
+
+      expect(getDatasetLabel(dataset)).toBe('foo (nm)');
+    });
+
+    it('should fall back to dataset name', () => {
```
loichuder, line 251: ...when other attributes are not string ?

src/h5web/visualizations/nexus/utils.test.ts
```diff
@@ -0,0 +1,330 @@
...
+      });
+
+      expect(getSilxStyle(group)).toEqual(silxStyle);
+    });
+
+    it('should ignore unknown `SILX_style` options and invalid values', () => {
+      const silxStyle1 = {
+        unknown: ScaleType.Log,
+        signal_scale_type: 'invalid',
+        axes_scale_type: ScaleType.Linear, // should be array
```
loichuder, line 281: For consistency with `axes`, should it not convert this as an array ?

Also, I would perhaps add a check in `useNxData` to see if `axes` and `axes_scale_type` have the same length, logging a warning if not.

loichuder (approved):

src/h5web/providers/mock/utils.ts
axelboc: I was looking for consistency with `makeStrAttr`, which makes me realise that it should have been called `makeIntAttr`, but happy to change to `makeIntegerAttr`.

src/h5web/visualizations/nexus/utils.test.ts
axelboc: This checks for a shape with `dims` equal to `[]`, which makes no sense.

axelboc: I mean you'd hope that a back-end would never return this, so perhaps it is indeed useless to assert it.

src/h5web/visualizations/nexus/utils.test.ts
axelboc: 👍 I've used `should fall back to dataset name when other attributes are absent or not string` and added a check for a dataset with no attributes.

src/h5web/visualizations/nexus/utils.test.ts
axelboc: 👍 I've used should fall back to dataset name when other attributes are absent or not string and added a check for a dataset with no attributes.

src/h5web/visualizations/nexus/utils.test.ts
axelboc: Totally agree for the first part.

For the second part, I've had a play with it and it doesn't seem worth it, especially since if the lengths don't match, nothing breaks -- you just don't get the scale types you were hoping for. If `SILX_style` was really important, then it'd make sense to check it, but just logging a warning is not very helpful: the user would still have to realise that something is not right.

loichuder: >  you just don't get the scale types you were hoping for.
That what I expected. I thought it might be worth a warning as it might not be obvious to debug. 

But in the end, the user can still change the scales themselves so I won't push strongly for this.

src/h5web/visualizations/nexus/utils.test.ts
```diff
@@ -0,0 +1,333 @@
...
+      expect(getSilxStyle(group)).toEqual({
+        axesScaleType: [ScaleType.Log],
+      });
+    });
+
+    it('should ignore unknown `SILX_style` options and invalid values', () => {
+      const silxStyle = {
+        unknown: ScaleType.Log,
+        signal_scale_type: 'invalid',
+        axes_scale_type: [ScaleType.Linear, 'invalid'], // should be array
```
loichuder, line 291: Obsolete comment

loichuder (approved):

src/h5web/visualizations/nexus/utils.test.ts
axelboc: 🎣

PR #398, opened 2020-12-17
Respect auto-scale when toggling errors

### LINKED ISSUE #396: Respect auto-scale config when computing domain of signal with errors
I was the one who suggested that we compute the domain of the signal without the errors when `showErrors` is `false`, but I now realise this behaviour can be annoying, for instance when toggling errors while zoomed in on the curve.

Perhaps the inclusion of the errors in the computation of the domain should also depend on the `autoScale` setting. If so, then the "Auto-scale" button should no longer be disabled if the data has `errors`.

axelboc:
Fix #396  -- well, this was quite trivial!

ezgif-7-f407ffa185a1


loichuder (approved):
:rocket:

PR #404, opened 2020-12-18
Implement CI release workflow

axelboc:
For #358. To be tested once merged on `master`...

I think it's pretty self-explanatory, but here are the links to the GitHub Actions that I've used for reference:

- https://github.com/geritol/match-tag-to-package-version
- https://github.com/nyaa8/package-version (may not this one if the previous action were to output the version number - cf. https://github.com/geritol/match-tag-to-package-version/issues/2)
- https://github.com/Reedyuk/NPM-Version
- https://github.com/marketplace/actions/npm-publish

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,60 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🏷️
+        uses: actions/checkout@v2
+
+      - name: Match tag to package version 🧐
+        uses: geritol/match-tag-to-package-version@0.0.2
+        env:
+          TAG_PREFIX: refs/tags/v
+
+      - name: Expose version as env var 🤓
+        uses: nyaayaya/package-version@v1
+
+      - name: Use Node 12 🕹️
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12'
+
+      - name: Install ⚙️
+        uses: bahmutov/npm-install@v1
+        with:
+          working-directory: packages
+
+      - name: Build 🛠️
+        run: cd packages && npm run build
+
+      - name: Set @h5web/app version 📌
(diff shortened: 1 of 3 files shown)


packages/package.json
```diff
@@ -17,7 +17,7 @@
     "prebuild": "rm -Rf */dist",
     "build": "webpack"
   },
-  "devDependencies": {
+  "dependencies": {
```
axelboc, line 20: We don't really need those to be `devDependencies`, since we don't need to differentiate dev and non-dev deps...

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
+name: Storybook
```
loichuder, line 1: Name to change to **Release**

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          working-directory: packages
+
+      - name: Build 🛠️
+        run: cd packages && npm run build
+
+      - name: Set @h5web/app version 📌
+        uses: reedyuk/npm-version@vlatest
+        with:
+          version: ${{ env.PACKAGE_VERSION }}
```
loichuder, line 39: Can't we instead use the tag ? Just to avoid the _expose as env var_ action.

packages/package.json
loichuder: It does not change the `npm build` ?

loichuder (approved):
Eager to try :nerd_face:

.github/workflows/release.yml
axelboc: 🎣

.github/workflows/release.yml
axelboc: The name of the tag can be parsed from env var `GITHUB_REF`, but it's something like `refs/tags/vX.Y.Z`, so it's not as trivial as using an action.

packages/package.json
axelboc: Oh no, the deps are available in the same way.

PR #405, opened 2020-12-21
Trigger Storybook deployment from release workflow

axelboc:


.github/workflows/release.yml
```diff
@@ -60,3 +60,10 @@ jobs:
           package: 'packages/lib/dist/package.json'
           token: ${{ secrets.NPM_TOKEN }}
           access: public
+
+      - name: Deploy Storybook 🚀
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: Storybook
+          token: ${{ secrets.USER_TOKEN }} # GH scret containing a Personal Access Token owned by @axelboc
```
loichuder, line 68:
```suggestion
          token: ${{ secrets.USER_TOKEN }} # GH secret containing a Personal Access Token owned by @axelboc
```

.github/workflows/release.yml
```diff
@@ -60,3 +60,10 @@ jobs:
...
+
```
loichuder, line 69: What about putting the two job steps `Build Storybook` and `Deploy` directly here ?

It would avoid redoing the checkout/install, no ?

loichuder (changes requested):

.github/workflows/release.yml
axelboc: hmm not a bad idea

loichuder (approved):

PR #408, opened 2020-12-22
Rethink cypress snapshots approval workflow

axelboc:
Re-running the same workflow with different parameters turns out to be quite brittle. The `update-snapshots` branch is rebased and the _Update Cypress reference snapshots_ PR re-opened at unexpected times.

As discussed, I've simplified things a lot by running Cypress with `--env approveSnapshots=true` directly in the workflow triggered when posting the approval comment. I was able to simplify this comment to just `/approve` as well.


.github/workflows/approve-snapshots.yml
```diff
@@ -0,0 +1,50 @@
+name: Approve snapshots
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  dispatch:
+    runs-on: ubuntu-latest
+    if: github.event.issue.pull_request && github.event.comment.body == '/approve'
```
axelboc, line 10: I don't use the slash-command action any more, I just do a manual check here to decide whether to run the workflow or not. This workflow no longer triggers another workflow -- everything is self-contained.

.github/workflows/slash-commands.yml
```diff
@@ -1,37 +0,0 @@
...
-    runs-on: ubuntu-latest
-    if: contains(github.event.comment.html_url, '/pull/')
-
-    steps:
-      - name: Get branch name based on PR ID
-        uses: andrevalentin/get-branch-name-by-pr@v1
-        id: getBranchName
-        with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
-          pr-id: ${{ github.event.issue.number }}
```
axelboc, line 18: We had to find the PR's branch in order for the Cypress worfklow to be triggered on the right reference branch. Since we don't trigger a separate workflow, this action is no longer required.

axelboc:
I'm going to try to introduce a regression for testing purposes.

loichuder:
> I'm going to try to introduce a regression for testing purposes.

I think the problem runs indeed deeper than this. Even the PR adding marks to sliders that is bound to change snapshots pass the e2e tests.

Looking at the logs, there are indeed some failed snapshots but the e2e tests do not fail: https://github.com/silx-kit/h5web/pull/353/checks?check_run_id=1593563290

loichuder (approved):

loichuder:
> 
> I think the problem runs indeed deeper than this. Even the PR adding marks to sliders that is bound to change snapshots pass the e2e tests.
> 
> Looking at the logs, there are indeed some failed snapshots but the e2e tests do not fail: https://github.com/silx-kit/h5web/pull/353/checks?check_run_id=1593563290

The e2e were succeeding due to https://github.com/jaredpalmer/cypress-image-snapshot/issues/156

PR #415, opened 2020-12-23
Fix sizing of line toolbar

### LINKED ISSUE #412: Toolbar overlaps tabs
image

axelboc:
Fix #412 

This had nothing to do with having `min-width` on the scale selectors as I initially thought. Turns out the `ScaleSelector` component was rendering a fragment with two children, and this wasn't playing well with the responsive `Toolbar`.


src/h5web/toolbar/LineToolbar.tsx
```diff
@@ -47,8 +47,6 @@ function LineToolbar(props: ToolbarProps): ReactElement {
         onScaleChange={setXScaleType}
       />
 
-      <Separator />
-
```
axelboc, line 51: I've removed the separator between the two selectors. Can't remember if we had talked about it when you first implemented it. I feel like it's more in line with what we've done so far of "grouping" controls of the same type (single toggle buttons, notably)

loichuder: Yes we did. I preferred having a clear separation between the two dropdowns. But in the end, it looks all right without it due to a bigger spacing between the two.

loichuder (approved):
Nice work :slightly_smiling_face:

PR #419, opened 2020-12-23
Upgrade dependencies and use new JSX transform

axelboc:
The highlight is the upgrade to TS 4.1, which allows us to use the [new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) and remove default `React` imports. The new JSX transform is a milestone towards React 17. I'll try to upgrade it next.

I also add a new `analyze` script to `package.json` to analyse the JS bundles after `npm run build`.


loichuder (approved):
image
:joy:

Issue #417, opened 2020-12-23
Data fetching strategy 2.0


axelboc:
### Background

Our data fetching strategy is built on the concept of **Data Providers** that act as interfaces between back-end APIs (typically REST) and the H5Web application. A data provider injects functions through React Context that allow any component to request the data it needs, when it needs it:

```jsx
// app entrypoint
<SilxProvider domain="bsa_002_000-integrate-sub">
  <App />
</SilxProvider>

// internal component of H5Web
const { getValue } = useContext(ProviderContext);
useEffect(() => {
  getValue(myDataset.id).then(...);
}, [])
```

Data providers must all respect the same context provider contract by rendering the same high-level `Provider` component provided by H5Web. In our initial implementation, this contract was as follows:

```jsx
return <Provider api={{ domain, getMetdata, getValue }}>{children}</Provider>
```

`getMetadata` and `getValue` are asynchronous functions. The former is expected to return the entire metadata tree, and the latter is expected to return the value of a dataset given its ID.

### Limitations

The main limitations of our initial implementation are as follows:

- **The entire metadata tree must be fetched at once** (entities, attributes, children). This is problematic for the following reasons:
  - Real-world HDF5 back-ends (e.g. HSDS, Jupyter Lab HDF5 extension), don't provide endpoints to fetch an HDF5 file's entire metadata tree. They only provide endpoints to fetch the metadata of a specific entity ... and even specific parts of the entity's metadata (i.e. only the children or the attributes). Fetching the whole metadata tree at once therefore requires performing a very large number of requests in a waterfall manner. This takes a long time even for simple HDF5 files and is impractical for large files.
  - For very large HDF5 files (or files with lots of external links pointing to other files with lots of external links), transferring the whole metadata tree over the network is time-consuming, or at least very wasteful, especially considering that a user may wish to inspect only a specific branch of the tree in H5Web.
- **Data providers must manage their own cache**. This is problematic for the following reasons:
  - A data provider could implement its cache poorly or not implement one at all, leading to poor performance in H5Web.
  - The caching requirements are similar for all providers, leading to duplicated logic and unnecessarily tedious implementation. **_EDIT_** Note that some providers may require more internal caching if the data expected by H5Web is more atomic than the available endpoints (e.g. if an endpoint returns both the metadata and the attributes of an entity while H5Web requests both separately).
[...]

PR #420, opened 2021-01-05
Move caching of values out of providers

axelboc:
As per the roadmap in #417, the caching of dataset values is now done in hooks `useDatasetValue` and `useDatasetValues`. The `values` cache itself is a `Map` located in `Provider.ts`.

**EDIT** the caching of the values is still out of the providers, but it is now handled solely by our suspense library, [react-suspense-fetch](https://github.com/dai-shi/react-suspense-fetch). Until now, we used `use-async-resource`, which already had its own internal cache but was tricky to use and inefficient, notably because it relied on a hook (one hook declaration = one cache, which prevented us from having multiple components/hooks share the same cache).

---

@loichuder I'll merge my PRs into branch `fetching-strategy-2` and I'll create a final PR for it when I'm done. This will allow me to keep merging while still allowing you to review everything when you're back.


axelboc:
Can't get around this regression in the tests...

Also, I'm not happy with the way our cache interacts with `useAsyncResource`, which also has its own cache: `useAsyncResource(asyncFn, params)` starts fetching right away, before we have the opportunity to check our own cache, so it feels wrong.

- We could just use `useAsyncResource`'s cache alone, but it wouldn't be shared between the two `useDatasetValue(s)` hooks, so it's not great...
- If I use the hook lazily with `useAsyncResource(asyncFn)`, then `valueReader()` can return `undefined`, which means we have to handle `undefined` everywhere.

Basically, I feel like `useAsyncResource` does more than we need, and the fact that it's a hook really limits the way we can use it. I'm going to look for a lower-level `Suspense` library

src/h5web/providers/Provider.tsx
```diff
@@ -12,32 +14,29 @@ function Provider(props: Props): ReactElement {
...
-    api,
-  ]);
+  const { value: metadata, error } = useAsync(async () => {
+    return api.fetchMetadata();
+  }, [api]);
+
+  const valuesStore = useMemo(
+    () => createFetchStore(api.fetchValue.bind(api)),
+    [api]
+  );
```
axelboc, line 24: Creating the values store here, with the provider API's `fetchValue` async method. The store has an internal cache that matches the signature of `fetchValue` (i.e. an object with `HDF5Id` keys and `HDF5Value` values).

src/h5web/providers/jupyter/api.ts
```diff
@@ -60,7 +54,7 @@ export class JupyterApi implements ProviderAPI {
     return data;
   }
 
-  private async fetchMetadata(path: string): Promise<JupyterMetaResponse> {
+  private async _fetchMetadata(path: string): Promise<JupyterMetaResponse> {
```
loichuder, line 57: Could be renamed `fetchEntityMetadata`

axelboc: FYI, I've renamed it back to `fetchMetadata`, but yeah, your suggestion is more explicit.

PR #421, opened 2021-01-07
Implement async loading of groups metadata in Explorer

axelboc:
After much staring into space trying to figure out the best approach, here is what I came up with:

- The `ProviderAPI` contract that providers must implement now includes a `getGroup` method, which will eventually replace `getMetadata`. `getGroup` accepts a string `path` and must return a `Group` entity. As an example, `getGroup('/')` must return the root `Group`. 
- `Provider` now includes a second suspense store called `groupsStore` that wraps the new `getGroup` method. In other words, **we now have a cache that stores groups against their paths**. `groupsStore` is available through context, which means that we can call `groupsStore.get('/some/path')` to either get the metadata of the group at `/some/path` from the cache, or suspend rendering while the group's metadata is being fetched for the first time.

As a first refactoring increment, I chose to use the new `groupsStore` in the `Explorer` only. The `Provider` still fetches the entire `metadata` tree on initial load, but the `Explorer` no longer relies on this object.

---

Here is how the Explorer UI behaves now:

- On page load, ignoring the "legacy" loading of the whole metadata tree, the first thing we see is the Explorer sidebar with the domain button (since the domain is known, this is instantaneous). Then, the root group's `EntityList` renders for the first time, triggering a call to get the group's metadata (i.e. `groupsStore.get('/')`). Since the root group hasn't been fetched yet, this call suspends the rendering of `EntityList`, falling back to a spinner positioned on top of the domain button. Once the root group is fetched, `EntityList` renders again, this time to completion, and the root group's direct children appear.
    
    | | |
    |---|---|
    | image | image |

- When expanding a nested group, the group's `EntityList` renders, triggering a call to fetch the group's metadata using its path (i.e. constructed by prefixing the group's name with the parent group's path). Again, the rendering suspends, revealing a spinner, until the group is fetched and its children are finally rendered.

    | | |
    |---|---|
    | image | image |

- If a group is collapsed and then expanded again, `groupsStore('/group/path')` is called again, but this time, since the group's metadata is already in the cache, its children are rendered right away and the spinner doesn't appear at all.

---

In the Explorer's code, the big change is that the core state is no longer a `selectedEntity` object, but a `selectedPath` string. This is really convenient, because it makes selecting the `DEFAULT_PATH` trivial: `useState(DEFAULT_PATH)`. When an entity is selected, we notify the `App`, which turns the path into an `Entity` object with `getEntityAtPath`. This works fine for now because we still fetch the entire `metadata` tree, but it will need to be refactored.

Since we now consider paths to be unique, we can also simplify the management of expanded groups. Instead of managing a complex state in `Explorer`, and resorting to a `useEffect` to expand the correct groups when a `DEFAULT_PATH` is present, we can now switch to a simple boolean state in a new component called `EntityItem`. If the default path begins with the entity's path, the group can be expanded simply by initialising the boolean state to `true`.

### DIFF
--- a/src/h5web/explorer/EntityList.tsx
+++ b/src/h5web/explorer/EntityList.tsx
@@ -1,27 +1,17 @@
-import {
-  CSSProperties,
-  ReactElement,
-  Suspense,
-  useContext,
-  useEffect,
-} from 'react';
+import { ReactElement, useContext } from 'react';
 import styles from './Explorer.module.css';
-import type { Entity } from '../providers/models';
-import Icon from './Icon';
-import { isGroup } from '../guards';
 import { ProviderContext } from '../providers/context';
-import { FiRefreshCw } from 'react-icons/fi';
+import EntityItem from './EntityItem';
 
 interface Props {
   level: number;
   parentPath: string;
   selectedPath: string;
-  expandedGroups: Set<string>;
-  onSelect: (entity: Entity, path: string) => void;
+  onSelect: (path: string) => void;
 }
 
 function EntityList(props: Props): ReactElement {
-  const { level, parentPath, selectedPath, expandedGroups, onSelect } = props;
+  const { level, parentPath, selectedPath, onSelect } = props;
 
   const { groupsStore } = useContext(ProviderContext);
   const entities = groupsStore.get(parentPath).children;
@@ -35,45 +25,16 @@ function EntityList(props: Props): ReactElement {
       {entities.map((entity) => {
         const { uid, name } = entity;
         const path = `${parentPath === '/' ? '' : parentPath}/${name}`;
-        const isExpanded = expandedGroups.has(entity.uid);
 
         return (
-          <li
+          <EntityItem
             key={uid}
-            className={styles.entity}
-            style={{ '--level': level } as CSSProperties}
-            role="none"
-          >
-            <button
-              className={styles.btn}
-              type="button"
-              role="treeitem"
-              aria-expanded={isGroup(entity) ? isExpanded : undefined}
-              aria-selected={path === selectedPath}
-              onClick={() => onSelect(entity, path)}
-            >
-              <Icon entity={entity} isExpanded={isExpanded} />
-              {name}
-            </button>
-
-            {isGroup(entity) && isExpanded && (
-              <Suspense
-                fallback={
-                  <FiRefreshCw className={styles.spinner}>
-                    Loading...
-                  </FiRefreshCw>
-                }
-              >
-                <EntityList
-                  level={level + 1}
-                  parentPath={path}
-                  selectedPath={selectedPath}
-                  expandedGroups={expandedGroups}
-                  onSelect={onSelect}
-                />
-              </Suspense>
-            )}
-          </li>
+            path={path}
+            entity={entity}
+            level={level}
+            selectedPath={selectedPath}
+            onSelect={onSelect}
+          />
         );
       })}
     </ul>
--- a/src/h5web/explorer/EntityList.tsx
+++ b/src/h5web/explorer/EntityList.tsx
@@ -1,19 +1,30 @@
-import type { CSSProperties, ReactElement } from 'react';
+import {
+  CSSProperties,
+  ReactElement,
+  Suspense,
+  useContext,
+  useEffect,
+} from 'react';
 import styles from './Explorer.module.css';
 import type { Entity } from '../providers/models';
 import Icon from './Icon';
 import { isGroup } from '../guards';
+import { ProviderContext } from '../providers/context';
+import { FiRefreshCw } from 'react-icons/fi';
 
 interface Props {
   level: number;
-  entities: Entity[];
-  selectedEntity?: Entity;
+  parentPath: string;
+  selectedPath: string;
   expandedGroups: Set<string>;
-  onSelect: (entity: Entity) => void;
+  onSelect: (entity: Entity, path: string) => void;
 }
 
 function EntityList(props: Props): ReactElement {
-  const { level, entities, selectedEntity, expandedGroups, onSelect } = props;
+  const { level, parentPath, selectedPath, expandedGroups, onSelect } = props;
+
+  const { groupsStore } = useContext(ProviderContext);
+  const entities = groupsStore.get(parentPath).children;
 
   if (entities.length === 0) {
     return <></>;
@@ -23,11 +34,13 @@ function EntityList(props: Props): ReactElement {
     <ul className={styles.group} role="group">
       {entities.map((entity) => {
         const { uid, name } = entity;
+        const path = `${parentPath === '/' ? '' : parentPath}/${name}`;
         const isExpanded = expandedGroups.has(entity.uid);
 
         return (
           <li
             key={uid}
+            className={styles.entity}
             style={{ '--level': level } as CSSProperties}
             role="none"
           >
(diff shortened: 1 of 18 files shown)


src/h5web/explorer/EntityList.tsx
```diff
@@ -23,38 +24,17 @@ function EntityList(props: Props): ReactElement {
     <ul className={styles.group} role="group">
       {entities.map((entity) => {
         const { uid, name } = entity;
-        const isExpanded = expandedGroups.has(entity.uid);
+        const path = `${parentPath === '/' ? '' : parentPath}/${name}`;
```
axelboc, line 27: We construct the entity's path ourselves. This is the key to the whole thing, really, since at the start, the only path we know exists is `/`. We could require providers to include a path in a group's children's metadata, but this seems easier and more robust.

src/h5web/explorer/Explorer.module.css
```diff
@@ -45,3 +50,23 @@
   margin-left: -0.375rem;
   margin-right: 0.125rem;
 }
+
+.spinner {
+  position: absolute;
+  top: 0.5rem;
+  right: 0.75rem;
```
axelboc, line 57: This is the bit that's a little hacky: the spinner is the fallback of a `Suspense` element, which is rendered _after_ a domain/group button. It needs to be taken off the flow of the page and _moved_ above the button.

axelboc: When React's Concurrent mode finally comes out, we'll be able to use `useTransition` instead of `Suspense` as explained here: https://reactjs.org/docs/concurrent-mode-patterns.html#transitions

src/h5web/explorer/Explorer.module.css
```diff
@@ -45,3 +50,23 @@
...
+  font-size: 0.75em;
+  animation: spin 1.5s ease-in-out infinite, fade 0.1s 0.1s backwards;
```
axelboc, line 59: The fade-in animation allows the spinner to not appear when a group's metadata is fetched almost instantly, for instance when using the `MockProvider`.

src/h5web/explorer/Explorer.tsx
```diff
@@ -1,76 +1,55 @@
...
-  const { onSelect, selectedEntity } = props;
-  const { domain, metadata: root } = useContext(ProviderContext);
+  const { onSelect } = props;
 
-  const [
-    expandedGroups,
-    { add: expandGroup, toggle: toggleGroup },
-  ] = useSet<string>();
+  const { domain } = useContext(ProviderContext);
+  const [selectedPath, setSelectedPath] = useState<string>(DEFAULT_PATH);
```
axelboc, line 17: The new local state of the `Explorer`. As explained, the selection of the `DEFAULT_PATH` is now trivial.

src/h5web/explorer/Explorer.tsx
```diff
@@ -1,76 +1,55 @@
...
-      <EntityList
-        level={0}
-        entities={root.children}
-        selectedEntity={selectedEntity}
-        expandedGroups={expandedGroups}
-        onSelect={handleSelect}
-      />
+      <Suspense
+        fallback={
+          <FiRefreshCw className={styles.spinner}>Loading...</FiRefreshCw>
```
axelboc, line 43: The domain button's spinner.

src/h5web/providers/Provider.tsx
```diff
@@ -15,13 +15,18 @@ function Provider(props: Props): ReactElement {
 
   // Wait until metadata is fetched before rendering app
   const { value: metadata, error } = useAsync(async () => {
-    return api.fetchMetadata();
+    return api.getMetadata();
```
axelboc, line 18: Reverting some name changes I made in the previous PR.

src/h5web/providers/Provider.tsx
```diff
@@ -15,13 +15,18 @@ function Provider(props: Props): ReactElement {
...
 
-  const valuesStore = useMemo(
-    () => createFetchStore(api.fetchValue.bind(api)),
-    [api]
-  );
+  const groupsStore = useMemo(() => {
+    const store = createFetchStore(api.getGroup.bind(api));
+    store.prefetch('/'); // pre-fetch root group
+    return store;
+  }, [api]);
```
axelboc, line 25: The new `groupsStore` that wraps the `getGroup` method. Here we also prefetch the root group to shave off a few millisecond as per React's "fetch as you render" pattern.

src/h5web/providers/mock/MockProvider.tsx
```diff
@@ -1,22 +1,38 @@
...
         domain,
-        fetchMetadata: async () => mockMetadata,
-        fetchValue: async (id: keyof typeof mockValues) => {
+        getMetadata: async () => mockMetadata,
+        getGroup: async (path: string) => {
+          if (path === slowOnPath) {
+            await new Promise((resolve) => {
+              setTimeout(resolve, 3000);
+            });
+          }
```
axelboc, line 27: The `slowOnPath` allows simulating a slow request for a group's metadata. I'll use it in a test when I get to that.

src/h5web/providers/mock/MockProvider.tsx
```diff
@@ -1,22 +1,38 @@
...
+
+          const group = getEntityAtPath(mockMetadata, path, true);
```
axelboc, line 29: Note that we use `getEntityAtPath` here. We could go back to a flat structure for the mock metadata (with groups stored against their paths), but the nested structure is a lot more readable...

axelboc:
Alright, I've now implemented `getGroup(path)` in all providers.

- In Jupyter, this was fairly simple since the endpoints expect a path. I fetch the group's "metadata" and "contents" in parallel and create a `Group` object with the results.
- In HSDS, this was more challenging, since the endpoints expect IDs internal to HSDS. So to get a group from a path, you need to first find its ID in the links of its parent. But to get the parent, you also need to the parent's ID, so you need to recursively get all of the ancestors up to the root. From here, we can fetch the root group's ID with the dedicated endpoint and then the group itself and all the ancestors we went through until we're back to the group we're interested in. Along the way, we cache the groups that we fetch against their paths so the process is faster on subsequent calls to `getGroup`... 

Obviously, it would be much better if HSDS were to offer endpoints that expect paths, but this works fine for now.

Note also that for now, since we only use `getGroup` in the Explorer, the HSDS implementation really does the bare minimum: it doesn't fetch attributes, and it doesn't fetch or parse the full metadata of entities (dataset type/shape, etc.)

src/h5web/explorer/Explorer.tsx
```diff
@@ -1,76 +1,56 @@
...
-    if (isGroup(entityToSelect)) {
-      expandGroup(entityToSelect.uid);
-    }
-
-    // Expand parent groups
-    getParents(entityToSelect).forEach((group) => expandGroup(group.uid));
-  }, [expandGroup, onSelect, root]);
+  useMount(() => {
+    onSelect(DEFAULT_PATH);
+  });
```
loichuder, line 27: I would move this under the `useState(DEFAULT_PATH)` as the two go together.

src/h5web/providers/Provider.tsx
```diff
@@ -15,13 +15,18 @@ function Provider(props: Props): ReactElement {
...
+
+  const valuesStore = useMemo(() => {
+    return createFetchStore(api.getValue.bind(api));
+  }, [api]);
```
loichuder, line 29: Braces could be removed.

src/h5web/providers/hsds/utils.ts
```diff
@@ -1,5 +1,13 @@
...
 
 export function isHsdsExternalLink(link: HsdsLink): link is HsdsExternalLink {
   return 'h5domain' in link;
 }
+
+export const COLLECTION_TO_KIND: Record<HDF5Collection, EntityKind> = {
+  [HDF5Collection.Groups]: EntityKind.Group,
+  [HDF5Collection.Datasets]: EntityKind.Dataset,
+  [HDF5Collection.Datatypes]: EntityKind.Datatype,
+};
```
loichuder, line 13: A bit awkward but there are no way around it I guess

loichuder (commented):
Attributes seem to be ignored in the new `get..` functions. Is that normal or did I misunderstood ?

src/h5web/providers/Provider.tsx
axelboc: If I do, Prettier splits `[api]` onto multiple lines, which is really ugly 😑

src/h5web/providers/hsds/utils.ts
axelboc: I've extracted this into a function in the new PR, but yeah, the HSDS provider code overall is pretty bad. Hopefully the situation will improve from HSDS' end eventually.

src/h5web/explorer/Explorer.tsx
axelboc: It's gone in a subsequent PR.

PR #422, opened 2021-01-11
Refactor `getGroup` in HSDS and Jupyter provider APIs

axelboc:
Part of #417.

The `getGroup` methods now return fuller `Group` objects (with attributes, type/shape of child datasets, etc.) They also make mroe use of the existing methods and utilities.

### DIFF
--- a/src/h5web/providers/hsds/api.ts
+++ b/src/h5web/providers/hsds/api.ts
@@ -10,7 +10,7 @@ import type {
   HsdsAttributeWithValueResponse,
   HsdsLink,
 } from './models';
-import { Entity, EntityKind, Group, Metadata } from '../models';
+import type { Group, Metadata } from '../models';
 import {
   HDF5Collection,
   HDF5Dataset,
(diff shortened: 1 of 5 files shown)


src/h5web/providers/hsds/api.ts
```diff
@@ -199,14 +182,16 @@ export class HsdsApi implements ProviderAPI {
 
   /* Processing methods to fetch links and attributes in addition to the entity. Also updates API members. */
 
-  private async processGroup(id: HDF5Id): Promise<void> {
+  private async processGroup(id: HDF5Id, depth = Infinity): Promise<void> {
```
axelboc, line 185: Adding an argument to control the recursion depth.

src/h5web/providers/hsds/api.ts
```diff
@@ -57,51 +58,18 @@ export class HsdsApi implements ProviderAPI {
...
-    assertGroup(child);
-
-    const group: Group = {
-      ...child,
-      children: await this.getGroupChildren(child.id),
-    };
+    const groupLink = await this.getLinkTo(path);
+    await this.processGroup(groupLink.id, 1);
 
+    const group = buildGroup(await this.getHdf5Metadata(), groupLink);
```
loichuder, line 72: Quite clever to delegate the handling of non-resolvable links to `buildGroup`.

Unfortunately, it means that the transformation from a `HSDSLink` to a `Link` is no longer made (previously made in `getGroupChildren` :
```
  const link = isHsdsExternalLink(hsdsLink)
    ? { ...hsdsLink, file: hsdsLink.h5domain }
    : hsdsLink;
```
)

axelboc: This is now done by `processGroup`, like it always should have been. I wasn't calling it in my early prototype.

PR #423, opened 2021-01-12
Lift `selectedPath` state up into App and use in BreadcrumbsBar

axelboc:
Part of #417. The `BreadcrumbsBar` is now much simpler, since it now has direct access to the path. The path is now the core state of the `App`. It will be used to retrieve the associated `Entity` metadata.

### DIFF
--- a/src/h5web/BreadcrumbsBar.tsx
+++ b/src/h5web/BreadcrumbsBar.tsx
@@ -1,30 +1,34 @@
-import { Fragment, ReactElement } from 'react';
-import { FiSidebar, FiChevronRight } from 'react-icons/fi';
+import { Fragment, ReactElement, useContext } from 'react';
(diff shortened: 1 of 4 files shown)


src/h5web/BreadcrumbsBar.tsx
```diff
@@ -35,21 +39,20 @@ function BreadcrumbsBar(props: Props): ReactElement {
         value={isExplorerOpen}
         onChange={onToggleExplorer}
       />
-      {selectedEntity && (
-        <h1 className={styles.breadCrumbs}>
-          {getParents(selectedEntity)
```
axelboc, line 40: No need for `getParents` any more.

loichuder (commented):
Much smoother :+1:

PR #424, opened 2021-01-12
Replace `getGroup` with `getEntity` in provider API

axelboc:
Part of #417. Now that the `selectedPath` is the core local state of `App`, sub-components will need a way to retrieve the metadata object of any selected entity from its path, not only selected groups. Therefore, I had to reconsider my approach with the `getGroup` API method and `groupsStore` suspense store, and switch to a more generic `getEntity` method and `entitiesStore` suspense store.

From a back-end point of view, `getEntity` still returns the same thing when passed the path of a group (i.e. the metadata of the group and its direct children). However, it now also accepts paths to datasets and datatypes and should return their metadata.

To avoid fetching the same dataset/datatype twice (once when selecting its parent group in the Explorer, and a second time when actually using the entity in the app), I manage a sort of "proxy cache" inside the `entitiesStore`. When a group is fetched, I store its children datasets/datatypes in the cache so that, if a component tries to retrieve them from the store directly, they are returned instantly instead of being re-fetched with `getEntity`.

Most likely, the `Visualizer` will eventually include a call such as `entitiesStore.get(selectedPath)`. If `selectedPath` refers to a dataset, this dataset's metadata will only be fetched from the back-end if:

- the `Visualizer` is used in an app on its own, without the `Explorer` (otherwise, the `Explorer` would have already fetched the dataset's parent group ... and therefore the dataset itself);
- the `Visualizer` is rendered with a `selectedPath` refering to a dataset before the `Explorer` had a chance to fetch the dataset's parent group (for instance, when providing a `DEFAULT_PATH` to a dataset).

### DIFF
--- a/src/h5web/providers/Provider.tsx
+++ b/src/h5web/providers/Provider.tsx
@@ -4,6 +4,8 @@ import { createFetchStore } from 'react-suspense-fetch';
 import { ProviderAPI, ProviderContext } from './context';
 import ErrorMessage from '../visualizer/ErrorMessage';
 import styles from '../visualizer/Visualizer.module.css';
+import type { Entity } from './models';
+import { isGroup } from '../guards';
 
 interface Props {
   api: ProviderAPI;
@@ -18,8 +20,28 @@ function Provider(props: Props): ReactElement {
     return api.getMetadata();
   }, [api]);
 
-  const groupsStore = useMemo(() => {
-    const store = createFetchStore(api.getGroup.bind(api));
+  const entitiesStore = useMemo(() => {
+    const childCache = new Map<string, Entity>();
+
+    const store = createFetchStore(async (path: string) => {
+      if (childCache.has(path)) {
+        return childCache.get(path) as Entity;
+      }
+
+      const entity = await api.getEntity(path);
+
+      if (isGroup(entity)) {
+        // Cache non-group children (datasets, datatypes and links)
+        entity.children.forEach((child) => {
+          if (!isGroup(child)) {
+            childCache.set(path, entity);
+          }
+        });
+      }
+
+      return entity;
+    });
+
     store.prefetch('/'); // pre-fetch root group
     return store;
   }, [api]);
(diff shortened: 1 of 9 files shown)


src/h5web/providers/Provider.tsx
```diff
@@ -18,8 +20,28 @@ function Provider(props: Props): ReactElement {
...
+        // Cache non-group children (datasets, datatypes and links)
+        entity.children.forEach((child) => {
+          if (!isGroup(child)) {
+            childCache.set(path, entity);
+          }
+        });
+      }
+
+      return entity;
+    });
```
loichuder, line 43: Even if in the end there is a single `entitiesStore`, to build it, you maintain two separate cache: `childCache` and the one from `store`.

Do you think it might be possible to use the `store` API to add directly the children to it and avoid keeping a separate cache ?

axelboc: The `store` API is pretty limited: https://github.com/dai-shi/react-suspense-fetch/blob/master/src/index.ts#L12-L16 -- it doesn't provide access to its internal cache.

src/h5web/providers/mock/MockProvider.tsx
```diff
@@ -19,18 +19,17 @@ function MockProvider(props: Props): ReactElement {
...
           if (path === slowOnPath) {
             await new Promise((resolve) => {
               setTimeout(resolve, 3000);
             });
           }
 
-          const group = getEntityAtPath(mockMetadata, path, true);
-          assertDefined(group);
-          assertGroup(group);
+          const entity = getEntityAtPath(mockMetadata, path, true);
```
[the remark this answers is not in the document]
axelboc: Ah right! 🎣

I'm working on replacing calls to `getEntityPath` to fetch the default `NXdata` group. Once this is done, `getEntityPath` will only be used in the mock and silx providers, with the root metadata object, so we'll be able to simplify its internal behaviour and remove the `allowSelf` flag.

PR #425, opened 2021-01-14
Add path to `Entity` model and remove `getParents` util

axelboc:


src/h5web/BreadcrumbsBar.tsx
```diff
@@ -15,19 +15,19 @@ interface Props {
...
   } = props;
 
-  if (!entityPath.startsWith('/')) {
+  if (!path.startsWith('/')) {
     throw new Error("Expected path to start with '/'");
   }
 
   const { domain } = useContext(ProviderContext);
-  const crumbs = `${domain}${entityPath === '/' ? '' : entityPath}`.split('/');
+  const crumbs = [domain, ...`${path === '/' ? '' : path}`.split('/').slice(1)];
```
axelboc, line 30: This was buggy when the domain contained slashes, like with HSDS.

loichuder (changes requested):
After a quick overview, I think using `alias` leads to a lot of unneeded complexity. 

You indeed get this "for free" for the Silx Provider but it means you have to construct a path in the HSDS Provider and force it into a `alias` field. Also, the fact that it is a `[string]` rather than a simple `string` leads to a lot of ugly `alias[0]`.

I would instead argue that `alias` should be directly replaced by the `path` to the entity. 
Each provider should then be responsible to build this path. For Silx, simply unpack `alias`. For HSDS, you already build the path so just put it as a field (and not force it in a `alias`). Parsing the path in the `build...` functions should then be clearer.

axelboc:
I totally agree, `alias` is just a stepping stone. My goal in this PR is to introduce the path on the `Entity` object so I can use it in the `MetadataViewer`. `alias` just felt like the simplest way to go for now.

Once I no longer need `getMetadata` in the provider API, `buildTree` will only be called by the Silx provider and no longer by the HSDS provider, which means I'll be able to remove it completely and make the Silx provider's `getEntity` method smarter instead (notably by converting `alias` to `path`).

The HSDS provider will eventually only call `buildEntity`, so my goal is also to refactor this utility so it accepts a single HDF5 entity instead of the entire `HDF5Metadata` tree. At this point, it will be trivial to replace `alias` with `path` in the models.

So all in all, `alias` is really convenient for now, but it won't stay long, don't worry 😁

axelboc:
Now that I think about it, another, faster option would be to completely remove the `SilxProvider` right now.

At the end of this refactoring, we'll have two working and relatively-efficient providers (since they will no longer be fetching the entire metadata tree on start-up). I think this makes the `SilxProvider` a lot less relevant. Especially since I don't see the long-term benefit of supporting a provider that can serve the output of `hdf5-json`.

loichuder:
> Now that I think about it, another, faster option would be to completely remove the `SilxProvider` right now.

And that we already have a different way of providing mock data.

loichuder (approved):
We can merge as it is but allow me to list a few points I would like to be adressed later:
- The `alias` issue that I already mentioned.
- In HSDS, we need to add `path` to every `process...` function which hampers readability.
- In the mock provider, `isRoot` and `prefixChildrenPath` feel a bit clunky. However, this can be fixed separately as it is not related to the fetching strategy.

PR #426, opened 2021-01-14
Remove `SilxProvider` and make Jupyter demo the default

axelboc:
As discussed in https://github.com/silx-kit/h5web/pull/425#issuecomment-760146345, I remove the `SilxProvider` completely and make https://h5web.panosc.eu/ point to the `JupyterProvider` demo.

This removal allows for a huge clean-up in the HDF5 models, notably. `HDF5Metadata`, `HDF5Collection`, `HDF5Group/Dataset/Datatype` were shared between the Silx and HSDS providers, which is what was causing a lot of complexity. Now that Silx is gone, these models can be moved to `/providers/hsds/models`. Same with the `buildTree/Entity/...` utilities!

Another simplification opportunity comes from the fact that, unlike Silx, HSDS does not expose the concept of an `HDF5RootLink`. We can therefore remove this concept completely and leave only hard links as "reachable" links, which simplifies a lot of function signatures.

Note that this is only the _first wave_ of simplifications! The second wave (my next PR) will consist of getting rid of `HDF5Metadata` and the `buildTree/buildEntity/...` utilities.


src/h5web/guards.ts
```diff
@@ -88,11 +87,8 @@ export function isLink(entity: Entity): entity is Link {
   return entity.kind === EntityKind.Link;
 }
 
-export function isReachable(
-  link: HDF5Link
-): link is HDF5HardLink | HDF5RootLink {
-  // Only hard and root links are considered as reachable for now
-  return link.class === HDF5LinkClass.Hard || link.class === HDF5LinkClass.Root;
+export function isHardLink(link: HDF5Link): link is HDF5HardLink {
+  return link.class === HDF5LinkClass.Hard;
```
axelboc, line 91: With `HDF5RootLink` gone, the concept of a "reachable link" feels less relevant, but happy to revert the naming.

src/h5web/providers/hdf5-models.ts
```diff
@@ -12,37 +12,6 @@ export enum HDF5Collection {
...
-  alias: [string];
-  attributes?: HDF5Attribute[];
-  shape: S;
-  type: T;
-}
-
-export interface HDF5Datatype<T = HDF5Type> {
-  alias: [string];
-  type: T;
-}
```
axelboc, line 44: I move all these now-HSDS-specific models to `providers/hsds/models`. I also rename them with the `Hsds` prefix and remove the generics. We can now see them almost as the response models of HSDS' endpoints (it will be the case even more after the second wave of changes).

src/h5web/providers/hsds/api.ts
```diff
@@ -204,15 +202,15 @@ export class HsdsApi implements ProviderAPI {
       );
 
     this.groups[id] = {
-      alias: [path || '/'], // path
+      path: path || '/',
```
axelboc, line 205: No more `alias`! 😁

src/h5web/visualizations/nexus/utils.ts
```diff
@@ -20,7 +18,7 @@ import { isScaleType } from '../shared/utils';
 import { getChildEntity, getEntityAtPath } from '../../utils';
 
 export function getAttributeValue(
-  entity: HDF5Dataset | HDF5Group | Entity,
+  entity: Entity,
```
axelboc, line 21: 🤷‍♂️

README.md
```diff
@@ -33,23 +33,15 @@ providers out of the box, which are both under active development:
...
 
-The demo at https://h5web.panosc.eu/ shows the content of an HDF5 file converted
-to JSON with [hdf5-json](https://github.com/HDFGroup/hdf5-json) and hosted on a
-remote static server.
-
-At https://h5web.panosc.eu/mock, you can view a set of mock data generated
-entirely on the front-end. This demo and its provider, `MockProvider` are used
-for end-to-end testing purposes.
-
 ### [Jupyter Lab HDF5 extension](https://github.com/jupyterlab/jupyterlab-hdf5)
```
loichuder, line 36: I think the title should be reworded as one may think it is a JLab extension and not a stand-alone app.

src/h5web/guards.ts
loichuder: I like the simplification. Let's keep `isHardLink` :+1:

src/h5web/guards.ts
```diff
@@ -145,10 +141,8 @@ export function assertGroup(
   }
 }
 
-export function assertReachable(
-  link: HDF5Link
-): asserts link is HDF5HardLink | HDF5RootLink {
-  if (!isReachable(link)) {
+export function assertHardLink(link: HDF5Link): asserts link is HDF5HardLink {
+  if (!isHardLink(link)) {
     throw new Error('Expected link to be reachable');
```
loichuder, line 146:
```suggestion
    throw new Error('Expected link to be hard');
```

src/h5web/providers/hsds/models.ts
```diff
@@ -38,11 +39,7 @@ export interface HsdsExternalLink extends Omit<HDF5ExternalLink, 'file'> {
   h5domain: string;
 }
 
-export type HsdsLink =
-  | HDF5RootLink
-  | HDF5HardLink
-  | HDF5SoftLink
-  | HsdsExternalLink;
+export type HsdsLink = HDF5HardLink | HDF5SoftLink | HsdsExternalLink;
```
loichuder, line 42: Could `HDF5ExternalLink` have the same signature as `HsdsExternalLink` ?

This way, there will no longer be a distinction between `HsdsLink` and `HDF5Link`.

loichuder (approved):
Ride the wave :ocean: !

src/h5web/providers/hsds/models.ts
axelboc: Not sure how. One has a `file` key; the other has an `h5domain` key. 🤔

README.md
axelboc: I've added a little intro, does it clarify things?

src/h5web/providers/hsds/models.ts
loichuder: Yes but is the model with the `file` key needed elsewhere ? or could we use the `h5domain` key all the way ?

README.md
loichuder: It does :+1:

src/h5web/guards.ts
```diff
@@ -145,11 +141,9 @@ export function assertGroup(
...
 }
 
-export function assertReachable(
-  link: HDF5Link
-): asserts link is HDF5HardLink | HDF5RootLink {
-  if (!isReachable(link)) {
-    throw new Error('Expected link to be reachable');
+export function assertHardLink(link: HDF5Link): asserts link is HDF5HardLink {
+  if (!isHardLink(link)) {
+    throw new Error('Expected link to be hard link');
```
loichuder, line 146: Or simply _Expected a hard link_ ?

axelboc: This function will be gone in my next PR 😉

src/h5web/providers/hsds/models.ts
axelboc: Well, just from the name, `h5domain` seems very specific to HSDS. The HDF5 spec uses the term "file": http://davis.lbl.gov/Manuals/HDF5-1.8.7/UG/09_Groups.html

loichuder: Ok nevermind, let's keep things as they are then

PR #427, opened 2021-01-15
Refactor HSDS provider to remove the need for `buildTree` utilities

axelboc:
Part of #417. Wave 2 of 2 of removing the `SilxProvider` and cleaning up the HSDS provider.


src/h5web/providers/hsds/api.ts
```diff
@@ -178,83 +143,121 @@ export class HsdsApi implements ProviderAPI {
     return Promise.all(attrsPromises);
   }
 
-  /* Processing methods to fetch links and attributes in addition to the entity. Also updates API members. */
-
   private async processGroup(
```
axelboc, line 146: All the `process` methods now return `Entity` objects (and objects that inherit `Entity`) directly. We don't go through intermediary `HDF5Dataset/Datatype/Group` objects any more and the `buildTree/Entity/...` utilities are obsolete!

axelboc (commented):
image

src/h5web/providers/hsds/api.ts
```diff
@@ -178,83 +143,121 @@ export class HsdsApi implements ProviderAPI {
...
     id: HDF5Id,
-    path = '',
-    depth = Infinity
-  ): Promise<void> {
+    path: string,
+    name: string,
+    depth = 1
```
loichuder, line 150: I think `depth` should not have a default value. This way, we know exactly at each call how deep the processing will go.

loichuder (approved):
Go with the flow :ocean: :man_dancing:

PR #428, opened 2021-01-15
Check for absolute paths and facilitate their creation

axelboc:
Part of #417

### DIFF
--- a/src/h5web/App.tsx
+++ b/src/h5web/App.tsx
@@ -7,8 +7,10 @@ import BreadcrumbsBar from './BreadcrumbsBar';
 import Visualizer from './visualizer/Visualizer';
 import { getEntityAtPath } from './utils';
 import { ProviderContext } from './providers/context';
+import { assertAbsolutePath } from './guards';
 
 const DEFAULT_PATH = process.env.REACT_APP_DEFAULT_PATH || '/';
+assertAbsolutePath(DEFAULT_PATH);
 
 function App(): ReactElement {
   const [selectedPath, setSelectedPath] = useState<string>(DEFAULT_PATH);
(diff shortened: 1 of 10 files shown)


src/h5web/App.tsx
```diff
@@ -7,8 +7,10 @@ import BreadcrumbsBar from './BreadcrumbsBar';
 import Visualizer from './visualizer/Visualizer';
 import { getEntityAtPath } from './utils';
 import { ProviderContext } from './providers/context';
+import { assertAbsolutePath } from './guards';
 
 const DEFAULT_PATH = process.env.REACT_APP_DEFAULT_PATH || '/';
+assertAbsolutePath(DEFAULT_PATH);
```
axelboc, line 13: Almost everything now works with absolute paths, so we need to check for them from time to time.

I had a go at using TypeScript's new template literal types (e.g. `type AbsolutePath = `/${string}``, but inference was pretty poor. For instance, `const path: AbsolutePath = `/${entityName}` would not work. I had to use `as AbsolutePath` everywhere, which was really ugly. So I kept the `string` type throughout for paths, and used simple runtime assertions where relevant.

src/h5web/explorer/EntityList.tsx
```diff
@@ -26,12 +27,11 @@ function EntityList(props: Props): ReactElement {
     <ul className={styles.group} role="group">
       {group.children.map((entity) => {
         const { uid, name } = entity;
-        const path = `${parentPath === '/' ? '' : parentPath}/${name}`;
 
         return (
           <EntityItem
             key={uid}
-            path={path}
+            path={buildEntityPath(parentPath, name)}
```
axelboc, line 34: Adding this little utility to help construct absolute paths. This will also be useful in my next PR for creating an absolute path from a relative NeXUS `defaultPath`.

src/h5web/providers/context.ts
```diff
@@ -13,6 +13,6 @@ export abstract class ProviderAPI {
 export const ProviderContext = createContext<{
   domain: string;
   metadata: Metadata;
-  entitiesStore: FetchStore<Entity, HDF5Id>;
+  entitiesStore: FetchStore<Entity, string>;
```
axelboc, line 16: Woops

loichuder (approved):

PR #430, opened 2021-01-19
Lazy-load entities in MetadataViewer and Visualizer

axelboc:
The whole metadata tree is no longer fetched on initial load! I remove `getMetadata` from the API and I use the `entitiesStore` in both the `Visualizer` and the `MetadataViewer`. There are still some fetching optimisations to be done here and there, but it works and it's pretty damn fast! I'll do a demo in half an hour.


src/h5web/metadata-viewer/AttributesTable.tsx
```diff
@@ -17,14 +17,16 @@ function AttributesTable(props: Props): ReactElement {
...
+            Attributes
+          </th>
         </tr>
       </thead>
       <tbody>
         {attributes.map(
           ({ name, value }: HDF5Attribute): ReactElement => (
             <tr key={name}>
-              <th>{name}</th>
+              <th scope="row">{name}</th>
```
axelboc, line 29: I had async troubles in the tests. This fix was not the issue after all, but since it doesn't hurt, I left it... :D

src/h5web/metadata-viewer/MetadataViewer.test.tsx
```diff
@@ -31,7 +31,7 @@ describe('MetadataViewer', () => {
     fireEvent.click(await screen.findByRole('tab', { name: 'Inspect' }));
     await selectExplorerNode('entities');
 
-    const column = screen.getByRole('columnheader', { name: /Group/u });
+    const column = await screen.findByRole('columnheader', { name: /Group/u });
```
axelboc, line 34: The entity's metadata is not longer available to the `MetadataViewer` instantly.

src/h5web/metadata-viewer/MetadataViewer.tsx
```diff
@@ -1,23 +1,18 @@
...
-  const { entity } = props;
+  const { path } = props;
 
-  if (!entity) {
-    return (
-      <div className={styles.empty}>
-        <p>No entity selected.</p>
-      </div>
-    );
-  }
```
axelboc, line 20: We can remove a bunch of `undefined` fallbacks like this one, and make a few props mandatory.

src/h5web/providers/Provider.tsx
```diff
@@ -15,11 +12,6 @@ interface Props {
 function Provider(props: Props): ReactElement {
   const { api, children } = props;
 
-  // Wait until metadata is fetched before rendering app
-  const { value: metadata, error } = useAsync(async () => {
-    return api.getMetadata();
-  }, [api]);
-
```
axelboc, line 22: We no longer fetch the whole metadata tree, so the provider no longer blocks the rendering of the app.

src/h5web/providers/hsds/api.ts
```diff
@@ -180,12 +168,6 @@ export class HsdsApi implements ProviderAPI {
       attributes,
       children,
     };
-
-    children.forEach((c) => {
-      c.parent = group;
-    });
```
axelboc, line 186: We no longer need to find the root group from any entity, so we no longer have to keep parent references.

src/h5web/providers/jupyter/api.ts
```diff
@@ -13,14 +13,13 @@ import type {
   JupyterMetaResponse,
 } from './models';
 import { nanoid } from 'nanoid';
-import { makeStrAttr, floatType } from '../mock/utils';
+import { makeStrAttr, floatType } from '../mock/data-utils';
```
axelboc, line 16: The mock data generation functions and mock types are moved to `data-utils` in order to have a file called `utils` for utilities like `findMockEntity` (previously `getEntityAtPath`. We have to have two separate files because of cyclic dependencies, but I think keeping the mock data generation utilities separate is best anyway.

src/h5web/providers/mock/MockProvider.tsx
```diff
@@ -18,18 +17,14 @@ function MockProvider(props: Props): ReactElement {
...
             await new Promise((resolve) => {
               setTimeout(resolve, 3000);
             });
           }
 
-          const entity = getEntityAtPath(mockMetadata, path);
-          assertDefined(entity);
-
-          return entity;
+          return findMockEntity(mockMetadata, path);
```
axelboc, line 27: Note that I moved `assertDefined` inside `findMockEntity`, which is why I renamed it from `get` to `find`.

src/h5web/providers/mock/data.ts
```diff
@@ -64,6 +55,9 @@ export const mockMetadata = makeNxGroup(mockDomain, 'NXroot', {
                 id: 'title_twoD',
               }),
             }),
+            makeNxGroup('absolute_default_path', 'NXentry', {
+              defaultPath: '/nexus_entry/nx_process/nx_data',
+            }),
```
axelboc, line 60: Since `supportsEntity` no longer checks for a default path, I had to remove a few unit tests. Most of them were testing things already covered by feature tests, but this was the exception, so I've added this group here as well as a corresponding feature test. Same goes with `default_not_string` below.

src/h5web/providers/mock/data.ts
```diff
@@ -161,18 +158,3 @@ export const mockValues = {
...
-  assertDefined(dataset, `Expected entity at path "${path}"`);
-  assertDataset(dataset, `Expected group at path "${path}"`);
-  assertNumericType(dataset);
-  assertSimpleShape(dataset);
-
-  const value = mockValues[dataset.id as keyof typeof mockValues];
-  assertArray<number>(value);
-
-  return ndarray(value.flat(Infinity), dataset.shape.dims);
-}
```
axelboc, line 178: Moved to `utils`.

src/h5web/providers/mock/utils.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        /path to start with '\/'/u
+      );
+    });
+
+    it('should throw if no entity is found at given path', () => {
+      expect(() => findMockEntity(rootGroup, '/child1/foo')).toThrow(
+        /entity at path/u
+      );
+    });
+  });
```
axelboc, line 28: I converted and simplified the old tests of `getEntityAtPath`, since `findMockEntity` now only expects the root group and an absolute path.

src/h5web/utils.test.ts
```diff
@@ -22,43 +22,18 @@ describe('Global utilities', () => {
...
-      expect(getEntityAtPath(rootGroup, '', false)).toBeUndefined();
-      expect(getEntityAtPath(rootGroup, '/', false)).toBeUndefined();
-      expect(getEntityAtPath(childGroup1, '', false)).toBeUndefined();
-      expect(getEntityAtPath(childGroup1, '/', false)).toBe(rootGroup);
+    it('should build absolute path with relative path', () => {
+      expect(buildEntityPath('/', 'group/dataset')).toBe('/group/dataset');
+      expect(buildEntityPath('/foo', 'group/dataset')).toBe(
+        '/foo/group/dataset'
+      );
     });
```
axelboc, line 37: Adding some tests for `buildEntityPath`.

src/h5web/visualizations/containers/NxImageContainer.tsx
```diff
@@ -10,17 +9,14 @@ function NxImageContainer(props: VisContainerProps): ReactElement {
   const { entity } = props;
   assertGroup(entity);
 
-  const nxDataGroup = findNxDataGroup(entity);
-  assertDefined(nxDataGroup, 'Expected to find NXdata group');
```
axelboc, line 14: Here, the `entity` should already be an `NXdata` group! 🎉 We make sure of it with an assertion in `useNxData`.

src/h5web/visualizations/index.test.ts
```diff
@@ -198,48 +197,7 @@ describe('Visualization definitions', () => {
...
-            ],
-          }),
-        ],
-      });
-
-      expect(supportsEntity(group)).toBe(true);
-    });
-
-    it('should not support group with no `NXdata` class and no `default` attribute', () => {
+    it('should not support group with no `NXdata` class', () => {
```
axelboc, line 200: The `supportsEntity` tests no longer need to worry about default paths. I've removed/replaced them, as explained earlier.

src/h5web/visualizations/index.tsx
```diff
@@ -110,16 +110,11 @@ export const VIS_DEFS: Record<Vis, VisDef> = {
     Toolbar: NxSpectrumToolbar,
     Container: NxSpectrumContainer,
     supportsEntity: (entity) => {
-      if (!isGroup(entity)) {
+      if (!isGroup(entity) || !isNxDataGroup(entity)) {
```
axelboc, line 113: This is all that's needed now! Note that there's already a test for this use case.

src/h5web/visualizations/nexus/hooks.ts
```diff
@@ -18,6 +19,8 @@ import {
 import { getChildEntity } from '../../utils';
 
 export function useNxData(group: Group): NxData {
+  assertNxDataGroup(group);
```
axelboc, line 22: This is the assertion that replaces `findNxDataGroup` in the vis containers, as mentioned.

src/h5web/visualizer/Visualizer.test.tsx
```diff
@@ -91,6 +91,15 @@ describe('Visualizer', () => {
...
   });
 
+  test('show "NX Image" tab for NXentry group with absolute path to 2D default signal', async () => {
+    renderApp();
+    await selectExplorerNode('nexus_entry/nx_process/absolute_default_path');
+
+    const tabs = await findVisSelectorTabs();
+    expect(tabs).toHaveLength(1);
+    expect(tabs[0]).toHaveTextContent(Vis.NxImage);
+  });
```
axelboc, line 101: Testing an absolute default path. This replaces the old `supportsEntity` tests, as mentioned.

src/h5web/visualizer/Visualizer.test.tsx
```diff
@@ -155,5 +169,8 @@ describe('Visualizer', () => {
 
     await selectExplorerNode('signal_not_found');
     expect(await screen.findByText(/to exist/u)).toBeVisible();
+
+    expect(consoleMock).toHaveBeenCalledTimes(8); // React logs two stack traces per error
+    resetConsole();
```
axelboc, line 174: The errors thrown in `supportsEntity` are now caught by the `ErrorBoundary` in `App`, which means that errors are now logged to the console.

src/h5web/visualizer/Visualizer.tsx
```diff
@@ -1,22 +1,28 @@
...
 function Visualizer(props: Props): ReactElement {
-  const { entity } = props;
+  const { path } = props;
 
-  const { supportedVis, error } = getSupportedVis(entity);
+  const { entitiesStore } = useContext(ProviderContext);
+  const entity = entitiesStore.get(path);
+
+  // Resolve any `default` attribute(s) to find entity to visualize
+  const defaultEntity = getDefaultEntity(entity, entitiesStore);
```
axelboc, line 23: The line/function that does all the default path resolution work. After this we don't worry about default paths ever again! 😄

src/h5web/visualizer/Visualizer.tsx
```diff
@@ -1,22 +1,28 @@
...
+
+  const supportedVis = getSupportedVis(defaultEntity);
```
axelboc, line 25: `getSupportedVis` no longer catches/returns errors. It lets them propagate to the `ErrorBoundary`.

src/h5web/visualizer/utils.ts
```diff
@@ -11,24 +15,34 @@ function removeRedundantVis(visArr: Vis[]): Vis[] {
...
-      },
-      []
-    );
-
-    return { supportedVis: removeRedundantVis(supportedVis) };
-  } catch (error) {
-    return { supportedVis: [], error };
+export function getDefaultEntity(
+  entity: Entity,
+  entitiesStore: FetchStore<Entity, string>
```
axelboc, line 32: I couldn't make it a hook and use `useContext` inside because of the recursion (the recursive call comes after an `if`/`return`, which breaks a rule of hooks).

src/h5web/visualizations/containers/NxImageContainer.tsx
loichuder: I wonder if it might be clearer to move the assertion out of `useNxData`

loichuder (approved):
Looks much better :slightly_smiling_face:

src/h5web/metadata-viewer/MetadataViewer.tsx
axelboc: It cannot: it should either return an `Entity` or throw.

src/h5web/visualizations/containers/NxImageContainer.tsx
axelboc: We can't type check it, so it's more of a fail safe, and I thought it safer to have this fail safe inside `useNxData` so we don't forget the next time we add a NeXus visualization.

src/h5web/visualizer/Visualizer.test.tsx
axelboc: Well, I do: I test the presence of the error message on screen displayed by the `ErrorBoundary`.

axelboc: By the way, the great thing about the way we test is that all of these `expect(await screen.findByText(/to exist/u)).toBeVisible();` did not actually break even though we now display errors with an `ErrorBoundary` instead of rendering them ourselves!

src/h5web/visualizer/Visualizer.tsx
axelboc: lol 😁

src/h5web/visualizer/Visualizer.test.tsx
loichuder: Ah yes indeed sorry :slightly_frowning_face:

PR #431, opened 2021-01-19
Improve loading fallbacks

axelboc:
Once React Concurrent Mode and `useTransition` is officially released, this loading fallback will no longer be needed, since React will be capable of keeping what's on the screen while suspending a component in order to show a loading spinner in a good place instead of replacing everything with a fallback.

I'm just showing a bar that looks either like the `EntityTable` header bar, or like the `Visualizer`'s tab bar.

Screenshot 2021-01-19 141149
Screenshot 2021-01-19 141124


src/h5web/LoadingFallback.tsx
```diff
@@ -0,0 +1,27 @@
...
+interface Props {
+  isInspecting: boolean;
+  message?: string;
+}
+
+function LoadingFallback(props: Props): ReactElement {
+  const { isInspecting, message = 'Loading' } = props;
+
+  // Wait a bit before showing loader to avoid flash
+  const [isReady] = useTimeout(100);
```
axelboc, line 14: As for the value loader, I'm delaying showing the text `Loading...` by 100ms. I can't get rid of the flash until React Concurrent Mode is finally out, but I can make the flash less perceptible.

src/h5web/visualizer/ValueLoader.tsx
```diff
@@ -1,16 +1,16 @@
...
   message?: string;
 }
 
-function Loader(props: Props): ReactElement {
+function ValueLoader(props: Props): ReactElement {
   const { message = 'Loading' } = props;
 
   // Wait a bit before showing loader to avoid flash
-  const [isReady] = useTimeout(50);
+  const [isReady] = useTimeout(100);
```
axelboc, line 13: Slight increase from my personal experience.

loichuder (approved):

PR #433, opened 2021-01-20
Moved ScaleOption in its own file

loichuder:
Perhaps the root cause of https://github.com/silx-kit/jupyterlab-h5web/issues/1 ?


src/h5web/toolbar/controls/ScaleOption.module.css
```diff
@@ -0,0 +1,3 @@
+.icon {
+  composes: icon from '../Toolbar.module.css';
+}
```
axelboc, line 3: You could keep this in `ScaleSelector.module.css` and import that file in `ScaleOption.tsx`

axelboc (approved):

PR #436, opened 2021-01-22
Get dataset values by path rather than ID

axelboc:
The signature is now `getValue(path: string): HDF5Value`. This allows us to remove `id` from the `Entity` model. Since this ID is still needed by HSDS, I had to compensate by creating a couple of models specific to HSDS.

### DIFF
--- a/src/h5web/providers/hsds/utils.ts
+++ b/src/h5web/providers/hsds/utils.ts
@@ -1,5 +1,28 @@
-import type { HsdsLink, HsdsExternalLink } from './models';
+import { isDataset, isGroup } from '../../guards';
+import type { Entity } from '../models';
+import type {
+  HsdsLink,
+  HsdsExternalLink,
+  HsdsDataset,
+  HsdsGroup,
+} from './models';
 
 export function isHsdsExternalLink(link: HsdsLink): link is HsdsExternalLink {
   return 'h5domain' in link;
+}
+
+export function isHsdsGroup(entity: Entity): entity is HsdsGroup {
+  return isGroup(entity) && 'id' in entity;
+}
+
+export function isHsdsDataset(entity: Entity): entity is HsdsDataset {
+  return isDataset(entity) && 'id' in entity;
+}
+
+export function assertHsdsDataset(
+  entity: Entity
+): asserts entity is HsdsDataset {
+  if (!isHsdsDataset(entity)) {
+    throw new Error('Expected entity to be HSDS dataset');
+  }
 }
--- a/src/h5web/providers/mock/metadata-utils.ts
+++ b/src/h5web/providers/mock/metadata-utils.ts
@@ -1,13 +1,5 @@
 import { nanoid } from 'nanoid';
-import {
-  Dataset,
-  Datatype,
-  Entity,
-  EntityKind,
-  Group,
-  Link,
-  ResolvedEntity,
-} from '../models';
+import { Datatype, Entity, EntityKind, Group, Link } from '../models';
 import {
   HDF5Attribute,
   HDF5CompoundType,
@@ -30,6 +22,8 @@ import {
 import type { NxInterpretation, SilxStyle } from '../../vis-packs/nexus/models';
 import { isGroup } from '../../guards';
 import { buildEntityPath } from '../../utils';
+import type { MockDataset, MockValueId } from './models';
+import { mockValues } from './values';
 
 /* -------------------------- */
 /* ----- TYPES & SHAPES ----- */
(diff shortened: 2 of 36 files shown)


src/h5web/providers/mock/metadata-utils.ts
```diff
@@ -138,19 +129,19 @@ export function makeDataset<S extends HDF5Shape, T extends HDF5Type>(
...
   return {
     uid: nanoid(),
-    id,
     name,
     path: `/${name}`,
     kind: EntityKind.Dataset,
     attributes,
     shape,
     type,
+    value: mockValues[valueId as MockValueId],
```
axelboc, line 144: We read the dataset's value from `mockValues` on creation, based on `valueId`. If a `valueId` is not provided, the dataset's `name` is used.

src/h5web/providers/mock/metadata.ts
```diff
@@ -52,7 +48,7 @@ export const mockMetadata = makeNxGroup(mockDomain, 'NXroot', {
               signal: makeNxDataset('twoD', intType, [20, 41]),
               silxStyle: { signalScaleType: ScaleType.SymLog },
               title: makeDataset('title', scalarShape, stringType, {
-                id: 'title_twoD',
+                valueId: 'title_twoD',
```
axelboc, line 51: The connection with the mock values is a little clearer now.

src/h5web/providers/mock/metadata.ts
```diff
@@ -117,44 +113,3 @@ export const mockMetadata = makeNxGroup(mockDomain, 'NXroot', {
     ]),
   ],
 });
-
-/* ------------------ */
-/* ----- VALUES ----- */
```
axelboc, line 122: `makeDataset` needs access to `mockValues`, so I had to move the mock values to a separate file to avoid cyclic imports.

src/h5web/providers/hsds/utils.ts
```diff
@@ -1,5 +1,24 @@
...
 }
+
+export function isHsdsGroup(entity: Entity): entity is HsdsGroup {
+  return isGroup(entity);
+}
+
+export function assertHsdsDataset(
+  entity: Entity
+): asserts entity is HsdsDataset {
+  if (!isDataset(entity)) {
```
loichuder, line 21: Same here with a dedicated method `isHsdsDataset`

src/h5web/providers/mock/utils.ts
```diff
@@ -33,15 +38,18 @@ export function findMockEntity(root: Metadata, path: string): Entity {
   return entity;
 }
 
-export function getMockDataArray(path: string): ndarray {
-  assertAbsolutePath(path);
-
+export function findMockDataset(path: string): MockDataset {
```
loichuder, line 41: I wonder if the introduction of this function is really worth it. To me, it doesn't add much value.

src/stories/Home.stories.mdx
```diff
@@ -89,9 +89,10 @@ The library exposes a utility function to retrieve a mock entity's metadata and
 You can browse through the available mock data at https://h5web.panosc.eu/mock.
 
 ```ts
-import { findMockEntity, getMockDataArray } from '@h5web/lib';
+import { findMockEntity, findMockDataset, getMockDataArray } from '@h5web/lib';
 
 const entity = findMockEntity('/nD_datasets/twoD');
+const dataset = findMockDataset('/nD_datasets/twoD');
 const dataArray = getMockDataArray('/nD_datasets/twoD');
```
loichuder, line 96: One may be tempted to ask here why you need `findMockDataset` when you have `findMockEntity`. 

Perhaps elaborate a bit on how it can be used ?

loichuder (changes requested):

src/h5web/providers/mock/utils.ts
axelboc: You're right 👍

I've also added `isMockDataset` with a check on `value`.

src/h5web/providers/mock/utils.ts
```diff
@@ -34,14 +47,13 @@ export function findMockEntity(root: Metadata, path: string): Entity {
...
-
   const dataset = findMockEntity(mockMetadata, path);
-  assertDataset(dataset);
+  assertMockDataset(dataset);
+
   assertNumericType(dataset);
   assertSimpleShape(dataset);
 
-  const value = mockValues[dataset.id as keyof typeof mockValues];
+  const { value } = dataset as MockDataset;
```
loichuder, line 56: You could `assertMockDataset` here instead of casting.

loichuder (approved):

src/h5web/providers/mock/utils.ts
axelboc: If I do, I loose the type narrowing of `assertNumericType` and `assertSimpleShape` and can't access `dataset.shape.dims`. 😢

loichuder: :crying_cat_face:

axelboc: Although I could store the dims in a constant before `assertMockDataset`!

axelboc: Or better, the `value`.

PR #437, opened 2021-01-22
Stop generating UIDs for entities and remove nanoid

axelboc:


package.json
```diff
@@ -42,7 +42,6 @@
     "d3-scale": "^3.2.3",
     "d3-scale-chromatic": "^2.0.0",
     "lodash-es": "^4.17.20",
-    "nanoid": "^3.1.20",
```
axelboc, line 45: Not often we remove a library... 🎉 😄

src/h5web/explorer/EntityList.tsx
```diff
@@ -26,11 +26,11 @@ function EntityList(props: Props): ReactElement {
   return (
     <ul className={styles.group} role="group">
       {group.children.map((entity) => {
-        const { uid, name } = entity;
+        const { name } = entity;
 
         return (
           <EntityItem
-            key={uid}
+            key={name}
```
axelboc, line 33: Since paths are unique, entity names within a group are unique as well.

axelboc (commented):
Note that this could have been done sooner; it's not linked to the removal of `id` from `Entity`.

loichuder (approved):

PR #438, opened 2021-01-25
Upgrade dependencies

axelboc:
These upgrades bring some nice improvements!

- **zustand**'s `persist` middleware is now much more feature-complete, with a `whitelist` and even versioning, so I've swiched to it and removed our custom middleware. Note that `persist` doesn't work well with `combine` so I had to do some refactoring.
- **react-slider** has a small fix that allows removing a `position: absolute`.
- **cypress-image-snapshot** fixes a compatibility with Cypress 5 allowing us to re-enable `retries` (and reduce the snapshot delay) to avoid randomly-failing tests.
- A few React-based libraries are now compatible with React 17. Note that I've decided to upgrade **react-use** even though the typing issue with `useMeasure` is still unresolved -- I've used a bit of type casting to work around the issue.


src/h5web/vis-packs/core/heatmap/config.ts
```diff
@@ -1,62 +1,71 @@
...
+    {
+      name: 'h5web:heatmap',
+      whitelist: [
+        'colorMap',
+        'scaleType',
+        'keepAspectRatio',
+        'showGrid',
+        'autoScale',
+      ],
+      version: 1,
```
axelboc, line 68: We just need to bump this version number the next time we modify the state, and we won't run into those nasty bugs of deprecated persisted states.

loichuder (approved):

PR #439, opened 2021-01-26
Reorder controls on line/spectrum toolbars

axelboc:
I've moved the Line/Points/Both selector last, since I don't reckon people would use it very often. Also, I've moved the errors toggle before the grid toggle, as it seems unlikely that someone would want to turn off the grid. Both arguably subjective choices, but it also just looks nicer 😄 

I've also simplified some typings and fixed the spacing of the errors toggle's icon.

image


src/h5web/vis-packs/nexus/spectrum/NxSpectrumToolbar.tsx
```diff
@@ -0,0 +1,31 @@
...
+          value={showErrors}
+          onChange={toggleErrors}
+          disabled={areErrorsDisabled}
+        />
+      }
+    />
+  );
+}
+
+export default NxSpectrumToolbar;
```
axelboc, line 31: Oh I've also moved this toolbar component out of the `toolbar` folder and inside the Nexus vis pack.

src/h5web/toolbar/LineToolbar.tsx
```diff
@@ -4,12 +4,16 @@ import ToggleBtn from './controls/ToggleBtn';
...
 import ToggleGroup from './controls/ToggleGroup';
-import Toolbar, { ToolbarProps } from './Toolbar';
+import Toolbar, { ToolbarControl } from './Toolbar';
 import Separator from './Separator';
 import ScaleSelector from './controls/ScaleSelector';
 
-function LineToolbar(props: ToolbarProps): ReactElement {
-  const { children } = props;
+interface Props {
+  errorControl?: ToolbarControl;
```
loichuder, line 12: This is so specific that it bothers me a bit. We could have a prop `additionnalControls?: ToolbarControl[] = []` ?

On the other hand, it is only used in `NxSpectrumToolbar`...

cypress/specs/app.spec.ts
```diff
@@ -1,4 +1,4 @@
-const SNAPSHOT_DELAY = 100;
+const SNAPSHOT_DELAY = 500;
```
loichuder, line 1: Even with the retries ?

loichuder (approved):

src/h5web/toolbar/LineToolbar.tsx
axelboc: Yeah, an array like `children` or `additionalControls` was too restrictive yet also too vague. If we ever use it in other places, then maybe we'll want to put controls in other places... If it comes down to that, we'll just duplicate the `LineToolbar`... For now, I thought it was clearer like this.

cypress/specs/app.spec.ts
axelboc: The snapshot approval job was failing, because it would pass the first attempt regardless of whether it had properly rendered or not... I've been investigating a better solution, but I still can't find a reliable way to know if something a R3F component has actually rendered on the canvas or not...

PR #445, opened 2021-01-28
Implement new auto-scale behaviour for image stacks

### LINKED ISSUE #443: Color bar disappears
image

axelboc:
The "Auto-scale" toggle moves next to the domain slider and the slider's reset button is removed.

image

The `autoScale` boolean is removed from the Heatmap config state, but the `dataDomain` and `customDomain` remain. They are sufficient to manage the state for the new auto-scale behaviour described below. The `resetDomains` action is replaced with a `setDataDomain` action that no longer resets the `customDomain` to `undefined`.

The behaviour is now as follows:

- When the auto-scale toggle is pressed and the user slices through a stack of images, the `dataDomain` of each slice is computed and set in the Heatmap config with `setDataDomain` and the domain slider updates. `customDomain` remains `undefined` at all times, and `dataDomain` is passed down as `domain` to the visualization.
- When the user depresses the auto-scale toggle or moves one of the the domain slider's thumbs, the Heatmap config receives and sets a `customDomain` through `setCustomDomain`. `customDomain` is now passed down as `domain` to the visualization.
  - If the user slices through a stack of images, the `dataDomain` for the slice is not computed and not set in the Heatmap config, and the domain slider keeps displaying the custom domain.
  - The custom domain is kept intact when the user switches to another dataset or when they switch back and forth between visualizations. It only gets reset if the user presses the auto-scale toggle.

With this new behaviour, we no longer compute the domain of an entire image stack, ever:

- when auto-scale is off and a custom domain is set, we don't compute a domain at all;
- when auto-scale is on, we compute the domain only of the current layer/2D slice.

---

This PR also fixes #443 by falling back to a default domain in the `HeatmapVis`, just like we already do in the `LineVis`. I've also added this fallback to the new `setDataDomain` action, so that the domain slider doesn't disappear when the computed `dataDomain` is `undefined`.


src/h5web/toolbar/controls/DomainSlider.module.css
```diff
@@ -42,19 +42,9 @@
...
-
-.resetIcon {
-  padding-top: 1px;
-}
-
-.resetBtn:hover > .resetBtnLike {
-  background-color: var(--secondary-light);
+  font-size: 0.875em;
+  border-radius: 0.75rem;
+  color: #222;
```
axelboc, line 49: Reducing the visual impact of the `zap` icon a little.

src/h5web/toolbar/controls/DomainSlider.tsx
```diff
@@ -60,6 +51,18 @@ function DomainSlider(props: Props): ReactElement {
         step={step}
         pearling
       />
+
+      <ToggleBtn
+        className={styles.autoBtnLike}
+        label="auto-scale"
+        icon={FiZap}
```
axelboc, line 58: A button with a label of `auto` looked a bit ugly. I hope the `zap` icon is clear enough.

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -85,21 +85,17 @@ function HeatmapVis(props: Props): ReactElement {
...
-        )}
+        <Mesh
+          rows={rows}
+          cols={cols}
+          values={dataArray.data as number[]}
+          domain={domain}
+          scaleType={scaleType}
+          colorMap={colorMap}
+          showLoader={showLoader}
+        />
```
axelboc, line 96: The new domain fallback brings a slight performance improvement since we no longer unmount/remount the mesh.

src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -40,24 +38,16 @@ function MappedHeatmapVis(props: Props): ReactElement {
...
+  }, [customDomain, dataArray, scaleType]);
 
   useEffect(() => {
-    resetDomains(dataDomain); // in config, update `dataDomain` and reset `customDomain`
-  }, [dataDomain, resetDomains]);
+    if (!customDomain) {
+      // If auto-scale is on, update `dataDomain`
+      setDataDomain(domain);
+    }
+  }, [customDomain, domain, setDataDomain]);
```
axelboc, line 50: A LOT simpler logic overall 😂

The `useMemo` allows not computing the `dataDomain` if we don't need to.

src/h5web/toolbar/controls/ToggleBtn.tsx
```diff
@@ -12,7 +13,8 @@ interface Props {
 }
 
 function ToggleBtn(props: Props): ReactElement {
-  const { label, icon: Icon, iconOnly, value, onChange, disabled } = props;
+  const { className, label, iconOnly, value, onChange, disabled } = props;
+  const { icon: Icon } = props;
```
loichuder, line 17: A workaround to avoid multiple line breaks, I presume ?

src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
loichuder: Nice !

src/h5web/toolbar/controls/DomainSlider.tsx
loichuder: To me, it is not :joy: !

What about keeping the previous icon with the auto label:
image

I think `Auto` needs to be here for someone to understand the button's role

loichuder (changes requested):

src/h5web/toolbar/controls/ToggleBtn.tsx
axelboc: yes 😄 -- I hesitated...

src/h5web/toolbar/controls/DomainSlider.tsx
axelboc: Aww okay 😢 😂

axelboc: Not sure the previous icon makes much sense with just "auto" as the label, though. What do you think?

loichuder: Nah, we can remove the icon but I found the button a bit bare without an icon. Want to put `zap` instead :zap: ?

PR #450, opened 2021-02-02
Move dimension mappers up into containers

axelboc:
The dimension mapping state now comes before the calls to `useDatasetValue`, which means we'll be able to pass the current slice's "address" to the hook.


axelboc:
Woops, I forgot `MatrixVisContainer` - 2s...

loichuder (approved):
Nice !

As we discussed, the `Mapped` components do not add much value now and could be removed in the future.

PR #451, opened 2021-02-02
Infer type of dataset values from HDF5 type/shape instead of asserting

axelboc:
I'll probably be able to use the same approach to infer of types of attribute values.

---

Also, remove the need to pass `mockMetadata` to `findMockEntity` and remove `Metadata` model.

### DIFF
--- a/src/h5web/guards.ts
+++ b/src/h5web/guards.ts
@@ -174,6 +174,14 @@ export function assertSimpleShape<T extends HDF5Type>(
   }
 }
 
+export function assertBaseType<S extends HDF5Shape>(
+  dataset: Dataset<S>
+): asserts dataset is Dataset<S, HDF5BaseType> {
+  if (!hasStringType(dataset) && !hasNumericType(dataset)) {
+    throw new Error('Expected dataset to have string or numeric type');
+  }
+}
+
 export function assertStringType<S extends HDF5Shape>(
   dataset: Dataset<S>
 ): asserts dataset is Dataset<S, HDF5StringType> {
--- a/src/h5web/providers/mock/utils.ts
+++ b/src/h5web/providers/mock/utils.ts
@@ -9,7 +9,7 @@ import {
   isGroup,
 } from '../../guards';
 import { getChildEntity } from '../../utils';
-import type { Entity, Metadata } from '../models';
+import type { Entity } from '../models';
 import { mockMetadata } from './metadata';
 import type { MockDataset } from './models';
 
@@ -25,11 +25,11 @@ export function assertMockDataset(
   }
 }
 
-export function findMockEntity(root: Metadata, path: string): Entity {
+export function findMockEntity(path: string): Entity {
   assertAbsolutePath(path);
 
   if (path === '/') {
-    return root;
+    return mockMetadata;
   }
 
   const pathSegments = path.slice(1).split('/');
@@ -39,7 +39,7 @@ export function findMockEntity(root: Metadata, path: string): Entity {
         ? getChildEntity(parentEntity, currSegment)
         : undefined;
     },
-    root
+    mockMetadata
   );
 
   assertDefined(entity, `Expected entity at path "${path}"`);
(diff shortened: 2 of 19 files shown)


src/h5web/vis-packs/core/hooks.ts
```diff
@@ -8,24 +8,31 @@ import type { DimensionMapping } from '../../dimension-mapper/models';
...
+import type { Dataset, Value } from '../../providers/models';
 
-export function useDatasetValue(
-  path: string | undefined
-): HDF5Value | undefined {
+export function useDatasetValue<
+  S extends HDF5Shape,
+  T extends HDF5Type,
+  D extends Dataset<S, T> | undefined
+>(dataset: D): D extends Dataset<infer S, infer T> ? Value<S, T> : undefined;
```
axelboc, line 19: Same as with `useBaseArray` and `useDataArray`: the conditional return type allows inferring for instance `number[]` instead of `number[] | undefined` (which would have to be followed by `assertDefined` or `as number[]`).

The other way to write it would be to replace the override above with two slightly simpler, but repetitive, overrides:

```ts
export function useDatasetValue<S extends HDF5Shape, T extends HDF5Type>(
  dataset: Dataset<S, T>
): Value<S, T>;

export function useDatasetValue<S extends HDF5Shape, T extends HDF5Type>(
  dataset: Dataset<S, T> | undefined
): Value<S, T> | undefined;
```

src/h5web/vis-packs/core/raw/RawVis.tsx
```diff
@@ -1,9 +1,8 @@
 import type { ReactElement } from 'react';
 import styles from './RawVis.module.css';
-import type { HDF5Value } from '../../../providers/hdf5-models';
 
 interface Props {
-  value: HDF5Value;
+  value: unknown;
```
axelboc, line 5: I chose to use an explicit type here for consistency with the other visualizations.

src/h5web/guards.ts
```diff
@@ -174,6 +174,14 @@ export function assertSimpleShape<T extends HDF5Type>(
   }
 }
 
+export function assertBaseType<S extends HDF5Shape>(
+  dataset: Dataset<S>
+): asserts dataset is Dataset<S, HDF5BaseType> {
+  if (!hasStringType(dataset) && !hasNumericType(dataset)) {
+    throw new Error('Expected dataset to have numeric type');
```
loichuder, line 181:
```suggestion
    throw new Error('Expected dataset to have base type');
```

src/h5web/providers/mock/utils.ts
```diff
@@ -39,15 +39,15 @@ export function findMockEntity(root: Metadata, path: string): Entity {
...
+    mockMetadata
   );
 
   assertDefined(entity, `Expected entity at path "${path}"`);
   return entity;
 }
 
 export function getMockDataArray(path: string): ndarray {
-  const dataset = findMockEntity(mockMetadata, path);
+  const dataset = findMockEntity(path);
```
loichuder, line 50: Good clean up :+1:

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -8,24 +8,31 @@ import type { DimensionMapping } from '../../dimension-mapper/models';
...
+export function useDatasetValues<S extends HDF5Shape, T extends HDF5Type>(
+  datasets: Dataset<S, T>[]
+) {
   const { valuesStore } = useContext(ProviderContext);
 
   return Object.fromEntries(
-    datasets.map(({ name, path }) => [name, valuesStore.get(path)])
+    datasets.map(({ name, path }) => [
+      name,
+      valuesStore.get(path) as Value<S, T>,
```
loichuder, line 34: A shame but better than before

loichuder (approved):

src/h5web/vis-packs/core/hooks.ts
axelboc: Yeah, it was that or adding an override, which seemed overly complicated.

Issue #406, opened 2020-12-21
Fix type-inferring in the Jupyter provider


loichuder:
A basic working version of the JLab provider was implemented in #394. This version assumes all attributes are string and all datasets contains floats because of the following issues:

- [x] The type and shape of attributes cannot be inferred yet as the backend does not provide the necessary info.
- [x] The type of datasets is also tough to find as the `dtype` field is the NumPy type and not the HDF5 type that we use. We could infer the HDF5 type from the `dtype` on the front but I would prefer that the backend provides directly the HDF5 type.

This needs to be fixed once the backend is updated.

loichuder:
In the end, implementing HDF5 types in the backend was too complex and redundant with `dtypes`. Instead, we chose to simplify the HDF5Type in the backend (#452) and to infer the HDF5Type from the `dtype` (#459).

Note that the issue still holds for attributes.

loichuder:
> This needs to be fixed once the backend is updated.
 It can be fixed now: https://github.com/jupyterlab/jupyterlab-hdf5/issues/53

PR #452, opened 2021-02-03
Change HDF5Type by parsing `base` into object fields

loichuder:
A refactoring in preparation of #406.

`HDF5FloatType` and `HDF5IntegerType` no longer holds a `base` string but directly the information that can be deduced from it (endianness, number of bits, sign). This will allow a much easier handling in the `JupyterProvider` (for type inferring) and in the `MetadataViewer` (for displaying the HDF5Type of a dataset).

The `base` string is now parsed in the `HsdsProvider` to add the relevant fields to `HDF5Type`. Note that some advanced types that hold reference to other types need to be recursively parsed.

### DIFF
--- a/src/h5web/providers/hdf5-models.ts
+++ b/src/h5web/providers/hdf5-models.ts
@@ -101,42 +101,19 @@ export enum HDF5TypeClass {
   Compound = 'H5T_COMPOUND',
 }
 
+export type HDF5Endianness = 'BE' | 'LE';
+
 export interface HDF5IntegerType {
   class: HDF5TypeClass.Integer;
-  base:
-    | 'H5T_STD_I8BE'
-    | 'H5T_STD_I8LE'
-    | 'H5T_STD_I16BE'
-    | 'H5T_STD_I16LE'
-    | 'H5T_STD_I32BE'
-    | 'H5T_STD_I32LE'
-    | 'H5T_STD_I64BE'
-    | 'H5T_STD_I64LE'
-    | 'H5T_STD_U8BE'
-    | 'H5T_STD_U8LE'
-    | 'H5T_STD_U16BE'
-    | 'H5T_STD_U16LE'
-    | 'H5T_STD_U32BE'
-    | 'H5T_STD_U32LE'
-    | 'H5T_STD_U64BE'
-    | 'H5T_STD_U64LE'
-    | 'H5T_STD_B8BE'
-    | 'H5T_STD_B8LE'
-    | 'H5T_STD_B16BE'
-    | 'H5T_STD_B16LE'
-    | 'H5T_STD_B32BE'
-    | 'H5T_STD_B32LE'
-    | 'H5T_STD_B64BE'
-    | 'H5T_STD_B64LE';
+  size: number;
+  unsigned?: boolean;
+  endianness: HDF5Endianness;
 }
 
 export interface HDF5FloatType {
   class: HDF5TypeClass.Float;
-  base:
-    | 'H5T_IEEE_F32BE'
-    | 'H5T_IEEE_F32LE'
-    | 'H5T_IEEE_F64BE'
-    | 'H5T_IEEE_F64LE';
+  size: number;
+  endianness: HDF5Endianness;
 }
 
 export interface HDF5StringType {
--- a/src/h5web/providers/hsds/models.ts
+++ b/src/h5web/providers/hsds/models.ts
@@ -1,14 +1,48 @@
 import type {
   HDF5Id,
   HDF5Shape,
-  HDF5Type,
   HDF5Attribute,
   HDF5Value,
   HDF5ExternalLink,
   HDF5HardLink,
   HDF5SoftLink,
+  HDF5TypeClass,
+  HDF5StringType,
+  HDF5Dims,
 } from '../hdf5-models';
 import type { Dataset, Group } from '../models';
+
+export type HsdsBaseType =
+  | { class: HDF5TypeClass.Integer; base: string }
+  | { class: HDF5TypeClass.Float; base: string }
+  | HDF5StringType;
+
+export type HsdsType =
+  | HsdsBaseType
+  | HsdsArrayType
+  | HsdsVLenType
+  | HsdsCompoundType;
+
+interface HsdsArrayType {
+  class: HDF5TypeClass.Array;
+  base: HsdsBaseType;
+  dims: HDF5Dims;
+}
+
+interface HsdsVLenType {
+  class: HDF5TypeClass.VLen;
+  base: HsdsBaseType;
+}
+
+export interface HsdsCompoundType {
+  class: HDF5TypeClass.Compound;
+  fields: HsdsCompoundTypeField[];
+}
+
+interface HsdsCompoundTypeField {
+  name: string;
+  type: HsdsType;
+}
 
 export interface HsdsRootResponse {
   root: HDF5Id;
@@ -23,13 +57,13 @@ export interface HsdsGroupResponse {
 export interface HsdsDatasetResponse {
   id: HDF5Id;
   shape: HDF5Shape;
-  type: HDF5Type;
+  type: HsdsType;
   attributeCount: number;
 }
 
 export interface HsdsDatatypeResponse {
   id: HDF5Id;
-  type: HDF5Type;
+  type: HsdsType;
 }
 
 export interface HsdsAttributesResponse {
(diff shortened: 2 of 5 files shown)


src/h5web/providers/hdf5-models.ts
```diff
@@ -103,40 +103,20 @@ export enum HDF5TypeClass {
...
-    | 'H5T_IEEE_F32LE'
-    | 'H5T_IEEE_F64BE'
-    | 'H5T_IEEE_F64LE';
+  length: number;
+  endianness: Endianness;
+}
+
+export enum Endianness {
+  Big = 'Big',
+  Little = 'Little',
```
loichuder, line 119: `enum` is perhaps too much here...

src/h5web/providers/hsds/models.ts
```diff
@@ -1,15 +1,49 @@
...
+  HDF5TypeClass,
+  HDF5StringType,
+  HDF5Dims,
 } from '../hdf5-models';
 import type { Dataset, Group } from '../models';
 
+export type HsdsBaseType =
+  | { class: HDF5TypeClass.Integer; base: string }
+  | { class: HDF5TypeClass.Float; base: string }
+  | HDF5StringType;
```
loichuder, line 18: I left `HDF5StringType` unchanged for now.

src/h5web/providers/hsds/utils.ts
```diff
@@ -26,3 +35,55 @@ export function assertHsdsDataset(
...
+  }
+
+  const [, sign, length, endian] = matches;
+
+  return {
+    class: hsdsClass,
+    endianness: endian === 'L' ? Endianness.Little : Endianness.Big,
+    length: Number.parseInt(length, 10),
+    unsigned: sign === 'U' || undefined,
+  };
```
loichuder, line 60: I intend to add some tests for this once we wrap up this PR

src/h5web/providers/hsds/utils.ts
```diff
@@ -26,3 +35,55 @@ export function assertHsdsDataset(
...
+}
+
+export function processHsdsType(hsdsType: HsdsType): HDF5Type {
+  if (isString(hsdsType)) {
+    return hsdsType;
```
loichuder, line 65: If `hsdsType` is a ref to another type (`HDF5Id`)

src/h5web/providers/hsds/utils.ts
```diff
@@ -26,3 +35,55 @@ export function assertHsdsDataset(
...
+    hsdsType.class === HDF5TypeClass.Array ||
+    hsdsType.class === HDF5TypeClass.VLen
+  ) {
+    return {
+      ...hsdsType,
+      base: processHsdsBaseType(hsdsType.base),
+    };
+  }
+
+  if (hsdsType.class === HDF5TypeClass.Compound) {
```
loichuder, line 78: For compound types, the `type` needs to be processed for each field

src/h5web/providers/mock/metadata-utils.ts
```diff
@@ -29,12 +30,14 @@ import { mockValues } from './values';
...
-  base: 'H5T_STD_I32LE',
+  endianness: Endianness.Little,
+  length: 32,
 };
 
 export const floatType: HDF5FloatType = {
   class: HDF5TypeClass.Float,
-  base: 'H5T_IEEE_F64LE',
+  endianness: Endianness.Little,
+  length: 64,
```
loichuder, line 40: This improves a little the readability of mocked types

src/h5web/providers/hsds/utils.ts
```diff
@@ -26,3 +35,55 @@ export function assertHsdsDataset(
...
 }
+
+export function processHsdsBaseType(hsdsBaseType: HsdsBaseType): HDF5BaseType {
+  if (hsdsBaseType.class === HDF5TypeClass.String) {
+    return hsdsBaseType;
+  }
+
+  const { class: hsdsClass, base } = hsdsBaseType;
+
+  const regex = /H5T_[A-z]+_([A-z])(\d+)(B|L)E/u;
```
axelboc, line 46: - I don't think this works with all the types listed here: https://support.hdfgroup.org/HDF5/doc/RM/PredefDTypes.html -- it works with `H5T_STD` and `H5T_IEEE` for sure, so may as well be specific about it.
- The base is all uppercase, so the character ranges can be shortened.
- It might be simpler to just have `(BE|LE)` at the end and just store that directly on the `endianness` prop, since we don't care much about it. 🤷 

So in summary: `/H5T_(STD|IEEE)_([A-Z])(\d+)(BE|LE)/u`

src/h5web/providers/hsds/utils.ts
```diff
@@ -26,3 +35,55 @@ export function assertHsdsDataset(
...
+    return {
+      ...hsdsType,
+      fields: hsdsType.fields.map((v) => ({
+        name: v.name,
+        type: processHsdsType(v.type),
+      })),
+    };
+  }
+
+  return processHsdsBaseType(hsdsType);
```
axelboc, line 88: Wouldn't this cause infinite recursion?

src/h5web/providers/hsds/utils.ts
```diff
@@ -26,3 +35,55 @@ export function assertHsdsDataset(
...
+
+  if (hsdsType.class === HDF5TypeClass.Compound) {
+    return {
+      ...hsdsType,
+      fields: hsdsType.fields.map((v) => ({
+        name: v.name,
+        type: processHsdsType(v.type),
+      })),
+    };
+  }
```
axelboc, line 86: These `if`s may be better written as a `switch`. Also, with `...hsdsType`, you keep the `base` property in the returned objects, which is unnecessary -- you could destructure `hsdsType` to get the `class` and specify it explicitly in each returned object.

axelboc (approved):

src/h5web/providers/hsds/utils.ts
loichuder: Nay: `processHsdsType` (now `convertHsdsType`) calls `processHsdsBaseType`

src/h5web/providers/hdf5-models.ts
```diff
@@ -101,42 +101,19 @@ export enum HDF5TypeClass {
   Compound = 'H5T_COMPOUND',
 }
 
+export type Endianness = 'BE' | 'LE';
```
axelboc, line 104: `HDF5Endianness` ?

src/h5web/providers/hsds/utils.ts
```diff
@@ -26,3 +30,66 @@ export function assertHsdsDataset(
...
+
+  const regex = /H5T_(?:STD|IEEE)_([A-Z])(\d+)(BE|LE)/u;
+  const matches = regex.exec(base);
+
+  if (!matches) {
+    throw new Error(`Unrecognized base ${base}`);
+  }
+
+  const [, sign, size, endianness] = matches;
+  assertEndianness(endianness);
```
axelboc, line 49: It's a case where it makes sense to do `endianness: endianness as HDF5Endianness`, because it cannot be anything else (unless there were to be a bug in the regex, in which case we'd fix it)

src/h5web/providers/hsds/utils.ts
```diff
@@ -26,3 +30,66 @@ export function assertHsdsDataset(
...
+        class: HDF5TypeClass.Array,
+        base: convertHsdsBaseType(hsdsType.base),
+        dims: hsdsType.dims,
+      };
+
+    case HDF5TypeClass.VLen:
+      return {
+        class: HDF5TypeClass.VLen,
+        base: convertHsdsBaseType(hsdsType.base),
+      };
```
axelboc, line 76: You can have two `case` in a row:

```
case HDF5TypeClass.Array:
case HDF5TypeClass.VLen:
  return { class: hsdsType.class, ... }
```

axelboc (approved):

src/h5web/providers/hsds/utils.ts
loichuder: There is a `dims` field for `HDF5TypeClass.Array` and not for `HDF5TypeClass.VLen`, right ?

axelboc: yeah, but you had them both in the same `if` earlier, didn't you?

axelboc: nevermind if it's clearer like this, though

loichuder: > yeah, but you had them both in the same `if` earlier, didn't you?

Yes because I was doing `...hsdsType` but I also find it much clearer like this.

PR #453, opened 2021-02-03
Added tests for convertHsdsType

loichuder:
Follow up of #452

### DIFF
--- a/src/h5web/providers/hsds/models.ts
+++ b/src/h5web/providers/hsds/models.ts
@@ -21,15 +21,16 @@ export type HsdsType =
   | HsdsBaseType
   | HsdsArrayType
   | HsdsVLenType
-  | HsdsCompoundType;
+  | HsdsCompoundType
+  | HDF5Id;
 
-interface HsdsArrayType {
+export interface HsdsArrayType {
   class: HDF5TypeClass.Array;
   base: HsdsBaseType;
   dims: HDF5Dims;
 }
 
-interface HsdsVLenType {
+export interface HsdsVLenType {
   class: HDF5TypeClass.VLen;
   base: HsdsBaseType;
 }
--- a/src/h5web/providers/hsds/utils.test.ts
+++ b/src/h5web/providers/hsds/utils.test.ts
@@ -1,0 +1,152 @@
+import {
+  HDF5BaseType,
+  HDF5Id,
+  HDF5StringType,
+  HDF5TypeClass,
+} from '../hdf5-models';
+import type {
+  HsdsArrayType,
+  HsdsBaseType,
+  HsdsCompoundType,
+  HsdsType,
+  HsdsVLenType,
+} from './models';
+import { convertHsdsType } from './utils';
+
+describe('HsdsProvider utilities', () => {
+  describe('convertHsdsType', () => {
+    interface TestType {
+      hsds: HsdsBaseType;
+      hdf5: HDF5BaseType;
+    }
+
+    const leIntegerType: TestType = {
+      hsds: {
+        class: HDF5TypeClass.Integer,
+        base: 'H5T_STD_I8LE',
+      },
+      hdf5: {
+        class: HDF5TypeClass.Integer,
+        size: 8,
+        endianness: 'LE',
+      },
+    };
+
+    const beIntegerType: TestType = {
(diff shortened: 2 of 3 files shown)


src/h5web/providers/hsds/models.ts
```diff
@@ -21,15 +21,16 @@ export type HsdsType =
   | HsdsBaseType
   | HsdsArrayType
   | HsdsVLenType
-  | HsdsCompoundType;
+  | HsdsCompoundType
+  | HDF5Id;
```
loichuder, line 25: Oops, forgot in the interface that the type can be `HDF5Id`

src/h5web/providers/hsds/utils.test.ts
```diff
@@ -0,0 +1,179 @@
...
+          },
+        ],
+      });
+    });
+
+    it('should throw when encountering an unknown type', () => {
+      const unknownType = { class: 'NO_CLASS' };
+      expect(() => convertHsdsType(unknownType as HsdsType)).toThrow(
+        /Unknown type/u
+      );
```
loichuder, line 176: Even if the error cannot happen according to TypeScript, we need to test it as TS has no control on what comes out of the HSDS requests.

src/h5web/providers/hsds/utils.test.ts
```diff
@@ -0,0 +1,179 @@
...
+        base: {
+          class: HDF5TypeClass.Integer,
+          size: 8,
+          endianness: 'LE',
+          unsigned: undefined,
+        },
+      });
+    });
+
+    it('should convert the field types of Compound type', () => {
```
loichuder, line 118: The test might seem over-complicated but it makes sure that all types are converted (even deeply nested ones).

It cannot happen in `Array` in `VLen` types as the base is a `HsdsBaseType` in these cases.

src/h5web/providers/hsds/utils.test.ts
```diff
@@ -0,0 +1,172 @@
...
+        class: HDF5TypeClass.VLen,
+        base: integerType,
+      };
+      const compoundType: HsdsCompoundType = {
+        class: HDF5TypeClass.Compound,
+        fields: [
+          {
+            name: 'f1',
+            type: floatType,
+          },
```
axelboc, line 132: maybe shrink those onto single lines?

src/h5web/providers/hsds/utils.test.ts
```diff
@@ -0,0 +1,172 @@
...
+          {
+            name: 'f2',
+            type: {
+              class: HDF5TypeClass.VLen,
+              base: {
+                class: HDF5TypeClass.Integer,
+                size: 8,
+                endianness: 'LE',
+              },
```
axelboc, line 158: You could probably reuse some of the expected `base` types from previous tests by storing them in constants at the top, like we do in other test files.

src/h5web/providers/hsds/utils.ts
```diff
@@ -52,12 +52,19 @@ export function convertHsdsBaseType(hsdsBaseType: HsdsBaseType): HDF5BaseType {
...
-    unsigned: sign === 'U' || undefined,
   };
+  if (sign === 'U') {
+    return {
+      ...hdf5Type,
+      unsigned: true,
+    };
+  }
+
+  return hdf5Type;
```
axelboc, line 67: Always annoying this one...

```ts
return {
  // ...
  ...(sign === 'U' ? { unsigned: true } : {})
};
```

axelboc (approved):

src/h5web/providers/hsds/utils.ts
```diff
@@ -52,11 +52,15 @@ export function convertHsdsBaseType(hsdsBaseType: HsdsBaseType): HDF5BaseType {
...
     class: hsdsClass,
     endianness: endianness as HDF5Endianness,
     size: Number.parseInt(size, 10),
-    unsigned: sign === 'U' || undefined,
+  };
+
+  return {
+    ...hdf5Type,
+    ...(sign === 'U' ? { unsigned: true } : {}),
   };
```
axelboc, line 64: oh, no need for the separate return here 😉

axelboc (approved):

src/h5web/providers/hsds/utils.ts
loichuder: :facepalm: thanks

PR #455, opened 2021-02-04
Made crumbs clickable to navigate to previous groups

### LINKED ISSUE #454: Clickable breadcrumbs
I thought we already had an issue for that... Well, what do I know

While breadcrumbs work fine, a common user interaction is to click on a crumb to navigate to the crumb level.

Ex:
image

Here, **results** is selected. One might expect to select **entry_0000** to display/inspect this group by clicking on it.

loichuder:
Fix #454 

I tried to improve the readability of the breadcrumbs implementation by extracting it in a separate component.


src/h5web/Breadcrumbs.tsx
```diff
@@ -0,0 +1,68 @@
...
+  showDomain: boolean;
+}
+
+function Breadcrumbs(props: Props): ReactElement {
+  const { path, onSelect, showDomain } = props;
+
+  assertAbsolutePath(path);
+  const { domain } = useContext(ProviderContext);
+
+  if (path === '/') {
```
loichuder, line 19: Having `Breadcrumbs` in its own component allows for an early return if the root is selected. In this case, we don't need to generate crumbs or to check if the domain should be shown.

src/h5web/Breadcrumbs.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <button
+            className={styles.crumbButton}
+            type="button"
+            onClick={() => onSelect('/')}
+          >
+            <span className={styles.crumb}>{domain}</span>
+          </button>
+          <FiChevronRight className={styles.separator} title="/" />
+        </>
+      )}
```
loichuder, line 45: I found clearer to duplicate a bit of code here for the domain crumb rather than trying to manipulate the `crumbs` array

src/h5web/BreadcrumbsBar.tsx
```diff
@@ -39,18 +34,11 @@ function BreadcrumbsBar(props: Props): ReactElement {
...
-        ))}
-        <span className={styles.crumb} data-current>
-          {crumbs[crumbs.length - 1]}
-        </span>
-      </h1>
+      <Breadcrumbs
+        path={path}
+        onSelect={setSelectedPath}
+        showDomain={!isExplorerOpen}
+      />
```
axelboc, line 41: Great, nice abstraction. 👍

src/h5web/Breadcrumbs.tsx
```diff
@@ -0,0 +1,68 @@
...
+            >
+              <span className={styles.crumb}>{crumb}</span>
+            </button>
+            <FiChevronRight className={styles.separator} title="/" />
+          </Fragment>
+        );
+      })}
+      <span className={styles.crumb} data-current>
+        {crumbs[crumbs.length - 1]}
+      </span>
```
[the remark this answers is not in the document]
axelboc: Neh, don't think so.

src/h5web/Breadcrumbs.tsx
axelboc: Might as well create a `Crumb` component and move all three components into a `breadcrumbs-bar` folder, maybe?

axelboc (changes requested):

loichuder:
e2e tests fail probably because some padding/margin was added when adding the `button` tag. I will approve the screenshots once we agree on the implementation.

src/h5web/breadcrumbs/Crumb.tsx
```diff
@@ -0,0 +1,34 @@
...
+interface Props {
+  name: string;
+  onClick?: () => void;
+  isCurrent?: boolean;
+}
+
+function Crumb(props: Props): ReactElement {
+  const { name, onClick, isCurrent } = props;
+
+  if (!onClick) {
```
axelboc, line 14: We've got a case of mutually exclusive props: `onClick` and `isCurrent`. Since `onClick` is needed, I would just compute `const isCurrent = !onClick`. Unless you think of a better solution.

axelboc (approved):
Nicely done!

src/h5web/breadcrumbs/Crumb.tsx
axelboc: Alternatively, you could keep the last crumb inlined in the parent.

loichuder: `const isCurrent = !onClick` is fine. 

While it is a big assumption to tell that the crumb is the current one if there is no `onClick` behaviour, we can easily revert in the future.

loichuder: Nah, it is clearer to keep it inlined in fact.

PR #456, opened 2021-02-04
Upgrade react-scripts to 4.0.2

axelboc:
https://github.com/facebook/create-react-app/releases/tag/v4.0.2

### DIFF
--- a/.env
+++ b/.env
@@ -3,8 +3,8 @@
 # Copy and configure the variables below in your own `.env.local` file
 #####
 
-# Path to entity to select on first mount in Explorer.
-# Provide an absolute path (e.g. `/entry_0000`) or leave blank to select root entity.
+# Path to entity to select on first mount in Explorer
+# Provide an absolute path (e.g. `/entry_0000`) or leave blank to select root entity
 REACT_APP_DEFAULT_PATH=
 
 # HSDS parameters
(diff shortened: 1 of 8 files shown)


.env
```diff
@@ -21,6 +21,10 @@ REACT_APP_JLAB_FALLBACK_DOMAIN="water_224.h5"
 # Rendering performance profiling (local development only)
 REACT_APP_PROFILING_ENABLED=false
 
+# Disable ESLint error modal in development (log warnings instead)
+ESLINT_NO_DEV_ERRORS=true
```
axelboc, line 25: Yay! No more bloody error modal for an unused var!! Note that TS errors will still bring up a modal.

package.json
```diff
@@ -101,11 +101,12 @@
     "prettier": "^2.2.1",
     "react-app-polyfill": "^2.0.0",
     "react-is": "^16.13.1",
-    "react-scripts": "4.0.1",
+    "react-scripts": "4.0.2",
     "typescript": "^4.1.3",
     "worker-loader": "^3.0.7"
   },
   "jest": {
+    "restoreMocks": true,
```
axelboc, line 109: Avoids mocks leaking into other tests.

.env
axelboc: Well, bad example, since we turn unused vars errors into warnings in our ESLint config, but you know what I mean.

.env
```diff
@@ -21,6 +21,10 @@ REACT_APP_JLAB_FALLBACK_DOMAIN="water_224.h5"
...
+
 # Logging limit of React Testing Library
-# https://testing-library.com/docs/dom-testing-library/api-helpers/#debugging
+# Don't override this variable in `.env.local`, as this file is not used during tests
```
loichuder, line 28: This means it should be set in `.env` to have an effect ?

loichuder (approved):

.env
axelboc: You can set it temporarily in `.env` and not commit it, on the terminal directly (before `npm test`), or in `.env.test.local` if you want to keep it around.

PR #459, opened 2021-02-04
HDF5type is now deduced from dtype in JupyterProvider

loichuder:
Fixes the first part of #406

### DIFF
--- a/src/h5web/providers/jupyter/api.ts
+++ b/src/h5web/providers/jupyter/api.ts
@@ -5,6 +5,7 @@ import {
   assertGroupContent,
   isDatasetResponse,
   isGroupResponse,
+  convertDtype,
 } from './utils';
 import type {
   JupyterAttrsResponse,
@@ -12,7 +13,7 @@ import type {
   JupyterDataResponse,
   JupyterMetaResponse,
 } from './models';
-import { makeStrAttr, floatType } from '../mock/metadata-utils';
+import { makeStrAttr } from '../mock/metadata-utils';
 import {
   HDF5LinkClass,
   HDF5ShapeClass,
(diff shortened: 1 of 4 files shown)


src/h5web/providers/jupyter/api.ts
```diff
@@ -108,12 +109,14 @@ export class JupyterApi implements ProviderAPI {
     }
 
     if (isDatasetResponse(response)) {
-      const { type, shape: dims } = response;
+      // `type` is the type of the fetched entity
```
axelboc, line 112: I feel this comment adds confusion 😅 -- the renaming in the destructuring and the second comment are sufficient.

src/h5web/providers/jupyter/utils.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        charSet: 'H5T_CSET_UTF8',
+        strPad: 'H5T_STR_NULLPAD',
+        length: 25,
+      });
+    });
+
+    it('should convert bytes string dtypes', () => {
+      expect(convertDtype('|S')).toEqual({
+        class: HDF5TypeClass.String,
+        charSet: 'H5T_CSET_ASCII',
```
axelboc, line 18: Should we store this more concisely like the endianness?

src/h5web/providers/jupyter/utils.test.ts
```diff
@@ -0,0 +1,89 @@
+import { HDF5TypeClass } from '../hdf5-models';
+import { convertDtype } from './utils';
+
+describe('JupyterProvider utilities', () => {
+  describe('convertDtype', () => {
+    it('should convert unicode string dtypes', () => {
+      expect(convertDtype('|U25')).toEqual({
+        class: HDF5TypeClass.String,
+        charSet: 'H5T_CSET_UTF8',
+        strPad: 'H5T_STR_NULLPAD',
```
axelboc, line 10: Should we just remove `strPad` from the model, since we can't get the information?

axelboc (approved):
Awesome!

src/h5web/providers/jupyter/utils.test.ts
loichuder: I would need to change the parsing of string types in the HsdsProvider but I would agree.

Having a field called `charSet` with `CSET` in the value is quite redundant.

src/h5web/providers/jupyter/utils.test.ts
loichuder: Again, I need to change in the HsdsProvider but yes it would make sense. We can restore it if the users ask for it.

PR #461, opened 2021-02-04
Make dimension mapper narrower

axelboc:
User feedback:

>  Toujours dans le heatmap je pense que le menu de gauche (slider + definition of x and y...) pourrait être un peu plus petit.

---

The screenshots don't do it justice, but the mapper becomes 19px narrower with a 2D dataset and 24px narrower with a 4D dataset, which is significant.

### BEFORE

image

### AFTER

image


axelboc:
Note that I also give a bit more vertical space to the sliders.

We could probably give the heatmap vis a bit more space by reducing the space to the left of the canvas and to the right of the color bar.

loichuder (approved):

loichuder:
> We could probably give the heatmap vis a bit more space by reducing the space to the left of the canvas and to the right of the color bar.

Indeed. This should be the focus of a dedicated issue: these spaces is quite big to allow for numbers with many digits but we could allocate the space more smartly.

PR #468, opened 2021-02-05
Adding data type row to the metadata inspector

loichuder:
Some examples:
- Scalar dataset with string type
image

- 2D dataset with integer type
image

### DIFF
--- a/src/h5web/metadata-viewer/EntityTable.tsx
+++ b/src/h5web/metadata-viewer/EntityTable.tsx
@@ -2,7 +2,7 @@ import { ReactElement, useContext } from 'react';
 import type { Entity } from '../providers/models';
 import styles from './MetadataViewer.module.css';
 import { hasSimpleShape, isDataset, isDatatype, isLink } from '../guards';
-import { renderShapeDims } from './utils';
+import { renderType, renderShapeDims } from './utils';
 import RawInspector from './RawInspector';
 import LinkInfo from './LinkInfo';
 import { capitalize } from 'lodash-es';
@@ -38,11 +38,7 @@ function EntityTable(props: Props): ReactElement {
         {(isDataset(entity) || isDatatype(entity)) && (
           <tr>
             <th scope="row">Type</th>
-            <td>
-              {typeof entity.type === 'string'
-                ? entity.type
-                : entity.type.class}
-            </td>
+            <td>{renderType(entity.type)}</td>
           </tr>
         )}
         {isDataset(entity) && (
--- a/src/h5web/metadata-viewer/utils.ts
+++ b/src/h5web/metadata-viewer/utils.ts
@@ -1,7 +1,51 @@
+import { HDF5Type, HDF5TypeClass } from '../providers/hdf5-models';
+
+const ENDIANNESS_LABELS = {
+  LE: 'little-endian',
+  BE: 'big-endian',
+  Native: 'native',
+  'Not applicable': '',
+};
+
 export function renderShapeDims(dims: number[]): string {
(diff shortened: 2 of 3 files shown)


src/h5web/metadata-viewer/EntityTable.tsx
```diff
@@ -37,14 +43,20 @@ function EntityTable(props: Props): ReactElement {
...
                 : entity.type.class}
             </td>
           </tr>
         )}
+        {isDataset(entity) && hasBaseType(entity) && (
+          <tr>
+            <th scope="row">Data type</th>
+            <td>{renderBaseType(entity.type)}</td>
+          </tr>
+        )}
```
axelboc, line 59: It feels a bit redundant to show the main type twice (e.g. Integer). Don't you think a single _Type_ row would suffice?

src/h5web/metadata-viewer/utils.ts
```diff
@@ -1,7 +1,45 @@
...
+    case 'Native':
+      return 'Native';
+
+    case 'Not applicable':
+      return '';
+
+    default:
+      return '';
+  }
+}
```
axelboc, line 32: This could just be a `Record`.

src/h5web/metadata-viewer/utils.ts
```diff
@@ -1,7 +1,45 @@
...
+
+export function renderBaseType(type: HDF5BaseType): string {
+  if (type.class === HDF5TypeClass.String) {
+    const { length, charSet } = type;
+    return `${charSet} string (size:
+      ${Number.isInteger(length) ? length : 'variable'})`;
```
axelboc, line 38: I would keep the term `length` and display "variable length" rather than "length: variable". Perhaps you could also replace the parens with a simple colon: `UTF8 string, variable length`.

src/h5web/metadata-viewer/utils.ts
```diff
@@ -1,7 +1,45 @@
...
+  }
+
+  const { endianness, size } = type;
+  return `${renderEndianness(endianness)}
+  ${type.class === HDF5TypeClass.Integer && type.unsigned ? 'unsigned' : ''}
+  ${size}-bit ${type.class.slice(4).toLowerCase()}`;
```
axelboc, line 44: If you agree with the single-row idea, I think it would make sense to show the most relevant information first rather than to make a grammatically correct phrase. For instance:

- "Float, 32-bit, big-endian"
- "Integer, unsigned 16 bit, little-endian"
- "String, variable length"
- "String, length=42

Not sure about the exact formatting/construction. Perhaps Silx can be of some inspiration here.

axelboc (changes requested):
Finally some more precise types in the metadata viewer! 👏 Here are some nitpicks.

src/h5web/metadata-viewer/utils.ts
loichuder: What about `string, variable length` or `string, n characters` ?

src/h5web/metadata-viewer/EntityTable.tsx
loichuder: This will make the condition for displaying the row more complex but I would agree, yes

src/h5web/metadata-viewer/utils.ts
axelboc: How about `n chars` so it's a bit more concise?

loichuder:
New aspect:
image
image

Also works for datatypes now:
image

axelboc (approved):
Perfect!

PR #469, opened 2021-02-05
Fetch values by slice in Heatmap and Matrix vis

axelboc:
In order to fetch values by slice, I need to pass an extra `selection` parameter (of the form `0,0,:,:`) along with the entity `path`. I also need the values store to use both the path and the selection as keys when caching results.

`react-suspense-fetch` accepts fetch functions with a single string or object parameter. An object parameter would do the job, but when `react-suspense-fetch` encounters an object parameter, it stores it in a `WeakMap` cache internally, which means that cached results would disappear as soon as the input objects were garbage collected...

The workaround is to stringify an input object before passing it to the store, and then parsing the string back into an object inside the fetch function:

```ts
const store = createFetchStore(async (key: string) => {
  const { path, selection } = JSON.parse(key);
  // ...
});

store.get(JSON.stringify({ path, selection }));
```

To avoid cluttering the code, enforce good typing and work around challenges posed by the fact that passing a `selection` should be optional, I implemented a thin wrapper around `react-suspense-fetch`: `createObjectKeyStore`. This store accepts fetch function with a single object parameter of any shape.

If an optional property of the object parameter is passed explicitly as `undefined`, the property is removed from the object so that `JSON.stringify` doesn't convert it to `null` -- this ensures that `{ path: 'foo', selection: undefined }` and `{ path: 'foo' }` are stored with the same key in the cache.

### DIFF
--- a/src/h5web/providers/mock/MockProvider.tsx
+++ b/src/h5web/providers/mock/MockProvider.tsx
@@ -2,6 +2,10 @@ import type { ReactElement, ReactNode } from 'react';
 import Provider from '../Provider';
 import { mockDomain } from './metadata';
 import { assertMockDataset, findMockEntity } from './utils';
+import type { GetValueParams } from '../context';
+import ndarray from 'ndarray';
+import { assertNumericType, assertSimpleShape } from '../../guards';
+import unpack from 'ndarray-unpack';
 
 interface Props {
   domain?: string;
@@ -26,7 +30,9 @@ function MockProvider(props: Props): ReactElement {
 
           return findMockEntity(path);
         },
-        getValue: async (path: string) => {
+        getValue: async (params: GetValueParams) => {
+          const { path, selection } = params;
+
           if (path === errorOnPath) {
             // Throw error when fetching value with specific ID
             throw new Error('error');
@@ -35,7 +41,26 @@ function MockProvider(props: Props): ReactElement {
           const dataset = findMockEntity(path);
           assertMockDataset(dataset);
 
-          return dataset.value;
+          const { value } = dataset;
+          if (!selection) {
+            return value;
+          }
+
+          assertSimpleShape(dataset);
+          assertNumericType(dataset);
+
+          const dataArray = ndarray(
+            (dataset.value as number[]).flat(Infinity),
+            dataset.shape.dims
+          );
+
+          const dataView = dataArray.pick(
+            ...selection
+              .split(',')
+              .map((s) => (s === ':' ? null : Number.parseInt(s, 10)))
+          );
+
+          return unpack(dataView);
         },
       }}
     >
--- a/src/h5web/providers/utils.ts
+++ b/src/h5web/providers/utils.ts
@@ -1,0 +1,43 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { pickBy } from 'lodash-es';
+import { createFetchStore } from 'react-suspense-fetch';
+
+export interface ObjectKeyStore<R, I extends {}> {
+  get: (obj: I) => R;
+  prefetch: (obj: I) => void;
+  evict: (obj: I) => void;
+}
+
+/**
+ * Create a suspense store with a reliable `Map` cache around an async
+ * function with an object parameter. `react-suspense-fetch` uses a
+ * `WeakMap` cache in this case.
+ *
+ * Interaction with the underlying `react-suspense-fetch` store is done by
+ * JSON-stringifying/parsing the input object.
+ */
+export function createObjectKeyStore<R, I extends {}>(
+  fetchFunc: (obj: I) => Promise<R>
+): ObjectKeyStore<R, I> {
+  const store = createFetchStore((key: string) => {
+    return fetchFunc(JSON.parse(key));
+  });
+
+  function wrapMethod<T>(method: (key: string) => T) {
+    return (params: I) => {
+      /* Remove optional params that are explicitely set to `undefined`.
+       * This is to guarantee that, for instance, `{ foo: undefined; }` and `{}`
+       * are both stringified to (and cached with) "{}" */
+      const pickedParams = pickBy(params, (v) => v !== undefined);
+
+      return method(JSON.stringify(pickedParams));
+    };
+  }
+
+  const { get, prefetch, evict } = store;
+  return {
+    get: wrapMethod(get),
+    prefetch: wrapMethod(prefetch),
+    evict: wrapMethod(evict),
+  };
+}
--- a/src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
+++ b/src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
@@ -10,6 +10,7 @@ import type {
   HDF5NumericType,
   HDF5SimpleShape,
 } from '../../../providers/hdf5-models';
+import { isAxis } from '../../../dimension-mapper/utils';
 import shallow from 'zustand/shallow';
 
 interface Props {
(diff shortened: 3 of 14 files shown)


src/h5web/vis-packs/core/hooks.ts
```diff
@@ -16,11 +16,30 @@ export function useDatasetValue<
...
+   * with a 2D dataset, as it ensures the selection string is always ":,:". */
+  const selection = dimMapping
+    ?.map((dim) => (!fetchAll && isNumber(dim) ? dim : ':'))
+    .join(',');
 
-export function useDatasetValue(dataset: Dataset | undefined) {
   const { valuesStore } = useContext(ProviderContext);
-  return dataset && valuesStore.get(dataset.path);
+  return dataset && valuesStore.get({ path: dataset.path, selection });
 }
```
axelboc, line 43: I hope the comment is clear. I don't see a simpler way to explain why the `LineVis` has to call `useDatasetValue` with its `dimMapping` and the `fetchAll` flag set to `true`, and I prefer the comment to be here than in `LineVis`.

src/h5web/providers/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+    };
+  }
+
+  const { get, prefetch, evict } = store;
+  return {
+    get: wrapMethod(get),
+    prefetch: wrapMethod(prefetch),
+    evict: wrapMethod(evict),
+  };
+}
```
axelboc, line 43: I tried, I really tried to implement a store that could accept a fetch function with any number of parameters, including optional parameters - e.g. `getValue(path: string, selection?: string)`. Between tuples and generics, I just couldn't get the typing right in the store, and I couldn't find a safe way to implement the work around for optional parameters (i.e. removing `undefined` values for caching purposes). 😢 So an object it is!

src/h5web/vis-packs/core/matrix/MappedMatrixVis.tsx
```diff
@@ -17,9 +18,18 @@ interface Props {
...
+  const [slicedDims, slicedMapping] = useMemo(
+    () => [
+      dimMapping.flatMap((dim, i) => (isAxis(dim) ? [dims[i]] : [])),
+      dimMapping.filter((dim) => isAxis(dim)),
+    ],
+    [dimMapping, dims]
+  );
+
+  const baseArray = useBaseArray(value, slicedDims);
+  const mappedArray = useMappedArray(baseArray, slicedMapping);
```
axelboc, line 32: Same code as in `MappedHeatmapVis`. Wasn't sure it was worth/a good idea to abstract it or to rethink this whole section of code. I don't mind the duplication too much, but if you have a better idea 🤷 

I can't just pass `dimMapping` and `dims` to `useBaseArray` and compute `slicedDims` in there, because I don't want to compute `slicedDims` in `MappedLineVis`... Same goes with `useMappedArray`...

src/h5web/vis-packs/nexus/containers/NxSpectrumContainer.tsx
```diff
@@ -34,6 +35,12 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
 
   const [dimMapping, setDimMapping] = useDimMappingState(signalDims, 1);
 
+  const { valuesStore } = useContext(ProviderContext);
+  valuesStore.prefetch({ path: signalDataset.path });
+  if (errorsDataset) {
+    valuesStore.prefetch({ path: errorsDataset.path });
+  }
```
axelboc, line 42: As discussed, I had to remove the prefetching of the signal from `useNxData`, so I decided to do the same with errors for consistency and move the prefetching here, in `NxSpectrumContainer`.

src/h5web/vis-packs/nexus/hooks.ts
```diff
@@ -27,40 +27,9 @@ import {
...
-  }
-
-  assertDataset(dataset);
-  assertSimpleShape(dataset);
-  assertNumericType(dataset);
-
-  valuesStore.prefetch(dataset.path);
-  return dataset;
-}
-
```
axelboc, line 63: Becomes `findErrorsDataset` in `utils.ts`.

src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -40,9 +41,18 @@ function MappedHeatmapVis(props: Props): ReactElement {
...
+  const [slicedDims, slicedMapping] = useMemo(
+    () => [
+      dimMapping.flatMap((dim, i) => (isAxis(dim) ? [dims[i]] : [])),
+      dimMapping.filter((dim) => isAxis(dim)),
+    ],
+    [dimMapping, dims]
+  );
+
+  const baseArray = useBaseArray(value, slicedDims);
+  const dataArray = useMappedArray(baseArray, slicedMapping);
```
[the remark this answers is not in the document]
loichuder: To be changed in a future PR as discussed

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -16,11 +16,32 @@ export function useDatasetValue<
...
+  }
+
+  // If no dim mapping is provided or dim mapping has no slicing dimension, fetch entire dataset
+  if (!dimMapping || !dimMapping.some(isNumber)) {
+    return valuesStore.get({ path: dataset.path });
+  }
+
+  // Create slice selection string from dim mapping - e.g. [0, 'y', 'x'] => "0,:,:"
+  const selection = dimMapping
+    .map((dim) => (isNumber(dim) ? dim : ':'))
```
loichuder, line 41: Wouldn't be clearer to use your new util `isAxis` ? `isAxis(dim) ? ":" : dim`

src/h5web/vis-packs/core/matrix/MappedMatrixVis.tsx
loichuder: The `useMemo` computing `slicedDims` and `slicedMapping` could be put in a shared hook. 

The rest is fine as it is bound to change anyway.

loichuder (changes requested):

src/h5web/providers/mock/MockProvider.tsx
```diff
@@ -35,7 +41,26 @@ function MockProvider(props: Props): ReactElement {
...
+            dataset.shape.dims
+          );
+
+          const dataView = dataArray.pick(
+            ...selection
+              .split(',')
+              .map((s) => (s === ':' ? null : Number.parseInt(s, 10)))
+          );
+
+          return unpack(dataView);
```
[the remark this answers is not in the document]
axelboc: Looks like I can! 🎣

src/h5web/vis-packs/core/matrix/MappedMatrixVis.tsx
axelboc: I'll do it in the next PR, in case this is impacted by the refactoring of the hooks.

src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -40,9 +41,18 @@ function MappedHeatmapVis(props: Props): ReactElement {
...
   } = useHeatmapConfig();
 
-  const value = useDatasetValue(dataset);
-  const baseArray = useBaseArray(value, dims);
-  const dataArray = useMappedArray(baseArray, dimMapping);
+  const value = useDatasetValue(dataset, dimMapping);
+
+  const [slicedDims, slicedMapping] = useMemo(
+    () => [
+      dimMapping.flatMap((dim, i) => (isAxis(dim) ? [dims[i]] : [])),
```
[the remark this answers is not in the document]
axelboc: Hehe nope, doesn't work. The flat map does this, for instance: `[0, 0, 'y', 'x']` => `[20, 41]`. Axes are converted to dimensions.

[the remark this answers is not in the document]
axelboc: Oooh I get it! Nice 😄

loichuder (approved):

PR #471, opened 2021-02-08
Upgrade dependencies and ESLint config

axelboc:


.eslintrc.js
```diff
@@ -37,8 +42,7 @@ module.exports = createConfig({
     {
       files: tsFiles,
       rules: {
-        '@typescript-eslint/no-confusing-void-expression': 'off', // not yet supported by CRA
```
axelboc, line 40: I've installed the ESLint plugin so we can get the latest version without waiting for CRA. This allows us to benefit from some new rules enabled in eslint-config-galex.

.eslintrc.js
```diff
@@ -37,8 +42,7 @@ module.exports = createConfig({
...
-        'testing-library/no-await-sync-events': 'off', // not yet supported by CRA
+        '@typescript-eslint/non-nullable-type-assertion-style': 'off', // conflicts with @typescript-eslint/no-non-null-assertion
```
axelboc, line 45: I'll remove once this is resolved: https://github.com/ljosberinn/eslint-config-galex/issues/235

package.json
```diff
@@ -90,13 +90,16 @@
     "@types/react-dom": "^16.9.10",
     "@types/react-measure": "^2.0.6",
     "@types/react-router-dom": "^5.1.7",
-    "@types/react-slider": "^1.1.0",
+    "@types/react-slider": "^1.1.1",
```
axelboc, line 93: Updating this as agreed.

package.json
```diff
@@ -90,13 +90,16 @@
...
+    "@typescript-eslint/eslint-plugin": "^4.14.2",
     "babel-loader": "8.1.0",
     "cypress": "^6.3.0",
     "cypress-image-snapshot": "^4.0.1",
-    "eslint": "^7.14.0",
-    "eslint-config-galex": "^2.9.4",
+    "eslint": "^7.19.0",
+    "eslint-config-galex": "^2.10.6",
+    "eslint-plugin-react": "^7.22.0",
+    "eslint-plugin-testing-library": "^3.10.1",
```
axelboc, line 102: So as mentioned, I'm declaring three of the plugins used by `eslint-config-galex` so we can get some of the new rules without waiting for CRA and without having to disable them in our ESLint config.

src/h5web/toolbar/HeatmapToolbar.tsx
```diff
@@ -22,7 +23,7 @@ function HeatmapToolbar(): ReactElement {
     toggleAspectRatio,
     showGrid,
     toggleGrid,
-  } = useHeatmapConfig();
+  } = useHeatmapConfig((state) => state, shallow);
```
axelboc, line 26: I noticed some useless re-rendenders and this was the cause.

We call setters in effects without checking if they're changing anything to the state, like:

```ts
  useEffect(() => {
    disableErrors(!errors);
  }, [disableErrors, errors]);
```

This is fine, but if we don't pass a selector to `useLineConfig` and `useHeatmapConfig`, this kind of state update always triggers a re-render. By passing a selector and `shallow`, we check if anything has changed before triggering a re-render.

Somehow I thought zustand did this check internally by default, but that's not the case.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -188,12 +188,7 @@ export function getIntegerTicks(domain: Domain, count: number): number[] {
   const stop = Math.floor(max / step);
   const numTicks = stop - start + 1;
 
-  const ticks = new Array(numTicks);
-  for (let i = 0; i < numTicks; i++) {
-    ticks[i] = (start + i) * step;
-  }
-
-  return ticks;
+  return Array.from({ length: numTicks }, (_, i) => (start + i) * step);
```
axelboc, line 191: Didn't know this syntax for array intialisation existed. According to https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-new-array.md, it's clearer than `new Array(n)` because `n` can't be confused with an element of the array.

I especially like the second parameter of `Array.from` -- it's feels nicer than chaining the constructor call with `fill` and `map` like we do in other places.

src/h5web/vis-packs/hooks.ts
```diff
@@ -22,7 +22,7 @@ export function useActiveVis<T extends VisDef>(supportedVis: T[]) {
 
 export function useDimMappingState(dims: number[], axesCount: number) {
   return useState<DimensionMapping>([
-    ...new Array(dims.length - axesCount).fill(0),
-    ...['y', 'x'].slice(-axesCount),
+    ...Array.from({ length: dims.length - axesCount }, () => 0),
+    ...['y' as const, 'x' as const].slice(-axesCount),
```
axelboc, line 26: I could also extract the x/y array and give it type `Axis[]`, but it doesn't seem worth it.

loichuder (approved):

PR #472, opened 2021-02-10
Various changes in the CSS to ease up implementation of themes

loichuder:
- Replaced colors with CSS variables (including default curve color that is now set though canvas color)
- Forced background app color to white
- Made `PanZoomMesh` and `TooltipMesh` transparent


src/h5web/vis-packs/core/shared/TooltipMesh.module.css
```diff
@@ -1,6 +1,12 @@
 .tooltip {
   color: inherit !important;
-  background-color: rgba(245, 251, 239, 0.7) !important;
+  background-color: var(--primary-light-bg) 0.7 !important;
```
loichuder, line 3: Actually, this is not exactly the same color but quite close: https://calcolor.co/blend/fafdf7/f5fbef

src/h5web/toolbar/controls/Selector/Selector.module.css
```diff
@@ -52,7 +52,7 @@
   margin-top: 0.75rem;
   margin-bottom: 0.25rem;
   padding-left: 0.75rem;
-  color: #777;
+  color: var(--gray);
```
loichuder, line 55: This is why snapshots needs to be updated: change from `#777777` to `#808080`

src/h5web/vis-packs/core/line/LineVis.module.css
```diff
@@ -5,3 +5,7 @@
   min-height: 0;
   margin: 1rem;
 }
+
+.root canvas {
+  color: var(--secondary-dark);
```
axelboc, line 10: A small comment here, maybe?

src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
```diff
@@ -93,6 +93,7 @@ function PanZoomMesh(): ReactElement {
 
   return (
     <mesh {...{ onPointerMove, onPointerUp, onPointerDown, onWheel }}>
+      <meshBasicMaterial attach="material" color="#000000" opacity={0} />
```
axelboc, line 96: I know the color is not visible, that's the point, but maybe use `#ffffff` for clarity?

src/h5web/vis-packs/core/shared/TooltipMesh.tsx
```diff
@@ -78,6 +72,7 @@ function TooltipMesh(props: Props): ReactElement {
   return (
     <>
       <mesh {...{ onPointerMove, onPointerOut, onPointerDown, onPointerUp }}>
+        <meshBasicMaterial attach="material" color="#000000" opacity={0} />
```
axelboc, line 75: same here

src/styles/vars.css
```diff
@@ -1,14 +1,21 @@
 :root {
   --black: #020402;
+  --dark-gray: #696969;
+  --gray: #808080;
+  --white: #ffffff;
+  --primary-axis-color: #2f4f4f;
+  --secondary-axis-color: var(--gray);
```
axelboc, line 7: A bit weird to use semantic names here, for these two variables. It would be good to store `#2f4f4f` in a colour variable, like `--dark-slate-gray` and then maybe move these two axis variables to `AxisSystem.module.css`, inside the `.axisSystem` declaration block.

axelboc (approved):

src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
loichuder: > maybe use `#ffffff` for clarity?

I see what you did there :smirk:

src/styles/vars.css
loichuder: I am looking this from the extension perspective: it would make more sense to me to set variables with semantic names. 

But I see your point.

axelboc: Even `primary-dark` and `secondary-light` are not semantic names in a way, since you'd have to override them with inverted colours and the names would no longer be "valid". With fully semantic names, we'd have to use names specific to each component, like `--axis-primary-color`. Having such variables at the root would not make much sense and moving them to their respective components would make them more difficult to override... You'd also have no clue what the underlying color is without refering back to the variable's value or how it relates to another colour (e.g. whether the foreground colour darker or lighter than the background colour).

What's good with kind-of-but-not-quite-semantic colour names is that they don't completely hide what the actual colour is (or at least not the darkness/brightness aspect). This is actually a good thing when overriding with inverted colors, since you can easily double check that the relations between colours remain logical (e.g. `--primary-dark` is lighter than `--primary`).

Not sure if my reasoning is clear?

loichuder: It is. This is why I said that I saw your point :smile:

PR #473, opened 2021-02-10
Start implementing shaders-based heatmap behind flag

axelboc:
This is an initial implementation to get things moving. For now, I use the inferno colour map function in the shader (instead of passing the colour map as a texture, as discussed). Also, only the linear scale is supported.

image

### DIFF
--- a/.env
+++ b/.env
@@ -18,6 +18,9 @@ REACT_APP_HSDS_FALLBACK_DOMAIN="/home/reader/water"
 REACT_APP_JLAB_URL=
 REACT_APP_JLAB_FALLBACK_DOMAIN="water_224.h5"
 
+# Development-only feature flags
+REACT_APP_DEV_SHADERS=false
+
 # Rendering performance profiling (local development only)
 REACT_APP_PROFILING_ENABLED=false
 
--- a/src/h5web/vis-packs/core/heatmap/Mesh.tsx
+++ b/src/h5web/vis-packs/core/heatmap/Mesh.tsx
@@ -45,7 +45,7 @@ function Mesh(props: Props): ReactElement {
     <>
       {material && (
         <mesh material={material}>
-          <planeBufferGeometry attach="geometry" args={[width, height]} />
+          <planeGeometry attach="geometry" args={[width, height]} />
         </mesh>
       )}
       {showLoader && (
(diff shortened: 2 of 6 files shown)


.env
```diff
@@ -18,6 +18,9 @@ REACT_APP_HSDS_FALLBACK_DOMAIN="/home/reader/water"
 REACT_APP_JLAB_URL=
 REACT_APP_JLAB_FALLBACK_DOMAIN="water_224.h5"
 
+# Development-only feature flags
+REACT_APP_DEV_SHADERS=false
+
```
axelboc, line 23: Just turn it on in `.env.local` to switch to the new shaders-based mesh. 😉

src/h5web/vis-packs/core/heatmap/Mesh.tsx
```diff
@@ -45,7 +45,7 @@ function Mesh(props: Props): ReactElement {
     <>
       {material && (
         <mesh material={material}>
-          <planeBufferGeometry attach="geometry" args={[width, height]} />
+          <planeGeometry attach="geometry" args={[width, height]} />
```
axelboc, line 48: Turns out that `PlaceBufferGeometry` is just an alias. `PlaneGeometry` is already a `BufferGeometry`.

src/h5web/vis-packs/core/heatmap/ShadersMesh.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const shader = useMemo(
+    () => ({
+      uniforms: {
+        data: {
+          value: new DataTexture(
+            Float32Array.from(values),
+            cols,
+            rows,
+            RedFormat,
```
axelboc, line 24: `RedFormat` avoids having to add three zeros for each value to our `values` array before passing it to the `DataTexture`.

src/h5web/vis-packs/core/heatmap/ShadersMesh.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <mesh scale={[width, height, 1]}>
+      <planeGeometry attach="geometry" args={[1, 1]} />
+      <shaderMaterial attach="material" args={[shader]} />
+    </mesh>
+  );
+}
+
+export default memo(ShadersMesh);
```
loichuder, line 79: If the `ShadersMesh` is memoized, it has the props as deps, right ? Meaning `[rows, cols, values, domain]`.

In this case, is the `useMemo` with the same deps really useful ?

loichuder (approved):
It is so faaaaaaaaast :rocket:

src/h5web/vis-packs/core/heatmap/ShadersMesh.tsx
axelboc: Excellent point!

PR #477, opened 2021-02-11
Implemented scale type support for ShadersMesh

loichuder:


axelboc (approved):
I did some more testing and couldn't find any issue. 🍻

PR #478, opened 2021-02-12
Switch to shaders-based Heatmap and remove old worker code

axelboc:
image

The shaders-based implementation is a lot lighter. We can remove three dependencies relating to web workers (61 kB off the unminified `main` bundle), and simplify the E2E tests a little since the renders should now be near instantaneous.

Looks like the shaders render the heatmaps the exact same way as before and that the E2E tests are more reliable (the delay is now down to 200):

image


.gitignore
```diff
@@ -16,4 +16,3 @@ npm-debug.log*
 *.log
 Thumbs.db
 .DS_Store
-.eslintcache
```
axelboc, line 19: In the latest patch version of CRA, this file has moved back to its proper location in the `node_modules/.cache` folder.

axelboc:
Took me a lot longer than I wanted, but here is a new demo GIF for the `README` where the heatmap is super fast. I couldn't demo the slicing of the image stack because it was making the GIF too big.

demo

src/h5web/vis-packs/core/heatmap/Mesh.tsx
```diff
@@ -1,61 +1,129 @@
...
-        <Html
-          className={styles.textureLoader}
-          data-visible={loading || undefined}
-        />
-      )}
-    </>
+    <mesh scale={[width, height, 1]}>
+      <planeGeometry attach="geometry" args={[1, 1]} />
+      <shaderMaterial attach="material" args={[shader]} />
+    </mesh>
```
loichuder, line 125: Why not keeping `width, height` on the geometry ?
```js
   <mesh>
      <planeGeometry attach="geometry" args={[width, height]} />
      <shaderMaterial attach="material" args={[shader]} />
    </mesh>
```

loichuder (approved):

src/h5web/vis-packs/core/heatmap/Mesh.tsx
axelboc: If I do, the `uv` values are not between `0` and `1` but, I think, between `0` and `width` or `height`, which doesn't work well with the `texture()` function -- we get near-solid heatmap with only the "highest" colour in the colour map.

PR #483, opened 2021-02-15
Don't round domain slider values on change

### LINKED ISSUE #444: Slider rounds data domain
1. Open `fourD` mock dataset, and switch to log scale.

2. Slider receives data domain
 
image

3. Move right-hand thumb

image

=> `customDomain` becomes `[0, 0.9]` because of rounding and visualization breaks.

axelboc:
Fix #444 

Hadn't realised this was our doing 😂 

The use case mentioned in the issue is fixed:

image

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider.tsx
@@ -1,7 +1,6 @@
 import type { ReactElement } from 'react';
 import ReactSlider from 'react-slider';
 import { format } from 'd3-format';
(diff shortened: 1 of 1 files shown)


loichuder (commented):
This fix means that there is now a discrepancy between the displayed value and the actual value used for the colormap domain.

Do you think this will be a problem ? For example, ticks are no longer nicely placed:

**Old**:
image

**New:**
image

axelboc:
Yeah, on the other hand, there was already a discrepancy when auto-scaling.

PR #489, opened 2021-02-16
Show two solid colors in color bar when domain is empty

axelboc:
When the domain has no size (i.e. `[x, x]`), the colour bar axis has a single tick, centred in the middle of the colour bar, but the colour bar itself still shows a gradient. In the visualization, any value below `x` is represented with the colour map's min colour and any value above `x` with the colour map's max colour, so the colour bar does not represent the visualization accurately.

This PR deals with this edge case by showing two solid colours in the colour bar instead of a gradient (though it is still a CSS linear-gradient under the hood).

FYI, my next PR will be about showing the min/max under/above the colour bar. In the case of an empty domain, I was thinking of actually showing `-Infinity` and `+Infinity` to accurately convey what's going on in the visualization.

### BEFORE

image

### AFTER

image

### DIFF
--- a/src/h5web/dimension-mapper/SlicingSlider.tsx
+++ b/src/h5web/dimension-mapper/SlicingSlider.tsx
@@ -30,6 +30,7 @@ function SlicingSlider(props: Props): ReactElement {
         // https://github.com/zillow/react-slider/issues/172
         key={`${mapperState.includes('y')}`}
         className={styles.slider}
+        ariaLabel="Dimension slider"
         trackClassName={styles.track}
         thumbClassName={styles.thumb}
         renderThumb={(thumbProps, state) => (
--- a/src/h5web/toolbar/controls/ColorMapSelector.tsx
+++ b/src/h5web/toolbar/controls/ColorMapSelector.tsx
@@ -8,7 +8,7 @@ import {
   SINGLE_HUE,
 } from '../../vis-packs/core/heatmap/interpolators';
 import type { ColorMap } from '../../vis-packs/core/heatmap/models';
-import { generateCSSLinearGradient } from '../../vis-packs/core/heatmap/utils';
+import { getLinearGradient } from '../../vis-packs/core/heatmap/utils';
 import Selector from './Selector/Selector';
 import styles from './ColorMapSelector.module.css';
 
@@ -22,10 +22,8 @@ const COLORMAP_GROUPS = {
 
 function ColorMapOption(props: { option: ColorMap }): ReactElement {
   const { option } = props;
-  const backgroundImage = generateCSSLinearGradient(
-    INTERPOLATORS[option],
-    'right'
-  );
+  const backgroundImage = getLinearGradient(INTERPOLATORS[option], 'right');
+
   return (
     <>
       {option}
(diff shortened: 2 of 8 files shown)


src/h5web/toolbar/controls/ColorMapSelector.tsx
```diff
@@ -22,10 +22,8 @@ const COLORMAP_GROUPS = {
 
 function ColorMapOption(props: { option: ColorMap }): ReactElement {
   const { option } = props;
-  const backgroundImage = generateCSSLinearGradient(
-    INTERPOLATORS[option],
-    'right'
-  );
+  const backgroundImage = getLinearGradient(INTERPOLATORS[option], 'right');
```
axelboc, line 25: I think a shorter name works just as well.

src/h5web/vis-packs/core/heatmap/utils.ts
```diff
@@ -2,20 +2,34 @@ import { range } from 'lodash-es';
 import type ndarray from 'ndarray';
 import type { D3Interpolator, Dims } from './models';
 
+const GRADIENT_PRECISION = 1 / 20;
```
axelboc, line 5: I'm doubling the precision (was `0.1`). It was a bit too low for cyclical interpolators with lots of colours, like `Rainbow`.

src/h5web/dimension-mapper/SlicingSlider.tsx
```diff
@@ -30,6 +30,7 @@ function SlicingSlider(props: Props): ReactElement {
         // https://github.com/zillow/react-slider/issues/172
         key={`${mapperState.includes('y')}`}
         className={styles.slider}
+        ariaLabel="Dimension slider"
```
loichuder, line 33: You could add the number `dimension` in the label (i.e. `Dimension {dimension} slider`)

src/h5web/vis-packs/core/heatmap/utils.ts
```diff
@@ -2,20 +2,34 @@ import { range } from 'lodash-es';
...
+}
 
-  return `linear-gradient(to ${direction},${gradientColors})`;
+export function getLinearGradient(
+  interpolator: D3Interpolator,
+  direction: 'top' | 'bottom' | 'right' | 'left',
+  minMaxOnly = false
+): string {
+  const colorStops = getColorStops(interpolator, minMaxOnly);
+  return `linear-gradient(to ${direction}, ${colorStops})`;
```
loichuder, line 32: Not sure `getColorStops` is necessary. You could have put the early return into `getLinearGradient`:

```js
  if (minMaxOnly) {
    const min = interpolator(0);
    const max = interpolator(1);
    return `linear-gradient(to ${direction}, ${min}, ${min} 50%, ${max} 50%, ${max});
  }

  return `linear-gradient(to ${direction}, GRADIENT_RANGE.map(interpolator).join(', ')`;
```

Feel free to keep it like this though.

loichuder (approved):
Looking good :sparkles:

src/h5web/dimension-mapper/SlicingSlider.tsx
axelboc: The dimension index, yes, but not the size of the current dimension... and I think `dimension` is the latter 😢 -- I'd have to pass an extra prop to get the index, which seems overkill for this.

src/h5web/vis-packs/core/heatmap/utils.ts
axelboc: Yeah, I had this at first, but the returns don't fit nicely on single lines -- I have to add `colorStops` constants above them... And I wasn't a fan of having `linear-gradient(to ${direction},` duplicated 🤷‍♂️

PR #495, opened 2021-02-16
Adding boolean type, a subset of enum type

### LINKED ISSUE #479: [Scalar] Boolean values
A user has reported seeing the error `Expected number or string` when trying to visualize a `boolean` value. This error probably comes from `ScalarVis`. We need to investigate how this data type is exposed by the back-ends and how we should model it in our front-end models and providers.

loichuder:
A step towards #479 as even if not a base type, the boolean can be displayed using the Raw visu:
image

Glad to have your feedback on this: on one hand, we add the support of the enum type, on the other, it might seem overthought only for booleans.

### DIFF
--- a/src/h5web/providers/jupyter/utils.ts
+++ b/src/h5web/providers/jupyter/utils.ts
@@ -51,6 +51,25 @@ export function convertEndianness(endianness: string): HDF5Endianness {
 }
 
 export function convertDtype(dtype: string): HDF5Type {
+  // Special case: booleans are stored as bytes
+  // See https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bool
+  if (dtype === '|b1') {
+    // Booleans are stored as Enum by h5py
+    // https://docs.h5py.org/en/stable/faq.html#what-datatypes-are-supported
+    return {
+      class: HDF5TypeClass.Enum,
+      base: {
+        class: HDF5TypeClass.Integer,
+        endianness: 'Not applicable',
+        size: 8,
+      },
+      mapping: {
+        FALSE: 0,
+        TRUE: 1,
+      },
+    };
+  }
+
   const regexp = /([<>=|])?([A-z])(\d*)/u;
   const matches = regexp.exec(dtype);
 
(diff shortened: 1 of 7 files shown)


src/h5web/providers/jupyter/utils.ts
```diff
@@ -124,6 +124,22 @@ export function convertDtype(dtype: string): HDF5Type {
         length: length || 'H5T_VARIABLE',
       };
 
+    // Assuming that only booleans are stored as bytes
+    // https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.bool_
+    case 'b':
```
loichuder, line 129: Actually, the check could be stricter as we know that boolean dtype is `|b1`. We could then return directly `booleanType` defined in `mock/metadata-utils.ts`

src/h5web/providers/mock/values.ts
```diff
@@ -22,6 +22,7 @@ export const mockValues = {
   raw_large: { str: '.'.repeat(1000) },
   scalar_int: 0,
   scalar_str: 'foo',
+  scalar_bool: true,
```
axelboc, line 25: `'true'` should be a string to be consistent with our providers

src/h5web/providers/jupyter/utils.ts
axelboc: Happy to be stricter here and to return `booleanType`. So `|b1` cannot be anything else than a boolean? Or could it be a integer from 0 to 255 as well?

src/h5web/providers/mock/values.ts
loichuder: Actually, I checked once again and data requests return booleans :roll_eyes:

src/h5web/providers/jupyter/utils.ts
loichuder: After a few tries, arrays with `|b1` are directly interpreted as booleans even in NumPy... I think we are safe :sweat_smile: 

Do we need to move `booleanType` out of `mock/metadata-utils.ts` ?

axelboc: Hmm good point -- just keep it duplicated it, then. Better to keep all the mocked stuff really separate.

Issue #310, opened 2020-11-17
[Nexus] Add support for complex data


loichuder:
By complex data, I mean numbers that have a real and an imaginary part :wink:

t20100:
A sample dataset: http://ftp.esrf.fr/pub/scisoft/PyNX/data/ptycho-siemens-star-id01-result.cxi

loichuder:
A first implementation of complex vis was done in https://github.com/silx-kit/h5web/pull/634: it uses the HeatmapVis to show the phase as color and the amplitude as transparency.

Based on recent feedback, we should in fact implement a three-fold vis where we can switch between:
- Phase as colour, amplitude as transparency (current behaviour)
- Only phase as colour
- Only amplitude as colour

axelboc:
I think a single "Complex" vis with a drop-down selector in the toolbar would make more sense than separate visualizations.

axelboc:
👏

PR #499, opened 2021-02-17
Matrix and ScalarVis now supports boolean values

### LINKED ISSUE #479: [Scalar] Boolean values

loichuder:
Fixes #479 

I changed my approach compared to #495 by adding a new base type `HDF5BooleanType`. More details below.


src/h5web/guards.ts
```diff
@@ -51,9 +51,15 @@ export function assertOptionalStr(
...
-    throw new TypeError('Expected number or string');
+export function assertNumOrStrOrBool(
+  val: unknown
+): asserts val is number | string | boolean {
+  if (
+    typeof val !== 'number' &&
+    typeof val !== 'string' &&
+    typeof val !== 'boolean'
+  ) {
+    throw new TypeError('Expected number or string or boolean');
```
loichuder, line 62: We could add a new type def for `number | string | boolean`. 

Their common point is the fact that they can easily be converted and displayed as strings so `StringLikeType`, `DisplayableType` ?

src/h5web/providers/hdf5-models.ts
```diff
@@ -101,13 +101,18 @@ export enum HDF5TypeClass {
...
 export type HDF5Endianness = 'BE' | 'LE' | 'Native' | 'Not applicable';
 
+export interface HDF5BooleanType {
+  class: HDF5TypeClass.Bool;
+}
+
 export interface HDF5EnumType {
   class: HDF5TypeClass.Enum;
-  base: HDF5BaseType;
+  base: HDF5Type;
```
loichuder, line 115: As a added `HDF5BooleanType`, I realized that the distinction between `HDF5BaseType` and `HDF5AdvancedType` is our doing and has nothing to do with the `base` of advanced types. 

Perhaps we should rename it `HDF5BasicType` to avoid confusion ?

src/h5web/providers/hsds/utils.ts
```diff
@@ -71,6 +71,14 @@ export function convertHsdsType(hsdsType: HsdsType): HDF5Type {
 
   switch (hsdsType.class) {
     case HDF5TypeClass.Enum:
+      // Booleans are stored as Enum by h5py
+      // https://docs.h5py.org/en/stable/faq.html#what-datatypes-are-supported
+      if (hsdsType.mapping === { FALSE: 0, TRUE: 1 }) {
+        return {
+          class: HDF5TypeClass.Bool,
+        };
+      }
```
loichuder, line 80: As I said, boolean types are now a special type in h5web. As HSDS returns them as an enum with a specific mapping, I check the mapping here to return the appropriate h5web type.

src/h5web/providers/jupyter/utils.ts
```diff
@@ -54,19 +54,8 @@ export function convertDtype(dtype: string): HDF5Type {
...
-      base: {
-        class: HDF5TypeClass.Integer,
-        endianness: 'Not applicable',
-        size: 8,
-      },
-      mapping: {
-        FALSE: 0,
-        TRUE: 1,
-      },
+      class: HDF5TypeClass.Bool,
```
loichuder, line 58: Also, it gets simpler in the JupyterProvider as I don't need to create an enum for nothing, I simply return `HDF5BooleanType`

src/h5web/vis-packs/core/containers/ScalarVisContainer.tsx
```diff
@@ -15,6 +16,7 @@ function ScalarVisContainer(props: VisContainerProps): ReactElement {
   assertBaseType(entity);
 
   const value = useDatasetValue(entity);
+  assertNumOrStrOrBool(value);
   return <ScalarVis value={value} />;
```
loichuder, line 20: This is where it gets tricky.

The result of `useDatasetValue` is `unknown` and `ScalarVis` now expects `number | string | boolean`. According to TS, these types are incompatible as `unknown` is not assignable to `true`. 

I therefore moved the `assert` from the Vis in the Container.

src/h5web/vis-packs/core/matrix/MappedMatrixVis.tsx
```diff
@@ -28,7 +29,9 @@ function MappedMatrixVis(props: Props): ReactElement {
     [dimMapping, dims]
   );
 
-  const baseArray = useBaseArray(value, slicedDims);
+  const baseArray = useBaseArray(value, slicedDims) as ndarray<
+    string | number | boolean
+  >;
```
loichuder, line 34: Same problem of incompatible types between `ndarray<unknown>` and `ndarray<string | number | boolean>`...

I could only fix it with a cast but perhaps you have a better idea...

src/h5web/guards.ts
```diff
@@ -108,9 +114,21 @@ export function hasBaseType<S extends HDF5Shape>(
...
+      HDF5TypeClass.Float,
+      HDF5TypeClass.String,
+      HDF5TypeClass.Bool,
+    ].includes(entity.type.class)
+  );
+}
+
+export function hasBoolType<S extends HDF5Shape>(
+  dataset: Dataset<S>
+): dataset is Dataset<S, HDF5StringType> {
```
axelboc, line 128: `HDF5BooleanType` 😅

src/h5web/providers/hsds/utils.ts
axelboc: This can't possibly work, since you're comparing two different object references. Maybe just check for `hsdsType.mapping.FALSE === 0` -- seems sufficient.

src/h5web/providers/jupyter/utils.ts
axelboc: You can put the returned object on a single line if you want ;)

src/h5web/vis-packs/core/containers/ScalarVisContainer.tsx
axelboc: This is because of my new super-fancy magical typing of `useDatasetValue`. It returns `unknown` because the `HDF5BooleanType` needs to be added to the conditional type we use there. Once that's done, you won't need `assertNumOrStrOrBool`.

src/h5web/vis-packs/core/scalar/ScalarVis.tsx
```diff
@@ -1,14 +1,12 @@
...
-import { assertNumOrStr } from '../../../guards';
 
 interface Props {
-  value: string | number;
+  value: string | number | boolean;
 }
 
 function ScalarVis(props: Props): ReactElement {
   const { value } = props;
-  assertNumOrStr(value);
```
axelboc, line 11: good catch

src/h5web/providers/hsds/utils.ts
loichuder: I indeed realized this as I was writing tests :smile:

src/h5web/guards.ts
loichuder: :facepalm:

loichuder:
Fixed the `unknown` issue and added some tests

src/h5web/providers/hdf5-models.ts
axelboc: Up to you, I'm fine with both.

src/h5web/vis-packs/core/visualizations.test.ts
```diff
@@ -8,16 +8,24 @@ import {
...
 const datasetInt1D = makeSimpleDataset('dataset_int_1d', intType, [5]);
 const datasetInt2D = makeSimpleDataset('dataset_int_2d', intType, [5, 3]);
 const datasetStr2D = makeSimpleDataset('dataset_str_2d', stringType, [5, 3]);
 const datasetFlt3D = makeSimpleDataset('dataset_flt_3d', intType, [5, 3, 1]);
+const datasetBool4D = makeSimpleDataset('dataset_flt_3d', booleanType, [
+  2,
+  3,
+  3,
+  1,
+]);
```
axelboc, line 28: Does this have to be a 4D array ? 1D would suffice and maybe fit on one line 😂  (also the name of the dataset is incorrect).

axelboc (approved):

src/h5web/providers/hdf5-models.ts
loichuder: I will leave it now and remove it when working on #310

PR #500, opened 2021-02-17
Adding toggle button to reverse color maps

loichuder:
image

I can add a e2e test if needed.


src/h5web/toolbar/HeatmapToolbar.tsx
```diff
@@ -38,6 +41,12 @@ function HeatmapToolbar(): ReactElement {
       {dataDomain && <Separator />}
 
       <ColorMapSelector value={colorMap} onChange={setColorMap} />
+      <ToggleBtn
+        label="Reverse"
+        icon={FiShuffle}
+        value={reverseColorMap}
+        onChange={toggleColorMapReversal}
+      />
```
loichuder, line 49: It would make more sense to have this `ToggleBtn` inside `ColorMapSelector` but this means another level of props to pass...

axelboc (commented):
Nice and simple 💯 

You're going to hate me, but I wonder whether "invert" would make more sense than "reverse"... You'd then have `invertColorMap` and `toggleColorMapInvestion` in the config, for instance, and "Invert" (or maybe even just "Inv.") as the toggle button label.

Also, since the toggle is a sub-control of the colour map selector, you should try to use the same styles as the "Auto" toggle. It's slightly smaller and more rounded.

Finally, I think an E2E test is worthwhile. 👍

src/stories/HeatmapVis.stories.tsx
```diff
@@ -142,6 +142,14 @@ WithTitle.args = {
   title: 'Pretty colors',
 };
 
+export const invertColorMap = Template.bind({});
```
axelboc, line 145: Needs to be capitalised.

axelboc (approved):

PR #501, opened 2021-02-17
Implement HSL color scale

### LINKED ISSUE #481: [Heatmap] HSV color map
A user has requested the [HSV color map](https://fr.mathworks.com/help/matlab/ref/hsv.html).

It is not currently implemented in [d3-scale-chromatic](https://github.com/d3/d3-scale-chromatic). We can request it ... or try implementing it ourselves 🤷.

axelboc:
Fixes #481.

HSL/HSV are the same thing since we only care about the hue. As discussed, since `Sinebow` is an improved version of `HSL`, I've put the former into the _Common_ group instead of `HSL`.

image

image

### DIFF
--- a/src/h5web/vis-packs/core/heatmap/interpolators.ts
+++ b/src/h5web/vis-packs/core/heatmap/interpolators.ts
@@ -1,3 +1,4 @@
+import { hsl } from 'd3-color';
 import {
   interpolateCool,
   interpolateMagma,
@@ -38,12 +39,17 @@ import {
   interpolateTurbo,
 } from 'd3-scale-chromatic';
 
+function interpolateHsl(t: number): string {
+  return hsl(t * 360, 1, 0.5).formatRgb();
+}
+
 export const MOST_USED = {
   Viridis: interpolateViridis,
   Inferno: interpolateInferno,
   Greys: interpolateGreys,
   RdBu: interpolateRdBu,
   Rainbow: interpolateRainbow,
+  Sinebow: interpolateSinebow,
 };
 
 export const SINGLE_HUE = {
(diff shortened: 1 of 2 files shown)


src/h5web/vis-packs/core/heatmap/interpolators.ts
```diff
@@ -81,6 +87,7 @@ export const MULTI_HUE = {
 export const CYCLICAL = {
   Rainbow: interpolateRainbow,
   Sinebow: interpolateSinebow,
+  HSL: interpolateHsl,
```
axelboc, line 90: I preferred to call it `HSL` since we use D3's `hsl` function...

src/h5web/vis-packs/core/heatmap/interpolators.ts
```diff
@@ -38,12 +39,17 @@ import {
   interpolateTurbo,
 } from 'd3-scale-chromatic';
 
+function interpolateHsl(t: number): string {
+  return hsl(t * 360, 1, 0.5).formatRgb();
+}
+
 export const MOST_USED = {
```
loichuder, line 46: Could we rename this in `COMMON` ?

loichuder (approved):

src/h5web/vis-packs/core/heatmap/interpolators.ts
axelboc: Or should we change the label of the group in the drop-down?

axelboc: Ah no, we talked about it, I remember. It sounded like it was the most used by the user. I'll rename the constant 👍

PR #502, opened 2021-02-17
Renamed domain (in the file sense) into filepath

loichuder:


axelboc (approved):
It's used in a lot more places than I thought. Thanks for remembering 💯

PR #507, opened 2021-02-18
Added support for complex in Matrix and ScalarVis

loichuder:
First step of #310. I indeed plan to implement complex-specific visus after that.

I perhaps went a bit overboard with the amount of change in this one :sweat_smile: 

A digest:
- I used the same trick as in #499 as I added a new complex type in h5web. The fact that a dataset contains complex data is checked in the providers to convert the type appropriately to the new complex type.
- I removed the distinction between `HDF5BaseType` and `HDF5AdvancedType` as it was artificial. I instead added `PrintableType` union that contains the types that can be displayed as strings (including the `HDF5ComplexType` now). I recycled the `HDF5BaseType` guards to check for this `PrintableType` for the `MatrixVis` and `ScalarVis`.
-  I needed a unified way to represent complex numbers h5web. I could have stuck with strings but anticipating the complex-specific operations to come, I used [complex.js](https://www.npmjs.com/package/complex.js) to have a dedicated `Complex` type.


loichuder:
Waiting for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51305 to be published to fix the lint but happy to get your review in the meantime @axelboc

src/h5web/providers/jupyter/models.ts
```diff
@@ -41,3 +41,9 @@ export type JupyterContentResponse =
 export type JupyterDataResponse = HDF5Value;
 
 export type JupyterAttrsResponse = Record<string, string>;
+
+export type JupyterComplex = JupyterComplexArray | JupyterComplexValue;
+export type JupyterComplexValue =
+  | `(${'' | '-'}${number}${'+' | '-'}${number}j)`
+  | `${'' | '-'}${number}j`;
+export type JupyterComplexArray = JupyterComplex[];
```
axelboc, line 49: Same, I don't think having `JupyterComplexArray` is required. `JupyterComplex[]` works just as well.

src/h5web/providers/models.ts
```diff
@@ -50,7 +52,7 @@ export interface Link<T extends HDF5Link = HDF5Link> extends Entity {
 
 type PrimitiveType<T extends HDF5Type> = T extends HDF5NumericType
   ? number
-  : T extends HDF5StringType
+  : T extends HDF5StringType | HDF5ComplexType
```
axelboc, line 55: I would rather see a separate logical branch that returns `Complex`. Here you're saying that the value is a `string`, which is wrong. `ScalarVis` and `MatrixVis` should explicitely accept `Complex` and `Complex[]` values.

src/h5web/providers/models.ts
```diff
@@ -60,3 +62,5 @@ export type Value<
   S extends HDF5Shape,
   T extends HDF5Type
 > = S extends HDF5SimpleShape ? PrimitiveType<T>[] : PrimitiveType<T>;
+
+export type ComplexArray = (ComplexArray | Complex)[];
```
axelboc, line 66: With my comment above, I think this will type no longer be relevant. It should be achieved with type `Value` above.

axelboc (changes requested):

axelboc (approved):

Issue #510, opened 2021-02-18
Rethink domain slider UI


axelboc:
The domain slider UI suffers from a few limitations:

- The values displayed in the thumbs sometimes don't match the actual values applied.
- The thumbs change size based on their values, which causes rendering issues with `ReactSlider`.
- The range is limited to +/-10% of the current dataset's domain (technically, the domain of the dataset that was selected when auto-scale was turned off -- see #486)
- We have no clue what the min/max values of the current dataset are.
- We can't auto-scale the min/max bounds individually from the other.
- We can't type in min/max bounds manually

To resolve these issues, we suggest the following interface:

### Mockups (v3.1)

![domain slider v3](https://user-images.githubusercontent.com/2936402/108857646-960b0100-75eb-11eb-941b-a62908fbf4d9.png)

<details>
![domain slider v2](https://user-images.githubusercontent.com/2936402/108846642-17a86200-75df-11eb-87d1-61e6f8a1bab5.png)

<summary>Previous versions</summary>
![image](https://user-images.githubusercontent.com/2936402/108386270-c843e980-720c-11eb-8335-6505b2e2d252.png)

![domain slider](https://user-images.githubusercontent.com/2936402/108715212-d51d5180-751a-11eb-9669-58f34f94a520.png)
</details>

### Specifications

- [x] The thumbs don't show values. #524 
- [x] The range of the slider is always an extension of the currently applied domain -- in other words, after clicking/dragging/releasing a thumb, it always comes back to its initial position and the thumbs never stay on top of each other, even when switching dataset with totally different data domains. #534 
  - [x] The only exception is if the two thumbs receive the same value. In this case, they end up in the middle of the slider, one on top of the other. #534 
- [x] If the data domain intersects with the range of the slider, the intersecting range is shown in dark on the slider's bar. #530 
- [x] When the slider is hovered, a popup appears that: #521
  - [x] shows the currently applied min/max bounds, as well as the min/max values of the data; #521
  - [x] allows entering custom min/max values (with an inline editing UI); #557 
  - [x] allows autoscaling the min/max bounds. #521
- ~The "Auto" toggle next to the slider toggles the autoscale of both bounds together, but if auto-scale is turned on or off for a specific bound from the popup, then the toggle shows "Auto (min)" or "Auto (max)". Clicking the toggle in this state turns auto-scale off for the corresponding bound.~
- [x] Moving a thumb or entering (and applying) a custom value for a bound turns autoscale off for this bound. #529 #557 
[...]

Issue #486, opened 2021-02-15
Domain slider bounds should update even when auto-scale is off


axelboc:

axelboc:
For the record, I'm not a fan of the behaviour. I think a bigger overhaul of the slider is required, as we've discussed.

axelboc:
Superseded by #510

PR #513, opened 2021-02-19
Refactor Heatmap config to support auto-scaling a single domain bound

axelboc:
Preparation for #510 

I change the `customDomain` state to `[number | undefined, number | undefined]` instead of `Domain | undefined`. This change is required if we want to support auto-scaling the domain bounds independently.

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider.tsx
@@ -2,7 +2,7 @@ import type { ReactElement } from 'react';
 import ReactSlider from 'react-slider';
 import { format } from 'd3-format';
 import styles from './DomainSlider.module.css';
-import type { Domain } from '../../vis-packs/core/models';
+import type { CustomDomain, Domain } from '../../vis-packs/core/models';
 import { extendDomain } from '../../vis-packs/core/utils';
 import ToggleBtn from './ToggleBtn';
 import { FiZap } from 'react-icons/fi';
@@ -12,9 +12,9 @@ const NB_DECIMALS = 1;
 
 interface Props {
   dataDomain: Domain;
-  value?: Domain;
+  value: CustomDomain;
   disabled?: boolean;
-  onChange: (value: Domain | undefined) => void;
+  onChange: (value: CustomDomain) => void;
 }
 
 function DomainSlider(props: Props): ReactElement {
@@ -22,6 +22,10 @@ function DomainSlider(props: Props): ReactElement {
 
   const [extendedMin, extendedMax] = extendDomain(dataDomain, EXTEND_FACTOR);
   const step = Math.max((extendedMax - extendedMin) / 100, 10 ** -NB_DECIMALS);
+
+  const isAutoMin = value[0] === undefined;
+  const isAutoMax = value[1] === undefined;
+  const isAutoScaling = isAutoMin || isAutoMax;
 
   return (
     <div className={styles.sliderWrapper}>
@@ -38,7 +42,7 @@ function DomainSlider(props: Props): ReactElement {
             </div>
           </div>
         )}
-        value={[...(value || dataDomain)]}
+        value={[value[0] ?? dataDomain[0], value[1] ?? dataDomain[1]]}
         onAfterChange={(bounds) => {
           onChange(bounds as Domain);
         }}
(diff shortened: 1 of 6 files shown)


src/h5web/toolbar/controls/DomainSlider.tsx
```diff
@@ -23,6 +24,8 @@ function DomainSlider(props: Props): ReactElement {
   const [extendedMin, extendedMax] = extendDomain(dataDomain, EXTEND_FACTOR);
   const step = Math.max((extendedMax - extendedMin) / 100, 10 ** -NB_DECIMALS);
 
+  const isAutoScaling = value.some(isUndefined);
```
loichuder, line 27: You might want to use the `isDefined` already defined in `h5web/guards`

loichuder: Or:
```js
const isAutoMin = value[0] === undefined && value[1] !== undefined
const isAutoMax = value[0] !== undefined && value[1] === undefined
const isAutoScaling = isAutoMin || isAutoMax
```

and then, you can use `isAutoMin` and `isAutoMax` to make the nested ternary below more readable

src/h5web/toolbar/controls/DomainSlider.tsx
```diff
@@ -50,11 +53,17 @@ function DomainSlider(props: Props): ReactElement {
...
       <ToggleBtn
         small
-        label="Auto"
+        label={`Auto${
+          value[0] === undefined && value[1] !== undefined
+            ? ' (min)'
+            : value[0] !== undefined && value[1] === undefined
+            ? ' (max)'
+            : ''
+        }`}
```
loichuder, line 62: According to my comment above, this would be:
```js
       label={`Auto${
          isAutoMin
            ? ' (min)'
            : isAutoMax
            ? ' (max)'
            : ''
        }`}
```

loichuder (approved):

src/h5web/toolbar/controls/DomainSlider.tsx
axelboc: Good idea, though in your suggestion, `isAutoScaling` doesn't end up with the same value. This works great, however:

```ts
const isAutoMin = value[0] === undefined;
const isAutoMax = value[1] === undefined;
const isAutoScaling = isAutoMin || isAutoMax;
```

src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -55,17 +60,12 @@ function MappedHeatmapVis(props: Props): ReactElement {
...
+    useDomain(dataArray.data as number[], scaleType) || DEFAULT_DOMAIN;
 
   useEffect(() => {
-    if (!customDomain) {
-      // If auto-scale is on, update `dataDomain`
-      setDataDomain(domain);
-    }
-  }, [customDomain, domain, setDataDomain]);
+    setDataDomain(dataDomain);
+  }, [dataDomain, setDataDomain]);
```
axelboc, line 71: Now, I always set the `dataDomain` in the config whenever it changes (i.e. when we switch dataset or scale type) -- I'm basically fixing #486, ahead of #510.

src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -56,16 +62,13 @@ function MappedHeatmapVis(props: Props): ReactElement {
   const baseArray = useBaseArray(value, slicedDims);
   const dataArray = useMappedArray(baseArray, slicedMapping);
 
-  const domain = useMemo(() => {
-    return customDomain || getDomain(dataArray.data as number[], scaleType);
-  }, [customDomain, dataArray, scaleType]);
+  const dataDomain =
+    useDomain(dataArray.data as number[], scaleType) || DEFAULT_DOMAIN;
+  const visDomain = useVisDomain(dataDomain, customDomain);
```
axelboc, line 67: This `useVisDomain` hook will evolve to check whether the domain is "valid". Right now it's just a `useMemo` with the `??` logic that was previously in the JSX below. I might end up using it in the domain slider itself too.

Note that I can't get `dataDomain` from the config inside the hook without risking a stale render. I have to pass it as an argument.

src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -56,16 +62,13 @@ function MappedHeatmapVis(props: Props): ReactElement {
   const baseArray = useBaseArray(value, slicedDims);
   const dataArray = useMappedArray(baseArray, slicedMapping);
 
-  const domain = useMemo(() => {
-    return customDomain || getDomain(dataArray.data as number[], scaleType);
-  }, [customDomain, dataArray, scaleType]);
+  const dataDomain =
+    useDomain(dataArray.data as number[], scaleType) || DEFAULT_DOMAIN;
```
axelboc, line 66: I need `|| DEFAULT_DOMAIN` in order to compute the `visDomain` below. I can't wait for `dataDomain` to be set in the config with the `useEffect`. I may eventually move the fallback into `useDomain`, if it also makes sense for the line vis.

loichuder (approved):

PR #514, opened 2021-02-19
Adding more CSS variables

loichuder:


src/styles/vars.css
```diff
@@ -26,4 +27,7 @@
     sans-serif;
   --btn-shadow-idle: -1px -1px 2px 0 inset;
   --btn-shadow-pressed: 1px 1px 2px 0 inset;
+  --btn-shadow-color: gray;
+  --btn-shadow-dark-color: #696969;
+  --btn-shadow-darker-color: darkslategray;
```
axelboc, line 32: Are these available as variables? If so, you can still refer to them instead of duplicating the colours. It won't prevent you from overriding them in JLab.

axelboc (approved):

Issue #470, opened 2021-02-05
[NeXus] Auxiliary signal (multiple curves)


axelboc:

loichuder:
- [] `useNxData` should get `auxiliary_signals` attributes and prefetch the values
- [] `MappedLineVis` and `LineVis` should accept several multiple `ndarrays` for multiple curves

PR #517, opened 2021-02-23
useNxData now gets auxiliary_signals attr and associated datasets

loichuder:
First step of #470

### DIFF
--- a/src/h5web/vis-packs/nexus/hooks.ts
+++ b/src/h5web/vis-packs/nexus/hooks.ts
@@ -48,16 +48,17 @@ function useTitleDataset(
   return dataset;
 }
 
-function useAxesDatasets(
-  group: Group
+function useAssociatedDatasets(
+  group: Group,
+  type: 'axes' | 'auxiliary_signals'
 ): (Dataset<HDF5SimpleShape, HDF5NumericType> | undefined)[] {
   const { valuesStore } = useContext(ProviderContext);
 
-  const axisList = getAttributeValue(group, 'axes') || [];
-  const axisNames = typeof axisList === 'string' ? [axisList] : axisList;
-  assertArray<string>(axisNames);
+  const dsetList = getAttributeValue(group, type) || [];
+  const dsetNames = typeof dsetList === 'string' ? [dsetList] : dsetList;
+  assertArray<string>(dsetNames);
 
-  return axisNames.map((name) => {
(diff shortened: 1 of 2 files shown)


src/h5web/vis-packs/nexus/hooks.ts
```diff
@@ -102,3 +106,31 @@ export function useAxisMapping(
...
+      ) {
+        console.warn(`${name} is not a valid auxiliary signal`); // eslint-disable-line no-console
+        return undefined;
+      }
+
+      valuesStore.prefetch({ path: dataset.path });
+      return dataset;
+    })
+    .filter(isDefined);
+}
```
loichuder, line 136: This hook is highly similar to `useAxesDatasets` but there are slight differences that made me prefer to write a separate hook.

axelboc: Yeah, makes sense since the return type is slightly different.

Why this algorithm, though? If `auxiliary_signals` contains the string `foo` and `foo` is not a valid auxiliary signal dataset, then shouldn't things break like they do with `axes`? At least the two behaviours should be consistent.

My vote goes to breaking hard because I prefer the `NxImage` visualization to not work at all rather than to work only partially. If it works but shows something the author did not intend, then users might think that everything is working as expected. Instead, if users see an error, they may try to get it fixed in the file. Also, not seeing the `NxImage` visualization doesn't prevent them from visualizing the raw data as a heatmap.

Another reason why I think breaking with an error is okay here is because all three of the four assertions would be highly unlikely to occur in practice (i.e. pointing to an entity that is not a dataset with a simple shape and numeric type). The most likely error would be a typo in the name of an auxiliary dataset, in which case I want authors to know as soon as possible that they've made a typo. If they check their file in `jupyterlab-h5web` and we don't show an error, they may not notice...

src/h5web/vis-packs/nexus/hooks.ts
```diff
@@ -102,3 +106,31 @@ export function useAxisMapping(
     );
   });
 }
+
+function useAuxiliaryDatasets(
+  group: Group
+): Dataset<HDF5SimpleShape, HDF5NumericType>[] {
```
axelboc, line 112: This function should go above `useNxData`.

axelboc (approved):
Looks good 👍

PR #521, opened 2021-02-25
Implement domain slider tooltip

axelboc:
Part of #510. The tooltip includes:

- the currently applied min/max domain bounds;
- the domain of the data;
- toggles to autoscale the min/max bounds separately.

image

image

The tooltip displays values rounded to 3 significant digits, but the raw values are displayed on hover.

image


axelboc:
The overflow menu now shares the same design as the domain tooltip:

image

Note that the background is very slightly transparent to not completely hide the visualization underneath.

src/h5web/toolbar/OverflowMenu.tsx
```diff
@@ -46,18 +46,20 @@ function OverflowMenu(props: Props): ReactElement {
         </span>
       </button>
 
-      <ul
+      <div
```
axelboc, line 49: This new wrapper is not technically needed for the overflow menu, but is required for the tooltip in order to have an invisible padding gap between the toolbar and the popup that doesn't "lose" the hover (unlike a gap that would created by absolutely positioning the popup away from the toolbar).

src/h5web/toolbar/Toolbar.module.css
```diff
@@ -56,6 +56,11 @@
   font-size: 0.875em;
 }
 
+.btn[data-raised] > .btnLike {
+  background-color: var(--white);
+  box-shadow: 0 0 1px var(--btn-shadow-color);
+}
```
axelboc, line 62: I needed a toggle with a background when idled (which the toggles in the toolbar don't have).

src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
```diff
@@ -0,0 +1,123 @@
+.root {
```
axelboc, line 1: `DomainSlider` and its CSS are moved into a subfolder.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
```diff
@@ -0,0 +1,123 @@
...
+}
+
+.autoBtnLike {
+  composes: btnLike from '../../Toolbar.module.css';
+  font-size: 0.875em;
+  border-radius: 0.875rem;
+  color: var(--black);
+}
+
+.tooltip {
```
axelboc, line 58: New code from here on 👇

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -0,0 +1,107 @@
...
+  if (isAutoMin && !isAutoMax) {
+    return 'Min';
+  }
+
+  if (isAutoMax && !isAutoMin) {
+    return 'Max';
+  }
+
+  return 'Auto';
+}
```
axelboc, line 28: This used to be implemented with nested ternaries inside `DomainSlider`. Extracting it is how I managed to get rid of the ESLint warning on cognitive complexity. I've kept the function here because it will be removed in a future PR, since the "Auto" toggle will be replaced with an edit button.

axelboc: BTW, when the label of the button changes, the slider moves on the toolbar. That's why I've shortened the labels to "Min" and "Max", so it's closer in length to "Auto". Another good reason to remove this toggle completely.

src/h5web/vis-packs/core/heatmap/ColorBar.module.css
```diff
@@ -51,6 +51,7 @@
   justify-content: center;
   line-height: 1;
   font-size: 0.875em;
+  white-space: nowrap;
```
axelboc, line 54: For consistency, I've changed the formatting of the bounds on the colour map to use exponent notation, and the browser wraps the value at `+/-` if there's not enough space. We want the bounds to overflow without wrapping.

src/h5web/vis-packs/core/heatmap/ColorBar.tsx
```diff
@@ -8,7 +8,7 @@ import type { ScaleType, Domain } from '../models';
 import type { ColorMap } from './models';
 import { format } from 'd3-format';
 
-const boundFormatter = format('.3~g');
+const boundFormatter = format('.3~e');
```
axelboc, line 11: I'll move this formatter in a utility later on so it can be shared.

src/styles/base.css
```diff
@@ -43,6 +43,12 @@ pre {
   font-family: var(--monospace);
 }
 
+abbr[title] {
+  cursor: help;
+  border-bottom: 1px dotted;
+  text-decoration: none;
+}
```
axelboc, line 50: The default styling is an underline, which is a bit too close to the text...

src/styles/vars.css
```diff
@@ -19,6 +20,7 @@
   --secondary-darker: #0e5846;
   --secondary-bg: #d9f4ec;
   --secondary-light-bg: #ecfaf6;
+  --secondary-light-bg-90: #ebfaf5e6;
```
axelboc, line 23: Two more variables to override in JLab.

loichuder (commented):
I think the `DomainSlider` is way too complex: I had to read it several times and I still do not get it (**and** I already took my coffee :coffee: )

This needs further discussion

PR #527, opened 2021-03-01
Use scale to interface with domain slider

axelboc:
Well, this was surprisingly easy.

As you suggested, @loichuder, the slider now always goes from `1` to `100` with a step of `1`, and we use a D3 scale to map the applied domain to values on the slider and back.

Our `ReactSlider` now works only with integers -- the scale takes care of rounding the values before passing them to the slider. When the slider's values change, the scale interpolates them back into our extended domain without any rounding. Thus, we bypass `ReactSlider`'s annoying [`toFixed(5)`](https://github.com/zillow/react-slider/blob/master/src/components/ReactSlider/ReactSlider.jsx#L83) and can support very small domains with a precision of `Number.EPSILON * 100`:

image

Nothing else changes, but this will make updating only one bound instead of both when moving a thumb (for the autoscale) much more robust, since we'll be able to compare the slider's integer values instead of floats rounded to 5 decimals...


src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -36,21 +33,32 @@ interface Props {
...
   const { dataDomain, customDomain, disabled, onCustomDomainChange } = props;
 
-  const [appliedDomain, setAppliedDomain] = useAppliedDomain(
-    dataDomain,
-    customDomain
-  );
-
-  const [tooltipOpen, toggleTooltip] = useToggle(false);
-  useKey('Escape', () => toggleTooltip(false));
+  const appliedDomain = useAppliedDomain(dataDomain, customDomain);
```
axelboc, line 36: Note that I now have an `appliedDomain` and a `sliderDomain`. It's the latter that I update in real time when a thumb is moving. Not really necessary yet, but still clearer. It will be more useful later, in my next PR.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -36,21 +33,32 @@ interface Props {
...
+    },
+    [sliderScale, setSliderDomain]
+  );
+
+  const handleSliderAfterChange = useCallback(
+    ([min, max]) => {
+      onCustomDomainChange([sliderScale.invert(min), sliderScale.invert(max)]);
+    },
+    [sliderScale, onCustomDomainChange]
+  );
```
axelboc, line 60: `handleSliderAfterChange` will diverge more from `handleSliderChange` in my next PR, which is why I didn't try to abstract them.

axelboc:
FYI, the snapshot diffs all look like this:

heatmap_2D diff

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
loichuder: I get that it is bound to change but can you add a small comment explaining the difference between `sliderDomain` and `appliedDomain` ?

It will be a useful memo (at least for me!)

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -65,13 +69,12 @@ function DomainSlider(props: Props): ReactElement {
         thumbClassName={styles.thumb}
         thumbActiveClassName={styles.thumbActive}
         pearling
-        value={appliedDomain}
-        min={extendedMin}
-        max={extendedMax}
-        step={step}
+        min={0}
+        max={100}
```
loichuder, line 73: Ideally, `min` and `max` should be set from `sliderScale.range`

src/h5web/toolbar/controls/DomainSlider/hooks.ts
```diff
@@ -1,18 +1,18 @@
...
+export const useSliderScale = createMemo(getSliderScale);
+
+export function useSliderDomain(appliedDomain: Domain) {
+  const state = useState(appliedDomain);
+  const [, setDisplayedDomain] = state;
 
   useEffect(() => {
-    setAppliedDomain(getSliderDomain(dataDomain, customDomain));
-  }, [customDomain, dataDomain, setAppliedDomain]);
+    setDisplayedDomain(appliedDomain);
```
loichuder, line 14: `setDomain` perhaps ?

I feel that there are already enough `...Domain` in there...

src/h5web/toolbar/controls/DomainSlider/hooks.ts
```diff
@@ -1,18 +1,18 @@
...
-export function useAppliedDomain(
-  dataDomain: Domain,
-  customDomain: CustomDomain
-) {
-  const state = useState(getSliderDomain(dataDomain, customDomain));
-  const [, setAppliedDomain] = state;
+export const useAppliedDomain = createMemo(getAppliedDomain);
+export const useSliderScale = createMemo(getSliderScale);
+
+export function useSliderDomain(appliedDomain: Domain) {
```
loichuder, line 9: Wild idea: do you think `useSliderScale` and `useSliderDomain` could be merged in a single hook ?

loichuder (changes requested):
I really find the `sliderDomain` term misleading in this context...

Could we abstract the scale in a `ScaledSlider` child component of `DomainSlider` ?

We could then:
- keep the name `sliderDomain` as it would be passed to `ScaledSlider` and make much more sense in this context) down to this slider and 
- make `DomainSlider` oblivious of the scale manipulations

src/h5web/toolbar/controls/DomainSlider/hooks.ts
axelboc: woops, sorry just a left over from a previous naming 😉

axelboc: `setState` even

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
axelboc: Ha, good point, especially since `min` should be `1` here...

src/h5web/toolbar/controls/DomainSlider/hooks.ts
axelboc: Hmm I don't see the benefit of squeezing those two hooks into one, sorry. Seems nice that `useSliderDomain` matches the signature of `useState`.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
axelboc: I could also rename this `visDomain` and `useVisDomain` 🤷

axelboc: But then will it make sense in the mapped visualization components? Hmm yeah, I think so. What do you think?

axelboc:
> Could we abstract the scale in a ScaledSlider child component of DomainSlider ?

It did cross my mind. It will lighten up `DomainSlider` as well, so I'll give it a try 👍

src/h5web/toolbar/controls/DomainSlider/hooks.ts
axelboc: Also, `useSliderScale` will ow be called in `ScaledSlider`, whereas `useSliderDomain` will stay in `DomainSlider`.

axelboc:
I've pushed a separate commit with all the changes. Note that I've removed `useSliderDomain` in the end -- `DomainSlider` is a bit lighter so it can handle a `useEffect`.

axelboc:
I got rid of the `useCallback`s too. I've made a POC to try to fix the autoscale issue when a thumb is moved, and it looks like they were a premature optimisation.

src/h5web/toolbar/controls/DomainSlider/hooks.ts
loichuder: No problem, I just thought it was worth mentioning it

src/h5web/toolbar/controls/DomainSlider/utils.ts
```diff
@@ -1,8 +1,37 @@
...
+    round: true,
+  });
+
+  function scaleDomain(domain: Domain) {
+    return domain.map(scale) as Domain;
+  }
+
+  scaleDomain.invert = function scaleInvDomain(domain: Domain) {
+    return domain.map((val) => scale.invert(val)) as Domain;
+  };
```
loichuder, line 34: Now that the scale is abstracted in `ScaledSlider`, do you think it is worth keeping these functions ?

My point is that having `map` explicitly in `ScaledSlider` would be simpler to understand.

On the other hand, it is an implementation detail...

I won't push hard for any of the options :sweat_smile:

loichuder (approved):
I really prefer the version with the `ScaledSlider` ! Well done :+1: 

I also prefer `visDomain`. All in all, I think it is much clearer now !

src/h5web/toolbar/controls/DomainSlider/utils.ts
axelboc: Neh, you're right. Also, the scale no longer needs to be memoized now that I've removed the callbacks. The fact that I was extending the domain inside `useSliderScale` was a bit convoluted too... So I've removed the hook completely and inlined the domain extension, the scale creation and the `map`s in `SliderScale`.

src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  rawDomain: Domain;
+  dataDomain: Domain;
+  disabled?: boolean;
+  isAutoMin: boolean;
+  isAutoMax: boolean;
+  onChange: (domain: Domain) => void;
+  onAfterChange: (domain: Domain) => void;
+}
+
+const EXTEND_FACTOR = 0.2;
```
loichuder, line 21: This should go above the `interface Props` with `SLIDER_RANGE` or both should go under `interface Props`

src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <ReactSlider
+      className={styles.slider}
+      trackClassName={styles.track}
+      thumbClassName={styles.thumb}
+      thumbActiveClassName={styles.thumbActive}
+      pearling
+      min={SLIDER_RANGE[0]}
+      max={SLIDER_RANGE[1]}
```
loichuder, line 44: Does TS complain if you use `scale.range[0]` and `scale.range[1]` here ?

src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  isAutoMin: boolean;
+  isAutoMax: boolean;
+  onChange: (domain: Domain) => void;
+  onAfterChange: (domain: Domain) => void;
+}
+
+const EXTEND_FACTOR = 0.2;
+
+function ScaledSlider(props: Props): ReactElement {
+  const { rawDomain, dataDomain, disabled, isAutoMin, isAutoMax } = props;
```
loichuder, line 24: `rawDomain` could be `unscaledValue` or even `value` as we don't need to know that a scale is used then ? :shrug:

loichuder (approved):

src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
axelboc: Now, I should be able to (I couldn't before, because I was returning a domain scale)

src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
axelboc: It was a bit too confusing when I had `useCallback`s, but now it's fine, you're right 👍

src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
axelboc: image

I find it more readable with the constant, anyway.

loichuder: I have no issue keeping the constant as well. It was out of curiosity :monocle_face:

PR #529, opened 2021-03-02
Update only the domain bounds that change when interacting with the slider

axelboc:
Not as straightforward as I was hoping... As a reminder, the initial goal was **to update only the `customDomain` bound that changed after moving a thumb**, so that the other bound could remain `undefined` (and therefore autoscaled) if applicable.

There were a couple of unforeseen challenges that made this more complicated than expected:

- The bound that doesn't change must also **_not_ be updated in `sliderDomain` _during_ the interaction**.

  Initially, I had `ScaledSlider` scale-invert both bounds and send the result to `DomainSlider` to set as is as the new `sliderDomain`. The problem is that `sliderDomain` is displayed in the popup (for real-time feedback when moving a thumb), and `scale()` rounds to the nearest integer, so `scale.invert()` doesn't always give back the same value... As a result, I'd see the value of the other bound change at the start of the interaction and be reset at the end...

- Because `ScaledSlider` is a controlled component, when `onAfterChange` is called, the slider's value that we get as argument to this callback has already been updated with the latest change, so it's impossible to determine as is which bound has changed in `onAfterChange`.

  To work around this, I had to use a couple of boolean states. I set them to `true` in `onChange`, and then reset them to `false` in `onAfterChange`.

- Because we enabled `pearling` on the slider, one thumb can push another during an interaction. Therefore, the two bounds can actually change at the same time, which means that sometimes, we do need to update both values in `sliderDomain` and `customDomain` during and after an interaction...

  That's why I use two separate boolean state and pass two separate arguments to the callbacks. Managing all this with a single enum value of some sort would have been too convoluted. That being said, the `onChange` callback could be smarter: I could pass a `CustomDomain` and convey that a bound has not changed by setting it to `undefined`. However, I'd like your opinion before implementing this solution, as you may find the current implementation satisfactory. Personally, I think two booleans make for a clearer callback signature, so I'm happy as it is.

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
@@ -61,7 +61,12 @@ function DomainSlider(props: Props): ReactElement {
         isAutoMax={isAutoMax}
         disabled={disabled}
         onChange={setSliderDomain}
-        onAfterChange={onCustomDomainChange}
+        onAfterChange={(hasMinChanged, hasMaxChanged) => {
+          onCustomDomainChange([
+            hasMinChanged ? sliderDomain[0] : customDomain[0],
+            hasMaxChanged ? sliderDomain[1] : customDomain[1],
+          ]);
+        }}
       />
 
       <ToggleBtn
--- a/src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
@@ -1,6 +1,7 @@
 import type { ReactElement, Ref } from 'react';
 import { FiSkipBack, FiSkipForward } from 'react-icons/fi';
 import ReactSlider from 'react-slider';
+import { useBoolean } from 'react-use';
 import { Domain, extendDomain, ScaleType } from '../../../../packages/lib';
 import { createAxisScale } from '../../../vis-packs/core/utils';
 import styles from './DomainSlider.module.css';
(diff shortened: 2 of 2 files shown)


src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -60,8 +60,18 @@ function DomainSlider(props: Props): ReactElement {
...
         disabled={disabled}
-        onChange={setSliderDomain}
-        onAfterChange={onCustomDomainChange}
+        onChange={(newValue, hasMinChanged, hasMaxChanged) => {
+          setSliderDomain([
+            hasMinChanged ? newValue[0] : sliderDomain[0],
+            hasMaxChanged ? newValue[1] : sliderDomain[1],
+          ]);
+        }}
+        onDone={(hasMinChanged, hasMaxChanged) => {
```
axelboc, line 69: No need to pass a `newValue` here, since by the time `onDone` is called, `sliderDomain` has already been updated with the latest value.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -60,8 +60,18 @@ function DomainSlider(props: Props): ReactElement {
...
         isAutoMax={isAutoMax}
         disabled={disabled}
-        onChange={setSliderDomain}
-        onAfterChange={onCustomDomainChange}
+        onChange={(newValue, hasMinChanged, hasMaxChanged) => {
+          setSliderDomain([
+            hasMinChanged ? newValue[0] : sliderDomain[0],
+            hasMaxChanged ? newValue[1] : sliderDomain[1],
+          ]);
+        }}
```
loichuder, line 68: You could keep `onChange={setSliderDomain}` here and put the ternaries inside `ScaledSlider`.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
loichuder: Why not keeping `onAfterChange` ?

loichuder (approved):
The implementation seems fine and quite understandable for me. 

Most of it is contained in `ScaledSlider` which is good.

PR #530, opened 2021-03-03
Show data domain on slider's track

axelboc:
image

image

... and a little bit of transparency on hover, to see the `dataDomain` range:

image


src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
```diff
@@ -23,13 +23,24 @@
...
 
+.dataTrack {
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  background-color: var(--secondary-dark);
+  box-shadow: 0 -2px 3px inset var(--secondary-darker);
+  border-radius: 2px;
+}
+
```
loichuder, line 43: Perhaps we should start breaking the CSS down in multiple files

src/h5web/toolbar/controls/DomainSlider/Track.tsx
```diff
@@ -0,0 +1,29 @@
...
+interface Props {
+  scale: AxisScale;
+  dataDomain: Domain;
+}
+
+function Track(props: Props): ReactElement {
+  const { scale, dataDomain } = props;
+
+  const [min, max] = dataDomain;
+  const cssScale = scale.copy().range([0, 100]) as AxisScale;
```
loichuder, line 14: Why do you need the cast here ?

loichuder (changes requested):
I like the transparent thumbs, I think it is a great idea !

However, they seem to be placed so that the edge matches the track. I would prefer if the center matched the track.

The rest looks good !

src/h5web/toolbar/controls/DomainSlider/Track.tsx
axelboc: Because the types of the chained methods in D3 aren't great. `visx` doesn't go as far as fixing that, unfortunately.

PR #533, opened 2021-03-04
Align track to middle of thumbs in domain slider

axelboc:
I'm much happier with this solution: I make sure the thumb containers always have the same measured size so `ReactSlider` doesn't go crazy, and I simply add a wrapper around the track with some padding (equal to half the width of a thumb on either side).

The autoscaled thumbs are technically misaligned by a tiny pixel, but since they are fully opaque and they stop being autoscaled right after you move them, it doesn't really matter. Aligning them properly would requiring reducing the track by 1 or 2 pixels when a thumb becomes autoscaled, which would look weird, I think. (We could compensate by increasing the size of the slider when this happens, but this would require a lot of JS code for not much gain...)


src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
```diff
@@ -27,11 +27,16 @@
...
   align-self: center;
   height: 0.5rem;
+  margin: 0 calc(1.375rem / 2); /* make track end in middle of thumbs */
   background-color: var(--secondary-dark-15);
   box-shadow: var(--btn-shadow-pressed) var(--btn-shadow-color);
   border-radius: 0.25rem;
 }
 
+.track[data-automin] {
+  margin: 0 calc(1.125rem / 2); /* make track end in middle of thumbs when auto-scaling min and/or max */
```
axelboc: I'm gonna be honest, I'm not 100% sure why this works.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
```diff
@@ -27,11 +30,17 @@
...
   height: 0.5rem;
+  margin: 0 calc(var(--thumb-size-default) / 2); /* make track end in middle of thumbs */
   background-color: var(--secondary-dark-15);
   box-shadow: var(--btn-shadow-pressed) var(--btn-shadow-color);
   border-radius: 0.25rem;
 }
 
+.track[data-automin] {
+  /* Make track end in middle of thumbs when auto-scaling min and/or max */
+  margin: 0 calc((var(--thumb-size-auto) - 0.125rem) / 2); /* MAGIC NUMBER */
```
axelboc, line 41: Let's be honest, I'm not 100% sure why this works. So ... better make it clear with a comment. Again, if you find better, please be my guest 😂

src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
```diff
@@ -42,52 +51,52 @@
...
   border-radius: 50%;
 }
 
-.thumbBtnLike > svg {
-  position: absolute;
-  top: 50%;
-  left: 50%;
-  /* Tweak centering of autoscale icons to compensate for inset shadows */
-  transform: translate(-52%, -50%);
-}
```
axelboc, line 67: just moving this down

src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
```diff
@@ -78,8 +78,6 @@ function ScaledSlider(props: Props): ReactElement {
   return (
     <ReactSlider
       className={styles.slider}
-      thumbClassName={styles.thumb}
-      thumbActiveClassName={styles.thumbActive}
```
axelboc, line 82: - The thumb class name is now declared in the `Thumb` component.
- The thumb active class name was no longer used.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
```diff
@@ -22,6 +25,12 @@
   pointer-events: none;
 }
 
+.trackWrapper {
+  flex: 1 1 0%;
+  display: flex;
+  padding: 0 calc(var(--thumb-size) / 2); /* make track end in middle of thumbs */
```
loichuder, line 31: Can't you simply add a `margin: 0 calc(var(--thumb-size) / 2);` to `.track` so that you don't need the wrapper ?

loichuder (approved):
Looks great !

Just a simple dumb question :sweat_smile:

src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
axelboc: Not at all a dumb question. The wrapper was left over from a previous attempt. I thought a `padding` was better at conveying what we were trying to do, but it's not worth the extra wrapper you're right.

PR #534, opened 2021-03-04
 Recompute slider extent when vis domain changes and support all scales

loichuder:


src/h5web/vis-packs/core/utils.test.ts
```diff
@@ -97,9 +97,23 @@ describe('Shared visualization utilities', () => {
...
+    it('should return a non-empty domain when given an empty domain', () => {
+      const extendedDomain = extendDomain([1, 1], 0.5);
+      expect(extendedDomain).toEqual([0, 2]);
+    });
+
+    it('should extend domain for use with log scale', () => {
+      const [min, max] = [10, 100];
+      const [extMin, extMax] = extendDomain([10, 100], 0.25, ScaleType.Log);
+      expect(extMin < min).toBe(true);
+      expect(extMax > max).toBe(true);
```
loichuder, line 109: It is now hard to check for the exact value...

axelboc: Isn't there a combination of domain and extension ratio that could result in a predictable extended domain?

src/h5web/vis-packs/core/utils.ts
```diff
@@ -120,16 +120,21 @@ export function getDomain(
 export function extendDomain(
   bareDomain: Domain,
   extendFactor: number,
-  isLog?: boolean
+  scaleType: ScaleType = ScaleType.Linear
```
axelboc, line 123: I think the scale could be mandatory and come before the `extendFactor`, since we now pass variables as the first two parameters and a constant as the third.

axelboc (approved):
Nicely done! Would love to hear how you got there.

loichuder:
> Nicely done! Would love to hear how you got there.

Back-of-the-envelope calculations :joy: 

Nah to be honest, it is the same principle as the scale that links canvas coordinates to the data coordinates.

src/h5web/vis-packs/core/utils.ts
loichuder: Not sure about this, I think it makes sense to have a "light" version of `extendDomain` that do not need any `scaleType` for simpler uses.

Also, that would imply having a fallback for `abscissaScale` in the `LineVis`.

src/h5web/vis-packs/core/utils.test.ts
loichuder: The extended bound for the log depends on `10**extendFactor` which is almost never an integer for `0 < extendFactor <1`. 

Mind, I could use`toBeCloseTo` to do the test with floating numbers. But to deduce the extended value, I was kinda writing the formula of `scale.invert` which makes the test eats its own tail...

I could perhaps use `extendFactor = 1` for the test :shrug:

src/h5web/vis-packs/core/utils.ts
axelboc: ah right, `abscissaScale` is optional, got you 👍

src/h5web/vis-packs/core/utils.test.ts
axelboc: Yeah, one test with `1`, one test with `0` and one in the middle where you check that the value `isBetween` two values.

axelboc: Maybe there are some edges cases to check too, like when the domain has a bound equal to `0` in linear or symlog.

axelboc: And maybe add checks in the empty domain test with the other scales? This will document the behaviour more precisely, and will avoid any future regressions.

PR #537, opened 2021-03-04
ExtendDomain is clamped to JS supported numbers

loichuder:


src/h5web/vis-packs/core/utils.ts
```diff
@@ -21,6 +21,8 @@ const TICK_FORMAT = format('0');
 
 export const DEFAULT_DOMAIN: Domain = [0.1, 1];
 
+const MIN_POSITIVE_VALUE = 1e-250;
```
loichuder, line 24: I can't use `Number.MIN_VALUE` as it is already too small.

For a log scale and numbers close to `Number.MIN_VALUE`, `scale.invert` indeed gives `0`. The `isSupported` check then passes giving an extended log domain of `[0, x]` and `NaN`s on the next render.

axelboc: Could you use the `scale` the default way to get the min number supported by `scale.invert`: `scale(Number.MIN_VALUE)`?

src/h5web/vis-packs/core/utils.ts
```diff
@@ -228,3 +236,11 @@ export function isScaleType(val: unknown): val is ScaleType {
     typeof val === 'string' && Object.values<string>(ScaleType).includes(val)
   );
 }
+
+function isSupported(val: number) {
+  if (val > 0) {
+    return val > MIN_POSITIVE_VALUE && val < Number.MAX_VALUE;
+  }
+
+  return val > -Number.MAX_VALUE && val < Number.MAX_VALUE;
```
axelboc, line 245: `&& val < -Number.MIN_POSITIVE_VALUE` ? But maybe you can simplify this with `Math.abs`.

src/h5web/vis-packs/core/utils.ts
axelboc: I feel like the logic should be the same for min and max . Could you pass the scale to `isSupported` to make sure that `scale(Number.MIN_VALUE) < Math.abs(val) < scale(Number.MAX_VALUE)`?

src/h5web/vis-packs/core/utils.test.ts
```diff
@@ -128,6 +128,12 @@ describe('Shared visualization utilities', () => {
       expect(extMin).toBe(10);
       expect(extMax).toBe(100);
     });
+
+    it('should not extend domain outside of JS supported values', () => {
+      const domain: Domain = [-Number.MAX_VALUE, Number.MAX_VALUE];
+      const extendedDomain = extendDomain(domain, 0.5);
+      expect(domain).toEqual(extendedDomain);
+    });
```
axelboc, line 136: One more little test with log and `MIN_VALUE` ?

src/h5web/vis-packs/core/utils.ts
```diff
@@ -228,3 +234,9 @@ export function isScaleType(val: unknown): val is ScaleType {
     typeof val === 'string' && Object.values<string>(ScaleType).includes(val)
   );
 }
+
+function isSupported(val: number) {
+  const absVal = Math.abs(val);
+
+  return val === 0 || (absVal > Number.EPSILON && absVal < Number.MAX_VALUE);
```
axelboc, line 241: I had suggested `1 / Number.EPSILON` instead of `Number.MAX_VALUE`, but we could also use `Number.MAX_SAFE_INTEGER`. The problem with `Number.MAX_VALUE` is that you don't really know that you've gone above it, because there's no possible value above it. I suggested `1 / Number.EPSILON` for consistency, and because it gives a number about equal to about half of `MAX_SAFE_INTEGER`, which is still sufficiently large for our use case (I think).

axelboc (approved):

PR #538, opened 2021-03-04
Fall back to data min/max when custom min/max is not supported by scale

axelboc:
Lots to unpack here.

image

- I've put the error messages at the bottom of the tooltip. If the messages are above the auto-scale buttons, if you click on them and the error message disappears, the buttons move up and the tooltip shrinks, which leads to the mouse leaving the tooltip and the tooltip closing. Working around this is not gonna be easy.
- The error icon looked a lot better than a red background on the thumb. It's also more accessible.
- I plan on having a separate stylesheet for the tooltip, and on testing the `getVisDomain` function, but I'll do both after your first review to make sure it's all going in the right direction.

New version with softer icon : 

image


src/h5web/vis-packs/core/heatmap/utils.ts
```diff
@@ -1,11 +1,44 @@
...
+    return [
+      [
+        visDomain[0] <= 0 ? dataDomain[0] : visDomain[0],
+        visDomain[1] <= 0 ? dataDomain[1] : visDomain[1],
+      ],
+      {
+        minError: visDomain[0] <= 0 ? BoundError.InvalidWithLog : undefined,
+        maxError: visDomain[1] <= 0 ? BoundError.InvalidWithLog : undefined,
+      },
+    ];
```
axelboc, line 36: Probably not the most clever code, but it will evolve once we start handling more errors.

loichuder (commented):
The code is very good. Things are taking the right direction in my opinion.

However, looking at your example, wouldn't it be better if the displayed `MIN` in the tooltip would be the invalid value rather than the valid value of `1e0`?

This plus the fact that the thumb shows an error icon while it is at a valid position makes things very misleading.

src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
```diff
@@ -48,11 +55,10 @@ function ScaledSlider(props: Props): ReactElement {
     type: scaleType,
     domain: sliderExtent,
     range: SLIDER_RANGE,
-    round: true,
     clamp: true,
   });
 
-  const scaledValue = value.map(scale) as Domain;
+  const scaledValue = value.map(scale).map(Math.round) as Domain;
```
loichuder, line 61: https://github.com/airbnb/visx/issues/1095

PR #539, opened 2021-03-05
Fix blurry SymLog icon

axelboc:
Fix #539


axelboc (commented):
I've rotated the icon in Inkscape and added it as a React component to the codebase, so we don't have to rotate it with CSS.

src/h5web/toolbar/controls/ScaleSelector/MdGraphicEqRotated.tsx
```diff
@@ -0,0 +1,27 @@
+/**
+ * Apache License 2.0 - https://github.com/google/material-design-icons
+ * Copyright (c) Google
```
axelboc, line 3: Couldn't find a copyright year on the repo.

loichuder (approved):

PR #544, opened 2021-03-08
Fix domain clamping and edge case with [0, 0]

axelboc:


src/h5web/vis-packs/core/utils.test.ts
```diff
@@ -92,58 +92,60 @@ describe('Shared visualization utilities', () => {
...
+      expect(extMax1).toBeGreaterThan(max);
 
-    it('should extend domain by factor 1 with log scale', () => {
       // Extension factor of 1 for log scale means one decade
-      const [extMin, extMax] = extendDomain([10, 100], 1, ScaleType.Log);
-      expect(extMin).toBeCloseTo(1);
-      expect(extMax).toBeCloseTo(1000);
+      const [extMin2, extMax2] = extendDomain([10, 100], 1, ScaleType.Log);
+      expect(extMin2).toBeCloseTo(1);
+      expect(extMax2).toBeCloseTo(1000);
```
axelboc, line 116: Merged the two log scale tests with two different factors.

src/h5web/vis-packs/core/utils.test.ts
```diff
@@ -92,58 +92,60 @@ describe('Shared visualization utilities', () => {
...
-      expect(extMax).toBeGreaterThan(max);
+    it('should extend negative single-value domain', () => {
+      const extendedDomain = extendDomain([-1, -1], 0.5);
+      expect(extendedDomain).toEqual([-2, 0]);
+    });
+
+    it('should return [-1, 1] regardless of factor when trying to extend [0, 0]', () => {
+      expect(extendDomain([0, 0], 0.5)).toEqual([-1, 1]);
+      expect(extendDomain([0, 0], 1)).toEqual([-1, 1]);
     });
```
axelboc, line 132: New test

src/h5web/vis-packs/core/utils.test.ts
```diff
@@ -92,58 +92,60 @@ describe('Shared visualization utilities', () => {
...
-      expect(extMax).toBe(100);
+    it('should not extend domain when factor is 0', () => {
+      const extendedDomain = extendDomain([10, 100], 0, ScaleType.Log);
+      expect(extendedDomain).toEqual([10, 100]);
     });
 
-    it('should not extend domain outside of JS supported values', () => {
-      const domain: Domain = [-1 / Number.EPSILON, 1 / Number.EPSILON];
+    it('should not extend domain outside of supported values', () => {
+      const domain: Domain = [-1 / Number.EPSILON + 1, 1 / Number.EPSILON - 1];
```
axelboc, line 140: I reduce the input domain slightly to check that it is extended and clamped. Previously, the input domain would be returned as is if extending it _would have_ exceeded the supported range.

src/h5web/vis-packs/core/utils.test.ts
```diff
@@ -92,58 +92,60 @@ describe('Shared visualization utilities', () => {
...
     });
 
-    it('should extend domain by factor with log scale', () => {
-      const [min, max] = [1, 10];
-      const [extMin, extMax] = extendDomain([min, max], 0.1, ScaleType.Log);
-      expect(extMin).toBeLessThan(min);
-      expect(extMax).toBeGreaterThan(max);
+    it('should extend positive single-value domain', () => {
+      const extendedDomain = extendDomain([1, 1], 0.5);
+      expect(extendedDomain).toEqual([0, 2]);
```
loichuder, line 121: Could you add a similar test for `ScaleType.Log` ?

loichuder (approved):
Thanks for the clean-up :100:

src/h5web/vis-packs/core/utils.test.ts
axelboc: Hmm I found another edge case... 

```ts
const extendedDomain = extendDomain([1, 1], 1, ScaleType.Log); // => `[0, +Infinity]` => `[0, 1 / Number.EPSILON]`
```

axelboc: Yeah, so we'd return `Number.EPSILON` in this case?

axelboc: Seems weird to return the largest possible log domain when passing `[1, 1]`, though.

loichuder (approved):

PR #545, opened 2021-03-08
Added explanatory abbr on the top of `n` in DimensionMapper

### LINKED ISSUE #462: More info in popup
From user feedback, the popup should include more info, notably the number of elements in each dimension.

loichuder:
Fix #462


axelboc (approved):
Good call

PR #549, opened 2021-03-09
Stop wheel events so page doesn't scroll while zooming

### LINKED ISSUE #520: Zooming on heatmap should not scroll page

axelboc:
Fix #520

### DIFF
--- a/src/h5web/vis-packs/core/hooks.ts
+++ b/src/h5web/vis-packs/core/hooks.ts
@@ -1,5 +1,5 @@
 import ndarray from 'ndarray';
-import { useContext, useMemo, useState } from 'react';
+import { useContext, useEffect, useMemo, useRef, useState } from 'react';
 import { isNumber } from 'lodash-es';
 import { assign } from 'ndarray-ops';
 import { createMemo } from 'react-use';
@@ -123,4 +123,23 @@ export function useCanvasScales(): {
     abscissaScale: getCanvasScale(abscissaConfig, width),
     ordinateScale: getCanvasScale(ordinateConfig, height),
   };
+}
+
+export function useWheelCapture() {
+  const ref = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    const elem = ref.current;
+
+    function onWheel(evt: WheelEvent) {
+      evt.preventDefault();
+    }
+
+    // Handler must be registed as non-passive for `preventDefault` to have an effect
+    // (React's `onWheel` prop registers handlers as passive)
+    elem?.addEventListener('wheel', onWheel, { passive: false });
+    return () => elem?.removeEventListener('wheel', onWheel);
+  });
+
+  return ref;
 }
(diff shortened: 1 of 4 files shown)


src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -1,56 +1,59 @@
...
+import { useState, useEffect, forwardRef } from 'react';
 import ReactDOM from 'react-dom';
 import { useThree } from 'react-three-fiber';
 
 // Simplified version of `drei`'s `<Html>` component
 // https://github.com/pmndrs/drei/blob/v2.2.3/src/Html.tsx
-function Html(props: React.HTMLAttributes<HTMLDivElement>): ReactElement {
-  const { className, style, children, ...divProps } = props;
+const Html = forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDivElement>>(
+  (props, ref) => {
```
axelboc, line 8: Just adding the possibility to do `<Html ref={ref} />` (by wrapping the component in `forwardRef`).

src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
```diff
@@ -91,11 +93,16 @@ function PanZoomMesh(): ReactElement {
...
-    <mesh {...{ onPointerMove, onPointerUp, onPointerDown, onWheel }}>
-      <meshBasicMaterial attach="material" opacity={0} transparent />
-      <planeGeometry attach="geometry" args={[width, height]} />
-    </mesh>
+    <>
+      <mesh {...{ onPointerMove, onPointerUp, onPointerDown, onWheel }}>
+        <meshBasicMaterial attach="material" opacity={0} transparent />
+        <planeGeometry attach="geometry" args={[width, height]} />
+      </mesh>
+      <Html ref={wheelCaptureRef} />
```
axelboc, line 104: The other option would be to have the `ref` on the `div` that R3F creates and wraps the canvas with, but it doesn't seem to be possible. We could wrap the canvas in an extra div (the current wrapper, `vis`, has padding for the axes...), but it's a bit hacky with regard to flexbox...

This option is simpler, and it makes more sense, since it's here that the zoom is implemented in the first place.

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -124,3 +124,22 @@ export function useCanvasScales(): {
...
+
+  useEffect(() => {
+    const elem = ref.current;
+
+    function onWheel(evt: WheelEvent) {
+      evt.preventDefault();
+    }
+
+    // Handler must be registed as non-passive for `preventDefault` to have an effect
+    // (React's `onWheel` prop registers handlers as passive)
```
loichuder, line 139: Good thing you added a comment there, that's far from obvious :roll_eyes:

loichuder (approved):

PR #553, opened 2021-03-10
Split domain slider styles into multiple modules

axelboc:
Also:

- Remove the line vis with a negative domain and log scale, as this no longer works with the throw I introduced.
- Fix a couple of problems in the line vis scales stories that led to the wrong display names being shown.


src/stories/LineVisScales.stories.tsx
```diff
@@ -45,15 +37,15 @@ SymLogForY.args = {
...
+LogForX.storyName = 'Log for X';
 LogForX.args = {
   dataArray: dataArrayForXLog,
   domain: domainForXLog,
   abscissaParams: { scaleType: ScaleType.Log },
 };
 
 export const SymLogForX = Template.bind({});
-LogForY.storyName = 'Sym Log for X';
+SymLogForX.storyName = 'Sym Log for X';
```
loichuder, line 48: :fishing_pole_and_fish:

loichuder (approved):

PR #556, opened 2021-03-11
Added support of auxiliary_signals in NxSpectrum

loichuder:


src/h5web/vis-packs/core/line/DataCurve.tsx
```diff
@@ -22,58 +22,16 @@ function DataCurve(props: Props): ReactElement {
...
-      if (Number.isFinite(yErrTop)) {
-        const errTopVector = new Vector3(x, yErrTop, 0);
-        errorBarSegments.push(dataVector, errTopVector);
-        errorCapPoints.push(errTopVector);
-      }
-    });
-
-    return { data: dataPoints, bars: errorBarSegments, caps: errorCapPoints };
-  }, [abscissaScale, abscissas, camera.far, errors, ordinateScale, ordinates]);
+  const points = useCanvasPoints(abscissas, ordinates, errors);
```
loichuder, line 34: `eslint` was bugging me about the complexity of this component due to the new ternary above. I resolved it by putting all the points computation in a new hook.

src/h5web/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -53,27 +47,25 @@ function MappedLineVis(props: Props): ReactElement {
...
-  const dataDomain = useDomain(
-    dataValues,
-    yScaleType,
-    showErrors || !autoScale ? errorValues : undefined
+  const dataDomain = useAutoScaledDomain(
+    dataArray,
+    showErrors ? errorArray : undefined,
+    auxArrays,
+    autoScale,
+    yScaleType
```
loichuder, line 68: Rather than making a new hook only for the `auxArrays`, I crossed the bridge and put the `autoScale` code in it as well.

Again, I welcome suggestions for the name.

src/h5web/vis-packs/core/line/models.ts
```diff
@@ -15,3 +16,8 @@ export const GLYPH_URLS = {
   Square: squareURL,
   Cap: capURL,
 };
+
+export interface MappedArray {
+  base: ndarray | undefined;
+  mapped: ndarray | undefined;
+}
```
loichuder, line 23: The new interface returned by `useArrays`.

`MappedArray` is **again** not an ideal name

src/h5web/providers/mock/metadata-utils.ts
```diff
@@ -254,6 +265,10 @@ export function makeNxDataGroup<
     | { axes: T; axesAttr: (Extract<keyof T, string> | '.')[] }
     | { axes?: never; axesAttr?: never }
   ) &
+    (
+      | { auxiliary: T; auxAttr: (Extract<keyof T, string> | '.')[] }
```
axelboc, line 269: I don't think `'.'` should be allowed here.

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -143,3 +163,17 @@ export function useWheelCapture() {
...
+  dimMapping: DimensionMapping
+) {
+  const values = useDatasetValues(auxDatasets);
+  return useMemo(() => {
+    return Object.entries(values).map(([name, value]) => {
+      const base = getBaseArray(value, dims);
+      return { name, base, mapped: getMappedArray(base, dimMapping) };
+    });
+  }, [dimMapping, dims, values]);
+}
```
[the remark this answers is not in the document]
axelboc: Yep, I get you, had the same problem with `useDatasetValues` and in other places.

src/h5web/vis-packs/core/line/DataCurve.tsx
```diff
@@ -22,58 +22,16 @@ function DataCurve(props: Props): ReactElement {
...
+    color = '--secondary-dark',
     curveType = CurveType.LineOnly,
   } = props;
 
-  const { camera, gl } = useThree();
-  const { abscissaScale, ordinateScale } = useCanvasScales();
+  const { gl } = useThree();
+  const curveColor = color.startsWith('--')
+    ? window.getComputedStyle(gl.domElement).getPropertyValue(color).trim()
+    : color;
```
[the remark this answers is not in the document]
axelboc: Clever 👍

src/h5web/vis-packs/core/line/LineVis.tsx
```diff
@@ -122,6 +124,17 @@ function LineVis(props: Props): ReactElement {
...
+          return (
+            <DataCurve
+              key={name}
+              color="--secondary"
+              curveType={curveType}
+              abscissas={abscissas}
+              ordinates={array.data as number[]}
+            />
+          );
+        })}
```
axelboc, line 137: you can condense this by removing the `return`

src/h5web/vis-packs/nexus/containers/NxSpectrumContainer.tsx
```diff
@@ -16,6 +16,7 @@ function NxSpectrumContainer(props: VisContainerProps): ReactElement {
 
   const nxData = useNxData(entity);
   const {
+    auxiliaryDatasets,
```
axelboc, line 19: Might be simpler to just name this property `auxDatasets` in the model.

axelboc (changes requested):
I get how it works and all, and kudos to you, but I'm sorry to say that this feels very convoluted. I don't think you've made the right abstractions in `MappedLineVis`. Passing those `{ base, mapped }` objects makes things particularly hard to follow.

**I would remove the `useArrays` hook, inline both the `useAuxDatasetArrays` and `useAutoScaledDomain` in `MappedLineVis` for now, and reassess from there.**

I know it's going to make `MappedLineVis` unreadable, with lots of new difficult-to-name constants, but I'm sure it will help us uncover a better way of abstracting the complexity. If you agree with this approach, we can LiveShare when you're done to discuss the best way to abstract the complexity. (If you get a complexity warning, just disable it at the top of the file for now.)

src/h5web/vis-packs/core/utils.ts
```diff
@@ -262,3 +271,65 @@ export function isScaleType(val: unknown): val is ScaleType {
...
+      Math.max(domain[1], domainToCombine[1]),
+    ],
+    otherDomains
+  );
+}
+
+export function getBaseArray<T extends unknown[] | undefined>(
+  value: T,
+  rawDims: number[]
+): T extends (infer U)[] ? ndarray<U> : undefined;
```
axelboc, line 301: I wonder whether we still need those override declarations, since `useMappedArray` has one already. Yeah... probably for `useMappedArrays`... Might be worth trying to remove the override declaration on `useMappedArray, actually. Or maybe you've already tried and didn't get anywhere. Not a merge blocker either way.

src/stories/Home.stories.mdx
```diff
@@ -99,7 +99,7 @@ const dataArray = getMockDataArray('/nD_datasets/twoD');
 The library exposes a number of utility functions, which are documented below, as well as all the enums and types used by these utility functions.
 If you use only the top-level visualization components, the only utilities you need are `getDomain` and/or `useDomain`.
 
-- **`getDomain(values: number[], scaleType: ScaleType = ScaleType.Linear): Domain | undefined`** - Find the min and max values contained in a flat array of numbers. If `scaleType` is `ScaleType.Log` and the domain crosses zero, clamp the min to the first positive value or return `undefined` if the domain is not supported (i.e. `[-x, 0]`).
+- **`getDomain(values: ndarray | number[], scaleType: ScaleType = ScaleType.Linear): Domain | undefined`** - Find the min and max values contained in a array (or ndarray) of numbers. If `scaleType` is `ScaleType.Log` and the domain crosses zero, clamp the min to the first positive value or return `undefined` if the domain is not supported (i.e. `[-x, 0]`).
```
axelboc, line 102: Maybe also change the example in the previous section to pass the `dataArray` rather than the flat values?

axelboc (changes requested):
Nicely done, I think it looks great! 👏 Only small tweaks, and one question about the override typing declarations of the hooks.

src/h5web/vis-packs/core/line/LineVis.tsx
```diff
@@ -62,6 +64,14 @@ function LineVis(props: Props): ReactElement {
     );
   }
 
+  auxArrays.forEach((auxArray) => {
+    if (auxArray.size !== dataArray.size) {
+      throw new Error(
+        `Auxiliary size (${auxArray.size}) does not match data length (${dataArray.size})`
+      );
+    }
+  });
```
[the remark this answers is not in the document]
loichuder: But then I cannot display `auxArray.size` right ?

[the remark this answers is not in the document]
axelboc: Ah right, hmm -- I don't think it's worth displaying the size given how unlikely the error is to occur -- or if it occurs how easy it is to check the dimensions of the auxiliary arrays....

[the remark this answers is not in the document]
axelboc: But I don't mind either way.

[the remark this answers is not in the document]
axelboc: By the way, is it really a big deal if some aux arrays are shorter than the data array?

src/h5web/vis-packs/core/utils.ts
```diff
@@ -262,3 +271,65 @@ export function isScaleType(val: unknown): val is ScaleType {
...
+
+export function getCombinedDomain(
+  domain: Domain | undefined,
+  domainsToCombine: (Domain | undefined)[]
+): Domain | undefined {
+  if (domainsToCombine.length === 0) {
+    return domain;
+  }
+
+  const [domainToCombine, ...otherDomains] = domainsToCombine;
```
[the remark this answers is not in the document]
loichuder: [domainToCombine, ...remainingDomains] ? :smile:

src/h5web/vis-packs/core/utils.ts
loichuder: Removing the override of `useMappedArray` works :fishing_pole_and_fish:

src/stories/Home.stories.mdx
loichuder: Good call as I saw that it was still named `findDomain`...

src/h5web/vis-packs/core/line/LineVis.tsx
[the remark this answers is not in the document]
loichuder: I don't really know.

Let's keep it this way for now to have a well-defined behaviour and let's see if we need to change this once we encounter real use cases

src/h5web/vis-packs/core/utils.ts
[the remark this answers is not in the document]
axelboc: whatev's

src/h5web/vis-packs/core/utils.ts
axelboc: Hmm still seems to be there, maybe you forgot to push?

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -143,3 +105,42 @@ export function useWheelCapture() {
...
+): T extends (infer U)[] ? [ndarray<U>, ndarray<U>] : [undefined, undefined];
+
+export function useMappedArray<T>(
+  value: T[] | undefined,
+  dims: number[],
+  mapping: DimensionMapping,
+  autoScale?: boolean
+) {
+  const baseArray = createMemo(getBaseArray)(value, dims);
+  const mappedArray = createMemo(applyMapping)(baseArray, mapping);
```
axelboc, line 125: Ooh you're recreating a hook on every call here 😅 -- just move them out above `useCombinedDomain`

axelboc (approved):

src/h5web/vis-packs/core/utils.ts
loichuder: Hm I don't know what I imagined but removing it doesn't work in fact.

It might be because we return either `[ndarray, ndarray]` or `[undefined, undefined]` which is interpreted as `(ndarray | undefined)[]` without the override.

PR #557, opened 2021-03-11
Implement editing of domain bounds

axelboc:
Some screenshots for 😮, 😋 and 😍 ... (at least! 😏)

image

image

image

image

image


src/h5web/toolbar/Toolbar.module.css
```diff
@@ -29,12 +29,6 @@
   padding: 0 0.25rem;
 }
 
-.btn:disabled,
-.btn[aria-disabled='true'] {
-  opacity: 0.5;
-  pointer-events: none;
-}
-
```
axelboc, line 37: Moved to global `btn-clean`.

src/h5web/toolbar/Toolbar.module.css
```diff
@@ -88,6 +82,10 @@
   box-shadow: var(--btn-shadow-idle) var(--btn-shadow-color);
 }
 
+.btn:active > .btnLike {
+  box-shadow: var(--btn-shadow-pressed) var(--btn-shadow-color);
+}
+
```
axelboc, line 88: Buttons now look pressed on `mousedown`.

src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+        className={styles.value}
+        type="text"
+        name="bound"
+        value={inputValue}
+        aria-labelledby={`${id}-label`} // https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/718
+        onChange={(evt) => setInputValue(evt.target.value)}
+        onFocus={() => {
+          if (!isEditing) {
+            onEditToggle(true);
+          }
```
axelboc, line 66: This "pins" the popup open when a field receives focus.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -38,38 +29,55 @@ function DomainSlider(props: Props): ReactElement {
   const { dataDomain, customDomain, scaleType, disabled } = props;
   const { onCustomDomainChange } = props;
 
-  const [visDomain, errors] = useVisDomain(dataDomain, customDomain, scaleType);
-  const [sliderDomain, setSliderDomain] = useState(visDomain);
+  const visDomain = useVisDomain(customDomain, dataDomain);
+  const [safeDomain, errors] = useSafeDomain(visDomain, dataDomain, scaleType);
 
+  const [sliderDomain, setSliderDomain] = useState(visDomain);
```
axelboc, line 35: The slider domain state now holds the "unsafe" vis domain. This allows the tooltip to show the custom domain even when it's invalid, while also showing the slider's domain while a thumb is being moved.

The `safeDomain` is passed to the `ScaledSlider` for computing the extended domain. `ScaledSlider` then recomputes a safe domain from its `value` on every render so the thumbs don't end up in weird places.

src/h5web/vis-packs/core/heatmap/utils.ts
```diff
@@ -14,15 +14,17 @@ const GRADIENT_PRECISION = 1 / 20;
...
-  dataDomain: Domain,
   customDomain: CustomDomain,
+  dataDomain: Domain
+): Domain {
+  return [customDomain[0] ?? dataDomain[0], customDomain[1] ?? dataDomain[1]];
+}
+
+export function getSafeDomain(
+  visDomain: Domain,
+  dataDomain: Domain,
```
loichuder, line 25: `dataDomain` could be renamed to `fallbackDomain`

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -79,34 +87,38 @@ function DomainSlider(props: Props): ReactElement {
...
+        open={hovered || isEditing}
+        sliderDomain={sliderDomain}
         dataDomain={dataDomain}
         errors={errors}
         isAutoMin={isAutoMin}
         isAutoMax={isAutoMax}
         onAutoMinToggle={() => {
           const newMin = isAutoMin ? dataDomain[0] : undefined;
           onCustomDomainChange([newMin, customDomain[1]]);
+          toggleEditingMin(isAutoMin);
```
loichuder, line 109: Not sure about this. It feels odd to toggle editing when deactivating the autoscale.

My expectation would be that the focus on the edition is lost when toggling a button (i.e. `toggleEditingMin(false)`).

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -79,34 +87,38 @@ function DomainSlider(props: Props): ReactElement {
       />
 
       <ToggleBtn
-        small
-        label={getAutoscaleLabel(isAutoMin, isAutoMax)}
-        icon={FiZap}
-        value={isAutoscaling}
+        iconOnly
+        label="Edit domain"
+        icon={MdTune}
```
loichuder, line 92: I feel the icon is not really appropriate here as it shows sliders.

Perhaps `MdEdit` or `FiEdit` would be more manifest ?

src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+      </label>
+
+      <input
+        id={id}
+        ref={inputRef}
+        className={styles.value}
+        type="text"
+        name="bound"
+        value={inputValue}
+        aria-labelledby={`${id}-label`} // https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/718
```
loichuder, line 61: Isn't it rather https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/566 ?

src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+      data-error={hasError || undefined}
+      data-editing={isEditing}
+      onSubmit={(evt) => {
+        evt.preventDefault();
+        onChange(Number.parseFloat(inputValue));
+        onEditToggle(false);
+      }}
+    >
+      <label id={`${id}-label`} className={styles.label} htmlFor={id}>
+        <abbr title={value.toString()}>{label}</abbr>
```
loichuder, line 51: I find counter-intuitive to have the `value` on the label.

Isn't there a way to have it on hover on the editing box when not editing ?

src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+function BoundEditor(props: Props): ReactElement {
+  const { label, value, isEditing, hasError, onEditToggle, onChange } = props;
+
+  const id = `${label}-bound`;
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const [inputValue, setInputValue] = useState(formatValue(value));
+
+  useEffect(() => {
+    setInputValue(isEditing ? value.toString() : formatValue(value));
```
loichuder, line 24: I would find more intuitive if the value was staying under `.3e` format when editing.

But that's my personal opinion so let's wait for user feedback before changing.

loichuder (changes requested):
I like the design very much ! Looking clean !

A few nitpicks here and there

src/h5web/vis-packs/core/heatmap/utils.ts
axelboc: and `visDomain` to just `domain`

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
axelboc: Oh yeah? I thought maybe users who disable the autoscale would want to edit the value right away in some cases. That being said, it does "pin" the popup, which means an extra click is required to close it if they're not interested in editing the value... Alright, I'll change this.

loichuder: > That being said, it does "pin" the popup, which means an extra click is required to close it if they're not interested in editing the value... Alright, I'll change this.

Yup that was my point

axelboc: However, I think if a field is in edit mode and the user disables the autoscale, I think the field should remain in edit mode.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
axelboc: Yeah, but those are really ugly on their own... Do you feel strongly about it? The icon is not meant to describe the UI but the fact that the domain can be "tuned". As long as users understand that... 🤷

src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
axelboc: haha yes, thanks, forgot to update the link 😄

src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
axelboc: I'll give it a try.

src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
axelboc: Ah yeah? It would allow me to shrink back the size of the input fields and to not have to reduce the font-size in editing mode. It wouldn't prevent people from entering values with as many decimals as they want, so... why not!

axelboc: > so let's wait for user feedback before changing.

yeah agreed

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
loichuder: > Yeah, but those are really ugly on their own...

Really ? I like `FiEdit`
image



> The icon is not meant to describe the UI but the fact that the domain can be "tuned".

I get you but the icon aspect is really at odds with the fact that clicking on it toggles editing. Well in any case:

> Do you feel strongly about it? 

No :smile:

src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
axelboc: Alright, as I thought, I can just use the `title` attribute on the input field. The downside is that there no longer is an indication that supplementary information is available like there was with `abbr` (the dotted underline). Not a huge problem.

image

loichuder (approved):
Great !

loichuder:
Before you merge, I noticed a small issue:

- Click on the **Tune** button to toggle edit
- Click on it again to toggle edit off

The tooltip disappears and do not appear again, even when moving thumbs. One must move the pointer away from the slider to make it reappear.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
axelboc: `FiEdit` isn't too bad, but it looks wrong visually because the square outline is not centred in the button (and I think tweaking the alignment would require reducing the icon's size too much).

But I'm not giving up yet! Here are a couple of alternatives (note that I can adjust the color/stroke to make some of these less imposing):

image

image

image

image

image

image

... and my new favourite: 

image

axelboc:
> The tooltip disappears and do not appear again, even when moving thumbs. One must move the pointer away from the slider to make it reappear.

Yeah, it's on purpose. You would only click this button to explicitly close the popup and cancel your edits. If you click and the popup stays, it just feels confusing... doesn't it?

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
axelboc: With the `small` variant applied to the toggle:

image

axelboc:
Neh you're right, now with the input fields, the behaviour of keeping the tooltip open makes more sense.

PR #559, opened 2021-03-11
Keep exponent notation when editing bounds (but increase precision)

axelboc:
Also, deal with minus sign to avoid `NaN`.


loichuder (approved):
Excellent !

PR #560, opened 2021-03-12
Make errors sticky in domain tooltip

axelboc:
The error "unsticks" when the popup is closed because the `BoundErrorMessage` components are rendered conditionally based on  the `open` state.

image

image


src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
```diff
@@ -30,11 +31,11 @@ function BoundEditor(props: Props): ReactElement {
       inputRef.current?.blur();
     }
 
-    if (isEditing && label === 'Min') {
+    if (isEditing && bound === 'min') {
```
axelboc, line 34: A bit stricter typing here.

loichuder (approved):
:+1:

PR #565, opened 2021-03-17
Fall back to data domain when min is greater than max

axelboc:
image

I've identified two remaining edge cases:

### `min > max` (fallback to `dataDomain`) + `custom max << data min` (or `custom min >> data max`)

1. Go to http://localhost:3000/mock 
2. Change max bound to `-400` 🡪 `visDomain=[-95, -400]`, so `min > max` and `safeVisDomain=[-95, 400]`. Note that custom max (-400) is much lower than data min (-95), so custom max is still lower than the slider's extent min.
3. Click to the left of the min bound 🡪 the custom min changes to a value still greater than custom max, so `safeVisDomain` and the slider's extent don't change.
4. Click to the left of the min bound again 🡪 the custom min doesn't change or very little, because the slider's extent stays the same after every click.

I don't believe there's a way around this. To restore normal behaviour, the user has to move the max thumb or edit the values manually.

### log + `custom min <= 0` (fallback to data min) + `custom max < data min`

1. Go to http://localhost:3000/mock 
2. Switch to log.
3. Enter a custom min value lower than or equal to 0.
4. Enter a custom max between 0 and data min (1), for instance 0.5 🡪 custom min is lower than max but not in `safeVisDomain`, so the slider is reversed.

I will fix this in my next PR by making custom min to fall back to custom max is this specific case.

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
@@ -1,7 +1,6 @@
-import { Fragment, ReactElement, Ref } from 'react';
+import { Fragment, ReactElement, Ref, useState } from 'react';
 import { FiSkipBack, FiSkipForward } from 'react-icons/fi';
 import ReactSlider from 'react-slider';
-import { useBoolean } from 'react-use';
 import { createAxisScale, extendDomain } from '../../../vis-packs/core/utils';
 import type {
   Domain,
@@ -48,10 +47,6 @@ function ScaledSlider(props: Props): ReactElement {
   const { minError, maxError } = errors;
 
   const sliderExtent = extendDomain(safeVisDomain, EXTEND_FACTOR, scaleType);
-
-  const [hasMinChanged, setMinChanged] = useBoolean(false);
-  const [hasMaxChanged, setMaxChanged] = useBoolean(false);
-
   const scale = createAxisScale({
     type: scaleType,
     domain: sliderExtent,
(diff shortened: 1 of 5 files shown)


src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
```diff
@@ -91,13 +80,14 @@ function ScaledSlider(props: Props): ReactElement {
       max={SLIDER_RANGE[1]}
       value={scaledValue}
       disabled={disabled}
+      onBeforeChange={(bounds) => setBeforeChangeValue(bounds as Domain)}
       onChange={(bounds) => handleChange(bounds as Domain)}
-      onAfterChange={handleAfterChange}
+      onAfterChange={(bounds) => handleChange(bounds as Domain, true)}
```
axelboc, line 85: I can now reuse the same function as `onChange` to avoid storing `hasMinChange` and `hasMaxChanged` in local states.

loichuder (commented):
I don't know if this proposal hampers your future fixes but what about adding `minGreaterThanMax` as a new type of `BoundError` ?

The error would be applied to both `minError` and `maxError` but you won't need any`minGreater ||` and to pass `minGreater` each time with `minError`/`maxError`.

axelboc:
The problem is that I don't want this error to appear below each bound.

src/h5web/vis-packs/core/heatmap/utils.ts
```diff
@@ -25,19 +25,21 @@ export function getSafeDomain(
   fallbackDomain: Domain,
   scaleType: ScaleType
 ): [Domain, DomainErrors] {
-  const [min, max] = domain;
+  const minGreater = domain[0] > domain[1];
+  const [min, max] = minGreater ? fallbackDomain : domain;
```
loichuder, line 29: Why not returning early here ?

loichuder (approved):
You could name `minMaxSwapped` all the way I think.

axelboc:
When I see `minMaxSwapped`, I think that the values have been swapped in the returned safe domain, which is not the case.

src/h5web/vis-packs/core/heatmap/utils.ts
axelboc: Sorry, missed your comment. Very true! Will update my next PR.

axelboc: Much better, I've made a separate commit as the diff is now a bit confusing.

PR #567, opened 2021-03-17
Let custom min fall back to custom max in one specific edge case

axelboc:
This resolves scenario 5 described in https://github.com/silx-kit/h5web/issues/510#issuecomment-797402408

To reproduce:

1. Go to http://localhost:3000/mock
2. Switch to log.
3. Enter a custom min value lower than or equal to 0.
4. Enter a custom max between 0 and data min (1), for instance 0.5

image

This introduces a noteworthy, but still logical, behaviour: in the same state as above, move the max thumb to the right 🡪 the two thumbs don't separate because the new custom max is still lower than data min. You have to move the max thumb again until you go past the data min:

image

image

... or instead, push the min thumb to the left with the max thumb so custom min becomes positive.

### DIFF
--- a/src/h5web/vis-packs/core/heatmap/utils.ts
+++ b/src/h5web/vis-packs/core/heatmap/utils.ts
@@ -29,11 +29,19 @@ export function getSafeDomain(
   const [min, max] = minGreater ? fallbackDomain : domain;
 
   if (scaleType === ScaleType.Log) {
+    const logSafeMin = min <= 0 ? fallbackDomain[0] : min;
+    const logSafeMax = max <= 0 ? fallbackDomain[1] : max;
+    const logSafeMinGreater = logSafeMin > logSafeMax;
+
     return [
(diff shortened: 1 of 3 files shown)


src/h5web/vis-packs/core/heatmap/utils.ts
```diff
@@ -25,21 +25,30 @@ export function getSafeDomain(
...
+    [logSafeMinGreater ? logSafeMax : logSafeMin, logSafeMax],
+    {
+      minError: logSafeMinGreater
+        ? BoundError.CustomMaxFallback
+        : min <= 0
+        ? BoundError.InvalidWithLog
+        : undefined,
+      maxError: max <= 0 ? BoundError.InvalidWithLog : undefined,
+    },
+  ];
```
loichuder, line 51: I preferred the reversed early return but whatever

loichuder (approved):
Clever :100: ! Well done :slightly_smiling_face:

PR #568, opened 2021-03-17
Add button to swap custom min and custom max

axelboc:
The button only appears if both min an max have custom values (i.e. only if the user purposefully entered two custom values in the wrong order):

image

Swapping resolves the situation:

image

If any bound is set to auto-scale, the button is not visible:

image


src/h5web/toolbar/controls/DomainSlider/MinGreaterErrorMessage.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <span>
+        Min greater than max
+        <br />
+        <FiCornerDownRight /> falling back to <strong>data range</strong>
+      </span>
+      {!isAuto && (
+        <button
+          className={styles.actionBtn}
+          type="button"
+          onClick={() => onSwap()}
```
loichuder, line 24: Missing some `aria` attributes ?

loichuder (changes requested):

src/h5web/toolbar/controls/DomainSlider/MinGreaterErrorMessage.tsx
axelboc: The icon inside the button has a label, so it's all good 👍

src/h5web/toolbar/controls/DomainSlider/MinGreaterErrorMessage.tsx
```diff
@@ -0,0 +1,33 @@
...
+interface Props {
+  isAuto: boolean;
+  onSwap: () => void;
+}
+
+function MinGreaterErrorMessage(props: Props) {
+  const { isAuto, onSwap } = props;
+
+  return (
+    <p className={styles.minMaxError}>
```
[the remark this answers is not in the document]
axelboc: Completely forgot about it, thanks! I'm abstracting the behaviour in a `StickyError` component. No need for error-specific components any more.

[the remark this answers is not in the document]
axelboc: Scratch that, as discussed, I've replaced the sticky behaviour with invisible padding around the tooltip. I still abstracted the error messages with a new component called `ErrorMessage`.

src/h5web/toolbar/controls/DomainSlider/ErrorMessage.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface Props {
+  message: string;
+  fallback: string;
+  showSwapBtn?: boolean;
+  onSwap?: () => void;
+}
+
+function ErrorMessage(props: Props) {
+  const { message, fallback, showSwapBtn = false, onSwap } = props;
```
loichuder, line 13: I would send the `error` as props and put the call of `getBoundErrorProps` (that would be renamed) in this component.

For `minGreater`, you could add a new type of `BoundError` created at https://github.com/silx-kit/h5web/pull/568/files#diff-0bff5f3cf4d4a01448d5c83c2021ad2d214a71511c758e866be6bad0698205aeR49. You would then need to add the appropriate associated message in `getBoundErrorProps`.

But I won't push strongly for this, feel free to keep as it is.

loichuder (approved):

src/h5web/toolbar/controls/DomainSlider/ErrorMessage.tsx
axelboc: No, you're right, great idea!

PR #569, opened 2021-03-18
Remove top-level describe blocks

axelboc:
Came across this: https://kentcdodds.com/blog/avoid-nesting-when-youre-testing ... and I whole-heartily agree: top-level describe blocks don't add much value when you consider that the file path/name already appears in Jest's output. They increase the nesting level for no good reason.

### DIFF
--- a/src/h5web/providers/hsds/utils.test.ts
+++ b/src/h5web/providers/hsds/utils.test.ts
@@ -1,6 +1,7 @@
 import Complex from 'complex.js';
 import { HDF5Type, HDF5Id, HDF5TypeClass } from '../hdf5-models';
 import type {
+  HsdsStringType,
   HsdsArrayType,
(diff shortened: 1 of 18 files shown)


src/h5web/providers/hsds/utils.test.ts
```diff
@@ -10,220 +10,222 @@ import type {
...
+    charSet: 'H5T_CSET_UTF8',
+    strPad: 'H5T_STR_NULLTERM',
+    length: 49,
+  },
+  hdf5: {
+    class: HDF5TypeClass.String,
+    charSet: 'UTF8',
+    length: 49,
+  },
+};
```
loichuder, line 44: Following the same principles, `asciiStrType` and `unicodeStrType` could be inlined in the according tests.

loichuder (approved):

PR #570, opened 2021-03-18
Make responsive toolbar more efficient and flexible

axelboc:
I've created separate commits, which will probably be easier to review.

The gist of it is that the previous implementation did not allow the toolbar controls to be feature tested. We used to render every control twice:

- Once visibly (or in the overflow menu) but only when measured.
- A second time (with `React.clone`) invisibly and disabled in order to measure them (and keep the measurements up to date).

Since JSDom does not support measuring elements, the controls were never "measured" and would remain hidden and disabled. React Testing Library excludes disabled elements when querying, so it would never find any controls in the toolbar.

The new implementation renders every control only once and never disabled. It uses CSS to hide the controls while they are being measured, and once all of them have been measured, it shows them (and moves those that don't fit in the overflow menu).

The behaviour is the same as before with one exception: the measurements of the controls in the overflow menu are not kept up to date. In other words, even if one control shrinks enough to fit on the toolbar, it will stay in the overflow menu. The reasoning is that it's best to avoid moving things around if we don't have to. This is very specific and affects neither the opposite behaviour (i.e. when a control visible on the toolbar shrinks or grows) nor the resize behaviour (controls from the overflow menu will still move in/out of the overflow menu on resize).

### DIFF
--- a/src/h5web/toolbar/MeasuredControl.tsx
+++ b/src/h5web/toolbar/MeasuredControl.tsx
@@ -1,0 +1,30 @@
+import type { ReactElement } from 'react';
+import Measure from 'react-measure';
+import styles from './Toolbar.module.css';
+
+interface Props {
+  children: ReactElement;
+  onMeasure: (width: number) => void;
+}
+
+function MeasuredControl(props: Props) {
+  const { children: child, onMeasure } = props;
+
+  return (
+    <Measure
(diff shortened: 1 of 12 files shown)


loichuder (approved):
I had some trouble to process the last commit but it works ! :smile:

PR #573, opened 2021-03-22
Upgrade ESLint config and remove `ReactElement` return types

axelboc:
- Upgrade `eslint-config-galex` and remove the now-obsolete Storybook overrides: https://github.com/ljosberinn/eslint-config-galex/issues/243
- As per https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript/, remove all `ReactElement` return types from React components. React warns us in other ways if we don't return what it expects.
  - This allows us to re-enable `react/jsx-no-useless-fragment` so that we can now return `null` instead of empty fragments.


src/h5web/toolbar/controls/ToggleGroup.tsx
```diff
@@ -63,7 +63,7 @@ type Props = ToggleGroupProps & {
   children: ReactNode;
```
loichuder, line 63: Should this be changed ?

src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
```diff
@@ -72,7 +72,6 @@ function PanZoomMesh(): ReactElement {
     (evt: WheelEvent) => {
       const factor = evt.deltaY > 0 ? ZOOM_FACTOR : 1 / ZOOM_FACTOR;
 
-      // eslint-disable-next-line no-param-reassign
       camera.zoom = Math.max(1, camera.zoom * factor);
```
loichuder, line 75: Why can you disable this now ?

loichuder (approved):
That's a lot of components :smile:

src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
axelboc: My guess is, the rule is smarter. It's meant to forbid `camera = ...`, not mutating the parameter like we are doing here.

src/h5web/toolbar/controls/ToggleGroup.tsx
axelboc: I guess so! `ReactElement<BtnProps>[]` does the trick.

PR #575, opened 2021-03-23
Inject zustand store hooks via context

### LINKED ISSUE #518: Config states are shared between tabs in JupyterLab
We need the config states to be injected with a provider à la Redux.

axelboc:
Fix #518 

I follow the advice I found on a Zustand issue (https://github.com/pmndrs/zustand/issues/128#issuecomment-673398578) to implement two new config providers: `HeatmapConfigProvider` and `LineConfigProvider`. The `Visualizer` retrieves the correct provider for the active visualization and renders it. The providers initialize their zustand stores on first render, and inject the `useStore` hooks that zustand creates though their contexts. `useHeatmapConfig` and `useLineConfig` then retrieve their `useStore` hooks with `useContext` and call them.

A few things to note:

- I chose to provide the configs on demand, that is, when a visualization becomes active. The downside is that the stores get recreated every time we switch visualization, but because the configs are persisted, this is not a problem.
- ... except for `customDomain` which now has to be persisted as well (otherwise it gets reset every time we switch visualization).

So the only difference with the current behaviour is that `customDomain` is now restored from `localStorage` on page load. I don't have a problem with that.

I could have chosen to inject both configs from the top-level of the app (or from `CorePack` and `NexusPack`), but it didn't seem very scalable:

```tsx
function CorePack(props: Props) {
  ...

  return (
    <LineConfigProvider>
      <HeatmapConfigProvider>
        <Visualizer
          entity={entity}
          activeVis={activeVis}
          supportedVis={supportedVis}
          onActiveVisChange={setActiveVis}
        />
      </HeatmapConfigProvider>
    </LineConfigProvider>
  );
}
```

Happy to reconsider, though, as both solutions work fine.


src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -36,8 +36,8 @@ function DomainSlider(props: Props) {
     setSliderDomain(visDomain);
   }, [visDomain, setSliderDomain]);
 
-  const isAutoMin = customDomain[0] === undefined;
-  const isAutoMax = customDomain[1] === undefined;
+  const isAutoMin = customDomain[0] === null;
+  const isAutoMax = customDomain[1] === null;
```
axelboc, line 40: I've added a comment to the model, but FYI, `undefined` gets persisted as `null` to `localStorage` (because of `JSON.stringify`), so changing from `undefined to `null` everywhere, means we don't have to change it dynamically in zustand's `persist` middleware.

src/h5web/vis-packs/core/heatmap/config.tsx
```diff
@@ -0,0 +1,105 @@
...
+  setScaleType: (scaleType: ScaleType) => void;
+
+  keepAspectRatio: boolean;
+  toggleAspectRatio: () => void;
+
+  showGrid: boolean;
+  toggleGrid: () => void;
+}
+
+function initialiseStore() {
```
axelboc, line 37: The diff is crap because I renamed `config.ts` to `config.tsx`, but here I just wrap the creation in a function so it can be called lazily by `useState` (see below).

src/h5web/vis-packs/core/heatmap/config.tsx
```diff
@@ -0,0 +1,105 @@
...
+  return useStore(selector, equalityFn);
+}
+
+// https://github.com/pmndrs/zustand/issues/128#issuecomment-673398578
+export function HeatmapConfigProvider(props: ConfigProviderProps) {
+  const { children } = props;
+  const [useStore] = useState(initialiseStore); // https://reactjs.org/docs/hooks-reference.html#lazy-initial-state
+
+  return <Provider value={useStore}>{children}</Provider>;
+}
```
axelboc, line 105: This is the new stuff.

src/h5web/vis-packs/core/line/config.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  return useStore(selector, equalityFn);
+}
+
+// https://github.com/pmndrs/zustand/issues/128#issuecomment-673398578
+export function LineConfigProvider(props: ConfigProviderProps) {
+  const { children } = props;
+  const [useStore] = useState(initialiseStore); // https://reactjs.org/docs/hooks-reference.html#lazy-initial-state
+
+  return <Provider value={useStore}>{children}</Provider>;
```
axelboc, line 99: This little `useState` "hack" is actually temporary. Zustand is working on a `ZustandProvider` that will take care of it: https://github.com/pmndrs/zustand/issues/182#issuecomment-803172558 (they use `useRef` instead of `useState`, but the behaviour is exactly the same).

src/h5web/visualizer/Visualizer.tsx
```diff
@@ -22,31 +22,33 @@ function Visualizer<T extends VisDef>(props: Props<T>) {
...
     <div className={styles.visualizer}>
-      <div className={styles.visBar}>
-        <VisSelector
-          activeVis={activeVis}
-          choices={supportedVis}
-          onChange={onActiveVisChange}
-        />
-        {Toolbar && <Toolbar />}
-      </div>
+      <ConfigProvider>
```
axelboc, line 29: I render the active visualization's config provider here. Every time the active vis changes, the provider is unmounted and then another one is mounted, which means the `useState` of the latter invokes its lazy initializer to recreate the config store.

src/setupTests.js
```diff
@@ -17,14 +15,9 @@ window.ResizeObserver = ResizeObserver;
...
-const initialLineConfig = useLineConfig.getState();
-const initialHeatmapConfig = useHeatmapConfig.getState();
-
-beforeEach(() => {
-  useLineConfig.setState(initialLineConfig, true);
-  useHeatmapConfig.setState(initialHeatmapConfig, true);
+// Clear localStorage after each test
+afterEach(() => {
+  window.localStorage.clear();
 });
```
axelboc, line 21: We no longer need to reset the stores themselves between tests; only `localStorage` like in `app.spec.ts`!

loichuder (approved):
Nice !

Thank for the comprehensive comments !

PR #576, opened 2021-03-23
Replace `fireEvent.click` with `userEvent.click`

axelboc:
`userEvent.click` simulates a proper browser click, with all the underlying events and behaviours: https://github.com/testing-library/user-event/blob/master/src/click.ts

### DIFF
--- a/src/h5web/test-utils.tsx
+++ b/src/h5web/test-utils.tsx
@@ -5,6 +5,7 @@ import {
   screen,
   within,
 } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import App from './App';
 import MockProvider from './providers/mock/MockProvider';
 import type { Vis } from './vis-packs/core/visualizations';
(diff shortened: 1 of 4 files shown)


src/h5web/test-utils.tsx
```diff
@@ -18,8 +19,11 @@ export function renderApp(): RenderResult {
 }
 
 export async function selectExplorerNode(path: string): Promise<void> {
-  for await (const name of path.split('/')) {
-    fireEvent.click(await screen.findByRole('treeitem', { name }));
+  const [name, ...remainingSegments] = path.split('/');
+  fireEvent.click(await screen.findByRole('treeitem', { name }));
+
+  if (remainingSegments.length > 0) {
+    await selectExplorerNode(remainingSegments.join('/'));
```
axelboc, line 26: Using recursion instead of a loop. Feels nicer.

For some reason, replacing this `fireEvent` with `userEvent` introduces a regression with one specific `NexusPack` test. React logs a warning "An update to MappedHeatmapVis inside a test was not wrapped in act(...).". I get what it means (the heatmap config is updated in a `useEffect` after the test has completed), but I don't get why I don't see the same regression in other tests, and why this particular change causes the regression. So... after an hour of investigation, it's time to give up.

loichuder (approved):

PR #578, opened 2021-03-24
Upgrade dependencies

axelboc:
Notable upgrades:

- **zustand** - better type safety + new `migrate` option in `persist` middleware called when store in `localStorage` has older version (to update it instead of erasing it): https://github.com/pmndrs/zustand/pull/292/files (I don't think we'll need this, but it's good to know.)
- **react-error-boundary** - fixes an issue I think I've seen happen: "First resetKeys update ignored when error triggered on mount" https://github.com/bvaughn/react-error-boundary/issues/84
- **react-reflex** - internal refactoring with `componentDidMount` and refs instead of `UNSAFE_componentWillReceiveProps` and `ReactDOM.findDOMNode()` - https://github.com/leefsmp/Re-Flex/commit/7cabce3eb18c83803a86b2243d080d23b1935ecb
- **@testing-library/user-event** - there's a new [`userEvent.keyboard()`](https://github.com/testing-library/user-event#keyboardtext-options) that can be used to simulate keyboard events outside of input fields (i.e. on the focused element). I should have upgraded before writing tests for the slider...
- **visx** - `@visx/grid` now supports [radial grids](https://airbnb.io/visx/lineradial) (`<GridRadial />` for cocentric circles, `<GridAngle />` for lines coming out of the centre, and `<GridPolar />` for both). Will be useful for Braggy!
- **TypeScript 4.2** - https://devblogs.microsoft.com/typescript/announcing-typescript-4-2/
  - `tsc --explainFiles` which tells us why each file is processed by TS (included by `tsconfig`, imported by a module, etc.) -- might be useful to debug issues in the future like the one we had because of an external libraries at some point.
  - They've reverted back to inferring template strings as `string` by default, unless you explicitely specify `as const`: https://devblogs.microsoft.com/typescript/announcing-typescript-4-2/#reverting-template-literal-inference


src/h5web/toolbar/controls/DomainSlider/DomainSlider.test.tsx
```diff
@@ -65,12 +65,12 @@ test('update domain when moving thumbs (with keyboard)', async () => {
...
 
   // Move thumb one step to the left
-  pressKey('ArrowLeft');
+  userEvent.keyboard('{ArrowLeft}');
   expect(minInput).toHaveValue('−1.04671e+2');
   expect(within(visArea).getByText('−1.047e+2')).toBeInTheDocument();
 
   // Move thumb five steps to the left (in a single press)
-  pressKey('ArrowLeft', 5);
+  pressKey('ArrowLeft', 5); // https://github.com/testing-library/user-event/issues/618
```
axelboc, line 73: I had to keep my `pressKey` utilitiy because of the issue I opened.

loichuder (approved):

PR #581, opened 2021-03-25
Minor optimisations

axelboc:


src/h5web/toolbar/controls/ColorMapSelector/ColorMapSelector.tsx
```diff
@@ -1,25 +1,11 @@
...
+import ToggleBtn from '../ToggleBtn';
 import { FiShuffle } from 'react-icons/fi';
-
-const COLORMAP_GROUPS = {
-  Recommended: Object.keys(RECOMMENDED) as ColorMap[],
-  'Single hue': Object.keys(SINGLE_HUE) as ColorMap[],
-  'Multi hue': Object.keys(MULTI_HUE) as ColorMap[],
-  Cyclical: Object.keys(CYCLICAL) as ColorMap[],
-  Diverging: Object.keys(DIVERGING) as ColorMap[],
-};
```
axelboc, line 22: I hard-code the arrays and move `COLORMAP_GROUPS` in a separate file (see below). It duplicates the `ColorMap` values a bit , but the typing is less hacky and we do a lot less work on initial load since there are no longer any `Object.keys()`.

src/h5web/toolbar/controls/DomainSlider/Thumb.tsx
```diff
@@ -4,12 +4,12 @@ import type { IconType } from 'react-icons/lib';
 import type { Bound } from '../../../vis-packs/core/models';
 import styles from './Thumb.module.css';
 
-type Props = HTMLProps<HTMLDivElement> & {
+interface Props extends HTMLProps<HTMLDivElement> {
```
axelboc, line 7: https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

src/h5web/vis-packs/core/heatmap/interpolators.ts
```diff
@@ -43,25 +43,16 @@ function interpolateHsl(t: number): string {
...
 }
 
-export const RECOMMENDED = {
-  Viridis: interpolateViridis,
-  Inferno: interpolateInferno,
-  Greys: interpolateGreys,
-  RdBu: interpolateRdBu,
-  Rainbow: interpolateRainbow,
-  Sinebow: interpolateSinebow,
-};
```
axelboc, line 53: The various interpolator groups are an implementation detail of the colormap selector. This file doesn't need to know about them. Declaring a single object also saves a few computations on initial load since we no longer need to destructure all the groups into a single object.

loichuder (approved):

Issue #519, opened 2021-02-23
Simplify internal HDF5 types


axelboc:
Apparently, most users are only interested in inspecting the `dtype` information and not the raw HDF5 types. For instance, h5py stores boolean datasets with a compound type (`H5T_COMPOUND`) with a `{ FALSE: 0, TRUE: 1 }` dictionary -- we don't need to show this -- the user just wants to know that it's a boolean, which can be inferred much more easily from the `dtype`.

We can therefore simplify a number of HDF5 types we have to better fit our front-end requirements.

axelboc:
Another simplification idea discussed in the chat: removing the concept of "shape" internally by treating scalars as datasets with no dimension `dims=[]` => `Dataset<HDF5Shape, HDF5Type>` would just become `Dataset<HDF5Type>` and every dataset would have a `dims` property.

axelboc:
We should [avoid large unions of interfaces](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-base-types-over-unions), like:

```ts
export type HDF5Type =
  | HDF5NumericType
  | HDF5StringType
  | HDF5BooleanType
  | HDF5ComplexType
  | HDF5Id
  | HDF5ArrayType
  | HDF5VLenType
  | HDF5CompoundType
  | HDF5EnumType;
```

axelboc:
Updates:

- The (soon-to-be-renamed) `HDF5Type` union stays because in some cases, TS is able to infer the correct union member inline without the need for type guards.
- HSDS cannot send the `dtype` in the dataset responses because dtypes are converted when datasets are loaded into object storage.
- The `Shape` generic must remain in `Dataset<Shape, Type>`. It is required to determine the type of the dataset's value from its type (i.e. for the conditional type in `Value<D extends Dataset>` to work).

With these updates in mind, the goal of this issue to simplify the internal HDF5 types remains. The scope is simply reduced. The work will be completed once `hdf5-models.ts` is deleted and all the remaining types unprefixed.

PR #586, opened 2021-03-31
Build `Value` type from dataset rather than shape and type

axelboc:
Part of #519 

A significant simplification of the `Value` type, which is used as the return type for the `useDatasetValue` and `useDatasetValues` hooks. Instead of two separate generics (`HDF5Shape` and `HDF5Type`), it now requires only one: `Dataset` (as this type already includes the types of the dataset's HDF5 shape and type).

### DIFF
--- a/src/h5web/providers/models.ts
+++ b/src/h5web/providers/models.ts
@@ -6,6 +6,7 @@ import type {
   HDF5Link,
   HDF5NumericType,
   HDF5Shape,
+  HDF5ScalarShape,
   HDF5SimpleShape,
   HDF5StringType,
   HDF5Type,
@@ -60,10 +61,11 @@ type PrimitiveType<T extends HDF5Type> = T extends HDF5NumericType
   ? Complex
   : unknown;
 
-export type Value<
-  S extends HDF5Shape,
-  T extends HDF5Type
-> = S extends HDF5SimpleShape ? PrimitiveType<T>[] : PrimitiveType<T>;
+export type Value<D extends Dataset> = D['shape'] extends HDF5SimpleShape
+  ? PrimitiveType<D['type']>[]
+  : D['shape'] extends HDF5ScalarShape
+  ? PrimitiveType<D['type']>
+  : never;
 
 export type ComplexArray = (ComplexArray | Complex)[];
 
--- a/src/h5web/vis-packs/core/hooks.ts
+++ b/src/h5web/vis-packs/core/hooks.ts
@@ -13,24 +13,20 @@ import {
 } from './utils';
 import AxisSystemContext from './shared/AxisSystemContext';
 import type { AxisScale } from './models';
-import type { HDF5Shape, HDF5Type } from '../../providers/hdf5-models';
 import { ProviderContext } from '../../providers/context';
 import type { Dataset, Value } from '../../providers/models';
 import { isAxis } from '../../dimension-mapper/utils';
+import type { HDF5Value } from '../../providers/hdf5-models';
 
-export function useDatasetValue<
-  S extends HDF5Shape,
-  T extends HDF5Type,
-  D extends Dataset<S, T> | undefined
->(
+export function useDatasetValue<D extends Dataset | undefined>(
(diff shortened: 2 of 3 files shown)


src/h5web/providers/models.ts
```diff
@@ -60,10 +61,11 @@ type PrimitiveType<T extends HDF5Type> = T extends HDF5NumericType
...
 
-export type Value<
-  S extends HDF5Shape,
-  T extends HDF5Type
-> = S extends HDF5SimpleShape ? PrimitiveType<T>[] : PrimitiveType<T>;
+export type Value<D extends Dataset> = D['shape'] extends HDF5SimpleShape
+  ? PrimitiveType<D['type']>[]
+  : D['shape'] extends HDF5ScalarShape
+  ? PrimitiveType<D['type']>
+  : never;
```
axelboc, line 68: `D['shape']` and `D['type']` are used to access the TS types of the dataset's HDF5 shape and type.

Note that I added a condition on `HDF5ScalarShape` with a fallback to deal with the `null` shape. We won't technically need it, but it feels more robust and complete.

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -13,24 +13,19 @@ import {
...
-export function useDatasetValue<
-  S extends HDF5Shape,
-  T extends HDF5Type,
-  D extends Dataset<S, T> | undefined
->(
+export function useDatasetValue<D extends Dataset | undefined>(
   dataset: D,
   dimMapping?: DimensionMapping
-): D extends Dataset<infer S, infer T> ? Value<S, T> : undefined;
+): D extends Dataset ? Value<D> : undefined;
```
axelboc, line 23: The override definition is much simpler ! No need to `infer` anything.

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -13,24 +13,19 @@ import {
...
 
 export function useDatasetValue(
   dataset: Dataset | undefined,
   dimMapping?: DimensionMapping
-) {
+): unknown {
```
axelboc, line 28: Adding explicit return type for clarity and performance reasons.

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -50,16 +45,15 @@ export function useDatasetValue(
...
+
+export function useDatasetValues(datasets: Dataset[]): Record<string, unknown> {
   const { valuesStore } = useContext(ProviderContext);
 
   return Object.fromEntries(
-    datasets.map(({ name, path }) => [
-      name,
-      valuesStore.get({ path }) as Value<S, T>,
-    ])
+    datasets.map(({ name, path }) => [name, valuesStore.get({ path })])
```
axelboc, line 56: Here I'm replacing the type cast with a return type of `Record<string, unknown>` and an override with the return type that the caller cares about, `Record<string, Value<D>>`, thus clearly separating the typing of the function's internal declaration and external usage.

src/h5web/vis-packs/core/hooks.ts
loichuder: Could you elaborate on this ? I don't get the reasons :smile:

loichuder (approved):

src/h5web/vis-packs/core/hooks.ts
axelboc: - Performance (granted, it's probably not a big deal in this case... 😄): https://github.com/microsoft/TypeScript/wiki/Performance#using-type-annotations
- The benefit I mention in my other comment also applies here: I'm now clearly identifying the typing of the function's internal declaration (as opposed to the typing of the function's external usage provided by the override). Since the two typings are different, explicitness seems worthwhile.
- The return type is short and leaving it out doesn't improve the formatting of the code (the function signature still gets wrapped over multiple lines), so there's no downside to having it, really.

That being said, I could use `HDF5Value` instead of `unknown`, I forgot about it!

PR #587, opened 2021-03-31
Simplify representation of dataset shapes and handle empty datasets

axelboc:
Part of #519 

I keep the `Dataset<S, T>` syntax so that we can still have a type that holds both the HDF5 shape and type of a given dataset. This simplifies the definition and use of type guards.

---

The alternative of having three types (`SimpleDataset<T>`, `ScalarDataset<T>` and `EmptyDataset<T>`) is problematic in a number of ways. Notably, it makes it too tedious to write `HDF5Type` type guards that "remember" the datasets' shapes, which means that when we write:

```ts
assertScalarShape(dataset);
assertNumericType(dataset);`
```

... then `dataset` gets inferred as `ScalarDataset<HDF5Type> & Dataset<HDF5NumericType>`, which is confusing and puts unnecessary strain on the TS compiler. We have to swap the two assertions in order to get `ScalarDataset<HDF5NumericType>`, which is too restrictive in my opinion.

---

I also add official support for datasets with `null` shape. Well, by support I mean show "Nothing to visualize", which I think is actually better than to show an error! I achieve this by simply changing the support function of the raw visualization, so that an empty dataset ends up with no supported visualization.

image

image


src/h5web/guards.ts
```diff
@@ -87,19 +86,36 @@ export function isHardLink(link: HDF5Link): link is HDF5HardLink {
...
 export function hasScalarShape<T extends HDF5Type>(
-  dataset: Dataset<HDF5Shape, T>
-): dataset is Dataset<HDF5ScalarShape, T> {
-  return dataset.shape.class === HDF5ShapeClass.Scalar;
+  dataset: Dataset<Shape, T>
+): dataset is Dataset<ScalarShape, T> {
+  return isScalarShape(dataset.shape);
+}
+
+export function hasSimpleShape<T extends HDF5Type>(
```
loichuder, line 99: `hasArrayShape` ? I feel `SimpleShape` is not very straightforward and we no longer need to match shape class.

src/h5web/providers/hsds/utils.ts
```diff
@@ -42,6 +44,17 @@ export function assertHsdsDataset(
...
+export function convertHsdsShape(shape: HDF5Shape): Shape {
+  switch (shape.class) {
+    case HDF5ShapeClass.Simple:
+      return shape.dims;
+    case HDF5ShapeClass.Scalar:
+      return [];
+    default:
+      return null;
+  }
+}
```
loichuder, line 56: Nice :+1:

src/h5web/providers/jupyter/api.ts
```diff
@@ -132,10 +127,7 @@ export class JupyterApi implements ProviderAPI {
         ...baseEntity,
         kind,
         type: convertDtype(dtype),
-        shape:
-          dims.length > 0
-            ? { class: HDF5ShapeClass.Simple, dims }
-            : { class: HDF5ShapeClass.Scalar },
+        shape: dims.length > 0 ? dims : [],
```
loichuder, line 130: This can be simplified to `shape: dims`, right ? This way, it will also be ready for the support for null shapes that will come in the next release.

Or is there an edge case that I don't get ?

src/h5web/guards.ts
```diff
@@ -180,26 +189,36 @@ export function assertGroup(
...
-  dataset: Dataset<HDF5Shape, T>
-): asserts dataset is Dataset<HDF5SimpleShape, T> {
+  dataset: Dataset<Shape, T>
+): asserts dataset is Dataset<SimpleShape, T> {
   if (!hasSimpleShape(dataset)) {
     throw new Error('Expected dataset to have simple shape');
   }
+}
+
+export function assertShape<T extends HDF5Type>(
```
loichuder, line 207: Do you find `assertShape` clear enough ?

loichuder (approved):
Going in the right direction !

Do you think we should now name it `dims` instead of `shape` ? It would solve the renaming in all VisContainers and make `assertShape` clearer (could be renamed to `assertDims` or `assertHasDims` )

src/h5web/providers/jupyter/api.ts
axelboc: 😂

PR #588, opened 2021-04-01
Simplify representation of attribute shapes

axelboc:
Part of #519 

Same as #587 but with attributes.

### DIFF
--- a/src/h5web/providers/hsds/utils.ts
+++ b/src/h5web/providers/hsds/utils.ts
@@ -3,6 +3,7 @@ import Complex from 'complex.js';
 import { isString } from 'lodash-es';
 import { isDataset, isGroup } from '../../guards';
 import {
+  HDF5Attribute,
   HDF5Endianness,
   HDF5FloatType,
   HDF5IntegerType,
@@ -11,7 +12,7 @@ import {
   HDF5Type,
   HDF5TypeClass,
 } from '../hdf5-models';
-import type { Entity, ComplexArray, Shape } from '../models';
+import type { Entity, ComplexArray, Shape, Attribute } from '../models';
 import type {
   HsdsLink,
   HsdsExternalLink,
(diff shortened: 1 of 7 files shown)


src/h5web/providers/mock/metadata-utils.ts
```diff
@@ -216,22 +214,15 @@ export function makeExternalLink(
 /* ----------------- */
 /* ----- NEXUS ----- */
 
-function makeNxStrArrayAttr(
-  array: string[],
-  attrName: 'axes' | 'auxiliary_signals'
-): HDF5Attribute {
-  return makeAttr(attrName, makeSimpleShape([array.length]), stringType, array);
-}
```
axelboc, line 224: Removing this function and calling `makeAttr` directly from the two functions below.

src/h5web/providers/models.ts
```diff
@@ -37,10 +37,7 @@ export interface Dataset<S extends Shape = Shape, T extends HDF5Type = HDF5Type>
   type: T;
 }
 
-export type Shape = ArrayShape | ScalarShape | NullShape;
-export type ArrayShape = HDF5Dims;
-export type ScalarShape = never[];
-export type NullShape = null;
+export type NumArrayDataset = Dataset<ArrayShape, HDF5NumericType>;
```
axelboc, line 40: Just moving these things around.

src/h5web/providers/models.ts
```diff
@@ -52,6 +49,18 @@ export interface Link<T extends HDF5Link = HDF5Link> extends Entity {
   rawLink: T;
 }
 
+export interface Attribute {
+  name: string;
+  shape: Shape;
+  type: HDF5Type;
+  value: HDF5Value;
+}
```
axelboc, line 57: New unprefixed type.

src/h5web/providers/hsds/utils.ts
```diff
@@ -149,6 +150,13 @@ export function convertHsdsType(hsdsType: HsdsType): HDF5Type {
   }
 }
 
+export function convertHsdsAttributes(attrs: HDF5Attribute[]): Attribute[] {
+  return attrs.map((attr) => ({
+    ...attr,
+    shape: convertHsdsShape(attr.shape),
```
loichuder, line 156: Actually, do the `type` of attribute need to be converted as well ?

The type `HDF5Attribute[]` might be wrong here.

loichuder (approved):

src/h5web/providers/hsds/utils.ts
axelboc: I think you're right. It probably comes from:

```ts
export type HsdsAttributeWithValueResponse = HDF5Attribute;
```

axelboc: Fixed in my other PR. Will merge in here once you confirm it's all good: https://github.com/silx-kit/h5web/pull/589/commits/52933a8c05342e4f51613d8505b6a19c66809d76

PR #589, opened 2021-04-01
Enforce stricter typing in HSDS API

axelboc:


src/h5web/providers/hsds/api.ts
```diff
@@ -36,14 +35,15 @@ import {
...
+  assertHsdsEntity,
 } from './utils';
 import { buildEntityPath, getChildEntity } from '../../utils';
 
 export class HsdsApi implements ProviderAPI {
   public readonly filepath: string;
   private readonly client: AxiosInstance;
 
-  private readonly entities = new Map<string, Entity>();
+  private readonly entities = new Map<string, HsdsEntity>();
```
axelboc, line 46: The entities we store have IDs, so I replace types `HsdsDataset` and `HsdsGroup` with a more generic `HsdsEntity<Entity>`.

src/h5web/providers/hsds/api.ts
```diff
@@ -82,6 +82,7 @@ export class HsdsApi implements ProviderAPI {
     const childName = path.slice(path.lastIndexOf('/') + 1);
     const child = getChildEntity(parentGroup, childName);
     assertDefined(child);
+    assertHsdsEntity(child);
 
     const entity = isHsdsGroup(child)
```
axelboc, line 87: `isHsdsGroup` no longer checks if the entity has an ID. Instead, it expects an HSDS entity.

loichuder (commented):
In which manner is this typing stricter ?

axelboc:
Beforehand, this would not fail if `getChildEntity` returned an entity without an `id` (`entity` would just end up with `child` even if it were a group entity):

image

Granted, though, it's not significantly stricter...

loichuder:
> , though, it's not significantly stricter...

I beg to differ: it catches a non-trivial bug so :+1: from me

PR #590, opened 2021-04-01
Move `HDF5Shape` types to HSDS provider and simplify them

axelboc:
I can simplify the new `HsdsShape` type _a lot_ by being a bit less descriptive of what HSDS returns and by simplifying the conversion to `Shape`.

I'm also able to remove `HDF5Attribute`! 🎉

### DIFF
--- a/src/h5web/providers/hsds/models.ts
+++ b/src/h5web/providers/hsds/models.ts
@@ -1,7 +1,5 @@
 import type {
   HDF5Id,
-  HDF5Shape,
-  HDF5Attribute,
   HDF5Value,
   HDF5ExternalLink,
   HDF5HardLink,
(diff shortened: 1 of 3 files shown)


src/h5web/providers/hsds/models.ts
```diff
@@ -11,6 +9,72 @@ import type {
...
+  attributeCount: number;
+}
+
+export interface HsdsDatatypeResponse {
+  id: HDF5Id;
+  type: HsdsType;
+}
+
+export interface HsdsAttributesResponse {
+  attributes: Omit<HsdsAttributeWithValueResponse, 'value'>[];
```
axelboc, line 38: I moved things around, so the diff is ugly, but here I'm now using `HsdsAttributeWithValueResponse` instead of `HDF5Attribute`, which is what allows me to remove the latter.

src/h5web/providers/hsds/models.ts
```diff
@@ -11,6 +9,72 @@ import type {
...
+export type HsdsComplex = HsdsComplex[] | HsdsComplexValue;
+export type HsdsComplexValue = [number, number];
+
+/* ------------------------ */
+/* ----- SHAPE & TYPE ----- */
+
+export interface HsdsShape {
+  class: 'H5S_SIMPLE' | 'H5S_SCALAR' | 'H5S_NUL';
+  dims?: HDF5Dims;
+}
```
axelboc, line 76: The new HSDS shape type. I don't specify that `dims` is only provided with `H5S_SIMPLE`, which allows the conversion function to be much simpler. 😄 We don't need separate shape types for inferring things later on, so it's not worth having them. I might do something similar with `HSDSType`, I have looked yet.

src/h5web/providers/hsds/models.ts
loichuder: Could we change the clunky `HsdsAttributeWithValueResponse` into `HsdsAttributeReponse` ?

loichuder (approved):

PR #591, opened 2021-04-07
Allow inspection of unknown types

axelboc:
- I add a type called `UnknownType` to catch any type we don't yet support so users can still inspect them in the metadata viewer.
- I stop supporting `HDF5Id` as a valid type since we don't yet have proof that this type is returned by our back-end providers. If we discover that it can, then we'll re-introduce it but in a more meaningful way (e.g. for instance by resolving the datatype in the provider so the front doesn't "see" the ID).

### DIFF
--- a/src/h5web/providers/hdf5-models.ts
+++ b/src/h5web/providers/hdf5-models.ts
@@ -51,7 +51,6 @@ export type HDF5Type =
   | HDF5StringType
   | HDF5BooleanType
   | HDF5ComplexType
-  | HDF5Id
   | HDF5ArrayType
   | HDF5VLenType
   | HDF5CompoundType
--- a/src/h5web/providers/hdf5-models.ts
+++ b/src/h5web/providers/hdf5-models.ts
@@ -54,20 +54,22 @@ export type HDF5Type =
   | HDF5ArrayType
   | HDF5VLenType
   | HDF5CompoundType
-  | HDF5EnumType;
+  | HDF5EnumType
+  | HDF5UnknownType;
 
 export type HDF5NumericType = HDF5IntegerType | HDF5FloatType;
 
 export enum HDF5TypeClass {
+  Bool = 'H5T_BOOL',
   Integer = 'H5T_INTEGER',
   Float = 'H5T_FLOAT',
+  Complex = 'H5T_COMPLEX',
   String = 'H5T_STRING',
+  Compound = 'H5T_COMPOUND',
   Array = 'H5T_ARRAY',
   VLen = 'H5T_VLEN',
-  Compound = 'H5T_COMPOUND',
   Enum = 'H5T_ENUM',
-  Bool = 'H5T_BOOL',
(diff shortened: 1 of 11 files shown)


src/h5web/providers/hdf5-models.ts
```diff
@@ -51,24 +51,25 @@ export type HDF5Type =
...
+  Bool = 'boolean',
+  Integer = 'integer',
+  Float = 'float',
+  Complex = 'complex',
+  String = 'string',
+  Compound = 'compound',
+  Array = 'array',
+  VLen = 'vlen',
+  Enum = 'enum',
+  Unknown = 'unknown',
```
axelboc, line 72: New `Unknown` class. This will allow every type to inherit the same `{ class: HDF5TypeClass }` interface in my next PR.

src/h5web/providers/hsds/api.ts
```diff
@@ -210,6 +210,7 @@ export class HsdsApi implements ProviderAPI {
       attributes: convertHsdsAttributes(attributes),
       shape: convertHsdsShape(shape),
       type: convertHsdsType(type),
+      rawType: type,
```
axelboc, line 213: Adding the raw type for inspection purposes.

src/h5web/providers/hdf5-models.ts
```diff
@@ -128,3 +129,7 @@ interface HDF5CompoundTypeField {
   name: string;
   type: HDF5Type;
 }
+
+interface HDF5UnknownType {
+  class: HDF5TypeClass.Unknown;
+}
```
loichuder, line 135: What about adding the rawType in this interface ?

loichuder (approved):

src/h5web/providers/hdf5-models.ts
axelboc: You mean instead of having it on the `Dataset` or `Datatype`? The benefit of the current solution is that we can inspect the raw type if we recognise the HSDS type but don't process it correctly. This could be useful for detecting problems with the special types like "boolean as enum" or "complex as compound", or maybe others in the future.

loichuder: I see. Still, it could make more sense to me to add to `HDF5Type` as `raw`, no ?

axelboc: Earlier this afternoon, I would have said yes, because I tried switching to `interface HDF5Type { class: HDF5TypeClass }` along with `extends HDF5Type`. I would then have been able to add a `raw` property to the base class.

However, I ran into a lot of issues with this typing and I had to revert back to `type HDF5Type = HDF5NumericType | HDF5StringType | ...`. And in this case, I'd have to add `raw` to every sub-type...

axelboc: I think having `rawType` on `Dataset` is actually consistent with `rawLink` on `Link`.

PR #592, opened 2021-04-07
Simplify HDF5 types and separate out HDF5 from HSDS type classes

axelboc:
I wasn't able to go as far as I'd planned by replacing the union of HDF5 types with `interface HDF5Type { class: HDF5TypeClass; }`. This led to some issues, notably when using `HDF5Type` as a return type of the provider's conversion functions. I realised that the significant benefit of a type union `A | B | C` is that TypeScript is able to infer the sub-types `A`, `B` and `C` inline, without the need for type guards.

So I've made only two changes:

1. I've merged a few HDF5 type interfaces together:
    - `HDF5FloatType` and `HDF5IntegerType` are now represented with a single `HDF5NumericType` interface.
    - Unsigned integers now have their own `HDF5TypeClass` entry and are also represented with `HDF5NumericType`.
    - `HDF5VLenType` is now represented by `HDF5ArrayType` (by making the latter's `dims` property optional).
2. I've completely separated out the HDF5 and HSDS type classes:
    - `HDF5TypeClass` now contains readable strings that we can render directly in the metadata viewer, like `integer`, 'boolean', etc.
    - `H5T_INTEGER` and the like, which are the classes returned by HSDS are no longer present anywhere outside of the HSDS provider code. This means we no longer have classes like `H5T_BOOL` that are never returned by HSDS.

### DIFF
--- a/src/h5web/metadata-viewer/utils.ts
+++ b/src/h5web/metadata-viewer/utils.ts
@@ -1,4 +1,4 @@
-import { isScalarShape } from '../guards';
+import { isNumericType, isScalarShape } from '../guards';
 import { HDF5Type, HDF5TypeClass } from '../providers/hdf5-models';
 import type { Shape } from '../providers/models';
 
@@ -24,32 +24,22 @@ export function renderShape(shape: Shape): string {
 }
 
 export function renderType(type: HDF5Type): string {
-  // Remove leading `H5T_`
-  const classLabel = type.class.slice(4).toLowerCase();
-
-  if (
-    type.class === HDF5TypeClass.Integer ||
-    type.class === HDF5TypeClass.Float
-  ) {
+  if (isNumericType(type)) {
     const { endianness, size } = type;
-    const typeInfos = `${classLabel}, ${
-      type.class === HDF5TypeClass.Integer && type.unsigned ? 'unsigned' : ''
-    } ${size}-bit`;
+    const endiannessLabel = ENDIANNESS_LABELS[endianness];
 
-    if (endianness === 'Not applicable') {
-      return typeInfos;
-    }
-
-    return `${typeInfos}, ${ENDIANNESS_LABELS[endianness]}`;
+    return `${type.class}, ${size}-bit${
+      endiannessLabel ? `, ${endiannessLabel}` : ''
+    }`;
   }
 
   if (type.class === HDF5TypeClass.String) {
     const { length, charSet } = type;
 
-    return `${classLabel},
-    ${charSet},
-    ${Number.isInteger(length) ? `${length} characters` : 'variable length'}`;
+    return `${type.class}, ${charSet}, ${
+      Number.isInteger(length) ? `${length} characters` : 'variable length'
+    }`;
   }
 
-  return classLabel;
+  return type.class;
 }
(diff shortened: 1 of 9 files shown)


src/h5web/metadata-viewer/utils.ts
```diff
@@ -24,32 +24,22 @@ export function renderShape(shape: Shape): string {
...
-      type.class === HDF5TypeClass.Integer && type.unsigned ? 'unsigned' : ''
-    } ${size}-bit`;
-
-    if (endianness === 'Not applicable') {
-      return typeInfos;
-    }
+    const endiannessLabel = ENDIANNESS_LABELS[endianness];
 
-    return `${typeInfos}, ${ENDIANNESS_LABELS[endianness]}`;
+    return `${type.class}, ${size}-bit${
```
axelboc, line 31: `HDF5TypeClass.Unsigned` is `"unsigned integer"`, so we no longer need the extra ternary here.

src/h5web/providers/hsds/models.ts
```diff
@@ -76,54 +75,37 @@ export interface HsdsShape {
...
   base: string;
 }
 
 export interface HsdsStringType {
-  class: HDF5TypeClass.String;
+  class: 'H5T_STRING';
   charSet: 'H5T_CSET_ASCII' | 'H5T_CSET_UTF8';
   strPad: 'H5T_STR_SPACEPAD' | 'H5T_STR_NULLTERM' | 'H5T_STR_NULLPAD';
   length: number | 'H5T_VARIABLE';
 }
```
axelboc, line 94: I'll make some further simplifications here in my next PR.

src/h5web/providers/hsds/utils.ts
```diff
@@ -74,76 +73,79 @@ export function convertHsdsNumericType(
...
+  return {
+    class: HDF5TypeClass.Compound,
+    fields: hsdsType.fields.map((v) => ({
+      name: v.name,
+      type: convertHsdsType(v.type),
+    })),
   };
 }
 
 export function convertHsdsType(hsdsType: HsdsType): HDF5Type {
```
axelboc, line 109: I've reorder the cases to match the `HsdsType` type union and extracted `convertHsdsCompoundType`

loichuder (changes requested):
While this is taking the right direction, the resulting naming brings a lot of confusion.

`HDF5Type` are now further and further from the "real" types of the HDF5 spec. In fact, "separate out HDF5 from HSDS type classes" is a non-sense as it is HSDS that holds the "real" HDF5 types.

We must clearly signify in the naming that the interface `HDF5Type` is now a convenience type for h5web (as we did with `Shape`).

Also, I would have copied the (now old) `enum` of `HDF5TypeClass` into the HSDS part.

axelboc:
Renaming/moving the types is the end goal, but renaming before simplifying would just make everything more complicated. As I've done in past refactorings, I'll do all the renaming and moving at the end.

`HDF5TypeClass` is _no longer_ an HSDS enum. It will be unprefixed and moved to `models` in my final PR.

loichuder:
I see. Sorry for jumping the gun :sweat_smile:, go ahead and merge :+1:

loichuder (approved):

PR #594, opened 2021-04-08
Move and unprefix dtypes

axelboc:
Part of #519


src/h5web/providers/hdf5-models.ts
```diff
@@ -41,78 +41,3 @@ export interface HDF5ExternalLink {
...
-
-interface HDF5EnumType {
-  class: HDF5TypeClass.Enum;
-  base: HDF5Type;
-  mapping: Record<string, number>;
-}
-
-interface HDF5UnknownType {
-  class: HDF5TypeClass.Unknown;
-}
```
loichuder, line 118: :smiling_imp:

loichuder (approved):

PR #595, opened 2021-04-09
Remove `HDF5Dims` type

axelboc:
Just a quick one so it doesn't clutter the next one with the links.

### DIFF
--- a/src/h5web/providers/hdf5-models.ts
+++ b/src/h5web/providers/hdf5-models.ts
@@ -3,7 +3,6 @@
 
 export type HDF5Id = string;
 export type HDF5Value = unknown;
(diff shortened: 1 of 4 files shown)


src/h5web/providers/hdf5-models.ts
```diff
@@ -3,7 +3,6 @@
 
 export type HDF5Id = string;
 export type HDF5Value = unknown;
-export type HDF5Dims = number[];
```
axelboc, line 6: I just don't think it adds any value, to be honest, that's all.

loichuder (approved):
Good catch :fishing_pole_and_fish:

PR #596, opened 2021-04-09
Rethink links

axelboc:
Part of #519 

This PR introduces a fundamental change in the way we represent HDF5 links: the `Link` model no longer extends `Entity` and it no longer is a "kind" of entity in its own right.

If an entity can't be resolved (i.e. if the back-end provider can't resolve the underlying HDF5 link), then it remains an `Entity` object. Otherwise, it becomes a `Dataset` or `Group` object.

Just like before, if the provider is "link-aware", it may store a `Link` object in the `link` property of each entity (previously `rawLink`).

Here is what an unresolved soft-link entity looks like in the metadata viewer with HSDS:

image

... and a mocked unresolved external-link entity:

image

I don't show anything for unresolved hard-link entities, because a hard link is, in my opinion, an internal optimisation performed by HDF5 libraries (as opposed to a feature that people use on purpose, like soft and external links).


src/h5web/explorer/Icon.tsx
```diff
@@ -34,7 +33,7 @@ function Icon(props: Props) {
     );
   }
 
-  const LeafIcon = LEAF_ICONS[entity.kind];
+  const LeafIcon = entity.kind ? LEAF_ICONS[entity.kind] : FiLink;
```
axelboc, line 36: The link icons is now used as a fallback, when an entity doesn't have a `kind` (i.e. when it's unresolved).

src/h5web/metadata-viewer/EntityTable.tsx
```diff
@@ -47,7 +46,15 @@ function EntityTable(props: Props) {
...
-        {isLink(entity) && entity.rawLink && <LinkInfo link={entity.rawLink} />}
+        {entity.link && entity.link.path && (
+          <tr>
+            <th scope="row">{entity.link.class} link</th>
+            <td>
+              {entity.link.file && `${entity.link.file}:`}
+              {entity.link.path}
+            </td>
+          </tr>
+        )}
```
axelboc, line 57: If the underlying `link` of the entity has a path (typically if it's soft or external), then we show a row labelled "Soft/External link" containing the `path` to the linked entity (including the `file`, for external links).

src/h5web/providers/hdf5-models.ts
```diff
@@ -1,42 +1,2 @@
...
-  title: string;
-  h5path: string;
-}
-
-export interface HDF5ExternalLink {
-  class: HDF5LinkClass.External;
-  title: string;
-  file: string;
-  h5path: string;
-}
```
axelboc, line 42: Moved to HSDS and simplified.

src/h5web/providers/hsds/api.ts
```diff
@@ -232,28 +221,31 @@ export class HsdsApi implements ProviderAPI {
   }
 
   private async resolveLink(
-    link: HDF5Link,
+    link: HsdsLink,
     path: string,
     depth: number
   ): Promise<Entity> {
-    if (!isHardLink(link)) {
+    if (link.class !== 'H5L_TYPE_HARD') {
```
axelboc, line 228: Like before, we don't attempt to resolve soft and external links.

src/h5web/providers/hsds/api.ts
```diff
@@ -232,28 +221,31 @@ export class HsdsApi implements ProviderAPI {
...
         name: link.title,
         path,
-        kind: EntityKind.Link,
         attributes: [],
-        rawLink: link,
+        link: {
+          class: link.class === 'H5L_TYPE_SOFT' ? 'Soft' : 'External',
+          path: link.h5path,
+          file: link.file,
+        },
```
axelboc, line 237: HSDS is "link-aware", so we store a `Link` object in the entity. Note that we no longer specify the `kind` property to indicate that the entity is unresolved.

src/h5web/providers/hsds/models.ts
```diff
@@ -51,17 +45,31 @@ export interface HsdsValueResponse {
...
+}
 
-export interface HsdsExternalLink extends Omit<HDF5ExternalLink, 'file'> {
-  h5domain: string;
+interface HsdsSymbolicLink {
+  class: 'H5L_TYPE_SOFT' | 'H5L_TYPE_EXTERNAL';
+  title: string;
+  file?: string;
+  h5path: string;
 }
```
axelboc, line 66: I've merged the soft and external links models into one. I think the name `HsdsSymbolicLink` makes sense, but note that it doesn't matter too much since it's not exported.

src/h5web/providers/jupyter/api.ts
```diff
@@ -132,15 +132,6 @@ export class JupyterApi implements ProviderAPI {
...
-    return {
-      ...baseEntity,
-      kind: EntityKind.Link,
-      rawLink: {
-        class: HDF5LinkClass.Soft,
-        title: path,
-        h5path: path,
-      },
-    };
+    throw new Error('Unknown response');
```
axelboc, line 135: JLab is not yet "link-aware", so this should not be reachable.

src/h5web/providers/mock/utils.test.ts
```diff
@@ -1,14 +1,11 @@
...
 
 describe('findMockEntity', () => {
   it('should return entity at given path', () => {
     expect(findMockEntity('/')).toBe(mockMetadata);
-    expect(findMockEntity('/nD_datasets')).toBe(mockMetadata.children[1]);
-    expect(findMockEntity('/entities/raw')).toBe(
-      (mockMetadata.children[0] as Group).children[4]
-    );
+    expect(findMockEntity('/nD_datasets').kind).toBe('group');
+    expect(findMockEntity('/entities/raw').name).toBe('raw');
```
axelboc, line 8: I keep having to fix this test when I make changes to the mock entities, so it's time I made it more reliable...

src/h5web/providers/models.ts
```diff
@@ -8,15 +8,14 @@ export enum EntityKind {
...
 }
 
 export interface Entity {
   name: string;
   path: string;
-  kind: EntityKind;
+  kind?: EntityKind;
   attributes: Attribute[];
-  rawLink?: HDF5Link;
+  link?: Link;
```
axelboc, line 18: The gist of it!

src/h5web/providers/models.ts
```diff
@@ -32,17 +31,17 @@ export interface Dataset<S extends Shape = Shape, T extends DType = DType>
...
 }
 
-export interface Link<T extends HDF5Link = HDF5Link> extends Entity {
-  kind: EntityKind.Link;
-  rawLink: T;
+export type LinkClass = 'Hard' | 'Soft' | 'External';
+export interface Link {
+  class: LinkClass;
+  path?: string;
+  file?: string;
```
axelboc, line 44: As discussed, those are the only information that seem relevant for H5Web. A hard link is represented with just `class`; a soft link with `class` and `path`; and an external link with `class`, `path` and `file`.

Links are only really used by the metadata viewer to display the path/file if present, so we don't need any more typing than that.

axelboc:
Next and final steps:

- move `HDF5Id` to HSDS and rename if `HsdsId`
- deal with `HDF5Value` somehow
- remove `hdf5-models.ts`

src/h5web/providers/jupyter/api.ts
loichuder: It will be reached for datatypes.

src/h5web/providers/hsds/models.ts
loichuder: Works fine for me

loichuder (approved):
I like how this turns out but I have a small remark.

An unresolved entity is identified by the fact that `kind` is `undefined`.  I feel that this is a somewhat implicit way of linking the two and you make use of it two times:
- For `LeafIcon`
- For the header of `EntityTable`

Looking at the PR, I don't see why we could not keep `kind`, make this relation explicit and keep the generic behaviour at the two places I mentioned. Perhaps `Link` is no longer appropriate but we could have `UnresolvedEntity` or something similar ?

loichuder (approved):

src/h5web/providers/jupyter/api.ts
axelboc: I've reverted the fallback, but now I'm just returning an unresolved entity without a link.

PR #597, opened 2021-04-12
Move and rename HDF5Id to HsdsId

axelboc:


src/h5web/providers/hdf5-models.ts
```diff
@@ -1,2 +1 @@
-export type HDF5Id = string;
 export type HDF5Value = unknown;
```
axelboc, line 1: Almost there... 😄

loichuder (approved):

PR #599, opened 2021-04-12
Update slicing sliders on resize and clip ends of tracks

### LINKED ISSUE #585: [DimensionMapper] Update when height changes
When H5Web is rendered in a container with `resize: vertical` and the container is resized, the sliders of the `DimensionMapper` don't detect the resize and break. We could use `useMeasure` and pass the height as `key` to the sliders.

axelboc:
Fix #585 

### BEFORE
image

### AFTER
image


src/h5web/dimension-mapper/SlicingSlider.module.css
```diff
@@ -4,7 +4,7 @@
   flex-direction: column;
   padding: 0 0.5rem;
   --thumb-height: 1.5rem;
-  --mark-height: 0.4375rem;
+  --mark-height: 0.5rem;
```
axelboc, line 7: So the marks are horizontally centred on the track. (There was a 1px misalignment.)

src/h5web/dimension-mapper/SlicingSlider.tsx
```diff
@@ -15,7 +16,14 @@ interface Props {
...
 function SlicingSlider(props: Props) {
   const { dimension, slicingIndex, rawDims, mapperState, onChange } = props;
+
   const [containerRef, { height }] = useMeasure();
+  const sliderRef = useRef<Component<ReactSlider.ReactSliderProps>>(null);
+
+  useEffect(() => {
+    // @ts-ignore
+    sliderRef.current?.resize();
+  }, [height]);
```
axelboc, line 26: The fix for the resizing.

src/h5web/dimension-mapper/SlicingSlider.tsx
```diff
@@ -25,29 +33,34 @@ function SlicingSlider(props: Props) {
...
       <span className={styles.label}>D{dimension}</span>
       <ReactSlider
-        // Force refresh when slider height changes - i.e. when Y-axis mapper appears/disappears
-        // https://github.com/zillow/react-slider/issues/172
-        key={`${mapperState.includes('y')}`}
+        ref={sliderRef}
         className={styles.slider}
         ariaLabel="Dimension slider"
-        trackClassName={styles.track}
-        thumbClassName={styles.thumb}
```
axelboc, line 34: I moved those class names inside the `renderThumb/Tracks` props and just reorder the other props.

src/h5web/dimension-mapper/SlicingSlider.tsx
```diff
@@ -25,29 +33,34 @@ function SlicingSlider(props: Props) {
...
+            {state.valueNow}
+          </div>
+        )}
+        renderTrack={({ key }, { index }) =>
+          index === 0 ? (
+            <div key={key} className={styles.track} />
+          ) : (
+            <Fragment key={key} />
+          )
+        }
```
axelboc, line 63: Custom render for the track so we only render one track and style it so it doesn't reach the ends of the slider.

loichuder: Can you now put `null` instead of `Fragment` ?

src/h5web/dimension-mapper/SlicingSlider.tsx
loichuder: Why the `// @ts-ignore` ?

loichuder (approved):
Thanks !

src/h5web/dimension-mapper/SlicingSlider.tsx
axelboc: That's because the types of ReactSlider don't cover this use case of getting a reference of the slider instance and calling methods on it.

src/h5web/dimension-mapper/SlicingSlider.tsx
axelboc: Hehe would have been awesome. I've just checked and ReactSlider expect the function to return `Element`.

PR #601, opened 2021-04-12
Make paths in metadata viewer clickable

### LINKED ISSUE #582: [MetadataViewer] Navigation for NeXus attributes
Make the values of any `default`, `signal`, `axes`, etc. attributes clickable in the `MetadataViewer` so users can quickly navigate to the entities.

axelboc:
Part of #582 

https://user-images.githubusercontent.com/2936402/114418536-e564bb00-9bb2-11eb-8fa2-dc1f1fc606cc.mp4

I haven't made attributes with array values clickable yet.

### DIFF
--- a/src/h5web/metadata-viewer/AttributesTable.tsx
+++ b/src/h5web/metadata-viewer/AttributesTable.tsx
@@ -1,12 +1,21 @@
+import { FiChevronsRight } from 'react-icons/fi';
 import type { Attribute } from '../providers/models';
 import styles from './MetadataViewer.module.css';
 
+const FOLLOWABLE_ATTRS = new Set([
+  'default',
+  'signal',
+  'axes',
+  'auxiliary_signals',
+]);
+
 interface Props {
   attributes: Attribute[];
+  onFollowPath: (path: string) => void;
 }
 
 function AttributesTable(props: Props) {
-  const { attributes } = props;
+  const { attributes, onFollowPath } = props;
 
   if (attributes.length === 0) {
     return null;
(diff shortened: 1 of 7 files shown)


src/h5web/metadata-viewer/AttributesTable.tsx
```diff
@@ -25,7 +34,21 @@ function AttributesTable(props: Props) {
         {attributes.map(({ name, value }) => (
           <tr key={name}>
             <th scope="row">{name}</th>
-            <td>{JSON.stringify(value)}</td>
+            <td>
+              {FOLLOWABLE_ATTRS.has(name) && typeof value === 'string' ? (
```
axelboc, line 38: I only deal with `string` attributes for now. Will deal with array of strings next.

loichuder: We might want to extract this in its own component then.

loichuder (approved):
Nice !

A simple question: would it be possible to dynamically check if an attribute is resolvable by checking in the `entityStore` ?

axelboc:
> A simple question: would it be possible to dynamically check if an attribute is resolvable by checking in the `entityStore` ?

Well, the `entityStore` doesn't know anything about the entities that haven't been visited yet, so it wouldn't be of much use. 

The `FOLLOWABLE_ATTRS` constant introduces a bit of an implicit coupling with NeXus, but at least it's simple and easy to understand, so I think it's reasonable for now.

loichuder:
> Well, the entityStore doesn't know anything about the entities that haven't been visited yet, so it wouldn't be of much use.

True, I forgot it only stores visited entities. I just wish we could notify the user if/when the path is not reachable.

axelboc:
I've fixed the regression and added a feature test.

PR #602, opened 2021-04-13
Improve design of Reflex splitter

axelboc:
It's been a while since I've wanted to make the splitter look a bit nicer and easier to grab.

image

On hover:

image

Before:

image


src/styles/vars.css
```diff
@@ -4,6 +4,7 @@
   --dark-slate-gray: #2f4f4f;
   --dark-gray: #696969;
   --gray: #808080;
+  --light-gray: #d6d6d6;
```
loichuder, line 7: :sob:

loichuder (approved):

PR #603, opened 2021-04-13
Upgrade to React v17 and R3F v6

axelboc:
Time for the React v17 and R3F v6 upgrade! There are still a bunch of peer-deps warnings, but much fewer than before.

No changes from React 17, of course, but R3F had a bunch of breaking changes, not all of which were documented...

Notable improvements:

- No need for `attach="material"` and `attach="geometry"`.
- `useThree` is now based on zustand, so we can pass it a selector to avoid unnecessary re-renders. When we need to get multiple properties from the state, I chose to write separate `useThree(state => state.prop)` calls instead of using zustand's `shallow` comparator. Having separate calls allows for destructing the objects we get, like `size`, `camera`, etc. right away.
- WebGL 2 is now the default.

Other notable upgrades:

- Three is still [making progress](https://github.com/mrdoob/three.js/pull/21437) with reducing its tree-shaken bundle size.
- `d3-array` [has a fix to `d3.ticks`](https://github.com/d3/d3-array/issues/199) that may benefit us somewhere.
- The ESLint config now warns us when we call a function with arguments of type `any`. I've had to fix a few of those.
- Storybook's `controls` add-on handles JSON arguments much better:

image

---

Planned work:

- `useUpdate` has been deprecated, so the code in `ErrorBars` is not as readable. I think there's a way to make this component more declarative by passing an array of floats directly to `<bufferGeometry args={[positions]} />` instead of calling `setFromPoints()` imperatively. Same in `DataCurve`.
- I've just thought of a way to extract the code of the tooltip formatters from `LineVis` and `HeatmapVis` with a hook called `useTooltipFormatters`. I might give that a try.


loichuder (approved):

src/h5web/vis-packs/core/line/ErrorBars.tsx
```diff
@@ -13,24 +13,25 @@ interface Props {
...
-  );
+  const barsGeometry = useRef<BufferGeometry>(null);
+  useLayoutEffect(() => {
+    barsGeometry.current?.setFromPoints(barsSegments);
+  }, [barsGeometry, barsSegments]);
+
+  const capsGeometry = useRef<BufferGeometry>(null);
+  useLayoutEffect(() => {
+    capsGeometry.current?.setFromPoints(capsPoints);
+  }, [capsGeometry, capsPoints]);
```
axelboc, line 24: `useUpdate` was removed. This is the recommended alternative. As mentioned in the PR description, I think there may be a way to refactor this.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -48,6 +48,7 @@ const Html = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
             left: 0,
             width,
             height,
+            pointerEvents: 'none', // let pointer events pass through to canvas
```
axelboc, line 51: Fix for the only regression caused by the React v17 upgrade. Now, `ReactDOM` registers its event handlers on the root element passed to `ReactDOM.render` instead of on the body. This means that our `Html` components now catch all the DOM events and stop their propagations. Since we want the `canvas` element to receive the events and yet still be positioned underneath the `Html` elements (so the SVG tooltip and grid remain visible), we have to make the `Html` elements "invisible" to DOM events with `pointer-events: none`.

loichuder (approved):
Not much to say: well done

PR #605, opened 2021-04-14
Followable attributes array members can be used for navigation

### LINKED ISSUE #582: [MetadataViewer] Navigation for NeXus attributes

loichuder:
Fix #582 

Not a huge fan of the resulting code, I may have missed a simpler solution. Perhaps I need my :coffee: :smile: 

**Before:**
image

**After:**
image

### DIFF
--- a/src/h5web/metadata-viewer/AttributeValueLink.tsx
+++ b/src/h5web/metadata-viewer/AttributeValueLink.tsx
@@ -1,0 +1,46 @@
+import { Fragment } from 'react';
+import { FiChevronsRight } from 'react-icons/fi';
+import styles from './MetadataViewer.module.css';
+
+interface Props {
+  onFollowPath: (value: string) => void;
+  value: unknown;
+}
+
+function AttributeValueLink(props: Props) {
+  const { onFollowPath, value } = props;
+
+  if (Array.isArray(value)) {
+    return (
+      <div className={styles.attrValueWrapper}>
+        <span>[</span>
+        {value.map((val, i) => (
+          // eslint-disable-next-line react/no-array-index-key
+          <Fragment key={i}>
+            <AttributeValueLink onFollowPath={onFollowPath} value={val} />
+            {i < value.length - 1 && <span>,</span>}
+          </Fragment>
+        ))}
+        <span>]</span>
+      </div>
+    );
+  }
+
+  if (typeof value === 'string' && value !== '.') {
+    return (
+      <button
+        className={styles.attrValueBtn}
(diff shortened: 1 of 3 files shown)


src/h5web/metadata-viewer/AttributeValueLink.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <AttributeValueLink onFollowPath={onFollowPath} value={val} />
+            {i < value.length - 1 && <span>,</span>}
+          </Fragment>
+        ))}
+        ]
+      </div>
+    );
+  }
+
+  if (typeof value === 'string' && value !== '.') {
```
loichuder, line 29: `.` is a placeholder and cannot be used for navigation so we might as well put it out of the way.

src/h5web/metadata-viewer/AttributesTable.tsx
```diff
@@ -35,16 +35,8 @@ function AttributesTable(props: Props) {
...
-                  className={styles.attrValueBtn}
-                  type="button"
-                  aria-label={`Inspect ${value}`}
-                  onClick={() => onFollowPath(value)}
-                >
-                  {value}
-                  <FiChevronsRight />
-                </button>
+              {FOLLOWABLE_ATTRS.has(name) ? (
+                <AttributeValueLink onFollowPath={onFollowPath} value={value} />
```
loichuder, line 39: Open to suggestions on the name

src/h5web/metadata-viewer/AttributeValueLink.tsx
```diff
@@ -0,0 +1,46 @@
...
+        aria-label={`Inspect ${value}`}
+        onClick={() => onFollowPath(value)}
+      >
+        {value}
+        <FiChevronsRight />
+      </button>
+    );
+  }
+
+  return <>{JSON.stringify(value)}</>;
```
axelboc, line 43: No need for the fragment in theory.

axelboc (approved):
Looks great to me. Can you just add spaces around the characters `[`, `]` and `,`?

src/h5web/metadata-viewer/AttributeValueLink.tsx
loichuder: I cannot remove it as I get `string | Element cannot be used as Element` when using `AttributeValueLink` recursively above.

PR #607, opened 2021-04-14
Added button to copy current path to clipboard

loichuder:
image

I used `navigator.clipboard.writeText`: https://caniuse.com/?search=clipboard

### DIFF
--- a/src/h5web/breadcrumbs/Breadcrumbs.tsx
+++ b/src/h5web/breadcrumbs/Breadcrumbs.tsx
@@ -48,10 +48,9 @@ function Breadcrumbs(props: Props) {
         type="button"
         title="Copy path to clipboard"
         onClick={() => navigator.clipboard.writeText(path)}
+        data-current
       >
-        <span className={styles.crumb} data-current>
-          {crumbs[crumbs.length - 1]}
-        </span>
+        <span className={styles.crumb}>{crumbs[crumbs.length - 1]}</span>
         <FiClipboard className={styles.copyIcon} />
       </button>
     </h1>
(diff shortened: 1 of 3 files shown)


src/h5web/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -42,9 +44,24 @@ function Breadcrumbs(props: Props) {
...
-      </span>
+      <button
+        className={styles.crumbButton}
+        type="button"
+        title="Copy path to clipboard"
+        onClick={() => {
+          navigator.clipboard.writeText(path);
+          setPathCopied(true);
+        }}
+        onMouseLeave={() => setPathCopied(false)}
```
axelboc, line 55: `onPointerOut` is more generic (for touch screens and such)

src/h5web/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -42,9 +44,24 @@ function Breadcrumbs(props: Props) {
...
+        data-current
+      >
+        <span className={styles.crumb}>{crumbs[crumbs.length - 1]}</span>
+        {isPathCopied ? (
+          <FiCheck className={styles.copyIcon} />
+        ) : (
+          <FiClipboard className={styles.copyIcon} />
+        )}
```
axelboc, line 63: You can store the icon in a constant like `const CopyIcon = isPathCopied ? FiCheck : FiClipoard` and then use it here `<CopyIcon className... />`

axelboc (approved):

PR #609, opened 2021-04-19
Matrix and Scalar vis support datasets of unsigned int types 

loichuder:
Fixes a regression introduced in #519 resolution


axelboc (approved):
Great catch 🎣

PR #610, opened 2021-04-19
Optimise LineVis when slicing

axelboc:
One big optimisation with `useMemo` and some minor ones.

### DIFF
--- a/src/h5web/vis-packs/core/line/DataCurve.tsx
+++ b/src/h5web/vis-packs/core/line/DataCurve.tsx
@@ -1,4 +1,4 @@
-import { Suspense, useMemo } from 'react';
+import { Suspense, useLayoutEffect, useState } from 'react';
 import { CurveType } from './models';
 import GlyphMaterial from './GlyphMaterial';
 import { extend, useThree } from '@react-three/fiber';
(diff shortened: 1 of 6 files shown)


src/h5web/vis-packs/core/line/DataCurve.tsx
```diff
@@ -38,22 +38,21 @@ function DataCurve(props: Props) {
...
   const points = useCanvasPoints(abscissas, ordinates, errors);
 
-  const dataGeometry = useMemo(() => {
-    const geometry = new BufferGeometry();
-    geometry.setFromPoints(points.data);
-    return geometry;
-  }, [points.data]);
+  useLayoutEffect(() => {
+    dataGeometry.setFromPoints(points.data);
+  }, [dataGeometry, points.data]);
```
axelboc, line 46: Minor optim: create `BufferGeometry` once and update it with `useLayoutEffect`, similarly to what is done in `ErrorBars` (except I don't use a `ref` here because we still need to create the geometry ourselves since it's shared between two meshes).

axelboc: I thought I could improve on `setFromPoints` by going lower level (by creating the `Float32Array`s and `BufferAttribute`s myself - cf. https://github.com/mrdoob/three.js/blob/master/src/core/BufferGeometry.js#L263), but I did a lot of benchmarking and couldn't justify the change, so I reverted.

src/h5web/vis-packs/core/line/LineVis.tsx
```diff
@@ -70,7 +70,10 @@ function LineVis(props: Props) {
     );
   }
 
-  const abscissaToIndex = getValueToIndexScale(abscissas, true);
+  const [abscissas, abscissaToIndex] = useMemo(() => {
+    const abscissas = abscissaValue || range(dataArray.size);
+    return [abscissas, getValueToIndexScale(abscissas, true)];
+  }, [abscissaValue, dataArray.size]);
```
axelboc, line 76: This constitutes the big optimisation: 

- When no custom abscissa values were passed, we were recreating the `abscissas` range on every render (for instance, when slicing). More importantly, the change to `abscissas` was invalidating the `useMemo`s further down, so the domain was being recomputed on every render as well.
- `getValueToIndexScale`, which is actually quite computationally intensive (it creates a `range` and loops through it), was not wrapped in `useMemo`, so it was also called on ever render.

src/h5web/vis-packs/core/line/hooks.ts
```diff
@@ -1,14 +1,14 @@
 import { useMemo } from 'react';
-import { useThree } from '@react-three/fiber';
 import { Vector3 } from 'three';
 import { useCanvasScales } from '../hooks';
 
+const CAMERA_FAR = 1000; // R3F's default
```
axelboc, line 5: It might make a tiny bit of an improvement to hard-code this value... Either way, it simplifies the code.

axelboc:
Sorry, I have a cognitive complexity warning. Will fix and let you know when it's done.

axelboc:
Hmm, actually, I'll silence it, as I'm planning a refactoring of the tooltip formatters that will extract them into a hook and most likely fix this warning.

loichuder (approved):

PR #611, opened 2021-04-19
Create tooltip formatters in custom hooks and fix perf issues on Heatmap

axelboc:
As mentioned, I move the tooltip formatters into custom hooks. I think it's a lot cleaner, since it also moves the creation of the scales that are specifically used for the tooltips.

I also fix performance bottlenecks in the Heatmap: `getPixelEdges` and `getValueToIndexScale` weren't wrapped in `useMemo`, so they were probably slowing down slicing through big image stacks significantly (since they both loop through the abscissas/ordinates arrays, which don't change when slicing). Do we have a large image stack somewhere? The one at `/entry_0000/0_measurement` in the `water` file that I usually use is unavailable because of compression. 😢

### DIFF
--- a/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
+++ b/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
@@ -1,5 +1,3 @@
-import { useMemo } from 'react';
-import { format } from 'd3-format';
 import type ndarray from 'ndarray';
 import styles from './HeatmapVis.module.css';
 import ColorBar from './ColorBar';
@@ -7,11 +5,13 @@ import TooltipMesh from '../shared/TooltipMesh';
 import PanZoomMesh from '../shared/PanZoomMesh';
 import Mesh from './Mesh';
 import VisCanvas from '../shared/VisCanvas';
-import { getDims, getPixelEdges } from './utils';
+import { getDims } from './utils';
 import { Domain, ScaleType, AxisParams } from '../models';
 import type { ColorMap } from './models';
-import { DEFAULT_DOMAIN, getDomain, getValueToIndexScale } from '../utils';
+import { DEFAULT_DOMAIN } from '../utils';
 import { assertDefined } from '../../../guards';
+import { useAxisValues, useTooltipFormatters } from './hooks';
+import { useDomain } from '../hooks';
 
 interface Props {
   dataArray: ndarray;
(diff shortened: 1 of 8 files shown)


src/h5web/vis-packs/core/heatmap/hooks.ts
```diff
@@ -1,5 +1,38 @@
...
+    formatIndex: ([x, y]) => {
+      return `${abscissaLabel || 'x'}=${abscissas[abscissaToIndex(x)]}, ${
+        ordinateLabel || 'y'
+      }=${ordinates[ordinateToIndex(y)]}`;
+    },
+
+    formatValue: ([x, y]) => {
+      return format('.3')(
+        dataArray.get(ordinateToIndex(y), abscissaToIndex(x))
+      );
```
loichuder, line 35: `return` could be omitted in both functions

src/h5web/vis-packs/core/line/LineVis.tsx
```diff
@@ -87,9 +80,17 @@ function LineVis(props: Props) {
...
+    errorsArray
+  );
+
   return (
     <figure className={styles.root} aria-labelledby="vis-title">
       <VisCanvas
+        canvasTitle={title}
         abscissaConfig={{
           domain: abscissaDomain,
           showGrid,
```
loichuder, line 96: Not modified in this PR but the ` isIndexAxis: !abscissaParams.value` below can be replaced by `isIndexAxis: !abscissaValue`

src/h5web/vis-packs/core/line/LineVis.tsx
```diff
@@ -71,9 +65,8 @@ function LineVis(props: Props) {
     );
   }
 
-  const [abscissas, abscissaToIndex] = useMemo(() => {
-    const abscissas = abscissaValue || range(dataArray.size);
-    return [abscissas, getValueToIndexScale(abscissas, true)];
+  const abscissas = useMemo(() => {
+    return abscissaValue || range(dataArray.size);
```
loichuder, line 69: `return` could be omitted to have a one-liner perhaps ?

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -40,50 +40,47 @@ function HeatmapVis(props: Props) {
...
+  const { label: abscissaLabel, value: abscissaValue } = abscissaParams;
+  const { label: ordinateLabel, value: ordinateValue } = ordinateParams;
+
   const { rows, cols } = getDims(dataArray);
   const aspectRatio = keepAspectRatio ? cols / rows : undefined; // width / height <=> cols / rows
 
-  const abscissas = getPixelEdges(abscissaParams.value, cols);
-  const abscissaToIndex = getValueToIndexScale(abscissas);
-  const abscissaDomain = useMemo(() => getDomain(abscissas), [abscissas]);
+  const abscissas = usePixelEdges(abscissaValue, cols);
```
loichuder, line 49: Looking back at this, I find `getPixelEdges` not clear enough... 

What about you ? Do you think `getPixelEdgesCoordinates` would be better ?

loichuder (approved):
Good call :+1:

src/h5web/vis-packs/core/line/LineVis.tsx
axelboc: Unfortunately no, the dependency array gets split onto a couple of lines 😢

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
axelboc: Actually, how about `useIndices`? I find the term "coordinates" inaccurate.

axelboc: Pushed. Let me know what you think.

loichuder: Well if `abscissaValue` is defined, these are not indices...

PR #612, opened 2021-04-19
Add same-length assertion function for arrays

axelboc:
Just to clean up the `LineVis` component, and `getDomain` a bit.


loichuder (approved):
:sparkles:

PR #617, opened 2021-04-20
Set up devApi for JupyterProvider

loichuder:


src/h5web/providers/jupyter/devApi.ts
```diff
@@ -0,0 +1,10 @@
+import { JupyterApi } from './api';
+
+export class JupyterDevApi extends JupyterApi {
+  /* API compatible with jupyterlab_hdf v0.5.1 */
+  public constructor(url: string, filepath: string) {
+    super(url, filepath);
+    // eslint-disable-next-line no-console
+    console.log('Using Jupyter dev API');
```
loichuder, line 8: I thought this could be handy for debugging but I don't mind removing it

.env
```diff
@@ -17,6 +17,8 @@ REACT_APP_HSDS_FALLBACK_FILEPATH="/home/reader/water"
 # JupyterLab backend parameters
 REACT_APP_JLAB_URL=
 REACT_APP_JLAB_FALLBACK_FILEPATH="water_224.h5"
+# Use dev version of the API
+REACT_APP_JLAB_DEV=
```
axelboc, line 21: Since it's a boolean, I would name it `REACT_APP_JLAB_DEV_ENABLED`

src/h5web/providers/jupyter/JupyterProvider.tsx
```diff
@@ -10,7 +11,13 @@ interface Props {
...
 function JupyterProvider(props: Props) {
   const { url, filepath, children } = props;
-  const api = useMemo(() => new JupyterApi(url, filepath), [filepath, url]);
+  const api = useMemo(
+    () =>
+      process.env.REACT_APP_JLAB_DEV === 'true'
+        ? new JupyterDevApi(url, filepath)
+        : new JupyterApi(url, filepath),
+    [filepath, url]
+  );
```
axelboc, line 20: You could store the right constructor in a constant outside of the component completely, since it's a build-time thing. Like `const JupyterApi = process.env.REACT_APP_JLAB_DEV === 'true' ? JupyterDevApi : JupyterStableApi`

src/h5web/providers/jupyter/devApi.ts
axelboc: Good call, it will save us a lot of time when debugging.

axelboc (approved):

src/h5web/providers/jupyter/devApi.ts
loichuder: I replaced `log` with `warn` to make it stand-out among profiling data

axelboc: Hmm, I would have suggested `debug` instead -- it doesn't stand out, but you can filter it in the dev tools.

axelboc: hmm no you're right -- if we're using the dev provider in a prod environment, we should see it clearly

PR #619, opened 2021-04-21
Updated metadata endpoints for jupyter dev api

loichuder:
This means #406 will be solved once this is part of the stable API.

### DIFF
--- a/src/h5web/providers/jupyter/api.ts
+++ b/src/h5web/providers/jupyter/api.ts
@@ -10,9 +10,12 @@ import {
 } from './utils';
 import type {
   JupyterAttrsResponse,
+  JupyterBaseEntity,
   JupyterComplex,
+  JupyterContentGroupResponse,
   JupyterContentResponse,
   JupyterDataResponse,
+  JupyterMetaGroupResponse,
   JupyterMetaResponse,
 } from './models';
 import { makeStrAttr } from '../mock/metadata-utils';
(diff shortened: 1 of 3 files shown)


src/h5web/providers/jupyter/models.ts
```diff
@@ -1,7 +1,6 @@
-import type { EntityKind } from '../models';
+import type { Attribute, EntityKind } from '../models';
 
 export interface JupyterMetaResponse {
-  id: string;
```
loichuder, line 4: `id` was removed in the dev API. As we do not make use of it, we might as well remove it from the stable API interface.

src/h5web/providers/jupyter/devApi.ts
```diff
@@ -1,10 +1,88 @@
...
+}
+
+interface DevJupyterMetaResponse extends JupyterMetaResponse {
+  attributes: DevJupyterAttrMeta[];
+}
+
+interface DevJupyterMetaGroupResponse extends JupyterMetaGroupResponse {
+  attributes: DevJupyterAttrMeta[];
+  childrenCount: number;
+}
```
loichuder, line 24: I introduced a few additional types. I thought that keeping them in the same file would ease the updating of the stable API.

src/h5web/providers/jupyter/api.ts
```diff
@@ -138,4 +127,33 @@ export class JupyterStableApi implements ProviderAPI {
...
+  protected async processContents(
+    path: string,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    response: JupyterMetaGroupResponse
+  ): Promise<JupyterContentGroupResponse> {
+    const contents = await this.fetchContents(path);
+    assertGroupContent(contents);
+
+    return contents;
+  }
```
loichuder, line 158: I encapsulated the varying process parts into these functions that are overriden in the dev API.

Once these are no longer needed in the dev API, we can inline them once again.

axelboc (approved):
Awesome! 🍻 I've got nothing to pick on.

PR #621, opened 2021-04-21
Refactor `useCombinedDomain` and add to lib package

axelboc:
Also:

- add `getDomains`/`useDomains` to compute multiple domains, 
- add story for auxiliary arrays,
- improve documentation of utilities:

image

### DIFF
--- a/src/stories/Home.stories.mdx
+++ b/src/stories/Home.stories.mdx
@@ -101,6 +101,10 @@ If you use only the top-level visualization components, the only utilities you n
 
 - **`getDomain(values: ndarray | number[], scaleType: ScaleType = ScaleType.Linear): Domain | undefined`** - Find the min and max values contained in a array (or ndarray) of numbers. If `scaleType` is `ScaleType.Log` and the domain crosses zero, clamp the min to the first positive value or return `undefined` if the domain is not supported (i.e. `[-x, 0]`).
 - **`useDomain(...args): Domain | undefined`** - Memoised React hook wrapper of `getDomain`.
+- **`getDomains(arrays: (ndarray | number[])[], scaleType: ScaleType = ScaleType.Linear): (Domain | undefined)[]`** - Find the domains of multiple arrays (or ndarray) of numbers. Useful when dealing with auxiliary curves.
+- **`useDomains(...args): (Domain | undefined)[]`** - Memoised React hook wrapper of `getDomains`.
+- **`getCombinedDomain(domains: (Domain | undefined)[]): Domain | undefined`** - Combine multiple domains into one. Useful to find the overall domain of a line visualization with auxiliary curves.
+- **`useCombinedDomain(...args): Domain | undefined`** - Memoised React hook wrapper of `getCombinedDomain`.
 - **`extendDomain(bareDomain: Domain, extendFactor: number, isLog?: boolean): Domain`** - Extend a domain by a given factor.
 - **`computeVisSize(availableSize: Size, aspectRatio?: number): Size | undefined`** - Compute the optimal size for a visualization to fill the available space given an optional aspect ratio.
 - **`getDims(dataArray: DataArray): Dims`** - Get the dimensions of a `ndarray` as a `{ rows, cols }` object.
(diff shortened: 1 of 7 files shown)


src/stories/Home.stories.mdx
```diff
@@ -99,9 +99,85 @@ const dataArray = getMockDataArray('/nD_datasets/twoD');
...
+
+```ts
+getCombinedDomain(domains: (Domain | undefined)[]): Domain | undefined
+
+const combinedDomain = getCombinedDomain([[-1, 10], [0, 30]]]); // [-1, 30]
+```
+
+#### extendDomain
+
+Extend a domain by a given factor.
```
loichuder, line 139:
```suggestion
Extend a domain by a given factor. The factor is scaled according to the given scale.
```

src/stories/Home.stories.mdx
```diff
@@ -99,9 +99,85 @@ const dataArray = getMockDataArray('/nD_datasets/twoD');
...
+
+```ts
+extendDomain(bareDomain: Domain, extendFactor: number, isLog?: boolean): Domain
```
loichuder, line 142: This declaration is not right.

src/stories/Home.stories.mdx
```diff
@@ -99,9 +99,85 @@ const dataArray = getMockDataArray('/nD_datasets/twoD');
...
+
+const extendedDomain1 = extendDomain([0, 100], 0.5]); // [-50, 150]
+const extendedDomain2 = extendDomain([10, 100], 1, ScaleType.Log); // approx. [1, 1000]
+const extendedDomain3 = extendDomain([2, 2], 0.5); // [1, 3]
+const extendedDomain4 = extendDomain([1, 1], 1, ScaleType.Log); // [0.1, 10]
+```
+
+#### computeVisSize
+
+Compute the optimal size for a visualization to fill the available space given an optional aspect ratio.
```
loichuder, line 152:
```suggestion
Compute the optimal size for a visualization to fill the available space given a aspect ratio. If this aspect ratio is omitted, it will return the available size.
```

src/stories/Home.stories.mdx
```diff
@@ -99,9 +99,85 @@ const dataArray = getMockDataArray('/nD_datasets/twoD');
...
+
+```ts
+computeVisSize(availableSize: Size, aspectRatio?: number): Size | undefined
+
+const size = computeVisSize({ width: 20, height: 10 }, 1.5); // { width: 15, height: 10 }
+```
+
+#### getDims
+
+Get the dimensions of a `ndarray` as a `{ rows, cols }` object.
```
loichuder, line 162: Remind me why are we exporting this ? :joy:

src/stories/Home.stories.mdx
```diff
@@ -99,9 +99,85 @@ const dataArray = getMockDataArray('/nD_datasets/twoD');
...
+const linearDomain = getDomain([10, 5, -1]); // [-1, 10]
+const logDomain = getDomain([-1, 2, 10], ScaleType.Log); // [2, 10]
+
+const myArray = ndarray([0, 1, 2, 3], [2, 2]);
+const myDomain = getDomain(myArray); // [0, 3]
+```
+
+#### getDomains
+
+Find the domains of multiple arrays (or ndarray) of numbers. Useful when dealing with auxiliary curves.
```
loichuder, line 118:
```suggestion
Find the domains of multiple arrays (or ndarrays) of numbers. Useful when dealing with auxiliary curves.
```

loichuder (approved):
Not a huge fan of `getDomains` but I understand the rationale behind it.

Actually, as we always follow its use by `getCombinedDomain`, would it not make sense to _combine_ `getDomains` and `getCombinedDomain` in one function ? That would return the domain of an array of arrays ?

axelboc:
> Actually, as we always follow its use by getCombinedDomain, would it not make sense to combine getDomains and getCombinedDomain in one function ? That would return the domain of an array of arrays ?

The tricky bit is that we compute the data domain separately. Sure, we could pass the data domain to the `getCombinedDomain` you suggest as an array like any other (since `getDomain(getDomain(dataArray))` would return the same value as `getDomain(dataArray)`), but it just felt wrong.

I think it's clearer for `getCombinedDomain` to have the single responsibility of merging domains, nothing else.

src/stories/Home.stories.mdx
axelboc: Hmm, I see no good reason 😂 - removing it now.

PR #622, opened 2021-04-22
Complex are treated as two-values arrays by providers, Matrix and ScalarVis

loichuder:
Part of #310

As reminder, complex values are already sent as two-values array by HSDS and JupyterDev providers but are sent as strings `a+bj` for JupyterStable.

Parsing such strings without `complex.js` would be a pain and is not worth the trouble as it is only a temp fix waiting for `jupyterlab_hdf` next release. Therefore, I kept `complex.js` as dep for now.


src/h5web/providers/models.ts
```diff
@@ -158,4 +156,5 @@ export type Value<D extends Dataset> = D['shape'] extends ScalarShape
   ? PrimitiveType<D['type']>[]
   : never;
 
-export type ComplexArray = (ComplexArray | Complex)[];
+export type H5WebComplex = [number, number];
+export type ComplexArray = (ComplexArray | H5WebComplex)[];
```
loichuder, line 160: The new `Complex` type is called `H5WebComplex` temporarily. Once `complex.js` is removed, I intend to rename it `Complex`

src/h5web/vis-packs/core/matrix/GridSettingsContext.tsx
```diff
@@ -1,15 +1,12 @@
...
 
 interface GridSettings {
   cellSize: Size;
   rowCount: number;
   columnCount: number;
-  valueAccessor: (
-    row: number,
-    col: number
-  ) => number | string | boolean | Complex;
+  valueAccessor: (row: number, col: number) => PrimitiveType<PrintableType>;
```
loichuder, line 9: While using PrimitiveType<PrintableType> is handy, this is not awfully readable. Glad to have your input on this.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -304,7 +304,7 @@ export function getBaseArray<T>(
   value: T[] | undefined,
   rawDims: number[]
 ): ndarray<T> | undefined {
-  return value && ndarray<T>(value.flat(Infinity) as T[], rawDims);
+  return value && ndarray<T>(value.flat(rawDims.length - 1) as T[], rawDims);
```
loichuder, line 307: This ensures that complex data does not get flattened all the way: complex arrays are `ndarrays` that contain two-value arrays as value.

It does not change anything for other types of arrays.

src/h5web/vis-packs/core/matrix/GridSettingsContext.tsx
axelboc: Good idea on using `PrintableType` 👍

It's true that the `Type` suffix is more for our adaptation of HDF5 types. Perhaps just rename it to `Primitive`?

src/h5web/vis-packs/core/utils.ts
axelboc: Can you please explain why this doesn't change anything for other types of arrays?

loichuder: Imagine a 2D array of integers:
```
[
  [0, 2],
  [1, 9]
]
```

Converting this array to 1D require a `.flat(1)` i.e. `.flat(nDims - 1)`. Yielding the same as `.flat(Infinity)`.

axelboc: Ah right because `rawDims` comes from the HDF5 type. Makes sense 👍

axelboc (approved):

src/h5web/vis-packs/core/utils.ts
axelboc: Maybe just add a comment to that regard somewhere?

PR #624, opened 2021-04-22
Fix scroll capture

### LINKED ISSUE #520: Zooming on heatmap should not scroll page

axelboc:
Fix #520

### DIFF
--- a/src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
+++ b/src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
@@ -2,7 +2,6 @@ import { useRef, useCallback, useEffect } from 'react';
 import type { Vector3 } from 'three';
 import { useThree } from '@react-three/fiber';
 import { clamp } from 'lodash-es';
-import Html from './Html';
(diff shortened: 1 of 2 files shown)


src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
```diff
@@ -94,16 +93,13 @@ function PanZoomMesh() {
...
-  const wheelCaptureRef = useWheelCapture();
+  useWheelCapture();
 
   return (
-    <>
-      <mesh {...{ onPointerMove, onPointerUp, onPointerDown, onWheel }}>
-        <meshBasicMaterial opacity={0} transparent />
-        <planeGeometry args={[width, height]} />
-      </mesh>
-      <Html ref={wheelCaptureRef} />
```
axelboc, line 105: It stopped working because of the `pointer-events: none` set on the root DOM element rendered by the `Html` compoent. Now, I hook straight onto the canvas' `domElement`, so we no longer need an `Html` element here!

loichuder (approved):
Great fix :slightly_smiling_face:

PR #625, opened 2021-04-22
Tweak styling of color bar axis and bounds + other minor tweaks

### LINKED ISSUE #623: [Storybook] Discussion after quick review
I had a quick look at the story book and would like to discuss the following points:

- [x] `[HeatmapVis]` Is the _Live Data_ story still relevant ?
- [x] `[HeatmapVis]` On _Scroll Capture_, zooming on the heatmap still make the page scroll (FF 87). Should we reopen #520 ?
- [x] `[ScalarVis]` Not the most interesting I agree, but should we add boolean and complex displays to the stories ?
- [x] `[ColorBar]` On _With Bounds_, the font size of bounds is much bigger than the one of the ticks. Perhaps we should change the styles so that the two font sizes are consistent ?

axelboc:
Fix #623 

- Remove `Live Data` HeatmapVis story.
- Add ScalarVis stories for boolean and complex values.
- Increase space available for abscissa axis label and space it out from the tick values.
- Increase font-size of color bar tick values.
- Tweak alignment and spacing of tick values with tick lines.
- Remove extraneous `svg` element around color bar tick values.
- Use same color as axis values/labels for color bar bounds.

image


src/h5web/vis-packs/core/shared/AxisSystem.module.css
```diff
@@ -22,37 +23,24 @@
 
 .axis[data-type='abscissa'] {
   grid-area: bottom-axis;
-  text-anchor: middle;
 }
 
 .axis[data-type='ordinate'] {
-  grid-area: axis-grid; /* `AxisLeft` aligns ticks to the outer left of its own bounding box  */
```
axelboc, line 29: The comment was on the wrong line, but the `text-anchor` fix is no longer needed because `AxisLeft` now seems to set the correct `text-anchor` by default.

loichuder (approved):

PR #627, opened 2021-04-22
Fixed display of complex data for attributes, Matrix and ScalarVis

loichuder:
**Attribute:**
image

**Matrix:**
image

**ScalarVis:**
image

### DIFF
--- a/src/h5web/metadata-viewer/AttributesTable.tsx
+++ b/src/h5web/metadata-viewer/AttributesTable.tsx
@@ -1,6 +1,12 @@
-import type { Attribute } from '../providers/models';
+import {
+  Attribute,
+  ComplexArray,
+  DTypeClass,
+  H5WebComplex,
+} from '../providers/models';
 import AttributeValueLink from './AttributeValueLink';
 import styles from './MetadataViewer.module.css';
+import { formatComplex } from './utils';
 
 const FOLLOWABLE_ATTRS = new Set([
   'default',
@@ -31,12 +37,14 @@ function AttributesTable(props: Props) {
         </tr>
       </thead>
       <tbody>
-        {attributes.map(({ name, value }) => (
+        {attributes.map(({ name, type, value }) => (
           <tr key={name}>
             <th scope="row">{name}</th>
             <td>
               {FOLLOWABLE_ATTRS.has(name) ? (
                 <AttributeValueLink onFollowPath={onFollowPath} value={value} />
+              ) : type.class === DTypeClass.Complex ? (
+                formatComplex(value as H5WebComplex | ComplexArray)
               ) : (
                 JSON.stringify(value)
               )}
--- a/src/h5web/utils.ts
+++ b/src/h5web/utils.ts
@@ -1,8 +1,17 @@
 import { format } from 'd3-format';
-import type { Entity, Group } from './providers/models';
+import type { Entity, Group, H5WebComplex } from './providers/models';
 
 export const formatValue = format('.3~e');
 export const formatPreciseValue = format('.5~e');
+export function formatComplexValue(value: H5WebComplex, specifier = '') {
+  const [real, imag] = value;
+  if (real === 0) {
+    return `${format(specifier)(imag)}i`;
+  }
+
+  const formatFn = format(specifier);
+  return `${formatFn(real)}${imag >= 0 ? '+' : ''}${formatFn(imag)}i`;
+}
 
 export function getChildEntity(
   group: Group,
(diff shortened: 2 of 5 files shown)


src/h5web/metadata-viewer/AttributesTable.tsx
```diff
@@ -31,14 +37,18 @@ function AttributesTable(props: Props) {
...
             <th scope="row">{name}</th>
             <td>
               {FOLLOWABLE_ATTRS.has(name) ? (
                 <AttributeValueLink onFollowPath={onFollowPath} value={value} />
               ) : (
-                JSON.stringify(value)
+                JSON.stringify(
+                  type.class === DTypeClass.Complex
+                    ? formatComplex(value as H5WebComplex | ComplexArray)
+                    : value
```
loichuder, line 50: We can surely avoid the cast here using a type guard. My issue is: what args does this type guard take ? 

I can't give the whole `attribute` as TS does not allow me to infer the type of an object field. But it would seem awkward to give `type` and `value`, no ?

src/h5web/vis-packs/core/matrix/Cell.tsx
```diff
@@ -3,6 +3,21 @@ import type { GridChildComponentProps } from 'react-window';
...
+  if (Array.isArray(value)) {
+    return formatComplexValue(value, '.1f');
+  }
+
+  if (typeof value === 'number') {
+    return format('.3e')(value);
+  }
+
+  return value.toString();
+}
```
loichuder, line 20: There is no `utils.ts` for `matrix`. So either this or nested ternaries...

src/h5web/vis-packs/core/scalar/ScalarVis.tsx
```diff
@@ -8,7 +9,12 @@ interface Props {
 
 function ScalarVis(props: Props) {
   const { value } = props;
-  return <div className={styles.scalar}>{value.toString()}</div>;
+
+  return (
+    <div className={styles.scalar}>
+      {Array.isArray(value) ? formatComplexValue(value) : value.toString()}
```
loichuder, line 15: The check leaves a lot to be desired in my opinion. It is sufficient as `value` is `string | number | boolean | [number, number]` but could we do better ?

src/h5web/metadata-viewer/AttributesTable.tsx
axelboc: Instinctively, it looks like the same problem as with datasets. We'd first want a guard that checks whether the attribute has a complex type and a scalar or array shape (`Attribute<Shape, ComplexType>`), then we'd want `formatComplex` to accept the entire attribute object so that the signature of the utility function can make use of `Value<Dataset | Attribute>`.

I wonder whether we could make all our existing guards smarter so they could deal with both datasets and attributes (with an extra generic).

src/h5web/utils.ts
```diff
@@ -1,8 +1,13 @@
 import { format } from 'd3-format';
-import type { Entity, Group } from './providers/models';
+import type { Entity, Group, H5WebComplex } from './providers/models';
 
 export const formatValue = format('.3~e');
 export const formatPreciseValue = format('.5~e');
+export function formatComplexValue(value: H5WebComplex, specifier = '') {
+  const [real, imag] = value;
+  const formatFn = format(specifier);
+  return `${formatFn(real)}${imag >= 0 ? '+' : ''}${formatFn(imag)}i`;
```
axelboc, line 13: The operator in the middle needs spaces on either side. One option would be add the minus sign in the ternary (with spaces around) and take the absolute value of `imag` in the third block.

Also I would use `Math.sign` in the ternary. What happens if `imag` is equal to 0? I'd argue that `0i` should _not_ be rendered, but maybe that's not what other libraries do 🤷.

src/h5web/vis-packs/core/matrix/Cell.tsx
axelboc: I reckon add the `utils.ts` 😉

axelboc: Just a thought for the future. Maybe the formatter should be passed by `MappedMatrixVis`. This way, we could pass the correct formatter for the dataset based on its type without having to check the value. `Array.isArray(value)` may be problematic in the future if other data types are stored as arrays, for instance.

src/h5web/vis-packs/core/scalar/ScalarVis.tsx
axelboc: Same remark as with `MatrixVis`. We could have a generic on `ScalarVis` and `MatrixVis`:

```ts
interface Props<T> {
  value: T;
  formatter: (value: T) => string;
}
```

... and a utility function that takes a dataset (or attribute, so it can also be used in the metadata viewer?) and returns a formatter with a conditional type similar to `Value`:

```ts
function getFormatter<T extends Dataset | Attribute>(datasetOrAttr: T): Formatter<T> { ... }
```

loichuder:
Thanks for the nice ideas ! 

They make a lot of sense to me so I will close this, do the needed refactorings and have another go when the environment is more hospitable :slightly_smiling_face:

PR #628, opened 2021-04-23
Change "Nothing to Visualize" message

### LINKED ISSUE #620: Reword displayed message `Nothing to visualize` when no vis is available
The message `Nothing to visualize` could imply that there is **absolutely nothing** to visualize in the whole file. Especially as it is displayed when landing on the root group.

Ideas:
1. `This entity cannot be visualized.`
2. `No visualization available for this entity.`
...

t20100:
closes #620


axelboc (approved):
The term `item` bothers me a little. What about `object`, which seems to be the term used by HDF5?

Issue #618, opened 2021-04-21
[NeXus] Error message when default is not found changes with the backend


loichuder:
When the `default` attribute is not found (see group `default_not_found`), we get different error messages depending on the backend:
- [Mock](https://h5web.panosc.eu/mock): **Expected entity at path "/test"**
- [Jupyter](https://h5web.panosc.eu/?file=nexus_malformed.h5): **Request failed with status code 500**
- [HSDS](https://h5web.panosc.eu/hsds/?file=/home/reader/nexus_malformed) : **Expected some value**

Perhaps we could have a more appropriate and unified error message for this particular case.

For the record, the other error messages (`default_not_string`, `no_signal` and `signal_not_found`) are consistent and clear.

axelboc:
Not a fan of the second approach after all. Will try a catch-and-rethrow approach instead.

axelboc:
This issue concerns also the error message displayed when navigating to an entity that doesn't exist, notably when following a NeXus attribute like `axis` or `signal` in the metadata viewer.

The fix requires back-end providers to respond with a 404 to the request made from the `getEntity` API method. This is currently the case with HSDS, but not with H5Grove and jupyerlab-hdf5:

- [ ] https://github.com/jupyterlab/jupyterlab-hdf5/issues/96
- [ ] https://github.com/silx-kit/h5grove/issues/24

axelboc:
Now fixed in all providers! :tada: :trophy:

PR #630, opened 2021-04-26
Improve error handling when entity is not found

axelboc:
Fix partially #618 for mock and HSDS providers.

I throw a specific error message, `ProviderError.NotFound`, from the mock and HSDS providers, which I catch from the calls to `entitiesStore.get(path)`. I then rethrow a more specific error that can be caught by the error boundary and displayed to the user.

image

Three error cases are now properly handled with the mock and HSDS providers:

1. In "Display", when the default path is not found (`getDefaultEntity` in `NexusPack`).
2. In "Inspect", when the user follows a default path that is not found (`MetadataViewer`).
3. In "Display", when the user switches back from the `MetadataViewer` after following a path that was not found (`VisPackChooser`).

### DIFF
--- a/src/h5web/metadata-viewer/MetadataViewer.tsx
+++ b/src/h5web/metadata-viewer/MetadataViewer.tsx
@@ -1,7 +1,8 @@
 import { memo, useContext } from 'react';
 import { isAbsolutePath } from '../guards';
 import { ProviderContext } from '../providers/context';
-import { buildEntityPath } from '../utils';
+import { ProviderError } from '../providers/models';
+import { buildEntityPath, handleError } from '../utils';
 import AttributesTable from './AttributesTable';
 import EntityTable from './EntityTable';
 import styles from './MetadataViewer.module.css';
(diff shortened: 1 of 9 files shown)


src/h5web/metadata-viewer/MetadataViewer.tsx
```diff
@@ -13,9 +14,13 @@ interface Props {
...
   const { path, onSelectPath } = props;
-
   const { entitiesStore } = useContext(ProviderContext);
-  const entity = entitiesStore.get(path);
+
+  const entity = handleError(
+    () => entitiesStore.get(path),
+    ProviderError.NotFound,
+    `No entity found at ${path}`
+  );
```
axelboc, line 23: I'm open to naming suggestions for `handleError` if you're not a fan.

Inlining the `try`/`catch` feels very repetitive and complex, as it requires the use of `let entity`:

```ts
let entity;
try {
  entity = entitiesStore.get(path);
} catch (...) {
  ...
}

// do something with `entity`
```

loichuder (approved):
:clap:

PR #632, opened 2021-04-28
Fixed flattening of slices of complex data in MockProvider

loichuder:
Same issue as https://github.com/silx-kit/h5web/pull/622#discussion_r618125585

Slices of complex data `[real, imag]` were flattened all the way down so that the dataset/slice shape was not conserved.

Ex: trying to fetch a slice of 2 complex elements containing `[[0, 1], [3, 2]]` was yielding `[0, 1, 3, 2]`


axelboc (approved):
May need to be fixed in `getMockDataArray` then?

loichuder:
> May need to be fixed in `getMockDataArray` then?

Currently no as `getMockDataArray` only supports `numericType` arrays. If we want it to support `complexType` as well, then yes.

Do you want to me to implement this in the same PR ?

axelboc:
Ah yeah, no it's fine.

PR #634, opened 2021-04-28
Added phase map visualisation for complex data (with amplitude as transparency)

loichuder:
A first visu for complex data (#310). The _Phase Map_ uses HeatmapVis to display the phase in color and the amplitude in transparency.

image

I can think of a dozen improvements as follow-ups but I would like first to integrate and discuss the current implementation.

In summary, this adds:
- A new `VisContainer` that supports nD complex array with n >=2 (note that the slicing with `DimensionMapper` work as expected).
- A new `MappedPhaseMapVis` that fetches slices of complex data. It will then transform these slices of `[real, imag]` into two `ndarray`: one containing the phase, the other the amplitude. As this requires to go through the whole slice, the domains are computed in the same loop for efficiency.
- Additional props to `HeatmapVis` and `Mesh` to handle transparency values. We may need to write specific components in the future but for this, these minimal changes were enough in my opinion.

The reuse of components make auxiliary operations such as changing the domain, the scale work as expected (albeit only on the phase).


src/h5web/providers/mock/values.ts
```diff
@@ -65,42 +65,42 @@ export const mockValues = {
...
       ],
       [
-        [9.5, 1.4],
-        [5, 0],
-        [0, -2],
-        [8, 2.1],
+        [-8, 0],
+        [6.93, -4],
+        [7.39, -3.06],
+        [8, 0],
```
loichuder, line 103: I changed the data so that I can have slices of same phases but different amplitudes

src/h5web/vis-packs/core/models.ts
```diff
@@ -63,7 +63,11 @@ export interface AxisParams {
 
 export type AxisMapping = (AxisParams | undefined)[];
 
-export type Bounds = [number, number, number];
+export interface Bounds {
+  min: number;
+  max: number;
+  positiveMin: number;
+}
```
loichuder, line 70: I changed `Bounds` into an object so that it is clear which number is the `min`, the `max` and the `positiveMin`. 

I made my visualisation crash a couple of times as I misplaced the bounds :smile:

src/h5web/vis-packs/core/complex/MappedPhaseMapVis.tsx
```diff
@@ -0,0 +1,99 @@
...
+      setScaleType(colorScaleType);
+    }
+  }, [setScaleType, colorScaleType]);
+
+  return (
+    <HeatmapVis
+      dataArray={phaseArray}
+      domain={safePhaseDomain}
+      transparencyArray={amplitudeArray}
+      transparencyDomain={amplitudeDomain}
```
axelboc, line 86: `alphaArray` and `alphaDomain` perhaps? It's a bit shorter.

src/h5web/vis-packs/core/complex/MappedPhaseMapVis.tsx
```diff
@@ -0,0 +1,99 @@
...
+      dimMapping.filter((dim) => isAxis(dim)),
+    ],
+    [dims, dimMapping]
+  );
+
+  const {
+    phaseArray,
+    phaseDomain = DEFAULT_DOMAIN,
+    amplitudeArray,
+    amplitudeDomain = DEFAULT_DOMAIN,
```
axelboc, line 62: You could put those domain fallbacks inside the hook, for simplicity.

src/h5web/vis-packs/core/complex/hooks.ts
```diff
@@ -0,0 +1,74 @@
...
+  dims: number[],
+  dimMapping: DimensionMapping,
+  scaleType: ScaleType = ScaleType.Linear
+): {
+  phaseArray: ndarray;
+  phaseDomain: Domain | undefined;
+  amplitudeArray: ndarray;
+  amplitudeDomain: Domain | undefined;
+} {
+  const [mappedArray] = useMappedArray(value, dims, dimMapping);
```
axelboc, line 20: Seems unnecessary and inconsistent with the other mapped vis to have this call to `useMappedArray` here instead of inlined in the mapped vis, especially since it requires passing two more props to `usePhaseAmplitude`.

src/h5web/vis-packs/core/heatmap/Mesh.tsx
```diff
@@ -55,6 +59,14 @@ function Mesh(props: Props) {
     return new DataTexture(valuesArr, cols, rows, RedFormat, FloatType);
   }, [cols, rows, values]);
 
+  const transparencyTexture = useMemo(() => {
```
axelboc, line 62: `alphaTexture`?

src/h5web/vis-packs/core/heatmap/Mesh.tsx
```diff
@@ -71,10 +83,16 @@ function Mesh(props: Props) {
   const shader = {
     uniforms: {
       data: { value: dataTexture },
+      useAlpha: { value: alphaValues ? 1 : 0 },
```
axelboc, line 86: I'd call this `withAlpha`.

src/h5web/vis-packs/core/heatmap/Mesh.tsx
```diff
@@ -119,6 +141,9 @@ function Mesh(props: Props) {
           gl_FragColor = nanColor;
         } else {
           gl_FragColor = texture2D(colorMap, vec2(scale(value), 0.5));
+          if(useAlpha == 1) {
+            gl_FragColor.a = oneOverAlphaRange*(texture2D(alpha, coords).r - alphaMin);
```
axelboc, line 145: spaces around `*` and after `if`, and a line break before `if`, maybe?

src/h5web/vis-packs/core/utils.ts
```diff
@@ -331,3 +325,17 @@ export function applyMapping<T>(
 
   return mappedArray;
 }
+
+export function getDomainValidForScale(
```
axelboc, line 329: `getValidDomainForScale` ?

src/h5web/vis-packs/core/complex/hooks.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    (mappedArray.data as H5WebComplex[]).forEach(([real, imag]) => {
+      const phase = Math.atan2(imag, real);
+      phaseBounds = getNewBounds(phaseBounds, phase);
+      phaseValues.push(phase);
+
+      const amplitude = Math.sqrt(real ** 2 + imag ** 2);
+      amplitudeBounds = getNewBounds(amplitudeBounds, amplitude);
+      amplitudeValues.push(amplitude);
+    });
```
axelboc, line 50: A `reduce` would have been nice 😁

src/h5web/vis-packs/core/complex/hooks.ts
```diff
@@ -0,0 +1,74 @@
...
+    ];
+  }, [mappedArray]);
+
+  const [phaseDomain, amplitudeDomain] = useMemo(
+    () => [
+      getDomainValidForScale(phaseBounds, scaleType),
+      getDomainValidForScale(amplitudeBounds, scaleType),
+    ],
+    [amplitudeBounds, phaseBounds, scaleType]
+  );
```
axelboc, line 66: The fact that you've separated these two `useMemo` makes me wonder whether there's an improvement to be made to `useDomain` as well. It seems we must be recomputing the domain by looping over the entire data array every time the scale changes, when this could be avoided.

axelboc (approved):
Impressive 👏

src/h5web/vis-packs/core/complex/hooks.ts
axelboc: Yep, testing with `entry_0000/0_measurement/images` confirms that switching the scale blocks the UI, indicating that the domain is being recomputed from scratch.

src/h5web/vis-packs/core/complex/hooks.ts
loichuder: I pushed a version with `reduce`: I am not fully convinced but tell me what you think ^^

loichuder:
@axelboc Can you have a last look ? I added a visualization test and a story for HeatmapVis

axelboc (approved):

Issue #616, opened 2021-04-20
Improve fetching for large datasets


loichuder:
Having played around with large datasets (~10M of points but a slice is ~100k), I find that the long fetch of the data breaks the flow... 

Let's make use of this issue to gather possible improvements:
- [ ] Allow the user to cancel requests
- [ ] Request domain separately to avoid computing in the front-end
- [ ] Make use of binary instead of JSON
- [ ] Implement specific strategies for fetching of large datasets (e.g. use subsampling; warn the user that the fetching can take a long time and ask them if they want to proceed anyway...)

axelboc:
I'd add that, when performing long downloads and computations, the UI should:

- be more informative (i.e. progress status, subsampling rate, etc.)
- remain responsive and allow cancelling slow computations (i.e. not just requests)

loichuder:
Our discussion on https://github.com/silx-kit/h5web/pull/632 also gave me an idea: we could make the flattening operation more consistent by encapsulating in the `get/useValue` method. 

It then becomes relevant to this issue as it would be a stepping stone to request the flattening on the back-end. Thus, avoiding another expensive computation in h5web.

axelboc:
In the providers' `getValue()` methods? Yeah, totally 👍

axelboc:
#635 implements cancellation on the front-end, but it doesn't resolve crashes on Bosquet when attempting to fetch (and cancel the fetch of) extremely large datasets.

axelboc:
#640 implements retrying after cancelling (including evicting cancellation errors from the value store's cache).

jreadey:
Just curious - for HSDS, have you tried using HTTP Compression?  That should reduce the payload size considerably.

loichuder:
Unfortunately, the impact will be limited as most of our heavy datasets are not compatible with HSDS due to https://github.com/HDFGroup/hsds/issues/76 :confused: 

But this is something that we still need to try !

loichuder:
The auto-scale feature in the `LineVis` that forces us to fetch the whole dataset can be a real limiter for huge datasets (https://github.com/silx-kit/jupyterlab-h5web/issues/71). 

Maybe, it is time to review it ? We could
- Disable it somehow for huge datasets. And give an indication to the user ?
- Request the domain separately (as proposed originally in https://github.com/silx-kit/h5web/issues/616#issue-862920494) to avoid the need to compute it in the front-end (and therefore the need to have the full dataset values)

[...]

PR #635, opened 2021-04-28
Allow cancelling fetching of dataset values

axelboc:
Part of #616

https://user-images.githubusercontent.com/2936402/116391251-33b6c280-a81f-11eb-89b0-6b049321cb02.mp4

- This is how cancellation is done with axios: https://github.com/axios/axios#cancellation
- To avoid duplication and provide a consistent implementation across all providers, I implemented cancellation in the providers' base class, `ProviderAPI`, which is therefore no longer `abstract`. I've also renamed it `ProviderApi` for consistency, but I've kept it in `context.ts` so you can see the diff. I'll move it into its own file before merging (any suggestion on how to name this file? `providerApi.ts` perhaps?
- The providers just need to wrap their cancellable fetches with `this.withCancelToken(...)`. This method accepts a callback function that receives a cancel token. The providers just need to pass this token to axios.
- `withCancelToken` stores the token sources in a `Set` so their `cancel` methods can be invoked when the user requests it (by clicking on the "Cancel" button below the loader).

I added some mock entities in a group called `resilience` and wrote some tests.


src/h5web/providers/context.ts
```diff
@@ -10,16 +16,45 @@ export type GetValueParams = {
...
+  public async withCancelToken<R>(
+    asyncFn: (cancelToken: CancelToken) => Promise<R>
+  ): Promise<R> {
+    const cancelSource = axios.CancelToken.source();
+    this.cancelSources.add(cancelSource);
+
+    try {
+      return await asyncFn(cancelSource.token);
+    } finally {
+      this.cancelSources.delete(cancelSource);
```
axelboc, line 45: We create one cancel source per request, so we need to clean it up if the request succeeds or fails. Reading this again, I wonder if I could have implemented this with a global, unique `cancelSource`...

loichuder (commented):
Good work !

I wonder if we could encapsulate all `cancelToken` logic inside the `ProviderApi`. Instead of `withCancelToken`, we could then have a method `cancellableGet(endpoint)` which would take care of adding the `cancelToken` in  `this.client.get` ?

Something like:
```js
function cancellableGet<T>(endpoint) {
    const cancelSource = axios.CancelToken.source();
    this.cancelSources.add(cancelSource);

    try {
      return await this.client.get<T>(endpoint), { cancelToken });;
    } finally {
      this.cancelSources.delete(cancelSource);
    }
}
```

src/h5web/providers/mock/MockProvider.tsx
```diff
@@ -1,76 +1,16 @@
...
-import { assertPrintableType, assertArrayShape } from '../../guards';
-import unpack from 'ndarray-unpack';
-import type { Primitive } from '../models';
-import type { PrintableType } from '../../vis-packs/core/models';
+import { MockApi } from './api';
 
 interface Props {
-  filepath?: string;
-  slowOnPath?: string;
-  errorOnPath?: string;
```
axelboc, line 15: I decided to remove those props and just hard-code the paths in the API. We used to make use of `errorOnPath` in the tests, but now that I've added a dedicated mock entity to test fetching errors, it's no longer needed.

src/h5web/providers/mock/MockProvider.tsx
```diff
@@ -1,76 +1,16 @@
...
   children: ReactNode;
 }
 
 function MockProvider(props: Props) {
-  const { filepath = mockFilepath, errorOnPath, slowOnPath, children } = props;
+  const { children } = props;
+  const api = useMemo(() => new MockApi(), []);
```
axelboc, line 11: We need a class now, so I've moved the mock API to its own file.

src/h5web/providers/mock/api.ts
```diff
@@ -0,0 +1,77 @@
...
+        return new Promise<void>((resolve, reject) => {
+          const timeout = setTimeout(() => resolve(), SLOW_TIMEOUT);
+
+          cancelToken.promise.then((error) => {
+            clearTimeout(timeout);
+            reject(error);
+          });
+        });
+      });
+    }
```
axelboc, line 52: This is the main change. I'm able to use the axios cancel token, since it's just a promise.

axelboc:
Ah I hadn't finished the review 😄 

Oh well, your suggestion makes sense 👍

src/h5web/providers/context.ts
loichuder: I don't see how: you might have several requests in parallel, right ? Which means several `cancelSource/Token` ?

loichuder (approved):
I like the addition of `fetchData` in the Jupyter api :+1: 

Did you test this with the different providers ? Does it behave as expected ?

src/h5web/providers/context.ts
axelboc: Yeah, but they could all reuse the same token. It's the same as having multiple tokens and cancelling them all at the same time. I've left it as is, because I suspect we'll need to cancel requests more intelligently in the future...

loichuder: I see. I don't mind keep it this way, this is as understandable if not more.

axelboc:
> Did you test this with the different providers ? Does it behave as expected ?

Of course. The cancellation in the browser works great for all providers. What goes around it still needs work (retrying and resetting the error boundaries, evicting errors from the stores so they don't get cached, making sure the server doesn't crash when dealing with extremely large datasets, etc.)

PR #638, opened 2021-04-28
Avoid recomputing bounds on scaleType changes

### LINKED ISSUE #636: Domain is recomputed from scratch when switching scale
1. Open `water_224::entry_0000/0_measurement/images`.
2. Switch to the line vis so the domain is computed over the entire, 10 million-point dataset.
2. Switch the scale => the UI gets blocked while the domain is being recomputed from scratch.

When switching scale, only `getValidDomainFromScale`, introduced in #634, needs to be called. The data array does not need to be looped over again.

loichuder:
Fix #636 

I had some inconsistent test failures locally on the `DomainSlider` but it passes on the CI...

 @axelboc can you try on your end to see if tests are passing ?


axelboc (approved):
Nice! Glad the public API (`(get|use)Domains?`) doesn't change. 🎉 

Tests fixed.

PR #639, opened 2021-04-29
Fix HSDS responses containing NaN (which is invalid JSON)

loichuder:
A temporary fix to work around https://github.com/HDFGroup/hsds/issues/87

To be removed when the issue is solved in HSDS.


axelboc (approved):
👏

PR #640, opened 2021-04-29
Allow user to retry after cancelling a fetch

axelboc:
It's all thanks to `react-error-boundary` and `react-suspense-fetch`:

1. In the boundary, when the error is a cancellation error, I show a button "Retry?", which invokes a callback provided by the boundary component called `resetErrorBoundary`.
2. Where the boundary is used (in the `Visualizer`), I pass it a prop called `onReset`, which is invoked after `resetErrorBoundary` is invoked.
3. In `onReset`, I evict the value of the entity at the current path from the store.
4. The boundary then triggers a re-render of its children, which calls `useDatasetValue`, which triggers a re-fetch since there's no longer a value at the current path in the cache.

https://user-images.githubusercontent.com/2936402/116522050-0298c980-a8d5-11eb-9904-ce9fc605a6d8.mp4

Dealing with cancelling and retrying when fetching slices was a bit trickier:

- I had to add error boundaries in the `HeatmapVisContainer` and the `NxImageVisContainer` so the dimension mapper wouldn't disappear while fetching a slice.
- Then, when retrying and calling the stores `evict` method, I had to pass the same slice selection string that was used to fetch the current slice so it would be properly evicted from the store's cache.

https://user-images.githubusercontent.com/2936402/116522060-0593ba00-a8d5-11eb-9d73-0f7f0e17d307.mp4


src/h5web/App.tsx
```diff
@@ -44,7 +44,7 @@ function App() {
         />
         <ErrorBoundary
           resetKeys={[selectedPath, isInspecting]}
-          FallbackComponent={ErrorMessage}
+          FallbackComponent={ErrorFallback}
```
axelboc, line 47: Renaming this component to distinguish it from the domain slider's `ErrorMessage` component. `ErrorFallback` (passed to `ErrorBoundary`) is consistent with `LoadingFallback` (passed to `Suspense`).

src/h5web/vis-packs/core/containers/HeatmapVisContainer.tsx
```diff
@@ -21,21 +25,34 @@ function HeatmapVisContainer(props: VisContainerProps) {
...
-          title={name}
-        />
-      </Suspense>
+      <ErrorBoundary
+        resetKeys={[dimMapping]}
+        FallbackComponent={ErrorFallback}
+        onReset={() => {
+          valuesStore.evict({
+            path: entity.path,
+            selection: getSliceSelection(dimMapping),
```
axelboc, line 43: This is how I evict a slice from the `valuesStore`'s cache.

src/h5web/visualizer/Visualizer.tsx
```diff
@@ -43,7 +45,10 @@ function Visualizer<T extends VisDef>(props: Props<T>) {
         <div className={styles.displayArea}>
           <ErrorBoundary
             resetKeys={[entity.path]}
-            FallbackComponent={ErrorMessage}
+            FallbackComponent={ErrorFallback}
+            onReset={() => {
+              valuesStore.evict({ path: entity.path });
```
axelboc, line 50: For other visualizations, I just need to pass the entity's `path` since there's no slice selection.

src/h5web/vis-packs/core/containers/HeatmapVisContainer.tsx
loichuder: What happens if the error caught by the boundary is not a `RequestCancelled` error ? Will it also remove the value from the store on reset ?

src/h5web/visualizer/Visualizer.test.tsx
```diff
@@ -82,4 +82,50 @@ test('cancel slow fetch of dataset value', async () => {
...
+test('cancel and retry slow fetch of dataset slice', async () => {
+  renderApp();
+
+  jest.useFakeTimers();
+  const errorSpy = mockConsoleMethod('error');
+
+  await selectExplorerNode('resilience/slow_slicing');
+  expect(await screen.findByText(/Loading data/)).toBeVisible();
+
+  // Cancel fetch of first slcie
```
loichuder, line 105:
```suggestion
  // Cancel fetch of first slice
```

loichuder (approved):
Excellent work on the tests !

I just had a question about the reset of the errorBoundary

src/h5web/vis-packs/core/containers/HeatmapVisContainer.tsx
axelboc: I only show the `retry` button when the error is a `RequestCancelled` error.

loichuder: Is this the only way to reset the error boundary from the UI ?

axelboc: What do you have in mind?

loichuder: Nothing in particular. I was just wondering if an user could remove the value from the store through a side-effect, which is not what we want.

axelboc: Right, I think I understand your concern: you're asking when `onReset` is called basically?

It only gets called when the boundary is reset manually by calling `resetErrorBoundary` inside it. When the boundary resets because one of the `resetKeys` passed to it changes, `onReset` is not called.

Does that make sense?

loichuder: It does ! It also lifts my concerns, thanks :slightly_smiling_face:

PR #642, opened 2021-04-29
Explicitly treat NaN in Vis

loichuder:


axelboc:
So this is no longer relevant, right? We're aiming to support `NaN` and `Infinity` in number arrays, right?

PR #643, opened 2021-04-29
Add suspense/error boundaries to containers and mock slow NXdata groups

axelboc:
1. Turns out that `NxImageContainer` and `HeatmapVisContainer` are not the only visualizations fetching slices of data, so we need to add error boundaries to more containers to handle retries properly, namely to `PhaseMapVisContainer`, `MatrixVisContainer`.
2. Seeing the dimension mapper and being able to interact with it while data is loading is really nice, so I've added a suspense boundary to `LineVisContainer` and `NxSpectrumContainer` as well.
3. When opening an `NXdata` group and cancelling, we potentially cancel multiple requests at once. So when we click "Retry", evicting only the signal dataset as we do at the moment doesn't always work. We need error boundaries in `NxImageContainer` and `NxSpectrumContainer` that evict every dataset that was potentially fetched. In this PR, I investigated this problem by adding two mock `NXdata` group in the `resilience` group that slowly fetch every possible NeXus-related dataset (signal, axis, auxiliary, title, errors). I'll work on fixing the eviction code in my next PR.
4. After creating the two slow mock `NXdata` groups, and adding suspense boundaries to `NxImageContainer` and `NxSpectrumContainer`, I identified a new problem: the fetching of the `title` and `axes` datasets was being done in the containers, which meant that they were triggering the parent suspense boundary (in `Visualizer.tsx`) and causing a loading waterfall effect, delaying the fetching of the signals and errors datasets. I fixed this by moving all the fetching to `MappedHeatmapVis` and `MappedLineVis`. Obviously this makes the mapped components even more complex, but now that the fetching is all done in one place, I think we'll be able to find better ways to organise the code (by separating the fetching from the mapping/slicing completely).


axelboc:
I give up. This bloody test was killing me, so I've removed it. There's another test in the same file that renders a 2D signal as NX image, so it's not worth spending more time on it.

src/h5web/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -66,6 +75,10 @@ function MappedLineVis(props: Props) {
   const auxiliaries = Object.values(useDatasetValues(auxDatasets));
   const [auxArrays, auxForDomain] = useMappedArrays(auxiliaries, ...hookArgs);
 
+  const title =
+    useDatasetValue(titleDataset) || valueLabel || valueDataset.name;
```
loichuder, line 79: Let's remove `valueLabel` from the props and use `getDatasetLabel(valueDataset)` as in `MappedHeatmapVis`

loichuder (approved):

loichuder:
For what it is worth, I managed to have the test pass by either:
- removing `defaultPath` from `nexus_entry`
- selecting `nexus_entry/nx_process/nx_data` in two steps:
```js
renderApp();
await selectExplorerNode('nexus_entry/nx_process');
await selectExplorerNode('nx_data');
```

Not sure if that helps... :shrug:

src/h5web/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -2,30 +2,40 @@ import { useEffect, useMemo } from 'react';
...
+import type { AxisDatasetMapping } from '../../nexus/models';
 
 interface Props {
   dataset: NumArrayDataset;
   dims: number[];
   dimMapping: DimensionMapping;
-  axisMapping?: AxisMapping;
-  title?: string;
+  titleDataset?: Dataset<ScalarShape, StringType>;
+  axisDatasetMapping?: AxisDatasetMapping;
```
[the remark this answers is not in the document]
axelboc: and keep the type as is?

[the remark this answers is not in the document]
loichuder: Yes

src/h5web/vis-packs/core/line/MappedLineVis.tsx
axelboc: There's a subtlety: if present, `valueLabel` is used as the ordinates label as well. This is the case with `NxSpectrumContainer` (but not `LineVisContainer`, which only displays the `title`).

axelboc:
> Not sure if that helps... 🤷

It might. I'll try refactoring the `selectExplorerNode` method with `userEvent` and with a loop instead of recursion... Let's see what happens... 🤷

src/h5web/vis-packs/core/line/MappedLineVis.tsx
loichuder: True but that does not prevent you from removing `valueLabel` from the props and instead define it as `valueLabel = getDatasetLabel(valueDataset)` in `MappedLineVis`.

axelboc: Left as is, as discussed.

axelboc:
I refactored `selectExplorerNode` as mentioned, but the test still failed. So I added a `default` path to `nx_data` as a temporary measure. There's definitely something going on that we don't quite understand. Hopefully the refactorings we discussed will make the issue easier to debug.

PR #645, opened 2021-04-30
Fix errors on Infinity and on slices with integers for HSDS

loichuder:
Linked to:
* https://github.com/HDFGroup/hsds/issues/87
* https://github.com/HDFGroup/hsds/issues/88

### DIFF
--- a/src/h5web/providers/hsds/api.ts
+++ b/src/h5web/providers/hsds/api.ts
@@ -56,7 +56,7 @@ export class HsdsApi extends ProviderApi {
         } catch {
           // https://github.com/HDFGroup/hsds/issues/87
           try {
-            return JSON.parse(data.replaceAll('NaN', '"NaN"'));
+            return JSON.parse(data.replaceAll(/(NaN|-?Infinity)/gu, '"$&"'));
           } catch {
             return data;
           }
@@ -104,7 +104,18 @@ export class HsdsApi extends ProviderApi {
     const entity = await this.getEntity(path);
     assertHsdsDataset(entity);
 
-    return this.fetchValue(entity.id, selection);
+    const value = await this.fetchValue(entity.id, selection);
+
+    // https://github.com/HDFGroup/hsds/issues/88
+    // HSDS does not reduce the number of dimensions when selecting indices
+    // This is an issue when trying to slice nD arrays
+    // Therefore the dimension reduction is done "manually" here with a flat
+    const numberIndices = selection.match(/\d+/gu)?.length;
+    if (!numberIndices) {
+      return value;
+    }
+
+    return (value as unknown[]).flat(numberIndices);
   }
 
   private async fetchRootId(): Promise<HsdsId> {


src/h5web/providers/hsds/api.ts
```diff
@@ -104,7 +104,18 @@ export class HsdsApi extends ProviderApi {
...
+    // https://github.com/HDFGroup/hsds/issues/88
+    // HSDS does not reduce the number of dimensions when selecting indices
+    // This is an issue when trying to slice nD arrays
+    // Therefore the dimension reduction is done "manually" here with a flat
+    const numberIndices = selection.match(/\d+/gu)?.length;
+    if (!numberIndices) {
+      return value;
+    }
+
+    return (value as unknown[]).flat(numberIndices);
```
loichuder, line 118: The problem arises in HeatmapVis because we make now use of the slice dimensions to flatten the array up to a certain point: https://github.com/silx-kit/h5web/pull/622#discussion_r618125585. Recall that this is needed as we do not want to flatten the complex data stored as `[number, number]`.

However, if the slice dimensions is not matching the true dimensions of the fetched data JSON, the data contained in the `ndarray` is bosh. This breaks the HeatmapVis and PhaseMapVis for nD arrays.

src/h5web/providers/hsds/api.ts
```diff
@@ -56,7 +56,7 @@ export class HsdsApi extends ProviderApi {
         } catch {
           // https://github.com/HDFGroup/hsds/issues/87
           try {
-            return JSON.parse(data.replaceAll('NaN', '"NaN"'));
+            return JSON.parse(data.replaceAll(/(NaN|-?Infinity)/gu, '"$&"'));
```
loichuder, line 59: Extends https://github.com/silx-kit/h5web/pull/639 for `Infinity` and `-Infinity` that have the same problem as `NaN`

src/h5web/providers/hsds/api.ts
axelboc: I don't get why this works. Doesn't `value.flat(rawDims.length - 1)` in `getBaseArray` then flattens the inner complex arrays?

loichuder: Let me take an example:

We have an array 4D of dimensions `[5, 4, 3, 2]`. We want to display the slice `0,0,:,:` in the `HeatmapVis` of dimensions `[4, 5]`.
- We request the slice in `MappedHeatmapVis` through `useDatasetValue`. As we pass a mapping, `value` will only contain the slice `0,0,:,:`/
- We build the ndarray in `useMappedArray` **using the `slicedDims` `[3, 2]**. `value.flat(rawDims.length - 1)`  in `getBaseArray` is therefore `value.flat(1)`.

That means we expect `value` to have only one level of nesting: `[ [x_00, x_01], [x_10, x_11], [x_21, x_22] ]`. This is what we get from the `JupyterProvider` and the `MockProvider`.

However, due to https://github.com/HDFGroup/hsds/issues/88, the slice we get keep the dimensionnalilty of the original array. We get then `[ [ [ [x_00, x_01], [x_10, x_11], [x_21, x_22] ] ] ]` which has the shape `[1, 1, 3, 2]`. The fix I implemented here is then required for the `value.flat(1)` to work as intended.

The fact `x_ij` can be `number` or `[number, number]` is only relevant as it prevents us from doing `value.flat(Infinity)`.

axelboc: Makes sense, thanks!

axelboc (approved):

PR #647, opened 2021-05-05
Fix retrying to fetch 2D dataset from Heatmap vis

axelboc:
There was a small bug that prevented retrying the fetch of a 2D dataset from the Heatmap vis.

### DIFF
--- a/src/h5web/providers/utils.ts
+++ b/src/h5web/providers/utils.ts
@@ -1,4 +1,3 @@
-import { pickBy } from 'lodash-es';
 import { createFetchStore } from 'react-suspense-fetch';
 
 export interface ObjectKeyStore<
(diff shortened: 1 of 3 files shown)


src/h5web/providers/utils.ts
```diff
@@ -27,14 +26,7 @@ export function createObjectKeyStore<
...
   function wrapMethod<T>(method: (key: string) => T) {
-    return (params: I) => {
-      /* Remove optional params that are explicitely set to `undefined`.
-       * This is to guarantee that, for instance, `{ foo: undefined; }` and `{}`
-       * are both stringified to (and cached with) "{}" */
-      const pickedParams = pickBy(params, (v) => v !== undefined);
-
-      return method(JSON.stringify(pickedParams));
-    };
+    return (params: I) => method(JSON.stringify(params));
```
axelboc, line 29: I don't know what I was on at the time, but `JSON.stringify({ foo: undefined })` does return `{}` out of the box... I think it's a left over from a previous attempt I had made with passing params separately instead of together in single `params` object.

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -34,11 +33,7 @@ export function useDatasetValue(
     return undefined;
   }
 
-  // If no dim mapping is provided or dim mapping has no slicing dimension, fetch entire dataset
-  if (!dimMapping || !dimMapping.some(isNumber)) {
-    return valuesStore.get({ path: dataset.path });
-  }
```
axelboc, line 40: This logic is now inside `getSliceSelection` which is called when performing the eviction in the Heatmap vis error boundary, so that `evict` is always called with the same object as `get`. The bug was that, when `dimMapping` had no slicing dimension, `evict` was called with an empty string `selection` (instead of just `undefined` like above)

loichuder (approved):

PR #652, opened 2021-05-07
Fix cancelling/retrying in NeXus visualizations

axelboc:
Retrying a cancelled NeXus vis now works, but leads to a waterfall effect.

The calls to `valuesStore.prefetch` are located in `useNxData`, itself called from the NeXus vis containers. When the error boundary in the container resets, it re-renders its children, not the container itself, so the prefetches aren't triggered and the requests in the mapped vis happen sequential instead of in parallel.

I will fix this behaviour in my next PR by moving the prefetching down into the mapped vis.

---

**UPDATE**

The initial solution had problems with the slicing. The new solution is much better, as it's completely generic.

Everything is done in in the base class of the provider APIs; the containers' error boundaries no longer have to evict requests from the store by passing the right cache keys. The API remembers the params of the requests that have been cancelled and allows the `Provider` to access them so it can evict them from its `valuesStore`.

This solution has the following added benefits:

- It does not evict values from the store that may have been fetched successfully.
- The error boundaries of both core and NeXus containers all share the same code: an `onError` handler that calls the `evictCancelledValues` method from the `ProviderContext`. This means we can extract the code more easily than we had previously discussed, into a single shared `VisContainer`.

### DIFF
--- a/src/h5web/providers/Provider.tsx
+++ b/src/h5web/providers/Provider.tsx
@@ -39,9 +39,19 @@ function Provider(props: Props) {
   }, [api]);
 
   const valuesStore = useMemo(() => {
-    return createFetchStore(api.getValue.bind(api), {
+    const store = createFetchStore(api.getValue.bind(api), {
       type: 'Map',
       areEqual: (a, b) => a.path === b.path && a.selection === b.selection,
+    });
+
+    return Object.assign(store, {
+      cancelOngoing: () => api.cancelValueRequests(),
+      evictCancelled: () => {
+        api.cancelledValueRequests.forEach(({ params }) => {
+          valuesStore.evict(params);
+        });
+        api.cancelledValueRequests.clear();
+      },
     });
   }, [api]);
 
@@ -51,7 +61,6 @@ function Provider(props: Props) {
         filepath: api.filepath,
         entitiesStore,
         valuesStore,
-        cancel: () => api.cancel(),
       }}
     >
       {children}
--- a/src/h5web/vis-packs/core/containers/ComplexVisContainer.tsx
+++ b/src/h5web/vis-packs/core/containers/ComplexVisContainer.tsx
@@ -12,7 +12,6 @@ import ValueLoader from '../../../visualizer/ValueLoader';
 import MappedComplexVis from '../complex/MappedComplexVis';
 import { ErrorBoundary } from 'react-error-boundary';
 import ErrorFallback from '../../../visualizer/ErrorFallback';
-import { getSliceSelection } from '../utils';
 import { ProviderContext } from '../../../providers/context';
 
 function ComplexVisContainer(props: VisContainerProps) {
@@ -37,12 +36,7 @@ function ComplexVisContainer(props: VisContainerProps) {
       <ErrorBoundary
         resetKeys={[dimMapping]}
         FallbackComponent={ErrorFallback}
-        onReset={() => {
-          valuesStore.evict({
-            path: entity.path,
-            selection: getSliceSelection(dimMapping),
-          });
-        }}
+        onError={() => valuesStore.evictCancelled()}
       >
         <Suspense fallback={<ValueLoader message="Loading current slice" />}>
           <MappedComplexVis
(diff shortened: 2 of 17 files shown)


src/h5web/vis-packs/nexus/NexusPack.test.tsx
```diff
@@ -83,3 +84,76 @@ test('show error when encountering malformed NeXus metadata', async () => {
...
+  await selectExplorerNode('resilience/slow_nx_spectrum');
+  expect(await screen.findByText(/Loading data/)).toBeVisible();
+
+  const errorSpy = mockConsoleMethod('error');
+  userEvent.click(await screen.findByRole('button', { name: /Cancel/ }));
+
+  expect(await screen.findByText('Request cancelled')).toBeVisible();
+  expect(errorSpy).toHaveBeenCalledTimes(2); // React logs two stack traces
+
+  // Move dimension slider
```
loichuder, line 101: I think the comment could be more meaningful here to explain the purpose of this part. Something like `Request should be still cancelled after moving dimension slider`

src/h5web/vis-packs/nexus/NexusPack.test.tsx
```diff
@@ -83,3 +84,76 @@ test('show error when encountering malformed NeXus metadata', async () => {
...
+
+  await selectExplorerNode('resilience/slow_nx_image');
+  expect(await screen.findByText(/Loading data/)).toBeVisible();
+
+  const errorSpy = mockConsoleMethod('error');
+  userEvent.click(await screen.findByRole('button', { name: /Cancel/ }));
+  expect(await screen.findByText('Request cancelled')).toBeVisible();
+  expect(errorSpy).toHaveBeenCalledTimes(2); // React logs two stack traces
+  errorSpy.mockRestore();
+
```
loichuder, line 142: If we move the slider here, it will request an another slice, right ?

loichuder (approved):

src/h5web/providers/Provider.tsx
```diff
@@ -51,7 +51,11 @@ function Provider(props: Props) {
         filepath: api.filepath,
         entitiesStore,
         valuesStore,
-        cancel: () => api.cancel(),
+        cancelValues: () => api.cancelValues(),
+        evictCancelledValues: () => {
+          api.cancelledValueParams.forEach((key) => valuesStore.evict(key));
+          api.cancelledValueParams.clear();
+        },
```
axelboc, line 58: The solution works in two stages:

1. When the user clicks on "Cancel", the `cancelValues` method is called. This cancels the requests and stores the params that were used to make them in a `Set` named `cancelledValueParams`.
2. The error boundary catches the "Request cancelled" error, then calls its `onError` handler. The handler then calls `evictCancelledValues`, which reads the params of the requests that have been cancelled and evict the corresponding cache entries from the `valuesStore`.

src/h5web/vis-packs/core/containers/ComplexVisContainer.tsx
```diff
@@ -37,12 +36,7 @@ function ComplexVisContainer(props: VisContainerProps) {
       <ErrorBoundary
         resetKeys={[dimMapping]}
         FallbackComponent={ErrorFallback}
-        onReset={() => {
-          valuesStore.evict({
-            path: entity.path,
-            selection: getSliceSelection(dimMapping),
-          });
-        }}
+        onError={() => evictCancelledValues()}
```
axelboc, line 39: Here is the magic call. It's the same in every container, so it will be trivial to remove the duplication later.

src/h5web/vis-packs/core/containers/LineVisContainer.tsx
```diff
@@ -19,20 +22,28 @@ function LineVisContainer(props: VisContainerProps) {
...
-          valueDataset={entity}
-          dims={dims}
-          dimMapping={dimMapping}
-        />
-      </Suspense>
+      <ErrorBoundary
+        resetKeys={[dimMapping]}
+        FallbackComponent={ErrorFallback}
+        onError={() => evictCancelledValues()}
+      >
```
axelboc, line 38: Just adding an `ErrorBoundary` in this container as well. The only containers without a boundary (and that rely on the `Visualizer`'s boundary) are now the `RawVisContainer` and `ScalarVisContainer`. Not sure yet how I'll deal with them in the containers' refactoring.

src/h5web/vis-packs/nexus/NexusPack.test.tsx
axelboc: - Moving the slider on an `NxSpectrum` visualization now automatically retries fetching the entire dataset, as well as any cancelled supporting datasets (`title`, `axes`, etc.) If some datasets were successfully fetched, they are not refetched.
- Moving the slider on an `NxImage` visualization fetches the new slice and retries fetching any cancelled supporting datasets. Coming back to a cancelled slice automatically retries fetching that slice as well.

Issue #631, opened 2021-04-26
[LineVis] The line does not appear when importing from @h5web/lib  


loichuder:
This is because the color of the line is set via `--secondary-dark` which is not set when importing from @h5web/lib. 

Actually, this is true for the color of ticks, grid and auxiliary signals which are also set via CSS variables.

Linked to #555

loichuder:
Bug reproduced in https://codesandbox.io/s/h5weblib-showcase-qbx5z

Issue #555, opened 2021-03-10
Review the use of CSS variables


loichuder:
Working on #470, we realized that setting line colors from CSS variables is a bit clunky and introduces obscure coupling between CSS and JS.

@axelboc proposed that we instead:
- use a `Provider` to explicitly set colors:
```jsx
<ThemeProvider theme="dark"> // or `theme={{ auxSignals: ['#000', '#111', '#222'] }}`
  <AppOrHeatmapOrWhatever />
</ThemeProvider>
```
- have then a hook `useTheme` that can return a default theme if the provider is not used
- use the `Provider` to read colors from the CSS or (the other way round) inject the colors in the CSS

axelboc:
This will make disabling transparency on the WebGL renderer [for performance reasons](https://discoverthreejs.com/tips-and-tricks/#renderer)) possible, as it will be easy to access the canvas' background colour so it can be [passed to the scene](https://github.com/pmndrs/react-three-fiber/issues/304#issuecomment-743038591)

axelboc:
After much thinking, I think a CSS-driven solution is cleaner and more flexible:

- Following @loichuder's investigation, it seems complicated to access the user's theme from JS in JupyterLab, which makes the idea of a `ThemeProvider` impractical. Overriding CSS custom properties based on the presence or absence of a `data-theme` attribute remains the most viable option.
- Users of `@h5web/lib` may not want to clutter their React application with a `ThemeProvider`. It seems overly complicated for users who just want to use one visualisation to have to add this provider.
- It is unclear whether a `ThemeProvider`-based solution would allow overriding variables contextually (e.g. in one part of the app but not another)
- We need access to the colours from React only in very few places, so adding this provider complexity to the React code seems unnecessary.

A CSS-driven solution allows for much more flexibility:

- Assuming we prefix our variables, users can just import a stylesheet in their project for all the variables to be made available on `:root`.
- If they prefer, users can declare only the variables they need where they need them.
- Users can easily override variables in some parts of their app and not others, or based on any CSS selector they choose (which makes it easy to change colours dynamically from React).

Reading the value of a CSS variable in React is not too complicated, and I've made a custom hook for it in #660. In the end, we'll probably need to use this hook only in the top-level visualization components, as low-level components like `DataCurve` can just accept valid colors as props (as done in #660).

PR #660, opened 2021-05-14
Read CSS custom property in `LineVis` instead of `DataCurve`

axelboc:
This relates to #631 and #555.

`DataCurve` no longer relies on H5Web's CSS custom properties. The `color` prop is now expected to be a valid color string. I've kept a default value for convenience, but I've chosen `black`. (If we add a `bgColor` prop to the canvas as per https://github.com/silx-kit/h5web/issues/555#issuecomment-821144768, we can use `white` as default.)

The CSS custom properties are now read by `LineVis` on the root `figure` element via a `ref` (instead of on the `canvas` element via `useThree`).

### DIFF
--- a/src/h5web/vis-packs/core/line/LineVis.module.css
+++ b/src/h5web/vis-packs/core/line/LineVis.module.css
@@ -5,8 +5,3 @@
   min-height: 0;
   margin: 1rem;
 }
-
-.root canvas {
-  /* This sets the default color of the line */
-  color: var(--secondary-dark);
-}
--- a/src/h5web/vis-packs/core/hooks.ts
+++ b/src/h5web/vis-packs/core/hooks.ts
@@ -1,5 +1,12 @@
 import type { NdArray } from 'ndarray';
-import { useContext, useEffect, useMemo, useState } from 'react';
+import {
+  RefCallback,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import { createMemo } from 'react-use';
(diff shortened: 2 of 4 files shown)


src/h5web/vis-packs/core/line/LineVis.module.css
```diff
@@ -5,8 +5,3 @@
   min-height: 0;
   margin: 1rem;
 }
-
-.root canvas {
-  /* This sets the default color of the line */
-  color: var(--secondary-dark);
-}
```
axelboc, line 12: Left-over from a previous refactoring. We don't read the `color` property but the CSS custom color properties themselves.

loichuder (approved):
A good step forward !

For the future, I wonder if we should instead give the possibility to change the color in `LineVis` (that is the exported component) and make use of the `useCSSCustomProperties` in `MappedLineVis` which is our internal component.

axelboc:
> For the future, I wonder if we should instead give the possibility to change the color in LineVis (that is the exported component) and make use of the useCSSCustomProperties in MappedLineVis which is our internal component.

Yeah, not sure. The counterpoint is that ticks for instance would still rely on CSS variables, so if users have to declare variables for ticks, they might as well declare one or two more for the curve(s). I think it makes sense to provide the same colour customisation technique for every element -- i.e. either only props or only variables. Let's talk further before I merge.

loichuder:
> Let's talk further before I merge.

To be clear, this can be merged as it is as it solves an immediate problem. My remark indeed needs further discussion :smile:

PR #661, opened 2021-05-17
Values are now flattened in providers 

loichuder:
Work towards https://github.com/silx-kit/h5web/issues/616#issuecomment-828238108

This means that the `valuesStore` now only stores flattened values.

There are several points I would like to discuss that I highlight below.

### DIFF
--- a/src/h5web/providers/hsds/hsds-api.ts
+++ b/src/h5web/providers/hsds/hsds-api.ts
@@ -21,7 +21,7 @@ import {
   Group,
   ProviderError,
 } from '../models';
-import { assertDefined, assertGroup } from '../../guards';
+import { assertDefined, assertGroup, hasArrayShape } from '../../guards';
 import { ProviderApi } from '../api';
 import {
   assertHsdsDataset,
@@ -100,7 +100,7 @@ export class HsdsApi extends ProviderApi {
   }
 
   public async getValue(params: ValueRequestParams): Promise<unknown> {
-    const { path, selection = '' } = params;
+    const { path } = params;
 
     const entity = await this.getEntity(path);
     assertHsdsDataset(entity);
@@ -109,14 +109,12 @@ export class HsdsApi extends ProviderApi {
 
     // https://github.com/HDFGroup/hsds/issues/88
     // HSDS does not reduce the number of dimensions when selecting indices
-    // This is an issue when trying to slice nD arrays
-    // Therefore the dimension reduction is done "manually" here with a flat
-    const numberIndices = selection.match(/\d+/gu)?.length;
-    if (!numberIndices) {
-      return value;
+    // Therefore the flattening must be done on all dimensions regardless of the selection
+    if (hasArrayShape(entity)) {
+      return (value as unknown[]).flat(entity.shape.length - 1);
     }
 
-    return (value as unknown[]).flat(numberIndices);
+    return value;
   }
 
   private async fetchRootId(): Promise<HsdsId> {
--- a/src/h5web/providers/jupyter/jupyter-api.ts
+++ b/src/h5web/providers/jupyter/jupyter-api.ts
@@ -24,7 +24,7 @@ import type {
   JupyterMetaResponse,
 } from './models';
 import { makeStrAttr } from '../mock/metadata-utils';
-import { assertDataset, hasComplexType } from '../../guards';
+import { assertDataset, hasArrayShape, hasComplexType } from '../../guards';
 
 export class JupyterStableApi extends ProviderApi {
   /* API compatible with jupyterlab_hdf v0.5.1 */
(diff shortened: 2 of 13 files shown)


src/h5web/providers/hsds/hsds-api.ts
```diff
@@ -109,14 +109,12 @@ export class HsdsApi extends ProviderApi {
...
     // HSDS does not reduce the number of dimensions when selecting indices
-    // This is an issue when trying to slice nD arrays
-    // Therefore the dimension reduction is done "manually" here with a flat
-    const numberIndices = selection.match(/\d+/gu)?.length;
-    if (!numberIndices) {
-      return value;
+    // Therefore the flattening must be done on all dimensions regardless of the selection
+    if (hasArrayShape(entity)) {
+      return (value as unknown[]).flat(entity.shape.length - 1);
     }
```
loichuder, line 115: https://github.com/HDFGroup/hsds/issues/88 works in our favour here as `value` will always have the same number of dimensions of the original dataset, even when slicing

src/h5web/providers/jupyter/jupyter-api.ts
```diff
@@ -161,4 +163,14 @@ export class JupyterStableApi extends ProviderApi {
 
     return contents;
   }
+
+  protected flattenValue(value: unknown, entity: Dataset, selection?: string) {
+    if (!hasArrayShape(entity)) {
+      return value;
+    }
+
+    const slicedDims = selection?.split(',').filter((s) => s.includes(':'));
```
loichuder, line 172: In Jupyter, the returned `value` dimensions change with the slicing. 

I detect when a dimension is conserved by checking for `:`.

src/h5web/providers/jupyter/jupyter-dev-api.ts
```diff
@@ -32,7 +33,15 @@ export class JupyterDevApi extends JupyterStableApi {
...
-    return this.fetchData(params);
+    const { path, selection } = params;
+    const [value, entity] = await Promise.all([
+      this.fetchData(params),
+      this.getEntity(path),
+    ]);
+
+    assertDataset(entity);
+
+    return this.flattenValue(value, entity, selection);
```
loichuder, line 44: As I need to know the dataset dimensions to flatten, I need to fetch the entity along with the data...

This was not an issue for `JupyterStable` and `Hsds` as the entity was already fetched for other reasons (complex type parsing in Jupyter and dataset assertion in Hsds)

This is not ideal but won't be a problem anymore if we move the flattening in the backend

src/h5web/providers/mock/mock-api.ts
```diff
@@ -39,28 +43,27 @@ export class MockApi extends ProviderApi {
...
+    const mappedArray = applyMapping(
+      dataArray,
+      selection
         .split(',')
-        .map((s) => (s === ':' ? null : Number.parseInt(s, 10)))
+        .map((s) => (s === ':' ? 'x' : Number.parseInt(s, 10)))
     );
 
-    return unpack(dataView);
+    return mappedArray.data;
```
loichuder, line 66: I cannot use `dataView.data` as it holds the data of the full `dataArray`. I therefore make use for `applyMapping` which allow me to remove `unpack` from the project dependencies.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -305,7 +305,7 @@ export function getBaseArray<T>(
   value: T[] | undefined,
   rawDims: number[]
 ): NdArray<T> | undefined {
-  return value && ndarray<T>(value.flat(rawDims.length - 1) as T[], rawDims);
+  return value && ndarray<T>(value, rawDims);
```
loichuder, line 308: `value` is now expected to be flat

src/h5web/providers/mock/mock-api.ts
```diff
@@ -39,28 +43,27 @@ export class MockApi extends ProviderApi {
       await this.cancellableDelay(params);
     }
 
-    const { value } = dataset;
+    const { value: rawValue } = dataset;
+    const value = hasArrayShape(dataset)
+      ? (rawValue as unknown[]).flat(dataset.shape.length - 1)
+      : rawValue;
```
loichuder, line 49: Note that we could remove this step if we store only flat arrays as mock data

src/h5web/providers/mock/mock-api.ts
axelboc: Clever. Seems a bit hacky to replace `:` with `x`, though. Maybe `applyMapping` could accept the type `(':' | number)[]`? Don't think you'd have to change the implementation.

src/h5web/vis-packs/core/complex/MappedComplexVis.tsx
```diff
@@ -45,13 +44,7 @@ function MappedComplexVis(props: Props) {
...
   const { visType } = useComplexConfig((state) => state, shallow);
 
-  const [slicedDims, slicedMapping] = useMemo(
-    () => [
-      dims.filter((_, i) => isAxis(dimMapping[i])),
-      dimMapping.filter((dim) => isAxis(dim)),
-    ],
-    [dims, dimMapping]
-  );
+  const [slicedDims, slicedMapping] = useSliceDimsAndMapping(dims, dimMapping);
```
axelboc, line 47: `useSlicedDimsAndMapping` (missing a `d`)

axelboc (approved):
Nicely done

src/h5web/vis-packs/core/utils.ts
```diff
@@ -305,24 +305,26 @@ export function getBaseArray<T>(
...
+  mapping: (number | Axis | ':')[]
 ): NdArray<T> | undefined {
   if (!baseArray) {
     return undefined;
   }
 
   const isXBeforeY =
-    mapping.includes('y') && mapping.indexOf('x') < mapping.indexOf('y');
+    mapping.includes('y') &&
+    mapping.includes('x') &&
```
axelboc, line 326: Ah nice 👏

axelboc (approved):

PR #662, opened 2021-05-17
Add "Customization" page to Storybook and first two semantic CSS vars for line vis

### LINKED ISSUE #631: [LineVis] The line does not appear when importing from @h5web/lib

### LINKED ISSUE #644: Document CSS requirement for parent element of visualizations
The parent needs `display: flex` and a height, which we don't document.

axelboc:
Fix #631, fix #644 and proof of concept for #555 

image

image

### DIFF
--- a/src/h5web/vis-packs/core/line/LineVis.tsx
+++ b/src/h5web/vis-packs/core/line/LineVis.tsx
@@ -38,7 +38,7 @@ function LineVis(props: Props) {
     ordinateLabel,
     title,
     errorsArray,
-    showErrors,
+    showErrors = false,
     auxArrays = [],
   } = props;
 
--- a/src/h5web/vis-packs/core/line/LineVis.tsx
+++ b/src/h5web/vis-packs/core/line/LineVis.tsx
@@ -77,7 +77,7 @@ function LineVis(props: Props) {
   const {
     colors: [curveColor, auxColor],
     refCallback: rootRef,
-  } = useCSSCustomProperties('--secondary-dark', '--secondary');
+  } = useCSSCustomProperties('--h5w-line--curveColor', '--h5w-line--auxColor');
 
   return (
     <figure ref={rootRef} className={styles.root} aria-labelledby="vis-title">
@@ -104,7 +104,7 @@ function LineVis(props: Props) {
           ordinates={dataArray.data as number[]}
           errors={errorsArray && (errorsArray.data as number[])}
           showErrors={showErrors}
-          color={curveColor}
+          color={curveColor || 'midnightblue'}
           curveType={curveType}
         />
         {auxArrays.map((array, i) => (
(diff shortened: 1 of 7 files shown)


src/h5web/vis-packs/core/line/LineVis.tsx
```diff
@@ -38,7 +38,7 @@ function LineVis(props: Props) {
     ordinateLabel,
     title,
     errorsArray,
-    showErrors,
+    showErrors = false,
```
axelboc, line 41: Random change here, just to be a bit more explicit and consistent with `showGrid`.

src/h5web/vis-packs/core/line/LineVis.tsx
```diff
@@ -104,15 +104,15 @@ function LineVis(props: Props) {
           ordinates={dataArray.data as number[]}
           errors={errorsArray && (errorsArray.data as number[])}
           showErrors={showErrors}
-          color={curveColor}
+          color={curveColor || 'midnightblue'}
```
axelboc, line 107: I cannot provide the default values in CSS, because these colours are not assigned to actual CSS properties. If I define `--h5w-line--curveColor` in `LineVis.module.css`, users won't be able to override it from further up the DOM tree.

For all the other variables, I'll be able to do something like `stroke: var(--h5web-axis--tickColor, mydefaultcolor)` in CSS directly.

src/packages/app.ts
```diff
@@ -1,9 +1,6 @@
 import 'react-reflex/styles.css';
 import 'normalize.css';
-import '../styles/base.css';
 import '../styles/index.css';
-import '../styles/utils.css';
-import '../styles/vars.css';
```
axelboc, line 6: `index.css` already imports all these partials.

src/stories/Customization.stories.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+## Customization
+
+Some design aspects of the visualizations, such as the font-size of the tick labels, the color of the title, the color of the axes, etc. are **controlled by CSS**.
+We provide sane defaults that should be sufficient to start with, but at some point
+you might like to tweak a few things so the visualizations integrate better with your app's design.
+
+### Inherited font properties
+
+Font properties (`font-size`, `font-family`, etc.) of textual elements (tick labels, plot title, etc.) are **inherited by default**, so tweaking them all at once is easy.
```
loichuder, line 13: What properties are inherited in addition to `font-size` and `font-family` ?

loichuder (approved):
Looks very good :sparkles: 

Perhaps, you should also address https://github.com/silx-kit/h5web/issues/644 in this or in one of the PR that will follow

src/stories/Customization.stories.mdx
axelboc: `color`, `line-height`, `font-size`, `font-weight` ...

axelboc: https://stackoverflow.com/a/5612360/758806

axelboc: I might just add `color`.

axelboc: I've added `font-weight` for good measure.

PR #666, opened 2021-05-19
Implement dark mode with CSS invert filter

axelboc:
image

image


src/h5web/vis-packs/core/heatmap/ColorBar.module.css
```diff
@@ -21,9 +21,15 @@
...
 
-.gradient {
+.gradientBar {
   grid-area: gradient;
+  display: flex;
   border: 1px solid var(--black);
+}
+
+.gradient {
+  composes: no-invert from global;
```
axelboc, line 31: I had move the gradient to a child element so as to not invert the border of the gradient bar.

src/h5web/vis-packs/core/shared/VisCanvas.module.css
```diff
@@ -13,6 +13,7 @@
 
 /* R3F's root element */
 .canvasWrapper {
+  composes: no-invert-canvas from global;
```
axelboc, line 16: R3F doesn't provide a way to add a class to the canvas, and composing a class from an element selector like `.canvasWrapper > canvas { composes: no-invert from global; }` is technically not possible.

src/styles/base.css
```diff
@@ -62,7 +62,27 @@ abbr[title] {
   *,
   *::before,
   *::after {
-    transition: none !important;
-    animation: none !important;
+    animation-duration: 0.01ms !important;
+    animation-iteration-count: 1 !important;
+    transition-duration: 0.01ms !important;
```
axelboc, line 67: This reset code has been going around. It avoids breaking some types of transition/animation code. Don't think it matters for us, but I thought we might as well be safe.

axelboc:
Ah hang on, I'll add an E2E test.

axelboc:
Hmm, couldn't find an easy way to tell Cypress to switch to dark mode... Oh well, not a big deal.

src/styles/index.css
```diff
@@ -1,3 +1,2 @@
 @import 'vars.css';
 @import 'base.css';
-@import 'utils.css';
```
axelboc, line 3: I've merged the content of `utils.css` into `base.css`. We had reflex overrides in there that weren't `utils` anyway, and in the dark mode query, I use a couple of utility classes, so it doesn't seem worth it to try to have two separate files.

loichuder (approved):
:sparkles: :new_moon_with_face:

PR #671, opened 2021-05-20
Support providing multiple auxiliary curve colors

axelboc:
`--h5w-line-auxColor` now accepts a comma-separated list of colours. If there are more auxiliary curves than colours, then the extra curves receive the last colour in the list.

Here is what the new default auxiliary colors look like:

image

They are similar to Matplotlib's colors: https://matplotlib.org/stable/_images/dflt_style_changes-1.png.


loichuder (approved):
:star_struck:

PR #673, opened 2021-05-31
Replace lodash-es with lodash

axelboc:
Applying some of the things I learned while working on Braggy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,8 @@ node_modules/
 /storybook-static/
 /packages/**/dist/
 
-/.env.*
+/.env.local
+/.env.*.local
 
 npm-debug.log*
 *.log
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,3 @@
 /build/
 /storybook-static/
 /packages/**/dist/
-/package-lock.json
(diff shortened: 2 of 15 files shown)


.gitignore
```diff
@@ -10,7 +10,8 @@ node_modules/
 /storybook-static/
 /packages/**/dist/
 
-/.env.*
+/.env.local
+/.env.*.local
```
axelboc, line 14: Not technically needed, but more accurate: ignores `.env.test.local` but not `.env.test`, for instance.

.prettierignore
```diff
@@ -1,4 +1,3 @@
 /build/
 /storybook-static/
 /packages/**/dist/
-/package-lock.json
```
axelboc, line 4: Ignored automatically.

package.json
```diff
@@ -43,7 +43,7 @@
     "d3-format": "^3.0.0",
     "d3-scale": "^3.3.0",
     "d3-scale-chromatic": "^2.0.0",
-    "lodash-es": "^4.17.21",
+    "lodash": "^4.17.21",
```
axelboc, line 46: I double-checked: the size of lodash in the output bundle stays exactly the same.

axelboc:
Hmm, it seems the tests now fail when importing `d3-format`... 🤷 Investigating now.

loichuder (approved):

PR #675, opened 2021-06-02
Add a provider for h5core example app

loichuder:


src/demo-app/H5CoreApp.tsx
```diff
@@ -0,0 +1,19 @@
+import { useLocation } from 'react-router-dom';
+import { App } from '../packages/app';
+import H5CoreProvider from '../h5web/providers/h5core/H5CoreProvider';
```
loichuder, line 3: For now, the `H5CoreProvider` is not exported in `packages/app`

src/h5web/providers/h5core/h5core-api.ts
```diff
@@ -0,0 +1,151 @@
...
+    );
+    return data;
+  }
+
+  protected async fetchMetadata(path: string): Promise<H5CoreMetaResponse> {
+    const { data } = await this.client.get<H5CoreMetaResponse>(
+      `/meta/${this.filepath}?path=${path}`
+    );
+    return data;
+  }
```
loichuder, line 68: The fetch functions are the same as in `JupyterApi` except that the url were adapted to h5core and that we don't have a content endpoint in h5core (`children` are returned by the meta endpoint)

src/h5web/providers/h5core/models.ts
```diff
@@ -0,0 +1,45 @@
+import type { EntityKind } from '../models';
+
+export type H5CoreAttrResponse = Record<string, unknown>;
+
+export type H5CoreDataResponse = unknown;
+
+export interface H5CoreMetaResponse {
```
loichuder, line 7: If I remember correctly our discussions, it is easier for TS to have a base interface like `H5CoreMetaResponse` like this with extensions rather than an union type `H5CoreMetaResponse = H5CoreDatasetMetaReponse | H5CoreGroupMetaReponse | ...`. 

Could you confirm that this is right ?

src/h5web/providers/jupyter/jupyter-api.ts
```diff
@@ -44,7 +44,7 @@ export class JupyterStableApi extends ProviderApi {
     ]);
 
     assertDataset(entity);
-    const flatValue = this.flattenValue(value, entity, selection);
+    const flatValue = flattenValue(value, entity, selection);
```
loichuder, line 47: Moved in `../utils` to be able to use it in `H5CoreApi`.

src/h5web/providers/jupyter/utils.test.ts
```diff
@@ -1,81 +1,5 @@
...
-      class: DTypeClass.Float,
-      size: 64,
-      endianness: undefined,
-    });
-  });
-
-  it('should handle unknown type', () => {
-    expect(convertDtype('>notAType')).toEqual({ class: DTypeClass.Unknown });
-  });
-});
```
loichuder, line 78: Test moved in `../utils.test.ts`

src/h5web/providers/jupyter/utils.ts
```diff
@@ -40,77 +28,6 @@ export function assertGroupContent(
...
-        length,
-      };
-
-    case 'O':
-      return { class: DTypeClass.String, charSet: 'UTF-8' };
-
-    default:
-      return { class: DTypeClass.Unknown };
-  }
-}
```
loichuder, line 112: Moved in `../utils.ts` for the same reason as `flattenValue`: use it in `H5CoreApi`

src/demo-app/DemoApp.tsx
```diff
@@ -26,6 +27,9 @@ function DemoApp() {
         <Route path="/hsds">
           <HsdsApp />
         </Route>
+        <Route path="/h5core">
+          <H5CoreApp />
+        </Route>
```
loichuder, line 32: A new route. We could make it the default at some point :shrug:

src/h5web/providers/h5core/h5core-api.ts
axelboc: nice ^^

src/h5web/providers/h5core/models.ts
axelboc: It's better for type-checking performance, but it requires the use of typeguards to narrow down the types, which is fine here. With type unions, if all types are mutually exclusive in some way (e.g. an enum property), then TypeScript can narrow down the type with a simple `if` or `switch`.

src/h5web/providers/h5core/models.ts
```diff
@@ -0,0 +1,47 @@
...
+
+interface H5CoreAttrMetadata {
+  dtype: string;
+  name: string;
+  shape: number[];
+}
+
+interface H5CoreMetaWithAttrResponse extends H5CoreMetaResponse {
+  attributes: H5CoreAttrMetadata[];
+}
```
axelboc, line 36: I feel like this type isn't particularly needed.

src/demo-app/DemoApp.tsx
axelboc: Maybe when it's ready to be exported in the package?

axelboc (approved):
awesome 😊

PR #676, opened 2021-06-03
Remove global utility classes

axelboc:
Global utility classes are problematic when importing components in other apps. They require importing global CSS, which isn't ideal in terms of namespacing and developer experience.

We had four global utility classes: `btn-clean`, `btn-link`, `no-invert` and `no-invert-canvas`. For the first two, I move them into a CSS module: `src/h5web/utils.module.css` so they become locally scoped. For the last two, since `jupyterlab-h5web` has to be able to write selectors targeting them, they cannot be locally scoped -- so instead of moving them into a CSS module, I replace them with `data-` attributes: `data-keep-colors` and `data-keep-canvas-colors`.

### DIFF
--- a/packages/webpack.config.js
+++ b/packages/webpack.config.js
@@ -83,6 +83,12 @@ function getConfig(pkg) {
         ],
       }),
     ],
+    stats: {
+      excludeAssets: /\.d\.ts$/u, // hide emitted type definitions
+      modules: false, // don't log individual modules
+      children: false, // hide `MiniCssExtractPlugin` logs
+    },
+    performance: { hints: false }, // hide max asset size warnings
   };
 }
 
--- a/src/h5web/vis-packs/core/shared/VisCanvas.module.css
+++ b/src/h5web/vis-packs/core/shared/VisCanvas.module.css
@@ -13,7 +13,6 @@
 
 /* R3F's root element */
 .canvasWrapper {
-  composes: no-invert-canvas from global;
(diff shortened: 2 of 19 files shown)


packages/webpack.config.js
```diff
@@ -83,6 +83,12 @@ function getConfig(pkg) {
         ],
       }),
     ],
+    stats: {
+      excludeAssets: /\.d\.ts$/u, // hide emitted type definitions
+      modules: false, // don't log individual modules
+      children: false, // hide `MiniCssExtractPlugin` logs
+    },
+    performance: { hints: false }, // hide max asset size warnings
```
axelboc, line 91: Just reducing the webpack logs when building packages so we can see the important information: the size of the main JS bundles.

src/h5web/vis-packs/core/shared/VisCanvas.module.css
```diff
@@ -13,7 +13,6 @@
 
 /* R3F's root element */
 .canvasWrapper {
-  composes: no-invert-canvas from global;
```
axelboc, line 16: This was actually inverting back the line vis as well, which I don't think we want. Now the `data-keep-canvas-colors` is set on the `HeatmapVis`' wrapping `figure` element.

loichuder (approved):
Great !

Can you simply add **CSS** to the commit message (like `Remove global CSS utility classes`) ? I had a doubt when seeing the PR name :joy:

PR #677, opened 2021-06-03
Cancel bound editors when closing domain tooltip

axelboc:
This PR fixes the following bug: 

1. Start editing min or max bound in domain tooltip.
2. Don't apply the edit -- instead, close the tooltip with Escape or by clicking outside of it.
3. Re-open the tooltip 👉 the field is no longer in edit mode and the edited value isn't applied, yet the field shows the edited value.

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
@@ -1,4 +1,10 @@
-import { useEffect, useRef, useState } from 'react';
+import {
+  forwardRef,
+  useEffect,
+  useImperativeHandle,
+  useRef,
+  useState,
+} from 'react';
 import { FiCheck, FiSlash } from 'react-icons/fi';
 import { formatPreciseValue } from '../../../utils';
 import type { Bound } from '../../../vis-packs/core/models';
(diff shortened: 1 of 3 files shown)


src/h5web/toolbar/controls/DomainSlider/DomainTooltip.tsx
```diff
@@ -39,6 +40,20 @@ function DomainTooltip(props: Props) {
...
 
   const { minGreater, minError, maxError } = errors;
+  const minEditorRef = useRef<BoundEditorHandle>(null);
+  const maxEditorRef = useRef<BoundEditorHandle>(null);
+
+  useEffect(() => {
+    if (!isEditingMin) {
+      minEditorRef.current?.cancel();
+    }
+  }, [isEditingMin]);
```
loichuder, line 50: Add a small comment perhaps ? (as above `useImperativeHandle`)

src/h5web/toolbar/controls/DomainSlider/BoundEditor.tsx
```diff
@@ -14,14 +20,27 @@ interface Props {
...
   const [inputValue, setInputValue] = useState('');
 
+  function cancel() {
+    onEditToggle(false);
+    setInputValue(formatPreciseValue(value));
+  }
+
+  /* Expose `cancel` function to parent component through ref handle so that
+    `inputValue` can be reset when the user closes the domain tooltip. */
+  useImperativeHandle(ref, () => ({ cancel }));
```
axelboc, line 42: First time I use this hook. The reason is as follows:

When the user closes the domain tooltip with Escape or by clicking outside of it, we want to cancel any edits made to the min/max fields. This means turning off edit mode (which is done in `DomainSlider`, but also resetting the `inputValue` here, in `BoundEditor`. Since the slider and tooltip components don't have access to the `inputValue` state, they can't reset it themselves. `useImperativeHandle` allows exposing the `cancel` function to the parent component via a forwarded `ref`: the parent component can then set a `ref` on each bound editor and call `ref.current.cancel()`.

loichuder (approved):
:fishing_pole_and_fish:

loichuder (approved):

loichuder (approved):
:octopus:

PR #678, opened 2021-06-03
Add toolbar components to lib package and document

axelboc:
- Add `Toolbar` and the main controls to `@h5web/lib`, as well as a couple more hooks and models.
- Create `GridToggler` control and expose to `@h5web/lib`.
- Document toolbar usage in storybook.
- Rename prop `onChange` of `ToggleBtn` to `onToggle`.

### DIFF
--- a/src/h5web/toolbar/controls/ToggleBtn.tsx
+++ b/src/h5web/toolbar/controls/ToggleBtn.tsx
@@ -9,7 +9,7 @@ interface Props extends AriaAttributes {
   small?: boolean;
   raised?: boolean;
   value: boolean;
-  onChange: () => void;
+  onToggle: () => void;
   disabled?: boolean;
 }
 
@@ -21,7 +21,7 @@ function ToggleBtn(props: Props) {
     small,
     raised,
     value,
-    onChange,
+    onToggle,
     disabled,
     ...ariaAttrs
   } = props;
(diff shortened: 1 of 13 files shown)


src/h5web/toolbar/controls/ToggleBtn.tsx
```diff
@@ -30,7 +30,7 @@ function ToggleBtn(props: Props) {
     <button
       className={styles.btn}
       type="button"
-      onClick={onChange}
+      onClick={() => onToggle()}
```
loichuder, line 33: What is the difference between:
```
onClick={() => onToggle()}
```
and
```
onClick={onToggle}
``` 
?

src/stories/GettingStarted.stories.mdx
```diff
@@ -7,10 +7,11 @@
 
 The **H5Web Component Library** provides data visualization components and utilities from H5Web for use in other front-end web applications.
 
-The components are organised in two categories:
+The components are organised in three categories:
 
 - 🎨 **Visualizations**: the top-level visualization components — i.e. the components you will most likely need.
 - 🧱 **Building Blocks**: the low-level components used by the visualization components — for advanced uses only.
+- 🧰 **Toolbar**: the responsive toolbar and toolbar controls.
```
loichuder, line 14: Is the emoji a toolbox ? :joy:

loichuder (approved):
We might want to build new code sandboxes with the toolbar

src/h5web/toolbar/controls/ToggleBtn.tsx
axelboc: In the second case, `onToggle` receives the `MouseEvent` as first argument, which means that if `onToggle` declares a parameter `force` and then does something with it, like `if (!!force)`, then it will pass the condition even though what we wanted was to call `onToggle()` without passing a `force` argument.

PR #679, opened 2021-06-03
Implement layout that take all available space while keeping aspect ratio

### LINKED ISSUE #674: Change handling of heatmap keep aspect ratio
As it is now, aspect ratio is kept at the plot area level (i.e, the plot area is a square).

It would be best to keep the aspect ratio of the displayed image inside the plot rather than at the plot area level.
This would allow the plot to use more space to display the data when zooming-in the image.

Some screenshots to give an idea:

Zoomed-out:
keep-ratio-zoom-out

Zooming-in:
keep-ratio-zoom-half

Zoomed-in:
keep-ratio-zoom-in

loichuder:
Fix #674 

image
image

From my point of view, this issue must be solved in three steps:
1. We want to keep the previous behaviour available.
2. `Mesh` dimensions and positions are based on the postulate that it will take all the space available in the `VisCanvas`. This needs to change so that `Mesh` is correctly scaled and positioned whatever the canvas limits.
3. Considering that `Mesh` is correctly scaled and positioned, to get the behaviour in #674, the canvas limits must be computed so that it keeps the aspect ratio of the image.

Each sub-issue is addressed by a separate commit (that might be easier to review separately):
1. The first commit changes the `keepAspectRatio` on/off state into a `Layout` that can hold three values: 
  - `tight` (the previous `keepAspectRatio=true`)
  - `fill` (the previous `keepAspectRatio=false`)
  - `loose` (the new layout described in #674)
2. The second adds a component `ScalingMesh` that takes care of positioning and scaling the `Mesh` thanks to the canvas scales.
3. The third changes `VisCanvas` so that the domain of abscissas and ordinates are adjusted to keep the right ratio. This is the commit that might need more discussion as `VisCanvas` is no longer "dumb" and can change the domains it gets in props. Unfortunately, adjusting the domains necessitates access to `availableSize` which lives in the `VisCanvas`...


src/h5web/vis-packs/core/heatmap/ScalingMesh.tsx
```diff
@@ -0,0 +1,34 @@
...
+interface Props extends MeshProps {
+  minAbscissa: number;
+  maxAbscissa: number;
+  minOrdinate: number;
+  maxOrdinate: number;
+}
+
+function ScalingMesh(props: Props) {
+  const { minAbscissa, maxAbscissa, minOrdinate, maxOrdinate, ...meshProps } =
+    props;
```
loichuder, line 13: `ScalingMesh` only needs the limit coordinates and forward the rest of the props to `Mesh`. 

We could also make use of WebGL to scale a new `mesh` and have `Mesh` a children of this new `mesh` so that it is relatively positioned.

src/h5web/vis-packs/core/heatmap/ScalingMesh.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const { abscissaScale, ordinateScale } = useCanvasScales();
+
+  const left = abscissaScale(minAbscissa);
+  const right = abscissaScale(maxAbscissa);
+  const x = abscissaScale(minAbscissa + (maxAbscissa - minAbscissa) / 2);
+
+  const bottom = ordinateScale(minOrdinate);
+  const top = ordinateScale(maxOrdinate);
+  const y = ordinateScale(minOrdinate + (maxOrdinate - minOrdinate) / 2);
```
loichuder, line 22: I make use of the canvasScales to translate the limits coordinates (and the center) in the canvas coordinate system.

src/h5web/vis-packs/core/shared/VisCanvas.tsx
```diff
@@ -3,23 +3,25 @@ import { Canvas } from '@react-three/fiber';
...
 import AxisSystemContext from './AxisSystemContext';
+import type { Layout } from '../heatmap/models';
 
 const AXIS_OFFSETS = { vertical: 72, horizontal: 40, fallback: 16 };
 
 interface Props {
   abscissaConfig: AxisConfig;
   ordinateConfig: AxisConfig;
+  layout?: Layout;
   aspectRatio?: number;
```
loichuder, line 17: `aspectRatio` is only used if `layout` is `tight` which is not great. I think `VisCanvas` will need a slight refactor

src/h5web/vis-packs/core/shared/VisCanvas.tsx
```diff
@@ -3,23 +3,25 @@ import { Canvas } from '@react-three/fiber';
...
   canvasTitle?: string;
   children: ReactNode;
 }
 
 function VisCanvas(props: Props) {
-  const { abscissaConfig, ordinateConfig, aspectRatio, canvasTitle, children } =
-    props;
+  const { aspectRatio, canvasTitle, children, layout = 'fill' } = props;
+  let { abscissaConfig, ordinateConfig } = props;
```
loichuder, line 24: I choose the most direct way for now for discussion by mutating the config domains if the `layout` is `loose`.

This ought to be changed once we agree on the roles of each component.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -365,3 +365,30 @@ export function getSliceSelection(
...
+
+export function computeSameRatioDomain(
+  availableSize: Size,
+  abscissaDomain: Domain,
+  ordinateDomain: Domain
+): [Domain, Domain] {
+  const { width, height } = availableSize;
+
+  const abscissaRatio = width / (abscissaDomain[1] - abscissaDomain[0]);
+  const ordinateRatio = height / (ordinateDomain[1] - ordinateDomain[0]);
```
loichuder, line 377: The `ratios` here are the number of pixels per coordinate.

Keeping the aspect ratio means that going from 0 to 1 on the abscissa and on the ordinate must span the same number of pixels. This is, both ratios must be equal.

To make the figure fit in the resulting domains, we need to take the **smaller** ratio as common value (i.e. increase the coordinates).

src/h5web/vis-packs/core/utils.ts
```diff
@@ -365,3 +365,30 @@ export function getSliceSelection(
...
+
+  if (abscissaRatio > ordinateRatio) {
+    const x0 =
+      (abscissaDomain[0] + abscissaDomain[1] - width / ordinateRatio) / 2;
```
loichuder, line 381: This formula was derived from solving the system of 2 equations gotten from the following conditions:
- the ratios must be equal
- the center of the canvas must stay the same

src/h5web/vis-packs/core/utils.ts
```diff
@@ -365,3 +365,30 @@ export function getSliceSelection(
...
+  }
+
+  if (abscissaRatio < ordinateRatio) {
+    const y0 =
+      (ordinateDomain[0] + ordinateDomain[1] - height / abscissaRatio) / 2;
+    const y1 = y0 + height / abscissaRatio;
+    return [abscissaDomain, [y0, y1]];
+  }
+
+  return [abscissaDomain, ordinateDomain];
```
loichuder, line 393: If both ratios are already equal, all is well and we simply return the domains :shrug:

loichuder (commented):
Names of new types/layout/components are open to changes btw

src/h5web/vis-packs/core/utils.ts
```diff
@@ -365,3 +365,38 @@ export function getSliceSelection(
...
+  const { domain: abscissaDomain } = abscissaConfig;
+  const { domain: ordinateDomain } = ordinateConfig;
+
+  const abscissaRatio = width / (abscissaDomain[1] - abscissaDomain[0]);
+  const ordinateRatio = height / (ordinateDomain[1] - ordinateDomain[0]);
+
+  if (abscissaRatio > ordinateRatio) {
+    const x0 =
+      (abscissaDomain[0] + abscissaDomain[1] - width / ordinateRatio) / 2;
+    const x1 = x0 + availableSize.width / ordinateRatio;
```
axelboc, line 384: just `width`

loichuder:
@axelboc I am not satisfied of the result but I think I have work too much on it and without enough caffeine to continue doing something relevant so I hope your review will allow me to move forward

You can look at [the changes of the last 3 commits](https://github.com/silx-kit/h5web/pull/679/files/11507053e2ff13d505c1c881d4aee08b0b120581..b73828d20fde72c027fa5e46a63779c11bd16002) for easier review

src/h5web/vis-packs/core/heatmap/HeatmapAxisSystem.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+function HeatmapAxisSystem(props: Props) {
+  const {
+    axisOffsets,
+    abscissaConfig,
+    ordinateConfig,
+    imageRatio,
+    title,
+    children,
+  } = props;
```
axelboc, line 24: To save a bit of space, you could spread the props that get forwarded in `...forwardedProps` (`children` `axisOffsets` and `title`), and then return:

```tsx
<AxisSystem
  abscissaConfig={abscissaConfig}
  ordinateConfig={ordinateConfig}
  {...forwardedProps}
/>
```

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -34,7 +36,7 @@ function HeatmapVis(props: Props) {
     domain = DEFAULT_DOMAIN,
     colorMap = 'Viridis',
     scaleType = ScaleType.Linear,
-    keepAspectRatio = true,
+    layout = 'fill',
```
axelboc, line 39: I wonder whether the default shouldn't rather be `cover`.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -365,3 +368,39 @@ export function getSliceSelection(
...
+  }
+
+  const y0 = (ordinateDomain[1] - ordinateDomain[0]) / 2;
+  return [
+    abscissaDomain,
+    ordinateDomain.map(
+      (y) => ((y - y0) * imageRatio) / canvasRatio + y0
+    ) as Domain,
+  ];
+}
```
axelboc, line 393: This algorithm is a bit challenging to understand: maybe rename `x0`, `y0`, `newAbscissaDomain`, etc., and add a few comments?

src/stories/VisCanvas.stories.tsx
```diff
@@ -1,8 +1,21 @@
...
-const Template: Story<VisCanvasProps> = (args) => <VisCanvas {...args} />;
+import { VisCanvas, AxisSystemProps, ScaleType } from '../packages/lib';
+import AxisSystem from '../h5web/vis-packs/core/shared/AxisSystem';
+import { getAxisOffsets } from '../h5web/vis-packs/core/utils';
+
+const Template: Story<{ aspectRatio?: number } & AxisSystemProps> = ({
+  aspectRatio,
+  ...args
+}) => {
+  const axisOffsets = getAxisOffsets({});
```
axelboc, line 11: You could make the argument of `getAxisOffsets` optional with `= {}`

src/stories/VisCanvas.stories.tsx
```diff
@@ -60,7 +73,7 @@ export const GraphTitle = Template.bind({});
 GraphTitle.args = {
   abscissaConfig: { domain: [0, 3], showGrid: true, isIndexAxis: true },
   ordinateConfig: { domain: [50, 100], showGrid: true, isIndexAxis: true },
-  canvasTitle: 'This is a graph',
+  title: 'This is a graph',
 };
```
axelboc, line 77: Maybe also add a story for `fill`? (assuming `cover` becomes the new default)

axelboc (approved):
Awesome! This feels right, doesn't it?

src/stories/HeatmapVis.stories.tsx
```diff
@@ -176,8 +176,12 @@ export default {
...
       },
     },
-    keepAspectRatio: {
-      defaultValue: true,
+    layout: {
+      defaultValue: 'fill',
+      control: {
+        type: 'inline-radio',
+        options: ['contain', 'cover', 'fill'],
+      },
```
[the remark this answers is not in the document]
loichuder: I added stories for `fill` and `container` as `cover` is the default

src/stories/VisCanvas.stories.tsx
loichuder: Well, `layout` is not supposed to be seen by `VisCanvas`...

axelboc: Ah of course, sorry 😄

PR #680, opened 2021-06-04
Let toolbar and controls be customized via CSS

axelboc:
Same idea: the toolbar and controls no longer reference H5Web's theme color variables but only namespaced, semantic variables with simple fallbacks values.

I purposely haven't tried to make every aspect of the toolbar customizable. For instance, I haven't provided a variable to set the background colour of buttons when idle (which would otherwise default to `transparent`). This is to ensure that a certain design coherence is maintained, as well as to help users make the customizations that really matter.

Here is what the toolbar looks like when no variables are provided:

image


src/h5web/toolbar/ComplexToolbar.tsx
```diff
@@ -55,6 +55,18 @@ function ComplexToolbar() {
...
 
+      <Selector
+        value={visType}
+        onChange={(value: ComplexVisType) => setVisType(value)}
+        options={Object.values(ComplexVisType)}
+        optionComponent={({ option }) => (
+          // eslint-disable-next-line react/jsx-no-useless-fragment
+          <>{`${VIS_TYPE_SYMBOLS[option]} ${option}`}</>
+        )}
+      />
```
axelboc, line 66: Just moving this up before the ratio/grid toggles. I feel like we had discussed doing that before.

src/h5web/toolbar/Toolbar.module.css
```diff
@@ -2,7 +2,9 @@
   flex: 1 1 0%;
   display: flex;
   min-width: 0;
+  height: 2.5rem;
   padding: 0 0.5rem;
+  background-color: var(--h5w-toolbar--bgColor, aliceblue);
```
axelboc, line 7: I set a height and bg color here, even though we don't need them in H5Web, so that the toolbar can be used on its own in Braggy, without a wrapper element.

src/h5web/toolbar/Toolbar.module.css
```diff
@@ -64,8 +66,7 @@
 }
 
 .btn[data-raised] > .btnLike {
-  background-color: var(--btn-bg-raised);
```
axelboc, line 67: "raised" buttons are used for min/max auto-scaling in the domain slider's tooltip. I give them a slightly bigger shadow and remove their background, which allows me to use them in toggle groups to avoid having another button variant (which would mean more CSS variables).

src/h5web/toolbar/Toolbar.module.css
```diff
@@ -91,37 +92,32 @@
...
 
 .btn:hover > .btnLike {
-  background-color: var(--secondary-lighter);
-  box-shadow: var(--btn-shadow-idle) var(--btn-shadow-color);
-}
-
-.btn:active > .btnLike {
-  box-shadow: var(--btn-shadow-pressed) var(--btn-shadow-color);
+  background-color: var(--h5w-btn-hover--bgColor, whitesmoke);
+  box-shadow: -1px -1px 2px inset var(--h5w-btn-hover--shadowColor, var(--h5w-btnRaised--shadowColor, gray));
```
axelboc, line 96: To maintain the design's coherence, the offsets and blur radii of all the box shadows are now hardcoded (and duplicated if needed).

src/h5web/toolbar/Toolbar.module.css
```diff
@@ -132,7 +128,10 @@
 }
 
 .popupInner {
-  background-color: var(--secondary-light-bg-90);
+  background-color: var(
+    --h5w-domainSlider-tooltip--bgColor,
+    rgba(255, 255, 255, 0.9)
```
axelboc, line 133: Switched to white for consistency with overflow and selector menus.

src/h5web/toolbar/controls/ScaleSelector/ScaleSelector.tsx
```diff
@@ -1,27 +1,24 @@
...
-      {label && <span className={styles.label}>{label}</span>}
-      <Selector
-        value={value}
-        onChange={onScaleChange}
-        options={Object.values(ScaleType)}
-        optionComponent={ScaleOption}
-      />
-    </div>
+    <Selector
+      label={label}
```
axelboc, line 16: Moving the label feature to `Selector`, as it seems to make sense and allows removing `ScaleSelector.module.css`.

src/h5web/toolbar/controls/ToggleGroup.module.css
```diff
@@ -16,20 +16,13 @@
...
 
-.btn[aria-checked='false'] > .btnLike {
-  background-color: var(--secondary-dark-15);
-  box-shadow: var(--btn-shadow-idle) var(--btn-shadow-color);
-}
-
-.btn[aria-checked='false']:hover > .btnLike {
-  box-shadow: var(--btn-shadow-idle) var(--btn-shadow-dark-color);
-}
-
```
axelboc, line 32: The buttons are now "raised" buttons, which are styled in `Toolbar.module.css`.

src/h5web/vis-packs/core/shared/TooltipMesh.module.css
```diff
@@ -1,6 +1,6 @@
 .tooltip {
   padding: 0.25rem 0.5rem;
-  background-color: var(--h5w-tooltip--bgColor, whitesmoke);
+  background-color: var(--h5w-tooltip--bgColor, rgba(255, 255, 255, 0.9));
```
axelboc, line 3: Making the background of the tooltip slightly transparent like the domain slider tooltip.

src/stories/Toolbar.stories.tsx
```diff
@@ -39,15 +39,7 @@ const Template: Story<{ narrow?: boolean }> = ({ narrow }) => {
...
-      <div
-        style={{
-          display: 'flex',
-          maxWidth: narrow ? '30rem' : undefined,
-          height: '2.5rem',
-          marginLeft: 'auto',
-          backgroundColor: 'var(--primary-bg)',
-        }}
-      >
+      <div style={narrow ? { maxWidth: '30rem', marginLeft: 'auto' } : {}}>
```
axelboc, line 42: The toolbar no longer needs a wrapper element. (This one is now just for the responsive toolbar story).

src/styles/vars.css
```diff
@@ -1,17 +1,9 @@
 :root {
-  --black: #020402;
-  --near-black: #333;
-  --dark-slate-gray: #2f4f4f;
-  --dark-gray: #696969;
-  --gray: #808080;
-  --light-gray: #d6d6d6;
-  --white: #fff;
```
axelboc, line 8: yay!!

src/h5web/toolbar/Toolbar.module.css
```diff
@@ -91,37 +92,32 @@
...
 
-.btn[aria-pressed='true'] > .btnLike,
-.btn[aria-checked='true'] > .btnLike {
-  background-color: var(--secondary-light);
-  box-shadow: var(--btn-shadow-pressed) var(--btn-shadow-dark-color);
+.btn:active > .btnLike,
+.btn[data-raised]:active > .btnLike,
+.btn[aria-pressed='true'] > .btnLike, /* toggle buttons, domain slider "edit" button */
+.btn[aria-checked='true'] > .btnLike, /* toggle group buttons */
+.btn[aria-haspopup][aria-expanded='true'] > .btnLike  /* selectors and overflow menu buttons */ {
```
axelboc, line 107: Adding `[aria-haspopup]` to the "expanded" selector so that the domain slider's "edit" button doesn't appear as pressed when hovered (as hovering it expands the domain slider's tooltip). Contrary to the overflow menu button and the buttons of the drop-down selectors, the "edit" button doesn't have `aria-haspopup`; only `aria-controls`.

src/h5web/visualizer/Visualizer.module.css
```diff
@@ -27,7 +27,7 @@
 
 .error {
   padding: 1em;
-  color: var(--danger);
+  color: brown;
```
loichuder, line 30: Should it be `var(--h5web-error--color, brown)` ?

src/stories/Customization.stories.mdx
```diff
@@ -48,41 +48,41 @@ as you see fit. For instance, if you'd like to change the color of the curve of
...
+| `--h5w-axisLabels--fontWeight` | `600`                      | Font weight of axis labels  |
+| `--h5w-grid--color`            | `gray`                     | Color of grid lines         |
+| `--h5w-grid--opacity`          | `0.33`                     | Opacity of grid lines       |
+| `--h5w-tooltip--color`         | `inherit`                  | Text color of tooltip       |
+| `--h5w-tooltip--bgColor`       | `rgba(255, 255, 255, 0.9)` | Background color of tooltip |
+| `--h5w-tooltip--fontFamily`    | `inherit`                  | Font family of tooltip      |
+| `--h5w-tooltip--fontSize`      | `0.75em`                   | Font size of tooltip        |
+| `--h5w-tooltip-guide--color`   | `gray`                     | Color of tooltip guide      |
+| `--h5w-tooltip-guide--width`   | `1.5`                      | Width of tooltip guide      |
+| `--h5w-tooltip-guide--opacity` | `0.5`                      | Opacity of tooltip guide    |
```
loichuder, line 76: I don't think I have seen `h5w-error--color` ?

loichuder (approved):

src/h5web/visualizer/Visualizer.module.css
axelboc: Well, `Visualizer` is not exposed to the lib, so there's no point in making it configurable. Since `--h5web-error--color` doesn't have a value in H5Web (because the default value works for us), then there's no point in using the `var(--h5web-error--color, brown)` syntax.

PR #682, opened 2021-06-07
Expose and document `getVisDomain` and `getSafeDomain`

axelboc:


src/stories/Utilities.stories.mdx
```diff
@@ -71,11 +71,39 @@ Generate a CSS linear gradient for a given D3 interpolator, to be used as `backg
 getLinearGradient(interpolator: D3Interpolator, direction: 'top' | 'bottom' | 'right' | 'left', minMaxOnly = false): string
 ```
 
+#### getVisDomain
+
+Determine the domain to be used for the visualization based on a user-selected custom domain. If a bound of the custom domain is `null`, it "auto-scales" by falling back to the corresponding bound of the data domain.
```
loichuder, line 76:
```suggestion
Determine the domain to be used for the visualization based on a user-selected custom domain. If a bound of the custom domain is `null`, it falls back to the corresponding bound of the data domain.
```

src/stories/Utilities.stories.mdx
```diff
@@ -71,11 +71,39 @@ Generate a CSS linear gradient for a given D3 interpolator, to be used as `backg
...
+
+```ts
+getVisDomain(customDomain: CustomDomain, dataDomain: Domain): Domain
+
+const visDomain1 = getVisDomain([null, null], [0, 100]]); // [0, 100]
```
loichuder, line 81:
```suggestion
const visDomain1 = getVisDomain([null, null], [0, 100]); // [0, 100]
```

src/stories/Utilities.stories.mdx
```diff
@@ -71,11 +71,39 @@ Generate a CSS linear gradient for a given D3 interpolator, to be used as `backg
...
+const visDomain2 = getVisDomain([null, 20], [0, 100]]); // [0, 20]
```
loichuder, line 82:
```suggestion
const visDomain2 = getVisDomain([null, 20], [0, 100]); // [0, 20]
```

src/stories/Utilities.stories.mdx
```diff
@@ -71,11 +71,39 @@ Generate a CSS linear gradient for a given D3 interpolator, to be used as `backg
...
+```
+
+#### getSafeDomain
+
+Determine a domain that is safe for the visualization. This is typically called with a user-defined `customDomain`, or with a `visDomain` as returned by `getVisDomain()`.
+If the domain is determined to be unsafe, a safe domain based on `fallbackDomain` is returned along with an error object. Note that `fallbackDomain` is assumed to be safe.
+The domain is considered unsafe notably if it's inverted (`min > max`), or if the scale is log and at least one of the bounds is lower than or equal to 0.
```
loichuder, line 89:
```suggestion
The domain is considered unsafe if it's inverted (`min > max`), or if the scale is log and at least one of the bounds is negative.
```

loichuder (approved):

PR #690, opened 2021-06-09
Turn grid on by default for Heatmap

axelboc:
Now that the axes extend around the image, it looks weird without the grid, so I think it should be visible by default.


loichuder (approved):
Agreed

PR #693, opened 2021-06-09
Upgrade deps

axelboc:
TypeScript 4.3 brings named imports autocompletion! 🎉 

image

image


src/h5web/dimension-mapper/SlicingSlider.tsx
```diff
@@ -17,19 +17,15 @@ interface Props {
 function SlicingSlider(props: Props) {
   const { dimension, slicingIndex, rawDims, mapperState, onChange } = props;
 
-  const [containerRef, { height }] = useMeasure();
+  const [containerSize, containerRef] = useMeasure<HTMLDivElement>();
```
axelboc, line 20: `@react-hookz/web`'s `useMeasure` puts the ref second and returns an `undefined` rect (whereas `react-use`'s `useMeasure` used to return `{ width: 0, height: 0, ... }`

src/h5web/dimension-mapper/SlicingSlider.tsx
```diff
@@ -17,19 +17,15 @@ interface Props {
...
 
   useEffect(() => {
     sliderRef.current?.resize();
-  }, [height]);
+  }, [containerSize]);
 
   return (
-    <div
-      key={dimension}
-      ref={containerRef as (element: HTMLElement | null) => void} // https://github.com/streamich/react-use/issues/1264
```
axelboc, line 30: ... but at least we can get rid of this workaround! 🎉

src/h5web/toolbar/controls/DomainSlider/ScaledSlider.tsx
```diff
@@ -78,7 +78,7 @@ function ScaledSlider(props: Props) {
       max={SLIDER_RANGE[1]}
       value={scaledValue}
       onBeforeChange={setBeforeChangeValue}
-      onChange={handleChange}
+      onChange={(bounds) => handleChange(bounds)}
```
axelboc, line 81: We cannot yet take advantage of https://github.com/zillow/react-slider/issues/212 because of https://github.com/zillow/react-slider/issues/229

src/h5web/dimension-mapper/SlicingSlider.tsx
loichuder: Good, I prefer the `undefined`. It feels clearer me.

src/h5web/vis-packs/core/shared/VisCanvas.tsx
```diff
@@ -14,8 +15,7 @@ interface Props {
 function VisCanvas(props: Props) {
   const { axisOffsets, children, aspectRatio } = props;
 
-  const [visAreaRef, visAreaSize] = useMeasure();
-
+  const [visAreaSize = EMPTY_RECT, visAreaRef] = useMeasure<HTMLDivElement>();
```
loichuder, line 18: Instead of using `visAreaSize`, you could add an early return here:

```js
if(!visAreaSize) {
      return (<div ref={visAreaRef} className={styles.visArea} />);
}
```

We would avoid computing `availableSize` and `visSize` for nothing below.

loichuder (approved):

src/h5web/vis-packs/core/shared/VisCanvas.tsx
axelboc: seems legit

axelboc: I removed `EMPTY_RECT` in the end (cf. `ColorBar.tsx`)

PR #695, opened 2021-06-10
Make `Html` move according to camera

loichuder:
Implements `calculatePosition` from https://github.com/pmndrs/drei/blob/v6.0.3/src/web/Html.tsx

This is a prop that can be overridden which is needed for `AxisSystem` and `Tooltip` that should not move.

### DIFF
--- a/src/h5web/vis-packs/core/shared/Html.tsx
+++ b/src/h5web/vis-packs/core/shared/Html.tsx
@@ -1,66 +1,103 @@
 import {
   useState,
-  useEffect,
   forwardRef,
   HTMLAttributes,
   useLayoutEffect,
+  useRef,
 } from 'react';
 import ReactDOM from 'react-dom';
-import { useThree } from '@react-three/fiber';
+import { GroupProps, useFrame, useThree } from '@react-three/fiber';
+import { Group, Vector3 } from 'three';
 
 // Simplified version of `drei`'s `<Html>` component
-// https://github.com/pmndrs/drei/blob/v2.2.3/src/Html.tsx
-const Html = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
-  (props, ref) => {
-    const { className, style, children, ...divProps } = props;
+// https://github.com/pmndrs/drei/blob/v6.0.3/src/web/Html.tsx
 
-    const { width, height } = useThree((state) => state.size);
-    const gl = useThree((state) => state.gl);
-    const { parentElement } = gl.domElement;
+const v1 = new Vector3();
 
-    // Container `div` for ReactDOM to render into, appended next to R3F's `canvas`
-    const [el] = useState(() => document.createElement('div'));
+export interface HtmlProps extends HTMLAttributes<HTMLDivElement> {
+  groupProps?: GroupProps;
+  followCamera?: boolean;
+}
 
-    useEffect(() => {
-      el.style.cssText = `position: absolute; top: 0; left: 0;`;
+const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
+  const {
+    className,
+    style,
+    children,
+    followCamera,
+    groupProps = {},
+    ...divProps
+  } = props;
 
-      if (parentElement) {
-        parentElement.append(el);
-      }
+  const { width, height } = useThree((state) => state.size);
+  const gl = useThree((state) => state.gl);
+  const { parentElement } = gl.domElement;
+  const camera = useThree((state) => state.camera);
 
-      return () => {
-        if (parentElement) {
-          el.remove();
-        }
+  // Container `div` for ReactDOM to render into, appended next to R3F's `canvas`
+  const [el] = useState(() => document.createElement('div'));
+  // Placeholder R3F group, the position of which is tracked and forwarded to `el` when `followCamera` is enabled
+  const group = useRef<Group>(null!); // eslint-disable-line @typescript-eslint/no-non-null-assertion
 
-        ReactDOM.unmountComponentAtNode(el);
-      };
-    }, [el, parentElement]);
+  function updateHtmlPosition() {
+    const objectPos = v1.setFromMatrixPosition(group.current.matrixWorld);
+    objectPos.project(camera);
+    const widthHalf = width / 2;
+    const heightHalf = height / 2;
+    const pos = [
+      objectPos.x * widthHalf + widthHalf,
+      -(objectPos.y * heightHalf) + heightHalf,
+    ];
 
-    useLayoutEffect(() => {
-      ReactDOM.render(
-        <div
-          ref={ref}
-          className={className}
-          style={{
-            position: 'absolute',
-            top: 0,
-            left: 0,
-            width,
-            height,
-            pointerEvents: 'none', // let pointer events pass through to canvas
-            ...style,
-          }}
-          {...divProps}
-        >
-          {children}
-        </div>,
-        el
-      );
-    }, [children, className, el, height, width, style, divProps, ref]);
-
-    return null;
+    el.style.transform = `translate3d(${pos[0]}px,${pos[1]}px,0)`;
   }
-);
+
+  useLayoutEffect(() => {
+    group.current.updateMatrixWorld(); // To place correctly `el` on first render
+    el.style.cssText = `position:absolute; top:0; left:0; transform-origin:0 0;`;
+    if (followCamera) {
+      updateHtmlPosition();
+    }
+
+    if (parentElement) {
+      parentElement.append(el);
+    }
+
+    return () => {
+      if (parentElement) {
+        el.remove();
+      }
+      ReactDOM.unmountComponentAtNode(el);
+    };
+  }, [el, parentElement]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  useLayoutEffect(() => {
+    ReactDOM.render(
+      <div
+        ref={ref}
+        className={className}
+        style={{
+          position: 'absolute',
+          top: 0,
+          left: 0,
+          pointerEvents: 'none', // let pointer events pass through to canvas
+          ...style,
+        }}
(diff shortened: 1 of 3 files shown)


src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,38 +4,82 @@ import {
...
+) {
+  const objectPos = v1.setFromMatrixPosition(el.matrixWorld);
+  objectPos.project(camera);
+  const widthHalf = size.width / 2;
+  const heightHalf = size.height / 2;
+  return [
+    objectPos.x * widthHalf + widthHalf,
+    -(objectPos.y * heightHalf) + heightHalf,
+  ];
+}
```
axelboc, line 33: If you go with the boolean prop, I would move this function inside the component to avoid passing the arguments. It would also allow you to directly set `el.style.transform` to reduce duplication between the `useEffect` and the `useFrame`. (You'll just have to set `el.style.cssTest` before calling this function.)

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,38 +4,82 @@ import {
...
     // Container `div` for ReactDOM to render into, appended next to R3F's `canvas`
     const [el] = useState(() => document.createElement('div'));
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const group = useRef<Group>(null!);
 
+    // eslint-disable-next-line consistent-return
     useEffect(() => {
-      el.style.cssText = `position: absolute; top: 0; left: 0;`;
+      if (group.current) {
+        scene.updateMatrixWorld();
```
axelboc, line 66: Do you think this is needed in our app? Same with `camera.updateMatrixWorld()` below.

axelboc (changes requested):

src/h5web/vis-packs/core/shared/Html.tsx
loichuder: Removed it

src/h5web/vis-packs/core/shared/AxisSystem.tsx
```diff
@@ -45,6 +45,7 @@ function AxisSystem(props: Props) {
     <>
       <Html
         className={styles.axisSystem}
+        fullscreen
```
loichuder, line 48: I added a `fullscreen` prop in accordance with drei's version.

When using `Html` to annotate things, we don't want to the annotating `div` to take the canvas dimensions.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,63 +4,107 @@ import {
...
 
-    const { width, height } = useThree((state) => state.size);
-    const gl = useThree((state) => state.gl);
-    const { parentElement } = gl.domElement;
+const v1 = new Vector3();
 
-    // Container `div` for ReactDOM to render into, appended next to R3F's `canvas`
-    const [el] = useState(() => document.createElement('div'));
+export interface HtmlProps extends HTMLAttributes<HTMLDivElement> {
+  groupProps?: GroupProps;
```
loichuder, line 19: I realized that `Html` must in fact be positioned as any Three.js object. `groupProps` allows to pass props to the parent `group` to set `position` and `scale` if needed

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,63 +4,107 @@ import {
...
+
+    return () => {
       if (parentElement) {
-        parentElement.append(el);
+        el.remove();
       }
+      ReactDOM.unmountComponentAtNode(el);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [el, parentElement]);
```
loichuder, line 75: Wow the diff is really all other the place here. 

I changed the `useEffect` to include the part that set the `cssText` and calls `updateHtmlPosition` but nothing was changed in the clean-up function.

There is also the addition of `// eslint-disable-next-line react-hooks/exhaustive-deps`

src/h5web/vis-packs/core/shared/AxisSystem.tsx
axelboc: I'm wondering whether the fullscreen behaviour should depend on `followCamera`. It seems the two props are mutually exclusive.

loichuder: > It seems the two props are mutually exclusive.

For our case yesy. But take the Braggy radial grid for example: you want it to take the whole heatmap while still following the camera.

loichuder: On second thought, the prop is perhaps not really adapted in our case. I will remove it and let the caller set the width/height.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,63 +4,107 @@ import {
...
+    const pos = [
+      objectPos.x * widthHalf + widthHalf,
+      -(objectPos.y * heightHalf) + heightHalf,
+    ];
+
+    el.style.transform = `translate3d(${pos[0]}px,${pos[1]}px,0)`;
+  }
+
+  useEffect(() => {
+    el.style.cssText = `position:absolute; top:0; left:0; transform-origin:0 0;`;
```
axelboc, line 59: I figured out why `scene.updateMatrixWorld()` is required here. Without it, when I comment out the `useFrame` block, I get this:

image

The annotation is rendered in the middle of the canvas. With the `useFrame` uncommented, you can barely see it, but the annotation does flash in the middle for a split millisecond.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,63 +4,107 @@ import {
...
+    }
+
+    return () => {
       if (parentElement) {
-        parentElement.append(el);
+        el.remove();
       }
+      ReactDOM.unmountComponentAtNode(el);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
axelboc, line 74: Just a stylistic choice: I would move the ESLint comments to the end of the lines they disable.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,63 +4,107 @@ import {
...
+    className,
+    style,
+    children,
+    followCamera,
+    fullscreen,
+    groupProps = {},
+    ...divProps
+  } = props;
 
+  const size = useThree((state) => state.size);
```
axelboc, line 35: Now that the `width` and `height` don't have to be passed around, maybe destructure `size` here?

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,63 +4,107 @@ import {
...
+  const gl = useThree((state) => state.gl);
+  const { parentElement } = gl.domElement;
+  const camera = useThree(({ camera }) => camera);
```
axelboc, line 38: `const camera = useThree((state) => state.camera);` for consistency?

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,63 +4,107 @@ import {
...
+
+  // Container `div` for ReactDOM to render into, appended next to R3F's `canvas`
+  const [el] = useState(() => document.createElement('div'));
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
axelboc, line 42: Maybe add a small comment here to explain what the group is for (i.e. for `followCamera` / position tracking).

axelboc (approved):

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,63 +4,104 @@ import {
...
+    const heightHalf = height / 2;
+    const pos = [
+      objectPos.x * widthHalf + widthHalf,
+      -(objectPos.y * heightHalf) + heightHalf,
+    ];
+
+    el.style.transform = `translate3d(${pos[0]}px,${pos[1]}px,0)`;
+  }
+
+  useEffect(() => {
```
axelboc, line 59: Found a small bug: when React hot-reloads (but probably also when `parentElement` changes, if ever), the clean up function of this `useEffect` runs _after_ the `useLayoutEffect` below, which means that `ReactDOM.unmountComponentAtNode(el);` is called _after_ `ReactDOM.render()`, which means that the annotation disappears from the screen.

Replacing this `useEffect` with a `useLayoutEffect` fixes the issue.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -4,63 +4,104 @@ import {
...
 
+  const { width, height } = useThree((state) => state.size);
+  const gl = useThree((state) => state.gl);
+  const { parentElement } = gl.domElement;
+  const camera = useThree((state) => state.camera);
+  const scene = useThree((state) => state.scene);
+
+  // Container `div` for ReactDOM to render into, appended next to R3F's `canvas`
+  const [el] = useState(() => document.createElement('div'));
+  // Placeholder R3F group in the canvas that will be the counterpart of HTML in the canvas.
```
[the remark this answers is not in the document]
loichuder: Great wording :+1:

src/h5web/vis-packs/core/shared/TooltipMesh.tsx
```diff
@@ -78,7 +78,7 @@ function TooltipMesh(props: Props) {
         <meshBasicMaterial opacity={0} transparent />
         <planeGeometry args={[width, height]} />
       </mesh>
-      <Html>
+      <Html style={{ width, height }}>
```
axelboc, line 81: I don't believe you actually need this, because the `svg` of the guides has a `width` and `height`, which makes the inner `div` of `Html` fullscreen.

PR #698, opened 2021-06-10
Fix overflow menu not opening

axelboc:
`useClickAway` was triggered when clicking the button. Not sure why it was working in H5Web before 🤷


loichuder (commented):
Sorry, the diff is a bit crappy. The only modification is the `ref` moved from `more-menu` to the enclosed div, right ?

axelboc:
yeah, and I moved the `Separator` outside of that `div` so it triggers a click-away.

loichuder (approved):

PR #701, opened 2021-06-11
Allow Html to scale according to the camera zoom

loichuder:
Follow-up of #695 

Simplified version of the [scaling of drei's Html](https://github.com/pmndrs/drei/blob/v6.0.3/src/web/Html.tsx#L269) as we only support Orthographic Camera.


src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -17,6 +17,7 @@ const v1 = new Vector3();
 export interface HtmlProps extends HTMLAttributes<HTMLDivElement> {
   groupProps?: GroupProps;
   followCamera?: boolean;
+  distanceFactor?: number;
```
axelboc, line 20: I would rename this `scalingFactor`

axelboc (approved):

PR #702, opened 2021-06-11
Constrain camera panning to mesh size

axelboc:
This is just a proof of concept. I need to rethink the code significantly.

### DIFF
--- a/src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
+++ b/src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
@@ -2,25 +2,38 @@ import { useRef, useCallback, useEffect } from 'react';
 import type { Vector3 } from 'three';
 import { useThree } from '@react-three/fiber';
 import { clamp } from 'lodash';
-import { useWheelCapture } from '../hooks';
+import { useCanvasScales, useWheelCapture } from '../hooks';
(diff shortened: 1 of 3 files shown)


src/h5web/vis-packs/core/shared/PanZoomMesh.tsx
```diff
@@ -2,25 +2,38 @@ import { useRef, useCallback, useEffect } from 'react';
...
 const ZOOM_FACTOR = 0.95;
 
-function PanZoomMesh() {
+interface Props {
+  abscissaDomain: Domain;
+  ordinateDomain: Domain;
+}
+
+function PanZoomMesh(props: Props) {
+  const { abscissaDomain, ordinateDomain } = props;
```
axelboc, line 17: I need the original domains to compute the width and height of the mesh, like in `Mesh`.

axelboc (commented):
I found it confusing that the two sets of abscissa and ordinate domains (one that refers to the domains of the canvas, and the other to the domains of the mesh) are referred to with the same names (`abscissaDomain` and `ordinateDomain`), and that one is provided through props and the other through context.

Perhaps `AxisSystem` should inject `abscissaConfig` and `ordinateConfig` as is, and inject the "canvas" domains separately. Actually, I don't think we need the canvas domains -- only the canvas scales, so maybe `AxisSystem` should just inject those, as well as some of the stuff we now use in multiple places, like the size of the mesh?

PR #704, opened 2021-06-14
Compute axis scales from vis size

axelboc:
... instead of from canvas size. Since the scales are not clamped, they behave exactly the same way as before, even with the `cover` layout on the Heatmap.


src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -86,7 +86,8 @@ function HeatmapVis(props: Props) {
         axisOffsets={axisOffsets}
         aspectRatio={layout === 'contain' ? cols / rows : undefined}
       >
-        <HeatmapAxisSystem
+        <AxisSystem
+          visRatio={layout === 'cover' ? cols / rows : undefined}
```
axelboc, line 90: Using a less specific term than `imageRatio`, as now we pass this ratio directly to `AxisSystem` (instead of adding an extra layer with `HeatmapAxisSystem`).

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -101,11 +102,10 @@ function HeatmapVis(props: Props) {
             isIndexAxis: !ordinateValue,
             label: ordinateLabel,
           }}
-          imageRatio={layout === 'cover' ? cols / rows : undefined}
         >
           <TooltipMesh {...tooltipFormatters} guides="both" />
           <PanZoomMesh />
-          <Mesh abscissaDomain={abscissaDomain} ordinateDomain={ordinateDomain}>
+          <Mesh>
```
axelboc, line 108: Since the abscissa/ordinate domains are now left untouched by `AxisSystem`, we can retrieve them from `AxisSystemContext`.

src/h5web/vis-packs/core/heatmap/hooks.ts
```diff
@@ -38,36 +31,3 @@ export function useTooltipFormatters(
       format('.3')(dataArray.get(ordinateToIndex(y), abscissaToIndex(x))),
   };
 }
-
-export function useImageSize(imageRatio: number | undefined): Size {
```
axelboc, line 42: Moved to `core/hooks` and renamed to `useVisSize`. I'm open to a better naming, especially since it kinda conflicts with `computeVisSize`, which we use for the `contain` layout (although we could rename that function instead).

src/h5web/vis-packs/core/heatmap/hooks.ts
```diff
@@ -38,36 +31,3 @@ export function useTooltipFormatters(
...
-    width: mismatch > 1 ? height * imageRatio : width,
-    height: mismatch > 1 ? height : width / imageRatio,
-  };
-}
-
-export function useCanvasDomains(
-  imageSize: Size,
-  abscissaDomain: Domain,
-  ordinateDomain: Domain
-): [Domain, Domain] {
```
axelboc, line 62: Removed completely, as we no longer need to extend the domains -- we just let the unbound scales do their job.

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -110,16 +110,32 @@ export function useFrameRendering(): void {
...
-  const { abscissaConfig, ordinateConfig } = useContext(AxisSystemContext);
-  const { width, height } = useThree((state) => state.size);
+  const { visSize, abscissaConfig, ordinateConfig } =
+    useContext(AxisSystemContext);
 
   return {
-    abscissaScale: getCanvasScale(abscissaConfig, width),
-    ordinateScale: getCanvasScale(ordinateConfig, height),
+    abscissaScale: getCanvasScale(abscissaConfig, visSize.width),
+    ordinateScale: getCanvasScale(ordinateConfig, visSize.height),
```
axelboc, line 138: The key change is here 😊

src/h5web/vis-packs/core/shared/AxisSystem.tsx
```diff
@@ -3,29 +3,37 @@ import { useThree } from '@react-three/fiber';
...
-  const size = useThree((state) => state.size);
-  const { width, height } = size;
+  const canvasSize = useThree((state) => state.size);
+  const { width, height } = canvasSize;
 
-  const abscissaScale = getCanvasScale(abscissaConfig, width);
-  const ordinateScale = getCanvasScale(ordinateConfig, height);
+  const visSize = useVisSize(visRatio);
+  const abscissaScale = getCanvasScale(abscissaConfig, visSize.width);
+  const ordinateScale = getCanvasScale(ordinateConfig, visSize.height);
```
axelboc, line 36: ... and here. Note that in my next PR, I will inject the scales through context directly, so we don't have to recompute them in two different places (here and in `useCanvasScales`).

src/h5web/vis-packs/core/shared/AxisSystem.tsx
```diff
@@ -64,17 +72,19 @@ function AxisSystem(props: Props) {
...
           config={ordinateConfig}
           domain={yVisibleDomain}
-          canvasSize={size}
+          canvasSize={canvasSize}
           flipAxis
         />
       </Html>
-      <AxisSystemContext.Provider value={{ abscissaConfig, ordinateConfig }}>
+      <AxisSystemContext.Provider
+        value={{ visSize, abscissaConfig, ordinateConfig }}
```
axelboc, line 86: Providing `visSize` for use in `useCanvasScales`. Again, in my next PR, I'll replace all this with `{ abscissaScale, ordinateScale }`.

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -86,8 +86,9 @@ function HeatmapVis(props: Props) {
         axisOffsets={axisOffsets}
         aspectRatio={layout === 'contain' ? cols / rows : undefined}
       >
-        <HeatmapAxisSystem
+        <AxisSystem
           axisOffsets={axisOffsets}
+          visRatio={layout === 'cover' ? cols / rows : undefined}
```
axelboc, line 91: Using a less specific term than `imageRatio`, as now we pass this ratio directly to `AxisSystem` (instead of adding an extra layer with `HeatmapAxisSystem`).

src/h5web/vis-packs/core/heatmap/hooks.ts
loichuder: As discussed, we can rename `computeVisSize` in `computeCanvasSize`

src/h5web/vis-packs/core/hooks.ts
```diff
@@ -110,16 +110,32 @@ export function useFrameRendering(): void {
...
+    return { width, height };
+  }
+
+  const canvasRatio = width / height;
+  const mismatch = canvasRatio / ratio;
+
+  return {
+    width: mismatch > 1 ? height * ratio : width,
+    height: mismatch > 1 ? height : width / ratio,
+  };
```
loichuder, line 126: What about:
```js
if(canvasRatio > ratio) {
  return {width: height*ratio, height}
}

return { width, height: width / ratio }
```

loichuder (approved):

src/h5web/vis-packs/core/hooks.ts
axelboc: I think the reason I like the single return is that the two branches have equal value, whereas the early return makes it look like `canvasRatio > ratio` is a special case.

axelboc: Though I'm happy to replace `mismatch > 1` with `canvasRatio > ratioToRespect`

loichuder: Good old forgotten pal `else` to the rescue:

```js
if(canvasRatio > ratio) {
  return {width: height*ratio, height}
}
else {
  return { width, height: width / ratio }
}
```

But I think that it prevented by our linting rules. I liked the fact that the `if` quickly shows that only one part of the size changes but I am not deeply against the current writing with ternaries.

PR #707, opened 2021-06-15
Add Annotation component and allow children in LineVis

loichuder:
`Annotation` is a simple wrapper around `Html` that takes vis coordinates instead of canvas coordinates.


axelboc (approved):
👏

PR #708, opened 2021-06-15
Restrict tooltip mesh to vis mesh

### LINKED ISSUE #694: [Heatmap] Tooltip appears outside image bounds
With the new layout #679, some free space can be left around the image and the tooltip appears when hovering this free space.

The tooltip should only appear when the mouse is in the image bounds.

loichuder:
Fix #694


src/packages/lib.ts
```diff
@@ -20,8 +20,8 @@ export { default as GridToggler } from '../h5web/toolbar/controls/GridToggler';
 export { default as VisCanvas } from '../h5web/vis-packs/core/shared/VisCanvas';
 export { default as PanZoomMesh } from '../h5web/vis-packs/core/shared/PanZoomMesh';
 export { default as TooltipMesh } from '../h5web/vis-packs/core/shared/TooltipMesh';
+export { default as VisMesh } from '../h5web/vis-packs/core/shared/Mesh';
```
loichuder, line 23: I moved `Mesh` in `shared` so that it does not make much sense to export it as `HeatmapMesh`. `VisMesh` sounded appropriate to me...

src/h5web/vis-packs/core/shared/TooltipMesh.tsx
```diff
@@ -74,10 +75,9 @@ function TooltipMesh(props: Props) {
 
   return (
     <>
-      <mesh {...{ onPointerMove, onPointerOut, onPointerDown, onPointerUp }}>
+      <Mesh {...{ onPointerMove, onPointerOut, onPointerDown, onPointerUp }}>
         <meshBasicMaterial opacity={0} transparent />
-        <planeGeometry args={[width, height]} />
-      </mesh>
+      </Mesh>
```
loichuder, line 80: Now using `Mesh` that has the sets the geometry to the vis size and places the `mesh` at the right position

axelboc (approved):
Perfect 💯

PR #709, opened 2021-06-15
Add thin axis line when grid is not displayed

### LINKED ISSUE #692: Show axes when grid is off
When `showGrid` is `false`, both in the Line and Heatmap vis, perhaps we should still show the axes?

image

loichuder:
Fix #692 

image


axelboc (approved):
Nickel

PR #710, opened 2021-06-15
Fix continuous re-renders by separating out provider from `AxisSystem`

axelboc:
... and bring `AxisSystem` and its provider back inside `VisCanvas`.

### DIFF
--- a/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
+++ b/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
@@ -12,8 +12,7 @@ import { DEFAULT_DOMAIN } from '../utils';
 import { assertDefined } from '../../../guards';
 import { useAxisValues, useTooltipFormatters } from './hooks';
 import { useDomain } from '../hooks';
-import VisMesh from '../shared/VisMesh';
-import MeshMaterial from './MeshMaterial';
+import HeatmapMesh from './HeatmapMesh';
 
 interface Props {
   dataArray: NdArray;
@@ -94,19 +93,17 @@ function HeatmapVis(props: Props) {
       >
         <TooltipMesh {...tooltipFormatters} guides="both" />
         <PanZoomMesh />
-        <VisMesh>
-          <MeshMaterial
-            rows={rows}
-            cols={cols}
-            values={dataArray.data as number[]}
-            domain={domain}
-            colorMap={colorMap}
-            invertColorMap={invertColorMap}
-            scaleType={scaleType}
-            alphaValues={alphaArray && (alphaArray.data as number[])}
-            alphaDomain={alphaDomain}
-          />
-        </VisMesh>
+        <HeatmapMesh
+          rows={rows}
+          cols={cols}
+          values={dataArray.data as number[]}
+          domain={domain}
+          colorMap={colorMap}
+          invertColorMap={invertColorMap}
+          scaleType={scaleType}
+          alphaValues={alphaArray && (alphaArray.data as number[])}
+          alphaDomain={alphaDomain}
+        />
         {children}
       </VisCanvas>
       <ColorBar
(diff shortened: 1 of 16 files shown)


src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -103,6 +102,7 @@ function HeatmapVis(props: Props) {
             label: ordinateLabel,
           }}
         >
+          <AxisSystem axisOffsets={axisOffsets} title={title} />
```
axelboc, line 105: `AxisSystem` now has to be rendered as a child of `AxisSystemProvider`.

src/h5web/vis-packs/core/line/hooks.ts
```diff
@@ -13,7 +13,7 @@ export function useCanvasPoints(
   ordinates: number[],
   errors?: number[]
 ) {
-  const { abscissaScale, ordinateScale } = useContext(AxisSystemContext);
+  const { abscissaScale, ordinateScale } = useAxisSystemContext();
```
axelboc, line 16: Adding a custom hook for convenience.

src/h5web/vis-packs/core/shared/AxisSystem.tsx
```diff
@@ -1,40 +1,26 @@
...
-    axisOffsets,
-    visRatio,
-    title,
-    abscissaConfig,
-    ordinateConfig,
-    children,
-  } = props;
+  const { axisOffsets, title } = props;
+
+  const { abscissaConfig, ordinateConfig, abscissaScale, ordinateScale } =
```
axelboc, line 17: `AxisSystem` now reads stuff from context.

src/h5web/vis-packs/core/shared/AxisSystemProvider.tsx
```diff
@@ -0,0 +1,37 @@
...
+  ordinateConfig: AxisConfig;
+  children: ReactNode;
+}
+
+function AxisSystemProvider(props: Props) {
+  const { visRatio, abscissaConfig, ordinateConfig, children } = props;
+
+  const visSize = useVisSize(visRatio);
+  const abscissaScale = getCanvasScale(abscissaConfig, visSize.width);
+  const ordinateScale = getCanvasScale(ordinateConfig, visSize.height);
```
axelboc, line 19: When the provider renders, it causes `PanZoomMesh` to render, which causes R3F's `invalidate`. to be called, which renders a new frame. Previously, since we had `useFrameRendering` inside the provider, it meant that the provider would re-render on every frame, which led to an infinite render loop.

Now only the child `AxisSystem` component re-renders on every frame. The provider now only renders when the size of the canvas changes.

loichuder (approved):
Idea: could we integrate `AxisSystemProvider` and `AxisSystem` back in `VisCanvas` to reduce boilerplate code ?

src/packages/lib.ts
```diff
@@ -18,13 +18,14 @@ export { default as GridToggler } from '../h5web/toolbar/controls/GridToggler';
 
 // Building blocks
 export { default as VisCanvas } from '../h5web/vis-packs/core/shared/VisCanvas';
+export { default as AxisSystemProvider } from '../h5web/vis-packs/core/shared/AxisSystemProvider';
+export { default as AxisSystem } from '../h5web/vis-packs/core/shared/AxisSystem';
```
loichuder, line 22: Is there still an interest to export these if everything is inside `VisCanvas` ?

src/h5web/vis-packs/core/shared/Mesh.tsx
```diff
@@ -15,7 +15,7 @@ function Mesh(props: Props) {
     abscissaScale,
     ordinateScale,
     visSize,
-  } = useContext(AxisSystemContext);
+  } = useAxisSystemContext();
```
loichuder, line 18: Unrelated but wow that I think about it, I think we should rename `Mesh` to `VisMesh` in our code.

loichuder (approved):

src/packages/lib.ts
axelboc: 🎣

src/h5web/vis-packs/core/shared/Mesh.tsx
axelboc: Ah thought you had done it. Yeah, definitely.

loichuder:
Oh and we should also export `useAxisSystemContext`

PR #711, opened 2021-06-15
Fix empty toolbar on some renders

axelboc:
The problem is that we store measured widths against children indices, so the fact that the domain slider and its separator are not rendered right away causes issues... In some cases, the toolbar remains blank thinking some of the children have not yet been measured... I spent ages on this and couldn't find any better solution. Reverting to using `keys` is no better unless we want to force users to specify a `key` on every control....

### DIFF
--- a/src/h5web/toolbar/HeatmapToolbar.tsx
+++ b/src/h5web/toolbar/HeatmapToolbar.tsx
@@ -29,16 +29,18 @@ function HeatmapToolbar() {
 
   return (
     <Toolbar>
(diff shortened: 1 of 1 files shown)


loichuder (approved):

axelboc:
I can't replicate in production, so maybe it's something to do with the double render in `StrictMode`. Not sure. Anyway, don't think it's worth the workaround. Maybe a better solution will arise in future versions of React, or we'll find a library to do this.

PR #713, opened 2021-06-16
Put H5Web's dependencies as dependencies and externals of the app/lib packages

loichuder:


packages/webpack.config.js
```diff
@@ -20,10 +23,9 @@ function getConfig(pkg) {
       filename: '[name].js',
       libraryTarget: 'commonjs2',
     },
-    externals: {
-      react: 'react',
-      'react-dom': 'react-dom',
-    },
+    externals: Object.fromEntries(
+      Object.keys(rootPkg.dependencies).map((dep) => [dep, `commonjs2 ${dep}`])
```
axelboc, line 27: I mark all the dependencies as externals.

packages/webpack.config.js
```diff
@@ -82,6 +84,18 @@ function getConfig(pkg) {
...
         ],
       }),
+      new JsonPostProcessPlugin({
+        matchers: [
+          {
+            // Add H5Web's dependencies to the package's output `package.json`
+            matcher: /^package\.json$/u,
+            action: (json) => ({
+              ...json,
+              dependencies: omit(rootPkg.dependencies, ['react', 'react-dom']),
```
axelboc, line 94: I add all the dependencies to the `package.json` of `@h5web/lib` and `@h5web/app`. This is so we don't have to keep the version numbers in sync in three different `package.json` files...

axelboc (approved):

packages/webpack.config.js
loichuder: I see.

Could we use `.filter` to avoid the `lodash` dependency ?

axelboc: Hmm it's quite convenient, and the dependency is only in `packages/package.json`, so it doesn't really matter.

axelboc:
Closing in favour of moving to a monorepo structure.

Issue #649, opened 2021-05-06
Add support for the HDF5 Image Specification (eg RGB data)


planetmarshall:
It would be useful to display RGB data as a bitmap - ie identifying datasets with the `IMAGE` and `IMAGE_SUBCLASS` attributes.

See [HDF5 Image Sepcification](https://support.hdfgroup.org/HDF5/doc/ADGuide/ImageSpec.html)

axelboc:
Hi @planetmarshall, thanks for brining this to our attention. The image specification seems quite complex, so I doubt we can support it fully, but we'd be willing to implement a very basic subset of it - e.g. to support RGB data as you mention.

Which features of the specification (attributes and exact values) are you interested in specifically? And do you have an example HDF5 file we could use for testing purposes?

planetmarshall:
Yes - basic support for RGB data would be perfect. I'm not familiar with the code, but if I can see how support for that is introduced I can always contribute additional support myself.

An [example h5](https://github.com/silx-kit/h5web/files/6514829/image_test.h5.gz) file is attached, it is created from the [Mandrill](https://github.com/scijs/baboon-image) test image as follows:

```.py
import skimage.io
import h5py

rgb = skimage.io.imread("https://raw.githubusercontent.com/mikolalysenko/baboon-image/master/baboon.png")
with h5py.File("image_test.h5", "w") as fp:
    image = fp.create_dataset("mandrill", data=rgb)
    image.attrs["CLASS"] = "IMAGE"
    image.attrs["SUBCLASS"] = len("IMAGE_TRUECOLOR")
    image.attrs["IMAGE_VERSION"] = "1.2"
```

When viewed with, for example, the Jupyter viewer ideally I would be able to visualize the dataset as an RGB image. 

Additionally it would be helpful to be able to switch between RGB and BGR interpretation (as some image are inverted on that axis, for example images from OpenCV) - in a similar fashion to how colour maps can be selected, however again this is always something I can contribute myself.

Many thanks.

vasole:
@t20100 

In principle the interpretation attribute can be used to interpret the RGB case. I do not remember if it is already supported in silx.

Should we check for the presence of the interpretation attribute **and** if not present the HDF5 Image attributes? I'm in favor.

planetmarshall:
Another issue which I've only just noticed is that the y-axis should be inverted for image data, so that (0,0) is at the top left.

loichuder:
[...]

PR #715, opened 2021-06-17
Add the possibility to flip the y axis in HeatmapVis

loichuder:
image

Should also help for #649

### DIFF
--- a/src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
+++ b/src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
@@ -8,6 +8,7 @@ import {
   UnsignedByteType,
 } from 'three';
 import { Domain, ScaleType } from '../models';
+import { useAxisSystemContext } from '../shared/AxisSystemContext';
 import VisMesh from '../shared/VisMesh';
 import type { ColorMap } from './models';
 import { getInterpolator } from './utils';
@@ -80,9 +81,11 @@ function HeatmapMesh(props: Props) {
     return new DataTexture(colors, CMAP_SIZE, 1, RGBFormat, UnsignedByteType);
   }, [colorMap, invertColorMap]);
 
+  const { ordinateConfig } = useAxisSystemContext();
   const shader = {
     uniforms: {
       data: { value: dataTexture },
+      originY: { value: ordinateConfig.flip ? 1 : 0 },
       withAlpha: { value: alphaValues ? 1 : 0 },
       alpha: { value: alphaTexture },
       colorMap: { value: colorMapTexture },
@@ -96,9 +99,10 @@ function HeatmapMesh(props: Props) {
     },
     vertexShader: `
       varying vec2 coords;
+      uniform float originY;
 
       void main() {
-        coords = uv;
+        coords = vec2(uv[0], abs(uv[1] - originY));
         gl_Position = projectionMatrix * modelViewMatrix * vec4( position, 1.0 );
       }
     `,
--- a/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
+++ b/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
@@ -27,6 +27,7 @@ interface Props {
   ordinateParams?: AxisParams;
   alphaArray?: NdArray;
   alphaDomain?: Domain;
+  flipYAxis?: boolean;
   children?: ReactNode;
 }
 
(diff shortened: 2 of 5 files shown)


src/h5web/vis-packs/core/utils.ts
```diff
@@ -235,7 +239,7 @@ export function getCanvasScale(
  * So we implement our own, simplified version of `d3.ticks` that always outputs integer values.
  */
 export function getIntegerTicks(domain: Domain, count: number): number[] {
-  const [min, max] = domain;
+  const [min, max] = domain[1] < domain[0] ? [domain[1], domain[0]] : domain;
```
loichuder, line 242: The method was not returning any ticks if the domain was inverted.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -218,11 +218,15 @@ export function getCanvasScale(
...
+  const { scaleType, visDomain, flip } = config;
+  const range = [-canvasSize / 2, canvasSize / 2];
+  if (flip) {
+    range.reverse();
+  }
 
   return createAxisScale({
     domain: visDomain,
-    range: [-canvasSize / 2, canvasSize / 2],
+    range,
```
loichuder, line 229: This inverts the scale. As a consequence, all components that rely on the scale will follow the inversion, namely:
- `Axes`
- `TooltipMesh`
- `DataCurve`
- `Annotation`

src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
```diff
@@ -96,9 +99,10 @@ function HeatmapMesh(props: Props) {
     },
     vertexShader: `
       varying vec2 coords;
+      uniform float originY;
 
       void main() {
-        coords = uv;
+        coords = vec2(uv[0], abs(uv[1] - originY));
```
loichuder, line 105: The shader does not use the scale and must be inverted "manually".

As a reminder, `uv` are the normalized coordinates that vary between 0 and 1. With this formula, when `originY = 1`, it will flip the coordinates: 1 will become 0, 0 will become 1, 0.7 will become 0.3 and so on... 

When `originY = 0`, there is no change to `coords` as `uv` are always positive.

axelboc: Great!

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -103,6 +106,7 @@ function HeatmapVis(props: Props) {
           scaleType={scaleType}
           alphaValues={alphaArray && (alphaArray.data as number[])}
           alphaDomain={alphaDomain}
+          flipYAxis={flipYAxis}
```
axelboc, line 109: Do you think this info should be retrieved from the axis system context (i.e. single source of truth)?

src/h5web/vis-packs/core/utils.ts
axelboc: `range: flip ? range.reverse() : range`

src/h5web/vis-packs/core/utils.ts
axelboc: Might be easier to do `const min = Math.min(...domain)` ?

axelboc (approved):
Cool! Very easy from here to add support for flipping the `x` axis if need be.

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
loichuder: I considered it as it would indeed feel cleaner. But then, I saw it required two destructurings so my laziness won:
```
  const { ordinateConfig } = useAxisSystemContext();
  const { flip: flipYAxis } = ordinateConfig;
```

Happy to implement it like that if you prefer though

src/h5web/vis-packs/core/utils.ts
loichuder: Feels a bit odd when `domain` only contains two values but why not.

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
axelboc: yeah the second one seems reasonable

src/h5web/vis-packs/core/utils.ts
axelboc: meh, up to you ^^

PR #717, opened 2021-06-18
Mark Three as dep/external of packages

axelboc:
Three is a peer dependency of R3F, so we need to also make it an explicit dependency.

### DIFF
--- a/packages/webpack.config.js
+++ b/packages/webpack.config.js
@@ -23,7 +23,7 @@ function getConfig(pkg) {
       libraryTarget: 'commonjs2',
     },
     externals: Object.fromEntries(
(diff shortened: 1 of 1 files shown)


packages/webpack.config.js
```diff
@@ -88,21 +88,19 @@ function getConfig(pkg) {
       }),
       new JsonPostProcessPlugin({
         // Add H5Web's R3F dependency to @h5web/lib's `package.json`
-        matchers:
-          pkg === 'lib'
```
axelboc, line 92: I realised that I was making R3F an external without adding it as a dependency, so v0.0.21 most likely would not work in `jupyterlab-h5web`.

loichuder (approved):

PR #722, opened 2021-06-24
Show file name instead of file path in headings

loichuder:
With https://github.com/silx-kit/jupyterlab-h5web/issues/37 next release, the filepaths in h5web can be absolute and therefore quite lengthy. I think it would be prettier to show only the filename and not the full path, which is the point of this PR.

**Before:**
image

**After:**
image

### DIFF
--- a/src/h5web/providers/Provider.tsx
+++ b/src/h5web/providers/Provider.tsx
@@ -56,10 +56,12 @@ function Provider(props: Props) {
     });
   }, [api]);
 
+  const filepathMembers = api.filepath.split('/');
+
   return (
     <ProviderContext.Provider
       value={{
-        filepath: api.filepath,
+        filename: filepathMembers[filepathMembers.length - 1],
         entitiesStore,
         valuesStore,
       }}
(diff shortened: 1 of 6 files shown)


src/h5web/providers/Provider.tsx
```diff
@@ -56,10 +56,12 @@ function Provider(props: Props) {
     });
   }, [api]);
 
+  const filepathMembers = api.filepath.split('/');
```
axelboc, line 59: Any risk of seeing a trailing slash?

loichuder: Sounds very unlikely: the last member is the filename and I never seen things like `file.h5/`

axelboc: I'm thinking more of HSDS, where the buckets aren't really files, but yeah, very unlikely.

loichuder: Good point ! I tried and the last member of the bucket name cannot have a trailing slash whether when loading the data in HSDS or when trying to fetch it in h5web

axelboc: Awesome 💯

PR #723, opened 2021-06-28
Add more features in RGB Vis

loichuder:
More work on #649 

- Add a control to switch between RGB and BGR interpretation (requested in https://github.com/silx-kit/h5web/issues/649#issuecomment-844928138)
- Add support for float types 

From https://support.hdfgroup.org/HDF5/doc/ADGuide/ImageSpec.html :
> Image data is stored as an HDF5 dataset with values of HDF5 class Integer or Float.


src/h5web/guards.ts
```diff
@@ -130,16 +129,6 @@ export function hasNumericType<S extends Shape>(
...
-function isIntegerType(type: DType): type is IntegerType {
-  return [DTypeClass.Integer, DTypeClass.Unsigned].includes(type.class);
-}
-
-export function hasIntegerType<S extends Shape>(
-  dataset: Dataset<S>
-): dataset is Dataset<S, IntegerType> {
-  return isIntegerType(dataset.type);
-}
-
```
loichuder, line 142: All the `IntegerType` guards are no longer needed as `RgbVis` supports `NumericType` now

src/h5web/vis-packs/core/rgb/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  const baseArray = ndarray(value, dims);
+  const steps = dims.map(() => 1);
+  steps[steps.length - 1] = -1;
+
+  const flippedView = baseArray.step(...steps);
+  const flippedArray = ndarray([], dims);
+  assign(flippedArray, flippedView);
+
+  return flippedArray.data as number[];
+}
```
loichuder, line 14: Based on the same principles as `applyMapping` https://github.com/silx-kit/h5web/blob/main/src/h5web/vis-packs/core/utils.ts#L321

The last dimension is stepped in reversed order (step = -1) to generate a **view** (i.e. `baseArray.data` is unchanged) with the last dimension flipped. The extraction of `data` is done by assigning in a new `ndarray`

src/h5web/toolbar/RgbVisToolbar.tsx
```diff
@@ -4,15 +4,29 @@ import Toolbar from './Toolbar';
...
+      <Selector
+        label="Image type"
+        value={imageType}
+        onChange={setImageType}
+        options={Object.values(ImageType)}
+        optionComponent={({ option }) => (
+          // eslint-disable-next-line react/jsx-no-useless-fragment
+          <>{option}</>
+        )}
+      />
```
axelboc, line 26: Out of curiosity, do we expect more image types? If not, a toggle group may be easier to use.

src/h5web/vis-packs/core/rgb/utils.ts
```diff
@@ -0,0 +1,14 @@
+import ndarray from 'ndarray';
+import { assign } from 'ndarray-ops';
+
+export function flipLastDimension(value: number[], dims: number[]) {
+  const baseArray = ndarray(value, dims);
+  const steps = dims.map(() => 1);
+  steps[steps.length - 1] = -1;
```
axelboc, line 7: I would put a ternary inside the `map`, based on the `index` of the dimension.

src/h5web/vis-packs/core/rgb/utils.ts
axelboc: Makes sense. Might be worth making a utility for this, which would take a `view` and `dims` and return a new, assigned ndarray.

src/h5web/vis-packs/core/rgb/RgbVis.tsx
```diff
@@ -3,39 +3,60 @@ import styles from '../heatmap/HeatmapVis.module.css';
...
+import { DataTexture, FloatType, RGBFormat, UnsignedByteType } from 'three';
 import VisMesh from '../shared/VisMesh';
+import { flipLastDimension } from './utils';
+import { ImageType } from './models';
+import { DTypeClass, NumericType } from '../../../providers/models';
 
 interface Props {
   value: number[];
   dims: number[];
+  dtype: NumericType;
```
axelboc, line 15: Creating the texture is the responsibility of the vis component, so there is no way around passing some sort of type information. That being said, we could just pass a boolean, like `withFloats` so that our "HDF5" models don't leak into the vis components.

axelboc (changes requested):
Awesome!

Two additional questions/requests:

1. Could you please add an E2E test? You could navigate to the image dataset, take a screenshot, then switch to BGR and take another.
2. Basically, what you've done so far amounts to supporting `IMAGE_SUBCLASS=IMAGE_TRUECOLOR`, right? I don't think we'll add support for the other subclasses, right? Should we add an assertion of sorts on `IMAGE_SUBCLASS`? Maybe throw if the attribute exist and its value is _not_ `IMAGE_TRUECOLOR`?

src/h5web/toolbar/RgbVisToolbar.tsx
loichuder: > Out of curiosity, do we expect more image types?

Not sure but the toggle group seems indeed handier

loichuder:
> Awesome!
> 
> Two additional questions/requests:
> 
>     1. Could you please add an E2E test? You could navigate to the image dataset, take a screenshot, then switch to BGR and take another.
> 
Done.

>     2. Basically, what you've done so far amounts to supporting `IMAGE_SUBCLASS=IMAGE_TRUECOLOR`, right? I don't think we'll add support for the other subclasses for now, right? Should we add an assertion of sorts on `IMAGE_SUBCLASS`? Maybe throw if the attribute exists AND its value is _not_ `IMAGE_TRUECOLOR`?

Not sure if it is worth enforcing strict compliance... Going this way, according to the [spec](https://support.hdfgroup.org/HDF5/doc/ADGuide/ImageSpec.html), an image dataset must have the `IMAGE_VERSION` attribute but I don't check it for now.

axelboc (approved):

axelboc:
> Not sure if it is worth enforcing strict compliance... Going this way, according to the [spec](https://support.hdfgroup.org/HDF5/doc/ADGuide/ImageSpec.html), an image dataset must have the `IMAGE_VERSION` attribute but I don't check it for now.  

It was more to prevent displaying images with non-rgb data. But maybe this is already sufficiently enforced by the 3-dimensions check.

PR #725, opened 2021-06-29
Make clear that IMAGE_SUBCLASS other than IMAGE_TRUECOLOR are unsupported

loichuder:
Follow-up of https://github.com/silx-kit/h5web/pull/723#pullrequestreview-693953694 and https://github.com/silx-kit/h5web/pull/723#issuecomment-870329229

### DIFF
--- a/src/h5web/vis-packs/core/containers/RgbVisContainer.tsx
+++ b/src/h5web/vis-packs/core/containers/RgbVisContainer.tsx
@@ -11,6 +11,7 @@ import RgbVis from '../rgb/RgbVis';
 import { useRgbVisConfig } from '../rgb/config';
 import shallow from 'zustand/shallow';
 import { DTypeClass } from '../../../providers/models';
+import { getAttributeValue } from '../../../utils';
 
 function RgbVisContainer(props: VisContainerProps) {
   const { entity } = props;
@@ -18,6 +19,11 @@ function RgbVisContainer(props: VisContainerProps) {
   assertArrayShape(entity);
   assertNumDims(entity, 3);
   assertNumericType(entity);
+
+  const subclassAttr = getAttributeValue(entity, 'IMAGE_SUBCLASS');
+  if (typeof subclassAttr === 'string' && subclassAttr !== 'IMAGE_TRUECOLOR') {
+    throw new Error('RGB Vis only supports IMAGE_TRUECOLOR.');
+  }
 
   const { shape: dims } = entity;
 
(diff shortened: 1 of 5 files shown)


src/h5web/vis-packs/core/containers/RgbVisContainer.tsx
```diff
@@ -19,6 +20,12 @@ function RgbVisContainer(props: VisContainerProps) {
   assertNumDims(entity, 3);
   assertNumericType(entity);
 
+  const subclassAttr = getAttributeValue(entity, 'IMAGE_SUBCLASS');
+  if (typeof subclassAttr !== 'string' || subclassAttr !== 'IMAGE_TRUECOLOR') {
+    // eslint-disable-next-line no-console
+    console.warn('RGB Vis only supports IMAGE_TRUECOLOR.');
```
axelboc, line 26: I was more suggesting throwing if the subclass attribute _is_ provided AND _is not_ `IMAGE_TRUECOLOR`. The idea is to fail early if we know for sure the image is not truecolor. If the subclass attribute is not provided, we can't tell, so we give the image the benefit of the doubt.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -387,3 +389,15 @@ export function createArrayFromView<T>(view: NdArray<T>): NdArray<T> {
...
+
+export function hasImageAttributes(dataset: Dataset): boolean {
+  const classAttr = getAttributeValue(dataset, 'CLASS');
+  const versionAttr = getAttributeValue(dataset, 'IMAGE_VERSION');
+
+  return (
+    typeof classAttr === 'string' &&
+    classAttr === 'IMAGE' &&
+    typeof versionAttr === 'string' &&
+    versionAttr === '1.2'
```
axelboc, line 401: I'm not convinced the version check is required. This seems like the kind of attribute people might forget. I'd be more for following the same guideline as we do for NeXus of checking the least amount of stuff possible (while still failing on critical syntax errors or unsupported features like `IMAGE_SUBCLASS=IMAGE_BITMAP`)

axelboc (changes requested):

src/h5web/vis-packs/core/containers/RgbVisContainer.tsx
loichuder: Sorry, I misunderstood. Your proposition makes sense :+1:

axelboc (approved):

PR #726, opened 2021-06-29
Prevent HTML annotations from overflowing canvas

axelboc:
image

Big refactoring, but worth it, I think. It's no longer the outer element that is transformed around the canvas, but the inner element. The outer element now takes the whole size of the canvas, which means it now has to have the `pointer-events: none`. This revealed an event propagation regression with the tooltip, which was easily fixed. Most importantly, I find that it makes the roles of all the `useLayoutEffect`, `useEffect` and `useFrame` hooks a lot clearer (as demonstrated by the fact that we no longer need to disable ESLint because of missing hook dependencies).

### DIFF
--- a/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
+++ b/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
@@ -94,8 +94,8 @@ function HeatmapVis(props: Props) {
           flip: flipYAxis,
         }}
       >
-        <TooltipMesh {...tooltipFormatters} guides="both" />
         <PanZoomMesh />
+        <TooltipMesh {...tooltipFormatters} guides="both" />
         <HeatmapMesh
           rows={rows}
           cols={cols}
--- a/src/h5web/vis-packs/core/shared/Html.tsx
+++ b/src/h5web/vis-packs/core/shared/Html.tsx
@@ -1,26 +1,27 @@
 import {
   useState,
-  forwardRef,
   HTMLAttributes,
   useLayoutEffect,
   useRef,
+  useCallback,
+  useEffect,
 } from 'react';
 import ReactDOM from 'react-dom';
 import { GroupProps, useFrame, useThree } from '@react-three/fiber';
 import { Group, Vector3 } from 'three';
-
-// Simplified version of `drei`'s `<Html>` component
-// https://github.com/pmndrs/drei/blob/v6.0.3/src/web/Html.tsx
+import { isDefined } from '../../../guards';
 
 const v1 = new Vector3();
 
-interface HtmlProps extends HTMLAttributes<HTMLDivElement> {
+interface Props extends HTMLAttributes<HTMLDivElement> {
   groupProps?: GroupProps;
   followCamera?: boolean;
   scaleOnZoom?: boolean;
 }
 
-const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
+// Customised version of drei's `<Html>` component
+// https://github.com/pmndrs/drei/blob/v6.0.3/src/web/Html.tsx
+function Html(props: Props) {
   const {
     className,
     style,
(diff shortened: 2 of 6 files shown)


src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -32,36 +33,50 @@ const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
...
-  }
+  const getInnerDivTransform = useCallback(() => {
+    const position = followCamera ? getGroupPosition() : undefined;
 
-  useLayoutEffect(() => {
-    group.current.updateMatrixWorld(); // To place correctly `el` on first render
-    el.style.cssText = `position:absolute; top:0; left:0; transform-origin:0 0;`;
-    if (followCamera) {
-      updateHtmlPosition();
-    }
```
axelboc, line 63: The diff is ugly, but this `useLayoutEffect` now only contains the append/remove logic.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -32,36 +33,50 @@ const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
...
+    return [
+      position && `translate3d(${position[0]}px, ${position[1]}px, 0)`,
+      scaleOnZoom ? `scale(${camera.zoom})` : undefined,
+    ]
+      .filter(isDefined)
+      .join(' ');
+  }, [camera, followCamera, getGroupPosition, scaleOnZoom]);
```
axelboc, line 76: This `useCallback` returns the `transform` string, which is empty if both `scaleOnZoom` and `followCamera` are disabled.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -70,20 +85,32 @@ const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
...
     };
-  }, [el, parentElement]); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [el, parentElement]);
 
+  // Update size and overflow of container `div`
   useLayoutEffect(() => {
+    el.style.width = `${width}px`;
+    el.style.height = `${height}px`;
+    el.style.overflow = followCamera ? 'hidden' : 'visible';
+  }, [el, width, height, followCamera]);
```
axelboc, line 98: In this `useLayoutEffect`, I give and update the size of the outer `div` and I set `overflow: hidden` when `followCamera` is enabled. This is what allows annotations to be cropped on the edges of the canvas.

As mentioned, since the outer `div` now has a size, it has to have the `pointer-events: none` declaration (set in the `useState` hook above).

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -70,20 +85,32 @@ const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
...
+
+  // Render inner `div`
+  useEffect(() => {
+    group.current.updateMatrixWorld(); // make sure we get the group's most up to date position
```
axelboc, line 102: `updateMatrixWorld` is now called here, before rendering the inner `div` (instead of before appending the outer `div`.

Note that this is now a `useEffect` instead of `useLayoutEffect`. This is because `useFrame` can now be safely called before this `useEffect`.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -70,20 +85,32 @@ const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
...
-        ref={ref}
+        ref={innerRef}
         className={className}
         style={{
           position: 'absolute',
           top: 0,
           left: 0,
-          pointerEvents: 'none', // let pointer events pass through to canvas
+          transformOrigin: 'left top',
+          transform: getInnerDivTransform(),
```
axelboc, line 113: As mentioned, we now set the `transform` on the inner `div`. We compute it here right away to avoid a positioning glitch.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -92,15 +119,16 @@ const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
       </div>,
       el
     );
-  }, [children, className, el, style, ref, divProps]);
+  }, [children, className, divProps, el, getInnerDivTransform, style]);
```
axelboc, line 122: The `getInnerDivTransform` is now an explicit dependency, which is much cleaner. This is why it has to be a `useCallback`.

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -92,15 +119,16 @@ const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
...
 
+  // Update inner `div` position on every frame
   useFrame(() => {
-    if (followCamera) {
-      updateHtmlPosition();
+    if (innerRef.current) {
+      innerRef.current.style.transform = getInnerDivTransform();
     }
```
axelboc, line 128: The condition checks whether the ref is ready, which makes `useFrame` safe if called before `useEffect`. Note that if `followCamera` is disabled, the `transform` is set to the empty string (i.e. the CSS property is not set, or removed if present).

src/stories/LineVisDisplay.stories.tsx
```diff
@@ -67,8 +67,8 @@ WithTitle.args = {
 
 export const WithAnnotation: Story<LineVisProps> = (args) => (
   <LineVis {...args}>
-    <Annotation x={30} y={10} style={{ width: '200px' }}>
-      <p>A simple line</p>
+    <Annotation x={30} y={10} style={{ width: '100px' }}>
+      a very simple line
```
axelboc, line 71: These changes test the wrapping of the annotation.

src/stories/HeatmapVis.stories.tsx
```diff
@@ -165,8 +165,8 @@ WithAlphaArray.args = {
 
 export const WithAnnotation: Story<HeatmapVisProps> = (args) => (
   <HeatmapVis {...args}>
-    <Annotation x={30} y={18} style={{ width: '200px' }}>
-      <p style={{ color: 'white' }}>An annotation</p>
+    <Annotation x={30} y={18} style={{ width: '200px', color: 'white' }}>
+      An annotation
```
axelboc, line 169: Just simplifying this a bit, now that `Annotation` puts its `style` on an inner `div`.

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -94,8 +94,8 @@ function HeatmapVis(props: Props) {
           flip: flipYAxis,
         }}
       >
-        <TooltipMesh {...tooltipFormatters} guides="both" />
         <PanZoomMesh />
+        <TooltipMesh {...tooltipFormatters} guides="both" />
```
loichuder, line 98: Should this be documented for `@h5web/lib` ?

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -70,20 +85,32 @@ const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
...
       ReactDOM.unmountComponentAtNode(el);
     };
-  }, [el, parentElement]); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [el, parentElement]);
 
+  // Update size and overflow of container `div`
   useLayoutEffect(() => {
+    el.style.width = `${width}px`;
+    el.style.height = `${height}px`;
+    el.style.overflow = followCamera ? 'hidden' : 'visible';
```
loichuder, line 97: Perhaps it should be a dedicated boolean ? Like `limitToCanvas` ?

src/h5web/vis-packs/core/shared/Html.tsx
```diff
@@ -32,36 +33,50 @@ const Html = forwardRef<HTMLDivElement, HtmlProps>((props, ref) => {
...
-  const camera = useThree((state) => state.camera);
 
   // Container `div` for ReactDOM to render into, appended next to R3F's `canvas`
-  const [el] = useState(() => document.createElement('div'));
-  // Placeholder R3F group, the position of which is tracked and forwarded to `el` when `followCamera` is enabled
+  const [el] = useState(() => {
+    const elem = document.createElement('div');
+
+    // Take out of flow and let pointer events pass through to canvas
+    elem.style.cssText = `position: absolute; top: 0; left: 0; pointer-events: none;`;
```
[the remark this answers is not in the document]
loichuder: Isn't it possible to set `overflow` here ?

loichuder (approved):
Impressive :100:

src/h5web/vis-packs/core/shared/Html.tsx
axelboc: Yeah, I did think about that, but in practice, would we ever want the overflow to be visible when following the camera, or hidden when not following the camera? I can't think of a use case where the two settings need to be decoupled, so I'm thinking that we should wait until one comes along.

src/h5web/vis-packs/core/shared/Html.tsx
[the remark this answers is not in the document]
axelboc: The `overflow` depends on a prop, so it needs to be updated dynamically. Note that since it's set in a `useLayoutEffect`, there's no chance of seeing a glitch on first render (i.e. the browser will always paint the DOM _after_ applying the correct overflow).

src/h5web/vis-packs/core/shared/Html.tsx
loichuder: Sensible :+1:

src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
axelboc: Went a bit too far with Storybook changes. Going to open a separate PR.

PR #734, opened 2021-06-29
Show filepath instead of filename in root group entity table

### LINKED ISSUE #727: Allow to see the full file path somewhere
Due to https://github.com/silx-kit/h5web/pull/722, there is no longer a way to see the full file path.

The Inspect panel for the root group may be an ideal location to put this.

loichuder:
image

Fix #727

### DIFF
--- a/src/h5web/metadata-viewer/EntityTable.tsx
+++ b/src/h5web/metadata-viewer/EntityTable.tsx
@@ -14,7 +14,9 @@ interface Props {
 function EntityTable(props: Props) {
   const { entity } = props;
   const { name, path, kind } = entity;
-  const { filename } = useContext(ProviderContext);
+  const { filepath } = useContext(ProviderContext);
+
+  const isRoot = path === '/';
 
   return (
     <table className={styles.table}>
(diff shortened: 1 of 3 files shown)


src/h5web/metadata-viewer/EntityTable.tsx
```diff
@@ -31,8 +31,8 @@ function EntityTable(props: Props) {
           <td>{path}</td>
         </tr>
         <tr>
-          <th scope="row">Name</th>
-          <td>{path === '/' ? filename : name}</td>
+          <th scope="row">{path === '/' ? 'File path' : 'Name'}</th>
+          <td>{path === '/' ? filepath : name}</td>
```
axelboc, line 35: Maybe introduce an `isRoot` or `isRootEntity` constant for clarity?

axelboc (approved):

src/h5web/metadata-viewer/EntityTable.tsx
loichuder: Agreed, I considered doing that :+1:

PR #735, opened 2021-06-30
Upgrade Storybook and fix source code blocks

axelboc:
I was able to revert my fix for the source code blocks. They've fixed the decorators issue.

The release notes are a bit obscure -- I had to dig through some PRs to get to some of the changes: https://github.com/storybookjs/storybook/releases/tag/v6.3.0.

Worth noting are two new buttons in the toolbar: one to draw an outline around every DOM element, and an other to measure DOM elements by hovering them. They've also stabilised support for webpack 5, which might be handy when `react-scripts` finally decides to upgrade, and they've got experimental support for Vite, which is a toolchain I've been looking at for building the library once we move to a monorepo. Finally, they allow stories to be unit tested with Jest, which could be handy.


.storybook/preview.js
```diff
@@ -3,6 +3,7 @@ import '../src/styles/index.css';
 
 export const parameters = {
   actions: { argTypesRegex: '^on[A-Z].*' },
+  docs: { source: { excludeDecorators: true } },
```
axelboc, line 6: This isn't yet documented ... I had to dig through a PR to figure out how to make it work...

src/stories/VisCanvasInteraction.stories.tsx
```diff
@@ -37,9 +37,8 @@ const Template: Story<TemplateProps> = (args) => {
 
 export const PanZoom = Template.bind({});
 
-PanZoom.args = {
-  panZoom: true,
-};
+PanZoom.args = { panZoom: true };
+PanZoom.argTypes = { guides: { table: { disable: true } } };
```
axelboc, line 41: Hiding the `guides`, since we're not rendering a tooltip in this story.

axelboc:
For some reason , the tests started failing because of a syntax issue in `d3-array` v3 that is not supported by Babel. I think upgrading Storybook changed the order of some dependencies in `package-lock.json`, leading to `d3-array` v3 being imported instead of v2 (which is still present because some other deps haven't upgraded `d3-array` to v3 yet).

Anyway, so I've downgraded `d3-array`, and I decided to start documenting problematic cases like this in `CONTRIBUTING.md`. I probably forgot some, but it's a start.

I've also removed `@babel/core` from `package.json`. Everything works fine without it, so I think whatever issue we had that made us add it is no longer relevant.

axelboc:
I'm really tempted to switch to `yarn` when we move to a monorepo, just so we can take advantage of this feature: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

CONTRIBUTING.md
```diff
@@ -169,3 +170,42 @@ npm link @h5web/lib
...
+
+- The major version number of `@types/node` must match the version of Node
+  specified in the `engine` field of `package.json`.
+- `d3-array` (and therefore `@types/d3-array`) cannot be upgraded to v3.x due to
+  a [bug in Babel](https://github.com/babel/babel/issues/11038).
+- Dev dependencies related to ESLint must be upgraded together with
+  `eslint-config-galex`. Their vesion numbers must match the ones in
+  [the shared config's `package.json`](https://github.com/ljosberinn/eslint-config-galex/blob/master/package.json#L66).
+  The reason for this is that `react-scripts` installs older versions of these
+  dependencies, and those versions risk being required when running
```
loichuder, line 192:
```suggestion
  dependencies which may be used when running
```

CONTRIBUTING.md
```diff
@@ -169,3 +170,42 @@ npm link @h5web/lib
...
+
+Beware that some dependencies cannot be upgraded because of issues or
+incompatibilities, while others must be upgraded by following a specific
+process. This section documents such cases:
+
+- The major version number of `@types/node` must match the version of Node
+  specified in the `engine` field of `package.json`.
+- `d3-array` (and therefore `@types/d3-array`) cannot be upgraded to v3.x due to
+  a [bug in Babel](https://github.com/babel/babel/issues/11038).
+- Dev dependencies related to ESLint must be upgraded together with
```
loichuder, line 188: I would put the list before:

> These eslint dev depedencies must be upgraded together with `eslint-config-galex:
>  - list
>  - ...
> Their **version (typo there)** numbers...
>

CONTRIBUTING.md
```diff
@@ -169,3 +170,42 @@ npm link @h5web/lib
...
+  `eslint-config-galex`. Their vesion numbers must match the ones in
+  [the shared config's `package.json`](https://github.com/ljosberinn/eslint-config-galex/blob/master/package.json#L66).
+  The reason for this is that `react-scripts` installs older versions of these
+  dependencies, and those versions risk being required when running
+  `npm run lint:eslint` instead of the more recent versions, leading to errors
+  suchs as unknown linting rules/options. This concerns the following dev
```
loichuder, line 194:
```suggestion
  such as unknown linting rules/options. This concerns the following dev
```

loichuder (approved):

loichuder:
> I'm really tempted to switch to `yarn` when we move to a monorepo, just so we can take advantage of this feature: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/. NPM is [working on a similar feature](https://github.com/npm/statusboard/issues/343), but it's gonna take ages.

As you wish, I don't have any preference

CONTRIBUTING.md
axelboc: Hmm, yeah but the fact that this is all inside a bullet point makes it tricky.

Issue #35, opened 2020-03-25
Copying/exporting datasets


axelboc-esrf:
The goal is to be able to use a dataset somewhere else, like in a Jupyter Notebook. The copy could be done either by reference or by value. The export could be to CSV or other raw formats.

axelboc:
User feedback suggested allowing rows/columns of the Matrix visualization to be selected and copied. Because the grid is virtualized, we can't rely on default text selection and copy; we'd have to provide a specific UI for this and make use of the [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API).

For the UI:
- We could have a button in the toolbar to copy the whole table, another to copy the current slice for 3D+ datasets, and an input of sorts to enter a range to copy (e.g. `0,0,:,:`).
- We could allow selection of visible rows and columns with click-and-drag -- when a selection is active, we could then show a button in the toolbar to copy the selection (and another button to clear the selection).
- We could allow copying a single row/column by showing a button in the row/column header cell on hover.
- ...

axelboc:
> User feedback suggested allowing rows/columns of the Matrix visualization to be selected and copied. 

We won't support this behaviour.

However, we should allow users to export data from other visualizations:

- Line as CSV
- Heatmap as CSV
- NxSpectrum as CSV, with custom axis values, and with axis label and plot title as column headers
- NxImage as CSV -- exact format to be discussed
- TIFF and NPY formats should also be considered

axelboc:
We're going to implement TIFF encoding in H5Grove (i.e. in the back-end) and use provider-level feature detection to determine whether to show the export button in the toolbars.

Once we have the feature detection system implemented, we'll be able to do the same for the CSV and NPY exports.

axelboc:
Now that we have a proof of concept for TIFF export, we know we can do the same with CSV and NPY (and even replace the current CSV export feature on the Matrix vis so it's done server-side). The question is then: should we provide all three exports on _all_ visualizations? Or does TIFF only make sense on Heatmap/NxImage, CSV on Matrix, etc.?

From a UI standpoint, we can merge the export/snapshot toolbar controls into a single "Export" drop-down to save space. Which means that if we want to display all four exports (TIFF/CSV/NPY/snapshot) on all visualization, it's no problem at all.

axelboc:
As per our discussion:

- `MatrixVis` : CSV ☑️ (to be delegated to H5Grove), NPY
- `LineVis` : CSV, NPY
[...]

PR #739, opened 2021-07-02
Implement CSV export for Matrix vis

axelboc:
Part of #35 

This is the simplest step I could think of to make some data exportable from H5Web. The Matrix vis now shows a button to download the current slice as CSV.

image

image


src/h5web/toolbar/controls/Btn.tsx
```diff
@@ -0,0 +1,47 @@
...
+  label: string;
+  icon?: IconType;
+  iconOnly?: boolean;
+  small?: boolean;
+  raised?: boolean;
+  onClick: () => void;
+  disabled?: boolean;
+}
+
+function Btn(props: Props) {
```
axelboc, line 15: This is unrelated (see commits) -- I'm just abstracting a `Btn` component out of `ToggleBtn` and I expose it to @h5web/lib.

src/h5web/toolbar/controls/DownloadBtn.tsx
```diff
@@ -0,0 +1,53 @@
...
+  label: string;
+  icon?: IconType;
+  iconOnly?: boolean;
+  // If specified, browser will prompt user to download the file instead of opening a new tab
+  filename?: string;
+  // Invoked on click; must return URL to set as `href`, or `false` to do nothing
+  getDownloadUrl: () => string | false;
+}
+
+function DownloadBtn(props: Props) {
```
axelboc, line 15: The CSV export button behaves the same way as the `SnapshotBtn`, so I thought I'd make an abstraction as well.

src/h5web/toolbar/MatrixToolbar.tsx
```diff
@@ -0,0 +1,39 @@
+import { FiDownload } from 'react-icons/fi';
+import { useMatrixConfig } from '../vis-packs/core/matrix/config';
+import DownloadBtn from './controls/DownloadBtn';
+import Toolbar from './Toolbar';
+
+function MatrixToolbar() {
+  const currentSlice = useMatrixConfig((state) => state.currentSlice);
```
axelboc, line 7: The toolbar needs access to the `ndarray` of the current slice. That's why I need a zustand state, which I set in `MappedMatrixVis`, via a `useEffect` (just like `dataDomain`).

loichuder (changes requested):
The export does not work for 1D data so you can either change the `throw` condition to reflect this or change the algo to handle 1D data.

src/h5web/toolbar/MatrixToolbar.tsx
```diff
@@ -0,0 +1,39 @@
...
+  if (currentSlice && currentSlice.shape.length > 2) {
+    throw new Error('Expected current slice to have at most two dimensions');
```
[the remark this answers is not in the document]
axelboc: Silly me... 😓

src/h5web/toolbar/MatrixToolbar.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <DownloadBtn
+          icon={FiDownload}
+          label="CSV"
+          filename="export.csv"
+          getDownloadUrl={() => {
+            let data = '';
+            for (let i = 0; i < currentSlice.shape[0]; i++) {
+              let line = '';
+              for (let j = 0; j < currentSlice.shape[1]; j++) {
+                line += `${currentSlice.get(i, j).toString()},`; // complex numbers are stringifyied as two values
```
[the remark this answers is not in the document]
axelboc: Yeah not sure. Two values is simpler as we don't have to check whether each value is an array or not (we don't currently have access to the dataset's type in the toolbar) -- `toString()` just outputs `a,b` when the value is of type `[number, number]`. The other advantage, I find, is for users: they can do phase/amplitude/... computations on the data without having to parse values of the form `a+b*i`.

The downside is that users end up with twice as many columns as they may expect and we don't make that clear anywhere. However, I think it's pretty easy to understand by just comparing a few values between the file and H5Web.

src/h5web/vis-packs/core/matrix/utils.ts
```diff
@@ -21,3 +23,29 @@ export function getFormatter(
...
+    return csv;
+  }
+
+  if (slice.shape.length === 2) {
+    for (let i = 0; i < slice.shape[0]; i++) {
+      let line = '';
+      for (let j = 0; j < slice.shape[1]; j++) {
+        line += `${slice.get(i, j).toString()},`; // complex numbers are stringifyied as two values
+      }
+      csv += `${line.replace(/,$/u, '')}\n`;
```
loichuder, line 44:
```suggestion
      csv += `${line.slice(0, -1)}\n`;
```
or
```suggestion
      csv += line.replace(/,$/u, '\n');
```

src/h5web/vis-packs/core/matrix/utils.ts
```diff
@@ -21,3 +23,29 @@ export function getFormatter(
...
+
+export function sliceToCsv(slice: NdArray<Primitive<PrintableType>>): string {
+  let csv = '';
+
+  if (slice.shape.length === 1) {
+    for (let i = 0; i < slice.shape[0]; i++) {
+      csv += `${slice.get(i).toString()}\n`; // complex numbers are stringifyied as two values
+    }
+
+    return csv;
```
loichuder, line 35: Alternatively, you can make use of `slice.data` for 1D slices:
```suggestion
    return slice.data.reduce((acc, val) => acc += `${val.toString()}\n`;
```

and keep `let csv = '';` inside `if (slice.shape.length === 2)` block.

loichuder (approved):

src/h5web/vis-packs/core/matrix/utils.ts
axelboc: TypeScript doesn't like it, even with a type cast on `slice.data` 😢 -- maybe once the typing improvements of `ndarray` are merged...

src/h5web/vis-packs/core/matrix/utils.ts
axelboc: Ha, clever 😄 -- I'll use the second one, as I like the self-documenting aspect of the regex.

PR #744, opened 2021-07-07
Implement Square root scale

### LINKED ISSUE #740: [Heatmap] Add square root scale for colormap

loichuder:
Fix #740 

This includes some refactorings. There was several problems with our implementation of scales:
1. **Adding a new scale requires to modify many files** as scale parameters (labels, icons, shaders...) are scattered
2. **The treatment of unsupported values is highly specific to the log** (i.e. supported value means `value > 0`) hampering the addition of scales with other supported values (namely, the square root that supports `value >= 0`) 
3. Combination of the two points above: **Treatment of unsupported values is scattered in many places**.

I address 1. by adding a single record `H5WEB_SCALES` that holds all the scale parameters and is indexed by the `scaleType`.

I address 2. by moving the `support` logic inside the record `H5WEB_SCALES`. I introduce `supportedDomain` which is the max extension supported domain and `supportsDomain` which tests if a domain is supported by the scale.

I address 3. by using the new `supportedDomain` and `supportsDomain` utilities in the appropriate functions: `clampBound`, `extendDomain` and `getValidDomainForScale`.

Adding the square root is then done by adding a new entry to `H5WEB_SCALES`.

### DIFF
--- a/src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
+++ b/src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
@@ -7,7 +7,8 @@ import {
   RGBFormat,
   UnsignedByteType,
 } from 'three';
-import { Domain, ScaleType } from '../models';
+import type { Domain } from '../models';
+import { ScaleType } from '../models';
 import { useAxisSystemContext } from '../shared/AxisSystemContext';
 import VisMesh from '../shared/VisMesh';
 import type { ColorMap } from './models';
@@ -25,19 +26,54 @@ interface Props {
   alphaDomain?: Domain;
 }
 
+const CMAP_SIZE = 256;
+
 const SCALE_FUNC: Record<ScaleType, (val: number) => number> = {
   [ScaleType.Linear]: (val) => val,
   [ScaleType.Log]: Math.log10,
-  [ScaleType.SymLog]: (val) => {
-    return Math.sign(val) * Math.log10(1 + Math.abs(val));
-  },
+  [ScaleType.SymLog]: (val) => Math.sign(val) * Math.log10(1 + Math.abs(val)),
+  [ScaleType.Sqrt]: Math.sqrt,
 };
 
-const CMAP_SIZE = 256;
-const CMAP_NORM: Record<ScaleType, number> = {
-  [ScaleType.Linear]: 0,
-  [ScaleType.Log]: 1,
-  [ScaleType.SymLog]: 2,
+const SCALE_SHADER: Record<ScaleType, string> = {
+  [ScaleType.Linear]: `
+    float scale(float value) {
+      return oneOverRange * (value - min); // Linear
+    }
+
+    bool isSupported(float value) {
+      return true;
+    }`,
+  [ScaleType.Log]: `
+    const float oneOverLog10 = 0.43429448190325176;
+    float scale(float value) {
+      return oneOverRange * (log(value) * oneOverLog10 - min); // Log
+    }
+
+    bool isSupported(float value) {
+      return value > 0.;
+    }`,
+  [ScaleType.SymLog]: `
+    const float oneOverLog10 = 0.43429448190325176;
+    float symlog(float x) {
+      return sign(x) * log(1. + abs(x)) * oneOverLog10;
+    }
+
+    float scale(float value) {
+      return oneOverRange * (symlog(value) - min);
+    }
+
+    bool isSupported(float value) {
+      return true;
+    }`,
+  [ScaleType.Sqrt]: `
+    float scale(float value) {
+      return oneOverRange * (sqrt(value) - min);
+    }
+
+    bool isSupported(float value) {
+      return value >= 0.;
+    }`,
 };
 
 function HeatmapMesh(props: Props) {
@@ -88,7 +124,6 @@ function HeatmapMesh(props: Props) {
       withAlpha: { value: alphaValues ? 1 : 0 },
       alpha: { value: alphaTexture },
       colorMap: { value: colorMapTexture },
-      scaleType: { value: CMAP_NORM[scaleType] },
       min: { value: scaledDomain[0] },
       alphaMin: { value: alphaDomain && alphaDomain[0] },
       oneOverRange: { value: 1 / (scaledDomain[1] - scaledDomain[0]) },
@@ -115,31 +150,16 @@ function HeatmapMesh(props: Props) {
       uniform float oneOverAlphaRange;
       uniform int withAlpha;
 
-      const float oneOverLog10 = 0.43429448190325176;
       const vec4 nanColor = vec4(255, 255, 255, 1);
 
       varying vec2 coords;
 
-      float symlog(float x) {
-        return sign(x) * log(1. + abs(x)) * oneOverLog10;
-      }
-
-      float scale(float value) {
-        if (scaleType == 1) {
-          return oneOverRange * (log(value) * oneOverLog10 - min); // Log
-        }
-
-        if (scaleType == 2) {
-          return oneOverRange * (symlog(value) - min); // SymLog
-        }
-
-        return oneOverRange * (value - min); // Linear
-      }
+      ${SCALE_SHADER[scaleType]}
 
       void main() {
         float value = texture2D(data, coords).r;
 
-        if (scaleType == 1 && value <= 0.) {
+        if (!isSupported(value)) {
           gl_FragColor = nanColor;
         } else {
           gl_FragColor = texture2D(colorMap, vec2(scale(value), 0.5));
--- a/src/h5web/vis-packs/core/models.ts
+++ b/src/h5web/vis-packs/core/models.ts
@@ -12,6 +12,7 @@ export enum ScaleType {
   Linear = 'linear',
   Log = 'log',
   SymLog = 'symlog',
+  Sqrt = 'sqrt',
 }
 
 export interface Size {
@@ -25,14 +26,14 @@ export type CustomDomain = [number | null, number | null]; // `null` for persist
 
 export interface DomainErrors {
   minGreater?: boolean;
-  minError?: DomainError.InvalidMinWithLog | DomainError.CustomMaxFallback;
-  maxError?: DomainError.InvalidMaxWithLog;
+  minError?: DomainError.InvalidMinWithScale | DomainError.CustomMaxFallback;
+  maxError?: DomainError.InvalidMaxWithScale;
 }
 
 export enum DomainError {
   MinGreater = 'min-greater',
-  InvalidMinWithLog = 'invalid-min-with-log',
-  InvalidMaxWithLog = 'invalid-max-with-log',
+  InvalidMinWithScale = 'invalid-min-with-scale',
+  InvalidMaxWithScale = 'invalid-max-with-scale',
   CustomMaxFallback = 'custom-max-fallback',
 }
 
@@ -46,7 +47,7 @@ export interface AxisConfig {
 }
 
 export type AxisScale = PickD3Scale<
-  ScaleType.Linear | ScaleType.Log | ScaleType.SymLog,
+  ScaleType.Linear | ScaleType.Log | ScaleType.SymLog | ScaleType.Sqrt,
   number,
   number
 >;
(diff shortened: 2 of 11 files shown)


src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
```diff
@@ -25,20 +26,7 @@ interface Props {
...
-    return Math.sign(val) * Math.log10(1 + Math.abs(val));
-  },
-};
-
 const CMAP_SIZE = 256;
-const CMAP_NORM: Record<ScaleType, number> = {
-  [ScaleType.Linear]: 0,
-  [ScaleType.Log]: 1,
-  [ScaleType.SymLog]: 2,
-};
```
loichuder, line 41: This was replaced by an explicit implementation of the scale shader in the h5web scale entry.

src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
```diff
@@ -115,31 +103,16 @@ function HeatmapMesh(props: Props) {
...
-
-        return oneOverRange * (value - min); // Linear
-      }
+      ${h5webScale.shaderImplementation}
 
       void main() {
         float value = texture2D(data, coords).r;
 
-        if (scaleType == 1 && value <= 0.) {
+        if (!isSupported(value)) {
```
loichuder, line 115: Again, making the "support" logic more generic.

src/h5web/vis-packs/core/models.ts
```diff
@@ -25,14 +26,14 @@ export type CustomDomain = [number | null, number | null]; // `null` for persist
 
 export interface DomainErrors {
   minGreater?: boolean;
-  minError?: DomainError.InvalidMinWithLog | DomainError.CustomMaxFallback;
-  maxError?: DomainError.InvalidMaxWithLog;
+  minError?: DomainError.MinNotSupportedByScale | DomainError.CustomMaxFallback;
+  maxError?: DomainError.MaxNotSupportedByScale;
```
loichuder, line 30: Removing ref to `Log` as the error could arise with `Sqrt`

src/h5web/vis-packs/core/scales.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  supportsDomain: () => true,
+  shaderImplementation: `
+  float scale(float value) {
+    return oneOverRange * (value - min); // Linear
+  }
+
+  bool isSupported(float value) {
+    return true;
+  }`,
```
loichuder, line 43: I tried to keep the `shaderImplementation` to the bare minimum. Unfortunately, this implies some implicit coupling as one needs to know which variables are defined and which functions needs to be defined by the shader. 

Open to suggestions on this.

src/h5web/vis-packs/core/scales.ts
```diff
@@ -0,0 +1,113 @@
...
+  supportsDomain: (domain: Domain) => boolean;
+  shaderImplementation: string;
+}
+
+const linearScale: H5WebScale = {
+  createScale: (config) => scaleLinear<number>(config),
+  Icon: MdSort,
+  label: 'Linear',
+  scalingFunction: (val) => val,
+  supportedDomain: [-Number.MAX_VALUE / 2, Number.MAX_VALUE / 2],
```
loichuder, line 33: This is used as extreme bounds in `clampBound`

src/h5web/vis-packs/core/utils.ts
```diff
@@ -50,16 +50,7 @@ export function createAxisScale(
...
-    case ScaleType.Linear:
-      return scaleLinear<number>(config);
-    case ScaleType.Log:
-      return scaleLog<number>(config);
-    case ScaleType.SymLog:
-      return scaleSymlog<number>(config);
-    default:
-      throw new Error('Unknown type');
-  }
+  return H5WEB_SCALES[type].createScale(config);
```
loichuder, line 53: I kept `createAxisScale` to reduce the diff but we could remove it in favour of a direct call to `H5WEB_SCALES[type].createScale(config)`

src/h5web/vis-packs/core/utils.ts
```diff
@@ -157,18 +148,13 @@ function extendEmptyDomain(
...
-    return (Math.sign(val) * Number.MAX_VALUE) / 2; // max domain length is `Number.MAX_VALUE`
-  }
+export function clampBound(
+  val: number,
+  supportedDomain: Domain = [-Number.MAX_VALUE / 2, Number.MAX_VALUE / 2]
+): number {
+  const [supportedMin, supportedMax] = supportedDomain;
 
-  return val;
+  return clamp(val, supportedMin, supportedMax);
```
loichuder, line 157: `clampBound` is now a simple `clamp` wrapper with optional `supportedDomain`

src/h5web/vis-packs/core/utils.ts
```diff
@@ -350,13 +343,20 @@ export function getValidDomainForScale(
...
+  if (h5webScale.supportsDomain([min, max])) {
     return [min, max];
   }
 
-  // Clamp domain minimum to first positive value,
-  // or return `undefined` if domain is not unsupported: `[-x, 0]`
-  return positiveMin !== Infinity ? [positiveMin, max] : undefined;
+  // Clamp domain minimum to first possible value
+  const [supportedMin] = h5webScale.supportedDomain;
+  const newMin = supportedMin > 0 ? positiveMin : supportedMin;
```
loichuder, line 354: Not really happy how this one turned out.

I want to use `positiveMin` in case of a log scale but `0` in case of a square root scale.

src/h5web/vis-packs/core/scales.ts
axelboc: It's the downside of having a centralised `Record`: it becomes difficult to know where all these properties (`supportedDomain`, `supportsDomain`, etc.) are used and what they are used for. All the functions are disconnected from the code that uses them.

With the vis def records `CORE_VIS` and `NEXUS_VIS`, this pattern works because the records are used in very few places, and most of the properties hold React components and providers.

Here, because the scales are used in so many places in the codebase, and because some of the things stored in the record are highly coupled with the caller code (e.g. `scalingFunction`, `supportsDomain`, `shaderImplementation`), it can be a bit difficult to understand.

src/h5web/vis-packs/core/utils.ts
axelboc: Hmm, I still like the abstraction, personally. Unless this is now called in very few places?

If we keep it, the problem is more the name... it's been wrong for a while now. Perhaps `createVisScale`? Let's change this in another PR, though.

src/h5web/vis-packs/core/utils.ts
axelboc: Very nice. I don't think it's worth having a `default` here, since `supportedDomain` is required in the scales record (which I like because it's explicit).

src/h5web/vis-packs/core/utils.ts
```diff
@@ -181,22 +167,29 @@ export function extendDomain(
...
+  const extendedDomain: Domain = [
+    clampBound(scale.invert(-extendFactor), supportedDomain),
+    clampBound(scale.invert(1 + extendFactor), supportedDomain),
+  ];
 
-  if (min <= 0 && isLog) {
-    throw new Error('Expected domain compatible with log scale');
+  const h5webScale = H5WEB_SCALES[scaleType];
+
+  if (!h5webScale.supportsDomain(domain)) {
```
axelboc, line 179: This function bothers me. For some scales, we have a `supportedDomain` and yet `supportsDomain: () => true`. This feels incoherent.

Also, I feel like this refactoring of `extendDomain` changes the behaviour of the function significantly: you create the scale and call it before checking if the scale is compatible. Not to mention that `min * max > 0` is not the same as checking `min <= 0`. Are you sure this refactoring doesn't have unintended consequences?

src/h5web/vis-packs/core/utils.ts
axelboc: Yeah, I feel like what needs changing is the condition used to find `positiveMin`, in the search for the domain. Here, you end up with `0` with the square root scale, even if the values don't contain `0`!

axelboc (changes requested):
I really like where this is going (the generalisation of the scale-related operations). However, I feel like you tried to get there a bit too fast.

What I really like:

- the generalisation of the error handling in the domain slider;
- the template string for the shader code that adds the right scale functions;
- the clamping of the domain with `supportedDomain` (although perhaps the name conflicts with `supportsDomain()` as per my comments);
- each scale defining its own `createScale` function.

What I'm not convinced with:

- the single, centralised record pattern, which adds too much distance between tightly coupled pieces of code;
- the refactorings of `extendDomain` and `getValueDomainForScale`, which I think might be wrong and/or are not very clear mostly because of the `supportsDomain` function.

I don't think separate records are a problem. It requires making changes in multiple places, but `Record<ScaleType, Whatever>` means that if you add a `ScaleType` and forget to add it to a record, TypeScript will scream at you.  It also allows each record to be closer to the code it's coupled with.

I think it would make sense to keep a "global" scale record with the label, icon and `createScale` function (perhaps also `supportedDomain`), as these are all clear and generic enough, but I think you should keep separate records (or even just functions with `if/return` blocks) for the rest, notably for the domain-related things and for the shader implementations (which should live in or near `ShaderMesh`).

src/h5web/vis-packs/core/utils.ts
axelboc: Looking more thoroughly into this, and as I mentioned in another comment, I feel like `MAX_VALUE` and `MIN_VALUE` should be an implementation detail of `clampBound`. Having them outside feels too brittle (e.g. risk of typo in the `supportedDomains` of one scale but not another...)

Moreover, is the clamping to 0 really needed for the square root scale in `extendDomain`? Can `extendDomain` ever generate a min value lower than 0? `extendEmptyDomain` can, but perhaps this can be dealt with there?

If I recall, the clamp to `MIN_VALUE` in log is required because the scale can output 0, but that's a very specific exception. Perhaps it is better to make this exception clear by keeping the code as it is? To me, the clamping code that needs more thinking is in `extendEmptyDomain`, so perhaps my suggestion of moving it to `extendDomain` is actually a bad idea.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -181,21 +166,24 @@ export function extendDomain(
...
+    throw new Error('Expected domain compatible with sqrt scale');
+  }
+
   if (min === max) {
     return extendEmptyDomain(min, extendFactor, scaleType);
   }
 
-  const scale = createAxisScale({ type: scaleType, domain, range: [0, 1] });
+  const { createScale, supportedDomain } = H5WEB_SCALES[scaleType];
+  const scale = createScale({ domain, range: [0, 1] });
```
axelboc, line 182: Since you're not using `createAxisScale` here, maybe it's not that interesting to keep it?

src/h5web/vis-packs/core/utils.ts
```diff
@@ -350,13 +338,15 @@ export function getValidDomainForScale(
...
+  if (scaleType === ScaleType.Log && min * max <= 0) {
+    return Number.isFinite(positiveMin) ? [positiveMin, max] : undefined;
+  }
+
+  if (scaleType === ScaleType.Sqrt && min * max < 0) {
+    return [0, max];
   }
 
-  // Clamp domain minimum to first positive value,
-  // or return `undefined` if domain is not unsupported: `[-x, 0]`
```
axelboc, line 358: Maybe keep that comment in the `if` clause for log above.

axelboc (changes requested):

src/h5web/vis-packs/core/utils.ts
```diff
@@ -180,23 +166,30 @@ export function extendDomain(
...
 
+  const extendedDomain = unsafeExtendDomain(domain, extendFactor, scaleType);
+
+  return [Math.max(validMin, extendedDomain[0]), extendedDomain[1]];
+}
+
+function unsafeExtendDomain(
+  domain: Domain,
+  extendFactor: number,
+  scaleType = ScaleType.Linear
```
axelboc, line 184: You don't need the default parameter here, since the function is local and used in `extendDomain` only.

axelboc (approved):
Looking great 🎉

PR #746, opened 2021-07-08
Show "NX" badge next to NeXus entities in Explorer

### LINKED ISSUE #503: [NeXus] Identification of NeXus groups in explorer
We need to add some sort of identification to NeXus groups in the explorer (icon, colour, text, etc.)

axelboc:
Fix #503 

A lot has changed since the issue was first created. The providers now fetch attributes of child groups, so it is now possible to identify NeXus groups in the explorer. However, we still cannot predict in advance the interpretation of the group to show a more specific icon, as this would require fetching the metadata of the signal dataset. Not convinced it's worth the coding complexity and extra request.

image

(Note that I've increased the vertical padding of the entities in the explorer slightly to give them more breathing room.)


src/h5web/vis-packs/nexus/utils.ts
```diff
@@ -20,35 +21,29 @@ import {
...
-    return longName;
-  }
-
-  const units = getAttributeValue(dataset, 'units');
-  if (units && typeof units === 'string') {
-    return `${dataset.name} (${units})`;
-  }
-
-  return dataset.name;
-}
```
axelboc, line 40: Just moving this utility down to the bottom, since it is chronologically called after all the others (not that it's a strict rule, it just feels more logical to have it after `getNxData` in this case).

src/h5web/vis-packs/nexus/utils.ts
```diff
@@ -20,35 +21,29 @@ import {
...
+export function assertNxDataGroup(group: Group): void {
   if (!isNxDataGroup(group)) {
     throw new Error('Expected NXdata group');
   }
 }
 
+export function isNxGroup(entity: Entity): boolean {
+  return (
+    isGroup(entity) &&
+    (isNxDataGroup(entity) || !!getAttributeValue(entity, 'default'))
```
loichuder, line 43: I think it should also check that the entity has a valid `NX_class` attribute.

Someone might create HDF entities with attributes called `default` but that have nothing to do with NeXus.

loichuder (approved):
Looks great ! The `aria-label` of the badge is a nice touch :+1:

src/h5web/vis-packs/nexus/utils.ts
axelboc: This would change the behaviour for detecting NeXus visualizations. Are you sure?

axelboc: I think it's good idea, just wanted to confirm.

loichuder: It should not change anything for properly formatted NeXus file right ? `default` attributes are always on valid NX entities.

axelboc: Yes, but if I recall, we had made the decision to be more lenient towards wrongly-formatted files (e.g. missing `NXroot` class on the root entity).

loichuder: Oh, yes, I remember now ! Then, forget it :smile:

PR #749, opened 2021-07-09
Upgrade JupyterAPI to jupyterlab_hdf@0.7

### LINKED ISSUE #406: Fix type-inferring in the Jupyter provider

axelboc:
Fix #406. I started by merging the dev provider into the stable provider, and then refactored the whole thing to take advantage of the changes in jupyterlab_hdf@0.7, namely that the `/hdf/meta` endpoint now returns:

- the full metadata of the entity's attributes - https://github.com/jupyterlab/jupyterlab-hdf5/pull/87;
- if the entity is a group: 
  - the full metadata of the group's children - https://github.com/jupyterlab/jupyterlab-hdf5/pull/91;
  - the full metadata of the attributes of the group's children.

There's no `attributeCount` or `childrenCount`; the attributes and children are directly included in the response (but down to one nesting level only).

I also take advantage of the fact that `jupyterlab_hdf` now returns complex numbers as 2-value arrays. This allows us to remove the `complex.js` library! 🎉 

I've made a couple of fetching optimisations:

- The values of the attributes of the children of a group are fetched in parallel.
- The values of the attributes of an entity are cached so they are not fetched twice (once when processing the entity's parent, and again when processing the entity itself, either because of a NeXus `default` attribute or because the user navigated to both entities directly).

One potential optimisation remains, as it needs more discussion:

- The values of the attributes of an entity are always fetched after the entity's metadata, when they could be fetched in parallel. This is done so that the attributes values request is not made if the entity has no attributes (which we can know only by fetching the entity's metadata first). However, it could be argued that _always_ making two parallel requests is more efficient than _sometimes_ making two consecutive requests... It all depends on whether the HDF5 file contains lots of entities with or without attributes.

---

Before merging, the following environment variables will need to be updated on Netlify:

- `REACT_APP_JLAB_URL` (set to stable URL instead of dev)
- `REACT_APP_JLAB_DEV_ENABLED` (set to `false` or remove)


src/h5web/providers/jupyter/jupyter-api.ts
```diff
@@ -44,33 +41,32 @@ export class JupyterStableApi extends ProviderApi {
...
+  protected async fetchEntity(path: string): Promise<JupyterEntityResponse> {
+    const { data } = await this.client.get<JupyterEntityResponse>(
       `/meta/${this.filepath}?uri=${path}`
     );
     return data;
   }
 
-  protected async fetchContents(path: string): Promise<JupyterContentResponse> {
-    const { data } = await this.client.get<JupyterContentResponse>(
-      `/contents/${this.filepath}?uri=${path}`
```
axelboc, line 72: No more contents requests! 🎉

package.json
```diff
@@ -37,7 +37,6 @@
     "@visx/shape": "^1.14.0",
     "@visx/tooltip": "^1.7.2",
     "axios": "^0.21.1",
-    "complex.js": "^2.0.13",
```
axelboc, line 40: No more parsing of complex numbers!

src/h5web/providers/jupyter/jupyter-api.ts
```diff
@@ -4,36 +4,33 @@ import {
...
-import { assertDataset, hasComplexType } from '../../guards';
+import { assertDataset } from '../../guards';
 import { convertDtype, flattenValue } from '../utils';
+import { buildEntityPath } from '../../utils';
 
 export class JupyterStableApi extends ProviderApi {
-  /* API compatible with jupyterlab_hdf v0.5.1 */
+  protected attrValuesCache = new Map<string, JupyterAttrValuesResponse>();
+
+  /* API compatible with jupyterlab_hdf v0.7.0 */
```
loichuder, line 26: Actually, it is `v0.6.0`. `v0.7.0` is the current dev version.

loichuder (approved):
Great ! Thanks a lot :+1:

src/h5web/providers/jupyter/models.ts
```diff
@@ -1,50 +1,29 @@
...
-export type JupyterComplexValue =
-  | `(${'' | '-'}${number}${'+' | '-'}${number}j)`
-  | `${'' | '-'}${number}j`;
-
-export interface JupyterBaseEntity {
-  name: string;
-  path: string;
-  attributes: Attribute[];
-}
+export type JupyterAttrValuesResponse = Record<string, string>;
```
loichuder, line 29: Shouldn't it be `Record<string, unknown>` ?

PR #751, opened 2021-07-09
Upgrade Storybook

### LINKED ISSUE #696: Typescript 4.3 breaks the storybook
The error comes from `react-docgen-typescript`: https://github.com/styleguidist/react-docgen-typescript/issues/356. 

`react-docgen-typescript` is requested by `react-docgen-typescript-plugin` that is requested by `@storybook/react`.

The [new version of react-docgen-typescript](react-docgen-typescript) fixes this but we should perhaps revert to TS 4.2 while the new version goes up the full dep chain ?

axelboc:
Fixes #696 by upgrading Storybook and reverting #697.

- Release notes: https://github.com/storybookjs/storybook/releases/tag/v6.3.2 
- Also fixes an Alt+Tab issue on Windows that was bugging me 😄


loichuder (approved):
Nice :+1:

PR #752, opened 2021-07-12
Upgrade @types/ndarray

axelboc:
I was able to merge https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53728#issuecomment-878220724. Unfortunately, the PR did not receive any review from maintainers of `@types/ndarray` or `@types/numjs`. Let's hope people are happy with the changes once they upgrade...

Previously, `NdArray` or `NdArray<number>` meant that the underlying data could either be `number[]` or a typed array (`Float32Array` and the like), which is why we always had to cast: `myArr.data as number[]`.

The main change now is that the generic describes the exact type of the underlying data stored in the ndarray - e.g. `NdArray<number[]>` (note the `[]`), `NdArray<TypedArray>`, `NdArray<Float32Array>`, etc. With `NdArray<number[]>`, the type of `myArr.data` no longer includes typed arrays -- only `number[]` -- which means we can remove our type casts.

Note that we have to specify `NdArray<number[]>` because the default value for the generic is `Data<number>`, which replicates the old behaviour for backwards-compatibility.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,7 +61,7 @@
         "@types/d3-scale-chromatic": "^3.0.0",
         "@types/jest": "^26.0.23",
         "@types/lodash": "^4.14.170",
-        "@types/ndarray": "^1.0.9",
+        "@types/ndarray": "^1.0.10",
         "@types/node": "^15.12.5",
         "@types/react": "^17.0.11",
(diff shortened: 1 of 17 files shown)


loichuder (approved):
Sweet !

PR #755, opened 2021-07-16
Fix position of TooltipMesh when X/Y scales are not linear

### LINKED ISSUE #748: [Line] Tooltip bounds broken in log
### Describe the bug

When visualizing a line In log scale, the tooltip shows up only in the top-right quarter of the visualization.

### To Reproduce

1. Go to https://h5web.panosc.eu/
2. Hover over the default NX spectrum visualization

loichuder:
Fix #748

### DIFF
--- a/src/h5web/vis-packs/core/shared/VisMesh.tsx
+++ b/src/h5web/vis-packs/core/shared/VisMesh.tsx
@@ -9,22 +9,10 @@ interface Props extends MeshProps {
 function VisMesh(props: Props) {
   const { children, ...meshProps } = props;
 
-  const {
-    abscissaConfig,
-    ordinateConfig,
-    abscissaScale,
-    ordinateScale,
-    visSize,
-  } = useAxisSystemContext();
-
-  const [minAbscissa, maxAbscissa] = abscissaConfig.visDomain;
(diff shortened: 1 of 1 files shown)


src/h5web/vis-packs/core/shared/VisMesh.tsx
```diff
@@ -9,22 +9,10 @@ interface Props extends MeshProps {
...
-    abscissaScale,
-    ordinateScale,
-    visSize,
-  } = useAxisSystemContext();
-
-  const [minAbscissa, maxAbscissa] = abscissaConfig.visDomain;
-  const [minOrdinate, maxOrdinate] = ordinateConfig.visDomain;
-
-  const x = abscissaScale(minAbscissa + (maxAbscissa - minAbscissa) / 2);
-  const y = ordinateScale(minOrdinate + (maxOrdinate - minOrdinate) / 2);
```
loichuder, line 24: Not only this calculation to find the center of the vis is wrong for non-linear scales, but it is completely indeed as since https://github.com/silx-kit/h5web/pull/705, the center of the vis correspond to the center of the canvas by construction.

Therefore, we can remove the calculation and simply set the `position` to the center of the canvas `[0, 0, 0]`

axelboc (approved):

PR #759, opened 2021-07-19
Upgrade RTL and fix tests

axelboc:
Upgrading React Testing Library to v12 (and therefore DOM Testing Library to v8), a bunch of tests started to throw React's famous "update not wrapped in `act` warnings".

After much investigation, it seems that the root cause is `userEvent.keyboard('{SomeKey}')`. When I replace these calls with `pressEvent('SomKey')`, the warnings go away. `pressEvent` is a small little utility that fires a `keydown` event followed by a `keyup` event; I had written it to be able to trigger multiple `keydown` events in a row to simulate a long key press. My guess is that `userEvent.keyboard()` does something more than `keydown`/`keyup` (probably `keypress`) that triggers extra state updates in React Slider. I couldn't figure out exactly why this happens, unfortunately.

A couple of tests were also failing because of asynchronous issues, which I solved easily with `waitFor`.


src/h5web/toolbar/controls/DomainSlider/DomainSlider.test.tsx
```diff
@@ -57,19 +57,17 @@ test('show min/max and data range in tooltip', async () => {
...
+  // Hover min thumb to reveal tooltip
+  userEvent.hover(minThumb);
   const minInput = screen.getByLabelText('min');
   const maxInput = screen.getByLabelText('max');
 
-  // Move thumb one step to the left
-  userEvent.keyboard('{ArrowLeft}');
+  // Move min thumb one step to the left with keyboard
+  minThumb.focus();
+  pressKey('ArrowLeft'); // `userEvent.keyboard('{ArrowLeft}');` causes `act` warnings
```
axelboc, line 70: Just re-ordering things a bit here. This line is the relevant change.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -50,20 +50,20 @@ function DomainSlider(props: Props) {
     toggleEditingMax(force);
   }
 
-  const rootRef = useRef(null);
-  const tooltipRef = useRef<DomainTooltipHandle>(null);
-
-  useClickOutside(rootRef, () => {
+  function cancelEditing() {
```
axelboc, line 53: Introducing a `cancelEditing` function called from both `useClickOutside` and `useKey` below. The goal is to avoid canceling the editing unnecessarily when pressing `Escape`. I did this to fix a test at first, but I'm not sure it's still needed. Regardless, I think the refactoring makes sense.

loichuder (approved):
:clap:

PR #760, opened 2021-07-21
Implement gamma scale

### LINKED ISSUE #741: [Heatmap] Add gamma scale for colormaps
To apply to scale:
- 1. Normalize data from colormap range to [0, 1]
- 2. Apply a gamma factor: value**gamma

Issue is that compared to other scales, there is a gamma parameter.

Example shader implementation: https://github.com/silx-kit/silx/blob/master/src/silx/gui/plot/backends/glutils/GLPlotImage.py#L195

loichuder:
Fix #741 

So messy :disappointed:

### DIFF
--- a/src/h5web/toolbar/ComplexToolbar.tsx
+++ b/src/h5web/toolbar/ComplexToolbar.tsx
@@ -14,6 +14,7 @@ import {
 } from '../vis-packs/core/complex/models';
 import Selector from './controls/Selector/Selector';
 import { useHeatmapConfig } from '../vis-packs/core/heatmap/config';
+import { ScaleType } from '../vis-packs/core/models';
 
 function ComplexToolbar() {
   const {
@@ -67,7 +68,16 @@ function ComplexToolbar() {
 
       <Separator />
 
-      <ScaleSelector value={scaleType} onScaleChange={setScaleType} />
+      <ScaleSelector
+        value={scaleType}
+        onScaleChange={setScaleType}
+        options={[
+          ScaleType.Linear,
+          ScaleType.Log,
+          ScaleType.SymLog,
+          ScaleType.Sqrt,
+        ]}
+      />
 
       <Separator />
 
--- a/src/h5web/vis-packs/core/gammaScale.test.ts
+++ b/src/h5web/vis-packs/core/gammaScale.test.ts
@@ -1,0 +1,150 @@
+import { gammaScale } from './gammaScale';
+
+describe('gammaScale', () => {
+  it('should initialize gamma scale', () => {
+    const scale = gammaScale({
+      domain: [-10, 10],
+      range: [0, 5],
+      exponent: 2,
+    });
+
+    expect(scale.domain()).toEqual([-10, 10]);
+    expect(scale.range()).toEqual([0, 5]);
+    expect(scale.exponent()).toEqual(2);
+  });
+
+  it('should initialize gamma scale with default properties', () => {
+    const scale = gammaScale();
+
+    expect(scale.domain()).toEqual([0, 1]);
+    expect(scale.range()).toEqual([0, 1]);
+    expect(scale.exponent()).toEqual(1);
+  });
+
+  it('should scale accordingly', () => {
+    const scale = gammaScale({
+      domain: [40, 50],
+      range: [0, 10],
+      exponent: 2,
+    });
+    expect(scale(40)).toEqual(0);
+    expect(scale(50)).toEqual(10);
+    expect(scale(42)).toBeCloseTo(0.4);
+  });
+
+  it('should support changing the domain', () => {
+    const scale = gammaScale({
+      domain: [0, 1],
+      range: [0, 10],
+      exponent: 2,
+    });
+
+    scale.domain([40, 50]);
+    expect(scale.domain()).toEqual([40, 50]);
+    expect(scale(40)).toEqual(0);
+    expect(scale(50)).toEqual(10);
+    expect(scale(42)).toBeCloseTo(0.4);
+  });
+
+  it('should support changing the range', () => {
+    const scale = gammaScale({
+      domain: [10, 20],
+      range: [0, 10],
+      exponent: 2,
+    });
+
+    scale.range([-5, 5]);
+    expect(scale.range()).toEqual([-5, 5]);
+    expect(scale(10)).toEqual(-5);
+    expect(scale(20)).toEqual(5);
+    expect(scale(16)).toBeCloseTo(-1.4);
+  });
+
+  it('should support changing the exponent', () => {
+    const scale = gammaScale({
+      domain: [10, 20],
+      range: [-5, 5],
+      exponent: 2,
+    });
+
+    scale.exponent(4);
+    expect(scale.exponent()).toEqual(4);
+    expect(scale(10)).toEqual(-5);
+    expect(scale(20)).toEqual(5);
+    expect(scale(18)).toBeCloseTo(-0.904);
+  });
+
+  it('should support changing to a rounding range', () => {
+    const scale = gammaScale({
+      domain: [10, 20],
+      range: [0, 1],
+      exponent: 4,
+    });
+
+    scale.rangeRound([-5, 5]);
+    expect(scale.range()).toEqual([-5, 5]);
+    expect(scale(10)).toEqual(-5);
+    expect(scale(20)).toEqual(5);
+    expect(scale(18)).toEqual(-1);
+  });
+
+  it('should invert accordingly', () => {
+    const scale = gammaScale({
+      domain: [40, 50],
+      range: [0, 10],
+      exponent: 2,
+    });
+    expect(scale.invert(0)).toEqual(40);
+    expect(scale.invert(10)).toEqual(50);
+    expect(scale.invert(0.4)).toEqual(42);
+  });
+
+  it('should clamp', () => {
+    const scale = gammaScale({
+      domain: [0, 1],
+      range: [0, 10],
+      exponent: 2,
+      clamp: true,
+    });
+    expect(scale(5)).toEqual(10);
+  });
+
+  it('should extend to a nice domain', () => {
+    const scale = gammaScale({
+      domain: [0.58, 5.98],
+      range: [0, 10],
+      exponent: 2,
+    });
+
+    scale.nice();
+    expect(scale.domain()).toEqual([0.5, 6]);
+  });
+
+  it('should return nice ticks values', () => {
+    const scale = gammaScale({
+      range: [0, 10],
+      exponent: 2,
+    });
+
+    scale.domain([-5.5, 5]);
+    expect(scale.ticks(5)).toEqual([-4, -2, 0, 2, 4]);
+  });
+
+  it('should return tick format', () => {
+    const scale = gammaScale({
+      domain: [0, 1],
+      range: [0, 10],
+      exponent: 2,
+    });
+    let formatter = scale.tickFormat(5);
+    // Ticks are rounded to the first decimal
+    expect(formatter(-3.4)).toEqual('−3.4');
+    expect(formatter(2.856)).toEqual('2.9');
+
+    scale.domain([-5.5, 5]);
+    formatter = scale.tickFormat(5);
+    // Ticks are rounded to the nearest integer
+    expect(formatter(-3.4)).toEqual('−3');
+    expect(formatter(2.856)).toEqual('3');
+  });
+});
--- a/src/h5web/vis-packs/core/gammaScale.ts
+++ b/src/h5web/vis-packs/core/gammaScale.ts
@@ -1,0 +1,175 @@
+import { scalePower } from '@visx/scale';
+import type { Domain, GammaScaleConfig } from './models';
+import { interpolateNumber, interpolateRound } from 'd3-interpolate';
+import type {
+  InterpolatorFactory,
+  NumberValue,
+  ScalePower,
+  UnknownReturnType,
+} from 'd3-scale';
+import { isNumber } from 'lodash';
+
+function normalize(a: number, b: number) {
+  const extent = b - a;
+  return (x: number) => (extent !== 0 ? (x - a) / extent : 0.5);
+}
+
+type Range = number;
+type Output = Range;
+type Unknown = never;
+export type GammaScale = ScalePower<Range, Output, Unknown>;
+
+export function gammaScale(config?: GammaScaleConfig): GammaScale {
+  let _domain: Domain = config?.domain || [0, 1];
+  let _range: Domain = config?.range || [0, 1];
+  let _exponent = config?.exponent ?? 1;
+  let _clamp = config?.clamp || false;
+  let _interpolate: InterpolatorFactory<Range, Output> = interpolateNumber;
+  let _unknown: Unknown;
+
+  const scaleFn = function (val: NumberValue) {
+    const x = isNumber(val) ? val : val.valueOf();
+
+    if (Number.isNaN(x)) {
+      return _unknown;
+    }
+
+    return _interpolate(..._range)(
+      normalize(..._domain)(clamper(x)) ** _exponent
+    );
+  };
+
+  function clamper(x: number) {
+    if (!_clamp) {
+      return x;
+    }
+    const [d0, d1] = _domain;
+    return Math.max(d0, Math.min(d1, x));
+  }
+
+  function domain(): Domain;
+  function domain(d: Domain): typeof scale;
+  function domain(d?: Domain) {
+    if (d) {
+      _domain = d;
+      return scale;
+    }
+    return [..._domain];
+  }
+
+  function range(): Domain;
+  function range(r: Domain): typeof scale;
+  function range(r?: Domain) {
+    if (r) {
+      _range = r;
+      return scale;
+    }
+    return [..._range];
+  }
+
+  function rangeRound(): Domain;
+  function rangeRound(r: Domain): typeof scale;
+  function rangeRound(r?: Domain) {
+    if (r) {
+      _range = r;
+      _interpolate = interpolateRound;
+      return scale;
+    }
+    return [..._range];
+  }
+
+  function clamp(): boolean;
+  function clamp(c: boolean): typeof scale;
+  function clamp(c?: boolean) {
+    if (c) {
+      _clamp = c;
+      return scale;
+    }
+    return _clamp;
+  }
+
+  function unknown(): UnknownReturnType<Unknown, never>;
+  function unknown() {
+    return _unknown;
+  }
+
+  function interpolate(): InterpolatorFactory<Range, Output>;
+  function interpolate(i: InterpolatorFactory<Range, Output>): typeof scale;
+  function interpolate(i?: InterpolatorFactory<Range, Output>) {
+    if (i) {
+      _interpolate = i;
+      return scale;
+    }
+    return _interpolate;
+  }
+
+  function exponent(): number;
+  function exponent(e: number): typeof scale;
+  function exponent(e?: number) {
+    if (e) {
+      _exponent = e;
+      return scale;
+    }
+    return _exponent;
+  }
+
+  function invert(val: Output) {
+    return clamper(
+      interpolateNumber(..._domain)(
+        normalize(..._range)(val) ** (1 / _exponent)
+      )
+    );
+  }
+
+  function _createGammaLikeScale() {
+    return scalePower({
+      domain: _domain,
+      range: _range,
+      exponent: _exponent,
+      clamp: _clamp,
+    });
+  }
+
+  function nice(count?: number | undefined) {
+    const scaleLike = _createGammaLikeScale().nice(count);
+    _domain = scaleLike.domain() as Domain;
+    return scale;
+  }
+
+  function ticks(count?: number | undefined) {
+    return _createGammaLikeScale().ticks(count);
+  }
+
+  function tickFormat(
+    count?: number | undefined,
+    specifier?: string | undefined
+  ) {
+    return _createGammaLikeScale().tickFormat(count, specifier);
+  }
+
+  function copy() {
+    return gammaScale()
+      .domain(_domain)
+      .range(_range)
+      .exponent(_exponent)
+      .clamp(_clamp)
+      .interpolate(_interpolate);
+  }
+
+  const scale = Object.assign(scaleFn, {
+    domain,
+    range,
+    rangeRound,
+    clamp,
+    interpolate,
+    invert,
+    unknown,
+    exponent,
+    nice,
+    ticks,
+    tickFormat,
+    copy,
+  });
+
+  return scale;
+}
--- a/src/h5web/vis-packs/core/heatmap/ColorBar.tsx
+++ b/src/h5web/vis-packs/core/heatmap/ColorBar.tsx
@@ -3,14 +3,14 @@ import { useMeasure } from '@react-hookz/web';
 import { adaptedNumTicks, createAxisScale } from '../utils';
 import styles from './ColorBar.module.css';
 import { getInterpolator, getLinearGradient } from './utils';
-import type { ScaleType, Domain } from '../models';
-import type { ColorMap } from './models';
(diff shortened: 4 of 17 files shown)


src/h5web/toolbar/ComplexToolbar.tsx
```diff
@@ -67,7 +68,16 @@ function ComplexToolbar() {
...
-      <ScaleSelector value={scaleType} onScaleChange={setScaleType} />
+      <ScaleSelector
+        value={scaleType}
+        onScaleChange={setScaleType}
+        options={[
+          ScaleType.Linear,
+          ScaleType.Log,
+          ScaleType.SymLog,
+          ScaleType.Sqrt,
+        ]}
```
loichuder, line 79: `ScaleSelector` now accepts a `options` prop to restrict the scale options.

src/h5web/toolbar/LineToolbar.tsx
```diff
@@ -8,6 +8,9 @@ import Toolbar from './Toolbar';
 import Separator from './Separator';
 import ScaleSelector from './controls/ScaleSelector/ScaleSelector';
 import shallow from 'zustand/shallow';
+import { ScaleType } from '../vis-packs/core/models';
+
+const SCALETYPE_OPTIONS = [ScaleType.Linear, ScaleType.Log, ScaleType.SymLog];
```
loichuder, line 13: `Sqrt` was removed from the available scales for the X/Y axes

src/h5web/vis-packs/core/gammaScale.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    const scale = gammaScale({
+      domain: [-10, 10],
+      range: [0, 10],
+      exponent: 2,
+    });
+
+    scale.domain([0, 5]);
+    expect(scale.domain()).toEqual([0, 5]);
+
+    expect(scale(10)).toEqual(10);
```
loichuder, line 26: The gamma scale does not behave as a d3-scale. When changing the `domain` via `scale.domain`, the real domain of the scale is not changed even if `scale.domain` returns the new value :disappointed:

src/stories/HeatmapVis.stories.tsx
```diff
@@ -19,7 +19,23 @@ const logSafeDomain = getDomain(dataValues, ScaleType.Log);
...
+  const props = {
+    scaleType:
+      scaleType === ScaleType.Gamma
+        ? ([ScaleType.Gamma, gammaExponent] as [ScaleType.Gamma, number])
+        : scaleType,
+    ...otherArgs,
+  };
+
+  return <HeatmapVis {...props} />;
+};
```
loichuder, line 38: Some tweaking to be able to test the gamma scale in the Storybook

src/h5web/vis-packs/core/models.ts
```diff
@@ -46,11 +53,17 @@ export interface AxisConfig {
...
+  PickScaleConfig<'linear' | 'log' | 'symlog', number>,
+  'type'
 >;
+export type GammaScaleConfig = Omit<PickScaleConfig<'pow', number>, 'type'>;
+
+export type AxisScale =
+  | ScaleLinear<number, number>
+  | ScaleLogarithmic<number, number>
+  | ScaleSymLog<number, number>
+  | ScalePower<number, number>;
```
axelboc, line 66: You're missing `ScaleSqrt` I think

src/stories/HeatmapVis.stories.tsx
```diff
@@ -19,7 +19,23 @@ const logSafeDomain = getDomain(dataValues, ScaleType.Log);
...
+  Omit<HeatmapVisProps, 'scaleType'> & {
+    scaleType: ScaleType;
+    gammaExponent: number;
+  }
+> = (args) => {
+  const { scaleType, gammaExponent, ...otherArgs } = args;
+  const props = {
+    scaleType:
+      scaleType === ScaleType.Gamma
+        ? ([ScaleType.Gamma, gammaExponent] as [ScaleType.Gamma, number])
```
axelboc, line 32: `as const` maybe ?

src/stories/HeatmapVis.stories.tsx
```diff
@@ -19,7 +19,23 @@ const logSafeDomain = getDomain(dataValues, ScaleType.Log);
...
+        : scaleType,
+    ...otherArgs,
+  };
+
+  return <HeatmapVis {...props} />;
```
axelboc, line 37: Put the spread of `otherArgs` here, it's simpler:

```
return <HeatmapVis scaleType={scaleType} {...otherArgs} />;
```

axelboc (approved):
Apart from a few minor suggestions, I'm pretty happy with this first implementation as is. It doesn't add too much complexity, considering... I'll try to play with the code a bit at some point to see if I can think of ways to simplify or avoid some of the `isArray`/`isScaleType` checks.

src/h5web/vis-packs/core/scales.ts
```diff
@@ -3,47 +3,106 @@ import {
...
+  const powScale = scalePower<number>({
+    ...config,
+    domain: [0, 1],
+  });
+
+  function gamma(value: NumberValue): number {
+    return powScale(linScale(value));
+  }
+
+  const result = scalePower<number>(config);
```
axelboc, line 63: Why not use the linear scale below instead of creating another scale?

src/h5web/vis-packs/core/scales.ts
```diff
@@ -3,47 +3,106 @@ import {
...
+  return Object.assign(gamma, {
+    invert: (val: number) => linScale.invert(powScale.invert(val)),
+    domain: result.domain,
+    range: result.range,
```
axelboc, line 67:
```ts
domain: linScale.domain, // this will fix the issue you mention in the tests
range: powScale.domin,
```

src/h5web/vis-packs/core/scales.ts
```diff
@@ -3,47 +3,106 @@ import {
...
+    rangeRound: result.rangeRound,
+    clamp: result.clamp,
+    unknown: result.unknown,
+    interpolate: result.interpolate, // To be fixed ?
```
axelboc, line 71: yep 😄 - right now it doesn't work with a domain other than `[0, 1]`, I assume?

axelboc (changes requested):
Sorry, I had skipped the gamma scale implementation and forgot to get back to it.

src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
```diff
@@ -154,7 +190,7 @@ function HeatmapMesh(props: Props) {
 
       varying vec2 coords;
 
-      ${SCALE_SHADER[scaleType]}
+      ${SCALE_SHADER[isScaleType(scaleType) ? scaleType : ScaleType.Gamma]}
```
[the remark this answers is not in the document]
loichuder: With `isArray(scaleType) ? scaleType[0] : scaleType`, the second member has the type `Omit<ScaleType, 'gamma'>`.

TS doesn't allow this type to be used as index of a record with keys `ScaleType`.

I could cast `as ScaleType` or change the `Omit` into an union ?

src/h5web/vis-packs/core/models.ts
loichuder: `ScaleSqrt` is a particular `ScalePower`. I could add it explicitly tho.

src/stories/HeatmapVis.stories.tsx
loichuder: Get casted as `readonly`. 
Then TS complains that `readonly [ScaleType.Gamma, number]` cannot be set to `[ScaleType.Gamma, number]` when doing `<HeatmapVis {...props} />` -____-

src/h5web/vis-packs/core/scales.ts
loichuder: Because the linear scale needs a `range` of `[0, 1]` while I want the resulting scale to have the `domain` and `range` of the config.

src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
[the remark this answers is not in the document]
axelboc: lol

src/h5web/vis-packs/core/scales.ts
loichuder: Hm, could work, I will test

src/h5web/vis-packs/core/scales.ts
loichuder: Good call, I will extend the tests to see if this is true

src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
```diff
@@ -20,20 +22,37 @@ interface Props {
...
-  [ScaleType.Sqrt]: Math.sqrt,
-};
+function computeScaledDomain(
+  scaleType: Omit<ScaleType, 'gamma'> | ScaleType.Gamma,
+  domain: Domain
+): Domain {
+  switch (scaleType) {
+    case ScaleType.Linear:
+    case ScaleType.Gamma:
+      return domain;
```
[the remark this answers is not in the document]
loichuder: > Shouldn't the gamma scale have this scale function here: `val => Math.pow(val, exponent)`, and in the shader: `return oneOverRange * (pow(value, gammaExponent) - min));`
> 
> It's equivalent,

It is not equivalent: the `pow(value, gammaExponent)` must be done when `value` is in `[0,1]` which is not the case with the formula you wrote

src/h5web/vis-packs/core/scales.ts
```diff
@@ -3,47 +3,106 @@ import {
...
+): AxisScale;
+
+export function createScale(
+  type: ScaleType.Linear | ScaleType.Log | ScaleType.Sqrt | ScaleType.SymLog,
+  config: VisxScaleConfig
+): AxisScale;
+
+export function createScale(
+  type: ScaleType,
+  config: VisxScaleConfig | GammaScaleConfig
```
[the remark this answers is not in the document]
loichuder: About this and https://github.com/silx-kit/h5web/pull/760#discussion_r673823481, I removed most uses of `@visx/scale` utils as I got TS errors saying that `Omit<ScaleType, ScaleType.Gamma>` is not assignable to the keys of the `@visx/scale` scale record (`"band"` in particular)

src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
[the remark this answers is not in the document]
axelboc: The root of my comment is the fact that this `return domain;` is inconsistent with the rest. It shows that there's a lot of coupling between this function (and previously the `SCALE_FUNC` record) and the shader implementations for each scale. It might be clearer to do the necessary operations on the domain from inside the shader scale implementations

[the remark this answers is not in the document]
axelboc: Adding more operations into the shader is probably not what we want, so perhaps we could pass more uniforms regardless of whether they're actually needed? I don't know... Anyway, not hugely important.

src/h5web/vis-packs/core/gammaScale.ts
```diff
@@ -0,0 +1,151 @@
...
+    return _clamp;
+  }
+
+  function unknown(): UnknownReturnType<Unknown, never>;
+  // function unknown(u: Unknown): typeof scale;
+  function unknown() {
+    // if (u) {
+    //   _unknown = u;
+    //   return scale;
+    // }
```
loichuder, line 93: I could not make this work with TS :disappointed:

src/h5web/vis-packs/core/gammaScale.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+    return _exponent;
+  }
+
+  function invert(val: Output) {
+    const [d0, d1] = _domain;
+    const [r0, r1] = _range;
+
+    return clamper(
+      interpolateNumber(d0, d1)(normalize(r0, r1)(val) ** (1 / _exponent))
```
loichuder, line 122: Note that I use `interpolateNumber` here and not `_interpolate`. The rounding is indeed done when scaling from `domain` to `range` but not the other way round.

src/h5web/vis-packs/core/gammaScale.ts
```diff
@@ -0,0 +1,151 @@
...
+    );
+  }
+
+  const { nice, ticks, tickFormat } = scalePower({
+    domain: _domain,
+    range: _range,
+  });
```
loichuder, line 129: A small shortcut for functions I did not want to reimplement :sweat_smile:

src/h5web/vis-packs/core/heatmap/ColorBar.tsx
```diff
@@ -3,14 +3,14 @@ import { useMeasure } from '@react-hookz/web';
...
-import type { ColorMap } from './models';
 import { formatBound } from '../../../utils';
+import type { Domain } from '../models';
+import type { ColorMap, ScaleParams } from './models';
 import Tick from '../shared/Tick';
 
 interface Props {
   domain: Domain;
-  scaleType: ScaleType;
+  scaleType: ScaleParams;
```
loichuder, line 13: Should we rename this prop to `scaleParams` ? Also in `HeatmapVis` and `HeatmapMesh` ?

src/h5web/vis-packs/core/models.ts
```diff
@@ -46,12 +48,26 @@ export interface AxisConfig {
...
-  number,
   number
 >;
 
+export interface GammaScaleConfig {
+  domain?: Domain;
+  range?: Domain;
+  exponent?: number;
+  clamp?: boolean;
+}
```
loichuder, line 61: I can't use `PickScaleConfigWithoutType<'pow', number>` as `domain` is in this case of type `ContinuousDomain = (number | Date)[]` which is not compatible with `Domain`...

src/h5web/vis-packs/core/scales.ts
```diff
@@ -3,47 +3,59 @@ import {
...
 import SqrtIcon from '../../toolbar/controls/ScaleSelector/SqrtIcon';
 import MdGraphicEqRotated from '../../toolbar/controls/ScaleSelector/MdGraphicEqRotated';
+import { gammaScale } from './gammaScale';
 
 interface H5WebScale {
-  createScale: (
-    config: Omit<PickScaleConfig<ScaleType, number>, 'type'>
-  ) => AxisScale;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  createScale: (config: any) => AxisScale;
```
loichuder, line 18: Some improvement to be made here. I could not find a working solution.

I tried using `createScale` from `@visx` as well but I ran into other TS problems... (overrides were not working as intended and using a template resulted in a infinite TS recursion :shrug: )

src/stories/HeatmapVis.stories.tsx
```diff
@@ -71,6 +89,7 @@ Alpha.args = {
   alphaDomain,
 };
 
+export { Template as HeatmapVisTemplate };
```
loichuder, line 92: Exporting the `Template` to reuse the specific gamma controls in other Heatmap stories files.

Note that I use `excludeStories` below to remove it from the list of stories of this file.

loichuder (commented):
Much better with a proper implementation :relieved:

src/h5web/vis-packs/core/gammaScale.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  it('should clamp', () => {
+    const scale = gammaScale({
+      domain: [0, 1],
+      range: [0, 10],
+      exponent: 2,
+      clamp: true,
+    });
+    expect(scale(5)).toEqual(10);
+  });
+});
```
axelboc, line 65: I guess you should add some tests for the other methods.

src/h5web/vis-packs/core/gammaScale.ts
```diff
@@ -0,0 +1,151 @@
...
+  let _exponent = config?.exponent ?? 1;
+  let _clamp = config?.clamp || false;
+  let _interpolate: InterpolatorFactory<Range, Output> = interpolateNumber;
+  let _unknown: Unknown;
+
+  const scaleFn = function (val: NumberValue) {
+    const [d0, d1] = _domain;
+    const [r0, r1] = _range;
+    const x = isNumber(val) ? val : val.valueOf();
+
```
axelboc, line 34: You could add a `NaN` check and early-return `_unknown` here, to make use of it.

src/h5web/vis-packs/core/gammaScale.ts
axelboc: Makes sense since `Unknown = never`. We just don't support changing the `unknown`, which is fine. You can remove the commented code.

src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
```diff
@@ -114,6 +115,27 @@ const SCALE_SHADER: Record<ScaleType, ScaleShader> = {
...
+    uniform float gammaExponent;
+
+    float scale(float value) {
+      return pow(oneOverRange * (value - min), gammaExponent);
+    }
+
+    bool isSupported(float value) {
+      return true;
+    }`,
+  },
```
axelboc, line 138: Good stuff with the uniforms defined in the same object 💯 

Maybe just fix the indentation of the fragment string? The extra spaces don't matter anyway, so might as well make it more readable. Also, the `exponent` uniform can fit on a single line.

src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
```diff
@@ -114,6 +115,27 @@ const SCALE_SHADER: Record<ScaleType, ScaleShader> = {
       return value >= 0.;
     }`,
   },
+  [ScaleType.Gamma]: {
+    uniforms: (domain: Domain, exponent = 1) => ({
```
axelboc, line 119: Don't think the default value is worth it for the exponent here, since it is mandatory in the `ScaleParams` type.

src/h5web/vis-packs/core/heatmap/ColorBar.tsx
axelboc: I like the prop being named `scaleType`, because even with the gamma scale, you're still passing a scale type -- you just pass something else along with it.

What about `VisScaleType` instead of `ScaleParams`?

src/h5web/vis-packs/core/scales.ts
```diff
@@ -3,47 +3,59 @@ import {
...
   [ScaleType.Sqrt]: {
-    createScale: (config) => scaleSqrt<number>(config),
+    createScale: (config: PickScaleConfigWithoutType<ScaleType.Sqrt, number>) =>
+      scaleSqrt<number>(config),
     Icon: SqrtIcon,
     label: 'Square root',
     validMin: 0,
   },
+  [ScaleType.Gamma]: {
+    createScale: (config: GammaScaleConfig) => gammaScale(config),
```
axelboc, line 56: You can just pass `gammaScale` without invoking it here.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -42,11 +41,20 @@ const adaptedLogTicksThreshold = scaleLinear({
 });
 
 export function createAxisScale(
-  config: PickScaleConfig<ScaleType, number>
+  scaleParams: ScaleParams,
+  config: VisxScaleConfig | GammaScaleConfig
 ): AxisScale {
-  const { type } = config;
+  if (isScaleType(scaleParams) && scaleParams !== ScaleType.Gamma) {
```
axelboc, line 47: If you switch to `VisScaleType`, then don't forget to rename `scaleParams` to `scaleType` here. Also, I think `Gamma` is not possible here, since we omit it from the type.

src/h5web/vis-packs/core/utils.ts
```diff
@@ -42,11 +41,20 @@ const adaptedLogTicksThreshold = scaleLinear({
...
+    return H5WEB_SCALES[scaleParams].createScale(config);
+  }
 
-  return H5WEB_SCALES[type].createScale(config);
+  const [scaleType, exponent] = scaleParams;
+  if (scaleType !== ScaleType.Gamma) {
+    throw new Error(
+      'scaleParams should be an array only for Gamma scale type !'
+    );
+  }
```
axelboc, line 56: Same, I don't get how this can happen based on the type of `VisScaleType`.

src/stories/HeatmapVis.stories.tsx
axelboc: I think it may be simpler to keep this template unchanged and just override the `scaleType` prop to only allow `ScaleType` values. Then, you can change the gamma scale story so it doesn't use the default template and so the gamma scale type is hardcoded. You can then add the `gammaExponent` control in that `stories` file and use it in the gamme scale story.

axelboc (changes requested):

src/h5web/vis-packs/core/gammaScale.ts
```diff
@@ -0,0 +1,151 @@
...
+    return _interpolate(r0, r1)(normalize(d0, d1)(clamper(x)) ** _exponent);
```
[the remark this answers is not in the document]
loichuder: Unfortunately, this raises `TS2556` as the type of range `Range[]` is not equivalent to `[number, number]` which is required for the spread

src/h5web/vis-packs/core/heatmap/HeatmapMesh.tsx
loichuder: It was more to satisfy the type constraint of `ScaleShader`:

```ts
export interface ScaleShader {
  uniforms: (
    domain: Domain,
    exponent?: number
  ) => Record<string, { value: number }>;
  fragment: string;
}
```

I changed it to:
```ts
export interface ScaleShader {
  uniforms:
    | ((domain: Domain) => Record<string, { value: number }>)
    | ((domain: Domain, exponent: number) => Record<string, { value: number }>);
  fragment: string;
}
```

src/h5web/vis-packs/core/gammaScale.ts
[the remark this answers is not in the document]
axelboc: and I assume it doesn't work if you were to change the type of `Range` to be `[number, number]` to match what the scale actually supports?

src/h5web/vis-packs/core/utils.ts
loichuder: https://github.com/silx-kit/h5web/pull/760#discussion_r686076029 and this were to help TypeScript which really struggles to infer the types.

It does not matter that much in the end as `createScale` accepts `any`.

src/h5web/vis-packs/core/heatmap/ColorBar.tsx
loichuder: We could instead have `VisxScaleType` for `Linear | Log | Symlog | Sqrt` and have `ScaleType = VisxScaleType | 'gamma'` ?

src/h5web/vis-packs/core/gammaScale.ts
loichuder: Got it. I did it differently but I think it did the trick

src/stories/HeatmapVis.stories.tsx
loichuder: Not sure I understood fully. Have a look at the latest changes and see if that is what you wanted

axelboc: Yep, looks good 😉  - you no longer have to export the `Template`, I think. Also, you may have to add a custom control for the `gammeExponent` to the gamma scale story.

src/h5web/vis-packs/core/gammaScale.ts
axelboc: Smart 👍

axelboc (approved):
Just the `Template` export in the stories to remove, but then it's good to go!

PR #761, opened 2021-07-21
Refactor formatting of complex numbers

axelboc:
I got side-tracked working on the tooltip. 😄 I was trying to collocate all the D3 formatters in `h5web/utils` so we could have the format specifiers all in one place and keep our sanity.

Here, I do so for the complex formatters used by the scalar vis, matrix vis and metadata viewer. I refactor the code that creates these formatters so we no longer have to pass specifiers around multiple functions.

I also did the following to improve the display of the matrix vis:

- I replaced the fancy `i` for imaginary numbers with a simple `i` with a space before it. The fancy `i` just looked off. I find the normal `i` works better, especially with that extra space to give it some breathing room.
- I centered the values inside the cells of the Matrix vis, since the formatters (and monospace font) guarantee that they will always take the same amount of space).
- ... except when the real/imag part is equal to 0 and not displayed. So I fixed this by always displaying the real/imag part even when equal to 0 (in the matrix vis only). I think it makes the values more readable as a result, but let me know what you think.

image

For reference, here is what complex numbers look like now in the scalar vis and the metadata viewer:

image

image

### DIFF
--- a/src/h5web/metadata-viewer/utils.ts
+++ b/src/h5web/metadata-viewer/utils.ts
@@ -6,7 +6,7 @@ import {
   H5WebComplex,
   Shape,
 } from '../providers/models';
-import { formatComplex } from '../utils';
+import { formatScalarComplex } from '../utils';
 
 export function renderShape(shape: Shape): string {
   if (shape === null) {
@@ -38,12 +38,9 @@ export function renderType(type: DType): string {
   return type.class;
 }
 
-export function renderComplex(
-  complex: H5WebComplex | ComplexArray,
-  specifier?: string
-): string {
+export function renderComplex(complex: H5WebComplex | ComplexArray): string {
   if (isH5WebComplex(complex)) {
-    return formatComplex(complex, specifier);
+    return formatScalarComplex(complex);
   }
 
   return `[ ${complex.map((c) => renderComplex(c)).join(', ')} ]`;
--- a/src/h5web/utils.ts
+++ b/src/h5web/utils.ts
@@ -9,21 +9,27 @@ import type { NxAttribute } from './vis-packs/nexus/models';
 
 export const formatValue = format('.3~e');
 export const formatPreciseValue = format('.5~e');
-export function formatComplex(value: H5WebComplex, specifier = '') {
-  const [real, imag] = value;
-  const formatFunction = format(specifier);
+export const formatMatrixValue = format('.3e');
+export const formatMatrixComplex = createComplexFormatter('.2e', true);
+export const formatScalarComplex = createComplexFormatter('.12~g');
 
-  if (imag === 0) {
-    return `${formatFunction(real)}`;
-  }
+export function createComplexFormatter(specifier: string, full = false) {
+  const formatVal = format(specifier);
 
-  if (real === 0) {
-    return `${formatFunction(imag)}ⅈ`;
-  }
+  return (value: H5WebComplex) => {
+    const [real, imag] = value;
 
-  return `${formatFunction(real)}${
-    Math.sign(imag) === 1 ? ' + ' : ' − '
-  }${formatFunction(Math.abs(imag))}ⅈ`;
+    if (imag === 0 && !full) {
+      return `${formatVal(real)}`;
+    }
+
+    if (real === 0 && !full) {
+      return `${formatVal(imag)} i`;
+    }
+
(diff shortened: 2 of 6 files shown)


src/h5web/metadata-viewer/utils.ts
```diff
@@ -38,12 +38,9 @@ export function renderType(type: DType): string {
...
 }
 
-export function renderComplex(
-  complex: H5WebComplex | ComplexArray,
-  specifier?: string
-): string {
+export function renderComplex(complex: H5WebComplex | ComplexArray): string {
   if (isH5WebComplex(complex)) {
-    return formatComplex(complex, specifier);
+    return formatScalarComplex(complex);
```
axelboc, line 43: `renderComplex` is now only used to render complex attributes in the metadata viewer, so I use the scalar complex formatter (see below).

src/h5web/utils.ts
```diff
@@ -9,21 +9,27 @@ import type { NxAttribute } from './vis-packs/nexus/models';
...
-  const formatFunction = format(specifier);
+export const formatMatrixValue = format('.3e');
+export const formatMatrixComplex = createComplexFormatter('.2e', true);
+export const formatScalarComplex = createComplexFormatter('.12~g');
 
-  if (imag === 0) {
-    return `${formatFunction(real)}`;
-  }
+export function createComplexFormatter(specifier: string, forceFull = false) {
+  const formatVal = format(specifier);
```
axelboc, line 17: Another benefit of the refactoring: the formatter is now created only once instead of for every value.

src/h5web/vis-packs/core/matrix/utils.ts
```diff
@@ -1,24 +1,23 @@
...
+import { formatMatrixComplex, formatMatrixValue } from '../../../utils';
 import type { PrintableType, ValueFormatter } from '../models';
-import { formatNumber } from '../utils';
 
 export function getFormatter(
   dataset: Dataset<ArrayShape, PrintableType>
 ): ValueFormatter<PrintableType> {
   if (hasComplexType(dataset)) {
-    return (val) => renderComplex(val as H5WebComplex, '.2e');
+    return (val) => formatMatrixComplex(val as H5WebComplex);
```
axelboc, line 16: We don't need the recursion behaviour of `renderComplex`, so there's no point in using it here. It means I can use the `formatMatrixComplex` formatter here, and the `formatScalarComplex` formmater in `renderComplex`, which is a lot more explicit than to pass an optional `specifier`.

src/h5web/utils.ts
```diff
@@ -9,21 +9,27 @@ import type { NxAttribute } from './vis-packs/nexus/models';
...
-  const [real, imag] = value;
-  const formatFunction = format(specifier);
+export const formatMatrixValue = format('.3e');
+export const formatMatrixComplex = createComplexFormatter('.2e', true);
+export const formatScalarComplex = createComplexFormatter('.12~g');
 
-  if (imag === 0) {
-    return `${formatFunction(real)}`;
-  }
+export function createComplexFormatter(specifier: string, forceFull = false) {
```
loichuder, line 16:
```suggestion
export function createComplexFormatter(specifier: string, displayAll = false) {
```

`forceFull` is fine otherwise

src/h5web/utils.ts
```diff
@@ -9,21 +9,27 @@ import type { NxAttribute } from './vis-packs/nexus/models';
...
-  }
+  return (value: H5WebComplex) => {
+    const [real, imag] = value;
+
+    if (imag === 0 && !forceFull) {
+      return `${formatVal(real)}`;
+    }
+
+    if (real === 0 && !forceFull) {
+      return `${formatVal(imag)}ⅈ`;
```
loichuder, line 27: **Fancy i** doesn't want to go !

loichuder (approved):
Excellent initiative !

The only downside I would see to display `0` instead of nothing is that a very small number (`0.0000001`) could be confused for a true `0` if rounding/truncating. But as we use the exponential notation in the matrix, this cannot happen.

src/h5web/utils.ts
axelboc: `displayFull`?

loichuder: Sounds good. It is not that important anyway

axelboc: or just `full` even ?

loichuder: Good :+1:

PR #762, opened 2021-07-22
Move Heatmap stories in sub-folders

axelboc:


src/stories/HeatmapVis.stories.tsx
```diff
@@ -7,183 +7,70 @@ import {
...
-CustomEdgesWithExtension.args = {
-  dataArray,
-  domain,
-  abscissaParams: {
-    value: Array.from({ length: dataArray.shape[1] }, (_, i) => 100 + 10 * i),
-  },
-  ordinateParams: {
-    value: Array.from({ length: dataArray.shape[0] }, (_, i) => -5 + 0.5 * i),
-  },
-};
```
axelboc, line 116: merged with previous story

src/stories/HeatmapVisScales.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+  domain: logSafeDomain,
+  scaleType: ScaleType.Log,
+};
+
+export const SqrtScale = Template.bind({});
+SqrtScale.args = {
+  dataArray,
+  domain: logSafeDomain,
+  scaleType: ScaleType.Sqrt,
+};
```
axelboc, line 37: new story for the square root scale

src/stories/LineVis.stories.tsx
```diff
@@ -44,6 +44,16 @@ Domain.args = {
...
+export const Abscissas = Template.bind({});
+
+Abscissas.args = {
+  dataArray,
+  domain,
+  abscissaParams: {
+    value: Array.from({ length: dataArray.size }, (_, i) => -10 + 0.5 * i),
+  },
+};
+
```
axelboc, line 56: Moved up from `Display` folder, as I think it's as important a feature as the error bars and auxiliary arrays.

src/stories/LineVisScales.stories.tsx
```diff
@@ -41,12 +58,12 @@ LogForX.args = {
...
-SymLogForX.storyName = 'Sym Log for X';
-SymLogForX.args = {
+export const SqrtForX = Template.bind({});
+SqrtForX.storyName = 'Sqrt for X';
+SqrtForX.args = {
   dataArray: dataArrayForXLog,
   domain: domainForXLog,
-  scaleType: ScaleType.SymLog,
+  scaleType: ScaleType.Sqrt,
 };
```
axelboc, line 67: new stories for the square scale

src/stories/HeatmapVisScales.stories.tsx
loichuder: I was meaning to do this, thanks :slightly_smiling_face: !

You could use a dedicated `sqrtSafeDomain` instead of the `logSafeDomain` and make the story name clearer:
```SqrtScale.storyName = 'Square root scale';```

src/stories/LineVisScales.stories.tsx
loichuder: Not sure it is worth adding those as we will remove `Sqrt` from the scale available for axes ?

loichuder (approved):

axelboc:
> Not sure it is worth adding those as we will remove Sqrt from the scale available for axes ?

Ah right, forgot about that. 👍

src/stories/HeatmapVisScales.stories.tsx
axelboc: > You could use a dedicated sqrtSafeDomain instead of the logSafeDomain and make the story name clearer:

🎣 had done it for the line vis scale stories but not here.

PR #763, opened 2021-07-22
Add support for custom tooltips

### LINKED ISSUE #757: [Tooltip] Allow the user to change the tooltip rendering function
We could have a prop in the `Vis` called `renderTooltip` which would take in input the maximum information:
- Abscissa `x` and ordinate `y` (unformatted) of the current point
- The corresponding indices `xi` and `yi`
- The value of the `dataArray` at this point (unformatted)

and would return a `ReactElement`

axelboc:
`HeatmapVis` and `LineVis` both accept a `renderTooltip` function that receives the tooltip's current coordinates, indices and abscissa/ordinate values. This allows the content of the tooltip to be customised (fix #757).

Screenshot from 2021-07-22 13-45-36

Screenshot from 2021-07-22 13-46-09

### DIFF
--- a/src/h5web/utils.ts
+++ b/src/h5web/utils.ts
@@ -7,13 +7,16 @@ import type {
 import type { ImageAttribute } from './vis-packs/core/models';
 import type { NxAttribute } from './vis-packs/nexus/models';
 
-export const formatValue = format('.3~e');
-export const formatPreciseValue = format('.5~e');
+export const formatTick = format('.5~g');
+export const formatBound = format('.3~e');
+export const formatBoundInput = format('.5~e');
+export const formatTooltipVal = format('.5~g');
+export const formatTooltipErr = format('.3~g');
 export const formatMatrixValue = format('.3e');
 export const formatMatrixComplex = createComplexFormatter('.2e', true);
 export const formatScalarComplex = createComplexFormatter('.12~g');
 
-export function createComplexFormatter(specifier: string, full = false) {
+function createComplexFormatter(specifier: string, full = false) {
   const formatVal = format(specifier);
 
   return (value: H5WebComplex) => {
--- a/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
+++ b/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
@@ -1,4 +1,4 @@
-import type { ReactNode } from 'react';
+import type { ReactElement, ReactNode } from 'react';
 import type { NdArray } from 'ndarray';
 import styles from './HeatmapVis.module.css';
 import ColorBar from './ColorBar';
@@ -7,12 +7,13 @@ import PanZoomMesh from '../shared/PanZoomMesh';
 import VisCanvas from '../shared/VisCanvas';
 import { getDims } from './utils';
 import { Domain, ScaleType, AxisParams } from '../models';
-import type { ColorMap, Layout } from './models';
+import type { ColorMap, Layout, TooltipData } from './models';
 import { DEFAULT_DOMAIN } from '../utils';
 import { assertDefined } from '../../../guards';
-import { useAxisValues, useTooltipFormatters } from './hooks';
-import { useDomain } from '../hooks';
+import { useAxisValues } from './hooks';
+import { useDomain, useValueToIndexScale } from '../hooks';
 import HeatmapMesh from './HeatmapMesh';
+import { formatTooltipVal } from '../../../utils';
 
 interface Props {
   dataArray: NdArray<number[]>;
@@ -28,6 +29,7 @@ interface Props {
   alphaArray?: NdArray<number[]>;
   alphaDomain?: Domain;
   flipYAxis?: boolean;
+  renderTooltip?: (data: TooltipData) => ReactElement;
   children?: ReactNode;
 }
 
@@ -46,6 +48,7 @@ function HeatmapVis(props: Props) {
     alphaArray,
     alphaDomain,
     flipYAxis,
+    renderTooltip,
     children,
   } = props;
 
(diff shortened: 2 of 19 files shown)


src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -95,7 +93,29 @@ function HeatmapVis(props: Props) {
...
+              <>
+                {`${abscissaLabel || 'x'}=${formatTooltipVal(abscissa)}, `}
+                {`${ordinateLabel || 'y'}=${formatTooltipVal(ordinate)}`}
+                <div className={styles.tooltipValue}>
+                  {formatTooltipVal(dataArray.get(yi, xi))}
+                </div>
+              </>
+            );
+          }}
+        />
```
axelboc, line 118: `TooltipMesh` now also accepts a `renderTooltip` function, but with only `x` and `y` as arguments. It is up to the visualization component to deal with these coordinates and render the content of the tooltip.

This is the simplest solution that we discussed. I really wasn't a fan of passing the `abscissas` and `ordinates` arrays (plus the `switchAtMidPoints` boolean) to `TooltipMesh` so it could compute the correct `abscissa` and `ordinate` values.

One reason is that it would have forced potential users of `TooltipMesh` to always generate `abscissas` and `ordinates` arrays, even when they'd be happy just doing `Math.round()` on the tooltip's raw coordinates. It would also have complexified the code and types of `TooltipMesh`, as we would have had to check whether an `ordinates` array was passed before creating the corresponding `ordinateToIndex` scale, and we would have had to enforce that `renderTooltip` would only receive `abscissa` and not `ordinate` when the `ordinates` array was not passed.

With this implementation, `TooltipMesh` remains completely uncoupled with the custom axis values feature.

Sure, it's a fair amount of code to add to `HeatmapVis` and `LineVis`... but I actually like the explicitness of it. I could have turned `useTooltipFormatters` into `useTooltipRenderer`, or moved the whole `<TooltipMesh ... />` block into a `HeatmapTooltipMesh` component, but it didn't feel necessary.

Let me know what you think.

src/h5web/vis-packs/core/line/LineVis.tsx
```diff
@@ -109,7 +106,33 @@ function LineVis(props: Props) {
...
+            const abscissa = abscissas[xi];
+
+            if (renderTooltip) {
+              return renderTooltip({ abscissa, xi, x });
+            }
+
+            const value = dataArray.get(xi);
+            if (value === undefined) {
+              return undefined;
+            }
```
axelboc, line 122: I believe this piece of code (previously in `useTooltipFormatters`) has been obsolete for a while.

When the user hovers over the extended parts of the live visualization, the `scaleThreshold` (in `getValueToIndexScale`) returns the nearest abscissa index (either 0 or `values.length - 1`). There's no way to disable this automatic "clamping" (since it's not a continuous scale), so we can never get an out-of-range index.

Is that right?

axelboc: If the code is indeed obsolete, then would we still need the ability to "hide" the popup by returning `undefined` in `renderTooltip`?

loichuder: > There's no way to disable this automatic "clamping" (since it's not a continuous scale), so we can never get an out-of-range index.
> 
> Is that right?

Good catch, I think you are right. You can remove this piece code for now and we can think about another way to handle this in another PR (it is not the worst thing to display the first/last value when hovering the extended parts).

Anticipating a bit, I think the best behaviour would be to still display the current abscissa value but do not display the value. We could test that `x` is included in `abscissas` for this ?

src/h5web/utils.ts
```diff
@@ -7,8 +7,11 @@ import type {
 import type { ImageAttribute } from './vis-packs/core/models';
 import type { NxAttribute } from './vis-packs/nexus/models';
 
-export const formatValue = format('.3~e');
-export const formatPreciseValue = format('.5~e');
+export const formatTick = format('.5~g');
+export const formatBound = format('.3~e');
+export const formatBoundInput = format('.5~e');
+export const formatTooltipVal = format('.5~g');
+export const formatTooltipErr = format('.3~g');
```
loichuder, line 14: Might be worth having a `formatters.ts` file

loichuder (approved):
Another well-stuck blow :clap:

axelboc:
> Anticipating a bit, I think the best behaviour would be to still display the current abscissa value but do not display the value. We could test that x is included in abscissas for this ?

Yes, it's an easy check. Good idea, it felt wrong hiding the tooltip completely 💯

src/h5web/utils.ts
axelboc: Hmm maybe. Not sure it's worth the effort though 😄

src/h5web/vis-packs/core/line/LineVis.tsx
axelboc: > Anticipating a bit, I think the best behaviour would be to still display the current abscissa value but do not display the value. We could test that x is included in abscissas for this ?

Yes, it's an easy check. Good idea, it felt wrong hiding the tooltip completely 💯

axelboc: Actually, the `switchAtMidPoints` throws a spanner in the works. If we want to show the first value in the tooltip over the same amount of space as the other values, then we need to know where the mid-point is to the left/bottom of the first abscissa/ordinate value, which isn't straightforward, obviously.

axelboc: I think the behaviour is great as it is. Let's not bother changing it 😄 I'll remove that dead piece of code, though.

loichuder: Then, we could disable the clamping of the scale. But agreed: not necessary to change for now.

PR #764, opened 2021-07-22
Upgrade deps and ESLint config

axelboc:


axelboc:
Updating Cypress led to a tiny, invisible change on one of the snapshots (on a button shadow).

loichuder (approved):

PR #765, opened 2021-07-22
Improve error message when entity is not found with H5Grove

axelboc:
See https://github.com/silx-kit/h5web/issues/618#issuecomment-875431360 and https://github.com/silx-kit/h5grove/issues/24

This can be merged before or after the latest version of H5Grove is deployed, it doesn't matter.

### DIFF
--- a/src/h5web/providers/utils.ts
+++ b/src/h5web/providers/utils.ts
@@ -1,3 +1,4 @@
+import axios from 'axios';
 import { hasArrayShape } from '../guards';
 import { Dataset, DType, DTypeClass, Endianness } from './models';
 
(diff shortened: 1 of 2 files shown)


src/h5web/providers/utils.ts
```diff
@@ -91,3 +92,19 @@ export function flattenValue(
   const dims = slicedDims || entity.shape;
   return (value as unknown[]).flat(dims.length - 1);
 }
+
+export function handleAxiosError<T>(
```
axelboc, line 96: Similar to `handleError` in `h5web/utils` but handles axios error responses with specific status codes instead of JS errors with specific messages.

loichuder:
I will deploy a new version while we deal with the other PR so you can test it out

loichuder (approved):

Issue #758, opened 2021-07-19
[DomainSlider] Add an optional histogram


loichuder:
We could have a histogram inside the domain slider by giving an optional `histogram` prop to the `DomainSlider`.

This histogram would:
- be located in the `DomainToolip`
- have a range determined by the slider extension
- have indicators giving the data min/max and current min/max
- **not** handle any interaction as the sliders/input fields already achieve this

loichuder:
> have indicators giving the data min/max and current min/max

I wonder if this is necessary given that the histogram bounds give the current min/max and that the bounds of the bars given the data min/max.

axelboc:
> I wonder if this is necessary given that the histogram bounds give the current min/max and that the bounds of the bars given the data min/max.

I think our initial reasoning was that the histogram would be static and that we would show the vis domain on the histogram with vertical guides, but your dynamic implementation in #771 looks great. Also you're right, showing the data min/max never made much sense, unless there's a tiny bar that you can barely see, but then you can still change the scale.

loichuder:
With the last PR merged, this is done !

PR #770, opened 2021-07-30
Add optional histogram to DomainSlider in the tooltip

loichuder:
image

First step of #758.

The histogram is really basic for now: no interplay with the current scaleType nor the current slider domain. It simply displays what is given in the `histogram` prop of the `DomainSlider`

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider/DomainTooltip.module.css
+++ b/src/h5web/toolbar/controls/DomainSlider/DomainTooltip.module.css
@@ -12,13 +12,15 @@
 .tooltipInner {
   composes: popupInner from '../../Toolbar.module.css';
   padding: 1rem 0.375rem 1rem 0.75rem;
+  display: flex;
+  align-items: center;
 }
 
-.tooltipInner > p {
+.tooltipControls > p {
   margin-bottom: 0.75rem;
 }
 
-.tooltipInner > p:last-child {
+.tooltipControls > p:last-child {
   margin-bottom: 0;
 }
 
--- a/src/h5web/toolbar/controls/DomainSlider/DomainTooltip.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider/DomainTooltip.tsx
@@ -1,9 +1,14 @@
 import { forwardRef, useImperativeHandle, useRef } from 'react';
 import type { Domain } from '../../../../packages/lib';
 import { formatBound } from '../../../utils';
-import { DomainError, DomainErrors } from '../../../vis-packs/core/models';
+import {
+  DomainError,
+  DomainErrors,
+  HistogramData,
+} from '../../../vis-packs/core/models';
 import ToggleBtn from '../ToggleBtn';
 import BoundEditor, { BoundEditorHandle } from './BoundEditor';
+import Histogram from './Histogram';
 import styles from './DomainTooltip.module.css';
 import ErrorMessage from './ErrorMessage';
 
@@ -24,6 +29,7 @@ interface Props {
   onChangeMin: (val: number) => void;
   onChangeMax: (val: number) => void;
   onSwap: () => void;
+  histogram?: HistogramData;
 }
 
 interface Handle {
@@ -31,7 +37,7 @@ interface Handle {
 }
 
 const DomainTooltip = forwardRef<Handle, Props>((props, ref) => {
-  const { id, open, sliderDomain, dataDomain, errors } = props;
+  const { id, open, sliderDomain, dataDomain, errors, histogram } = props;
   const { isAutoMin, isAutoMax, isEditingMin, isEditingMax } = props;
   const {
     onAutoMinToggle,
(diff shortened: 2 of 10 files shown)


src/h5web/toolbar/controls/DomainSlider/DomainTooltip.tsx
```diff
@@ -47,6 +54,8 @@ const DomainTooltip = forwardRef<Handle, Props>((props, ref) => {
   const minEditorRef = useRef<BoundEditorHandle>(null);
   const maxEditorRef = useRef<BoundEditorHandle>(null);
 
+  const [controlsRef, controlsSize] = useMeasure<HTMLDivElement>();
```
loichuder, line 57: I measure the size of the `div` containg the controls to give the same size to the histogram. That gives a visually pleasing ratio.

src/stories/Customization.stories.mdx
```diff
@@ -164,3 +164,5 @@ as you see fit. For instance, if you'd like to change the color of the curve of
 | `--h5w-domainSlider-boundInput-focus--shadowColor`   | `royalblue`                | Box shadow color of bound inputs on focus      |
 | `--h5w-domainSlider-boundInput-editing--bgColor`     | `white`                    | Background color of bound inputs when editing  |
 | `--h5w-domainSlider-boundInput-editing--borderColor` | `royalblue`                | Border color of bound inputs when editing      |
+| `--h5w-domainSlider-histogram-bar--color`            | `royalblue`                | Fill color of histogram bars                   |
+| `--h5w-domainSlider-histogram-bgColor`               | `white`                    | Background color of histogram                  |
```
loichuder, line 168: Added two customization CSS variables

src/stories/DomainSlider.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+  },
+  argTypes: {
+    scaleType: {
+      control: {
+        type: 'inline-radio',
+        options: [ScaleType.Linear, ScaleType.Log, ScaleType.SymLog],
+      },
+    },
+  },
+} as Meta;
```
loichuder, line 53: Even if I think that `DomainSlider` does not need a story of its own, I found way it easier to test the histogram with it

src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
```diff
@@ -1,6 +1,7 @@
 .root {
   position: relative;
   display: flex;
+  justify-content: center;
```
loichuder, line 4: This is to keep the track aligned with the tooltip even if the parent div takes more width than the track

src/h5web/toolbar/controls/DomainSlider/DomainTooltip.tsx
```diff
@@ -65,66 +74,74 @@ const DomainTooltip = forwardRef<Handle, Props>((props, ref) => {
...
-            showSwapBtn={!isAutoMin && !isAutoMax}
-            onSwap={onSwap}
+        {histogram && (
+          <DataHistogram
+            bins={histogram.bins}
+            data={histogram.data}
+            size={controlsSize}
           />
         )}
+        <div ref={controlsRef} className={styles.tooltipControls}>
```
loichuder, line 84: Crappy diff here: I put all the controls into this `div` to be able to position the histogram on the left of the controls

src/h5web/toolbar/controls/DomainSlider/DomainTooltip.module.css
```diff
@@ -12,13 +12,15 @@
...
+  align-items: center;
 }
 
-.tooltipInner > p {
+.tooltipControls > p {
   margin-bottom: 0.75rem;
 }
 
-.tooltipInner > p:last-child {
+.tooltipControls > p:last-child {
```
loichuder, line 23: Adapting the CSS as I introduced a new `div` enclosing the controls (see below)

src/h5web/toolbar/controls/DomainSlider/DataHistogram.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <div className={styles.container}>
+      <svg width={width} height={height} className={styles.svg}>
+        {data.map((d, i) => (
+          <rect
+            className={styles.bar}
+            key={i} // eslint-disable-line react/no-array-index-key
+            x={xScale(bins[i])}
+            y={height - yScale(d)}
+            width={xScale(bins[i + 1]) - xScale(bins[i])}
```
axelboc, line 37: There seems to be a very slight gap between the bars. We may have to add `0.5` here or something to fix it 🤷

src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
axelboc: The root `div` of the domain slider is not supposed to grow to fill the available space. Is this for the storybook? Maybe the issue comes from the decorator with `display: grid`. Do you need the decorator or can you remove it?

axelboc (approved):
Awesome! Just some minor things.

src/h5web/toolbar/controls/DomainSlider/DomainTooltip.module.css
```diff
@@ -12,13 +12,15 @@
 .tooltipInner {
   composes: popupInner from '../../Toolbar.module.css';
   padding: 1rem 0.375rem 1rem 0.75rem;
+  display: flex;
+  align-items: center;
```
axelboc, line 16: One more thing: I'm confused about the HTML/CSS layout. You measure `tooltipControls` and give the histogram's SVG the same size. Why do you need `align-items: center` here and why do you need an additional flex `container` inside `DataHistogram` with a `2rem` padding? I feel like this could be simplified.

src/h5web/toolbar/controls/DomainSlider/DataHistogram.tsx
```diff
@@ -0,0 +1,49 @@
+import { AxisBottom, AxisLeft } from '@visx/axis';
+import { scaleLinear } from '@visx/scale';
+import { useDomain } from '../../../vis-packs/core/hooks';
+import type { Size } from '../../../vis-packs/core/models';
+import styles from './DataHistogram.module.css';
+
+interface Props {
+  bins: number[];
+  data: number[];
+  size: Size | undefined;
```
axelboc, line 10: You could pass `width` and `height` directly? You'd then have `type Props = HistogramData & Size`.

src/h5web/toolbar/controls/DomainSlider/DomainTooltip.module.css
loichuder: The `2rem` padding was to make space for the axes that overflow from the `svg`. But I can indeed remove the container and put a margin of `2rem` to the `svg`.

Well spotted :owl: !

src/h5web/toolbar/controls/DomainSlider/DomainSlider.module.css
loichuder: In fact, I don't use the decorator. I am not sure why the root `div` takes the whole space in the Storybook...

loichuder: I removed this and used a new decorator `Center` to center the slider in the storybook.

Tell me what you think

axelboc:
The decorator looks great 👍 

I had a play with the histogram and noticed a couple of things:

- The tick labels are wrapped in `svg` elements. You need to use the custom `Tick` component that we use in other axes.

    image

- I'm not a huge fan of the 2rem margin all around. On the left, sure, but it feels a bit too much everywhere else.

    image

    I think your initial decision to use a container with some padding was fine in the end. I hadn't realised that the padding was to give space to the axes and that it ended up increasing the height of the tooltip, which is why you had to center `div.tooltipControls` with `align-items: center`. Padding makes more sense than margin in this case - i.e. to create space in order to display something in that space.

    That being said, I think the way you size the histogram is overly complex. I think it would be much simpler to just hard-code the size of the container `div` in CSS, give it some padding, let the `svg` element fit its container with `width="100%" height="100%"`, and measure the `svg` element in React for the scales. You could make the dimensions of the container `div` customisable with two CSS variables: either width/height (with default values in `rem`), or just the `width` and an aspect ratio).

PR #771, opened 2021-08-03
Make histogram extension match the slider extension

loichuder:
Second step of #758 

Also accounts for `scaleType`

### Slider domain bigger than data domain
image

### Slider domain smaller than data domain
image

### Symlog scale
image

Ticks are not great in the last case. <s>I will try to reuse our `Axis` components in a future PR.</s> Done in a522f25f9086ee6a3690ddb06dda4c47cf634005.

image

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider/DataHistogram.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider/DataHistogram.tsx
@@ -1,43 +1,54 @@
 import { AxisBottom, AxisLeft } from '@visx/axis';
 import { scaleLinear } from '@visx/scale';
 import { useDomain } from '../../../vis-packs/core/hooks';
-import type { Size } from '../../../vis-packs/core/models';
+import type { ScaleType, Size, Domain } from '../../../vis-packs/core/models';
+import { H5WEB_SCALES } from '../../../vis-packs/core/scales';
 import styles from './DataHistogram.module.css';
 
 interface Props {
   bins: number[];
   data: number[];
   size: Size | undefined;
+  scaleType: ScaleType;
+  sliderDomain: Domain;
 }
 
 function DataHistogram(props: Props) {
-  const { data, bins, size = { width: 0, height: 0 } } = props;
+  const {
+    data,
+    bins,
+    scaleType,
+    sliderDomain,
+    size = { width: 0, height: 0 },
+  } = props;
   const { height, width } = size;
 
-  const domain = useDomain(data);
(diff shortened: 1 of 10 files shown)


src/h5web/toolbar/controls/DomainSlider/DataHistogram.tsx
```diff
@@ -1,43 +1,54 @@
...
+            <rect
+              className={styles.bar}
+              key={i} // eslint-disable-line react/no-array-index-key
+              x={xScale(bins[i])}
+              y={height - yScale(d)}
+              width={xScale(bins[i + 1]) - xScale(bins[i])}
+              height={yScale(d)}
+            />
+          ))}
+        </svg>
```
loichuder, line 51: Enclosing the bars in a new `svg` that has `overflow: hidden`. Otherwise, bars overflow from the parent svg that needs to have to `overflow: visible` for the axes.

src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -136,6 +136,7 @@ function DomainSlider(props: Props) {
         onChangeMin={(val) => onCustomDomainChange([val, customDomain[1]])}
         onChangeMax={(val) => onCustomDomainChange([customDomain[0], val])}
         onSwap={() => onCustomDomainChange([customDomain[1], customDomain[0]])}
+        scaleType={scaleType}
```
loichuder, line 139: `scaleType` needs to be forwarded from `DomainSlider` to `DataHistogram` through `DomainTooltip`

src/h5web/toolbar/controls/DomainSlider/DataHistogram.tsx
axelboc: Can you use a group (`g`) instead of a nested `svg`?

axelboc (approved):

src/h5web/toolbar/controls/DomainSlider/DataHistogram.tsx
loichuder: I don't think you can put CSS properties to a group ? Or that there is a SVG equivalent to `overflow: hidden` ?

axelboc: Boom! `<g clip-path="inset(0 0 0 0) view-box">` 😤

loichuder:
Closing in favour of a more "silx-like" implementation:
image

PR #772, opened 2021-08-03
Fix alignement of ticks of the color bar

loichuder:
The use `border` makes the gradient bar slightly bigger than the axis. As a result, on a horizontal colormap, the first tick is slightly too much on the right and the last one slightly too much on the left.

#### Horizontal colormap with border
image

Changing to `outline` fixes the problem

#### Horizontal colormap with outline
image

#### Vertical colormap
| border      | outline |
| ----------- | ----------- |
| image       | image       |


loichuder:
Actually, this seems to depend on the zoom level. Curious to see how it will fare on your screen @axelboc

src/h5web/vis-packs/core/heatmap/ColorBar.module.css
```diff
@@ -25,6 +27,7 @@
   grid-area: gradient;
   display: flex;
   border: 1px solid currentColor;
+  z-index: 1; /* above axis ticks */
```
axelboc, line 30: Making sure the border of the color bar always covers the ticks.

src/h5web/vis-packs/core/heatmap/ColorBar.module.css
```diff
@@ -38,6 +41,14 @@
   overflow: visible;
 }
 
+.colorBar:not([data-horizontal]) > .axis {
+  border-top: 1px solid transparent; /* fix alignment of ticks */
+}
+
+.colorBar[data-horizontal] > .axis {
+  border-left: 1px solid transparent; /* fix alignment of ticks */
+}
```
axelboc, line 50: The outline fix was smart, but it made the color bar visually bigger than its container, which isn't ideal. What do you think of this? The problem is that the first tick is vertically (or horizontally) positioned in the middle of its theoretical position. Sometimes, because of rounding, it ends up being rendered one pixel too high. By adding a 1px transparent border, I move every tick down by 1px, which gives good-enough alignment in all cases.

image

image

image

loichuder: Clever :100:

loichuder: Could we use `padding` or `margin` instead of `border` ? It would make more sense, no ?

axelboc: Hmm, I thought visx would ignore the padding when positioning the ticks, and that the margin would push the last tick outside of the colorbar, but I was wrong: both seem work fine. I went with the padding.

axelboc (approved):
Kinda reviewing myself here, but oh well... 😄

PR #773, opened 2021-08-03
Add color bar below the histogram

loichuder:
Last step of #758

image

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
@@ -4,16 +4,17 @@ import { scaleLinear } from '@visx/scale';
 import { useCombinedDomain, useDomain } from '../../../vis-packs/core/hooks';
 import type {
   Domain,
-  HistogramData,
+  HistogramParams,
   ScaleType,
 } from '../../../vis-packs/core/models';
 import { H5WEB_SCALES } from '../../../vis-packs/core/scales';
 import Tick from '../../../vis-packs/core/shared/Tick';
 import { adaptedNumTicks, DEFAULT_DOMAIN } from '../../../vis-packs/core/utils';
-import styles from './Histogram.module.css';
+import HistogramColorBar from './HistogramColorBar';
 import HistogramIndicators from './HistogramIndicators';
+import styles from './Histogram.module.css';
 
-interface Props extends HistogramData {
+interface Props extends HistogramParams {
   scaleType: ScaleType;
   dataDomain: Domain;
   sliderDomain: Domain;
@@ -21,9 +22,12 @@ interface Props extends HistogramData {
 
 function Histogram(props: Props) {
   const { values, bins, scaleType, sliderDomain, dataDomain } = props;
+  const { colorMap, invertColorMap } = props;
 
   const binDomain = useCombinedDomain([sliderDomain, dataDomain]);
   const barDomain = useDomain(values);
+  const yMax = barDomain ? barDomain[1] : DEFAULT_DOMAIN[1];
+  const yMin = colorMap ? -yMax / 10 : 0;
 
   const [size, ref] = useMeasure<HTMLDivElement>();
 
@@ -38,9 +42,11 @@ function Histogram(props: Props) {
     range: [0, width],
   });
   const yScale = scaleLinear({
-    domain: barDomain ? [0, barDomain[1]] : DEFAULT_DOMAIN,
+    domain: [yMin, yMax],
     range: [0, height],
   });
+
+  const indicatorPositions = sliderDomain.map(xScale) as Domain;
 
   return (
     <div ref={ref} className={styles.container}>
@@ -52,13 +58,20 @@ function Histogram(props: Props) {
             x={xScale(bins[i])}
             y={height - yScale(d)}
             width={xScale(bins[i + 1]) - xScale(bins[i]) + 0.5} // +0.5 removes the small gap between bars
-            height={yScale(d)}
+            height={yScale(d) - yScale(0)}
           />
         ))}
-        <HistogramIndicators
-          positions={sliderDomain.map(xScale) as Domain}
-          height={height}
-        />
+        {colorMap && (
+          <HistogramColorBar
+            x={indicatorPositions[0]}
+            y={height - yScale(0)}
+            width={indicatorPositions[1] - indicatorPositions[0]}
+            height={yScale(0) - yScale(yMin)}
+            colorMap={colorMap}
+            invertColorMap={invertColorMap}
+          />
+        )}
+        <HistogramIndicators positions={indicatorPositions} height={height} />
 
         <AxisBottom
           top={height}
--- a/src/h5web/vis-packs/core/models.ts
+++ b/src/h5web/vis-packs/core/models.ts
@@ -7,6 +7,7 @@ import type {
   DType,
   Primitive,
 } from '../../providers/models';
+import type { ColorMap } from './heatmap/models';
 
 export enum ScaleType {
   Linear = 'linear',
@@ -85,7 +86,9 @@ export type ValueFormatter<T extends DType> = (val: Primitive<T>) => string;
 
 export type ImageAttribute = 'CLASS' | 'IMAGE_SUBCLASS';
 
-export interface HistogramData {
+export interface HistogramParams {
   values: number[];
   bins: number[];
+  colorMap?: ColorMap;
+  invertColorMap?: boolean;
 }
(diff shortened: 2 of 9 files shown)


src/h5web/toolbar/controls/DomainSlider/DataHistogram.tsx
```diff
@@ -8,11 +9,13 @@ import { adaptedNumTicks } from '../../../vis-packs/core/utils';
 import styles from './DataHistogram.module.css';
 
 interface Props {
-  bins: number[];
   data: number[];
-  size: Size | undefined;
+  bins: number[];
   scaleType: ScaleType;
   sliderDomain: Domain;
+  colorMap?: ColorMap;
```
loichuder, line 16: `colorMap` is kept optional as the domain slider could be used to other purposes that setting the domain of a color bar.

src/h5web/toolbar/controls/DomainSlider/DataHistogram.tsx
```diff
@@ -52,22 +57,33 @@ function DataHistogram(props: Props) {
...
+        {!colorMap && (
+          <AxisBottom
+            top={height}
+            scale={xScale}
+            numTicks={adaptedNumTicks(width)}
+            tickClassName={styles.tick}
+            tickComponent={Tick}
+            tickFormat={xScale.tickFormat(adaptedNumTicks(width), '.3~g')}
+          />
+        )}
```
loichuder, line 76: If `colorMap` is defined, the Axis is part of `ColorBar`. Else, we display a `AxisBottom`.

src/h5web/vis-packs/core/heatmap/ColorBar.module.css
```diff
@@ -1,6 +1,5 @@
 .colorBar {
   display: grid;
-  padding: 1rem;
```
loichuder, line 3: I removed the padding as it was better in the `DomainSlider` to have `ColorBar` directly against the bars.

As a consequence, I added wrappers with padding in the other components using `ColorBar`

src/h5web/vis-packs/core/models.ts
```diff
@@ -85,7 +86,9 @@ export type ValueFormatter<T extends DType> = (val: Primitive<T>) => string;
 
 export type ImageAttribute = 'CLASS' | 'IMAGE_SUBCLASS';
 
-export interface HistogramData {
+export interface HistogramParameters {
   data: number[];
   bins: number[];
+  colorMap?: ColorMap;
+  invertColorMap?: boolean;
```
loichuder, line 93: The `colorMap` props are enclosed in the `HistogramParameters` as they are only used if there is an histogram and it was easier this way to make the prop go through the `DomainSlider` and `DomainTooltip`

src/stories/ColorBar.stories.tsx
```diff
@@ -70,6 +80,7 @@ InvertColorMap.args = {
   invertColorMap: true,
 };
 
+export { Template as ColorMapTemplate };
```
loichuder, line 83: I now reuse the whole `Template` instead of only `TemplateProps`

axelboc (changes requested):
I have to admit that I'm not convinced about reusing the `ColorBar` component in the histogram... We already have a bottom axis, so we don't need the one from `ColorBar` (which requires its own `useMeasure`...) Also, I don't think `ColorBar` fits within the design of the histogram: it's too thick and I feel like the borders are too much.

All we want is a gradient, that's it. In other words, this part of the `ColorBar`, which you can simplify even more since we want it to always be horizontal anyway:

```tsx
        <div
          className={styles.gradient}
          data-keep-colors
          style={{
            backgroundImage: getLinearGradient(
              getInterpolator(colorMap, invertColorMap),
              horizontal ? 'right' : 'top',
              domain[0] === domain[1]
            ),
          }}
        />
```

src/h5web/toolbar/controls/DomainSlider/ColorBarGradient.tsx
```diff
@@ -0,0 +1,38 @@
+import { range } from 'lodash';
+import type { SVGProps } from 'react';
+import type { ColorMap } from '../../../../packages/lib';
+import { getInterpolator } from '../../../vis-packs/core/heatmap/utils';
+
+const GRADIENT_PRECISION = 1 / 20;
+const GRADIENT_RANGE = range(0, 1 + GRADIENT_PRECISION, GRADIENT_PRECISION);
```
loichuder, line 7: I could not find an easy way to reuse the `getColorStops` and `getLinearGradient` methods as the `svg` gradient is quite different from the CSS implementation.

src/h5web/toolbar/controls/DomainSlider/ColorBarGradient.tsx
```diff
@@ -0,0 +1,38 @@
...
+              <stop
+                key={v}
+                offset={`${v * 100}%`}
+                stopColor={interpolator(v)}
+              />
+            );
+          })}
+        </linearGradient>
+      </defs>
+      <rect {...rectProps} fill="url(#ColorMapGradient)" />
```
loichuder, line 33: I could restrict `rectProps` to `x,y,height,width` if you think that's better

src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
```diff
@@ -4,26 +4,30 @@ import { scaleLinear } from '@visx/scale';
...
 }
 
 function Histogram(props: Props) {
   const { values, bins, scaleType, sliderDomain, dataDomain } = props;
+  const { colorMap, invertColorMap } = props;
 
   const binDomain = useCombinedDomain([sliderDomain, dataDomain]);
   const barDomain = useDomain(values);
+  const yMax = barDomain ? barDomain[1] : DEFAULT_DOMAIN[1];
+  const yMin = colorMap ? -yMax / 10 : 0;
```
loichuder, line 30: If there is a colorMap, I extend the limits `yAxis` to make space for it.

src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
```diff
@@ -52,13 +58,20 @@ function Histogram(props: Props) {
             x={xScale(bins[i])}
             y={height - yScale(d)}
             width={xScale(bins[i + 1]) - xScale(bins[i]) + 0.5} // +0.5 removes the small gap between bars
-            height={yScale(d)}
+            height={yScale(d) - yScale(0)}
```
loichuder, line 61: This was not needed before as `yScale(0)` was equal to `0` but this is no longer true if there is a colorbar.

src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
```diff
@@ -52,13 +58,20 @@ function Histogram(props: Props) {
...
-        <HistogramIndicators
-          positions={sliderDomain.map(xScale) as Domain}
-          height={height}
-        />
+        {colorMap && (
+          <ColorBarGradient
+            x={indicatorPositions[0]}
+            y={height - yScale(0)}
+            width={indicatorPositions[1] - indicatorPositions[0]}
+            height={yScale(0) - yScale(yMin)}
```
loichuder, line 69: Placing the colorbar between the indicators in X and between `0` and `yMin` in Y

src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
```diff
@@ -52,13 +58,20 @@ function Histogram(props: Props) {
...
-            height={yScale(d)}
+            height={yScale(d) - yScale(0)}
           />
         ))}
-        <HistogramIndicators
-          positions={sliderDomain.map(xScale) as Domain}
-          height={height}
-        />
+        {colorMap && (
+          <ColorBarGradient
```
axelboc, line 65: I would maybe call this component `HistogramColorBar` instead.

src/stories/ColorBar.stories.tsx
axelboc: Nice, good call 👍

src/h5web/toolbar/controls/DomainSlider/ColorBarGradient.tsx
axelboc: Yeah it will be stricter, and with `Required<Pick<SVGProps<SVGRectElement>>, 'x' | ...>>` so they are required too. Unless those four props are already required in `SVGProps<SVGRectElement>`? If so, then don't bother.

axelboc (approved):

PR #774, opened 2021-08-05
Encode query parameters in `H5GroveProvider`

loichuder:
To fix https://github.com/silx-kit/jupyterlab-h5web/issues/45

#### Previously
The raw URL `/meta/?file=stack+.h5&path=/` was unchanged. 

The `file` arg gets interpreted as  `stack .h5` by `h5grove`.

#### Now
The URL `/meta/?file=stack+.h5&path=/` gets encoded as `/meta/?file=stack%2B.h5&path=%2F`. 

The `file` arg gets interpreted as  `stack+.h5` by `h5grove`.

### DIFF
--- a/src/h5web/providers/utils.ts
+++ b/src/h5web/providers/utils.ts
@@ -108,3 +108,7 @@ export async function handleAxiosError<T>(
     throw error;
   }
 }
+
+export function encodeQueryParams(params: Record<string, string>) {
+  return new URLSearchParams(params).toString();
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,7 @@
   "requires": true,
   "packages": {
     "": {
-      "version": "0.0.24",
+      "version": "0.0.25",
       "license": "MIT",
       "dependencies": {
         "@react-hookz/web": "^5.0.0",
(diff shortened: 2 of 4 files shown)


src/h5web/providers/utils.ts
```diff
@@ -108,3 +108,7 @@ export async function handleAxiosError<T>(
     throw error;
   }
 }
+
+export function encodeQueryParams(params: Record<string, string>) {
+  return new URLSearchParams(params).toString();
```
loichuder, line 113: I used `URLSearchParams` as it was easier for what I wanted to do. There was other ways of doing this and here why I discarded them:
- [qs.stringify](https://github.com/ljharb/qs#stringifying): adds a new dependency
- `encodeURIComponent`: require looping by hand (see https://stackoverflow.com/questions/59889140/different-output-from-encodeuricomponent-vs-urlsearchparams for an example implementation)
- pass query params with the `params` field of the `AxiosConfig`: tempting but require additional changes to `cancellableFetchValue`

axelboc: Great call 💯

axelboc (approved):

Issue #641, opened 2021-04-29
Test visualisations for datasets with NaN


loichuder:
Working on #639 made me wonder if h5web is really robust for datasets containing NaN. 

I think we should test the fetching and visualizations for:
- a scalar dataset containing NaN
- an array dataset containing NaN and regular values
- an array dataset containing only NaNs

loichuder:
This poses an issue at fetching: `NaN` and `Infinity`are not supported by the JSON format. There are several ways of deal with such values.

The default behaviour or the `json` Python serializer is to replace `Infinity` and `NaN` by their JS equivalent: https://docs.python.org/3/library/json.html#json.dump. Unfortunately, this produces invalid JSON as in HSDS: https://github.com/HDFGroup/hsds/issues/87.

A common workaround is to serialize `Infinity` and `NaN` as `null`. This is the approach currently in use in `jupyterlab_hdf` and `orjson`. While the JSON is valid, we cannot distinguish between `NaN` and `Infinity` in h5web. This also means that strictly speaking, the typing of our `ndarray<number>` is not valid as we could get `ndarray<number | null>` instead. I started some work on https://github.com/silx-kit/h5web/pull/642 but we wish to instead support `NaN` and `Infinity` explicitly and be able to distinguish the two.

The only approach I see to enable this is to serialize `NaN` and `Infinity` as strings as suggested in https://github.com/jupyterlab/jupyterlab-hdf5/issues/22 (proposed implementation in https://github.com/jupyterlab/jupyterlab-hdf5/pull/97). Note the use of `orjson` (that looks very promising for performance reasons) excludes this approach (https://github.com/jupyterlab/jupyterlab-hdf5/pull/98).

loichuder:
> A common workaround is to serialize Infinity and NaN as null. This is the approach currently in use in jupyterlab_hdf and orjson.

Update: this behaviour was implemented in HSDS when using `ignore_nan`: https://github.com/HDFGroup/hsds/issues/87

loichuder:
Here is the current status for our three providers:

### HSDS
#### When fetching
- As we don't use the [`ignore_nan` parameter](https://github.com/HDFGroup/hsds/issues/87), we receive invalid JSON containing `NaN` and `(-)Infinity`
[...]

PR #775, opened 2021-08-06
Do not support NaN in domain computation and shaders

loichuder:
Fixes HSDS issues reported in https://github.com/silx-kit/h5web/issues/641#issuecomment-894235661:
> :x: LineVis show a blank canvas even for slices containing only valid values
> :x: Computation of the domain in the HeatmapVis yields NaN or Infinity values

Also pixels with `NaN` are not considered as unsupported in `HeatmapVis`:
image

### DIFF
--- a/src/h5web/vis-packs/core/utils.ts
+++ b/src/h5web/vis-packs/core/utils.ts
@@ -94,20 +94,23 @@ export function getBounds(
   const values = toArray(valuesArray);
   const errors = errorArray && toArray(errorArray);
 
-  if (values.length === 0) {
-    return undefined;
-  }
-
-  return values.reduce<Bounds>(
+  const bounds = values.reduce<Bounds>(
     (acc, val, i) => {
+      // Ignore NaN and Infinity from the bounds computation
+      if (!Number.isFinite(val)) {
+        return acc;
+      }
       const newBounds = getNewBounds(acc, val);
       const err = errors && errors[i];
       return err
         ? getNewBounds(getNewBounds(newBounds, val - err), val + err)
         : newBounds;
     },
-    { min: values[0], max: values[0], positiveMin: Infinity }
+    { min: Infinity, max: -Infinity, positiveMin: Infinity }
   );
+
+  // Return undefined if min is Infinity (values is empty or contains only NaN/Infinity)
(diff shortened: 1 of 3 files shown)


axelboc (approved):
👍

src/h5web/vis-packs/core/utils.ts
```diff
@@ -349,7 +356,7 @@ export function getValidDomainForScale(
     return [0, max];
   }
 
-  return [min, max];
+  return Number.isFinite(min) ? [min, max] : undefined;
```
axelboc, line 359: Could you combine the two finite checks in a single `if` block above?

axelboc, line 359: The way these `Number.isFinite` checks are used is confusing, because it looks like you're checking if `min` is equal to `-Infinity` or if `max` is equal `+Infinity`, whereas it's actually the opposite. Moreover, `min` and `max` are either both equal to `+/-Infinity` or neither of them is, so the check for `Number.isFinite(min)` is superfluous.

In a way it would be clearer to stick with the idea that `getBounds` should return `undefined` when there are no supported values in the array. You can still remove the `values.length === 0` check, but maybe instead of directly returning the result of the `reduce`, you could instead add a ternary and return `undefined` if the `min` is `Infinity`:

```tsx
const bounds = values.reduce(...);
return Number.isFinite(bounds.min) ? bounds : undefined; // return `undefined if `values` has no finite values
```

axelboc (changes requested):
On second thoughts...

src/h5web/vis-packs/core/utils.ts
```diff
@@ -100,13 +100,16 @@ export function getBounds(
 
   return values.reduce<Bounds>(
     (acc, val, i) => {
+      if (!Number.isFinite(val)) {
+        return acc;
```
axelboc, line 104: Also, maybe just add a comment here to say that this ignores `Infinity` and `NaN` values from the domain computation.

src/h5web/vis-packs/core/utils.ts
loichuder: Good call :nerd_face:

axelboc (approved):

PR #778, opened 2021-08-10
Histogram extension match the slider extension if bigger than data domain

loichuder:
Peek 2021-08-10 15-55
Also tuned `Axis` aspect as in https://github.com/silx-kit/h5web/commit/a522f25f9086ee6a3690ddb06dda4c47cf634005

Leaving indicators for the next PR


axelboc (approved):
Looks great 💯

PR #779, opened 2021-08-11
Added indicators for slider domain in histogram

loichuder:
Peek 2021-08-10 17-45

Not a huge fan of the `Min`/`Max` legend in silx so I tried to add another type of visual indicator on the limit lines. Tell me if think that is fine or if that looks too much like a golf course :golf:

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider/Histogram.module.css
+++ b/src/h5web/toolbar/controls/DomainSlider/Histogram.module.css
@@ -16,3 +16,9 @@
 .bar {
   fill: var(--h5w-domainSlider-histogram--color, royalblue);
 }
+
+.indicators {
+  color: var(--h5w-domainSlider-histogram-indicator--color, midnightblue);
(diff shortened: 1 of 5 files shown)


src/h5web/toolbar/controls/DomainSlider/Histogram.module.css
```diff
@@ -16,3 +16,8 @@
 .bar {
   fill: var(--h5w-domainSlider-histogram--color, royalblue);
 }
+
+.indicators {
+  stroke: var(--h5w-domainSlider-histogram-indicator--color, midnightblue);
+  fill: var(--h5w-domainSlider-histogram-indicator--color, midnightblue);
```
axelboc, line 22: You can set the custom property on `color` and then use `currentColor` for both `fill` and `stroke`

axelboc (approved):
Hahaha it's fun, I love it

PR #780, opened 2021-08-11
Fix vis sizing feedback loop in some contexts

axelboc:
This should fix the issue with long names on Jupyter Lab. The problem was actually caused by a feedback loop with the sizing of the visualizations. Taking the visualizations out of the flow with `position: absolute` seems to solve the issue.

### DIFF
--- a/src/h5web/vis-packs/core/matrix/MatrixVis.module.css
+++ b/src/h5web/vis-packs/core/matrix/MatrixVis.module.css
@@ -1,10 +1,10 @@
 .wrapper {
+  position: relative;
   flex: 1 1 0%;
-  min-width: 0;
-  min-height: 0;
 }
 
 .grid {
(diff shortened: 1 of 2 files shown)


src/h5web/vis-packs/core/matrix/MatrixVis.module.css
```diff
@@ -1,10 +1,10 @@
 .wrapper {
+  position: relative;
   flex: 1 1 0%;
-  min-width: 0;
-  min-height: 0;
 }
 
 .grid {
+  position: absolute !important; /* prevent feedback loop with `useMeasure` on wrapper */
```
axelboc, line 7: `react-window` adds `position: relative`, so we need to override it.

axelboc (commented):
This is a technique we should keep in mind for when we measure the container in order to size the child.

PR #781, opened 2021-08-12
Encode query params with axios

loichuder:
Common encoding issues (such as not including `undefined` fields) were not solved by #774. Using axios avoid reimplementing fixes for such issues.

Also, the encoding was extended to all providers and not solely `h5grove`.

### DIFF
--- a/src/h5web/providers/api.ts
+++ b/src/h5web/providers/api.ts
@@ -4,10 +4,10 @@ import axios, {
   AxiosResponse,
   CancelTokenSource,
 } from 'axios';
-import { Entity, ProviderError, ValueRequestParams } from './models';
+import { Entity, ProviderError, ValuesStoreParams } from './models';
 
 interface ValueRequest {
-  params: ValueRequestParams;
+  storeParams: ValuesStoreParams;
   cancelSource: CancelTokenSource;
 }
 
@@ -37,15 +37,19 @@ export abstract class ProviderApi {
 
   protected async cancellableFetchValue<T>(
     endpoint: string,
-    params: ValueRequestParams
+    storeParams: ValuesStoreParams,
+    queryParams?: Record<string, string | undefined>
   ): Promise<AxiosResponse<T>> {
     const cancelSource = axios.CancelToken.source();
-    const request = { params, cancelSource };
+    const request = { storeParams, cancelSource };
     this.valueRequests.add(request);
 
     try {
       const { token: cancelToken } = cancelSource;
-      return await this.client.get<T>(endpoint, { cancelToken });
+      return await this.client.get<T>(endpoint, {
+        cancelToken,
+        params: queryParams || storeParams,
+      });
     } finally {
       // Remove cancellation source when request fulfills
       this.valueRequests.delete(request);
@@ -53,5 +57,5 @@ export abstract class ProviderApi {
   }
 
   public abstract getEntity(path: string): Promise<Entity>;
-  public abstract getValue(params: ValueRequestParams): Promise<unknown>;
+  public abstract getValue(params: ValuesStoreParams): Promise<unknown>;
 }
--- a/src/h5web/providers/h5grove/h5grove-api.ts
+++ b/src/h5web/providers/h5grove/h5grove-api.ts
@@ -1,5 +1,5 @@
 import {
-  ValueRequestParams,
+  ValuesStoreParams,
   Dataset,
   Entity,
   EntityKind,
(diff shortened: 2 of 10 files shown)


src/h5web/providers/api.ts
```diff
@@ -4,10 +4,14 @@ import axios, {
...
-import { Entity, ProviderError, ValueRequestParams } from './models';
+import {
+  Entity,
+  ProviderError,
+  ValueRequestParams as ValueStoreParams,
+} from './models';
 
 interface ValueRequest {
-  params: ValueRequestParams;
+  params: ValueStoreParams;
```
loichuder, line 14: Renamed this type to make the difference with the request query parameters

src/h5web/providers/h5grove/h5grove-api.ts
```diff
@@ -18,20 +18,15 @@ import type {
...
+import { convertDtype, flattenValue, handleAxiosError } from '../utils';
 import { buildEntityPath } from '../../utils';
 
 export class H5GroveApi extends ProviderApi {
   protected attrValuesCache = new Map<string, H5GroveAttrValuesResponse>();
 
   /* API compatible with h5grove@0.0.4 */
   public constructor(url: string, filepath: string) {
-    super(filepath, { baseURL: url });
+    super(filepath, { baseURL: url, params: { file: filepath } });
```
loichuder, line 29: Slight improvment: the same way that `domain` is set in the HsdsApi constructor, `file` can be set in the constuctor of `H5GroveApi` to avoid setting it for each fetch.

src/h5web/providers/api.ts
axelboc: Maybe just add an `s` to match the store's exact name: `ValuesStoreParams`

src/h5web/providers/h5grove/h5grove-api.ts
axelboc: nice !!

src/h5web/providers/api.ts
```diff
@@ -37,21 +41,25 @@ export abstract class ProviderApi {
 
   protected async cancellableFetchValue<T>(
     endpoint: string,
-    params: ValueRequestParams
+    storeParams: ValueStoreParams,
+    queryParams: Record<string, string | undefined>
   ): Promise<AxiosResponse<T>> {
     const cancelSource = axios.CancelToken.source();
-    const request = { params, cancelSource };
+    const request = { params: storeParams, cancelSource };
```
axelboc, line 44: might as well rename `params` to `storeParams` in the `ValueRequest` interface

src/h5web/providers/jupyter/jupyter-api.ts
```diff
@@ -63,20 +64,22 @@ export class JupyterStableApi extends ProviderApi {
...
+    params: ValuesStoreParams
   ): Promise<JupyterDataResponse> {
-    const { path, selection = '' } = params;
+    const { path, selection } = params;
     const { data } = await this.cancellableFetchValue<JupyterDataResponse>(
-      `/data/${this.filepath}?uri=${path}${selection && `&ixstr=${selection}`}`,
-      params
+      `/data/${this.filepath}`,
+      params,
+      { uri: path, selection }
```
axelboc, line 82: `ixstr: selection` 😉

src/h5web/providers/hsds/hsds-api.ts
```diff
@@ -173,12 +173,13 @@ export class HsdsApi extends ProviderApi {
...
+    params: ValuesStoreParams
   ): Promise<unknown> {
-    const { selection = '' } = params;
+    const { selection } = params;
     const { data } = await this.cancellableFetchValue<HsdsValueResponse>(
-      `/datasets/${entityId}/value${selection && `?select=[${selection}]`}`,
-      params
+      `/datasets/${entityId}/value`,
+      params,
+      { selection }
```
axelboc, line 182:
```
{ select: selection && `[${selection}]` }
```

src/h5web/providers/h5grove/h5grove-api.ts
```diff
@@ -74,19 +67,21 @@ export class H5GroveApi extends ProviderApi {
...
   private async fetchData(
-    params: ValueRequestParams
+    params: ValuesStoreParams
   ): Promise<H5GroveDataResponse> {
     const { data } = await this.cancellableFetchValue<H5GroveDataResponse>(
-      `/data/?${encodeQueryParams({ file: this.filepath, ...params })}`,
-      params
+      `/data/`,
+      params,
+      { ...params }
```
axelboc, line 84: You could let `queryParams` fall back to `storeParams` in `cancellableFetchValue` : `queryParams = storeParams`. This way you don't have to pass them twice here.

axelboc (changes requested):
Great stuff, it's a lot cleaner! Just found a couple of refactoring issues.

axelboc (approved):

PR #784, opened 2021-08-16
Make responsive toolbar more robust and flexible

axelboc:
- `Toolbar` now accepts fragments as children. I use https://github.com/grrowl/react-keyed-flatten-children to flatten the children into an array instead of `React.Children.toArray()`, as the latter doesn't traverse fragments.
- I also fix an issue with the toolbar remaining empty sometimes with the Heatmap vis. This was related to the fact that I was storing the measured widths of the children against their indices, which can change between renders. I now store the widths against the children's keys. I was able to confirm that `react-keyed-flatten-children` guarantees stable string keys for every children regardless of the JSX logic (i.e. even when a child is rendered conditionally, like the `DomainSlider`).
- While I was working on the toolbar, I noticed a few styling issues, notably:
  - a `z-index` issue with the selector menus when two of them were stacked in the toolbar's overflow menu;
  - a missing margin between the icons and the labels in the scale selector;
  - an alignment issue with the labels of the scale selector options;
  - two alignment issues between the selector menus and their buttons;
  - a text wrapping issue with the `Scale root` option in the scale selector.

Here is the scale selector before and after:

| Before | After |
| --- | --- |
| image | image |


src/h5web/toolbar/ComplexToolbar.tsx
```diff
@@ -37,16 +37,17 @@ function ComplexToolbar() {
...
+            dataDomain={dataDomain}
+            customDomain={customDomain}
+            scaleType={scaleType}
+            onCustomDomainChange={setCustomDomain}
+          />
+          <Separator />
+        </>
       )}
 
-      {dataDomain && <Separator />}
```
axelboc, line 48: We can now use fragments! 🎉

src/h5web/toolbar/Toolbar.tsx
```diff
@@ -1,32 +1,34 @@
...
 }
 
 function Toolbar(props: Props) {
   const { children } = props;
 
-  // Convert `children` to array, as it cannot be used directly (and remove `undefined` children)
-  const allChildren = Children.toArray(children) as ReactElement[];
+  /* Convert `children` to flat array by traversing nested arrays and fragments.
+   * (Note that `flattenChildren` guarantees stable string keys regardless of JSX logic.) */
+  const allChildren = flattenChildren(children).filter(isReactElement);
```
axelboc, line 19: The types are simpler:

- `children` is now just `ReactNode` (i.e. `ReactChild | ReactFragment | ReactPortal | boolean | null | undefined`), which brings flexibility to users of the `Toolbar` component;
- `flattenChildren` returns `ReactChild[]` (i.e. `(ReactElement | string | number)[]` -- note that `ReactFragment` is gone), which I then filter to ignore strings and numbers.

So we end with what we really want, `ReactElement[]`, in a much less convoluted way than before.

src/h5web/toolbar/Toolbar.tsx
```diff
@@ -1,32 +1,34 @@
...
+  const childrenWidths = useMap<string, number>();
 
   // Group visible and hidden children based on their accumulated width
   const [visibleChildren, hiddenChildren, allMeasured] = allChildren.reduce<
     [ReactElement[], ReactElement[], boolean, number]
   >(
-    ([accVisible, accHidden, accAllMeasured, accWidth], child, index) => {
-      const width = childrenWidths[index] ?? 0;
+    ([accVisible, accHidden, accAllMeasured, accWidth], child) => {
+      const width = childrenWidths.get(child.key as string) ?? 0;
```
axelboc, line 31: As mentioned, I've confirmed that all children elements have stable, string keys thanks to `flattenChildren` (though I think this was also the case with `React.Children.toArray`...), so I can store the widths against those keys. Note that `as string` is required here because `ReactElement.key` has type `string | number | null`...

src/h5web/toolbar/controls/ColorMapSelector/ColorMapOption.tsx
```diff
@@ -0,0 +1,28 @@
...
+  getLinearGradient,
+  INTERPOLATORS,
+} from '../../../../packages/lib';
+import styles from './ColorMapSelector.module.css';
+
+interface Props {
+  option: ColorMap;
+}
+
+function ColorMapOption(props: Props) {
```
axelboc, line 12: Moved into its own file for consistency with `ScaleOption`.

src/h5web/toolbar/controls/ColorMapSelector/ColorMapSelector.module.css
```diff
@@ -2,6 +2,12 @@
   display: flex;
 }
 
+.option {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+}
+
```
axelboc, line 10: This style was previously named `.colormap` and located in `Selector.module.css`. As a result it was being applied to the scale selector's options as well, which explains why the labels were spaced out from their icons.

src/h5web/toolbar/controls/ScaleSelector/ScaleOption.tsx
```diff
@@ -1,15 +1,16 @@
...
   const { option } = props;
   const { Icon, label } = H5WEB_SCALES[option];
+
   return (
-    <>
+    <div className={styles.option}>
       <Icon className={styles.icon} />
-      {label}
-    </>
+      <span>{label}</span>
```
axelboc, line 12: The `span` brings back the margin of the `icon` (because of `.icon:last-child` in `Toolbar.module.css`).

src/h5web/toolbar/controls/ScaleSelector/ScaleSelector.module.css
```diff
@@ -0,0 +1,8 @@
+.option {
+  display: flex;
+  align-items: center;
```
axelboc, line 3: No more `justify-content: space-between` here 😄

src/h5web/toolbar/controls/Selector/Selector.module.css
```diff
@@ -37,8 +36,10 @@
 
 .menu {
   position: absolute;
-  top: 100%;
-  left: 0;
+  top: calc(100% + 0.75rem);
+  left: 50%;
+  transform: translateX(-50%); /* center menu with button */
+  z-index: 1; /* above other selectors */
```
axelboc, line 42: This fixes the alignment and stacking of the selector menus.

src/h5web/toolbar/controls/Selector/Selector.module.css
```diff
@@ -68,21 +69,15 @@
...
-  composes: colorMap;
   flex: 1 1 0%;
 }
 
 .option {
-  composes: colorMap;
+  flex: 1 1 0%;
   padding: 0.5rem 0.75rem;
   cursor: pointer;
+  white-space: nowrap;
```
axelboc, line 80: This fixes the text wrapping of `Square root`.

src/h5web/guards.ts
```diff
@@ -25,6 +26,10 @@ export function isDefined<T>(val: T): val is T extends undefined ? never : T {
   return val !== undefined;
 }
 
+export function isReactElement(child: ReactChild): child is ReactElement {
```
loichuder, line 29: It could be clearer to use the real type rather than the alias (i.e. `ReactElement | string | number` instead of `ReactChild`) but that's all right like this

src/h5web/toolbar/Toolbar.tsx
loichuder: Would it be safer to `assertString` instead of `as string` ? Or do you trust `flattenChildren` enough ?

src/h5web/toolbar/controls/ColorMapSelector/ColorMapSelector.module.css
loichuder: :fishing_pole_and_fish:

loichuder (approved):
Great !

> I also fix an issue with the toolbar remaining empty sometimes with the Heatmap vis.

I also got this issue (usually on first render of the root NeXus vis) but forgot to report it. Nice to fix this as well :+1:

src/h5web/toolbar/Toolbar.tsx
axelboc: I trust it enough. Also, if it's anything else, then the `get` will return `undefined` anyway.

src/h5web/guards.ts
axelboc: yeah, I might do this in the last PR then, to avoid having to rebase 👍

PR #785, opened 2021-08-16
Fix controls jumping in/out of overflow menu and refactor controls measuring flow

axelboc:
The controls have a tendency to jump in and out of the overflow menu when switching back and forth between a shorter and longer option in a selector control - e.g. between "Log" and "Square root" in a scale selector. This shifts the position of the controls in the toolbar (including the one being interacted with), which can be annoying:

jump1

**This PR fixes this issue by storing only the largest known width of each control.** This is done by ignoring new measurements of a control if they are shorter than the measurement already present in the state.

This change has the side-effect of decreasing **by half** the number of renders of the `Toolbar` used in `HeatmapToolbar`. Indeed, until now, we weren't checking if a measured width had changed before updating the state, which was triggering a lot of unnecessary renders when the domain slider appeared for the first time.

The result is that overflowing controls jump into the overflow menu but don't come out of it again:

jump3

---

The above fix is actually the second commit of the PR. The first commit is a refactoring that consists in moving the logic of showing/hiding controls from `Toolbar` to `MeasuredControl` so that each control is shown/hidden independently from all the others while it is being measured for the first time.

Apart from hugely simplifying the logic in `Toolbar` by removing the need for an `allMeasured` boolean, it also means that we no longer re-hide all the controls when the domain slider appears for the first time (i.e. when the domain is computed, after all the other controls have already been measured). This should improve the perceived rendering time of the toolbar by a few milliseconds (which is a little bit noticeable when switching between the line and heatmap vis).

### DIFF
--- a/src/h5web/toolbar/MeasuredControl.tsx
+++ b/src/h5web/toolbar/MeasuredControl.tsx
@@ -4,23 +4,29 @@ import styles from './Toolbar.module.css';
 
 interface Props {
   children: ReactElement;
+  knownWidth: number | undefined;
   onMeasure: (width: number) => void;
 }
 
 function MeasuredControl(props: Props) {
-  const { children: child, onMeasure } = props;
+  const { children: child, knownWidth, onMeasure } = props;
 
   return (
     <Measure
       onResize={({ entry }: { entry?: DOMRect }) => {
-        if (entry) {
+        if (entry && entry.width !== knownWidth) {
           onMeasure(entry.width);
         }
       }}
     >
       {({ measureRef }) => (
-        <div ref={measureRef} className={styles.control}>
-          {child}
+        <div
+          className={styles.controlWrapper}
+          data-measured={knownWidth !== undefined} // hide control until measured for the first time
+        >
+          <div ref={measureRef} className={styles.control}>
+            {child}
+          </div>
         </div>
       )}
     </Measure>
--- a/src/h5web/toolbar/MeasuredControl.tsx
+++ b/src/h5web/toolbar/MeasuredControl.tsx
@@ -14,7 +14,7 @@ function MeasuredControl(props: Props) {
   return (
     <Measure
       onResize={({ entry }: { entry?: DOMRect }) => {
-        if (entry && entry.width !== knownWidth) {
+        if (entry && entry.width > (knownWidth || 0)) {
           onMeasure(entry.width);
         }
       }}
--- a/src/h5web/toolbar/Toolbar.tsx
+++ b/src/h5web/toolbar/Toolbar.tsx
@@ -23,53 +23,43 @@ function Toolbar(props: Props) {
 
   const childrenWidths = useMap<string, number>();
 
-  // Group visible and hidden children based on their accumulated width
-  const [visibleChildren, hiddenChildren, allMeasured] = allChildren.reduce<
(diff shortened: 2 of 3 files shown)


src/h5web/toolbar/Toolbar.tsx
```diff
@@ -24,48 +24,43 @@ function Toolbar(props: Props) {
...
-        isMeasured && !isOverflowing ? [...accVisible, child] : accVisible,
-        isMeasured && isOverflowing ? [...accHidden, child] : accHidden,
-        accAllMeasured && isMeasured,
+        !isOverflowing ? [...accVisible, child] : accVisible,
+        isOverflowing ? [...accHidden, child] : accHidden,
         accWidth + width,
       ];
     },
-    [[], [], true, 0]
+    [[], [], 0]
```
axelboc, line 40: Much simpler! 😄

src/h5web/toolbar/MeasuredControl.tsx
```diff
@@ -4,23 +4,29 @@ import styles from './Toolbar.module.css';
...
 
 function MeasuredControl(props: Props) {
-  const { children: child, onMeasure } = props;
+  const { children: child, knownWidth, onMeasure } = props;
 
   return (
     <Measure
       onResize={({ entry }: { entry?: DOMRect }) => {
-        if (entry) {
+        if (entry && entry.width > (knownWidth || 0)) {
```
axelboc, line 17: This is the line that fixes the jumping in/out issue. Much more elegant than if I had to do it in `Toolbar`.

src/h5web/toolbar/Toolbar.tsx
```diff
@@ -24,48 +24,41 @@ function Toolbar(props: Props) {
...
   );
 
   const isSeparatorLast =
     visibleChildren[visibleChildren.length - 1]?.type === Separator;
 
-  const allOrVisibleChildren = allMeasured
-    ? isSeparatorLast
-      ? visibleChildren.slice(0, -1)
-      : visibleChildren
-    : allChildren;
```
axelboc, line 52: We no longer have to switch between rendering all the children or only the visible ones. We now assume that all children have a size of `0` to start, which means that they all end up in `visibleChildren` and they all get rendered and wrapped in a `MeasuredControl`.

src/h5web/toolbar/Toolbar.tsx
```diff
@@ -24,48 +24,41 @@ function Toolbar(props: Props) {
   const childrenWidths = useMap<string, number>();
 
   // Group visible and hidden children based on their accumulated width
-  const [visibleChildren, hiddenChildren, allMeasured] = allChildren.reduce<
-    [ReactElement[], ReactElement[], boolean, number]
+  const [visibleChildren, hiddenChildren] = allChildren.reduce<
```
loichuder, line 27: Rereading this, I feel the naming `visible/hiddenChildren` can be misleading.

What about `displayed` instead of `visible` and `overflowing` instead of `hidden` ?

loichuder (approved):
:point_right: :brain:

src/h5web/toolbar/Toolbar.tsx
axelboc: I went with `inView` and `outOfView` to keep the names short and avoid excessive line wrapping.

PR #786, opened 2021-08-16
Prevent flickering of controls when switching scale type of Heatmap

axelboc:
Another case of the toolbar controls moving in/out of the overflow menu occurs when changing the scale type on the heatmap/complex vis:

jump2

When I switch the `scaleType`, the `dataDomain` in the config store gets reset to `undefined`, which leads to the domain slider being unmounted in the toolbar. The space left by the slider is given to the "Screenshot" button, which was previously hidden in the overflow menu. When the `dataDomain` is recomputed and the `useEffect` that updates it in the store is called, the domain slider is re-mounted and the "Screenshot" button goes back in the overflow menu.

I scratched my head a lot to try to fix this in the toolbar (either in `Toolbar` or in `HeatmapToolbar`/`ComplexToolbar`), but it went nowhere. The problem is not with the toolbar, it is with resetting the `dataDomain` in the store. We've never really liked this solution anyway (at least I haven't), so I tried to find an alternative.

Turns out I found one, and I think it's actually pretty clean. The great part is that it's almost entirely contained within the config store of the Heatmap! Here is how the toolbar behaves after the fix (and the one from #785):

jump4

---

Another minor change: I've reworded "Phase and amplitude" to "Phase & Amp." to keep the selector to a reasonable size. Here is what the drop-down looks like with this change and the one from #784:

| Before #784 | Before this PR | Now |
| --- | --- | --- |
| image | image | image |
| | image | image |

### DIFF
--- a/src/h5web/toolbar/ComplexToolbar.tsx
+++ b/src/h5web/toolbar/ComplexToolbar.tsx
@@ -24,6 +24,7 @@ function ComplexToolbar() {
     colorMap,
     setColorMap,
     scaleType,
+    staleDomainScaleType,
     setScaleType,
     layout,
     setLayout,
@@ -41,7 +42,7 @@ function ComplexToolbar() {
           <DomainSlider
             dataDomain={dataDomain}
             customDomain={customDomain}
-            scaleType={scaleType}
+            scaleType={staleDomainScaleType || scaleType}
             onCustomDomainChange={setCustomDomain}
           />
           <Separator />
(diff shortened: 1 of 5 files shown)


axelboc (commented):
To avoid having `staleScaleType || scaleType` in both `HeatmapToolbar` and `ComplexToolbar`, I could add a computed state called `safeScaleType`, or `domainSliderScaleType`. However, I feel like this could add confusion to a fairly straightforward solution.

Looking forward to knowing what you think 😉

loichuder (approved):
I would simply rename `staleScaleType` into `staleDomainScaleType` or something equivalent to make clear that this is for the `domainSlider`.

PR #787, opened 2021-08-18
Switch to yarn and upgrade deps

axelboc:
I ran into weird issues trying to update visx: NPM screwed up `package-lock.json` and uninstalled a bunch of deps like `webpack`... I don't get why. Anyway, so I got sick of NPM and switched to `yarn` as I had mentioned I was planning on doing anyway.

The upside is that we can now use the `resolutions` field in `package.json` to resolve incompatibles versions more explicitely. 🎉 

Along with `yarn`, I updated all the dependencies, including `visx`, which finally has `react@17` as peer dep.


src/h5web/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -63,7 +62,7 @@ function DomainSlider(props: Props) {
   const tooltipRef = useRef<DomainTooltipHandle>(null);
 
   useClickOutside(rootRef, cancelEditing);
-  useKey('Escape', () => {
+  useKeyboardEvent('Escape', () => {
```
axelboc, line 65: Oh yeah, one fewer `react-use` hook -- not many left, which is good.

src/h5web/vis-packs/core/complex/config.tsx
```diff
@@ -1,7 +1,7 @@
 import create from 'zustand';
 import { persist } from 'zustand/middleware';
 import type { ConfigProviderProps } from '../../models';
-import createContext from 'zustand/context';
+import createContext from 'zustand/context.js';
```
axelboc, line 4: In adding support for ES modules in Node, zustand broke our imports... Until CRA switches to `webpack@5`, we have to load the CommonJS module `context.js` explicitly (instead of the ES module `context.mjs`, which is the default when no extension is specified).

Note that this is not a problem with `zustand/middleware`, as this module doesn't `import` any other module.

loichuder: They rolled back some of the changes in 3.5.10: https://github.com/pmndrs/zustand/releases/tag/v3.5.10.

You might want to see if this is still needed if we update to 3.5.10 instead of 3.5.9

CONTRIBUTING.md
```diff
@@ -149,17 +148,17 @@ To build the packages, run the following commands:
...
 
 To test a package locally, build the packages then run the following commands:
 
 ```bash
 cd lib/dist
-npm link
+yarn link
 cd /your/test/app
-npm link @h5web/lib
+yarn link @h5web/lib
```
loichuder, line 161: We might want to check if this still works. I seem to remember that linking with yarn was a bit different.

loichuder (approved):
:yarn: :+1:

src/h5web/vis-packs/core/complex/config.tsx
axelboc: Ah yes, saw this. Thanks for the reminder.

CONTRIBUTING.md
axelboc: It works the same way as NPM and gives the same "react hook call" issues 😂

PR #790, opened 2021-08-30
SqrtScale: use positive min rather than 0 when domain crosses 0

### LINKED ISSUE #789: [SqrtScale] Domain can be `[0, max]` even when data doesn't contain `0`
On line 353 of `src\h5web\vis-packs\core\utils.ts`, in `getValidDomainForScale()`, we have this code:

```ts
  if (scaleType === ScaleType.Sqrt && min * max < 0) {
    return [0, max];
  }
```

I'm thinking that this can potentially return a min bound that doesn't actually match the min "safe" value of the data. If the data is `[-1, 1, 2]`, for instance, then the returned domain will be `[0, 2]` instead of `[1, 2]`.

The `Bounds` object should probably include an additional  property (`positiveOrZeroValue`, perhaps?)

loichuder:
Fix #789

### DIFF
--- a/src/h5web/vis-packs/core/complex/utils.ts
+++ b/src/h5web/vis-packs/core/complex/utils.ts
@@ -6,6 +6,7 @@ const INITIAL_BOUNDS: Bounds = {
   min: Infinity,
   max: -Infinity,
   positiveMin: Infinity,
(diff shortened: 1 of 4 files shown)


axelboc (approved):
Great naming 👍

PR #791, opened 2021-08-31
Consider domain of bins when computing domain of histogram x axis

loichuder:
#### Problem

The domain of the histogram x axis is given by the combination of the slider domain and of the data domain. 
if the domain of bins is not contained in the slider domain or the data domain, the histogram can then overflow (reminder: the parent `svg` has `overflow: visible` to show the axes) :

image

Note that this occurs if the bin domain is not included in the data domain. This should not occur as normally, when computing a histogram of the data, the bins are spread along the data values. But this is not enforced in h5web.

#### What I could have done

I could have added a group with `clip-path="view-box"` as mentioned in https://github.com/silx-kit/h5web/pull/771#discussion_r685932539 to clip the overflowing histogram:

image

But depending on the chosen data domain, the histogram can be really zoomed-in and is only discoverable by extending the slider domain. 

#### What this PR does instead

If we want the histogram to be relevant, I think it should to be **completely** visible at all times. I then combined the slider domain and the data domain with the bin domain to compute the x axis.

image

### DIFF
--- a/src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
+++ b/src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
@@ -24,7 +24,8 @@ function Histogram(props: Props) {
   const { values, bins, scaleType, sliderDomain, dataDomain } = props;
   const { colorMap, invertColorMap } = props;
 
-  const binDomain = useCombinedDomain([sliderDomain, dataDomain]);
+  const binDomain = useDomain(bins);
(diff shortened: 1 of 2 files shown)


axelboc (approved):
Great, and the extension behaviour of the abscissa domain still works perfectly 👏

loichuder:
> Great, and the extension behaviour of the abscissa domain still works perfectly clap

Yeah `useCombinedDomain` is really handy

PR #792, opened 2021-09-01
Add `iconOnly` prop to `ToggleGroup` buttons and export `Selector` in `lib`

loichuder:


src/h5web/toolbar/controls/ToggleGroup.tsx
```diff
@@ -53,7 +54,7 @@ function Btn(props: BtnProps) {
     >
       <span className={styles.btnLike}>
         {Icon && <Icon className={styles.icon} />}
-        <span className={styles.label}>{label}</span>
+        {!iconOnly && <span className={styles.label}>{label}</span>}
```
axelboc, line 57: You also need to add an `aria-label` to the button, so that it remains accessible (like in `Btn`)

axelboc (approved):

Issue #714, opened 2021-06-17
Switch to monorepo structure


axelboc:
We're starting to see the limitations of our current setup:

- `@h5web/lib` and `@h5web/app` each contain a single JS bundle with all the files and dependencies (except React). This means that apps that use these packages cannot optimize away unused code with tree-shaking.
- If those apps need some of the dependencies already present in the bundle (like Three, R3F, D3 or zustand), those dependencies will end up duplicated in the app's output bundle.
- Because all the dependencies are listed in a single `package.json` at the root of the project, there's no way to know which one is used by which package. 
- Because we don't list dependencies in `@h5web/lib`, they are not installed when the package is installed, which means that typings and source maps for those dependencies don't work. Listing the dependencies in `@h5web/lib` is tedious because of the problem above.
- webpack isn't the most appropriate tool for building libraries yet, as it doesn't support outputting ES modules and in a one-to-one manner. Unfortunately, we cannot replace it with Babel because it's difficult to tell which files are needed for the library and need to be transpiled, and which ones aren't. Using Rollup would be possible, but difficult because of the single `package.json`, as it needs to know which dependencies are needed by the lib.

For these reasons, I believe the time has come to switch to a proper monorepo structure that will make the project more scalable, optimised and easier to maintain.

A few solutions are available, like [Lerna](https://github.com/lerna/lerna), [Yarn Workspaces](https://classic.yarnpkg.com/en/docs/workspaces/), etc. but my eyes are currently on [Rush](https://rushjs.io/), which is developed by Microsoft and seems like a more robust, stable, flexible and all-in-one solution than other tools.

Here is what the structure could look like:

```
+ .vscode
+ .github
+ node_modules
+ packages // NPM packages, each with their own Storybook, potentially
  + app
  + lib
    - src
    - package.json
    - README.md
    - etc.
  + nexus
+ demo // CRA (note that Rush doesn't force every projects and packages to be under a single folder like Lerna (`packages`)
  + cypress
  - etc.
- .env
- .eslintrc.js
-  package.json // Rush
- tsconfig.json
- etc.
```

axelboc:
Remaining known issues/limitations/awkwardness:

- [ ] CSS modules ordering issues in output `styles.css` when composing from relative paths -- forces us to use global utilities
- [ ] CSS modules compilation issues with Vite on Windows (`root: '.'`)
[...]

PR #795, opened 2021-09-21
Switch to pnpm monorepo structure

axelboc:
See #714 

Known issues/unknowns/awkward things:

- [x] Externalize dependencies of `@h5web/lib` when building `@h5web/app`
- [x] Externalize dependencies of `@h5web/shared` when building `@h5web/app` and `@h5web/lib` (need to redeclare dependencies of `@h5web/shared` in both packages, since the former is not actually published to NPM)
- [x] "Non-existent export `TextProps` is imported from `../../node_modules/.pnpm/@visx+text@2.1.0_react@17.0.2/node_modules/@visx/text/esm/types.js`" when build `@h5web/app`
- [x] How to test a built package without having to remove `publishConfig` from its `package.json` and uncomment the `dist/styles.css` import in the demo?
- [x] Shared global styles (`@h5web/<app|lib>/src/styles.css` aren't currently built/published -- can we bundle them in `with dist/style.css`?
- [x] Can types of `@h5web/shared` be accessed from published NPM packages? Probably not, since `@h5web/shared` is not published to NPM and output declaration files import `@h5web/shared`... How to bundle them in? With project references maybe?
- [x] Jest: `testTimeout` ignored on Windows when running tests locally, and `jest.setTimeout()` ignored on Linux...


packages/lib/package.json
```diff
@@ -1,18 +1,92 @@
 {
   "name": "@h5web/lib",
   "version": "0.0.1",
-  "description": "Data visualization library",
+  "description": "HDF5 visualization toolkit",
```
loichuder, line 4: `lib` is not really bound to `HDF5` right ?

packages/lib/src/h5web/toolbar/controls/ToggleGroup.module.css
```diff
@@ -5,7 +5,7 @@
 
 .btn {
   composes: btn from '../Toolbar.module.css';
-  padding: 0;
+  padding: 0 !important; /* FIX style ordering issue with Vite */
```
loichuder, line 8: Could you put the link to the vite issue somewhere ? In the linked issue or in the comment ?

packages/lib/src/h5web/vis-packs/core/heatmap/HeatmapVis.tsx
```diff
@@ -107,8 +106,8 @@ function HeatmapVis(props: Props) {
 
             return (
               <>
-                {`${abscissaLabel || 'x'}=${formatTooltipVal(abscissa)}, `}
-                {`${ordinateLabel || 'y'}=${formatTooltipVal(ordinate)}`}
+                {`${abscissaLabel ?? 'x'}=${formatTooltipVal(abscissa)}, `}
+                {`${ordinateLabel ?? 'y'}=${formatTooltipVal(ordinate)}`}
```
loichuder, line 110: Nice catch :+1:

packages/lib/src/h5web/vis-packs/core/line/GlyphMaterial.tsx
```diff
@@ -10,7 +10,7 @@ interface Props {
 
 function GlyphMaterial(props: Props) {
   const { color, size, glyphURL } = props;
-  const sprite = useLoader(TextureLoader, glyphURL || GLYPH_URLS.Cross);
+  const sprite = useLoader(TextureLoader, glyphURL ?? GLYPH_URLS.Cross);
```
loichuder, line 13: Hm, I wonder if `||` isn't better in this case. 

I mean, setting `glyphURL` to `""` will make this crash with `??` while it will fallback to `Cross` with `||`, right ?

src/h5web/providers/utils.test.ts
```diff
@@ -1,77 +0,0 @@
...
-      class: DTypeClass.Float,
-      size: 64,
-      endianness: undefined,
-    });
-  });
-
-  it('should handle unknown type', () => {
-    expect(convertDtype('>notAType')).toEqual({ class: DTypeClass.Unknown });
-  });
-});
```
loichuder, line 77: Did you move these tests elsewhere or removed them ? I did not find them :sweat_smile: 

Same question for:
- src/h5web/metadata-viewer/utils.test.ts
- src/h5web/providers/hsds/utils.test.ts
- src/h5web/providers/mock/utils.test.ts

loichuder (approved):

apps/demo/src/DemoApp.tsx
```diff
@@ -1,28 +1,17 @@
...
 import JupyterApp from './JupyterApp';
 
-// Split mock data generation code out of main bundle
-const MockProvider = lazy(
-  () =>
-    import(
-      /* webpackChunkName: "mock" */ '../h5web/providers/mock/MockProvider'
-    )
-);
-
```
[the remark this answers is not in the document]
axelboc: It was giving me trouble, can't remember why. Anyways, it prevents loading 15kB on initial page lod, roughly, which isn't significant-enough to warrant the trouble.

apps/storybook/.storybook/main.js
```diff
@@ -0,0 +1,21 @@
...
+module.exports = {
+  stories: ['../src/**/*.stories.@(tsx|mdx)'],
+  addons: [
+    '@storybook/preset-create-react-app',
+    '@storybook/addon-links',
+    '@storybook/addon-essentials',
+  ],
+  webpackFinal: (config) => {
+    // Let webpack process raw TS files imported from local packages
+    config.module.rules[2].oneOf[2].include.push(
```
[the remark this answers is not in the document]
axelboc: Yeah, since it's not CRA, I can't use the utility from `react-app-rewire-alias`... I think it's better to have something very targetted and clear rather than convoluted and hard to debug. When we upgrade Storybook, if this breaks, at least we'll know it's just because the position of the rule changed, not because the code we use to find it uses the wrong regex or whatever...

apps/storybook/.storybook/preview.js
```diff
@@ -1,4 +1,6 @@
 import 'normalize.css';
+import '@h5web/lib/src/styles.css';
+// import '@h5web/lib/dist/style.css';
```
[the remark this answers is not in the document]
axelboc: I totally forgot about Storybook when I introduced this env var. Thanks, I'll take a look.

packages/app/src/h5web/metadata-viewer/MetadataViewer.tsx
```diff
@@ -1,8 +1,7 @@
-import { memo, useContext } from 'react';
-import { isAbsolutePath } from '../guards';
 import { ProviderContext } from '../providers/context';
 import { ProviderError } from '../providers/models';
-import { buildEntityPath, handleError } from '../utils';
+import { buildEntityPath, handleError, isAbsolutePath } from '@h5web/shared';
+import { memo, useContext } from 'react';
```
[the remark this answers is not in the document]
axelboc: Yeah, I'm actually thinking of re-enabling the ESLint rule to sort imports after merging, to see if it can save us from doing this manually.

packages/app/tsconfig.build.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": ["**/*.test.ts"],
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "dist-ts",
```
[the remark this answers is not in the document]
axelboc: Exactly :+1:

packages/lib/src/h5web/toolbar/controls/ToggleGroup.module.css
axelboc: There isn't one for the ordering specifically. The existing issue is for `composes` and relative paths like `../../my-shared-module.css`, which we don't hit because of the ordering issue and my decision to switch to global utilities. I'm planning on opening an issue for the ordering specifically asap.

packages/lib/src/h5web/vis-packs/core/line/GlyphMaterial.tsx
axelboc: Yeah but that's true with any string that is not a valid URL or doesn't correctly encode a resource. The goal here is not to prevent misuse, but to support not passing a glyph URL at all.

packages/app/src/h5web/metadata-viewer/MetadataViewer.tsx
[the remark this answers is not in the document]
loichuder: Ah yes, good point. Ignore my other similar comments then

packages/lib/src/h5web/vis-packs/core/line/GlyphMaterial.tsx
loichuder: I see your point, makes sense

src/h5web/providers/utils.test.ts
axelboc: Ooh well spotted! I will go through them again and compare the number of tests  and test suites reported by Jest.

axelboc: I found two more tests for `getAttributeValue` that I hadn't moved. Thanks :+1:

apps/storybook/.storybook/preview.js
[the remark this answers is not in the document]
axelboc: Done, same thing but with `STORYBOOK_DIST`, which I've mentioned in the `CONTRIBUTING` guide.

PR #802, opened 2021-10-04
Upgrade dependencies

axelboc:
Nothing interesting to report.


apps/storybook/.storybook/preview.js
```diff
@@ -1,11 +1,11 @@
 import 'normalize.css';
+import '../src/styles/index.css';
+import '@h5web/shared/src/utils.css';
```
axelboc: For some reason, the shared styles aren't imported in development (even though the lib imports them). It's probably a webpack config issue, but I couldn't fix it, so I'm importing it explicitly. In prod, this means the shared styles are duplicated, but it's not a big deal.

apps/storybook/.storybook/preview.js
```diff
@@ -1,11 +1,11 @@
 import 'normalize.css';
+import '../src/styles/index.css';
+import '@h5web/shared/src/utils.css'; // fix missing shared styles in development
```
axelboc, line 3: For some reason, the shared styles aren't imported in development (even though the lib imports them). It's probably a webpack config issue, but I couldn't fix it, so I'm importing it explicitly. In prod, this means the shared styles are duplicated, but it's not a big deal.

loichuder (approved):

PR #803, opened 2021-10-04
Include base and vars css in @h5web/app

loichuder:


axelboc (commented):
The branch doesn't look up to date with `main`.

axelboc (approved):

axelboc:
The Storybook also has the variables duplicated, but it doesn't use `@h5web/app`... We could either leave it as is, or install `@h5web/app` in the Storybook so we can import the styles, or ... put the styles in `@h5web/shared` and import them from both Storybook and `@h5web/app` (which might make sense for the variables actually), or ... rethink things a bit? I can add this to the monorepo issue, but I feel like maybe we should take a good look at the global styles to see which ones we really need and whether we can do without them. For instance, I don't think we use HTML elements like `h1`, `h2`, `dl`, etc.

PR #806, opened 2021-10-04
Make histogram domain safe with respect to data domain 

### LINKED ISSUE #793: [DomainSlider] Histogram domain is not safe with respect to the scale type
### Describe the bug

The implementation in https://github.com/silx-kit/h5web/pull/791 does not take the scale type into account. The bin domain can therefore be unsupported by the current scale type, leading to disappearance of the histogram x axis.

### To Reproduce

See this [URL on the Storybook[(https://h5web-docs.panosc.eu/?path=/story/toolbar-domainslider--histogram&args=scaleType:log;histogram.bins[0]:-2) where I have done the following steps:

1. Go to Toolbar/DomainSlider > Histogram
2. Change the scale type to `Log`
3. Change the first bin to `-2` to have a bin domain unsupported by log
4. The histogram x axis disappears and a bunch of errors are logged in the console.

### Expected behaviour

The histogram should restrict its domain to a safe domain as does the slider.

loichuder:
Fix #793 

The problem was in fact twofold (see below)

### DIFF
--- a/packages/lib/src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
+++ b/packages/lib/src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
@@ -3,6 +3,7 @@ import { useMeasure } from '@react-hookz/web';
 import { AxisBottom, AxisLeft } from '@visx/axis';
 import { scaleLinear } from '@visx/scale';
 
+import { useSafeDomain } from '../../../vis-packs/core/heatmap/hooks';
 import { useCombinedDomain, useDomain } from '../../../vis-packs/core/hooks';
 import type { HistogramParams } from '../../../vis-packs/core/models';
 import { H5WEB_SCALES } from '../../../vis-packs/core/scales';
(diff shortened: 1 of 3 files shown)


packages/lib/src/h5web/toolbar/controls/DomainSlider/Histogram.tsx
```diff
@@ -22,8 +23,10 @@ function Histogram(props: Props) {
   const { values, bins, scaleType, sliderDomain, dataDomain } = props;
   const { colorMap, invertColorMap } = props;
 
-  const binDomain = useDomain(bins);
-  const xDomain = useCombinedDomain([binDomain, sliderDomain, dataDomain]);
+  const binDomain = useDomain(bins, scaleType) || DEFAULT_DOMAIN;
```
loichuder, line 26: 1. The `binDomain` could be unsupported by the current `scaleType`. This is fixed by taking `scaleType` into account when computing `binDomain`

PR #831, opened 2021-10-28
Support typed arrays throughout, assert values, and fix values provided by H5Grove

axelboc:
See individual commits.


axelboc:
Superseded by #836

PR #833, opened 2021-10-28
Add NX support for entities without default but valid NX children

### LINKED ISSUE #821: [NeXus] Support NX entities without default
h5web only  finds plottable data by following default attributes. 

However, the [NeXuS procedure](https://manual.nexusformat.org/datarules.html#version-3) shows that `default` is not required to find plottable data:
>  If the root has an attribute default, then its value is the name of the NXentry group to be used. **Otherwise, pick any NXentry group. This is trivial if there is only one NXentry group.**

> Open the NXentry group selected above. If it has an attribute default, then its value is the name of the NXdata group to be used. **Otherwise, pick any NXdata group. This is trivial if there is only one NXdata group.**

See https://h5web.panosc.eu/?file=201805_WSe2_arpes.nxs for an example file with only one NXentry and one NXdata group and no default.

loichuder:
Fix #821 

Support must be added for the two following entity types:
- NXroots with no `default` but having at least one NXentry child
- NXentries with no `default` but having at least a NXdata child

This PR solves this in two times:
1. `isNxGroup` is changed so that the above entity types are recognized as NeXus in `VisPackChooser` to use the `NexusPack`.
2. `getDefaultEntity` is changed so that if there is no `default` attribute, it searches for a valid child entity.


packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -33,17 +34,38 @@ export function isNxDataGroup(group: Group): boolean {
...
+  // ... a NXentry group OR
+  if (getAttributeValue(entity, 'default') || isNxEntry(entity)) {
+    return true;
+  }
+
+  // ... a NXroot that has a NXentry child
+  return hasChildren(entity) &&
+    getAttributeValue(entity, 'NX_class') === 'NXroot'
+    ? entity.children.some(isNxEntry)
+    : false;
```
axelboc, line 68: If I understand `getDefaultEntityPath` correctly, it calls `isNxGroup` for every _child_ of an entity. So I don't think `isNxGroup` can ever be called with the root entity.

If you remove this, then you can shrink this function back down to a one-liner:

```ts
  return (
    isGroup(entity) &&
    (isNxDataGroup(entity) || !!getAttributeValue(entity, 'default')) || isNxEntry(entity)
  );
```

That being said, I also think `isNxEntry` is unnecessarily restrictive: we could just check for the presence of `NX_class` rather than check for `NXentry` specifically. This way we allow for `NXprocess` and the like to be selected:

```ts
  return (
    isGroup(entity) &&
    (isNxDataGroup(entity) || !!getAttributeValue(entity, 'default')) || !!getAttributeValue(entity, 'NX_class')
  );
```

packages/shared/src/mock/metadata.ts
```diff
@@ -209,5 +209,20 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
...
+          children: [
+            makeNxDataGroup('spectrum', {
+              signal: makeNxDataset('oneD', intType, [41]),
+              axes: { X: makeNxDataset('X', floatType, [41]) },
+              axesAttr: ['X'],
+            }),
+          ],
+        }),
+      ],
+    }),
```
axelboc, line 226: I don't think it makes much sense to have an `NXroot` group that is not actually at the root. Maybe just switch to `NXprocess`? Or even better, just remove the extra NX group to just test the `NXentry` without a default path -- I think that would still test the two behaviours we're interested in: 1) consider a group without a `default` attribute but with an `NX_class` attribute as an NX group; 2) look through the children to find the first NX group to use as default entity. The recursion part is not new, so there's no reason to test it here, I think.

axelboc (changes requested):

packages/app/src/vis-packs/nexus/utils.ts
loichuder: > If I understand `getDefaultEntityPath` correctly, it calls `isNxGroup` for every _child_ of an entity. 

Yes if needed.

> So I don't think `isNxGroup` can ever be called with the root entity.

In fact, it will be called for the root entity in the `VisPackChooser` to know if `NexusPack` should be used for the root entity.

So your first proposed implementation of `isNxGroup` would consider that the NexusPack should not be used for a `NXroot` without `default`.

This being said, your second implementation would work in this case and simplifies things :+1:

packages/shared/src/mock/metadata.ts
loichuder: > I don't think it makes much sense to have an `NXroot` group that is not actually at the root.

Yes, this would not appear in a regular Nexus file but it was a way to test the *NXroot without default* usecase.

I can remove it if you want :smile:

axelboc: With my second implementation, testing `NXroot` specifically is indeed unnecessary. Might as well keep the mock data compliant, I reckon.

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -41,8 +41,9 @@ function assertNxDataGroup(group: Group): void {
 
 export function isNxGroup(entity: Entity): boolean {
   return (
-    isGroup(entity) &&
-    (isNxDataGroup(entity) || !!getAttributeValue(entity, 'default'))
+    (isGroup(entity) &&
+      (isNxDataGroup(entity) || !!getAttributeValue(entity, 'default'))) ||
+    !!getAttributeValue(entity, 'NX_class')
```
axelboc, line 46: There's a parenthesis issue, I think.

axelboc (approved):

packages/app/src/vis-packs/nexus/utils.ts
loichuder: Thanks, I added a simple `hasAttribute` wrapper to lighten the syntax a bit

axelboc: Ah yeah, good call

PR #836, opened 2021-11-03
Fix values provided by H5Grove

### LINKED ISSUE #824: [types] Strengthen `Value<D>` type
#817 raised a number of issues related to the `Value<D>` type:

### Mismatch if D is a numeric array dataset
Its value can be `TypedArray` instead of `number[]` inferred by `Value<D>`. It stills works as the value is used later to construct a `ndarray` which is possible with `TypedArray` and `number[]`.

### Mismatch if D is a numeric scalar dataset
Its value can be a `TypedArray` of one element instead of `number`. It still works as `ScalarVis` will convert the `TypedArray` to a string, yielding to same result if it was a `number`.

### Uncaught mismatches
The previous type mismatches were not caught by our type assertions which is a problem. The cast from `unknown` to `Value<D>` takes place in `ValueFetcher` which is where we discussed adding an assertion.

axelboc:
- Return scalar values instead of typed arrays of length `1`.
- Restore flattening of non-numeric arrays.
- Return classic JS arrays instead of typed arrays until we decide whether/how to support typed arrays in the visualizations.
- Assert values returned by providers against datasets type and shape to avoid future regressions - fix #824


packages/app/src/vis-packs/nexus/containers/NxComplexImageContainer.tsx
```diff
@@ -34,13 +35,13 @@ function NxComplexImageContainer(props: VisContainerProps) {
       />
       <VisBoundary resetKey={dimMapping}>
         <NxValuesFetcher
-          nxData={nxData}
+          nxData={nxData as NxData<ComplexType>}
```
axelboc, line 38: Unrelated improvement. This type cast is a bit nicer than the one below. The problem is that `assertComplexType` above asserts the type of the dataset and not of the `nxData` object... I was about to add a guard called `assertNxDataComplexType` or something, but it felt too much, especially sicne we'd need another guard checking for the other type in `NxImageContainer`.

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -46,12 +52,16 @@ export class H5GroveApi extends ProviderApi {
...
+      return hasScalarShape(entity) ? array[0] : [...array];
     }
 
-    const buffer = await this.fetchBinaryData(params);
-    return new DTypedArray(buffer);
+    const value = await this.fetchData(params);
+    return hasArrayShape(entity)
+      ? (value as unknown[]).flat(entity.shape.length - 1)
+      : value;
   }
```
axelboc, line 65: Sorry for the diff, I've swapped the early return logic.

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -46,10 +47,13 @@ export function useDatasetValue(
   }
 
   // If `dimMapping` is not provided or has no slicing dimension, the entire dataset will be fetched
-  return valuesStore.get({
+  const value = valuesStore.get({
     path: dataset.path,
     selection: getSliceSelection(dimMapping),
   });
+
+  assertDatasetValue(value, dataset);
```
axelboc, line 55: The new assertion function is called here and below.

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -95,14 +106,12 @@ export const useCombinedDomain = createMemo(getCombinedDomain);
...
-  value: T,
+export function useMappedArray<T, U extends T[] | undefined>(
+  value: U,
   dims: number[],
   mapping: DimensionMapping,
   autoScale?: boolean
-): T extends (infer U)[]
-  ? [NdArray<U[]>, NdArray<U[]>]
-  : [undefined, undefined];
+): U extends T[] ? [NdArray<U>, NdArray<U>] : [undefined, undefined];
```
axelboc, line 114: Unrelated typing simplification. `infer` is not needed.

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -46,12 +52,16 @@ export class H5GroveApi extends ProviderApi {
...
+      const buffer = await this.fetchBinaryData(params);
+      const array = new DTypedArray(buffer);
+      return hasScalarShape(entity) ? array[0] : [...array];
     }
 
-    const buffer = await this.fetchBinaryData(params);
-    return new DTypedArray(buffer);
+    const value = await this.fetchData(params);
+    return hasArrayShape(entity)
+      ? (value as unknown[]).flat(entity.shape.length - 1)
```
loichuder, line 63: There is a complexity layer that you did not take into account there: the presence of `selection` in the `params` changes the shape of the array returned from the request.

This means the checks of shapes must be done on the resulting slice rather than on the full dataset.

loichuder (changes requested):
There is a complexity layer that is missed by your assertions: the presence of `selection` in the `params` changes the shape of the array returned from the request. It can even return a scalar value instead of an array.

This means the checks of shapes (such as in h5grove provider and in `assertDatasetValue`) must be done according to the asked selection rather than considering the full dataset metadata. Same thing for the flattening.

axelboc:
Right, I need to re-add the `flattenValue` utility that we've removed. That being said, I don't think it ever supported the case of the selection of a scalar. When we ignore this edge case, `assertDatasetValue` works fine, as it always receives an array.

Since this edge case isn't something we encounter in the viewer, I don't really see the point of supporting it. At least not in this PR.

packages/app/src/providers/utils.ts
```diff
@@ -81,6 +81,17 @@ export function convertDtype(dtype: string): DType {
   }
 }
 
+export function flattenValue(
+  value: unknown,
+  entity: Dataset<ArrayShape>,
+  selection?: string
+) {
+  assertArray(value);
```
axelboc, line 89: This will throw if we ask the API to select a scalar, so at least we'll know (in the unlikely event we write code that does this inadvertently).

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -41,17 +48,22 @@ export class H5GroveApi extends ProviderApi {
...
+      const array = new DTypedArray(buffer);
+      return hasScalarShape(entity) ? array[0] : [...array];
     }
 
-    const buffer = await this.fetchBinaryData(params);
-    return new DTypedArray(buffer);
+    const value = await this.fetchData(params);
+    return hasArrayShape(entity)
+      ? flattenValue(value, entity, selection)
+      : value;
```
loichuder, line 66: You could put the check inside `flattenValue`

packages/app/src/vis-packs/nexus/containers/NxComplexImageContainer.tsx
loichuder: >  I was about to add a guard called `assertNxDataComplexType` or something, but it felt too much, especially sicne we'd need another guard checking for the other type in `NxImageContainer`.

I don't see why it would feel too much. I think it would make sense to assert the type of `NxData` as everything would follow, no ? (i.e it would allow to infer the type of `signalDataset` which would in turn would allow to infer the type of `signal` )

packages/app/src/providers/utils.ts
loichuder: I understand that given our usecases, we will never get a scalar from an array dataset.

Still it bothers me to prevent it for the following reasons:

- Here and in `assertDatasetValue` usage locations, we have access to `selection`. So we could assert according to the `selection`. 
- The type inference in `useDatasetValue` could make use of `dimMapping` to have a `Value<D>` corresponding to the gotten slice.

Anyway, this will do for now.

loichuder (approved):

packages/app/src/providers/utils.ts
axelboc: Sure, it's all doable, but adding this level of type safety to `useDatasetValue` means that we then have to assert that the returned value is not scalar inside the relevant containers, which leads to exactly the same result as the assertion that the value returned by the provider is an array regardless of `selection`.

If anything, it would make more sense to prevent `useDatasetValue` from being called with a `dimMapping` that only contains numbers.

axelboc: Either way, we still have implicit coupling between `useDatasetValue` and the provider APIs' `getValue` methods... Perhaps a better solution would be to rethink the signature of `getValue` to make this coupling more explicit - e.g. by letting it accept a dataset rather than an entity and an optional dimension mapping array with at least one axis in it.

packages/app/src/providers/h5grove/h5grove-api.ts
axelboc: Hmm I think it's clearer if `flattenValue` accepts `Dataset<ArrayShape>` and returns `unknown[]` (which I'm going to make explicit), as it doesn't make sense to "flatten" a scalar.

PR #838, opened 2021-11-03
Check for `NXentry` or `NXroot` specifically when looking for NX groups

axelboc:
This ensures we don't show the `NX` badge next to non-plotable entities like `NXcollection` or `NXinstrument`. It also ensures that such NX entities are skipped when looking for a default `NXdata` group to plot.


packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -48,7 +48,7 @@ export function isNxGroup(entity: Entity): boolean {
     isGroup(entity) &&
     (isNxDataGroup(entity) ||
       hasAttribute(entity, 'default') ||
-      hasAttribute(entity, 'NX_class'))
+      getAttributeValue(entity, 'NX_class') === 'NXentry')
```
loichuder, line 51: This would remove the support for NXroot without `default` as pointed out in https://github.com/silx-kit/h5web/pull/833#discussion_r741092093

packages/shared/src/mock/metadata.ts
```diff
@@ -151,6 +151,17 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
...
+      defaultPath: undefined,
+      children: [
+        makeNxGroup('ignore_me', 'NXprocess'),
+        makeNxDataGroup('spectrum', {
+          signal: makeNxDataset('oneD', intType, [41]),
+          axes: { X: makeNxDataset('X', floatType, [41]) },
+          axesAttr: ['X'],
+        }),
+      ],
+    }),
```
axelboc, line 164: I've just moved this group up and renamed it to get this:

image

I've also added an `NXprocess` entity to check that it is correctly skipped when looking for the default plot.

packages/shared/src/mock/metadata.ts
```diff
@@ -151,6 +151,17 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
         }),
       ],
     }),
+    makeNxGroup('nexus_entry_no_default', 'NXentry', {
```
loichuder, line 154: :scream: 

Now witness all the tests relying on `^nexus_entry` break :smile:

loichuder (changes requested):

packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -89,6 +89,15 @@ test('visualize NXdata group with "rgb-image" interpretation', async () => {
...
 
+test('visualize NXentry group with implicit default child NXdata group', async () => {
+  await renderApp();
+  await selectExplorerNode('nexus_entry_no_default');
+
+  const tabs = await findVisSelectorTabs();
+  expect(tabs).toHaveLength(1);
+  expect(tabs[0]).toHaveTextContent(NexusVis.NxSpectrum);
+});
+
```
axelboc, line 100: New test here.

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -48,7 +48,8 @@ export function isNxGroup(entity: Entity): boolean {
     isGroup(entity) &&
     (isNxDataGroup(entity) ||
       hasAttribute(entity, 'default') ||
-      hasAttribute(entity, 'NX_class'))
+      getAttributeValue(entity, 'NX_class') === 'NXentry' ||
+      getAttributeValue(entity, 'NX_class') === 'NXroot')
```
axelboc, line 52: `['NXentry', 'NXroot'].includes()` doesn't work because the return type of `getAttributeValue` is `unknown`. I went with this instead of adding an assertion since we're most likely going to improve the type safety of `getAttributeValue` once we start fetching attribute values on demand.

loichuder: Yep, it is fine

loichuder (approved):

PR #840, opened 2021-11-03
Make copy path button more "catchable"

loichuder:
As I wanted to copy the path by clicking on the copy icon, I realized that the icon was quite shy:

Peek 2021-11-03 16-24

I changed the CSS a bit so that the icon is included in the crumb and is more easily "catchable"

_PS: Yes, I know I could click directly on the crumb but my brain forced me to click on the icon_

### DIFF
--- a/packages/app/src/breadcrumbs/BreadcrumbsBar.module.css
+++ b/packages/app/src/breadcrumbs/BreadcrumbsBar.module.css
@@ -15,7 +15,7 @@
   min-width: 0;
   margin-top: 0;
   margin-bottom: 0;
-  padding: 0 1rem;
+  padding: 0 0 0 1rem;
   font-size: inherit;
   font-weight: inherit;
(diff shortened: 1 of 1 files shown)


axelboc (changes requested):
Arf but this shifts the text on hover, and you get that weird flicker sometimes like on your GIF. You should be able to get around it with some padding: try moving the padding-right that is currently on the `h1` to the last crumb's button.

loichuder:
> Arf but this shifts the text on hover, and you get that weird flicker sometimes like on your GIF.

So **that** was the reason for the `position: absolute` ! I didn't remember :smile: ,thanks for pointing it out !

loichuder:
I found the origin of the "hover-gap" between the icon and the crumb: it is due to the shift of the icon `right: -0.125rem`.

I bridge the gap with some padding as you suggested and it seems to fix the issue

PR #842, opened 2021-11-05
Simplify `getValue` signature to prepare for stricter typing

axelboc:
`getValue`'s `params` parameter now contains a complete `dataset` object instead of just a dataset `path`. This will allow me later to statically type the return value of `getValue` to `Value<D>` instead of `unknown`.

### DIFF
--- a/packages/app/src/providers/hsds/hsds-api.ts
+++ b/packages/app/src/providers/hsds/hsds-api.ts
@@ -33,7 +33,7 @@ import type {
   HsdsId,
 } from './models';
 import {
-  assertNonNullHsdsDataset,
+  assertHsdsDataset,
   isHsdsGroup,
   convertHsdsShape,
   convertHsdsType,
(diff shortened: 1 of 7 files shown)


packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -108,17 +108,15 @@ export class HsdsApi extends ProviderApi {
   }
 
   public async getValue(params: ValuesStoreParams): Promise<unknown> {
-    const { path } = params;
+    const { dataset } = params;
+    assertHsdsDataset(dataset);
```
axelboc, line 112: We still have to make sure here that the dataset includes the HSDS-specific `id` property. It's for TS, but also to ensure that the `id` property of the dataset doesn't get lost from the time it is returned by `getEntity` to the time it arrives here.

axelboc: The alternative to this implicit coupling between `getEntity` and `getValue` would be to keep a private map between entity paths and IDs. I don't think it's worth the complexity.

packages/app/src/providers/mock/mock-api.ts
```diff
@@ -36,21 +36,17 @@ export class MockApi extends ProviderApi {
   }
 
   public async getValue(params: ValuesStoreParams): Promise<unknown> {
-    const { path, selection } = params;
+    const { dataset, selection } = params;
 
-    if (path.includes('error_value')) {
-      // Throw error when fetching value with specific path
+    if (dataset.name === 'error_value') {
```
axelboc, line 41: We can make this check and the one below more specific.

packages/app/src/providers/models.ts
```diff
@@ -1,5 +1,7 @@
+import type { ArrayShape, Dataset, ScalarShape } from '@h5web/shared';
+
 export interface ValuesStoreParams {
-  path: string;
+  dataset: Dataset<ScalarShape | ArrayShape>;
```
axelboc, line 4: Note that I'm actually enforcing that `valuesStore.getValue()` never gets called with a dataset with a null shape, which is another benefit.

packages/app/src/providers/hsds/hsds-api.ts
loichuder: > I don't think it's worth the complexity.

Yeah agreed. I can live with this.

loichuder (approved):
Very nice :+1:

Issue #807, opened 2021-10-06
Review attribute fetching to improve performance of processing of groups with many children


loichuder:
## Problem

Opening a group with many children is slow. 

This was found to be the case for the `JupyterProvider` (see point 3. of https://github.com/silx-kit/jupyterlab-h5web/issues/51) and should be the same for `H5GroveProvider` and `HsdsProvider`.

## Reason

When processing a group, children are processed as well (https://github.com/silx-kit/h5web/blob/main/packages/app/src/h5web/providers/jupyter/jupyter-api.ts#L109). This processing triggers the fetching of attribute values **for each child**. This abundance of requests is the reason of the slowness.

## Proposed improvement

We could instead have a fetching on-demand of attribute values, removing the fetching of attribute values from the entity processing.

Attribute values are needed in the following scenarios:
1. When inspecting an entity (via `Inspect` button)
1. When processing a NX entity (as a `default` attribute could point to a `NXData` that needs to be displayed). This scenario can be discriminated without additional requests as the group metadata holds the child metadata that include the names of their attributes.

axelboc:
In the end, we always fetch the values of all the attributes of an entity at once, even if we just need one. This doesn't seem to matter much with H5Grove, since it provides an endpoint for this, which is why I haven't gone further. It's more of a problem for HSDS, since the endpoint it provides only allows fetching the value of a single attribute, so we end up making consecutive requests to fetch all the values. I think this is fine for now -- we can optimise things later if the need arises.

PR #844, opened 2021-11-08
Fetch attribute values independently in MetadataViewer

axelboc:
I'm trying to do the refactoring for #807 in a safe way, so I've added the new behaviour on top of the current one. Therefore, for now, the attribute values:

- are fetched twice: first as part of `getEntity` and now also in `getAttrValues`, which is invoked when inspecting a dataset;
- are stored in two places: in the `Entity` objects, and in the new `attrValuesStore`.

I've also limited the refactoring to the `MetadataViewer` for now. The next step will be do deal with NeXus, and the one after that to remove the old attribute-fetching behaviour.

attr-loading

You can see the new suspense loader in the `MetadataViewer` above with HSDS but not with H5Grove, because in `H5GroveApi`, `fetchAttrValues` keeps its own cache (and `getAttrValues` is always called after `getEntity`, so `getAttrValues` hits the cache). Hopefully at the end of this refactoring, this caching will no longer be required.


packages/app/src/metadata-viewer/AttributesInfo.tsx
```diff
@@ -13,28 +14,21 @@ const FOLLOWABLE_ATTRS = new Set([
...
   return (
-    <table className={styles.table}>
-      <thead>
-        <tr>
-          <th scope="col" colSpan={2}>
-            Attributes
-          </th>
-        </tr>
-      </thead>
-      <tbody>
```
axelboc, line 36: I've extracted the `table` wrapper in a separate component for re-use.

packages/app/src/metadata-viewer/AttributesInfo.tsx
```diff
@@ -13,28 +14,21 @@ const FOLLOWABLE_ATTRS = new Set([
...
-function AttributesTable(props: Props) {
-  const { attributes, onFollowPath } = props;
+function AttributesInfo(props: Props) {
+  const { entity, onFollowPath } = props;
 
-  if (attributes.length === 0) {
-    return null;
-  }
+  const { attrValuesStore } = useContext(ProviderContext);
+  const attrValues = attrValuesStore.get(entity);
```
axelboc, line 25: This is from where the new behaviour is triggered.

packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -161,12 +185,9 @@ export class HsdsApi extends ProviderApi {
...
 
-    const attrsPromises = data.attributes.map(async (attr) => {
-      const { data } = await this.client.get<HsdsAttributeResponse>(
-        `/${entityCollection}/${entityId}/attributes/${attr.name}`
-      );
-      return data;
-    });
+    const attrsPromises = data.attributes.map(async (attr) =>
+      this.fetchAttributeWithValue(entityCollection, entityId, attr.name)
+    );
```
axelboc, line 190: The `fetchAttributes` method currently fetches all the attributes' metadata at once, and then each attribute's value one at a time like this. At the end of this refactoring, this look will be gone from here and only occur in `getAttrValues`.

packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -184,8 +205,20 @@ export class HsdsApi extends ProviderApi {
     return data.value;
   }
 
+  private async fetchAttributeWithValue(
+    entityCollection: HsdsCollection,
```
axelboc, line 209: The endpoint needs to know the HSDS entity's `collection`, which is like the `type` but with an `s`... So far we were discarding this information which comes from the link resolution phase. Instead of trying to guess the collection from the type, which is tricky for unresolved entities... I decided to keep the `collection` around like we do for the `id`. That's why you see it a bit everywhere in the diff.

packages/shared/src/mock/utils.ts
```diff
@@ -1,28 +1,25 @@
...
-function isMockDataset(entity: Entity): entity is MockDataset {
-  return isDataset(entity) && 'value' in entity;
-}
-
-export function assertMockDataset(
-  entity: Entity
-): asserts entity is MockDataset {
-  if (!isMockDataset(entity)) {
+export function assertMockDataset<S extends Shape, T extends DType>(
+  dataset: Dataset<S, T>
```
axelboc, line 20: Unrelated change: the guard now accepts a dataset instead of an entity, which means that callers just need to make sure to call `assertDataset` beforehand if they need to (which is not always the case).

packages/shared/src/models-hdf5.ts
```diff
@@ -32,11 +32,6 @@ export interface Dataset<S extends Shape = Shape, T extends DType = DType>
   rawType?: unknown;
 }
 
-export interface MockDataset<S extends Shape = Shape, T extends DType = DType>
-  extends Dataset<S, T> {
-  value: unknown;
-}
-
```
axelboc, line 39: dead code

packages/app/src/metadata-viewer/MetadataViewer.tsx
```diff
@@ -22,19 +31,42 @@ function MetadataViewer(props: Props) {
...
+        <MetadataTable title="Attributes">
+          <ErrorBoundary
+            resetKeys={[path]}
+            FallbackComponent={AttributesErrorFallback}
+          >
+            <Suspense
+              fallback={
+                <tr>
+                  <td className={styles.fallback}>Loading...</td>
+                </tr>
```
loichuder, line 53: Might be worth putting in its own component ? :man_shrugging:

loichuder (approved):
Well-thought-out :+1:

PR #845, opened 2021-11-12
Refactor vis pack resolution

axelboc:
As discussed, before refactoring the attribute fetching, I think we should first get a handle on our pack detection strategy and NeXus default path resolution algorithm. We currently have a few problems caused by the recently added "implicit default entity resolution" feature:

- An entity can be considered to support NeXus in `VisPackChooser` and yet not have any supported NeXus visualization -- e.g. an `NXentry` without a `default` attribute and without an implicit default child.
- An entity can be considered _not_ to support NeXus in `VisPackChooser` and yet have a child that could be considered its implicit default entity -- e.g. `NXprocess` with `NXdata` child (like `/entry_0000/1_integration` in the water file).
- An `NXdata` group is correctly identified as supporting NeXus in `VisPackChooser`, but if it contains a child `NXentry`, this child is considered its implicit default entity and passed to the `Visualizer`.

As a reminder, the current algorithm works roughly like this:

```
- In `VisPackChooser`, if the entity is NXdata or NXentry or NXroot or has `default` attribute, render `NexusPack` ; otherwise, render `CorePack`.
  - In `NexusPack`, look for the default entity, get its supported visualizations, and render the `Visualizer`.
  - In `CorePack`, get the entity's supported visualization and render the `Visualizer`.
    - In `Visualizer`, if the entity supports a visualization, render it ; otherwise, show "no vis" message.
```

The new algorithm removes the implicit coupling between the NeXus pack detection code and the visualization support checks. It works like this:

```
- In `VisPackResolver`, get the entity's supported visualizations.
- If there's at least one supported visualization, render the `Visualizer` ; otherwise, look for a potential default entity (i.e. `default` attribute, or child `NXdata`, or child `NXentry`).
- If a potential default entity is found, recurse by rendering `VisPackResolver` with that entity's path.
- If no entity is found, show "no vis" message.
```

Unfortunately, there's one downside to this algorithm: when we switch between two entities that point to the same default entity (e.g. root entity of `water_224.h5` and `/entry_0000`), **the visualization flashes**. The recursion of `VisPackResolver` causes the `Visualizer` to be rendered at different levels in the React tree, so React has no choice but to unmount it and then re-mount it.

--- 

### UPDATE 1

I moved the recursion back to a utility function called `resolvePath`, which returns the resolved entity along with its supported visualizations. The algorithm now looks like this:

```
- In `VisPackResolver`, call `resolvePath` with the path of the selected entity.
  - In `resolvePath`, fetch the entity and get its supported visualizations.
  - If there's at least one supported vis, return them along with the entity itself ; otherwise, look for an implicit default entity.
  - If an implicit default entity is found, recurse with its path ; otherwise, return `undefined`.
- If `resolvePath` returns `undefined`, show "no vis" message ; otherwise, render the `Visualizer`.
```

In other words the visualization support checks are now part of the default entity resolution algorithm and always happen first (i.e. before recursing).

---

### UPDATE 2

The CI and some more manual testing revealed a couple of remaining issues related to the `isNxGroup` check:

1. An `NXprocess` group with a implicit default child candidate doesn't get the NX badge in the explorer.
2. An `NXentry` child that comes before an `NXdata` child is wrongly chose as the implicit default child.

For 1), I decided to write a custom `isNxGroup` utility for the `Explorer` to remove the coupling with `resolvePath`. For now, I've just added `NXprocess` to the list of NeXus classes that qualify an entity for the `NX` badge. This is not ideal, as an entity can still get the NX badge and not have a visualization (and reciprocally)... but it's a compromise -- we can't do better without actually fetching child entities. In theory, we could fix any false positive/negative once the user has selected the groups and their children have been fetched, but the implementation of the `Explorer` doesn't currently allow for this.

For 2), I replaced `entity.children.find((c) => isNxGroup(c))` with a `getImplicitDefaultChild` utility that loops through the children twice: once to find an `NXdata` child, and then again if needed to find an `NXentry` child.


packages/app/src/vis-packs/nexus/pack-utils.ts
```diff
@@ -2,57 +2,36 @@ import {
...
-import type { FetchStore } from 'react-suspense-fetch';
 
-import { ProviderError } from '../../providers/models';
 import { getAttributeValue } from '../../utils';
 import type { VisDef } from '../models';
 import { NxInterpretation } from './models';
 import { findSignalDataset, isNxDataGroup, isNxGroup } from './utils';
 import { NexusVis, NEXUS_VIS } from './visualizations';
 
-function getDefaultEntityPath(entity: Entity): string | undefined {
```
axelboc, line 22: Basically `getDefaultEntity` becomes `resolvePath` but with the visualisation support check added, and `getDefaultEntityPath` becomes `getNxDefaultPath`.

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -49,8 +49,7 @@ export function isNxGroup(entity: Entity): boolean {
     isGroup(entity) &&
     (isNxDataGroup(entity) ||
       hasAttribute(entity, 'default') ||
-      getAttributeValue(entity, 'NX_class') === 'NXentry' ||
-      getAttributeValue(entity, 'NX_class') === 'NXroot')
```
axelboc, line 53: `isNxGroup` is now called only on children entities, never on the root group.

packages/app/src/visualizer/Visualizer.tsx
```diff
@@ -1,37 +1,37 @@
...
-      <p className={styles.fallback}>
-        No visualization available for this entity.
-      </p>
+  const [activeVis, setActiveVis] = useActiveVis(supportedVis);
+  const { Container, Toolbar } = activeVis;
+
+  if (!('ResizeObserver' in window)) {
+    throw new Error(
+      "Your browser's version is not supported. Please upgrade to the latest version."
     );
```
axelboc, line 25: I've moved the `ResizeObserver` check here, and the "no vis" message is now in `VisPackResolver`.

packages/app/src/visualizer/Visualizer.tsx
```diff
@@ -1,37 +1,37 @@
...
+function Visualizer(props: Props) {
+  const { entity, supportedVis } = props;
+  assertDefined(supportedVis[0], 'Expected supported visualization');
 
-  if (!activeVis) {
-    return (
-      <p className={styles.fallback}>
-        No visualization available for this entity.
-      </p>
+  const [activeVis, setActiveVis] = useActiveVis(supportedVis);
```
axelboc, line 19: `useActiveVis` is now called here instead of duplicated in the now-removed `CorePack` and `NexusPack`. It is now guaranteed to return a visualization.

axelboc (commented):
I'm gonna look at the failing tests, but I wouldn't mind your opinion in the meantime.

packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -103,22 +103,22 @@ test('show error when encountering malformed NeXus metadata', async () => {
...
+  expect(await screen.findByText(/'signal' attribute/)).toBeVisible(); // `no_signal` is default `NXdata` child group
 
   await selectExplorerNode('default_not_string');
   expect(await screen.findByText(/to be a string/)).toBeVisible();
 
   await selectExplorerNode('default_not_found');
-  expect(
-    await screen.findByText(/No entity found at NeXus default path/)
-  ).toBeVisible();
+  expect(await screen.findByText(/No entity found at/)).toBeVisible();
```
axelboc, line 112: We lose the precision of the error message, but it's not a huge deal.

packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -103,22 +103,22 @@ test('show error when encountering malformed NeXus metadata', async () => {
 
   const errorSpy = mockConsoleMethod('error');
   await selectExplorerNode('nexus_malformed');
+  expect(await screen.findByText(/'signal' attribute/)).toBeVisible(); // `no_signal` is default `NXdata` child group
```
axelboc, line 106: The `resilience` and `nexus_malformed` groups now have implicit default `NXdata` children.

packages/app/src/explorer/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+    return true;
+  }
+
+  if (hasAttribute(entity, 'NX_class')) {
+    const nxClass = getAttributeValue(entity, 'NX_class');
+    assertStr(nxClass);
+    return SUPPORTED_NX_CLASSES.has(nxClass);
+  }
+
+  return false;
```
axelboc, line 23: Obviously, this algorithm is not as accurate as the default entity resolution algorithm: if an entity has an implicit default child but doesn't have an `NX_class`, it won't have the NX badge (whereas it should). Conversely, if an `NXprocess` entity doesn't have an implicit default child, it will have the NX badge (whereas it shouldn't). But it will do for now.

packages/app/src/utils.ts
```diff
@@ -9,3 +9,10 @@ export function getAttributeValue(
 ): unknown {
   return entity.attributes?.find((attr) => attr.name === attributeName)?.value;
 }
+
+export function hasAttribute(
+  entity: Entity,
+  attributeName: NxAttribute | ImageAttribute
+) {
+  return entity.attributes.some((attr) => attr.name === attributeName);
```
axelboc, line 17: Moved from `vis-packs/nexus/utils.ts` and refactored in preparation for the fetching of attribute values - i.e. it no longer calls `getAttributeValue`, which was unnecessary -- checking the existence of the attribute is sufficient.

packages/shared/src/mock/metadata.ts
```diff
@@ -151,10 +151,10 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
         }),
       ],
     }),
-    makeNxGroup('nexus_entry_no_default', 'NXentry', {
+    makeNxGroup('nexus_no_default', 'NXprocess', {
```
axelboc, line 154: Replacing the class here to check that we no longer care about it.

packages/shared/src/mock/metadata.ts
```diff
@@ -151,10 +151,10 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
...
       defaultPath: undefined,
       children: [
-        makeNxGroup('ignore_me', 'NXprocess'),
+        makeNxGroup('ignore_me', 'NXentry'),
```
axelboc, line 157: Replacing the class here to check that we look for an `NXdata` child first, regardless of its position.

axelboc (commented):
Ok, it's good to go now. The E2E tests fail because I renamed a mock entity again.

packages/app/src/vis-packs/utils.ts
```diff
@@ -0,0 +1,36 @@
+import type { Entity } from '@h5web/shared';
+
+import { getSupportedVis as getSupportedCoreVis } from './core/pack-utils';
+import type { VisDef } from './models';
+import {
+  getNxDefaultPath,
+  getSupportedVis as getSupportedNxVis,
+} from './nexus/pack-utils';
```
loichuder, line 8: As we no longer have a `visPack` concept, it should not be kept in the structure of the files

packages/shared/src/mock/metadata.ts
loichuder: > to check that we no longer care about it.

Rather to test that `NXprocess` is treated as `NXentry`.

packages/app/src/vis-packs/nexus/pack-utils.ts
```diff
@@ -2,57 +2,49 @@ import {
...
+  // Look for an `NXdata` child group first
+  const nxDataChild = groups.find((g) => hasNxClass(g, 'NXdata'));
+  if (nxDataChild) {
+    return nxDataChild;
   }
 
-  const nxChild = entity.children.find((c) => isNxGroup(c));
-  return nxChild?.path;
+  // Then for an `NXentry` child group
+  return groups.find((g) => hasNxClass(g, 'NXentry'));
```
loichuder, line 28: I wonder... If this checks fails, should we have a fallback to `NXprocess` as well, given the similarity with `NXentry` ?

loichuder (approved):

packages/app/src/vis-packs/nexus/pack-utils.ts
```diff
@@ -2,57 +2,36 @@ import {
...
-    ProviderError.EntityNotFound,
-    `No entity found at NeXus default path "${path}"`
-  );
+    return defaultPath.startsWith('/')
+      ? defaultPath
+      : buildEntityPath(entity.path, defaultPath);
+  }
 
-  return getDefaultEntity(defaultEntity, entitiesStore);
+  assertGroupWithChildren(entity);
```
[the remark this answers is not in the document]
loichuder: For clearer user feedack, you could add a custom error message ? Like `expected Nx group to have children` ?

packages/app/src/vis-packs/VisPackResolver.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const { entitiesStore } = useContext(ProviderContext);
+
+  function getEntity(entityPath: string): Entity {
+    return handleError(
+      () => entitiesStore.get(entityPath),
+      ProviderError.EntityNotFound,
+      `No entity found at ${entityPath}`
+    );
+  }
```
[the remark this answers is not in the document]
axelboc: No, I agree, it's definitely not great.

packages/app/src/vis-packs/nexus/pack-utils.ts
```diff
@@ -2,57 +2,36 @@ import {
...
-  entity: Entity,
-  entitiesStore: FetchStore<Entity, string>
-): Entity {
-  const defaultPath = getDefaultEntityPath(entity);
-  if (defaultPath === undefined) {
-    return entity;
-  }
-  const path = defaultPath.startsWith('/')
-    ? defaultPath
-    : buildEntityPath(entity.path, defaultPath);
```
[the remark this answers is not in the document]
axelboc: `getDefaultEntityPath` used to return the value of the `default` attribute if present, or fall back to returning the path of a child NX entity. In this latter case, the path is necessarily absolute but the ternary would still be processed.

packages/app/src/explorer/utils.ts
axelboc: That's correct. What I find annoying is that we get both false positives (badge shown when it shouldn't) and false negatives (badge not shown when it should). We'll have to see if someone complains about either when confronted to real files.

packages/shared/src/mock/metadata.ts
axelboc: Hmm, for the NX badge yes, but not for the implicit child resolution -- if the `NX_class` is not `NXdata`, then we don't care what it is; it's the `NX_class` of the children that we care about. Does that make sense?

packages/app/src/vis-packs/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+function findSupportedVis(entity: Entity): VisDef[] {
+  const nxVis = getSupportedNxVis(entity);
+  if (nxVis) {
+    return [nxVis];
+  }
```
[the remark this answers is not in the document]
axelboc: I think it make `getSupportedNxVis` less clear, to be honest. I like that it is clear that an entity can only have one supported NX visualization at a time.

packages/app/src/vis-packs/utils.ts
axelboc: I'll investigate this after I've renamed `VisPackResolver` and `Visualizer` and merged the PR :+1:

packages/app/src/vis-packs/nexus/pack-utils.ts
axelboc: Yeah not sure. Maybe you're right. That would actually be more coherent with the NX badge logic.

axelboc: done

packages/app/src/vis-packs/nexus/pack-utils.ts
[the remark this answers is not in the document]
loichuder: Oh all right. I forgot that `assertGroupWithChildren` is to test if the `children` array is present and not if it is not empty.

packages/app/src/explorer/utils.ts
loichuder: These cases should be documented in an issue so that we can see how to move forward

packages/app/src/vis-packs/VisPackResolver.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { ProviderError } from '../providers/models';
+import Visualizer from '../visualizer/Visualizer';
+import styles from '../visualizer/Visualizer.module.css';
+import { resolvePath } from './utils';
+
+interface Props {
+  path: string;
+}
+
+function VisPackResolver(props: Props) {
```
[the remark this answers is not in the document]
axelboc: Will rename `Visualizer` to `VisManager` and `VisPackChooser` to `Visualizer`.

PR #848, opened 2021-11-15
Rename/restructure @h5web/app files

axelboc:
- `VisPackResolver` becomes `Visualizer`
- `Visualizer` becomes `VisManager`
- Utilities for finding supported vis and default entity, previously in `vis-packs/*/pack-utils.ts` are now collocated in `visualizer/utils.ts`.
- The following components used in `App` are now located in the root `src` folder instead of the `visualizer` or `vis-packs` folders: `ErrorFallback`, `VisConfigProvider`.
- The root of the `vis-packs` folder is now reserved for components and models shared by the core/nexus vis containers (i.e. `VisBoundary`) -- anything happening above the level of the vis containers is located in the `visualizer` folder. Arguably, since the nexus vis pack already uses things from the core vis pack, we could move the `VisBoundary` inside the core pack... not sure.


packages/app/src/visualizer/Visualizer.module.css
```diff
@@ -1,17 +1,11 @@
-.visualizer {
+.visPresenter {
```
loichuder, line 1: `visManager` ?

loichuder (approved):

packages/app/src/visualizer/Visualizer.module.css
axelboc: :fishing_pole_and_fish: :eyes: :mag:

PR #849, opened 2021-11-16
Fetch attribute values on demand for NX badges in explorer

axelboc:
Part of #807 

Here I'm dealing with the NX badge case. I introduce an `NxBadge` component that I wrap with error/suspense boundaries. I give these boudaries empty fallbacks so that the NX badges load silently, and also fail silently in the event of network errors.

For the moment, it's difficult to see whether our NX badge heuristics lead to an appropriate number of requests. We'll have to wait until we no longer fetch any attribute values ahead of time before we can tweak things.

### DIFF
--- a/packages/app/src/explorer/EntityItem.tsx
+++ b/packages/app/src/explorer/EntityItem.tsx
@@ -3,12 +3,13 @@ import type { Entity } from '@h5web/shared';
 import { useToggle } from '@react-hookz/web';
 import type { CSSProperties } from 'react';
 import { Suspense, useEffect } from 'react';
+import { ErrorBoundary } from 'react-error-boundary';
 import { FiRefreshCw } from 'react-icons/fi';
 
 import EntityList from './EntityList';
 import styles from './Explorer.module.css';
 import Icon from './Icon';
-import { isNxGroup } from './utils';
+import NxBadge from './NxBadge';
 
 interface Props {
   path: string;
@@ -57,11 +58,12 @@ function EntityItem(props: Props) {
       >
         <Icon entity={entity} isExpanded={isExpanded} />
         <span className={styles.name}>{entity.name}</span>
-        {isNxGroup(entity) && (
-          <span className={styles.nx} aria-label=" (NeXus group)">
-            NX
-          </span>
-        )}
+
+        <ErrorBoundary fallback={<> </>}>
+          <Suspense fallback={null}>
+            <NxBadge entity={entity} />
+          </Suspense>
+        </ErrorBoundary>
       </button>
 
       {isGroup(entity) && isExpanded && (
--- a/packages/app/src/explorer/NxBadge.tsx
+++ b/packages/app/src/explorer/NxBadge.tsx
@@ -1,0 +1,30 @@
+import type { Entity } from '@h5web/shared';
+import { useContext } from 'react';
+
+import { ProviderContext } from '../providers/context';
+import styles from './Explorer.module.css';
+import { needsNxBadge } from './utils';
+
+interface Props {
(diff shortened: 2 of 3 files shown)


packages/app/src/explorer/EntityItem.tsx
```diff
@@ -57,11 +58,12 @@ function EntityItem(props: Props) {
       >
         <Icon entity={entity} isExpanded={isExpanded} />
         <span className={styles.name}>{entity.name}</span>
-        {isNxGroup(entity) && (
-          <span className={styles.nx} aria-label=" (NeXus group)">
-            NX
-          </span>
-        )}
+
+        <ErrorBoundary fallback={<> </>}>
```
axelboc, line 62: `ErrorBoundary` doesn't accept `null` as fallback. The space is to avoid an ESLint warning.

packages/app/src/explorer/utils.ts
```diff
@@ -1,21 +1,23 @@
...
+  group: Group,
+  attrValuesStore: FetchStore<AttributeValues, Entity>
+): boolean {
+  if (hasAttribute(group, 'default')) {
     return true;
   }
 
-  if (hasAttribute(entity, 'NX_class')) {
-    const nxClass = getAttributeValue(entity, 'NX_class');
+  if (hasAttribute(group, 'NX_class')) {
```
axelboc, line 17: Note that I'm fetching the attribute values only if I know there's an attribute called `NX_class`. For now we're fetching the value of all the attributes. We can revisit this later on.

packages/app/src/explorer/utils.ts
```diff
@@ -1,21 +1,23 @@
...
 
 const SUPPORTED_NX_CLASSES = new Set(['NXdata', 'NXentry', 'NXprocess']);
 
-export function isNxGroup(entity: Entity): boolean {
-  if (!isGroup(entity)) {
-    return false;
-  }
-
-  if (hasAttribute(entity, 'default')) {
+export function needsNxBadge(
```
loichuder, line 9: Good call on the renaming, this is much clearer :+1:

packages/app/src/explorer/NxBadge.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface Props {
+  entity: Entity;
+}
+
+function NxBadge(props: Props) {
+  const { entity } = props;
+  const { attrValuesStore } = useContext(ProviderContext);
+
+  if (!isGroup(entity) || !needsNxBadge(entity, attrValuesStore)) {
```
loichuder, line 17: The fact that non-group entities will never get a NX badge is a side-effect from the NX spec. As a consequence, it would make more sense to me to check for `isGroup` inside `needsNxBadge` (which would accept an entity)

packages/app/src/explorer/utils.ts
```diff
@@ -1,21 +1,23 @@
...
+): boolean {
+  if (hasAttribute(group, 'default')) {
     return true;
   }
 
-  if (hasAttribute(entity, 'NX_class')) {
-    const nxClass = getAttributeValue(entity, 'NX_class');
+  if (hasAttribute(group, 'NX_class')) {
+    const attrValues = attrValuesStore.get(group);
+    const nxClass = attrValues?.NX_class;
```
loichuder, line 19: At some point, we may put this inside a `getAttributeValue` function.

loichuder (approved):

packages/app/src/explorer/utils.ts
axelboc: Yeah, I'd like to improve the stores' APIs.

PR #850, opened 2021-11-16
Fetch attribute values on demand in `resolvePath`

axelboc:
Part of #807 

Pretty straight-forward. Again, we may find we need to optimise things later on. For now, I'm always fetching all the attribute values after fetching the entity, but we could imagine fetching only the `default` and `interpretation` attributes when we know they're there.


loichuder (approved):

axelboc:
Arf have to rethink the tests...

axelboc:
It no longer makes much sense to unit test `getSupportedCore/NxVis` since we have to fetch stuff, so I removed the unit tests and added a couple of feature tests with mock entities to cover the cases that I thought were still worth testing.

packages/app/src/vis-packs/nexus/models.ts
```diff
@@ -23,12 +23,6 @@ export type NxAttribute =
   | 'SILX_style'
   | 'auxiliary_signals';
 
-export enum NxInterpretation {
-  Spectrum = 'spectrum',
-  Image = 'image',
-  RGB = 'rgb-image',
-}
-
```
axelboc, line 31: Already in @h5web/shared.

packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -89,6 +89,24 @@ test('visualize NXdata group with "rgb-image" interpretation', async () => {
   expect(tabs[0]).toHaveTextContent(NexusVis.NxRGB);
 });
 
+test('visualize NXdata group with unknown interpretation', async () => {
```
loichuder, line 92: We should add somewhere (comment or test title) that the NXdata group should also have more than 1 dimension. Otherwise, the vis will be `NxSpectrum` and not `NxImage`.

loichuder (approved):
Great, features tests already covered a lot of stuff anyway

PR #851, opened 2021-11-16
Fetch `NX_class` on demand when checking if entity is `NXdata` group

axelboc:
Part of #807 

I add a new `attrValuesStore.getSingle()` method that allows retrieving the value of a specific attribute. It still fetches the values of all the attributes, but it returns only the value of the attribute we're interested in. Moreover, the fetch only happens if we know that the attribute exists.

Also, as per our discussion in the meeting this morning, I know consider that a group without a `signal` attribute is _not_ an `NXdata` group. I even check this first, before the value of the `NX_class`, so that we don't bother fetching the attribute values at all if the entity has no `signal` attribute. As a result of this change, if an entity has an `NXdata` class but no `signal` attribute, we no longer show the "Expected `signal` attribute` error -- we show the "no vis" fallback message instead.

### DIFF
--- a/packages/app/src/__tests__/NexusPack.test.tsx
+++ b/packages/app/src/__tests__/NexusPack.test.tsx
@@ -116,30 +116,35 @@ test('visualize NXentry group with implicit default child NXdata group', async (
   expect(tabs[0]).toHaveTextContent(NexusVis.NxSpectrum);
 });
 
-test('show error when encountering malformed NeXus metadata', async () => {
+test('show error when `default` entity is not found', async () => {
   await renderApp();
 
   const errorSpy = mockConsoleMethod('error');
-  await selectExplorerNode('nexus_malformed');
-  expect(await screen.findByText("Expected 'signal' attribute")).toBeVisible(); // `nexus_malformed` resolves to implicit default child `no_signal`
-
-  await selectExplorerNode('default_not_string');
-  expect(
-    await screen.findByText("Expected 'default' attribute to be a string")
-  ).toBeVisible();
-
-  await selectExplorerNode('default_not_found');
+  await selectExplorerNode('nexus_malformed/default_not_found');
   expect(await screen.findByText('No entity found at /test')).toBeVisible();
 
-  await selectExplorerNode('no_signal');
-  expect(await screen.findByText("Expected 'signal' attribute")).toBeVisible();
+  expect(errorSpy).toHaveBeenCalledTimes(2); // React logs two stack traces
+});
 
-  await selectExplorerNode('signal_not_found');
+test('show error when `signal` entity is not found', async () => {
+  await renderApp();
+
+  const errorSpy = mockConsoleMethod('error');
+  await selectExplorerNode('nexus_malformed/signal_not_found');
   expect(
     await screen.findByText('Expected "unknown" signal entity to exist')
   ).toBeVisible();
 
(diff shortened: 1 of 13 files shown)


packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -116,30 +116,35 @@ test('visualize NXentry group with implicit default child NXdata group', async (
   expect(tabs[0]).toHaveTextContent(NexusVis.NxSpectrum);
 });
 
-test('show error when encountering malformed NeXus metadata', async () => {
+test('show error when `default` entity is not found', async () => {
```
axelboc, line 119: I've split the `nexus_malformed` tests, since the `no_signal` one no longer shows an error. I've also removed the `default_not_string` test (and mock entity), as I no longer think it's worth having given how unlikely it is to occur.

packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -116,30 +116,35 @@ test('visualize NXentry group with implicit default child NXdata group', async (
...
   await renderApp();
 
   const errorSpy = mockConsoleMethod('error');
-  await selectExplorerNode('nexus_malformed');
```
axelboc, line 123: Note that the `nexus_malformed` entity, which still has `no_signal` as its implicit default child, now shows the "no vis" fallback message too.

packages/app/src/explorer/utils.ts
```diff
@@ -18,10 +18,8 @@ export function needsNxBadge(
     return true;
   }
 
-  if (hasAttribute(entity, 'NX_class')) {
-    const attrValues = attrValuesStore.get(entity);
-    const nxClass = attrValues?.NX_class;
-
+  const nxClass = attrValuesStore.getSingle(entity, 'NX_class');
+  if (nxClass) {
```
axelboc, line 22: The `hasAttribute` check/optimisation is now done in `attrValuesStore.getSingle()`, so we don't have to worry about it here.

packages/app/src/vis-packs/nexus/hooks.ts
```diff
@@ -1,10 +1,38 @@
...
+  findSignalDataset,
+  findTitleDataset,
+  getDatasetLabel,
+  getSilxStyle,
+} from './utils';
+
+export function useNxData(group: GroupWithChildren): NxData {
+  const { attrValuesStore } = useContext(ProviderContext);
+
+  assertNxDataGroup(group, attrValuesStore);
```
axelboc, line 22: `useNxData` is now a hook so I can get `attrValuesStore` and pass it to `assertNxDataGroup`, which passes it to `isNxDataGroup`.

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -25,15 +24,25 @@ import type {
...
 
-export function isNxDataGroup(group: Group): boolean {
-  return getAttributeValue(group, 'NX_class') === 'NXdata';
+export function isNxDataGroup(
+  group: Group,
+  attrValuesStore: AttrValuesStore
+): boolean {
+  return (
+    hasAttribute(group, 'signal') &&
+    attrValuesStore.getSingle(group, 'NX_class') === 'NXdata'
```
axelboc, line 37: As mentioned:

- we no longer show an error if an `NXdata` group has no `signal` attribute (via the assertion below)
- we avoid the fetch of the `NX_class` if the entity has no `signal` attribute

PR #858, opened 2021-11-23
Fetch `NX_class` attributes on demand when looking for implicit default child

axelboc:
Part of #807

### DIFF
--- a/packages/app/src/visualizer/utils.ts
+++ b/packages/app/src/visualizer/utils.ts
@@ -12,14 +12,11 @@ import {
 } from '@h5web/shared';
 
 import type { AttrValuesStore } from '../providers/context';
+import { hasAttribute } from '../utils';
 import type { CoreVisDef } from '../vis-packs/core/visualizations';
 import { Vis, CORE_VIS } from '../vis-packs/core/visualizations';
 import type { VisDef } from '../vis-packs/models';
-import {
-  findSignalDataset,
-  hasNxClass,
-  isNxDataGroup,
-} from '../vis-packs/nexus/utils';
+import { findSignalDataset, isNxDataGroup } from '../vis-packs/nexus/utils';
 import { NexusVis, NEXUS_VIS } from '../vis-packs/nexus/visualizations';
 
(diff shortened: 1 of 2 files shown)


packages/app/src/visualizer/utils.ts
```diff
@@ -121,21 +118,34 @@ function getSupportedNxVis(
...
 
-function getImplicitDefaultChild(children: Entity[]): Entity | undefined {
-  const groups = children.filter(isGroup);
+function getImplicitDefaultChild(
+  children: Entity[],
+  attrValueStore: AttrValuesStore
+): Entity | undefined {
+  const nxGroups = children
+    .filter(isGroup)
+    .filter((g) => hasAttribute(g, 'NX_class'));
```
axelboc, line 127: We're only fetching attribute values of groups that we know have an `NX_class` attribute

packages/app/src/visualizer/utils.ts
```diff
@@ -121,21 +118,34 @@ function getSupportedNxVis(
...
 
   // Look for an `NXdata` child group first
-  const nxDataChild = groups.find((g) => hasNxClass(g, 'NXdata'));
+  const nxDataChild = nxGroups.find(
+    (g) => attrValueStore.getSingle(g, 'NX_class') === 'NXdata'
+  );
```
axelboc, line 132: This will trigger requests for attribute values sequentially until it finds the first `NXdata` group. Note that the other `find` loops below won't, since the responses will be cached, so the worst case scenario is when a group has thousands of `NXcollection` children. Since we don't check the `NX_class` of the parent group when looking for an implicit default child, this can potentially happen on any group that contains NX child groups. We could mitigate this by disabling implicit child resolution for non-`NXentry` groups for instance, but I would rather wait for the need to arise in a real use case.

loichuder: Yes, I see how that could become a problem but it seems unlikely. I agree that waiting for the case to actually show up is better than premature optimization.

loichuder (approved):

PR #861, opened 2021-11-24
Fetch attributes on demand when checking for RBG vis support

axelboc:


packages/app/src/vis-packs/core/visualizations.test.ts
```diff
@@ -47,150 +51,150 @@ describe('Raw', () => {
   const { supportsDataset } = CORE_VIS.Raw;
 
   it('should support any dataset', () => {
-    expect(supportsDataset(datasetIntScalar)).toBe(true);
-    expect(supportsDataset(datasetStr2D)).toBe(true);
+    expect(supportsDataset(datasetIntScalar, mockStore)).toBe(true);
+    expect(supportsDataset(datasetStr2D, mockStore)).toBe(true);
```
axelboc, line 55: The downside here is that we need to pass a mock `attrValuesStore` in which `getSingle` is proxied to `getAttributeValue` on every call... That's why I didn't do this with other unit tests, and decided to replace them with feature tests. Here, this decision is more debatable, as we're checking for a lot more edge cases...

loichuder: Yeah, tossing around `attrValuesStore` is getting a bit ugly. 

Could we think about something more elegant/encapsulated such as using hooks to get attribute values (putting the `useContext` inside ? Or could we imagine fetching all attributes for the current entity to have something closer than our `getAttributeValue` from before while still tackling #807 ?

loichuder (approved):
Nah, after review, passing `attrValueStore` as arg is only done in `visualizer/utils.ts` which make sense.

This will do.

PR #863, opened 2021-11-25
Stop fetching attribute values automatically when fetching entities

### LINKED ISSUE #807: Review attribute fetching to improve performance of processing of groups with many children

axelboc:
At last, this closes #807.

- Attribute values are no longer fetched automatically as part of `getEntity()`.
- I remove the `value` property from the `Attribute` model, and create a new `MockAttribute` model with a `value` property for the mock provider.
- I remove `getAttributeValue` (rather, I move it to the mock store in `visualizations.test.ts` and adapt it to assert that the attribute is a mock attribute).


packages/app/src/providers/Provider.tsx
```diff
@@ -68,7 +68,10 @@ function Provider(props: Props) {
   const filepathMembers = api.filepath.split('/');
 
   const attrValuesStore = useMemo(() => {
-    const store = createFetchStore(api.getAttrValues.bind(api));
+    const store = createFetchStore(api.getAttrValues.bind(api), {
+      type: 'Map',
+      areEqual: (a, b) => a.path === b.path,
```
axelboc, line 73: I realised when removing the internal cache from `h5grove-api` that the store's cache wasn't working properly.

loichuder (approved):

Issue #862, opened 2021-11-25
Add components to select/mask part of the data


loichuder:
Not needed for the viewer but useful for other apps that use `@h5web/lib` as a data visualization lib.

A starting point could be offering the possibility to the user to draw a selection on the data of rectangular shape. The selection should then be translated into a range of abscissas/ordinates (and/or into the corresponding data) to make further operations (e.g. removing the selection from domain computation).

More involved topics to be tackled later include:
- Resizing/moving the selection
- Drawing multiple selections
- Selections of different shapes

Pinging @t20100 here as the masking features of `silx` can help in designing this.

loichuder:
Selection components are demonstrated in https://h5web-docs.panosc.eu/?path=/story/building-blocks-selection

PR #864, opened 2021-11-25
Add a component to be able to draw selections on data

loichuder:
Answering the most basic need of #862

This is mostly to discuss integration/organization in the exisiting projects and packages.

### DIFF
--- a/apps/storybook/src/VisCanvasInteraction.stories.tsx
+++ b/apps/storybook/src/VisCanvasInteraction.stories.tsx
@@ -1,4 +1,4 @@
-import { PanZoomMesh, TooltipMesh, VisCanvas } from '@h5web/lib';
+import { PanZoomMesh, TooltipMesh, SelectionMesh, VisCanvas } from '@h5web/lib';
 import type { TooltipMeshProps } from '@h5web/lib';
 import { formatTooltipVal } from '@h5web/shared';
 import type { Meta, Story } from '@storybook/react';
@@ -9,10 +9,11 @@ interface TemplateProps {
   panZoom?: boolean;
   tooltipValue?: string;
   guides?: TooltipMeshProps['guides'];
+  drawSelection?: boolean;
 }
 
 const Template: Story<TemplateProps> = (args) => {
-  const { panZoom = false, tooltipValue, guides } = args;
+  const { panZoom = false, tooltipValue, guides, drawSelection = false } = args;
 
   return (
     <VisCanvas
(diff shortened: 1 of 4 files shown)


apps/storybook/src/VisCanvasInteraction.stories.tsx
```diff
@@ -9,10 +9,11 @@ interface TemplateProps {
   panZoom?: boolean;
   tooltipValue?: string;
   guides?: TooltipMeshProps['guides'];
+  drawSelection?: boolean;
 }
 
 const Template: Story<TemplateProps> = (args) => {
-  const { panZoom = false, tooltipValue, guides } = args;
+  const { panZoom = false, tooltipValue, guides, drawSelection = false } = args;
```
loichuder, line 16: For now, one can test this in the `VisCanvasInteraction` story

packages/lib/src/index.ts
```diff
@@ -36,6 +36,7 @@ export { default as HeatmapMesh } from './vis/heatmap/HeatmapMesh';
 export { default as DataCurve } from './vis/line/DataCurve';
 export { default as Html } from './vis/shared/Html';
 export { default as Annotation } from './vis/shared/Annotation';
+export { default as SelectionMesh } from './vis/shared/SelectionMesh';
```
loichuder, line 39: I made `SelectionMesh` part of `lib` since it is quite close to `TooltipMesh` and `PanZoomMesh`. I have to export it to reuse it in the storybook.

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -0,0 +1,53 @@
...
+    (target as Element).releasePointerCapture(pointerId);
+    setEndPoint(new Vector2(unprojectedPoint.x, unprojectedPoint.y));
+    setDrag(false);
+  }, []);
+
+  return (
+    <>
+      <mesh {...{ onPointerMove, onPointerUp, onPointerDown }}>
+        <meshBasicMaterial opacity={0} transparent color="#0000ff" />
+        <planeGeometry args={[width, height]} />
```
axelboc, line 44: We actually have a `VisMesh` for this :wink: -- which, I'm noticing, we should use in `PanZoomMesh` too.

axelboc (approved):
Looks great!

So do you want to move it to a separate package now or later?

packages/lib/src/vis/shared/SelectionMesh.tsx
loichuder: Good point !

I am not sure for `PanZoomMesh` though: `VisMesh` covers only the vis space and not the canvas space which would prevent from panning/zooming when the mouse is outside the vis but in the canvas.

loichuder:
> So do you want to move it to a separate package now or later?
 
Um, later ? :sweat_smile:

PR #867, opened 2021-11-25
Wait for color map scale to change in E2E tests to avoid failures

axelboc:
Let's see how this goes.


axelboc:
All passed without any failed attempts! :tada:

loichuder (approved):

PR #871, opened 2021-11-25
Let providers throw user-facing errors

axelboc:
This simplifies things a bit. We no longer have to wrap `entitiesStore.get()` with `handleError`, or create a `getEntity` function in `Visualizer` to pass to `resolvePath`.

### DIFF
--- a/packages/app/src/explorer/utils.ts
+++ b/packages/app/src/explorer/utils.ts
@@ -1,7 +1,7 @@
 import type { Entity } from '@h5web/shared';
 import { isGroup, assertStr } from '@h5web/shared';
 
-import type { AttrValuesStore } from '../providers/context';
+import type { AttrValuesStore } from '../providers/models';
 import { hasAttribute } from '../utils';
 
 const SUPPORTED_NX_CLASSES = new Set(['NXdata', 'NXentry', 'NXprocess']);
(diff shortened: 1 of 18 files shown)


packages/app/src/explorer/utils.ts
```diff
@@ -1,7 +1,7 @@
 import type { Entity } from '@h5web/shared';
 import { isGroup, assertStr } from '@h5web/shared';
 
-import type { AttrValuesStore } from '../providers/context';
+import type { AttrValuesStore } from '../providers/models';
```
axelboc, line 4: I've moved the models of the stores to `models.ts`.

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -81,7 +80,17 @@ export class H5GroveApi extends ProviderApi {
           return undefined;
         }
 
-        return getMatchingProviderError(errorData);
+        if (errorData.includes('File not found')) {
+          return `File not found: '${this.filepath}'`;
```
axelboc, line 84: I need `this.filepath`, so it was easier to just inline the logic here.

loichuder (approved):
:sparkles:

PR #874, opened 2021-11-26
Fetch only slice for LineVis of huge datasets

loichuder:
A tentative to solve https://github.com/silx-kit/h5web/issues/616#issuecomment-964256705, open for discussion.

`LimitedValueFetcher` is a `ValueFetcher` that checks if the full dataset can be fetched. If not, it fetches the corresponding slice and passes a boolean to the render function to tell that only a slice was fetched. I make use of this boolean to know if autoscale should be disabled.

Note that because I added some code to compute the size of the value according to the dataset metadata, I had to ensure the fact that a complex dtype holds two numeric dtypes.

If we go this way, `LimitedValueFetcher` should also be used in `NxSpectrum` and `NxComplexSpectrum` vis.

### DIFF
--- a/packages/app/src/providers/hsds/utils.ts
+++ b/packages/app/src/providers/hsds/utils.ts
@@ -1,25 +1,25 @@
 import type {
-  Group,
-  Dataset,
-  Entity,
-  Shape,
-  NumericType,
-  CompoundType,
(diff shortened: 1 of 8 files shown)


loichuder:
Superseded by #877

PR #875, opened 2021-11-26
Show readable dtype in tooltip

### LINKED ISSUE #508: Show data type in visualizer
According to some user feedback, it may be relevant to display the type of data (int8, etc..) in the visualizer, for instance in the dimension mapper. It's not great to have to switch to the inspector for this, and this information is not present in the inspector anyway for NeXus entities.

axelboc:
Friday POC to fix #508, which has been open for a long time.

I pass the dataset type object (`NumericType`) from the container down to the visualization so it can be displayed in the tooltip, next to the hovered value. I let the tooltip's render prop convert the type to a readable string. It seems more straightforward, from a naming point of view, to do it this way than to format the readable string in the container and pass that down to the visualization.

I've only done `LineVisContainer` as a proof of concept for now:

image

image

### DIFF
--- a/packages/app/src/vis-packs/core/heatmap/HeatmapVisContainer.tsx
+++ b/packages/app/src/vis-packs/core/heatmap/HeatmapVisContainer.tsx
@@ -39,6 +39,7 @@ function HeatmapVisContainer(props: VisContainerProps) {
               dims={dims}
               dimMapping={dimMapping}
               title={entity.name}
+              dtype={entity.type}
             />
           )}
         />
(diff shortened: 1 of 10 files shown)


axelboc:
If you recall, showing the readable dtype makes particular sense in NeXus visualization, as switching to "Inspect" doesn't give you the dtype of the signal.

I think the tooltip is an easy and convenient place to show this information in all visualizations, but it does mean adding a few props... One alternative solution would be to show the dtype of the signal for NeXus groups in the `MetadataViewer`, but it's trickier.

loichuder:
It looks very good: very nice idea to put it in the tooltip ! I have no objection to generalise it to other vis.

Complexes will be tricky though. I think you will need some of the improvements I made in #874 to enforce numerical dtypes for complex fields.

axelboc:
image

image

loichuder (approved):

PR #877, opened 2021-11-26
 Fetch whole dataset only when autoScale is off (and slices otherwise)

loichuder:
Improvement for #616 

Also:
- Enforce numeric types for complex fields
- Remove persistence of `autoScale`

Drafting it for discussion as I feel we can do better here:
- There is a lot of common behaviour in the various `LineVis` containers that we could refactor
- I liked that fact that `useLineConfig` was always used in `Mapped` components but now, I need it outside...


packages/app/src/vis-packs/core/complex/ComplexLineVisContainer.tsx
```diff
@@ -20,19 +21,19 @@ function ComplexLineVisContainer(props: VisContainerProps) {
...
       <DimensionMapper
         rawDims={dims}
         mapperState={dimMapping}
         onChange={setDimMapping}
       />
-      <VisBoundary
-        resetKey={dimMapping}
-        loadingMessage="Loading entire dataset"
-      >
+      <VisBoundary resetKey={dimMapping} loadingMessage="Loading current slice">
```
axelboc, line 33: The loading message depends on `autoScale`. Might be worth adding a boolean prop to move the logic of the message inside `VisBoundary`.

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -57,13 +57,17 @@ export function useDatasetValue<D extends Dataset<ArrayShape | ScalarShape>>(
...
+  dimMapping?: DimensionMapping
 ): Record<string, Value<D>> {
   const { valuesStore } = useContext(ProviderContext);
 
   return Object.fromEntries(
     datasets.map((dataset) => {
-      const value = valuesStore.get({ dataset });
+      const value = valuesStore.get({
+        dataset,
+        selection: getSliceSelection(dimMapping),
```
axelboc, line 69: You could store  the selection at the start of the function.

packages/app/src/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -85,7 +91,7 @@ function MappedLineVis(props: Props) {
   }, [disableErrors, errors]);
 
   useEffect(() => {
-    // Disable `autoScale` for 1D datasets (baseArray and dataArray are the same)
+    // Disable `autoScale` for 1D datasets (baseArray and dataArray are the same) or if the value is a slice
```
axelboc, line 94: Are you sure about this? `dims` is the shape of the full dataset, not the slice. I think it still works the same, which is good.

axelboc (approved):
Apart from the loading message logic, which can be moved inside `VisBoundary`, I don't see a lot of duplication between the line and spectrum containers. Where there's more duplication now is between the line and heatmap containers. However, I don't see it as a problem, as all the business logic is nicely encapsulated in shared hooks and components.

Comparing the containers, I've actually just noticed that we have duplicate NeXus type guards: `assertNumericNxData` and `assertNumericSignal`, and `assertComplexNxData` and `assertComplexSignal`.

The fact that we use `useLineConfig` in the containers can indeed feel like we're breaking one of our design decisions, but perhaps this wasn't a great decision in the first place. It makes sense that the user's preferences would eventually affect the fetching of the data somehow. Perhaps relaxing this decision will open the door to some refactoring to reduce duplication/complexity in the mapped vis  components?

packages/app/src/vis-packs/core/line/MappedLineVis.tsx
loichuder: Sorry, leftover from my previous attempt, I will remove this

packages/app/src/vis-packs/core/complex/ComplexLineVisContainer.tsx
axelboc: Well, I was thinking this would apply to `HeatmapVisContainer` and the like as well, but maybe you're right, it's a bit more complex than that since for NeXus visualizations, we load more than one dataset...

loichuder: ALl right, I will but a ternary on both `ComplexLineVisContainer` and `LineVisContainer` and call it a day

packages/app/src/__tests__/CorePack.test.tsx
```diff
@@ -72,3 +73,54 @@ test('visualize 2D datasets', async () => {
...
+
+  // Check that entire dataset is fetched
+  d0Slider.focus();
+  userEvent.keyboard('{PageUp}');
+  expect(await screen.findByRole('figure')).toBeVisible();
+  d0Slider.blur(); // remove focus to avoid state update after unmount
+
+  jest.runOnlyPendingTimers();
+  jest.useRealTimers();
+});
```
loichuder, line 126: New test to test the new auto scale behaviour. It should perhaps be put in `Visualizer.test` ?

axelboc: It's quite intense :sweat_smile: -- I would perhaps split the test in two: "Visualize 3D dataset as line with auto-scale enabled" and  "Visualize 3D dataset as line with auto-scale disabled".

axelboc (changes requested):

packages/app/src/__tests__/CorePack.test.tsx
loichuder: Well, in the end, I still need to go through all the previous steps to test auto-scale disabled so I might as well keep the same test.

I changed the comments to be more meaningful instead.

packages/app/src/__tests__/CorePack.test.tsx
```diff
@@ -72,3 +73,61 @@ test('visualize 2D datasets', async () => {
...
+    name: 'Dimension slider',
+  })[0];
+  d0Slider.focus();
+  userEvent.keyboard('{PageUp}');
+  expect(await screen.findByText(/Loading current slice/)).toBeVisible();
+
+  // Let the new slice fetch succeed
+  jest.runAllTimers();
+  expect(await screen.findByRole('figure')).toBeVisible();
+  d0Slider.blur(); // remove focus to avoid state update after unmount
```
axelboc, line 115: I don't think you need this one, since you're focusing something else later. You can also use `{ArrowUp/Down}` instead of `{PageUp/Down}` to simplify things.

axelboc (approved):
nice!

PR #878, opened 2021-11-29
Render toolbars from mapped vis components

axelboc:
This is in prevision of adding a button to export an image to TIFF. Currently, toolbars are rendered from `VisManager`, which means they don't have access to information such as the data domain, the dim mapping state, or even the kind/shape/type of the entity. So far, we had to store such information in the toolbar's config store -- we'd call the relevant setters from the mapped vis via `useEffect`. In addition to the complexity, this asynchronous behaviour would cause race conditions and other issues in feature and E2E tests.

Now, the toolbars are rendered from the mapped vis components in portals, which means that they can be passed the information they need directly through props.

The only visible impact for users is that the toolbars are now rendered _after_ the data has finished loading. It means the toolbars disappear/reappear when slicing, for instance. If we render the toolbars in the containers instead of the mapped vis, outside of the `VisBoundary`, we can't pass them the data, domain, etc.

To demonstrate the benefit of this refactoring as it stands, I pass the `currentSlice` ndarray to the `MatrixToolbar` via a prop instead of setting it in the matrix config store (store which I can now remove). More refactoring to come (`dataDomain`, `SILX_style`, etc.)


packages/app/src/visualizer/VisManager.tsx
```diff
@@ -16,29 +17,37 @@ function VisManager(props: Props) {
...
   if (!('ResizeObserver' in window)) {
     throw new Error(
       "Your browser's version is not supported. Please upgrade to the latest version."
     );
   }
 
+  const [activeVis, setActiveVis] = useActiveVis(supportedVis);
+  const { Container } = activeVis;
+
+  const [visBarElem, setVisBarElem] = useState<HTMLDivElement>();
```
axelboc, line 29: I can't use `useRef`, because I need a re-render to be triggered once I have a reference to the element.

loichuder (approved):
:+1:

PR #885, opened 2021-11-30
Add "Export" menu to toolbars with CSV/NPY/TIFF exports

### LINKED ISSUE #35: Copying/exporting datasets

axelboc:
Fix #35 

The feature detection is actually quite simple: I've discovered that an abstract class can have an optional method! So if the provider implements the method, it is defined; otherwise it isn't. This means that the toolbar can just check if `getTiffUrl` is defined to decide whether to show the export button.

I've provided a tentative implementation of the method for H5Grove. It returns a URL of this form: `https://<h5grove.url>/tiff/?<axios-params>&path=/foo&selection=0:,:`. Up to H5Grove to then respond at this URL with the TIFF data and the correct `Content-type` HTTP header (`image/tiff`).

image

---

### Update

I've merged #890 into this branch, so now it includes CSV and NPY exports as well.

### DIFF
--- a/packages/app/src/providers/h5grove/h5grove-api.ts
+++ b/packages/app/src/providers/h5grove/h5grove-api.ts
@@ -33,7 +33,7 @@ import {
 } from './utils';
 
 export class H5GroveApi extends ProviderApi {
-  /* API compatible with h5grove@0.0.9 */
+  /* API compatible with h5grove@0.0.13 */
   public constructor(
     url: string,
     filepath: string,
--- a/packages/app/src/providers/h5grove/h5grove-api.ts
+++ b/packages/app/src/providers/h5grove/h5grove-api.ts
@@ -1,10 +1,12 @@
 import type {
+  ArrayShape,
   Attribute,
   AttributeValues,
   Dataset,
   Entity,
   Group,
   GroupWithChildren,
+  NumericType,
   UnresolvedEntity,
 } from '@h5web/shared';
 import {
@@ -69,6 +71,23 @@ export class H5GroveApi extends ProviderApi {
   public async getAttrValues(entity: Entity): Promise<AttributeValues> {
     const { path, attributes } = entity;
     return attributes.length > 0 ? this.fetchAttrValues(path) : {};
+  }
+
+  public getTiffUrl(
+    dataset: Dataset<ArrayShape, NumericType>,
+    selection: string | undefined
+  ): string | undefined {
+    const { baseURL, params } = this.client.defaults;
+
+    const searchParams = new URLSearchParams(params as Record<string, string>);
+    searchParams.set('path', dataset.path);
+    searchParams.set('format', 'tiff');
+
+    if (selection) {
+      searchParams.set('selection', selection);
+    }
+
+    return `${baseURL as string}/data/?${searchParams.toString()}`;
   }
 
   private async fetchEntity(path: string): Promise<H5GroveEntityResponse> {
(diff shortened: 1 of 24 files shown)


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -71,6 +73,22 @@ export class H5GroveApi extends ProviderApi {
...
+    const { baseURL, params } = this.client.defaults;
+
+    const searchParams = new URLSearchParams(params as Record<string, string>);
+    searchParams.set('path', dataset.path);
+
+    if (selection) {
+      searchParams.set('selection', selection);
+    }
+
+    return `${baseURL as string}/tiff/?${searchParams.toString()}`;
```
axelboc, line 89: This is a bit complicated, because it turns out that axios doesn't currently provide a way to retrieve the URL it constructs internally based on the initial configuration (`baseUrl` and `params` in our case)...

packages/lib/src/toolbar/controls/DownloadBtn.tsx
```diff
@@ -8,8 +8,8 @@ interface Props extends AriaAttributes {
   iconOnly?: boolean;
   // If specified, browser will prompt user to download the file instead of opening a new tab
   filename?: string;
-  // Invoked on click; must return URL to set as `href`, or `false` to do nothing
-  getDownloadUrl: () => string | false;
+  // Invoked on click; must return URL to set as `href`, or `undefined` to do nothing
+  getDownloadUrl: () => string | undefined;
```
axelboc, line 12: On second thought, I'm not a fan of this API... `undefined` makes more sense.

packages/app/src/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -58,7 +63,9 @@ function MappedHeatmapVis(props: Props) {
       {toolbarContainer &&
         createPortal(
           <HeatmapToolbar
+            dataset={dataset}
```
axelboc, line 66: I'm now passing the whole dataset. It's just easier.

packages/app/src/providers/h5grove/h5grove-api.ts
loichuder: [axios.getUri](https://github.com/axios/axios#axiosgeturiconfig) does not give what you want ?

loichuder (approved):

packages/app/src/providers/h5grove/h5grove-api.ts
axelboc: https://github.com/axios/axios/pull/3737 :cry:

axelboc:
Marking the PR as "waiting" until you find the time to implement the TIFF endpoint in H5Grove.

packages/app/src/providers/h5grove/h5grove-api.ts
loichuder: Unfortunate. You could still make use of it for the parameters serialization but at the risk of being broken when/if the return value include `baseUrl`.

Let's keep it like this for now :+1:

loichuder:
> Marking the PR as "waiting" until you find the time to implement the TIFF endpoint in H5Grove.

The demo is now requesting the tentative implementation at https://github.com/silx-kit/h5grove/pull/54. You can give it a try

axelboc:
I'm getting a black TIFF file, so I assume it works!

axelboc:
We'll just have to update the compatibility comment in `h5grove-api` once you've merged the PR on H5Grove.

PR #886, opened 2021-11-30
Cancel on-going requests when changing entity or vis

### LINKED ISSUE #881: Cancel ongoing requests when switching entity

loichuder:
Fix #881


axelboc (approved):
:tada:

PR #888, opened 2021-12-02
Debounce dim mapping state to reduce fetching rate when slicing

### LINKED ISSUE #879: [DimensionMapper] Be less eager to fetch when slicing
This is a problem that appears when slicing through a dataset with heavy slices. 

Say I want to look at the latest slice of the slicing dimension: I need to make the thumb go through many (if not all) slices between the first and the last.

This triggers many slow slice fetches which are not useful as I am only interested in the last one in my example.

We could have some throttling or launch the fetch only when releasing the thumb.

loichuder:
Fix #879

### DIFF
--- a/packages/app/src/dimension-mapper/hooks.ts
+++ b/packages/app/src/dimension-mapper/hooks.ts
@@ -1,10 +1,13 @@
-import { useState } from 'react';
+import { useDebouncedState } from '@react-hookz/web';
 
 import type { DimensionMapping } from './models';
 
 export function useDimMappingState(dims: number[], axesCount: number) {
-  return useState<DimensionMapping>([
-    ...Array.from({ length: dims.length - axesCount }, () => 0),
-    ...['y' as const, 'x' as const].slice(-axesCount),
-  ]);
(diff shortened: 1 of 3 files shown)


packages/app/src/dimension-mapper/hooks.ts
```diff
@@ -1,10 +1,13 @@
...
-  return useState<DimensionMapping>([
-    ...Array.from({ length: dims.length - axesCount }, () => 0),
-    ...['y' as const, 'x' as const].slice(-axesCount),
-  ]);
+  return useDebouncedState<DimensionMapping>(
+    [
+      ...Array.from({ length: dims.length - axesCount }, () => 0),
+      ...['y' as const, 'x' as const].slice(-axesCount),
+    ],
+    50
```
loichuder, line 11: This number may be changed but `50` is already enough to prevent fetching when doing a continuous motion with the slider thumb

axelboc (approved):
Ha, very simple and elegant. I had thought of debouncing the request, but this is much better.

PR #889, opened 2021-12-02
Improve matrix vis performance through virtualisation of index/header cells

### LINKED ISSUE #880: [Matrix] Virtualisation performance issue with large datasets

axelboc:
Fix #880 

The header cells are now virtualized in the same way as the content cells. I tried a lot of things to keep the sticky behaviour and ended up with the current solution, which relies on `react-window`'s `onRenderedItems` callback. The challenge was then to provide the rendered items to the `StickyGrid` (which uses the info to decide which header cells to render) through a separate context from the grid settings context to avoid re-rendering the whole `Grid` on scroll.

I've also replaced the "anchor" button in the top left-hand cell with a proper toolbar button:

image


packages/lib/src/vis/matrix/MatrixVis.tsx
```diff
@@ -1,60 +1,41 @@
...
-            rowHeight={CELL_HEIGHT}
-            columnCount={columnCount}
-            rowCount={rowCount}
-            width={wrapperSize.width}
-            height={wrapperSize.height}
-          >
-            {Cell}
-          </IndexedGrid>
-        )}
-      </div>
```
axelboc, line 56: Moved to `Grid`.

packages/lib/src/vis/matrix/MatrixVis.tsx
```diff
@@ -1,60 +1,41 @@
...
-      }
+      cellSize={{ width: cellWidth, height: CELL_HEIGHT }}
+      cellFormatter={cellFormatter}
+      sticky={sticky}
     >
-      <div ref={wrapperRef} className={styles.wrapper}>
-        {wrapperSize && (
-          <IndexedGrid
-            className={styles.grid}
-            innerElementType={forwardRef(StickyGrid)}
```
axelboc, line 45: This `forwardRef` invocation led to a new `innerElementType` component being passed to `IndexedGrid` on every render of `MatrixVis`, and thus to `StickyGrid` being umounted/re-mounted each time. This was problematic with some of my previous attempts ... not too much with the current solution, but still, not idea. `forwardRef` is now called statically in `StickyGrid`.

packages/lib/src/vis/matrix/StickyGrid.tsx
```diff
@@ -1,39 +1,58 @@
...
+    useContext(SettingsContext);
 
-  const [sticky, setSticky] = useState(true);
+  const renderedItems = useContext(RenderedItemsContext);
+  const {
+    overscanColumnStartIndex: colStart = 0,
+    overscanColumnStopIndex: colStop = 0,
+    overscanRowStartIndex: rowStart = 0,
+    overscanRowStopIndex: rowStop = 0,
+  } = renderedItems || {};
```
axelboc, line 24: This is where I make use of `renderedItems`. Note that I'm using the `overscan` indices instead of the `visible` inidices which are also available. This allows the same overscan UX improvement to apply to header cells.

packages/lib/src/vis/matrix/StickyGrid.tsx
```diff
@@ -1,39 +1,58 @@
...
     >
-      <IndexTrack className={styles.indexRow} cellCount={columnCount - 1}>
-        <AnchorCell sticky={sticky} onToggle={() => setSticky(!sticky)} />
-      </IndexTrack>
+      <div className={styles.colHeaders}>
+        <div className={styles.topLeftCell} style={cellSize} data-bg />
+        <HeaderCells
+          indexMin={colStart}
+          indexMax={colStop}
+          transform={`translateX(${cellSize.width * colStart}px)`}
```
axelboc, line 41: This is the CSS `transform` I mentioned earlier. Since we're not rendering header cells from `0` to `colStart`, we have to shift the ones that are rendered by the amount of space the unrendered cells would have taken. I've left a component explaining this against the `transform` prop in `HeaderCells`.

packages/lib/src/vis/matrix/context.tsx
```diff
@@ -0,0 +1,53 @@
...
+export const RenderedItemsContext = createContext<
+  GridOnItemsRenderedProps | undefined
+>(undefined);
+
+function GridProvider(props: Props) {
+  const { rowCount, columnCount, cellSize, sticky, cellFormatter, children } =
+    props;
+
+  const [renderedItems, setRenderedItems] =
+    useState<GridOnItemsRenderedProps>();
```
axelboc, line 30: The `renderedItems` state, which I provide through its own, independent context as explained by Kent C. Dodds: https://kentcdodds.com/blog/how-to-optimize-your-context-value

loichuder (approved):
:100:

PR #890, opened 2021-12-03
Add controls to export dataset or slice to NPY and CSV

### LINKED ISSUE #35: Copying/exporting datasets

axelboc:
Fix #35

- `MatrixVis` : CSV ✅, NPY ✅
- `LineVis` : CSV ✅, NPY ✅
- `HeatmapVis` : NPY ✅ TIFF ✅ (already done in #885)

The export links/buttons sit inside a drop-down menu (similar to a `Selector` but without any selection management). When the `selection` is defined, the menu reads "Export slice"; otherwise, just "Export".

The API method `getTiffUrl` has been generalized to handle other formats. It is now called `getExportURL`. Providers may implement this method but return `undefined` for any unsupported formats. The `ExportMenu` will display download links only for supported formats.

For NeXus visualizations, contrary to what I had initially written in https://github.com/silx-kit/h5web/issues/35#issuecomment-985365726 (now edited), it was simpler to implement the export of the signal dataset rather than to not show the export menu at all (because NeXus visualization re-use the core visualizations' toolbars).

I can confirm that both NPY and CSV work with H5Grove on Bosquet already 🎉 

image

image

image

### DIFF
--- a/packages/app/src/vis-packs/core/line/LineVisContainer.tsx
+++ b/packages/app/src/vis-packs/core/line/LineVisContainer.tsx
@@ -23,6 +23,7 @@ function LineVisContainer(props: VisContainerProps) {
   const [dimMapping, setDimMapping] = useDimMappingState(dims, 1);
 
   const autoScale = useLineConfig((state) => state.autoScale);
+  const selection = autoScale ? getSliceSelection(dimMapping) : undefined;
 
   return (
     <>
@@ -39,14 +40,15 @@ function LineVisContainer(props: VisContainerProps) {
       >
         <ValueFetcher
           dataset={entity}
-          selection={autoScale ? getSliceSelection(dimMapping) : undefined}
+          selection={selection}
           render={(value) => (
             <MappedLineVis
+              dataset={entity}
+              selection={selection}
               value={value}
               dims={dims}
               dimMapping={dimMapping}
               title={entity.name}
-              dtype={entity.type}
               toolbarContainer={toolbarContainer}
             />
           )}
--- a/packages/lib/src/toolbar/controls/ExportMenu.tsx
+++ b/packages/lib/src/toolbar/controls/ExportMenu.tsx
@@ -1,0 +1,60 @@
+import { isDefined } from '@h5web/shared';
+import { Button, Wrapper, Menu } from 'react-aria-menubutton';
+import { FiDownload } from 'react-icons/fi';
+import { MdArrowDropDown } from 'react-icons/md';
+
+import styles from './Selector/Selector.module.css';
+
+interface Props<F extends string> {
+  formats: F[];
+  isSlice: boolean;
+  getFormatURL: (format: F) => string | undefined; // `undefined` if format is not supported
+}
+
+function ExportMenu<F extends string>(props: Props<F>) {
+  const { formats, isSlice, getFormatURL } = props;
+
+  const urls = formats.map(getFormatURL);
+  const hasSupportedFormats = urls.some(isDefined);
+
+  return (
+    <Wrapper className={styles.wrapper}>
+      <Button
+        className={styles.btn}
+        tag="button"
+        disabled={!hasSupportedFormats}
+      >
+        <div className={styles.btnLike}>
+          <FiDownload className={styles.icon} />
+          <span className={styles.selectedOption}>
+            Export{isSlice && ' slice'}
+          </span>
+          <MdArrowDropDown className={styles.arrowIcon} />
+        </div>
+      </Button>
+      <Menu className={styles.menu}>
+        <div className={styles.list}>
+          {formats.map(
+            (format, index) =>
+              urls[index] && (
+                <a
+                  key={format}
+                  className={styles.linkOption}
+                  href={urls[index]}
+                  target="_blank"
+                  download={`data.${format}`}
+                  rel="noreferrer"
+                >
+                  <span
+                    className={styles.label}
+                  >{`Export to ${format.toUpperCase()}`}</span>
+                </a>
+              )
+          )}
+        </div>
+      </Menu>
+    </Wrapper>
+  );
+}
+
+export default ExportMenu;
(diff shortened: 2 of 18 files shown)


packages/lib/src/toolbar/controls/Selector/Selector.module.css
```diff
@@ -79,18 +87,15 @@
...
   background-color: var(--h5w-selector-option-selected--bgColor, #eee);
 }
-
-.gradient {
-  position: relative;
-  top: 1px;
-  width: 2rem;
-  height: 1.25em;
-  margin-left: 0.625rem;
-}
```
axelboc, line 96: dead code, which I'm sure I had removed before 🤷

packages/lib/src/toolbar/controls/ExportMenu.tsx
```diff
@@ -0,0 +1,59 @@
...
+interface Props<F extends string> {
+  formats: F[];
+  isSlice: boolean;
+  getFormatURL: (format: F) => string | undefined; // `undefined` if format is not supported
+}
+
+function ExportMenu<F extends string>(props: Props<F>) {
+  const { formats, isSlice, getFormatURL } = props;
+
+  if (!getFormatURL) {
```
loichuder, line 16: It would be nice to check also if `formats.map(getFormatURL)` gives defined values.

Otherwise, we can have an export menu with no entries.

packages/app/src/vis-packs/core/line/LineVisContainer.tsx
```diff
@@ -39,14 +40,15 @@ function LineVisContainer(props: VisContainerProps) {
       >
         <ValueFetcher
           dataset={entity}
-          selection={autoScale ? getSliceSelection(dimMapping) : undefined}
+          selection={selection}
           render={(value) => (
             <MappedLineVis
+              dataset={entity}
+              selection={selection}
```
loichuder, line 47: Given how often `dataset` and `selection` come together, it might be tempting to merge them in a single object

loichuder (approved):

loichuder:
Also, I wonder if people would not expect to export the **dataset** rather than the **slice**. But this is fine for a first go.

packages/lib/src/toolbar/controls/ExportMenu.tsx
axelboc: Yeah... That's what I had initially, but I found the code a little too complex compared to the likelihood of this happening. Perhaps we can revisit it if it ever happens?

packages/app/src/vis-packs/core/line/LineVisContainer.tsx
axelboc: Interesting thought, but do you mind if we keep thinking about this one for a bit? I feel like there might be a broader refactor to be done after all of the recent toolbar work.

packages/lib/src/toolbar/controls/ExportMenu.tsx
loichuder: Well, now that we are knee-deep in this feature, we might as well implement the check and be done with it, no :smile: ?

axelboc:
> Also, I wonder if people would not expect to export the dataset rather than the slice. But this is fine for a first go.

Yeah, fair, but we come back to the same concern as with the auto-scale. If people want to download entire datasets, we don't want to encourage them to do so from the viewer, or we risk overloading the server. I see this more as a convenience feature for people to be able to extract the raw data that is _actually_ displayed by the viewer.

loichuder:
> > Also, I wonder if people would not expect to export the dataset rather than the slice. But this is fine for a first go.
> 
> Yeah, fair, but we come back to the same concern as with the auto-scale. If people want to download entire datasets, we don't want to encourage them to do so from the viewer, or we risk overloading the server. I see this more as a convenience feature for people to be able to extract the raw data that is _actually_ displayed by the viewer.

Good analysis :+1:

packages/lib/src/toolbar/controls/ExportMenu.tsx
axelboc: Haha fair, okay I'll put it back.

axelboc: Found a simpler way than what I had initially tried, so all good. Let me know what you think.

packages/lib/src/toolbar/controls/ExportMenu.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const urls = getFormatURL ? formats.map(getFormatURL) : [];
+  const hasSupportedFormats = urls.some(isDefined);
+
+  return (
+    <Wrapper className={styles.wrapper}>
+      <Button
+        className={styles.btn}
+        tag="button"
+        disabled={!hasSupportedFormats}
```
axelboc, line 25: I now disable the menu if the provider doesn't implement `getExportURL` or doesn't support any of the given formats. This avoids ending up with a `Separator` that has nothing to separate.

packages/lib/src/toolbar/controls/ExportMenu.tsx
```diff
@@ -0,0 +1,60 @@
...
+interface Props<F extends string> {
+  formats: F[];
+  isSlice: boolean;
+  getFormatURL: (format: F) => string | undefined; // `undefined` if format is not supported
+}
+
+function ExportMenu<F extends string>(props: Props<F>) {
+  const { formats, isSlice, getFormatURL } = props;
+
+  const urls = getFormatURL ? formats.map(getFormatURL) : [];
```
loichuder, line 17: Given that `getFormatURL` is always defined, you can get rid of the ternary

loichuder (approved):

packages/lib/src/toolbar/controls/ExportMenu.tsx
axelboc: Well spotted

PR #896, opened 2021-12-06
Show progress while fetching dataset values

axelboc:
Proof of concept for now. I'd like your view on the implementation before I add a proper progress bar.

image

At the moment, I keep the progress state in `H5GroveProvider` so I can pass an `onValueProgress` callback when the API is instantiated. If I were to keep the progress state in `Provider`, I would have to add a `setOnValueProgress` or `setValueProgressCb method to the API to set the callback post-instantiation, which wouldn't be very elegant.

Since I only need the callback when actually fetching a value, I could add an argument to the API's `getValue` method. The downside is that I then need to propagate it to the `fetchValue`, `fetchBinaryValue` and `cancellableFetchValue` methods, which complexifies the signatures (especially for the latter method)...


axelboc:
Superseded by #908

PR #899, opened 2021-12-09
Keep viewport center ... centered on resize

axelboc:
keep-centre-on-resize

Next up, I'll do some optimising in `VisCanvas`, as we're currently over-rendering on resize because of `useMeasure`. I'll also tweak the `debounce` settings of R3F, as it is currently inconsistent (no debounce at all when resizing the window vs. a significant debounce when resizing the explorer panel).

### DIFF
--- a/packages/lib/src/vis/shared/PanZoomMesh.tsx
+++ b/packages/lib/src/vis/shared/PanZoomMesh.tsx
@@ -3,6 +3,7 @@ import type { ThreeEvent } from '@react-three/fiber/dist/declarations/src/core/e
 import { clamp } from 'lodash';
 import { useRef, useCallback, useEffect } from 'react';
 import type { Vector3 } from 'three';
+import { Vector2 } from 'three';
 
 import { useWheelCapture } from '../hooks';
 import { useAxisSystemContext } from './AxisSystemContext';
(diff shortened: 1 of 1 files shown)


packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -3,27 +3,38 @@ import type { ThreeEvent } from '@react-three/fiber/dist/declarations/src/core/e
...
   const invalidate = useThree((state) => state.invalidate);
 
   const startOffsetPosition = useRef<Vector3>(); // `useRef` to avoid re-renders
+  const viewportCenter = useRef<Vector2>();
 
   const moveCameraTo = useCallback(
-    (x: number, y: number) => {
-      const { position, zoom } = camera;
+    (x: number, y: number, saveCoords = false) => {
+      if (saveCoords) {
```
loichuder, line 26: Is `saveCoords` really necessary ? I mean, it does not hurt to save the viewport center once again when restoring it ?

We could also move the `saveCoords` part out of `moveCameraTo` to be put in a dedicated callback ?

loichuder (approved):
:clap:

packages/lib/src/vis/shared/PanZoomMesh.tsx
axelboc: > Is saveCoords really necessary ? I mean, it does not hurt to save the viewport center once again when restoring it ?

I guess not, good point. It's not triggering any re-renders since it's in a ref, so the impact is minimal.

PR #900, opened 2021-12-09
Flatten arrays in backend in h5grove

loichuder:
Since [h5grove@0.0.12](https://github.com/silx-kit/h5grove/releases/tag/v0.0.12), it is possible to request a flattened array by setting the query parameter `flatten` to `true`.

The `flatten` has no effect for scalar datasets.


axelboc (approved):
Nice!

PR #904, opened 2021-12-13
Measure canvas area only when a specific canvas ratio is requested

axelboc:
The goal is to avoid unnecessary re-renders of the heatmap vis' canvas when the layout is `fill` or `cover` (which is always the case in @h5web/app).

To make this behaviour a bit more readable, I actually reworked a few things:

- I merged `computeVisSize` and `useVisSize` into a single `getSizeToFit`, which takes an `availableSize` and a `ratioToRespect`.
- I moved the axis offsets padding to the measured container (now called `canvasArea` instead of `visArea`), removed the early return, etc.

As mentioned in a comment, one issue is that when the layout changes to/from `contain` dynamically (i.e. when the `canvasRatio` prop of `VisCanvas` changes between a number and `undefined`), the measuring either doesn't start (and nothing gets rendered), or doesn't stop (everything keeps re-rendering on resize).

### DIFF
--- a/packages/lib/src/vis/shared/VisCanvas.tsx
+++ b/packages/lib/src/vis/shared/VisCanvas.tsx
@@ -1,25 +1,34 @@
 import { useMeasure } from '@react-hookz/web';
 import { Canvas } from '@react-three/fiber';
+import type { PropsWithChildren } from 'react';
 
-import { computeCanvasSize, getAxisOffsets } from '../utils';
+import type { AxisConfig } from '../models';
+import { getSizeToFit, getAxisOffsets } from '../utils';
 import AxisSystem from './AxisSystem';
 import AxisSystemProvider from './AxisSystemProvider';
-import type { AxisSystemProviderProps } from './AxisSystemProvider';
 import styles from './VisCanvas.module.css';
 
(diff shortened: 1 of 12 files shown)


packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -1,77 +1,84 @@
...
-    <div ref={visAreaRef} className={styles.visArea}>
-      <div
-        className={styles.vis}
-        style={{
-          ...visSize,
-          paddingBottom: axisOffsets.bottom,
-          paddingLeft: axisOffsets.left,
-          paddingTop: axisOffsets.top,
-          paddingRight: axisOffsets.right,
-          boxSizing: 'content-box',
```
axelboc, line 47: The padding is now on the parent, which means I no longer need to use `boxSizing`.

axelboc:
So now, the only re-renders that happen when resizing are those triggered by R3F. Since they are debounced thanks to `resize={{ debounce: { scroll: 20, resize: 200 }, scroll: false }}`, we now get much better performance when resizing.

loichuder (approved):

PR #907, opened 2021-12-14
Add button to breadcrumbs bar to leave feedback

axelboc:
I've moved things around a little, hope you like it:

image

I've tried a few other things but wasn't convinced:

1. image

2. image

3. image


packages/lib/src/index.ts
```diff
@@ -12,7 +12,7 @@ export type { HeatmapVisProps } from './vis/heatmap/HeatmapVis';
 export { default as Toolbar } from './toolbar/Toolbar';
 export { default as Separator } from './toolbar/Separator';
 export { default as Btn } from './toolbar/controls/Btn';
-export { default as DownloadBtn } from './toolbar/controls/DownloadBtn';
+export { default as LinkBtn } from './toolbar/controls/LinkBtn';
```
axelboc, line 15: I've made `DownloadBtn` more generic to just be a link button (i.e. with an `<a>` tag). `SnapshotBtn` now uses `LinkBtn`.

packages/lib/src/toolbar/controls/Btn.tsx
```diff
@@ -28,11 +28,12 @@ function Btn(props: Props) {
     <button
       className={styles.btn}
       type="button"
-      onClick={() => onClick()}
+      title={iconOnly ? label : undefined}
```
axelboc, line 31: now showing a tooltip on hover when `iconOnly` is `true`.

axelboc:
image

packages/app/src/breadcrumbs/utils.ts
```diff
@@ -0,0 +1,21 @@
+import { version } from '../../package.json';
+
+export function prepareFeedback(filepath: string, entityPath: string): string {
+  return `Hi,
+
+<< Please provide your feedback here. >>
+<< To report an issue, please include screenshots, reproduction steps, etc. >>
+<< To suggest a new feature, please describe the needs this feature would fulfill. >>
```
loichuder, line 8: I would add a line saying `You can also open an issue on https://github.com/silx-kit/h5web/issues` to make use of the issue template.

Speaking of which, I would also be in favour of adding a feature request template like for jupyterlab-h5web: https://github.com/silx-kit/jupyterlab-h5web/blob/main/.github/ISSUE_TEMPLATE/feature_request.md

loichuder (commented):
Not a fan of the new layout. 1. seems more visually appealing to me.

So, the idea would be to add this button for all projects using `@h5web/app` ?

axelboc:
> Not a fan of the new layout. 1. seems more visually appealing to me.

- What I don't like about layout 1 (below for reference). is that the breadcrumbs are no longer centered within the right-hand side of the UI. They weren't quite centered before, but the Inspect/Visualize control was small enough and the whitespace around the breadcrumbs was big enough that this wasn't noticeable. The feedback button now pushes the breadcrumbs to the left too much in my opinion ... and it pushes them too much to be able to compensate with padding on the left (the whitespace around the breadcrumbs would then feel too unbalanced - i.e. too much whitespace on the left and not enough on the right).
  
image

- What I like about the layout I'm proposing (below for reference) is that the alignment of the breadcrumbs is more what we expect it to be on the web (i.e. left aligned); it also equates to less shifting (i.e. the parent crumbs don't move when selecting a child entity). What I also like is that the Inspect/Visualize control is more easily reachable -- we typically want to switch to inspect before/after interacting with the explorer, and at the moment, these two UI elements are on opposite sides of the screen.

image

> So, the idea would be to add this button for all projects using `@h5web/app` ?

With the current implementation, yes. Unless that's not what we want? Do you see a downside?

packages/app/src/breadcrumbs/utils.ts
axelboc: I was going to add the line, but I'm having second thoughts. I'm wondering whether this wouldn't just confuse users. They're about to write an email and then we tell them "oh, actually, maybe just create an issue on GitHub" -- they may end up trying to create an issue, look at the issue template and give up; they may have trouble remembering their GitHub credentials are giving up at the 2FA step; worse, they may open the issue but forget to paste the contextual information pre-filled in the email body!

FYI, I've just discovered that it's possible to generate GitHub URLs to create issues with pre-filled content: https://github.com/silx-kit/h5web/issues/new?title=Foo&body=Bar. You can even select a template: https://github.com/silx-kit/h5web/issues/new?title=Foo&template=bug_report.md. Unfortunately, you can't do both - i.e. pre-fill a template with contextual information; otherwise I would have been happy to add a link with such a URL in the feedback email.

In conclusion, I think the pre-filled email with no mention of GitHub is the least confusing solution that maximises our chances of receiving feedback.

loichuder: > In conclusion, I think the pre-filled email with no mention of GitHub is the least confusing solution that maximises our chances of receiving feedback.

Sounds good

loichuder (approved):
I understand the reasons behind the new layout so we can merge this

loichuder:
> Unless that's not what we want? Do you see a downside?

I don't have much experience in this so sorry if my question is dumb but won't exposing such a button on https://h5web.panosc.eu/ attract a lot of spam ?

axelboc:
Ah true, true. Also, it seems `@h5web/app` is now being used outside the ESRF: #914. 🙌

t20100:
Only downside I can think of with the new layout, is that it adds a bit of visual clutter next to the breadcrumbs.
+1 for left aligned breadcrumbs.

t20100 (approved):

t20100:
BTW, if the "Give feedback" button was smaller, I would have voted for 3. since I'm not sure switching between `Inspect` and `Display` is that regular.

packages/app/src/breadcrumbs/CopyableCrumb.tsx
```diff
@@ -23,7 +23,7 @@ function CopyableCrumb(props: Props) {
         void navigator.clipboard.writeText(path);
         setPathCopied(true);
       }}
-      onPointerOut={() => setPathCopied(false)}
+      onPointerLeave={() => setPathCopied(false)}
```
axelboc, line 26: Bugfix: after copying the path, the tick icon would revert back to the copy icon when moving the mouse from the label to the icon.

axelboc:
As suggested, trying with layout 3 and a shorter label for the feedback button:

image

packages/app/src/breadcrumbs/utils.ts
```diff
@@ -0,0 +1,28 @@
+import type { FeedbackContext } from './models';
+
+export function getFeedbackMailto(
```
loichuder, line 3: Not sure if this is worth documenting/exporting.

I would inline this in the `getFeedbackURL` of `demo` and be done with it

axelboc: Hmm but then you'll have to re-implement it in `jupyter-h5web` 😄

loichuder: Didn't think about this :clown_face:

wlievens:
@axelboc If you want some more details on our use case:

I work for a company that designs image sensors. We use the HDF5 format for in image sensor validation campaigns, and also plan to use it for production test data. A web-based view of HDF5 data would allow users to quickly inspect issues with sensors visually as they get flagged as defective by the production test software. I adapted our (work-in-progress) backend application to comply with h5grove endpoints, and now plan to use h5web in the (Angular) front-end.

axelboc:
Haha no that's great, thanks a lot for sharing! How did you find the implementation of the H5Grove endpoints? Did you make use of the [h5grove](https://github.com/silx-kit/h5grove/) library itself? Don't hesitate to open an issue over there if you have any suggestions/questions. 😉

wlievens:
The backend uses FastAPI for the REST endpoints, so I made a few GET endpoints there that then call the appropriate functions in h5grove. It wasn't too hard to figure out based on the flaskutils example code.

I already had a (simpler, more limited) HDF5 extraction endpoint for use in our application, so I refactored that client code to re-use the h5grove-compliant hdf5 endpoints. It all works really smooth, except for #914 that is ;-)

PR #908, opened 2021-12-14
Show progress while fetching dataset values

### LINKED ISSUE #892: Progress bar
Like on Braggy, we could show a progress bar on the `ValueLoader`, for dataset values requests.

axelboc:
Fix #892 

New version with listeners in the API.

Still the problem that the back-end doesn't send the total size of the data for some reason (probably missing `Content-Length` HTTP header).

### DIFF
--- a/packages/app/src/providers/api.ts
+++ b/packages/app/src/providers/api.ts
@@ -1,3 +1,4 @@
+/* eslint-disable promise/prefer-await-to-callbacks */
 import type {
   ArrayShape,
   AttributeValues,
@@ -14,7 +15,11 @@ import type {
 } from 'axios';
 import axios from 'axios';
 
-import type { ExportFormat, ValuesStoreParams } from './models';
+import type {
+  ExportFormat,
+  ProgressCallback,
+  ValuesStoreParams,
+} from './models';
 import { CANCELLED_ERROR_MSG } from './utils';
 
 interface ValueRequest {
@@ -28,6 +33,9 @@ export abstract class ProviderApi {
 
   protected readonly client: AxiosInstance;
   protected readonly valueRequests = new Set<ValueRequest>();
+  protected progress = new Map<ValuesStoreParams, number>();
+
+  private readonly progressListeners = new Set<ProgressCallback>();
 
   public constructor(filepath: string, config?: AxiosRequestConfig) {
     this.filepath = filepath;
@@ -39,6 +47,15 @@ export abstract class ProviderApi {
     selection: string | undefined,
     format: ExportFormat
   ): string | undefined; // `undefined` if format is not supported
+
+  public addProgressListener(cb: ProgressCallback): void {
+    this.progressListeners.add(cb);
+    cb([...this.progress.values()]); // notify once
+  }
+
+  public removeProgressListener(cb: ProgressCallback): void {
+    this.progressListeners.delete(cb);
+  }
 
   public cancelValueRequests(): void {
     // Cancel every active value request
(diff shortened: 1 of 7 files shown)


loichuder (dismissed):
:partying_face: 

I will have a look at the back-end

packages/app/src/vis-packs/ValueLoader.tsx
```diff
@@ -32,7 +40,9 @@ function ValueLoader(props: Props) {
         <div />
         <div />
       </div>
-      <p>{message}...</p>
+      <p>
+        {message}... ({progress} %)
```
loichuder, line 44: I tried by adding the `Content-Length` headers in h5grove and it works !

I would simply suggest to round `progress` as now, it displays a number with many decimals.

Also, the `ValueLoader` displays `0%` just before disappearing which is not great. Perhaps due to the `removeProgressListener` ?

axelboc: Great! Okay, I've rounded the number and I've made sure the `progress` is reset when a new request starts rather than when a request fulfills.

axelboc:
I'll wait until you've merged/deployed the back-end on bosquet. I'd like to double check what happens when multiple value requests occur in parallel (i.e. with NeXus).

axelboc:
Alright, I've pushed something I'm starting to feel okay with. I now display progress bars for ongoing requests:

image

If multiple requests are happening in parallel (e.g. NeXus), the `ValueLoader` shows as many progress bars, stacked on top of one another. The bars disappear when the corresponding requests fulfill.

I'm still not 100% happy with the code, but at least having multiple progress bars makes sense conceptually and means we can keep everything decoupled (i.e. the `ValueLoader` doesn't need to know which requests are in progress; we don't need to initialize a `progress` state in every container; we don't need to pass a callback from every call to `useDatasetValue(s)`; etc.)

One problem I see is that since we don't cancel requests when slicing, the progress bars add up:

image

I think it's worth cancelling requests like we do when switching vis/entity -- if we don't want to do that, then we'll be back in a pickle :smile: 

Anyways... to be continued.

axelboc:
... or I can just show the five "oldest" progress bars. Once the fetch of a previous slice succeeds, the progress bar disappears and leaves room for a more recent progress bar. Sounds reasonable to you @loichuder?

loichuder:
That sounds sensible. I will wait to test it out to give my final opinion :smile:

axelboc:
Okay, I've limited it to 3 simultaneous progress bars. Enough to show that multiple things are loaded, yet not too many so as for the UI to not feel as cluttered as in the previous screenshot. It works for NeXus groups with multiple children datasets to fetch, as well as when slicing. We may decide to cancel requests while fetching, as previously discussed (and as suggested in #920), but with the new 3-bars limit, this no longer impacts this PR.

packages/app/src/providers/api.ts
```diff
@@ -55,26 +72,41 @@ export abstract class ProviderApi {
   protected async cancellableFetchValue<T>(
     endpoint: string,
     storeParams: ValuesStoreParams,
-    queryParams?: Record<string, string | boolean | undefined>,
+    queryParams: Record<string, string | boolean | undefined>,
     responseType?: ResponseType
```
axelboc, line 76: Note a minor change here: `queryParams` defaulting to `storeParams` was prone to bugs in my opinion. We no longer relied on it anyway.

packages/app/src/providers/api.ts
```diff
@@ -55,26 +72,41 @@ export abstract class ProviderApi {
...
-      // Remove cancellation source when request fulfills
+      // Remove cancellation source and progress when request fulfills
       this.valueRequests.delete(request);
+      this.progress.delete(storeParams);
+      this.notifyProgressChange();
     }
   }
 
+  private notifyProgressChange() {
+    this.progressListeners.forEach((cb) => cb([...this.progress.values()]));
```
loichuder, line 107: Just a thought, if `progressListeners` was a `Map` instead of a `Set`, you could notify only the relevant change, right ?

You would need to pass the `storeParams` as arg of most callbacks though.

packages/app/src/vis-packs/ValueLoader.module.css
```diff
@@ -6,8 +6,34 @@
   flex-direction: column;
 }
 
+.progressBars {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  max-width: 20rem;
+  margin-bottom: 0.75rem;
```
loichuder, line 14: I would change the CSS (add a `min-height` ?) so that the `Cancel` button is not moved around when progress bars disappear. 

It might be frustrating for the user to try to click on the cancel button if the layout changes dynamically...

loichuder (approved):
Nice :+1:

packages/app/src/providers/api.ts
axelboc: You mean so listeners only get notified of the request that has "progressed"? Not sure there's a need for this. The value loader loops through all the progress values on every render anyway. We don't have a component that would benefit from re-rendering only when a specific request's progress changes.

packages/app/src/vis-packs/ValueLoader.module.css
axelboc: Good call. I've added a `min-height` as you suggested, as well as `justify-content: center` so the whitespace is shared equally above/below the bar(s) when there's only one or two.

Issue #769, opened 2021-07-26
[Zoom] Give more zooming options


loichuder:
### Example use case 
With a plot like this:
![image](https://user-images.githubusercontent.com/42204205/126961158-c88b4abf-0327-4709-9339-7699e4303c23.png)

I would like to be able to zoom on the y-axis while keeping the full x-axis. The current zoom only allows me to zoom on both axes at once.

### Implementation
We could imagine 
- Draw a zooming box with drag and drop ? This could interfere with the panning though.
- Zoom only on one axis with Alt/Ctrl+wheel ?
- Being able to zoom only on one axis by hovering on the desired axis and use the wheel

axelboc:
- Drawing a rectangle to zoom horizontally doesn't make much sense to me. However, I do like UIs that allow selecting a horizontal area, like in the dev tools. However, in such UIs, the selection is usually displayed somewhere else on the UI -- i.e. it's not a zoom per se, more like an inspection.

    ![image](https://user-images.githubusercontent.com/2936402/128701278-1a9eaeff-0c99-46a8-b303-af5415cad06c.png)

    Assuming a zoom-as-you-select UI, it would indeed interfere with panning. We'd have to either bind the action to another mouse button (not discoverable) or add a button to the UI to switch between pan and select-to-zoom (tedious to use, one more button in the toolbar, and challenging labeling/iconography to convey that the button activates horizontal zooming as this pattern is more commonly used for x/y zooming).

- Ctrl+Wheel is tempting, but the problem is discoverability. We could add a help dialog somewhere to document all the interactions, but it's not great...

- Zooming by hovering on the desired axis is promising, as it is a lot more discoverable and doesn't require any additional UI. On the technical side, it may be difficult to make the axis system mesh communicate with the pan/zoom mesh.

- The simplest (technically) and clearest (for users), would be to add a button in the toolbar to change the behaviour of the wheel-zoom to horizontal-only. Compared to the select-to-zoom feature, it has the advantage of not drastically changing the interaction (mouse wheel to zoom) and not interfering with the pan. However, adding a button for this in the toolbar gives the feature a bit too much prominence for my taste. So far we only have buttons that control the display of the plot -- we don't have buttons that control the interactions with the plot. Perhaps interactions controls could be located somewhere else, like on the right of the plot title?

    ![image](https://user-images.githubusercontent.com/2936402/128704621-050e1d0d-ba6a-434a-81d7-c562c4c29e49.png)

---

[...]

PR #909, opened 2021-12-14
Make zoom work with camera FOV

loichuder:
Our implementation of the zoom on the vis was using the Three.js camera zoom. 

While this is very handy, this prevents from scaling differently in X and in Y making issues like #769 impossible to resolve.

Zooming **with a orthographic camera** consists simply in reducing the field of the view of the camera (see https://threejs.org/manual/examples/cameras-orthographic-2-scenes.html). The field of the view is constrained by the properties `left`, `right`, `top` and `bottom` of the camera.

So, instead of using `camera.zoom`, I change the field of view "manually" by changing the aforementioned parameters. This will allow to change the zoom in X (`left`/`right`) independently from the zoom in Y (`top`/`bottom`) when needed.

### DIFF
--- a/packages/lib/src/vis/hooks.ts
+++ b/packages/lib/src/vis/hooks.ts
@@ -5,6 +5,7 @@ import type { NdArray } from 'ndarray';
 import { useCallback, useMemo, useState } from 'react';
 import type { RefCallback } from 'react';
 import { createMemo } from 'react-use';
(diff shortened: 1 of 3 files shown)


loichuder:
Converting this to draft as I have an issue on resize: the camera FOV gets reset.

loichuder:
Superseded by #932

PR #911, opened 2021-12-15
Add TS checks to linting

loichuder:


apps/demo-cra-lib/package.json
```diff
@@ -16,7 +16,9 @@
   "scripts": {
     "start": "react-app-rewired start",
     "build": "react-app-rewired build",
-    "lint": "eslint \"**/*.{js,cjs,ts,tsx}\" --max-warnings=0"
+    "lint": "run-p lint:eslint lint:tsc",
```
axelboc, line 19: I wouldn't bother with these `lint` scripts in all the projects (especially since it requires installing `npm-run-all`. The important bit is that we have one command at the root that runs everything, so I'd instead rename `lint:all` to `lint` in the root `package.json`. (It's great that we're freeing up the bare name `lint` in the root like this!)

axelboc (approved):
Good call :100:

Note that the contributing guide may need updating.

loichuder:
With https://github.com/DefinitelyTyped/DefinitelyTyped/pull/57746, we are almost there !

I tried to address https://github.com/silx-kit/h5web/pull/911#discussion_r770345331 by having only `npm-run-all` at the project root. I also updated the relevant parts in CONTRIBUTING.

I still have an issue with the storybook: TypeScript reports that it cannot find CSS modules files from `@h5web/lib` ? :shrug: 

Could you have a look ?

axelboc:
I finally managed to fix the Storybook errors as discussed. I'm still not a huge fan of the fact that workspace dependencies are type-checked multiple times and that the `react-scripts` types have to be referenced again from the `storybook` project, but the result is worth it.

Note my other two commits, notably the one that restores the lint script for the workspace root (i.e. the JS config files at the root and the `cypress` folder). I had to create an extra `tsconfig.lint.json` file and run `tsc` twice to avoid a problem with missing Cypress types.

loichuder (commented):
Yaay, thanks a lot :smile_cat: ! Good to ship for me :ship:

axelboc (approved):

PR #912, opened 2021-12-17
Project points with camera rather than manually

loichuder:
While working on #909, I realized that we made a lot of similar computations when going from canvas coordinates to HTML coordinates. These computations were "manually" using the camera parameters (zoom and position).

I refactored these computations to use the camera projection matrix instead, that already contain all the camera parameters. 

This has several advantages:
- The computations are more generic, taking in account *both* the camera field of view and the zoom. This will greatly simplify #909 and allow a seamless transition.
- The computations are more concise and easier to understand in my opinion.

I see one disadvantage tho, we have to deal with (_shudder_) mutable `Vector3`...

Let me know what you think !

_PS: This image from [this SO thread](https://stackoverflow.com/questions/46829113/transpose-z-position-from-perspective-to-orthographic-camera-in-three-js) was very useful:_

image

_Note, however, that the X and Z are switched._

### DIFF
--- a/apps/storybook/src/HeatmapVisDisplay.stories.tsx
+++ b/apps/storybook/src/HeatmapVisDisplay.stories.tsx
@@ -109,6 +109,28 @@ WithAnnotation.args = {
   domain,
 };
 
+export const WithAnnotationZoom: Story<HeatmapVisProps> = (args) => (
+  <HeatmapVis {...args}>
+    <Annotation
+      x={30}
+      y={18}
+      style={{
+        width: '100px',
+        color: 'white',
+        textAlign: 'center',
+        fontSize: '0.75rem',
+      }}
+      scaleOnZoom
+    >
+      An annotation scaling with zoom
+    </Annotation>
+  </HeatmapVis>
+);
+WithAnnotationZoom.args = {
+  dataArray,
+  domain,
+};
+
 export default {
   ...HeatmapVisStoriesConfig,
   title: 'Visualizations/HeatmapVis/Display',
--- a/packages/lib/src/vis/shared/AxisSystem.tsx
+++ b/packages/lib/src/vis/shared/AxisSystem.tsx
@@ -1,5 +1,6 @@
 import type { Domain } from '@h5web/shared';
 import { useThree } from '@react-three/fiber';
+import { Vector3 } from 'three';
 
 import { useFrameRendering } from '../hooks';
 import type { AxisOffsets } from '../models';
@@ -13,25 +14,30 @@ interface Props {
   title?: string;
 }
 
+const CAMERA_BOTTOM_LEFT = new Vector3(-1, -1, 0);
+const CAMERA_TOP_RIGHT = new Vector3(1, 1, 0);
+
 function AxisSystem(props: Props) {
   const { axisOffsets, title } = props;
 
   const { abscissaConfig, ordinateConfig, abscissaScale, ordinateScale } =
     useAxisSystemContext();
 
-  const { position, zoom } = useThree((state) => state.camera);
+  const camera = useThree((state) => state.camera);
   const canvasSize = useThree((state) => state.size);
   const { width, height } = canvasSize;
 
-  // Find visible domains from camera's zoom and position
+  const worldBottomLeft = CAMERA_BOTTOM_LEFT.clone().unproject(camera);
+  const worldTopRight = CAMERA_TOP_RIGHT.clone().unproject(camera);
+
   const xVisibleDomain: Domain = [
-    abscissaScale.invert(-width / (2 * zoom) + position.x),
-    abscissaScale.invert(width / (2 * zoom) + position.x),
+    abscissaScale.invert(worldBottomLeft.x), // left
+    abscissaScale.invert(worldTopRight.x), // right
   ];
 
   const yVisibleDomain: Domain = [
-    ordinateScale.invert(-height / (2 * zoom) + position.y),
-    ordinateScale.invert(height / (2 * zoom) + position.y),
+    ordinateScale.invert(worldBottomLeft.y), // bottom
+    ordinateScale.invert(worldTopRight.y), // top
   ];
 
   // Re-render on every R3F frame (i.e. on every change of camera zoom/position)
--- a/packages/lib/src/vis/shared/Html.tsx
+++ b/packages/lib/src/vis/shared/Html.tsx
@@ -13,7 +13,7 @@ import ReactDOM from 'react-dom';
 import type { Group } from 'three';
 import { Vector3 } from 'three';
 
-const v1 = new Vector3();
+import { projectCameraToHtml } from '../utils';
 
 interface Props extends HTMLAttributes<HTMLDivElement> {
   groupProps?: GroupProps;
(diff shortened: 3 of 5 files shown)


packages/lib/src/vis/shared/AxisSystem.tsx
```diff
@@ -13,25 +14,31 @@ interface Props {
...
 
   const { abscissaConfig, ordinateConfig, abscissaScale, ordinateScale } =
     useAxisSystemContext();
 
-  const { position, zoom } = useThree((state) => state.camera);
+  const camera = useThree((state) => state.camera);
   const canvasSize = useThree((state) => state.size);
   const { width, height } = canvasSize;
 
+  const unprojectedLeftBottom = v1.set(-1, -1, 0).unproject(camera);
```
loichuder, line 30: `unproject` allows to move from the normalized camera space to the world space. 

The nice thing about the normalized space is that the FOV is always delimited by `(-1, -1)` and `(1,1)` which means I don't have to care about the `canvasSize` anymore: all is handled by `unproject`.

packages/lib/src/vis/shared/Html.tsx
```diff
@@ -68,14 +69,17 @@ function Html(props: Props) {
 
   const getInnerDivTransform = useCallback(() => {
     const position = followCamera ? getGroupPosition() : undefined;
+    const scale = v1
+      .setFromMatrixScale(camera.projectionMatrix) // scaling from projected to normalized
+      .multiply(sizeVector.set(width / 2, height / 2, 0)); // scaling from normalized to canvas size
```
loichuder, line 74: Technically, I would have to add the scale from world to local but this has no effect as our camera scale is always `1`. 

I don't know if it is worth the added complexity ?

packages/lib/src/vis/shared/TooltipMesh.tsx
```diff
@@ -38,15 +38,13 @@ function TooltipMesh(props: Props) {
   // When panning, events are handled and stopped by texture mesh and do not reach this mesh (which is behind)
   const onPointerMove = useCallback(
     (evt: ThreeEvent<PointerEvent>) => {
-      const { zoom } = camera;
-      const projectedPoint = camera.worldToLocal(evt.unprojectedPoint.clone());
-
       const abscissaCoord = abscissaScale.invert(evt.unprojectedPoint.x);
       const ordinateCoord = ordinateScale.invert(evt.unprojectedPoint.y);
 
+      const projectedPoint = evt.unprojectedPoint.clone().project(camera);
```
loichuder, line 44: Previously, the point was projected from the world space into the local space.

I go one step further with `project` to project from the world space into the local space into the normalized space.

packages/lib/src/vis/shared/TooltipMesh.tsx
```diff
@@ -38,15 +38,13 @@ function TooltipMesh(props: Props) {
...
       showTooltip({
-        tooltipLeft: projectedPoint.x * zoom + width / 2,
-        tooltipTop: -projectedPoint.y * zoom + height / 2,
+        tooltipLeft: (width / 2) * (projectedPoint.x + 1),
+        tooltipTop: -(height / 2) * (projectedPoint.y - 1),
```
loichuder, line 47: As `projectedPoint` is now in the normalized space, I don't have to care about the camera parameter but only to move from the normalized space into the HTML space.

I shift the center (HTML `0,0` is the top) and scale by `width / 2` in X and `height / 2` in Y. There is a sign change for Y as the Y-axis needs to be inverted.

packages/lib/src/vis/shared/AxisSystem.tsx
loichuder: Oh, and I need to set `v1` at each computation as `unproject` mutates it...

axelboc: This is awesome. Instead of mutating the vector, you could clone it. Here is what I'm thinking, with some naming suggestions:

```ts
const NORM_BOTTOM_LEFT= new Vector(-1, -1, 0);
...
const worldBottomLeft = NORM_BOTTOM_LEFT.clone().unproject(camera);
```

packages/lib/src/vis/shared/Html.tsx
axelboc: Not sure I follow here.

packages/lib/src/vis/shared/AxisSystem.tsx
```diff
@@ -13,25 +14,31 @@ interface Props {
...
+  const unprojectedRightTop = v2.set(1, 1, 0).unproject(camera);
+
   // Find visible domains from camera's zoom and position
```
axelboc, line 33: Some comments may need updating.

packages/lib/src/vis/shared/Html.tsx
axelboc: Yeah no, no need to add confusion.

axelboc (changes requested):
Took me a while to get my head around all this, but it was well worth it. It's a lot better indeed. Not dealing with the zoom/position of the camera is amazing. I wish we had known earlier about those two ThreeJS method...

(By the way, I personally found [this page](http://www.codinglabs.net/article_world_view_projection_matrix.aspx) much clearer than previous materials I had come across.)

axelboc:
Just remembered that when I was trying to understand the code in `Html`, I noticed that we don't have a story demonstrating an annotation that scales on zoom. Might be worth adding one.

packages/lib/src/vis/shared/AxisSystem.tsx
```diff
@@ -13,25 +14,31 @@ interface Props {
   title?: string;
 }
 
+const v1 = new Vector3();
+const v2 = new Vector3();
```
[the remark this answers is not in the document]
loichuder: I saw https://docs.pmnd.rs/react-three-fiber/advanced/pitfalls#don-t-re-create-objects-in-loops but perhaps that is an over-optimization in this case (we are not in a `useFrame`)

loichuder:
Thanks for the review. I tried to address your comments in 8ac901b0caa03f1712eeaf652c17f9ff54974bb8

apps/storybook/src/HeatmapVisDisplay.stories.tsx
```diff
@@ -109,6 +109,28 @@ WithAnnotation.args = {
   domain,
 };
 
+export const WithAnnotationZoom: Story<HeatmapVisProps> = (args) => (
```
axelboc, line 112: Awesome, thanks!

packages/lib/src/vis/shared/Html.tsx
```diff
@@ -57,25 +60,27 @@ function Html(props: Props) {
   const group = useRef<Group>(null!); // eslint-disable-line @typescript-eslint/no-non-null-assertion
 
   const getGroupPosition = useCallback(() => {
-    const groupPos = v1.setFromMatrixPosition(group.current.matrixWorld);
-    groupPos.project(camera);
+    const rawGroupPos = v1.setFromMatrixPosition(group.current.matrixWorld);
```
axelboc, line 63: I think you can initialize a new vector on the fly here (and below as well with `v1` again and `sizeVector`). Also, maybe rename the vars to `cameraPos` and `htmlPos` similarly to what you did for the scale in `TooltipMesh`?

```ts
const worldPos = new Vector3().setFromMatrixPosition(group.current.matrixWorld);
const cameraPos = worldPos.clone().project(camera);
const htmlPos = projectCameraToHtml(cameraPos , width, height);
```

You could also chain the call to `project` to avoid the extra `worldPos` constant initialization and cloning, but as discussed I reckon the performance impact is not significant, and two variables makes the code more readable in my opinion.

packages/lib/src/vis/utils.ts
```diff
@@ -293,3 +295,15 @@ const TYPE_STRINGS: Record<NumericType['class'], string> = {
...
 }
+
+export function projectCameraToHtml(
+  cameraVector: Vector3,
+  width: number,
+  height: number
+): Vector3 {
+  const cameraToHtmlMatrix = new Matrix4().makeScale(width / 2, -height / 2, 1);
+  // Account for shift of (0,0) position (center for camera, top-left for HTML)
+  cameraToHtmlMatrix.setPosition(width / 2, height / 2);
```
axelboc, line 306: Neat!

axelboc (approved):

Issue #915, opened 2022-01-04
Clicking on a group that contain a broken link prevents navigation


loichuder:
### Describe the bug

Clicking on a group that contain a broken link raises an error `Could not resolve soft link` that takes the whole screen. It prevents navigating back to another group or to look at other entities in the group containing the broken link.

### To Reproduce

1. Go to https://h5web.panosc.eu/?file=epics.h5
2. Click on `links` -> `soft links`
3. See error

### Expected behaviour

The error should be displayed in the main area to keep the explorer visible to go back to another group.

We could also imagine catching the link error to be able to display the broken link as an entity in the explorer. This would require some work in the back-end though.

### Context

- H5Web context: demo

loichuder:
Fixed by 5c7e3708012a5cda7b9f3f17a92f42fa2b0df791 and updating h5grove to 0.0.14

PR #918, opened 2022-01-05
Add link info for soft and external links in h5grove provider

loichuder:
Related to #915 

In our entity model, there is a `link` model but we didn't make use of it in the h5grove provider.

This PR adds the parsing of soft and external links responses to fill the `link` field of entities. 

For now, I did not change the model of entity but I think it will worth discussing if links should be first-class entities or not.


loichuder:
For now, this brings no operational changes. It will need https://github.com/silx-kit/h5grove/pull/58 to fix #915.

axelboc (approved):

PR #924, opened 2022-01-11
Support switching VisCanvas's sizing behaviour dynamically

axelboc:
We can now toggle the value of `VisCanvas`'s `canvasRatio` prop (between a ratio and `undefined`) and the the sizing behaviour of the canvas will change accordingly (i.e. adapt its ratio on resize or just fill the screen). This is thanks to https://github.com/react-hookz/web/commit/4c6f074eca31604d0975a9c8de1262b2fa8bda48, which now allows `useMeasure` to be toggled on/off dynamically.

### DIFF
--- a/packages/lib/src/vis/shared/VisCanvas.tsx
+++ b/packages/lib/src/vis/shared/VisCanvas.tsx
@@ -27,7 +27,7 @@ function VisCanvas(props: PropsWithChildren<Props>) {
   } = props;
 
   const shouldMeasure = !!canvasRatio;
-  const [areaSize, areaRef] = useMeasure<HTMLDivElement>();
+  const [areaSize, areaRef] = useMeasure<HTMLDivElement>(shouldMeasure);
   const canvasSize = areaSize && getSizeToFit(areaSize, canvasRatio);
 
   const axisOffsets = getAxisOffsets({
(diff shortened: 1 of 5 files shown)


packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -48,7 +48,10 @@ function VisCanvas(props: PropsWithChildren<Props>) {
       }}
     >
       {(!shouldMeasure || canvasSize) && (
-        <div className={styles.canvasWrapper} style={canvasSize}>
+        <div
+          className={styles.canvasWrapper}
+          style={shouldMeasure ? canvasSize : undefined}
```
loichuder, line 53: Why is the ternary needed ? Won't `canvasSize` be `undefined` anyway if `shouldMeasure` is `false` ?

axelboc:
Note that you cannot test this in Storybook. According to https://github.com/storybookjs/storybook/issues/13259#issuecomment-841011922, you cannot set a control's value to `undefined` if if was given a value in the story. Unfortunately, it seems you can't test it from another story either, as the reset button doesn't seem to reset the value of `canvasRatio` to `undefined` as suggested in the comment.

packages/lib/src/vis/shared/VisCanvas.tsx
axelboc: Unfortunately, no. `useMeasure` doesn't trigger a re-render with `undefined` when it receives `false` after receiving `true`, so `canvasSize` keeps its latest value.

axelboc: Maybe I'll open a PR in react-hookz at some point to fix this.

loichuder: All right then

loichuder (approved):

PR #925, opened 2022-01-12
Use shaders instead of sprites for glyphs

loichuder:
On the new ScatterVis, I noticed that some of the circles had ragged outlines due to mishandled transparency (see the top-right of the screenshot of https://github.com/silx-kit/h5web/pull/913).

To solve this, I reworked our `GlyphMaterial` to use shaders to define the shapes rather than sprites. This should also ease the addition of new shapes.


packages/lib/src/vis/line/GlyphMaterial.tsx
```diff
@@ -1,29 +1,96 @@
...
+            discard;
+        } else {
+            gl_FragColor = vec4(vColor.rgb, alpha);
+        }
+      }
+    `,
+  };
+
+  if (color) {
+    // If color is given, make it a uniform in the shader
```
loichuder, line 82: [Your recommendation](https://github.com/silx-kit/h5web/pull/913#discussion_r782226696) to set `vertexColors` only if `color` is not set made a lot of sense when using shaders so I implemented there.

packages/lib/src/vis/line/GlyphMaterial.tsx
```diff
@@ -1,29 +1,96 @@
...
+      <shaderMaterial
+        uniforms={{ color: { value: new Color(color) }, ...shader.uniforms }}
+        vertexShader={`uniform vec3 color; ${shader.vertexShader}`}
+        fragmentShader={shader.fragmentShader}
+      />
+    );
+  }
+
+  // If not, use vertex colors
+  return <shaderMaterial args={[shader]} vertexColors />;
```
loichuder, line 93: When setting `vertexColors`, `attribute vec3 color` is added to the vertex shader with the value of the current vertex attribute.

See https://stackoverflow.com/questions/25740054/three-js-passing-vertex-colors-to-fragment-shader/41863039 for more info

packages/lib/src/vis/line/DataCurve.tsx
```diff
@@ -66,7 +66,7 @@ function DataCurve(props: Props) {
           color={color}
         />
       )}
-    </Suspense>
```
loichuder, line 69: `Suspense` is no longer needed as we no longer load an external file

packages/lib/src/vis/line/GlyphMaterial.tsx
```diff
@@ -1,29 +1,96 @@
...
 
 interface Props {
   color?: string;
   size: number;
-  glyphURL?: string;
-  vertexColors?: boolean;
+  glyphType?: GlyphType;
 }
 
+// From https://github.com/silx-kit/silx/blob/11f764176ede6ad68d7551e215773a47bcb4614d/src/silx/gui/plot/backends/glutils/GLPlotCurve.py#L581-L799
```
loichuder, line 11: Big :+1: to @t20100 for pointing us towards silx implementations

packages/lib/src/vis/line/models.ts
```diff
@@ -1,20 +1,15 @@
-import capURL from './sprites/cap.png';
-import circleURL from './sprites/circle.png';
-import crossURL from './sprites/cross.png';
-import squareURL from './sprites/square.png';
```
axelboc, line 4: In `jest.config.js` of @h5web/app, there's a `transform` config to handle those. We should be able to remove it.

packages/lib/src/vis/line/GlyphMaterial.tsx
```diff
@@ -1,29 +1,96 @@
...
+  `,
+  [GlyphType.Circle]: `
+    float alphaSymbol(vec2 coord, float size) {
+      float radius = 0.5;
+      float r = distance(coord, vec2(0.5, 0.5));
+      return clamp(size * (radius - r), 0.0, 1.0);
+    }
+  `,
+  [GlyphType.Square]: `
+    float alphaSymbol(vec2 coord, float size) {
```
axelboc, line 42: Do the arguments have to be specified even though they're not used?

packages/lib/src/vis/line/GlyphMaterial.tsx
```diff
@@ -1,29 +1,96 @@
...
-  return (
-    <pointsMaterial
-      map={sprite}
-      color={color}
-      size={size}
-      vertexColors={vertexColors}
-      transparent
-    />
-  );
+  const { color, size, glyphType = GlyphType.Cross } = props;
```
axelboc, line 49: I would maybe not specify a default here, as it feels a bit arbitrary. What do you think?

packages/lib/src/vis/line/GlyphMaterial.tsx
```diff
@@ -1,29 +1,96 @@
...
+      void main() {
+        vColor = color;
+        gl_PointSize = size;
+        gl_Position = projectionMatrix * modelViewMatrix * vec4( position, 1.0 );
+      }
+    `,
+    fragmentShader: `
+      varying vec3 vColor;
+      uniform float size;
+      ${GLYPH_SHADERS[glyphType]}
```
axelboc, line 68: line break above since it adds a function

packages/lib/src/vis/line/GlyphMaterial.tsx
axelboc: Instead of the early return, how about conditionally adding the `color` uniform's value and declaration directly in the `uniforms` object and vertex shader with ternaries and a template string evaluation block? This would keep all the shader creation logic in one place.

axelboc (approved):
Awesome!

packages/lib/src/vis/line/GlyphMaterial.tsx
loichuder: Yes. The call in the fragment shader will still pass the two arguments so all `alphaSymbol` functions must have the same signature.

GLSL makes no compromise :joy:

packages/lib/src/vis/line/GlyphMaterial.tsx
loichuder: Good point, I will remove it

packages/lib/src/vis/line/GlyphMaterial.tsx
loichuder: That is what I had in the first place but the duplicated ternaries felt too contrived to me. Also, it made the placement of the comments a bit awkward.

I preferred to gather the conditional logic rather than the shader creation logic.

PR #928, opened 2022-01-13
Compute camera bounds from camera matrix rather than manually

loichuder:
Forgotten from #912

### DIFF
--- a/packages/lib/src/vis/shared/PanZoomMesh.tsx
+++ b/packages/lib/src/vis/shared/PanZoomMesh.tsx
@@ -2,13 +2,14 @@ import { useThree } from '@react-three/fiber';
 import type { ThreeEvent } from '@react-three/fiber/dist/declarations/src/core/events';
 import { clamp } from 'lodash';
 import { useRef, useCallback, useEffect } from 'react';
-import type { Vector3 } from 'three';
-import { Vector2 } from 'three';
+import { Vector2, Vector3 } from 'three';
 
 import { useWheelCapture } from '../hooks';
 import { useAxisSystemContext } from './AxisSystemContext';
 
 const ZOOM_FACTOR = 0.95;
+
+const CAMERA_TOP_RIGHT = new Vector3(1, 1, 0);
 
 function PanZoomMesh() {
   const { abscissaScale, ordinateScale, visSize } = useAxisSystemContext();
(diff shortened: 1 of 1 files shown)


packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -29,10 +30,14 @@ function PanZoomMesh() {
         abscissaScale.invert(x),
         ordinateScale.invert(y)
       );
+      const { position, projectionMatrixInverse } = camera;
+
+      const cameraBounds = CAMERA_TOP_RIGHT.clone().applyMatrix4(
```
loichuder, line 35: The projection is this time done from camera (normalized) space to local camera space. 

The reason is that the local camera space is centered on `camera.position`. That means the projection of the top right point is enough to get the bounds of the camera FOV in X and Y. Projecting the bottom left point would get the same values with a `-`.

axelboc: `cameraLocalBounds` and maybe a comment explaining this part "Projecting the bottom left point would get the same values with a -."?

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -43,16 +48,7 @@ function PanZoomMesh() {
...
-      abscissaScale,
-      camera,
-      height,
-      invalidate,
-      ordinateScale,
-      visWidth,
-      visHeight,
-      width,
-    ]
+    [abscissaScale, ordinateScale, camera, visWidth, visHeight, invalidate]
```
axelboc, line 51: Maybe the destructuring of `position` and `projectionMatrixInverse` should be done outside `useCallback` so the two objects can be referenced as dependencies (the `camera` object probably gets mutated by Three).

axelboc (approved):

packages/lib/src/vis/shared/PanZoomMesh.tsx
loichuder: Is it really needed given that `camera` is already needed as dependencies due to `camera.updateMatrixWorld()` below ?

axelboc: Totally, you're right. `camera.*` don't re-render the component, so there's no point in having them as dependencies (since the `useCallback` won't ever know they have changed). https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/

axelboc: This is why ESLint doesn't complain about `viewportCenter` not being a dep (it knows it's a mutable ref). Live and learn.

Issue #926, opened 2022-01-12
H5Web is no longer just a stand-alone HDF5 viewer


axelboc:
We should update the README(s) to convey that H5Web is more than a stand-alone viewer and that the library can be used to visualize data arrays from any source. We should also update the GIF, since the UI has evolved a little since it was recorded.

axelboc:
While we're at it, we can add more example files and maybe make them links for convenience.

loichuder:
We have the GIF that demonstrates `@h5web/app` but I think we should add some screenshots of `@h5web/lib` to highlight more the visualization capabilities. 

We could for example add screenshots of the example codesandboxes below the links ?

PR #929, opened 2022-01-13
Update README with new files and links

loichuder:
Part of #926 

I thought it might be easier to be able to specify the subdomain of HSDS so that files can be referred the same way whatever the provider.

I also cleaned-up the files on the back-end side.


axelboc (approved):
Awesome!

PR #930, opened 2022-01-13
Upgrade dependencies

axelboc:
Only the safe stuff for now, will try to upgrade CRA tomorrow.

### DIFF
--- a/apps/storybook/src/react-app-env.d.ts
+++ b/apps/storybook/src/react-app-env.d.ts
@@ -1,1 +1,0 @@
-/// <reference types="react-scripts" />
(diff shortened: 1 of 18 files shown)


apps/storybook/src/react-app-env.d.ts
```diff
@@ -1 +0,0 @@
-/// <reference types="react-scripts" />
```
axelboc, line 1: Switching to the file below, which saves us from having to install `react-scripts` in Storybook.

packages/app/README.md
```diff
@@ -32,9 +32,11 @@ import { App, MockProvider } from '@h5web/app';
 
 function MyApp() {
   return (
-    <MockProvider>
-      <App />
-    </MockProvider>
+    <div style={{ height: '100vh' }}>
```
axelboc, line 35: Adding this `div` for a smoother getting-started experience.

loichuder (approved):

Issue #914, opened 2021-12-30
Integration in Angular failing to import flattenChildren()


wlievens:
### Describe the bug

I'm attempting to integrate the component in an Angular application.
This is partially successful, I made an Angular wrapper component where I instantiate the h5web component as follows:

```
ReactDOM.render(
  React.createElement(H5GroveProvider, {
    url: url,
    filepath: '',
    children: React.createElement(App, {
      startFullscreen: false
    })
  }),
  document.getElementById(this.rootId)
);
```

My backend is h5grove compatible at the URL where I am testing it. I can browse the HDF5 tree just fine and can see metadata, attributes, and scalar values.

However, when I click a multidimensional dataset node, I get the following error:

```react_keyed_flatten_children__WEBPACK_IMPORTED_MODULE_5__ is not a function```

I traced this to the following line in the `Toolbar` function/class and checked with some console prints, and it indeed isn't a function ... it's the `react-keyed-flatten-children` module object being referred at the following line:

```const allChildren = flattenChildren(children).filter(isReactElement);```

If I replace that line with:

```const allChildren = flattenChildren.default(children).filter(isReactElement);```

everything works fine! So somehow it is failing to recognise the **default export** on `flattenChildren` when I integrate the component in Angular.

Now, I don't know much about TypeScript and JavaScript's import magic, so I tried various combinations of setting the tsconfig values for `esModuleInterop`, `allowSyntheticDefaultImports`, `isolatedModules` and so forth but to no avail.

### To Reproduce

I'm not sure anyone is interested in debugging my Angular setup since this is a React component so I was hoping for some wisdom from above regarding the peculiarities of TypeScript and/or JavaScript's import behaviour. If that leads us nowhere I can always try setting up a clean Angular sample app.

### Expected behaviour

No errors on a trivial function import.

### Screenshots

![image](https://user-images.githubusercontent.com/1073326/147791790-cc7bc9fc-b555-4ea8-95da-d66a125b53ae.png)

### Context

- OS: Windows
- Browser: Chrome
- Version: 1.3.0
- H5Web context: @h5web/lib, @h5web/app

wlievens:
This issue seems related, however I don't know what role webpack plays in the whole setup

https://github.com/webpack/webpack/issues/12675

axelboc:
[...]

PR #931, opened 2022-01-17
Remove `type: module` from published app/lib package.json

axelboc:
Trying this fix for #914. I've tested @h5web/app with CRA v5 outside of the repo and came across a couple of errors that I managed to fix by removing `"type": "module"` from the package, so I thought I'd give it a try.

Note that `"type": "module"` is required in development, so I "revert" it by setting to `"commonjs"` inside `"publishConfig"`.

I'll publish a beta version once merged.

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -15,6 +15,7 @@
   ],
   "main": "src/index.ts",
   "publishConfig": {
+    "type": "commonjs",
     "main": "dist/index.cjs",
     "module": "dist/index.js",
     "types": "dist/index.d.ts",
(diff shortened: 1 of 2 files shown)


loichuder (approved):
If I understand correctly, the commit message could be clearer: `Replace type: module by type: commonjs in published app/lib` package.json` ?

axelboc:
Well, it's equivalent since `commonjs` is the default behaviour.

loichuder:
> Well, it's equivalent since `commonjs` is the default behaviour.

Didn't know that :sweat_smile:

PR #932, opened 2022-01-17
Make zoom work with camera scale

loichuder:
Re-implementation of #909

Using `camera.scale` allows to circumvent the FOV reset on resize while still having the possibility to scale differently in X and Y for #769 

I also had to change a bit the computations like in #912 and #928 so that the change goes smoothly. I expect the computations to be generic enough so that reverting to use `zoom` instead of `scale` should be a breeze.

### DIFF
--- a/packages/lib/src/vis/shared/Html.tsx
+++ b/packages/lib/src/vis/shared/Html.tsx
@@ -66,20 +66,25 @@ function Html(props: Props) {
     return [htmlPos.x, htmlPos.y];
   }, [camera, height, width]);
 
+  const getGroupScale = useCallback(() => {
+    const groupScale = new Vector3().setFromMatrixScale(
+      camera.matrixWorldInverse
+    ); // Scale inverse to the camera scale
+
+    return [groupScale.x, groupScale.y];
+  }, [camera]);
+
   const getInnerDivTransform = useCallback(() => {
     const position = followCamera ? getGroupPosition() : undefined;
-    // Scale from canvas to camera space and from camera space to world space
-    const scale = new Vector3()
-      .setFromMatrixScale(camera.projectionMatrix)
-      .multiply(new Vector3().set(width / 2, height / 2, 0));
+    const scale = scaleOnZoom ? getGroupScale() : undefined;
 
     return [
       position && `translate3d(${position[0]}px, ${position[1]}px, 0)`,
-      scaleOnZoom ? `scale(${scale.x}, ${scale.y})` : undefined,
+      scale && `scale(${scale[0]}, ${scale[1]})`,
     ]
       .filter(isDefined)
       .join(' ');
-  }, [camera, followCamera, getGroupPosition, scaleOnZoom, width, height]);
+  }, [followCamera, getGroupPosition, getGroupScale, scaleOnZoom]);
 
   // Append/remove container `div` next to R3F's `canvas`
   useLayoutEffect(() => {
--- a/packages/lib/src/vis/shared/PanZoomMesh.tsx
+++ b/packages/lib/src/vis/shared/PanZoomMesh.tsx
@@ -10,6 +10,7 @@ import { useAxisSystemContext } from './AxisSystemContext';
 const ZOOM_FACTOR = 0.95;
 
 const CAMERA_TOP_RIGHT = new Vector3(1, 1, 0);
+const ONE_VECTOR = new Vector3(1, 1, 1);
 
 function PanZoomMesh() {
   const { abscissaScale, ordinateScale, visSize } = useAxisSystemContext();
@@ -30,12 +31,12 @@ function PanZoomMesh() {
         abscissaScale.invert(x),
         ordinateScale.invert(y)
       );
-      const { position, projectionMatrixInverse } = camera;
+      const { position } = camera;
 
-      // Project from normalized camera space (-1, -1) to (1, 1) to local camera space (-Xbound, -Ybound) to (Xbound, Ybound)
-      const cameraLocalBounds = CAMERA_TOP_RIGHT.clone().applyMatrix4(
-        projectionMatrixInverse
-      );
+      // Unproject from normalized camera space (-1, -1) to (1, 1) to world space and subtract camera position to get bounds
+      const cameraLocalBounds = CAMERA_TOP_RIGHT.clone()
+        .unproject(camera)
+        .sub(position);
 
       const xBound = Math.max(visWidth / 2 - cameraLocalBounds.x, 0);
       const yBound = Math.max(visHeight / 2 - cameraLocalBounds.y, 0);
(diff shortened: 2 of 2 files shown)


packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -30,12 +30,12 @@ function PanZoomMesh() {
...
+      const { position } = camera;
 
-      // Project from normalized camera space (-1, -1) to (1, 1) to local camera space (-Xbound, -Ybound) to (Xbound, Ybound)
-      const cameraLocalBounds = CAMERA_TOP_RIGHT.clone().applyMatrix4(
-        projectionMatrixInverse
-      );
+      // Unproject from normalized camera space (-1, -1) to (1, 1) to world space and subtract camera position to get bounds
+      const cameraLocalBounds = CAMERA_TOP_RIGHT.clone()
+        .unproject(camera)
+        .sub(position);
```
loichuder, line 38: Because the `scale` of the camera changes, the point needs to be unprojected in the world space. The bounds are then computed by removing the position offset.

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -52,17 +52,13 @@ function PanZoomMesh() {
...
+    const { sourceEvent, unprojectedPoint } = evt;
+    const { target, pointerId } = sourceEvent;
+    (target as Element).setPointerCapture(pointerId); // https://stackoverflow.com/q/28900077/758806
 
-      const projectedPoint = camera.worldToLocal(unprojectedPoint.clone());
-      startOffsetPosition.current = camera.position.clone().add(projectedPoint);
-    },
-    [camera]
-  );
+    startOffsetPosition.current = unprojectedPoint.clone();
```
loichuder, line 60: Rather than doing a mixture of projection/unprojection as before, I make `startOffsetPosition` the coordinates in the world space and work from here.

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -81,14 +77,12 @@ function PanZoomMesh() {
...
-      const projectedPoint = camera.worldToLocal(evt.unprojectedPoint.clone());
-      const { x: pointerX, y: pointerY } = projectedPoint;
-      const { x: startX, y: startY } = startOffsetPosition.current;
-
-      const targetX = startX - pointerX;
-      const targetY = startY - pointerY;
+      const delta = startOffsetPosition.current
+        .clone()
+        .sub(evt.unprojectedPoint);
+      const target = camera.position.clone().add(delta);
```
loichuder, line 83: Using world coordinates for `startOffsetPosition.current` makes possible to remove the projection here.

As a reminder, when panning, we want to compute the change in position from the starting position of the pointer to its current position (`delta`) and impact this change to the camera position (giving `target`).

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -98,17 +92,18 @@ function PanZoomMesh() {
       const { sourceEvent } = evt;
       const factor = sourceEvent.deltaY > 0 ? ZOOM_FACTOR : 1 / ZOOM_FACTOR;
 
-      camera.zoom = Math.max(1, camera.zoom * factor);
+      camera.scale.x = Math.min(1, camera.scale.x / factor);
+      camera.scale.y = Math.min(1, camera.scale.y / factor);
+
       camera.updateProjectionMatrix();
+      camera.updateMatrixWorld();
```
loichuder, line 99: Changing `scale` changes the `matrixWorld`

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -98,17 +92,18 @@ function PanZoomMesh() {
       const { sourceEvent } = evt;
       const factor = sourceEvent.deltaY > 0 ? ZOOM_FACTOR : 1 / ZOOM_FACTOR;
 
-      camera.zoom = Math.max(1, camera.zoom * factor);
+      camera.scale.x = Math.min(1, camera.scale.x / factor);
+      camera.scale.y = Math.min(1, camera.scale.y / factor);
```
loichuder, line 96: When zooming, the scale should decrease for the camera FOV to be smaller

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -98,17 +92,18 @@ function PanZoomMesh() {
...
       const projectedPoint = camera.worldToLocal(evt.unprojectedPoint.clone());
-      const { x: pointerX, y: pointerY } = projectedPoint;
-      const { x: camX, y: camY } = camera.position;
-
-      moveCameraTo(
-        camX + pointerX * (1 - 1 / factor),
-        camY + pointerY * (1 - 1 / factor)
+      const scaledPoint = camera.localToWorld(
+        projectedPoint.clone().multiplyScalar(1 - 1 / factor)
       );
```
loichuder, line 104: Again, I made the computation more generic by using matrices so that the change in scale is taken into account.

I could simplify this computation to remove projection and unprojection in a future commit.

axelboc (approved):
Nicely done!! So the panning just works thanks to the removal of the extra projection step in `startOffsetPosition`? Pretty cool 🎉

loichuder:
Could you have another look ? I had to change the scaling code in Html (I don't know who thought [our camera scale is always 1](https://github.com/silx-kit/h5web/pull/912#discussion_r771309678 ) :roll_eyes: )

I also modified the computation of the scaling of the position when zooming (https://github.com/silx-kit/h5web/pull/932#discussion_r785769836)

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -81,34 +79,35 @@ function PanZoomMesh() {
...
   const onWheel = useCallback(
     (evt: ThreeEvent<WheelEvent>) => {
-      const { sourceEvent } = evt;
+      const { sourceEvent, unprojectedPoint } = evt;
       const factor = sourceEvent.deltaY > 0 ? ZOOM_FACTOR : 1 / ZOOM_FACTOR;
 
-      camera.zoom = Math.max(1, camera.zoom * factor);
-      camera.updateProjectionMatrix();
+      const zoomVector = new Vector3(1 / factor, 1 / factor, 1);
+      camera.scale.multiply(zoomVector).clamp(ZERO_VECTOR, ONE_VECTOR);
```
axelboc, line 98: We can't really go below 0, though. Could we just clamp to `ONE_VECTOR` with `min()` or something?

packages/lib/src/vis/shared/Html.tsx
```diff
@@ -66,20 +66,25 @@ function Html(props: Props) {
     return [htmlPos.x, htmlPos.y];
   }, [camera, height, width]);
 
+  const getGroupScale = useCallback(() => {
+    const groupScale = new Vector3()
+      .setFromMatrixScale(camera.matrixWorldInverse) // Scale inverse to the camera scale
+      .multiplyScalar(camera.zoom); // Scale to zoom
```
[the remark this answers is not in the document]
loichuder: Yes, this is not needed in our case but I thought it might be better to keep the old behaviour :shrug: ?

packages/lib/src/vis/shared/PanZoomMesh.tsx
loichuder: Good call

packages/lib/src/vis/shared/Html.tsx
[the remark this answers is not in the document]
axelboc: Ah in case `Html` is used with another `PanZoomMesh` of sorts? Hmm I feel like it's more confusing than anything now ^^

axelboc (approved):

packages/lib/src/vis/shared/Html.tsx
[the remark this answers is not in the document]
loichuder: I get you, I removed it :+1:

PR #935, opened 2022-01-18
Fix selections moving with respect to data on resize

loichuder:
On resize, selection rectangles or line moved with respect to the data coordinates (see stories [Selecting regions](https://h5web-docs.panosc.eu/?path=/story/building-blocks-viscanvas-interaction--selecting-regions) or [Selecting lines](https://h5web-docs.panosc.eu/?path=/story/building-blocks-viscanvas-interaction--selecting-lines)).

I fix this by storing the starting and end point in data coordinates rather than world coordinates (making use of `abscissaScale` and `ordinateScale`). 

I use the scales once again to go from data coordinates in world coordinates to display the `Selection` component. This fixes the behaviour on resize as changes in canvas size will change the scales accordingly.


packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -14,48 +16,73 @@ interface Props {
...
+  const worldToData = useCallback(
+    (vec: Vector3) =>
+      new Vector2(abscissaScale.invert(vec.x), ordinateScale.invert(vec.y)),
+    [abscissaScale, ordinateScale]
+  );
+
+  const dataToWorld = useCallback(
+    (vec: Vector2) => new Vector2(abscissaScale(vec.x), ordinateScale(vec.y)),
+    [abscissaScale, ordinateScale]
+  );
```
axelboc, line 30: These are very nice. Could you maybe move them to the axis system context so we can re-use them in other places? They'll be very convenient to use in the new `useVisibleDomain` hook I was about to work on...

axelboc (approved):

packages/lib/src/vis/shared/SelectionMesh.tsx
loichuder: Sounds good.

Should we keep `Vector2` or use a more generic structure like `[number, number]` ?

axelboc: `Vector2` is great actually!

axelboc: it still represent a point in a different space.

axelboc (approved):

PR #936, opened 2022-01-18
Add hook to compute visible axis domains

axelboc:
@t20100 this should make your life much easier, as you requested :wink:


loichuder (approved):
Great ! Thanks for updating the doc, I forgot about it :sweat_smile:

PR #937, opened 2022-01-20
Remove REACT_APP env vars from `@h5web/app` package

axelboc:
The package should not be coupled with Create React App.

- `REACT_APP_PROFILING_ENABLED`: remove altogether, along with `Profiler` component which was set up to profile only the visualization components. For future performance debugging, we can use `ReactProfiler` directly and anywhere we want.
- `REACT_APP_DEFAULT_PATH`: replace env var with new `initialPath` prop on `App`. A value can be passed temporarily via this prop in the demo in development.


packages/app/README.md
```diff
@@ -66,6 +66,17 @@ visualization, which is useful when H5Web is embeded inside another app.
...
 
+### `initialPath?: string` (optional)
+
+The path to select within the file when the app is first rendered. Defaults to
+`'/'`.
+
+```tsx
+<MockProvider>
+  <App initialPath="/nD_datasets/threeD" />
+</MockProvider>
```
loichuder, line 77: It seems strange to pass this to `App` rather than through the `Provider` that is supposed to handle the files.

We should perhaps add the `initialPath` to the provider context and get it back in the `App`.

This being said, I have no scruple to merge this as-is as this is not the main focus of the PR.

loichuder (approved):

packages/app/README.md
axelboc: `initialPath` is just a UI state, and the `App` is the UI, so it still makes sense IMO. That being said, I get that there's coupling between the path and the file that is passed to the provider, but since the `App` and its provider are always rendered in the same place, I don't reckon there'd be a real gain in making it more convoluted by passing the path through the provider's context.

PR #939, opened 2022-01-20
Remove `demo-cra-lib` project from repo (replaced with CodeSandbox)

axelboc:


cypress/specs/app.spec.ts
```diff
@@ -18,7 +18,7 @@ describe('App', () => {
     cy.findByRole('figure', { name: 'oneD' }).should('exist');
 
     if (Cypress.env('TAKE_SNAPSHOTS')) {
-      cy.wait(SNAPSHOT_DELAY);
+      cy.wait(SNAPSHOT_DELAY * 2); /* first spec seems to fail more often */
```
axelboc, line 21: I've noticed that the first spec tends to fail randomly quite often, so I'm just increase the delay here to see.

loichuder (approved):
Should we also modify the section `Use built packages in demos` in `CONTRIBUTING.md` to put `demo` instead of `demos` ?

PR #941, opened 2022-01-21
Add a prop to SelectionMesh to run a callback on selection

loichuder:


apps/storybook/src/VisCanvasInteraction.stories.tsx
```diff
@@ -18,31 +21,56 @@ interface TemplateProps {
...
+            onSelection={(start, end) => setSelectedVectors([start, end])}
+          />
+        )}
+        {selection === 'rectangle' && (
+          <RectSelectionMesh
+            onSelection={(start, end) => setSelectedVectors([start, end])}
+          />
+        )}
+      </VisCanvas>
+    </>
```
loichuder, line 73: The template of the story is becoming very "selection-specific". Do you think I should move it to a new file `VisCanvasSelection.stories` ?

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -57,8 +58,11 @@ function SelectionMesh(props: Props) {
       (target as Element).releasePointerCapture(pointerId);
       setEndPoint(point);
       setDrag(false);
+      if (onSelection && startPoint && endPoint) {
+        onSelection(startPoint, endPoint);
+      }
```
loichuder, line 63: I thought it was best to run the callback after `onPointerUp` rather than use a `useEffect`.

axelboc: Yep, that's great 👍

apps/storybook/src/VisCanvasInteraction.stories.tsx
axelboc: Yeah, with the local state, it's looking a bit bloated, you're right.

axelboc (approved):

PR #938, opened 2022-01-20
Add CRA and Vite sandboxes and fix @h5web/lib stylesheet export

axelboc:
Next up, I'll remove the `demo-cra-lib` project from the repo.

PR #942, opened 2022-01-25
Migrate demo from CRA to Vite

axelboc:
Tested on FF 60 (i.e. the "browser not supported" error shows up as expected).

---

**EDIT** - Summary of changes:

- The `demo` is now based on Vite instead of CRA and no longer requires a `config-override.js` file to work.
- Environment variables `REACT_APP_DIST` and `STORYBOOK_DIST` have been removed, as testing distributed packages this way is not sufficiently reliable and not worth the complexity it adds to the monorepo. Distributed packages can be tested by publishing beta versions. The four Codesandboxes (#938) are available to test the published packages reliably inside both CRA and Vite apps before publishing official releases.
- The process of building the packages' styles is now more robust and a lot clearer. It is performed after the JS build and is based on a separate Vite config (`vite.styles.config.js`) and a separate entrypoint (`src/styles.ts`).
- Consumers of `@h5web/app` now need to import a single stylesheet instead of two.
- Building the packages, their styles and their type declarations, can now be done by running a single script: `pnpm packages` (no need to run `pnpm packages:dts` afterwards, as this is now done automatically).


.gitignore
```diff
@@ -1,7 +1,7 @@
 node_modules/
 
-/apps/**/build/
-/packages/**/dist*/
+build/
+dist*/
```
axelboc, line 4: Vite builds the demo to a folder named `dist`.

apps/demo/index.html
```diff
@@ -7,11 +7,13 @@
     <title>H5Web</title>
 
     <meta name="description" content="Web-based HDF5 file viewer." />
-    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/favicon192.png" />
+    <link rel="icon" href="/favicon.ico" />
+    <link rel="apple-touch-icon" href="/favicon192.png" />
+    <script src="https://polyfill.io/v3/polyfill.min.js?features=globalThis"></script>
```
axelboc, line 12: Using `polyfill.io` instead of `react-app-polyfill`. Polyfill.io returns a JS file with only the polyfills that are relevant to the browser that makes the request. So in modern browsers, the JS file is empty.

The downside is that the loading of the app is blocked until the polyfill script is loaded. The upside is that we no longer bundle a whole bunch of polyfills that modern browsers don't need. Since this project is a demo, I think it's a good compromise given how simple it is.

Note that the current URL loads Polyfill.io's `default` set of polyfills plus the `gloablThis` polyfill: https://polyfill.io/v3/url-builder/. This is enough for FF 60 to not crash and show the "browser version not supported" message.

apps/demo/index.html
```diff
@@ -7,11 +7,13 @@
...
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
+    <script type="module" src="/src/index.tsx"></script>
```
axelboc, line 17: Contrary to CRA, Vite's entrypoint is not `index.tsx`, but `index.html` (just like a browser), so we need to reference `index.tsx` here so Vite can find it a process it.

apps/demo/src/H5GroveApp.tsx
```diff
@@ -3,8 +3,8 @@ import { useLocation } from 'react-router-dom';
 
 import { getFeedbackURL } from './utils';
 
-const URL = process.env.REACT_APP_H5GROVE_URL || '';
-const FILEPATH = process.env.REACT_APP_H5GROVE_FALLBACK_FILEPATH || '';
+const URL = import.meta.env.VITE_H5GROVE_URL as string;
+const FILEPATH = import.meta.env.VITE_H5GROVE_FALLBACK_FILEPATH as string;
```
axelboc, line 7: Note the change from `process.env` to `import.meta.env`. I've also changed the empty string fallbacks into type casts because Vite types env vars as `string | boolean | undefined` -- the `boolean` part is weird, as in my tests I have not been able to get an env var parsed as boolean... See https://github.com/vitejs/vite/discussions/6637

apps/demo/src/index.tsx
```diff
@@ -1,17 +1,14 @@
-import 'react-app-polyfill/stable';
-
 import { StrictMode } from 'react';
 import ReactDOM from 'react-dom';
 
-import 'normalize.css';
```
axelboc, line 6: Moved to `styles.css` for consistency with a change I had to make in the @h5web/app package to import React-Reflex' global styles.

apps/demo/src/index.tsx
```diff
@@ -1,17 +1,14 @@
...
 import DemoApp from './DemoApp';
 
-if (process.env.REACT_APP_DIST === 'true') {
-  // Import distributed lib/app styles (can't use `import()` easily due to asynchronicity and lack of top-level `await` support)
-  require('@h5web/app/dist/style-lib.css'); // eslint-disable-line @typescript-eslint/no-require-imports
-  require('@h5web/app/dist/style.css'); // eslint-disable-line @typescript-eslint/no-require-imports
+if (import.meta.env.VITE_DIST === 'true') {
+  // Import distributed lib/app styles
+  import('@h5web/app/dist/style-lib.css');
+  import('@h5web/app/dist/style.css');
```
axelboc, line 11: It seems that Vite works better than CRA with regard to importing stylesheets conditionally/dynamically. The asynchronicity issue mentioned in the comment is nowhere to be seen.

package.json
```diff
@@ -52,10 +52,7 @@
   },
   "pnpm": {
     "overrides": {
-      "eslint": ">=8",
-      "react-scripts>eslint-plugin-testing-library": ">=5",
-      "react-scripts>@typescript-eslint/parser": ">=5",
-      "react-scripts>@typescript-eslint/eslint-plugin": ">=5"
+      "eslint": ">=8"
```
axelboc, line 55: Still need to keep this because of `vite-plugin-eslint`, as explained in `CONTRIBUTING.md`.

packages/app/src/explorer/Explorer.module.css
```diff
@@ -55,6 +55,7 @@
 }
 
 .name {
+  padding: 0.375rem 0;
```
axelboc, line 58: Here and above, I'm fixing a random cropping issue I've noticed in the Explorer:

image

This is caused by `overflow: hidden` below, which is required for the ellipsis to appear when the explorer is too narrow. The fix is to move the padding from the parent to the child with the overflow (`.name`).

tsconfig.json
```diff
@@ -22,6 +22,7 @@
     "noFallthroughCasesInSwitch": true,
     "forceConsistentCasingInFileNames": true,
     "importsNotUsedAsValues": "error",
+    "useDefineForClassFields": true,
```
axelboc, line 25: Vite enables this option when it bootstraps a new project: https://www.typescriptlang.org/tsconfig#useDefineForClassFields. Sounds like a good option to enable, so I do so in the root `tsconfig`.

axelboc:
Huge drop in `pnpm-lock.yaml` -- Vite has much fewer dependencies than CRA! 🎉  

image

apps/demo/src/H5GroveApp.tsx
loichuder: Isn't the type cast `as string` wrong if the variable is not set ? It will be `undefined` right ?

apps/demo/src/index.tsx
loichuder: This part prevents me from running the demo...

I get an error `Internal server error: Failed to resolve import "@h5web/app/dist/style-lib.css" from "src/index.tsx".` even if `import.meta.env.VITE_DIST` is `false`

I also get an inline eslint error in VSCode: `dynamic imports require a leading comment with the webpack chunknameeslintimport/dynamic-import-chunkname`

.github/workflows/approve-snapshots.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         run: pnpm build
 
       - name: Serve ⚡️
-        run: pnpm serve & pnpm dlx wait-on http://localhost:3000
+        run: pnpm serve & pnpm dlx wait-on http://localhost:5000
```
loichuder, line 52: Why do we need to change the port ?

tsconfig.json
loichuder: I struggle to see what kind of changes it brings (especially in our code) but all right

apps/demo/src/H5GroveApp.tsx
axelboc: All the variables are set as long as they're listed in `.env` - i.e. `VITE_H5GROVE_URL=` assigns an empty string to the variable. Also, it's slightly better to fail hardly because of `undefined` rather than silently because of an empty string.

Ideally, we'd want to assert that all the env vars respect some sort of schema (e.g. either `'true'` or `''` for `VITE_DIST`), but I haven't found a way to do this cleanly at runtime. It works with `assertStr` or equivalent, but only when the guard is invoked inside the function in which the env var is used. I can't assert all the en vars globally without writing unnecessarily complex code to access the asserted variables (in my opinion).

If Vite had a plugin of sorts, it'd be amazing. There's https://github.com/ElMassimo/vite-plugin-environment but I'm not a fan of the API (setting an env var's default value to `undefined` to say that it's required...) So I went with the status quo... 😄

.github/workflows/approve-snapshots.yml
axelboc: `vite preview` starts on port 5000 -- I think port 5000 is more commonly used for this sort of thing.

loichuder: On my end, `pnpm start` still runs on `http://localhost:3000/`. That means `start` and `serve` will end up on different ports which no longer makes them interchangeable.

For example, I can no longer run the cypress tests by having the app running through `pnpm start` :confused:

apps/demo/src/index.tsx
loichuder: Possibly linked to https://github.com/vitejs/vite/issues/6007

.github/workflows/approve-snapshots.yml
axelboc: Excellent point!

apps/demo/.env
```diff
@@ -3,27 +3,18 @@
 # Copy and configure the variables below in your own `.env.local` file
 #####
 
-# Whether to import @h5web/app's built files instead of its source files
-REACT_APP_DIST=
-
```
axelboc, line 8: Removed in favor of `import.meta.env.PROD` -- note that Vite is able to build the demo without using the distributed JS bundles of the packages. `PROD` is only used to import the distributed styles, as for some reason, global styles are not processed when building for production. I'm still investigating this.

apps/demo/src/index.tsx
axelboc: These two imports are no longer required and have been removed.

axelboc:
Alright, so as mentioned on the chat, removing the global styles imports from the `index.ts` files of the packages means that the packages' output stylesheets `dist/style.css` no longer include said global styles...

Instead of trying to find a sub-optimal solution that works for both the dev/build of the demo/Storybook and the build of the packages, I decided to take care of the problem on its own.

In the process, I managed to solve the problem of having to import two stylesheets when consuming the `@h5web/app` package. Now consumers just have to import `@h5web/app/dist/styles.css` (note the `s`). Also worth noting that I now run the `build:dts` script from the `build` script, so I've removed the `packages:dts` script from the root `package.json`.

I've explained how the packages are now build in details in the `CONTRIBUTING` guide. Hopefully it is clear. Relevant files (inside each package) are: `package.json`, `vite.styles.config.js` and `src/styles.ts`, `README.md`. Relevant commit: https://github.com/silx-kit/h5web/pull/942/commits/71ae4275b496843313417b488b047cc860f47659

apps/storybook/.storybook/main.js
```diff
@@ -6,11 +6,7 @@ module.exports = {
   },
   stories: ['../src/**/*.stories.@(tsx|mdx)'],
   addons: [
-    {
-      name: '@storybook/preset-create-react-app',
-      // Override react-scripts package path as it resolves to the wrong path on Linux
-      options: { scriptsPackageName: '../node_modules/react-scripts' },
```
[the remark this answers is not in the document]
loichuder: Nice, it looks cleaner

packages/app/src/styles.ts
```diff
@@ -0,0 +1,3 @@
+// Entry point of `vite.styles.config.js` to build global styles for distribution
+import 'react-reflex/styles.css';
+import '@h5web/lib/dist/styles.css';
```
loichuder, line 3: Why not using `import ./styles.css` as in `lib` ?

loichuder (approved):
:clap:

CONTRIBUTING.md
```diff
@@ -114,27 +112,40 @@ Icons can be imported as React components from `react-icons/fi`.
...
+
+### Package builds
+
+The build process of the packages works as follows:
+
+1. First, Vite builds the JS bundles in library mode starting from the package's
+   entrypoint: `src/index.ts`. The output bundles (ESM and CommonJS) are placed
+   in the output directory `dist` and referenced from `package.json`.
+
+   The JS build also generates a file called `style.css` in the `dist` folder
```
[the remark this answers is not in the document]
axelboc: Unfortunately, I don't have control over the filename 😢

CONTRIBUTING.md
```diff
@@ -114,27 +112,40 @@ Icons can be imported as React components from `react-icons/fi`.
...
+   that contains the compiled CSS modules that Vite comes across while building
+   the React components. These styles are called "local" styles.
+
+2. Second, we run two scripts in parallel: `build:css` and `build:ts`.
+   - The job of `build:css` is to build the package's global styles and
+     concatenate them with the local styles compiled at the first step. To do
+     so, we run Vite again but with a different config: `vite.styles.config.js`,
+     and a different entrypoint: `src/styles.ts`. The ouptut files are placed in
+     a temporary folder: `dist/temp`. We then concatenate `dist/temp/style.css`
```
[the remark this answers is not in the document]
axelboc: Same problem, Vite always outputs a file called `style.css`...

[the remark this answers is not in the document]
loichuder: Unfortunate. Btw, your comment made me notice a typo on the line above (`ouptut`)

apps/storybook/.storybook/main.js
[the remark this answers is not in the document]
loichuder: Note that the installation of `react-scripts` from the storybook does not solve the path resolution issue I had on Linux.

The `options: { scriptsPackageName: '../node_modules/react-scripts' },` should be kept

packages/app/src/styles.ts
axelboc: Because in this case, I want to import the distributed lib files, not the global lib styles. This is how we allow consumers to import only one CSS file when using `@h5web/app`. I have clarified the comments in `styles.ts/css` in both packages and replaced `import ./styles.css` with the full shared styles import in `lib/src/styles.ts` to avoid any confusion.

loichuder: Oh I see

axelboc: I missed this behaviour when documenting the package build process in `CONTRIBUTING.md`, sorry. I'm working on it.

axelboc: Done: https://github.com/silx-kit/h5web/pull/942/commits/b1eab4e873b87cf35835c75108848d364f46ddae

apps/storybook/.storybook/main.js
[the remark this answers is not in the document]
axelboc: 👍 it's back in

PR #945, opened 2022-01-31
Flip only the needed meshes rather than the whole canvas

loichuder:
This is in essence a revert of https://github.com/silx-kit/h5web/pull/813

While this seemed very elegant for texture-based meshes, this introduces an issue for meshes that make use of the axis scales to position themselves (through `dataToWorld`). 

Indeed, the `flip` is already accounted for in the scales (which make sense for axes) so that coordinates computed through them **already have the correct flipped coordinates**. As a consequence, having an inverting group around such meshes flip them once more **which is incorrect**.

Examples of such meshes include `SelectionMesh` or the ones used in `LineVis`.

### DIFF
--- a/packages/lib/src/vis/rgb/RgbVis.tsx
+++ b/packages/lib/src/vis/rgb/RgbVis.tsx
@@ -78,7 +78,7 @@ function RgbVis(props: Props) {
         }}
       >
         <PanZoomMesh />
-        <VisMesh>
+        <VisMesh scale={[1, -1, 1]}>
           <meshBasicMaterial map={texture} />
         </VisMesh>
         {children}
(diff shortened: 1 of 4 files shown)


packages/lib/src/vis/rgb/RgbVis.tsx
```diff
@@ -78,7 +78,7 @@ function RgbVis(props: Props) {
         }}
       >
         <PanZoomMesh />
-        <VisMesh>
+        <VisMesh scale={[1, -1, 1]}>
```
axelboc, line 81: no ternary here ?

loichuder: Unneeded as we force the flip of Y is set to `true` above (part of unchanged code)

axelboc (approved):

PR #946, opened 2022-01-31
Let HeatmapVis accept typed ndarrays

axelboc:
This adds support for typed ndarrays in both `HeatmapVis` and `HeatmapMesh`. I haven't touched `@h5web/app` yet, this is just for `@h5web/lib`.

Both `dataArray` and `alpha.array` can now have type `NdArray<number[] | CompatibleTypedArray>`, where `CompatibleTypedArray` is `TypedArray` without `Uint8ClampedArray` and `Float64Array`, neither of which are supported by WebGL (at least to my knowledge, feel free to correct me @t20100).

I've added a new story too. You can tell it works because the `dataArray` control shows that the `data` property of the ndarray has 820 "keys" (instead of "items" when the ndarray contains a simple JS array):

image

### DIFF
--- a/apps/storybook/src/HeatmapVis.stories.tsx
+++ b/apps/storybook/src/HeatmapVis.stories.tsx
@@ -6,7 +6,7 @@ import {
   getMockDataArray,
   INTERPOLATORS,
 } from '@h5web/lib';
-import { assertDefined } from '@h5web/shared';
+import { assertDefined, toTypedNdArray } from '@h5web/shared';
 import type { Meta, Story } from '@storybook/react/types-6-0';
 import ndarray from 'ndarray';
 
@@ -92,6 +92,13 @@ Alpha.args = {
   dataArray,
   domain,
   alpha: { array: alphaArray, domain: alphaDomain },
+};
+
+export const TypedArray = Template.bind({});
+TypedArray.args = {
+  dataArray: toTypedNdArray(dataArray),
+  domain,
+  alpha: { array: toTypedNdArray(alphaArray), domain: alphaDomain },
 };
 
 export default {
--- a/packages/lib/src/vis/heatmap/HeatmapMesh.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapMesh.tsx
@@ -1,30 +1,25 @@
 import type { Domain } from '@h5web/shared';
 import { ScaleType, assertDefined, isScaleType } from '@h5web/shared';
 import { rgb } from 'd3-color';
+import type { NdArray } from 'ndarray';
 import { memo, useMemo } from 'react';
-import {
-  RedFormat,
-  DataTexture,
-  FloatType,
-  RGBFormat,
-  UnsignedByteType,
-} from 'three';
+import type { TextureDataType } from 'three';
+import { DataTexture, RGBFormat, UnsignedByteType } from 'three';
 
 import { useAxisSystemContext } from '../..';
 import type { VisScaleType } from '../models';
 import VisMesh from '../shared/VisMesh';
-import type { ColorMap, ScaleShader } from './models';
-import { getInterpolator } from './utils';
+import type { ColorMap, CompatibleTypedArray, ScaleShader } from './models';
+import { getDataTexture, getInterpolator } from './utils';
 
 interface Props {
-  rows: number;
-  cols: number;
-  values: number[];
+  values: NdArray<CompatibleTypedArray>;
   domain: Domain;
-  colorMap: ColorMap;
   scaleType: VisScaleType;
+  colorMap: ColorMap;
   invertColorMap?: boolean;
-  alphaValues?: number[];
+  textureType?: TextureDataType; // override default texture type (determined from `values.dtype`)
+  alphaValues?: NdArray<CompatibleTypedArray>;
   alphaDomain?: Domain;
 }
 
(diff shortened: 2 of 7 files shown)


packages/lib/src/vis/heatmap/utils.ts
```diff
@@ -111,3 +151,15 @@ export function getInterpolator(colorMap: ColorMap, reverse: boolean) {
...
 }
+
+export function getTextureFormatFromType(type: TextureDataType): PixelFormat {
+  if (type === FloatType || type === HalfFloatType) {
+    return RedFormat;
+  }
+
+  if (UNSUPPORTED_TEXTURE_TYPES.has(type)) {
+    throw new Error('Texture type not supported');
+  }
```
axelboc, line 162: TS won't let me create a narrower version of `TextureDataType` with the unsupported types excluded... I could only manage to do a runtime check...

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -142,31 +141,34 @@ const SCALE_SHADER: Record<ScaleType, ScaleShader> = {
...
   const { ordinateConfig } = useAxisSystemContext();
 
   const dataTexture = useMemo(() => {
-    const valuesArr = Float32Array.from(values);
-    return new DataTexture(valuesArr, cols, rows, RedFormat, FloatType);
-  }, [cols, rows, values]);
+    const { data, shape } = values;
+    const format = getTextureFormatFromType(textureType);
+    return new DataTexture(data, shape[1], shape[0], format, textureType);
+  }, [textureType, values]);
```
loichuder, line 160: Perhaps it would be worth abstracting the hook/utility to construct a new `DataTexture` from a ndarray

packages/lib/src/vis/heatmap/HeatmapVis.tsx
```diff
@@ -116,14 +125,12 @@ function HeatmapVis(props: Props) {
           }}
         />
         <HeatmapMesh
-          rows={rows}
-          cols={cols}
-          values={dataArray.data}
+          values={toTypedNdArray(dataArray)}
```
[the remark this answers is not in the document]
loichuder: At a cost of a copy. But that can be avoided by passing a TypedNdArray. Nice.

packages/shared/src/utils.ts
```diff
@@ -40,6 +40,16 @@ export function toArray(arr: NdArray<number[]> | number[]): number[] {
   return 'data' in arr ? arr.data : arr;
 }
 
+export function toTypedNdArray<T extends number[] | TypedArray>(
```
loichuder, line 43: Would it be possible for the typing to return `NdArray<Float32Array>` when `T extends number[]` ?

So that we know that `toTypedNdArray` does not convert to _any_ TypedArray ?

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -142,31 +141,34 @@ const SCALE_SHADER: Record<ScaleType, ScaleShader> = {
...
   const {
-    rows,
-    cols,
     values,
     domain,
-    colorMap,
     scaleType,
+    colorMap,
     invertColorMap = false,
+    textureType = TEXTURE_TYPE_BY_DTYPE[values.dtype],
```
loichuder, line 149: Would it be possible to have a `dtype` that is not a key of `TEXTURE_TYPE_BY_DTYPE` ? Making `textureType` `undefined` ?

If yes, we should account for it

packages/lib/src/vis/heatmap/utils.ts
loichuder: Unfortunate but that will work too

apps/storybook/src/HeatmapVis.stories.tsx
```diff
@@ -6,7 +6,7 @@ import {
   getMockDataArray,
   INTERPOLATORS,
 } from '@h5web/lib';
-import { assertDefined } from '@h5web/shared';
+import { assertDefined, toTypedNdArray } from '@h5web/shared';
```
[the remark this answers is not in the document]
loichuder: I think `@h5web/shared` is a great place for it

loichuder (approved):

packages/lib/src/vis/heatmap/models.ts
```diff
@@ -29,3 +30,8 @@ export interface ScaleShader {
   ) => Record<string, { value: number }>;
   fragment: string;
 }
+
+export type CompatibleTypedArray = Exclude<
+  TypedArray,
+  Uint8ClampedArray | Float64Array
```
t20100, line 36: I don't know if they are always present depending on JS version, but `BigInt64Array` and `BigUInt64Array` should also be excluded.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
t20100: It's the case for `float64` and "big ints".

axelboc: This line is safe, since I'm restricting the type of `values` to `NdArray<CompatibleTypedArray>`.

t20100 (approved):
Nice, this will be great to use float16!

packages/lib/src/vis/heatmap/models.ts
axelboc: `TypedArray` comes from `@types/ndarray`, in which bigint types are treated separately for backwards compatibility (because only the latest versions of TS know what bigints are).

packages/shared/src/utils.ts
axelboc: Re-reading your comment, I'd like to clarify that the return type `NdArray<Exclude<T, number[]> | Float32Array>`  does not mean _any_ TypedArray. It keeps whichever TypeArrays were allowed in the input parameter. So when we call it in `HeatmapVis` with `NdArray<number[] | CompatibleTypedArray>`, we get back `NdArray<CompatibleTypedArray>` (since `CompatibleTypedArray` already includes `Float32Array`).

The only downside of the current TS signature is that if you call the function with an ndarray of type `NdArray<Uint16Array>`, you'll get back `NdArray<Uint16Array | Float32Array>`.

But the function is not really meant to be called with an already typed ndarray -- it's meant to be used when you don't know whether you have a typed of `number[]` ndarray like in the `HeatmapVis`.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
axelboc: Good call, it's much cleaner. As discussed, I didn't go as far as creating a memoized hook so as to not have to handle the `undefined` case with function overrides...

packages/shared/src/utils.ts
loichuder: > Re-reading your comment, I'd like to clarify that the return type `NdArray<Exclude<T, number[]> | Float32Array>` does not mean _any_ TypedArray. It keeps whichever TypedArrays were allowed in the input parameter. So when we call it in `HeatmapVis` with `NdArray<number[] | CompatibleTypedArray>`, we get back `NdArray<CompatibleTypedArray>` (since `CompatibleTypedArray` already includes `Float32Array`).

Understood, but I thought we could make the return type stronger. You could have `T extends number[] ? NdArray<Float32Array> : NdArray<Exclude<T, number[]>` to avoid the issue you mention below.
> 
> The only downside of the current TS signature is that if you call the function with an ndarray of type `NdArray<Uint16Array>`, you'll get back `NdArray<Uint16Array | Float32Array>`.

axelboc: Doesn't work, unfortunately. I get errors on the two return statements saying that the type of what I'm returning is not assignable to the function's return conditional type without further explanations...

loichuder: Yeah, I figured that it would be a bit too tortuous for TypeScript to make its way. Let's keep it like this then

PR #947, opened 2022-02-02
Allow modifier keys for SelectionMesh and PanZoomMesh

loichuder:
To use `SelectionMesh` and the `PanZoomMesh` on the same vis, we have to differentiate the clicks linked to the panning and the one linked to the selection.

This PR allows to specify a modifier key (`Alt`, `Ctrl` or `Shift`) to these meshes so that only clicks accompanied by the press of the key trigger the mesh callbacks. 

Also, previously, mesh callbacks were triggered on any clicks (including the one accompanied by press of modifier keys) so I had to change this so that clicks can be differentiated. 

I also added some stories to test things out.

### DIFF
--- a/packages/lib/src/vis/models.ts
+++ b/packages/lib/src/vis/models.ts
@@ -90,3 +90,5 @@ export interface HistogramParams {
   colorMap?: ColorMap;
   invertColorMap?: boolean;
 }
+
+export type ModifierKey = 'Alt' | 'Control' | 'Shift';
--- a/packages/lib/src/vis/shared/LineSelectionMesh.tsx
+++ b/packages/lib/src/vis/shared/LineSelectionMesh.tsx
@@ -1,14 +1,8 @@
-import type { MeshProps } from '@react-three/fiber';
-import type { Vector2 } from 'three';
-
 import SelectionLine from './SelectionLine';
+import type { SelectionMeshProps } from './SelectionMesh';
 import SelectionMesh from './SelectionMesh';
 
-interface Props extends MeshProps {
-  onSelection?: (startPoint: Vector2, endPoint: Vector2) => void;
-}
-
-function LineSelectionMesh(props: Props) {
+function LineSelectionMesh(props: SelectionMeshProps) {
   return <SelectionMesh selectionComponent={SelectionLine} {...props} />;
 }
 
--- a/packages/lib/src/vis/shared/PanZoomMesh.tsx
+++ b/packages/lib/src/vis/shared/PanZoomMesh.tsx
@@ -5,13 +5,19 @@ import { useRef, useCallback, useEffect } from 'react';
 import { Vector2, Vector3 } from 'three';
 
 import { useWheelCapture } from '../hooks';
-import { CAMERA_TOP_RIGHT } from '../utils';
+import type { ModifierKey } from '../models';
+import { CAMERA_TOP_RIGHT, isEventValid } from '../utils';
 import { useAxisSystemContext } from './AxisSystemContext';
 
 const ZOOM_FACTOR = 0.95;
 const ONE_VECTOR = new Vector3(1, 1, 1);
 
-function PanZoomMesh() {
+interface Props {
+  panKey?: ModifierKey;
+}
+
+function PanZoomMesh(props: Props) {
+  const { panKey } = props;
   const { abscissaScale, ordinateScale, visSize } = useAxisSystemContext();
   const { width: visWidth, height: visHeight } = visSize;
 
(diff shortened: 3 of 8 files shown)


packages/lib/src/vis/models.ts
```diff
@@ -90,3 +90,9 @@ export interface HistogramParams {
   colorMap?: ColorMap;
   invertColorMap?: boolean;
 }
+
+export enum ModifierKey {
+  Alt = 'Alt',
+  Ctrl = 'Control',
+  Shift = 'Shift',
+}
```
loichuder, line 98: Modifier keys can be any string but I thought it was better to restrict them at first

packages/lib/src/vis/shared/LineSelectionMesh.tsx
```diff
@@ -1,14 +1,8 @@
...
 import SelectionLine from './SelectionLine';
+import type { SelectionMeshProps } from './SelectionMesh';
 import SelectionMesh from './SelectionMesh';
 
-interface Props extends MeshProps {
-  onSelection?: (startPoint: Vector2, endPoint: Vector2) => void;
-}
-
-function LineSelectionMesh(props: Props) {
+function LineSelectionMesh(props: SelectionMeshProps) {
```
loichuder, line 5: `modifierKey` is added in `SelectionMeshProps`. I added this interface to avoid copying the `Props` between this mesh and `RectSelectionMesh`

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -90,8 +103,11 @@ function PanZoomMesh() {
   const onWheel = useCallback(
     (evt: ThreeEvent<WheelEvent>) => {
       const { sourceEvent, unprojectedPoint } = evt;
-      const factor = sourceEvent.deltaY > 0 ? ZOOM_FACTOR : 1 / ZOOM_FACTOR;
+      if (!isEventValid(sourceEvent, modifierKey)) {
+        return;
+      }
```
loichuder, line 108: I added the check for `modifierKey` for the wheel event as well. Not sure if that is worth it ? :shrug:

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -51,6 +66,9 @@ function SelectionMesh(props: Props) {
 
   const onPointerUp = useCallback(
     (evt: ThreeEvent<PointerEvent>) => {
+      if (!isDragging) {
+        return;
+      }
```
loichuder, line 71: This prevents the execution of the callback if `onPointerDown` a valid starting point was not found

packages/lib/src/vis/utils.ts
```diff
@@ -309,3 +310,15 @@ export function projectCameraToHtml(
...
 }
+
+export function isEventValid(
+  event: MouseEvent,
+  modifierKey: ModifierKey | undefined
+) {
+  if (!modifierKey) {
+    // If no modifier key is given, the event is deemed valid only if no modifier is are pressed
+    return !event.altKey && !event.ctrlKey && !event.shiftKey;
+  }
```
loichuder, line 321: If `modifierKey` is present, the event is valid if the key is pressed.

If `modifierKey` is not present, we expect the event to be valid for "simple" clicks (i.e. clicks with no modifier key pressed). I also considered using `!Object.values(ModifierKey).some((k) => event.getModifierState(k))` instead of the chained conditions.

packages/lib/src/vis/models.ts
axelboc: Might be simpler to use a union `'Alt' | 'Control' | 'Shift'`. Avoids having to import the enum in the consumer component.

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -52,13 +58,20 @@ function PanZoomMesh() {
...
 
-  const onPointerDown = useCallback((evt: ThreeEvent<PointerEvent>) => {
-    const { sourceEvent, unprojectedPoint } = evt;
-    const { target, pointerId } = sourceEvent;
-    (target as Element).setPointerCapture(pointerId); // https://stackoverflow.com/q/28900077/758806
+  const onPointerDown = useCallback(
+    (evt: ThreeEvent<PointerEvent>) => {
+      const { sourceEvent, unprojectedPoint } = evt;
+      const { target, pointerId } = sourceEvent;
+      (target as Element).setPointerCapture(pointerId); // https://stackoverflow.com/q/28900077/758806
```
axelboc, line 65: Doesn't capturing the pointer here interfere with the mesh selection (or breaks the pointer capture in the selection mesh)? Might be worth moving this down after the event validity check.

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -28,13 +39,17 @@ function SelectionMesh(props: Props) {
...
 
+      if (!isEventValid(sourceEvent, modifierKey)) {
+        return;
+      }
+
       const point = worldToData(unprojectedPoint);
 
       setStartPoint(point);
       setEndPoint(point);
       setDrag(true);
```
axelboc, line 50: Random thought: maybe this should be called `setPanning` (and `isPanning`)?

packages/lib/src/vis/utils.ts
axelboc: It's nice and clear the way it is (and allows using a type union).

axelboc (approved):
👍 

(Up to you if you want to change the format of the prop in this PR or later.)

packages/lib/src/vis/shared/PanZoomMesh.tsx
loichuder: It seems to work fine in the story.

This being said, I think the pointer capture must be revisited: I noticed some hiccups when releasing the pointer outside the canvas.

packages/lib/src/vis/shared/SelectionMesh.tsx
loichuder: I wanted to convey the idea that we are _dragging_ the pointer to draw the shape. In my opinion, `isPanning` makes think of the pan from the `PanZoomMesh`.

packages/lib/src/vis/shared/PanZoomMesh.tsx
loichuder: Still, it makes sense to capture the pointer only if the rest of the callback is executed, you are right.

packages/lib/src/vis/shared/SelectionMesh.tsx
axelboc: Oh sorry, mixed up the components, thought this was in `PanZoomMesh` 😅

PR #948, opened 2022-02-02
Allow passing custom `magFilter` to `HeatmapMesh`

axelboc:
- I simply add a prop called `magFilter` to `HeatmapMesh` that I pass along to the new `getDataTexture` utility.
- I also document `HeatmapMesh` as a new building block in Storybook.

One of the stories tests the texture override feature by passing `HalfFloatTexture` with a `Uint16Array`. For some reason, I get a black heatmap... Not sure if I'm doing things wrong in the story, or if the texture override feature does not work as expected. Any idea?

### DIFF
--- a/apps/storybook/src/HeatmapMesh.stories.tsx
+++ b/apps/storybook/src/HeatmapMesh.stories.tsx
@@ -1,0 +1,121 @@
+import {
+  getDomain,
+  getMockDataArray,
+  HeatmapMesh,
+  VisCanvas,
+} from '@h5web/lib';
+import type { HeatmapMeshProps, Domain } from '@h5web/lib';
+import { ScaleType, toTypedNdArray } from '@h5web/shared';
+import type { Meta, Story } from '@storybook/react/types-6-0';
+import ndarray from 'ndarray';
+import {
+  ByteType,
+  FloatType,
+  HalfFloatType,
+  IntType,
+  LinearFilter,
+  NearestFilter,
+  ShortType,
+  UnsignedByteType,
+  UnsignedIntType,
+  UnsignedShortType,
+} from 'three';
+
+import FillHeight from './decorators/FillHeight';
+
+const dataArray = getMockDataArray('/nD_datasets/twoD');
+const domain = getDomain(dataArray.data);
+
+const uint16Values = [0x49_00, 0x4d_00, 0x4f_80, 0x51_00]; // 10, 20, 30, 40
+const uint16DataArray = ndarray(Uint16Array.from(uint16Values), [2, 2]);
+const uint16Domain: Domain = [10, 40];
+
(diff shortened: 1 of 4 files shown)


loichuder:
> One of the stories tests the texture override feature by passing `HalfFloatTexture` with a `Uint16Array`. For some reason, I get a black heatmap... Not sure if I'm doing things wrong in the story, or if the texture override feature does not work as expected. Any idea?

Perhaps, I guess it is because you can take the buffer from `Uint16Array` and interpret it as half floats. Values are passed through the texture but are ridiculously small (<1e-4). We should generate a buffer of half floats to store it in a `Uint16Array` as @t20100 is doing in his prototype.

Also I can't make the `HeatmapMesh` work with `Uint16Array` even without the texture override. Perhaps it is because `texture2D` in the shader is always expected to contain floats ? :shrug:

axelboc:
Such a counter-intuitive format... I managed to create a 2x2 test array with https://evanw.github.io/float-toy/. I also improved the controls for `magFilter` and `textureType` (not that changing the texture type does anything interesting, but at least it shows the options instead of a number constant).

axelboc:
image

apps/storybook/src/HeatmapMesh.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+} from 'three';
+
+import FillHeight from './decorators/FillHeight';
+
+const dataArray = getMockDataArray('/nD_datasets/twoD');
+const domain = getDomain(dataArray.data);
+const [rows, cols] = dataArray.shape;
+
+const uint16DataArray = ndarray(
+  Uint16Array.from([0x49_00, 0x4d_00, 0x4f_80, 0x51_00]),
```
loichuder, line 31: It would be nice to put the decimal values above as comments

loichuder (approved):
Great :+1:

apps/storybook/src/HeatmapMesh.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+  [2, 2]
+);
+const uint16Domain: Domain = [10, 40];
+
+const Template: Story<HeatmapMeshProps> = (args) => {
+  return (
+    <VisCanvas
+      abscissaConfig={{ visDomain: [0, cols] }}
+      ordinateConfig={{ visDomain: [0, rows] }}
```
loichuder, line 40: Given that you set different arrays for `values`, it might be better to get `cols` and `rows` from `args.values` in here

axelboc: Ha, good catch, didn't even notice that the axes were wrong.

axelboc: I've also set `isIndexAxis` for obvious reasons.

PR #949, opened 2022-02-03
Allow disabling pan/zoom and modifier keys to zoom in X or Y

loichuder:
Work for #769 

Follow-up of #947

### DIFF
--- a/apps/storybook/src/VisCanvasInteraction.stories.tsx
+++ b/apps/storybook/src/VisCanvasInteraction.stories.tsx
@@ -1,26 +1,24 @@
+import type { PanZoomProps, TooltipMeshProps } from '@h5web/lib';
 import { PanZoomMesh, TooltipMesh, VisCanvas } from '@h5web/lib';
-import type { TooltipMeshProps } from '@h5web/lib';
 import { formatTooltipVal } from '@h5web/shared';
 import type { Meta, Story } from '@storybook/react';
 
 import VisCanvasStoriesConfig from './VisCanvas.stories';
 
-interface TemplateProps {
-  panZoom?: boolean;
-  panKey?: 'Alt' | 'Control' | 'Shift';
+interface TemplateProps extends PanZoomProps {
   tooltipValue?: string;
   guides?: TooltipMeshProps['guides'];
 }
 
 const Template: Story<TemplateProps> = (args) => {
-  const { panZoom = false, tooltipValue, guides, panKey } = args;
+  const { tooltipValue, guides, ...panZoomProps } = args;
 
   return (
     <VisCanvas
       abscissaConfig={{ visDomain: [-10, 0], showGrid: true }}
       ordinateConfig={{ visDomain: [50, 100], showGrid: true }}
     >
-      {panZoom && <PanZoomMesh panKey={panKey} />}
+      <PanZoomMesh {...panZoomProps} />
       {tooltipValue && (
         <TooltipMesh
           guides={guides}
@@ -39,32 +37,64 @@ const Template: Story<TemplateProps> = (args) => {
 };
 
 export const PanZoom = Template.bind({});
-PanZoom.args = { panZoom: true };
-PanZoom.argTypes = { guides: { table: { disable: true } } };
-
-export const PanZoomModifiers = Template.bind({});
-PanZoomModifiers.args = {
-  panZoom: true,
-  panKey: 'Alt',
+PanZoom.args = {
+  pan: true,
+  zoom: true,
 };
 PanZoom.argTypes = { guides: { table: { disable: true } } };
 
+export const PanModifier = Template.bind({});
+PanModifier.args = {
+  pan: true,
+  zoom: true,
+  panKey: 'Alt',
+};
+PanModifier.argTypes = { guides: { table: { disable: true } } };
+
+export const PanZoomInXAndY = Template.bind({});
+PanZoomInXAndY.args = {
+  pan: true,
+  zoom: true,
+  xZoom: true,
+  yZoom: true,
+};
+PanZoomInXAndY.argTypes = { guides: { table: { disable: true } } };
+
+export const ZoomModifierInX = Template.bind({});
+ZoomModifierInX.args = {
+  pan: true,
+  zoom: true,
+  xZoom: true,
+  xZoomKey: 'Control',
+};
+ZoomModifierInX.argTypes = { guides: { table: { disable: true } } };
+
 export const Tooltip = Template.bind({});
 Tooltip.args = {
+  pan: false,
+  zoom: false,
   tooltipValue: 'no value to display',
 };
 
 export const TooltipWithGuides = Template.bind({});
 TooltipWithGuides.args = {
+  pan: false,
+  zoom: false,
   tooltipValue: 'guides="both"',
   guides: 'both',
 };
 
 export const TooltipWithPanZoom = Template.bind({});
 TooltipWithPanZoom.args = {
-  panZoom: true,
+  pan: true,
+  zoom: true,
   tooltipValue: '<PanZoomMesh /> must come first',
   guides: 'vertical',
+};
+
+const keyArgType = {
+  control: { type: 'inline-radio' },
+  options: ['Alt', 'Control', 'Shift'],
 };
 
 export default {
@@ -73,7 +103,15 @@ export default {
   parameters: {
     ...VisCanvasStoriesConfig.parameters,
     controls: {
-      include: ['panZoom', 'tooltipValue', 'guides', 'panKey'],
+      include: [
+        'pan',
+        'panKey',
+        'zoom',
+        'xZoomKey',
+        'yZoomKey',
+        'tooltipValue',
+        'guides',
+      ],
     },
   },
   argTypes: {
@@ -81,9 +119,8 @@ export default {
       control: { type: 'inline-radio' },
       options: ['horizontal', 'vertical', 'both'],
     },
-    panKey: {
-      control: { type: 'inline-radio' },
-      options: ['Alt', 'Control', 'Shift'],
-    },
+    panKey: keyArgType,
+    xZoomKey: keyArgType,
+    yZoomKey: keyArgType,
   },
 } as Meta;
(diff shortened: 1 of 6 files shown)


apps/storybook/src/VisCanvasInteraction.stories.tsx
```diff
@@ -1,26 +1,25 @@
 import { PanZoomMesh, TooltipMesh, VisCanvas } from '@h5web/lib';
 import type { TooltipMeshProps } from '@h5web/lib';
+import type { PanZoomProps } from '@h5web/lib/src/vis/shared/PanZoomMesh';
```
loichuder, line 3: Should I expose this type in `@h5web/lib` ?

apps/storybook/src/VisCanvasInteraction.stories.tsx
```diff
@@ -1,26 +1,25 @@
...
-  const { panZoom = false, tooltipValue, guides, panKey } = args;
+  const { tooltipValue, guides, ...panZoomProps } = args;
 
   return (
     <VisCanvas
       abscissaConfig={{ visDomain: [-10, 0], showGrid: true }}
       ordinateConfig={{ visDomain: [50, 100], showGrid: true }}
     >
-      {panZoom && <PanZoomMesh panKey={panKey} />}
+      <PanZoomMesh {...panZoomProps} />
```
loichuder, line 22: `panZoom` is no longer necessary in the story as we can equivalently disable panning and zooming though the props `pan` and `zoom`

apps/storybook/src/VisCanvasInteraction.stories.tsx
```diff
@@ -39,51 +38,71 @@ const Template: Story<TemplateProps> = (args) => {
...
 };
-PanZoom.argTypes = { guides: { table: { disable: true } } };
+PanZoomModifiers.argTypes = { guides: { table: { disable: true } } };
+
+export const PanZoomInXAndY = Template.bind({});
+PanZoomInXAndY.args = {
+  zoomKeyX: 'Alt',
+  zoomKeyY: 'Shift',
+};
+PanZoomInXAndY.argTypes = { guides: { table: { disable: true } } };
```
loichuder, line 54: This is the story to look at if you want to check the new behaviour

apps/storybook/src/VisCanvasInteraction.stories.tsx
```diff
@@ -39,51 +38,71 @@ const Template: Story<TemplateProps> = (args) => {
...
 TooltipWithGuides.args = {
+  pan: false,
+  zoom: false,
   tooltipValue: 'guides="both"',
   guides: 'both',
 };
 
 export const TooltipWithPanZoom = Template.bind({});
 TooltipWithPanZoom.args = {
-  panZoom: true,
```
loichuder, line 65: I could explicitely set `pan` and `zoom` to `true` rather than relying on the defaults. What do you think ?

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -103,8 +108,17 @@ function PanZoomMesh(props: Props) {
...
+        return;
+      }
+
       const factor = sourceEvent.deltaY > 0 ? ZOOM_FACTOR : 1 / ZOOM_FACTOR;
-      const zoomVector = new Vector3(1 / factor, 1 / factor, 1);
+      const noKeyPressed = noModifierKeyPressed(sourceEvent);
+      const zoomVector = new Vector3(
+        noKeyPressed || isEventValid(sourceEvent, zoomKeyX) ? 1 / factor : 1,
+        noKeyPressed || isEventValid(sourceEvent, zoomKeyY) ? 1 / factor : 1,
+        1
```
loichuder, line 120: As discussed, a WheelEvent with no key pressed should always allow to zoom/unzoom in both directions. If a key a pressed, I check if this corresponds to the X zoom or the Y zoom to set the `zoomVector` values.

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -1,23 +1,27 @@
...
+  pan?: boolean;
   panKey?: ModifierKey;
+  zoom?: boolean;
+  zoomKeyX?: ModifierKey;
+  zoomKeyY?: ModifierKey;
 }
 
 function PanZoomMesh(props: Props) {
-  const { panKey } = props;
+  const { pan = true, zoom = true, panKey, zoomKeyX, zoomKeyY } = props;
```
axelboc, line 24: Seeing this API in practice now, I realize that it's a bit unclear that `zoomKeyX` and `zoomKeyY` serve to enable an additional feature on top of `zoom` - i.e. it's not because `zoom` is `true` that the zoom on X/Y is enabled. This is a bit inconsistent with the behaviour of the `panKey` prop, which does not enable a new feature - it just configures a keybinding. In other words, I feel like configuring the keybindings for zooming on X/Y should be possible independently of enabling zooming on X/Y.

Perhaps we should provide two additional boolean props for this. Doing so would allow us to set sane defaults for `zoomKeyX` and `zoomKeyY` so consumers could just write `<PanZoomMesh zoomX zoomY />` without worrying about key bindings. It would also simplify your `isEventValid` logic.

When it comes to naming, since `zoomXKey` is a bit ugly, I propose `xZoom` and `xZoomKey`.

What do you think?

loichuder: If we go this way, I would like to clarify a few usecases. Tell me if you think this is right or not:
- If setting only `zoomX` to `true` _without a key_, the wheel (with no key pressed) should zoom only in X and not Y
- If setting only `zoomX` to `true` _with a key_, the wheel (with no key pressed) should  **not** allow to zoom. It should zoom in X **only** if the wheel is scrolled with the key pressed.
- If setting `zoomX` and `zoomY` to `true` with keys for both, we agreed that the wheel (with no key pressed) should zoom in X and Y.
- If setting `zoomX` to `true` _with a key_ and `zoomY` to `true` _without a key_, the wheel (with no key pressed) should still zoom in X and Y. The only axial zoom allowed is the X zoom when pressing the corresponding key when scrolling the wheel.

If we agree on the behaviour above, I don't see how it will simplify the logic. Or did I miss something ?

axelboc: I was suggesting that the x/y zoom keys would always be set (because we'd give them the default values `Alt` and `Shift`). The goal was solely to make the API clearer by making the activation of the x-only and y-only zoom features clearer. I was _not_ suggesting that enabling either x-only or y-only zoom without a key would disable the global x/y zoom. Sorry for the confusion.

Enabling axial zooming in one direction without a modifier key is a use case (perhaps interesting for line visualizations) that we hadn't considered before, I think. It definitely complexifies the problem. 😅 Perhaps it's not worth worrying about for now.

axelboc: Here are all the use cases I can think of:

| Use case | Current API | Proposed API |
| --- | --- | --- |
| No global zoom, no axial zoom | `zoom: false` | `zoom: false` |
| Global zoom only (no modifier key) | `zoom: true` | `zoom: true` |
| Global zoom + x-only zoom with `Alt` +  | `zoom: true, zoomKeyX: 'Alt'` | `zoom: true, xZoom: true` |
| Global zoom + y-only zoom with `Shift` +  | `zoom: true, zoomKeyY: 'Shift'` | `zoom: true, yZoom: true` |
| Global zoom + x-only zoom with `Alt` + y-only zoom with `Shift` +  | `zoom: true, zoomKeyX: 'Alt', zoomKeyY: 'Shift'` | `zoom: true, xZoom: true, yZoom: true` |
| Global zoom + x-only zoom with `Shift` + y-only zoom with `Alt` +  | `zoom: true, zoomKeyX: 'Shift', zoomKeyY: 'Alt'` | `zoom: true, xZoom: true, xZoomKey: 'Shift', yZoom: true, yZoomKey: 'Alt'` |
| Global zoom with `Ctrl` | not supported | not supported |
| X-only zoom (no modifier key, no global zoom) | not supported | not supported |
| Y-only zoom (no modifier key, no global zoom) | not supported | not supported |
| X-only zoom (no modifier key, no global zoom) + Y-only zoom with `Ctrl` | not supported | not supported |
| Y-only zoom (no modifier key, no global zoom) + X-only zoom with `Ctrl` | not supported | not supported |

apps/storybook/src/VisCanvasInteraction.stories.tsx
loichuder: In the end, I made `pan` and `zoom` explicit for each story

packages/lib/src/vis/shared/PanZoomMesh.tsx
loichuder: Thanks, that's way clearer. I think the implementation should now match this

packages/lib/src/vis/shared/PanZoomMesh.tsx
axelboc: That being said, is there a point in preventing zoom when a modifier key is pressed and `x/yZoom` is `false`? It may be convenient to be able to zoom while drawing a selection? 

In this case, the ternary could be simplified and inverted: `!xZoom || (xZoom && isEventValid(sourceEvent, xZoomKey)) ? 1 : 1 / factor` and since `xZoomKey` is always defined, this could just become `!xZoom || (xZoom && sourceEvent.getModifierState(xZoomKey)) ? 1 : 1 / factor`.

Since `isEventValid` and `noModifierKeyPressed` would no longer be used for the zoom, you'd then be able to inline the `isEventValid` logic in the pan callback: 

```ts
if (!pan) {
  return;
}

const isPanAllowed = panKey ? event.getModifierState(panKey) : noModifierKeyPressed(event);
if (isPanAllowed) {
  (target as Element).setPointerCapture(pointerId); // https://stackoverflow.com/q/28900077/758806
  startOffsetPosition.current = unprojectedPoint.clone();
}
```

axelboc (approved):
Huge feature, took a lot of work to get here, congrats 👏

packages/lib/src/vis/shared/PanZoomMesh.tsx
loichuder: As discussed, breaking down the pan/zoom in multiple components should allow to solve such quandaries.

Before that, removing `isEventValid` is appealing to me so I will implement your proposed changes, merge as it is and work on the breaking down next.

axelboc: Yep, no worries 👍

PR #952, opened 2022-02-04
Let `MappedMatrixVis` and `MatrixVis` accept typed arrays

axelboc:


packages/lib/src/vis/matrix/MatrixVis.tsx
```diff
@@ -8,14 +8,14 @@ import GridProvider from './context';
...
+  dataArray: NdArray<ArrayValue<PrintableType>>;
   formatter: (value: Primitive<PrintableType>) => string;
   cellWidth: number;
-  sticky: boolean;
+  sticky?: boolean;
 }
 
 function MatrixVis(props: Props) {
-  const { dataArray, formatter, cellWidth, sticky } = props;
+  const { dataArray, formatter, cellWidth, sticky = true } = props;
```
axelboc, line 18: Just adding a default value to `sticky`.

packages/shared/src/models-hdf5.ts
```diff
@@ -155,6 +157,10 @@ export type Primitive<T extends DType> = T extends NumericType
   ? H5WebComplex
   : unknown;
 
+export type ArrayValue<T extends DType> =
+  | Primitive<T>[]
+  | (T extends NumericType ? TypedArray : never);
```
axelboc, line 162: This is the key new type for this PR. `ArrayValue<PrintableType>` resolves to `Primitive<PrintableType>[] | TypedArray`, or `number[] | string[] | ... | TypedArray`.

Once we're ready for providers to return typed arrays, I'll use `ArrayValue<D['type']>` below on line 167.

loichuder: I think we could make the name `ArrayValue` more explicit but that will do.

loichuder (approved):

PR #953, opened 2022-02-07
Split PanZoom mesh in Pan and Zoom meshes

loichuder:
Also refactored the viewport centering on resize: it was extracted from the interaction meshes into a dedicated component `ViewportCenter`.

### DIFF
--- a/packages/lib/src/vis/shared/PanMesh.tsx
+++ b/packages/lib/src/vis/shared/PanMesh.tsx
@@ -1,0 +1,80 @@
+import { useThree } from '@react-three/fiber';
+import type { ThreeEvent } from '@react-three/fiber';
+import { useRef, useCallback } from 'react';
+import type { Vector3 } from 'three';
+
+import type { ModifierKey } from '../models';
+import { noModifierKeyPressed } from '../utils';
+import { useMoveCameraTo } from './hooks';
+
+interface Props {
+  disabled?: boolean;
+  modifierKey?: ModifierKey;
+}
+
+function PanMesh(props: Props) {
+  const { disabled, modifierKey } = props;
+
+  const camera = useThree((state) => state.camera);
+  const { width, height } = useThree((state) => state.size);
+
+  const startOffsetPosition = useRef<Vector3>(); // `useRef` to avoid re-renders
+
+  const moveCameraTo = useMoveCameraTo();
+
+  const onPointerDown = useCallback(
+    (evt: ThreeEvent<PointerEvent>) => {
+      const { sourceEvent, unprojectedPoint } = evt;
+      const { target, pointerId } = sourceEvent;
+
+      if (disabled) {
+        return;
+      }
+
+      const isPanAllowed = modifierKey
+        ? sourceEvent.getModifierState(modifierKey)
+        : noModifierKeyPressed(sourceEvent);
+      if (isPanAllowed) {
+        (target as Element).setPointerCapture(pointerId); // https://stackoverflow.com/q/28900077/758806
+        startOffsetPosition.current = unprojectedPoint.clone();
+      }
+    },
+    [disabled, modifierKey]
+  );
+
+  const onPointerUp = useCallback((evt: ThreeEvent<PointerEvent>) => {
+    const { sourceEvent } = evt;
+    const { target, pointerId } = sourceEvent;
+    (target as Element).releasePointerCapture(pointerId); // https://stackoverflow.com/q/28900077/758806
+
+    startOffsetPosition.current = undefined;
+  }, []);
+
+  const onPointerMove = useCallback(
+    (evt: ThreeEvent<PointerEvent>) => {
+      if (disabled || !startOffsetPosition.current) {
+        return;
+      }
+
+      // Prevent events from reaching tooltip mesh when panning
+      evt.stopPropagation();
+
+      const delta = startOffsetPosition.current
+        .clone()
+        .sub(evt.unprojectedPoint);
+      const target = camera.position.clone().add(delta);
+
+      moveCameraTo(target.x, target.y);
+    },
+    [camera, disabled, moveCameraTo]
+  );
+
+  return (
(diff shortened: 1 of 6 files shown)


packages/lib/src/vis/shared/PanMesh.tsx
```diff
@@ -0,0 +1,80 @@
...
+    [camera, moveCameraTo]
+  );
+
+  return (
+    <mesh {...{ onPointerMove, onPointerUp, onPointerDown }}>
+      <meshBasicMaterial opacity={0} transparent />
+      <planeGeometry args={[width, height]} />
+    </mesh>
+  );
+}
```
loichuder, line 78: No changes except for the removal of the `viewportCenter` code

packages/lib/src/vis/shared/PanZoomMesh.tsx
```diff
@@ -23,154 +13,13 @@ interface Props {
 }
 
 function PanZoomMesh(props: Props) {
```
loichuder, line 15: For ease of refactoring and testing in the stories, I kept `PanZoomMesh` and its `Props` for now. 

Now, it only calls the new meshes `PanMesh` and `ZoomMesh`.

packages/lib/src/vis/shared/ViewportCenterer.tsx
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  useEffect(() => {
+    if (viewportCenter.current) {
+      // On resize, move camera to the latest saved viewport center coordinates
+      const { x, y } = dataToWorld(viewportCenter.current);
+      moveCameraTo(x, y);
+    }
+  }, [viewportCenter, moveCameraTo, dataToWorld, position.x, position.y]);
+  return <mesh />;
```
loichuder, line 26: I was unsure of what kind of component to put here. As everything happens in hooks, we might instead put the logic in a custom hook and call this hook in `AxisSystem` instead ? :shrug:

packages/lib/src/vis/shared/ViewportCenterer.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const moveCameraTo = useMoveCameraTo();
+
+  useFrame(() => {
+    viewportCenter.current = worldToData(position);
+  });
+
+  useEffect(() => {
+    if (viewportCenter.current) {
+      // On resize, move camera to the latest saved viewport center coordinates
+      const { x, y } = dataToWorld(viewportCenter.current);
```
loichuder, line 22: Note that I used the new utilities `worldToData` and `dataToWorld` rather than the scales. I find it easier to read.

packages/lib/src/vis/shared/ViewportCenterer.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+function ViewportCenterer() {
+  const { dataToWorld, worldToData } = useAxisSystemContext();
+  const viewportCenter = useRef<Vector2>();
+  const { position } = useThree((state) => state.camera);
+
+  const moveCameraTo = useMoveCameraTo();
+
+  useFrame(() => {
+    viewportCenter.current = worldToData(position);
```
loichuder, line 16: Rather than computing the new camera position in interaction meshes and store the computed position in the callback, the camera position (in data coordinates) is now saved at each frame.

packages/lib/src/vis/shared/ZoomMesh.tsx
```diff
@@ -0,0 +1,54 @@
...
+        ? 1 / factor
+        : 1,
+      noKeyPressed || (yZoom && sourceEvent.getModifierState(yZoomKey))
+        ? 1 / factor
+        : 1,
+      1
+    );
+  };
+
+  const onWheel = useZoomOnWheel(zoomVector, disabled);
```
loichuder, line 44: By passing `zoomVector` to the callback, I expect to be able to reuse `useZoomOnWheel` for axial zooms as well.

packages/lib/src/vis/shared/PanMesh.tsx
axelboc: Nice!

I think the `onPointerMove` callback should also return early if `disabled` is `true`, in case the pan gets disabled while the user is panning.

packages/lib/src/vis/shared/ViewportCenterer.tsx
axelboc: That's what it's for! :tada:

packages/lib/src/vis/shared/ViewportCenterer.tsx
axelboc: Can you return `null`? I think it's good to have a component that can be rendered explicitly from `VisCanvas`.

packages/lib/src/vis/shared/ZoomMesh.tsx
axelboc: Nice solution :100:

packages/lib/src/vis/shared/ViewportCenterer.tsx
axelboc: Could `moveCameraTo` just accept a vector now?

axelboc (approved):
Awesome, this opens up a lot of possibilities.

packages/lib/src/vis/shared/ViewportCenterer.tsx
loichuder: `null` seems to do the trick

packages/lib/src/vis/shared/ViewportCenterer.tsx
loichuder: We could but two times out of three, we call it with a `Vector3`.

So we could have two possibility:
- Accept `Vector2 | Vector3` and discard the `z` component in the callback
- Accept only `Vector2` and create a new `Vector2` from the `Vector3` in the caller.

Both solutions are not very satisfactory to me. With `x, y`, we make clear that `z` is not changed.

Issue #853, opened 2021-11-23
Keep active visualization when switching between Visualize/Inspect


andygotz:

axelboc:
Fixed by #886

PR #954, opened 2022-02-07
Restore last used vis when coming back to vis panel

loichuder:
Fix https://github.com/silx-kit/h5web/issues/853


axelboc (changes requested):
Smart call to use `localStorage`! I had wondered how this could be done with React state without introducing unnecessary coupling/complexity.

You may want to use https://react-hookz.github.io/web/?path=/docs/side-effect-uselocalstoragevalue--example, though, as this will take care of feature detection and has features that may help simplify the `useEffect`. You'll probably want to pass `handleStorageEvent: false`, the active vis as `defaultValue`, and `storeDefaultValue: true`.

Also, make sure to pass a dependency array to `useEffect`.

axelboc:
I feel like there may be a way to simplify `useActiveVis` as well, or to move this logic to `useActiveVis`, as this is where we deal with keeping the same visualization active when switching from one dataset to another. Perhaps the `localStorage` solution can take care of both problems at once?

Also, note that the `useEffect` introduces an asynchronicity problem: the `VisManager` is most likely rendered with the last supported vis for a split second before the latest vis is restored from `localStorage`. To solve this, we were setting the active vis state synchronously in `useActiveVis`, so this is probably the way to go. `localStorage` access is synchronous, so this shouldn't be a problem.

loichuder:
> I feel like there may be a way to simplify `useActiveVis` as well, or to move this logic to `useActiveVis`, as this is where we deal with keeping the same visualization active when switching from one dataset to another. Perhaps the `localStorage` solution can take care of both problems at once?

So I made use of `useLocalStorageState` to store our `activeVis` state in local storage rather than on the component. 

The behaviour when switching from one dataset to another is then kept but if the state is no longer reset when the component is unmounted (e.g. when switching from Display to Inspect and back).

Please give a try and tell me how it worked for you

axelboc:
Alright, so as discussed, I've change the active vis behaviour. Please try it out and let me know what you think @loichuder @t20100. I'll fix the tests afterwards.

### BEFORE my commit

1. Select `nD_datasets/oneD` => live vis is active
2. Select `twoD` => line vis remains active
3. Switch to inspect and back => line vis remains active

### AFTER my commit

1. Select `nD_datasets/oneD` => live vis is active (no preferred vis, so default to last supported)
2. Select `twoD` => heatmap is active (no preferred vis, so default to last supported)
5. Switch to inspect and back => heatmap remains active (no preferred vis, so default to last supported)
3. Switch to matrix vis (=> matrix becomes the preferred vis in local storage)
4. Select `oneD` again => matrix vis remains active (preferred vis)
5. Switch to inspect and back => matrix vis remains active (preferred vis)
5. Switch to line vis - i.e. `oneD`'s last supported vis (=> preferred vis is removed)
6. Switch to heatmap vis => heatmap vis is active (no preferred vis, so default to last supported)

loichuder:
Sounds good to me.

I only have a minor issue: there is no way to set the last supported vis as the preferred vis. For example, when selecting `Line` on a `oneD` dataset, one might want to keep this vis when selecting `twoD` but this is impossible as selecting `Line` on `oneD` will clear the preferred vis.

t20100:
Sounds good to me, too.

The important selection to persist for me is when changing between 2 datasets with same number of dimensions, and also peristing the selection when switching back from "inspect" is nice to have.

axelboc:
I've added some tests @loichuder

packages/app/src/visualizer/Visualizer.tsx
```diff
@@ -24,7 +24,9 @@ function Visualizer(props: Props) {
   }
 
   const { entity, supportedVis } = resolution;
-  return <VisManager entity={entity} supportedVis={supportedVis} />;
+  return (
```
loichuder, line 27: Little comment to explain the presence of the `key` ?

loichuder (commented):
Tests are great :clap:

PR #955, opened 2022-02-07
Remove PanZoomMesh in favour of individual meshes

### LINKED ISSUE #769: [Zoom] Give more zooming options

loichuder:
Fix #769

This time, `PanZoomMesh` is truly gone ! We now favour the use of the individual meshes:
- `PanMesh` to enable panning in vis bounds
- `ZoomMesh` to enable zooming in X and Y on the mouse position
- `XAxisZoomMesh` to enable zooming in X on the mouse position
- `YAxisZoomMesh` to enable zooming in Y on the mouse position

All meshes can be disabled and accept a `modifierKey` through props (with `Alt` the default for X and `Shift` the default for Y, the others have no default).

Stories and visualisations have been updated with the new components. Notably, zooming in X and/or Y is now allowed in `Line` (and its cousin `NxSpectrum`).


apps/storybook/src/VisCanvasInteraction.stories.tsx
```diff
@@ -1,24 +1,52 @@
...
+    tooltipValue,
+    guides,
+    disablePan,
+    panKey,
+    disableZoom,
+    disableXZoom,
+    xZoomKey,
+    disableYZoom,
+    yZoomKey,
+  } = args;
```
axelboc, line 39: Because of all the prop aliases, I feel like this template no longer documents the new meshes very well. Perhaps we should have one template per mesh? We can always store the canvas' axis configs into a shared object that we then destructure. Might also be worth splitting all the pan/zoom interactions into a separate file to stay clear of the tooltip? All of this can be left to a future PR.

packages/lib/src/vis/shared/XAxisZoomMesh.tsx
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+  const onWheel = useZoomOnWheel(zoomVector, disabled);
+
+  return (
+    <mesh onWheel={onWheel}>
+      <meshBasicMaterial opacity={0} transparent />
+      <planeGeometry args={[width, height]} />
+    </mesh>
+  );
```
axelboc, line 36: Can be left to another PR, but I feel like this could be abstracted to a `CanvasMesh` of sorts.

packages/lib/src/vis/shared/XAxisZoomMesh.tsx
```diff
@@ -0,0 +1,39 @@
...
+  modifierKey?: ModifierKey;
+}
+
+function XAxisZoomMesh(props: Props) {
+  const { disabled, modifierKey = 'Alt' } = props;
+
+  const { width, height } = useThree((state) => state.size);
+
+  const zoomVector = (sourceEvent: WheelEvent) => {
+    const factor = sourceEvent.deltaY > 0 ? ZOOM_FACTOR : 1 / ZOOM_FACTOR;
```
axelboc, line 20: Could this be computed in `useZoomOnWheel` and passed as an argument? `ZOOM_FACTOR` is the same everywhere, so there's no point to duplicate this logic for now, I reckon.

axelboc (approved):
:tada: :champagne: 

FYI, I've added a reference to the issue in the PR description.

apps/storybook/src/VisCanvasInteraction.stories.tsx
loichuder: No you are right, I might as well do it in this PR to properly document each mesh.

loichuder: So I split into two files: one for the tooltip and one for the PanZoom meshes. In PanZoom, each story documents an individual mesh with controls that reflect its props and that act only on the relevant mesh. 

But all meshes are still present in each story as `Zoom` is needed for `Pan` to be used.

Tell me what you think.

packages/lib/src/vis/shared/XAxisZoomMesh.tsx
loichuder: Abstracted in `InteractionMesh`

packages/lib/src/vis/shared/XAxisZoomMesh.tsx
loichuder: I changed this to return booleans instead. The conversion in `zoomVector` is then done in the callback

PR #950, opened 2022-02-04
Let `MappedHeatmapVis` and related utilities accept typed arrays

axelboc:
This has already impacted quite a few files, so I thought I'd keep further changes (line vis, providers, etc.) for separate PRs.

PR #956, opened 2022-02-08
Let `RgbVis` and new `MappedRgbVis` accept typed arrays

axelboc:


packages/lib/src/vis/rgb/RgbVis.tsx
```diff
@@ -23,41 +24,28 @@ interface Props {
...
     children,
   } = props;
 
-  const rgbValue = useMemo(
-    () =>
-      imageType === ImageType.BGR ? flipLastDimension(value, dims) : value,
-    [value, dims, imageType]
-  );
+  const values = useMemo(() => {
+    const typedDataArray = toTypedNdArray(dataArray, Uint8Array);
```
axelboc, line 36: Converting to typed ndarray here.

packages/lib/src/vis/rgb/utils.ts
```diff
@@ -1,10 +1,12 @@
 import { createArrayFromView } from '@h5web/shared';
-import ndarray from 'ndarray';
+import type { NdArray, TypedArray } from 'ndarray';
 
-export function flipLastDimension(value: number[], dims: number[]) {
-  const baseArray = ndarray(value, dims);
```
axelboc, line 5: No longer converting to `ndarray` here so I can call `toTypedNdArray` in `RgbVis` directly (and skip the conversion if the ndarray is already typed).

packages/shared/src/utils.ts
```diff
@@ -41,13 +46,14 @@ export function getValues(arr: AnyNumArray): NumArray {
 }
 
 export function toTypedNdArray<T extends NumArray>(
-  arr: NdArray<T>
-): NdArray<Exclude<T, number[]> | Float32Array> {
+  arr: NdArray<T>,
+  TypedArrayConstructor: Float32ArrayConstructor | Uint8ArrayConstructor
```
axelboc, line 50: Passing the typed array constructor as mentioned. Unfortunately, there doesn't seem to be a union type of all possible typed array constructors...

packages/shared/src/utils.ts
```diff
@@ -65,10 +71,19 @@ export function buildEntityPath(
...
+  view: NdArray<U>
+): NdArray<U> {
+  const { data } = view;
+
+  const array = ndarray(
+    // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
+    // @ts-ignore (`data.constructor` resolves to `Function`)
+    (isTypedArray(data) ? new data.constructor(view.size) : []) as U,
+    view.shape
+  );
```
axelboc, line 84: If the input ndarray is typed, `createArrayFromView` must now initialize an empty ndarray with the same typed array (and buffer length). `new data.constructor()` works perfectly fine, but was too challenging to type because of the generic.

packages/shared/src/utils.ts
loichuder: Unfortunate.

You could create the `TypedArrayConstructor` type on the same model of `TypedArray`:
```
    type TypedArrayConstructor =
        | Int8ArrayConstructor
        | Int16ArrayConstructor
        | Int32ArrayConstructor
        | Uint8ArrayConstructor
        | Uint8ClampedArrayConstructor
        | Uint16ArrayConstructor
        | Uint32ArrayConstructor
        | Float32ArrayConstructor
        | Float64ArrayConstructor;
```

packages/app/src/vis-packs/core/heatmap/utils.ts
```diff
@@ -0,0 +1,10 @@
+import type { TextureTypedArray } from '@h5web/lib/src/vis/heatmap/models';
```
loichuder, line 1: Should we expose the type `TextureTypedArray` in `@h5web/lib` to clean up the import ?

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -106,7 +106,18 @@ export function useDomains(
 
 export const useCombinedDomain = createMemo(getCombinedDomain);
 
-const useBaseArray = createMemo(getBaseArray);
```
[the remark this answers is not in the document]
loichuder: Was it the source of the typing errors you said you had in https://github.com/silx-kit/h5web/pull/950#discussion_r799393977 ?

packages/lib/src/vis/rgb/RgbVis.tsx
```diff
@@ -1,19 +1,20 @@
...
 import VisCanvas from '../shared/VisCanvas';
 import VisMesh from '../shared/VisMesh';
 import { ImageType } from './models';
 import { flipLastDimension } from './utils';
 
 interface Props {
-  value: number[];
-  dims: number[];
-  floatFormat?: boolean;
+  dataArray: NdArray<number[] | TextureTypedArray>; // if `number[]`, assumes uint8: [0, 255]
```
loichuder, line 17: I think the comment should rather be on line 36 (where the conversion takes place)

packages/lib/src/vis/rgb/utils.ts
loichuder: Great idea :+1:

loichuder (approved):

packages/app/src/vis-packs/core/hooks.ts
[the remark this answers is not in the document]
axelboc: It might have been, yes.

PR #957, opened 2022-02-09
Split PanZoom stories into individual files

loichuder:
Follow-up of #955


apps/storybook/src/VisCanvasXZoom.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const ModifierKey = Template.bind({});
+ModifierKey.args = {
+  disabled: false,
+  modifierKey: 'Shift',
+};
+
+export default {
+  ...VisCanvasStoriesConfig,
+  title: 'Building Blocks/VisCanvas/XZoom',
```
axelboc, line 46: I'd be tempted to remove the nesting under `VisCanvas`, like for `HeatmapMesh`.

axelboc (approved):
Awesome!!

apps/storybook/src/VisCanvasXZoom.stories.tsx
loichuder: Or a new section `PanZoom` ?

loichuder: I meant to replace the nesting `VisCanvas` section by a new section `PanZoom`

axelboc: Yep, sounds good!

PR #960, opened 2022-02-10
Finish dealing with typed arrays

axelboc:
Can't think of a better title for this PR 😂 

1. First, I extend the type of axis and histogram values to include typed arrays.
2. Second, I let the H5Grove and Mock providers return raw typed array instead of `number[]` from `getValue`.
3. Third, I change the way `HeatmapMesh`, `HeatmapVis` and `RgbVis` deal with typed arrays:
    - `HeatmapMesh` now accepts only `Float32Array` and `Uint16Array`, which are the only two typed arrays we can reasonably support because of WebGL. With `Float32Array`, we create a texture with `FloatType` and `RedFormat`; with `Uint16Array`, we create a texture with `HalfFloatType` and `RedFormat`. The `textureType` prop is removed to prevent incorrect uses of `HeatmapMesh` -- we always assume that a `Uint16Array` is used to display half floats.
    - `HeatmapVis` now accepts any `TypedArray` and takes care of the conversion to `Float32Array` (with utility `toTextureTypedArray()`). I no longer throw on `Float64Array` or `Uint8ClampedArray` -- they both get converted as well. There's no support for passing half floats in a `Uint16Array` at the visualization level -- users have to use `HeatmapMesh` for that.
    - `RgbVis` now accepts any `TypedArray` and takes care of the conversion to `Uint8Array` or `Float32Array` (with utility `toRgbSafeNdArray()` - cf. https://github.com/silx-kit/h5web/pull/960#discussion_r805830425).
4. Fourth, I refactor the code that puts together the shader in `HeatmapMesh`. The scaling of the domain is now done in a single utility function called `scaleDomain`. Uniforms `min`, `oneOverRange` and `gammaExponent` are now always declared and passed to the shader code (instead of conditionally based on the scale type). The alpha domain is given a default value to remove a number of annoying ternaries.
5. Fifth, I add some mock datasets for testing purposes:

image

image


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -53,7 +53,7 @@ export class H5GroveApi extends ProviderApi {
     if (DTypedArray) {
       const buffer = await this.fetchBinaryData(params);
       const array = new DTypedArray(buffer);
-      return hasScalarShape(dataset) ? array[0] : [...array];
+      return hasScalarShape(dataset) ? array[0] : array;
```
axelboc, line 56: The most important change 🎉 😁

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -136,7 +134,11 @@ export function useMappedArray<T>(
   autoScale?: boolean
 ) {
   const baseArray = useBaseArray(value, dims);
-  const mappedArray = useApplyMapping(baseArray, mapping);
+
+  const mappedArray = useMemo(
+    () => applyMapping(baseArray, mapping),
+    [baseArray, mapping]
+  );
```
axelboc, line 141: Since it's only used here and is not quite properly typed, I ended up inlining it.

packages/app/src/vis-packs/core/rgb/MappedRgbVis.tsx
```diff
@@ -18,7 +17,6 @@ interface Props {
 function MappedRgbVis(props: Props) {
   const { value, dims, title, toolbarContainer } = props;
 
-  assertTextureNumArray(value);
```
axelboc, line 21: `RgbVis` now accepts any typed array.

packages/lib/src/toolbar/controls/DomainSlider/Histogram.tsx
```diff
@@ -53,7 +53,7 @@ function Histogram(props: Props) {
   return (
     <div ref={ref} className={styles.container}>
       <svg width="100%" height="100%" className={styles.histogram}>
-        {values.map((d, i) => (
+        {(values as number[]).map((d, i) => (
```
axelboc, line 56: Typing issue with `number[] | TypedArray` 😢

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -1,159 +1,95 @@
...
     domain,
-    scaleType,
+    scaleType: visScaleType,
     colorMap,
     invertColorMap = false,
-    textureType,
     magFilter,
     alphaValues,
-    alphaDomain,
+    alphaDomain = DEFAULT_DOMAIN,
```
axelboc, line 85: The default alpha domain to remove a few ternaries.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -174,21 +110,25 @@ function HeatmapMesh(props: Props) {
...
   const shader = {
     uniforms: {
       data: { value: dataTexture },
-      withAlpha: { value: alphaValues ? 1 : 0 },
-      alpha: { value: alphaTexture },
       colorMap: { value: colorMapTexture },
-      alphaMin: { value: alphaDomain?.[0] },
+      min: { value: scaledDomain[0] },
+      oneOverRange: { value: 1 / (scaledDomain[1] - scaledDomain[0]) },
+      gammaExponent: { value: gammaExponent },
```
axelboc, line 125: These three uniforms are now passed here instead of via `scaleUniforms` (now removed). `gammeExponent` just ends up being `undefined` if the scale is not `Gamma`.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -174,21 +110,25 @@ function HeatmapMesh(props: Props) {
...
+      alpha: { value: alphaTexture },
+      withAlpha: { value: alphaValues ? 1 : 0 },
+      alphaMin: { value: alphaDomain[0] },
       oneOverAlphaRange: {
-        value: alphaDomain && 1 / (alphaDomain[1] - alphaDomain[0]),
+        value: 1 / (alphaDomain[1] - alphaDomain[0]),
       },
```
axelboc, line 131: The alpha domain is now always passed. It's not used in the shader when `withAlpha` is `0`, so it doesn't change anything. It just removes a couple of ternary/optional chaining. It was simplifying a lot more things when I had to normalize the domains when dealing with `uint8`, but I still think it's worth it. I'm hoping that the performance impact of this is negligible, but please correct me if I'm wrong and I'll revert.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -201,26 +141,27 @@ function HeatmapMesh(props: Props) {
...
+      uniform float oneOverRange;
+      uniform float gammaExponent;
+
       uniform sampler2D alpha;
+      uniform float alphaMin;
       uniform float oneOverAlphaRange;
       uniform int withAlpha;
 
       const vec4 nanColor = vec4(255, 255, 255, 1);
+      const float oneOverLog10 = 0.43429448190325176;
```
axelboc, line 155: Required for log/symlog. Again, I think the performance impact of declaring this regardless of the scale is negligible.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -201,26 +141,27 @@ function HeatmapMesh(props: Props) {
     fragmentShader: `
       uniform sampler2D data;
       uniform sampler2D colorMap;
-      uniform int scaleType;
```
axelboc, line 204: Doesn't seem to be used/passed anywhere, so must be a leftover.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -201,26 +141,27 @@ function HeatmapMesh(props: Props) {
...
 
-      ${
-        SCALE_SHADER[isScaleType(scaleType) ? scaleType : ScaleType.Gamma]
-          .fragment
-      }
+      ${SCALE_SHADER[isScaleType(scaleType) ? scaleType : ScaleType.Gamma]}
 
       void main() {
         float value = texture2D(data, coords).r;
-        bool isNotFinite = isnan(value) || isinf(value);
```
axelboc, line 221: I find `isNotFinite` quite confusing, so I just inlined the conditions. 😄

packages/lib/src/vis/heatmap/HeatmapVis.tsx
```diff
@@ -127,12 +122,12 @@ function HeatmapVis(props: Props) {
           }}
         />
         <HeatmapMesh
-          values={toTypedNdArray(dataArray, Float32Array)}
+          values={toTextureSafeNdArray(dataArray)}
```
axelboc, line 125: This is still where the conversion to `Float32Array` takes place, but now, in addition to converting `number[]`, it also converts other typed arrays.

packages/lib/src/vis/heatmap/utils.ts
```diff
@@ -40,25 +25,14 @@ export const GRADIENT_RANGE = range(
...
-  DataType<TextureTypedArray>,
-  TextureDataType
-> = {
-  int8: ByteType,
-  int16: ShortType,
-  int32: IntType,
-  uint8: UnsignedByteType,
-  uint16: UnsignedShortType,
-  uint32: UnsignedIntType,
-  float32: FloatType,
```
axelboc, line 53: We no longer need this. It's either `FloatType` or `HalfFloatType` (see further down).

packages/lib/src/vis/heatmap/utils.ts
```diff
@@ -40,25 +25,14 @@ export const GRADIENT_RANGE = range(
...
+const SCALE_FNS: Record<ScaleType, (val: number) => number> = {
+  [ScaleType.Linear]: (val) => val,
+  [ScaleType.Log]: Math.log10,
+  [ScaleType.SymLog]: (val) => Math.sign(val) * Math.log10(1 + Math.abs(val)),
+  [ScaleType.Sqrt]: Math.sqrt,
+  [ScaleType.Gamma]: (val) => val,
```
axelboc, line 33: This is where the scale functions end up. `SCALE_FNS` is then used in `scaleDomain()` below.

packages/lib/src/vis/heatmap/utils.ts
```diff
@@ -127,15 +101,15 @@ export function getLinearGradient(
...
+  rawValues: NumArray | undefined,
   pixelCount: number
 ): number[] {
   if (!rawValues) {
     return range(pixelCount + 1);
   }
 
   if (rawValues.length === pixelCount + 1) {
-    return rawValues;
+    return isTypedArray(rawValues) ? [...rawValues] : rawValues;
```
axelboc, line 112: Even though axis values are now fetched as typed arrays, I still convert them to `number[]` here. Turns out it was a bit complicated to handle typed array in `getValueToIndexScale`. Maybe I'll revisit this later on.

packages/lib/src/vis/heatmap/utils.ts
```diff
@@ -156,33 +130,27 @@ export function getInterpolator(colorMap: ColorMap, reverse: boolean) {
...
-  const format = getTextureFormatFromType(textureType);
 
   return new DataTexture(
     data,
     cols,
     rows,
-    format,
-    textureType,
+    RedFormat,
+    values.dtype === 'uint16' ? HalfFloatType : FloatType,
```
axelboc, line 153: Always `RedFormat`, combined with either `FloatType` or `HalfFloatType`, as explained.

packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -79,7 +80,11 @@ function LineVis(props: Props) {
   auxArrays.forEach((arr) => assertDataLength(arr, dataArray, 'auxiliary'));
 
   const abscissas = useMemo(() => {
-    return abscissaValue ?? range(dataArray.size);
+    if (abscissaValue) {
+      return isTypedArray(abscissaValue) ? [...abscissaValue] : abscissaValue;
```
axelboc, line 84: The conversion of axis values to `number[]` in `getAxisValues()` that I pointed out further up only applies to `HeatmapVis`. I do the same here.

packages/shared/src/guards.ts
```diff
@@ -318,7 +326,7 @@ export function assertDatasetValue<D extends Dataset<ScalarShape | ArrayShape>>(
   dataset: D
 ): asserts value is Value<D> {
   if (hasArrayShape(dataset)) {
-    assertArray(value);
+    assertArrayOrTypedArray(value);
```
axelboc, line 329: This allows providers to actually return typed arrays at runtime.

packages/shared/src/guards.ts
```diff
@@ -361,11 +369,17 @@ export function isNdArray<T extends Data>(
 }
 
 export function isTypedArray<T, U extends TypedArray>(arr: U | T[]): arr is U {
-  return isTypedArrayLodash(arr);
+  return !Array.isArray(arr);
```
axelboc, line 372: Since this guard only accepts `TypedArray | T[]`, we can assume that anything that is now a plain JS array is a typed array. This is most likely faster than lodash's `isTypedArray`, which runs a regex against the result of `toString()`.

packages/shared/src/mock/metadata.ts
```diff
@@ -56,6 +56,14 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
...
       makeDataset('fourD', intType, [3, 9, 20, 41]),
     ]),
+    makeGroup('typed_arrays', [
+      makeDataset('uint8', intType, [2, 2]),
+      makeDataset('int16', intType, [2, 2]),
+      makeDataset('float32', intType, [2, 2]),
+      makeDataset('float64', intType, [2, 2]),
+      withImageAttributes(makeDataset('uint8_rgb', intType, [2, 2, 3])),
+      withImageAttributes(makeDataset('float32_rgb', intType, [2, 2, 3])),
+    ]),
```
axelboc, line 66: Adding some mock datasets. Since all values except `float32` are converted to `float32`, I haven't bothered covering every possible typed array.

packages/shared/src/models-hdf5.ts
```diff
@@ -164,7 +164,7 @@ export type ArrayValue<T extends DType> =
 export type Value<D extends Dataset> = D['shape'] extends ScalarShape
   ? Primitive<D['type']>
   : D['shape'] extends ArrayShape
-  ? Primitive<D['type']>[]
+  ? ArrayValue<D['type']>
```
axelboc, line 167: Another very important change, that I had mentioned in my previous PR, which lets providers return typed arrays (from TS' point of view).

packages/shared/src/utils.ts
```diff
@@ -47,17 +52,21 @@ export function getValues(arr: AnyNumArray): NumArray {
...
-export function toTypedNdArray<
-  T extends NumArray,
-  U extends TypedArrayConstructor
->(arr: NdArray<T>, TypedArrayConstructor: U) {
-  if (isTypedNdArray(arr)) {
+export function toTypedNdArray<T extends TypedArrayConstructor>(
+  arr: NdArray<NumArray>,
+  Constructor: T
+): NdArray<InstanceType<T>> {
+  return ndarray(Constructor.from(arr.data) as InstanceType<T>, arr.shape);
```
axelboc, line 59: Just renamed an argument and moved the type assertion closer to where it needs to be.

packages/shared/src/utils.ts
```diff
@@ -47,17 +52,21 @@ export function getValues(arr: AnyNumArray): NumArray {
...
+}
+
+export function toTextureSafeNdArray(
+  arr: NdArray<number[] | TypedArray>
+): NdArray<Float32Array> {
+  if (isTypedNdArray(arr) && isTextureSafeNdArray(arr)) {
     return arr;
```
axelboc, line 66: In other words, if `arr` is `NdArray<Float32Array>`, do nothing. I thought the names `toTextureSafeNdArray` and `isTextureSafeNdArray` would better convey the purpose of the functions than `toFloat32NdArray` and `isFloat32NdArray`.

packages/shared/src/utils.ts
```diff
@@ -81,9 +90,9 @@ export function createArrayFromView<T, U extends TypedArray | T[]>(
   const { data } = view;
 
   const array = ndarray(
-    // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
-    // @ts-ignore (`data.constructor` resolves to `Function`)
-    (isTypedArray(data) ? new data.constructor(view.size) : []) as U,
+    (isTypedArray(data)
+      ? new (data.constructor as TypedArrayConstructor)(view.size)
+      : []) as U,
```
axelboc, line 95: I managed to avoid the `ts-ignore` with another assertion.

packages/app/src/vis-packs/core/hooks.ts
loichuder: As a side remark, it would be nice to track other instances of `createMemo` to see if they also bring improper types. Removing them would also be a step towards shying away from unmaintained `react-use`.

packages/lib/src/toolbar/controls/DomainSlider/Histogram.tsx
loichuder: :disappointed:

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -1,159 +1,95 @@
...
-  values: NdArray<TextureTypedArray>;
+  values: NdArray<Float32Array | Uint16Array>; // if `Uint16Array`, it must contain half floats
   domain: Domain;
   scaleType: VisScaleType;
   colorMap: ColorMap;
   invertColorMap?: boolean;
-  textureType?: TextureDataType; // override default texture type (determined from `values.dtype`)
   magFilter?: TextureFilter;
-  alphaValues?: NdArray<TextureTypedArray>;
+  alphaValues?: NdArray<Float32Array>;
```
loichuder, line 23: Should be allow `UInt16Array` for `alphaValues` as well ?

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
loichuder: All right, why not

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
loichuder: I don't see how this will impact performance: the `uniform` was always there even if it could be `undefined` before.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
loichuder: Good idea, it is indeed clearer

packages/lib/src/vis/heatmap/utils.ts
loichuder: Could you add some comments to sum up what motivated this choice in the end ? I fear that we might have forgotten if we need to revisit this later

packages/lib/src/vis/hooks.ts
```diff
@@ -21,9 +21,9 @@ const useValidDomainForScale = createMemo(getValidDomainForScale);
 export const useAxisDomain = createMemo(getAxisDomain);
 
 export function useDomain(
-  valuesArray: NdArray<number[]> | number[],
+  valuesArray: AnyNumArray,
```
loichuder, line 24: We ought to find a better name for this... Or perhaps removing the alias would make things clearer

packages/lib/src/vis/line/LineVis.tsx
loichuder: I would have reverted the condition to early return `range(dataArray.size)` but no big deal

packages/lib/src/vis/rgb/RgbVis.tsx
```diff
@@ -34,19 +34,18 @@ function RgbVis(props: Props) {
   } = props;
 
   const values = useMemo(() => {
-    const typedDataArray = toTypedNdArray(dataArray, Uint8Array); // if `number[]`, assume uint8: [0, 255]
+    const typedDataArray = toTextureSafeNdArray(dataArray);
```
loichuder, line 37: This introduces a regression: it removes support for `uint8` arrays. See the mandrill image in https://h5web.panosc.eu/?file=grove.h5 that is now blank.

packages/shared/src/mock/metadata.ts
loichuder: `*_rgb` datasets should have the shape `[3, 3, 3]` to display all the values in `rgb`. 

Otherwise, we end up only with the first 4 values that only includes channels that are `0` or `255` which makes us miss problems for other values.

packages/shared/src/utils.ts
loichuder: Rereading this again, I wonder: does it makes a copy of the data or not ?

packages/shared/src/utils.ts
loichuder: For `toTextureSafeNdArray`, I agree but I think renaming `isTextureSafeNdArray` to `isFloat32NdArray` would make this easier to understand

packages/shared/src/utils.ts
loichuder: Meh. No choice :shrug:

loichuder (changes requested):
Huge work, congrats :clap: 

A few nitpicks here and there but more importantly, we need to look into the support of RGB `uint8` arrays.

packages/app/src/vis-packs/core/hooks.ts
axelboc: `createMemo` is not bad in itself, it just makes it harder to add signature overrides where needed. Fortunately `useApplyMapping` was the only remaining case of a `createMemo` being used to memoise a utility with signature overrides.

packages/lib/src/vis/line/LineVis.tsx
axelboc: Sounds good :+1:

packages/shared/src/mock/metadata.ts
axelboc: Oh right, well spotted :+1:

packages/shared/src/utils.ts
axelboc: Yes, definitely. That's why I was hoping to be able to do the BGR flip in the shader at some point. But the use in `applyMapping` is worth a look as well, as I'm just realising that we always call `createArrayFromView` from there, regardless of whether the base array is sliced/mapped...

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
t20100: I guess current main (only?) use is for complex numbers, right?
If so, it is not necessary, but would make the API consistent.

packages/shared/src/utils.ts
axelboc: I've opened issue #961

packages/lib/src/vis/rgb/RgbVis.tsx
t20100: Looks like supporting `uint8` normalized texture is useful here.

packages/lib/src/vis/heatmap/utils.ts
axelboc: Great point. I've added a few paragraphs to explain the reasoning.

packages/lib/src/vis/hooks.ts
axelboc: Just to clarify, this is the type definition of `AnyNumArray` (not just `NdArray<number[]> | number[]`):

```ts
export type NumArray = TypedArray | number[];
export type AnyNumArray = NdArray<NumArray> | NumArray;
```

I find the term "any" quite accurate, since it really allows any kind of numeric array: typed array, JS `number` array, ndarray (with typed array of JS `number` array). There are 12 occurrences of it throughout the codebase, which makes it quite useful in my opinion.

packages/lib/src/vis/rgb/RgbVis.tsx
axelboc: Indeed, same problem with the grey that was cropped by mistake in the mock datasets.

axelboc: At least `mandrill_float` works now :joy:

axelboc: I've updated the PR description.

packages/lib/src/vis/heatmap/utils.ts
```diff
@@ -127,15 +101,15 @@ export function getLinearGradient(
 }
 
 export function getAxisValues(
-  rawValues: number[] | undefined,
+  rawValues: NumArray | undefined,
   pixelCount: number
 ): number[] {
```
loichuder, line 106: Could we return `NumArray` here ?

packages/lib/src/vis/heatmap/utils.ts
loichuder, line 112: ...avoiding the ternary here

packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -79,7 +80,11 @@ function LineVis(props: Props) {
   auxArrays.forEach((arr) => assertDataLength(arr, dataArray, 'auxiliary'));
 
   const abscissas = useMemo(() => {
-    return abscissaValue ?? range(dataArray.size);
+    if (!abscissaValue) {
+      return range(dataArray.size);
+    }
+
+    return isTypedArray(abscissaValue) ? [...abscissaValue] : abscissaValue;
```
loichuder, line 87: Same as above: what prevents us from keeping `abscissas` a `NumArray` ?

packages/lib/src/vis/rgb/utils.ts
```diff
@@ -1,5 +1,37 @@
...
+
+  if (ndArr.dtype === 'int8') {
+    return ndarray(
+      Uint8Array.from([...ndArr.data].map((val) => val + 128)),
+      ndArr.shape
+    );
+  }
+
+  return toTypedNdArray(ndArr, Uint8Array);
+}
```
loichuder, line 34: Great :+1:

loichuder (approved):

PR #964, opened 2022-02-14
Support raw uint8 values in `HeatmapMesh`

axelboc:
Very elegant solution @t20100, thanks!

image

I took the opportunity to refactor the shader's scale functions a bit further.

### DIFF
--- a/packages/lib/src/vis/heatmap/HeatmapMesh.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapMesh.tsx
@@ -1,5 +1,5 @@
 import type { Domain } from '@h5web/shared';
-import { ScaleType, isScaleType } from '@h5web/shared';
+import { ScaleType } from '@h5web/shared';
 import { rgb } from 'd3-color';
 import type { NdArray } from 'ndarray';
 import { memo, useMemo } from 'react';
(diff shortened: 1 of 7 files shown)


packages/lib/src/vis/heatmap/utils.ts
```diff
@@ -138,22 +152,27 @@ export function scaleDomain(
   return [scaleFn(domain[0]), scaleFn(domain[1])];
 }
 
+export function toTextureSafeNdArray(
+  arr: NdArray<NumArray>
+): NdArray<TextureSafeTypedArray> {
+  if (arr.dtype === 'float32' || arr.dtype.startsWith('uint8')) {
+    return arr as NdArray<TextureSafeTypedArray>;
```
axelboc, line 159: Seemed over the top to add a type guard to check specifically for these dtypes like I had done before for `float32`...

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -123,6 +89,7 @@ function HeatmapMesh(props: Props) {
       min: { value: scaledDomain[0] },
       oneOverRange: { value: 1 / (scaledDomain[1] - scaledDomain[0]) },
       gammaExponent: { value: gammaExponent },
+      normRevertFactor: { value: values.dtype === 'uint8' ? 255 : 1 }, // to revert automatic normalization
```
loichuder, line 92: It might be useful to specify what does the normalization in the first place (is it three.js with a particular texture format ?)

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -10,67 +10,33 @@ import { useAxisSystemContext } from '../..';
...
-    bool isSupported(float value) {
-      return true;
-    }`,
+const IMPL_SCALE: Record<ScaleType, string> = {
+  [ScaleType.Linear]: 'norm(value)',
+  [ScaleType.Log]: 'norm(log(value) * oneOverLog10)',
+  [ScaleType.SymLog]: 'norm(sign(value) * log(1. + abs(value)) * oneOverLog10)',
+  [ScaleType.Sqrt]: 'norm(sqrt(value))',
+  [ScaleType.Gamma]: 'pow(norm(value), gammaExponent)',
+};
```
loichuder, line 35: Not a huge fan of the refactoring, to be honest.

Yes, it is way shorter but it relies on knowing how the template string is interpolated in the shader below instead of using abstractions as before.

I can live with it, though.

loichuder (approved):

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
axelboc: As suggested by Thomas, `gammaExponent` now defaults to `1` and the gamma correction is always applied. This means we no longer need the `norm` function and can simplify this record even more with only the relevant scaling code.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -110,26 +74,25 @@ function HeatmapMesh(props: Props) {
...
-      oneOverRange: { value: 1 / (scaledDomain[1] - scaledDomain[0]) },
-      gammaExponent: { value: gammaExponent },
-      alpha: { value: alphaTexture },
-      withAlpha: { value: alphaValues ? 1 : 0 },
-      alphaMin: { value: alphaDomain[0] },
-      oneOverAlphaRange: {
-        value: 1 / (alphaDomain[1] - alphaDomain[0]),
-      },
-    },
+    uniforms: getUniforms({
```
axelboc, line 84: New `getUniforms` utility to avoid having to pass objects with a `value` key.

PR #967, opened 2022-02-15
Add tiles image with multiple level of details layer support

t20100:
This PR adds a component to render (large) images as tiles with pyramid of level-of-details image support.

The axes are used to compute the level of detail to use rather than the size of the array, this needs to be solved.

~~commit https://github.com/silx-kit/h5web/commit/cc293bc86d51e0d37a991e23a746280d278f908c provides a proof of concept for h5grove-based fetch of subsampled slices. It is there for demo and should be removed before merging.~~

### DIFF
--- a/apps/storybook/src/TiledHeatmap.stories.tsx
+++ b/apps/storybook/src/TiledHeatmap.stories.tsx
@@ -118,8 +118,7 @@ class MandelbrotTilesApi extends TilesApi {
 }
 const Template: Story<TiledHeatmapProps> = (args) => {
   const { api, ...tiledHeatmapProps } = args;
-  const [layer0Size] = api.layerSizes;
-  const { width, height } = layer0Size;
+  const { width, height } = api.imageSize;
 
   return (
     <VisCanvas
--- a/apps/storybook/src/TiledHeatmap.stories.tsx
+++ b/apps/storybook/src/TiledHeatmap.stories.tsx
@@ -59,7 +59,7 @@ function mandelbrot(
   return array;
 }
 
-interface TileKey {
+interface TileParams {
   lod: number;
   offset: Vector2;
 }
@@ -80,7 +80,7 @@ class MandelbrotTilesApi extends TilesApi {
     this.yDomain = yDomain;
 
     this.store = createFetchStore(
-      async (tile: TileKey) => {
+      async (tile: TileParams) => {
         const { lod, offset } = tile;
         const layerSize = this.layerSizes[lod];
         // Clip slice to size of the level
@@ -106,7 +106,7 @@ class MandelbrotTilesApi extends TilesApi {
       },
       {
         type: 'Map',
-        areEqual: (a: TileKey, b: TileKey) =>
+        areEqual: (a: TileParams, b: TileParams) =>
           a.lod === b.lod && a.offset.equals(b.offset),
       }
     );
--- a/apps/storybook/src/TiledHeatmap.stories.tsx
+++ b/apps/storybook/src/TiledHeatmap.stories.tsx
@@ -1,0 +1,182 @@
+import {
+  getLayerSizes,
+  PanMesh,
+  VisCanvas,
+  ZoomMesh,
+  TiledHeatmap,
+  TilesApi,
+} from '@h5web/lib';
+import type { Domain, Size, TiledHeatmapProps } from '@h5web/lib';
+import { ScaleType } from '@h5web/shared';
+import type { Meta, Story } from '@storybook/react/types-6-0';
+import { clamp } from 'lodash';
+import ndarray from 'ndarray';
+import type { NdArray } from 'ndarray';
+import { createFetchStore } from 'react-suspense-fetch';
+import type { Vector2 } from 'three';
+
+import FillHeight from './decorators/FillHeight';
+
+// See https://en.wikipedia.org/wiki/Mandelbrot_set
+function mandelbrot(
+  iterations: number,
+  xDomain: Domain,
+  yDomain: Domain,
+  size: Size
+): NdArray<Float32Array> {
+  const { width, height } = size;
+  const xRange = xDomain[1] - xDomain[0];
+  const yRange = yDomain[1] - yDomain[0];
+
+  const array = ndarray(new Float32Array(height * width), [height, width]);
+
+  for (let row = 0; row < height; row += 1) {
+    const cImag = yDomain[0] + yRange * (row / (height - 1));
+    for (let col = 0; col < width; col += 1) {
+      let value = 0;
+
+      const cReal = xDomain[0] + xRange * (col / (width - 1));
+      // z = c
+      let zReal = cReal;
+      let zImag = cImag;
+      for (let index = 0; index <= iterations; index += 1) {
+        // z = z**2 + c
+        const zRealSq = zReal ** 2;
+        const zImagSq = zImag ** 2;
+        zImag = 2 * zReal * zImag + cImag;
+        zReal = zRealSq - zImagSq + cReal;
+
+        // check divergence
+        if (zRealSq + zImagSq > 4) {
+          value = index / iterations;
+          break;
+        }
+      }
+
+      array.set(row, col, value);
+    }
+  }
+  return array;
+}
+
+interface TileKey {
+  lod: number;
+  offset: Vector2;
+}
+
+class MandelbrotTilesApi extends TilesApi {
+  public readonly xDomain: Domain;
+  public readonly yDomain: Domain;
+  private readonly store;
+
+  public constructor(
+    size: Size,
+    tileSize: Size,
+    xDomain: Domain,
+    yDomain: Domain
+  ) {
+    super(tileSize, getLayerSizes(size, tileSize));
+    this.xDomain = xDomain;
+    this.yDomain = yDomain;
+
+    this.store = createFetchStore(
+      async (tile: TileKey) => {
+        const { lod, offset } = tile;
+        const layerSize = this.layerSizes[lod];
+        // Clip slice to size of the level
+        const width = clamp(layerSize.width - offset.x, 0, this.tileSize.width);
+        const height = clamp(
+          layerSize.height - offset.y,
+          0,
+          this.tileSize.height
+        );
+
+        const xScale = (this.xDomain[1] - this.xDomain[0]) / layerSize.width;
+        const xRange: Domain = [
+          this.xDomain[0] + xScale * offset.x,
+          this.xDomain[0] + xScale * (offset.x + width),
+        ];
+        const yScale = (this.yDomain[1] - this.yDomain[0]) / layerSize.height;
+        const yRange: Domain = [
+          this.yDomain[0] + yScale * offset.y,
+          this.yDomain[0] + yScale * (offset.y + height),
+        ];
+
+        return mandelbrot(50, xRange, yRange, { width, height });
+      },
+      {
+        type: 'Map',
+        areEqual: (a: TileKey, b: TileKey) =>
+          a.lod === b.lod && a.offset.equals(b.offset),
+      }
+    );
+  }
+
+  public get(lod: number, offset: Vector2): NdArray<Float32Array> {
+    return this.store.get({ lod, offset });
+  }
+}
+const Template: Story<TiledHeatmapProps> = (args) => {
+  const { api, ...tiledHeatmapProps } = args;
+  const [layer0Size] = api.layerSizes;
+  const { width, height } = layer0Size;
+
+  return (
+    <VisCanvas
+      abscissaConfig={{
+        visDomain: [0, width],
+        isIndexAxis: true,
+        showGrid: false,
+      }}
+      ordinateConfig={{
+        visDomain: [0, height],
+        isIndexAxis: true,
+        showGrid: false,
+      }}
+      visRatio={width / height}
+    >
+      <PanMesh />
+      <ZoomMesh />
+      <TiledHeatmap api={api} {...tiledHeatmapProps} />
+    </VisCanvas>
+  );
+};
+
+export const Default = Template.bind({});
+Default.args = {
+  api: new MandelbrotTilesApi(
+    { width: 1e9, height: 1e9 },
+    { width: 128, height: 128 },
+    [-2, 1],
+    [-1.5, 1.5]
+  ),
+  domain: [0, 1],
+};
+
+export default {
+  title: 'Building Blocks/TiledHeatmap',
+  component: TiledHeatmap,
+  decorators: [FillHeight],
+  parameters: { layout: 'fullscreen', controls: { sort: 'requiredFirst' } },
+  args: {
+    invertColorMap: false,
+    colorMap: 'Viridis',
+    scaleType: ScaleType.Linear,
+    displayLowerResolutions: true,
+    qualityFactor: 1,
+  },
+  argTypes: {
+    scaleType: {
+      control: { type: 'inline-radio' },
+      options: [
+        ScaleType.Linear,
+        ScaleType.Log,
+        ScaleType.SymLog,
+        ScaleType.Sqrt,
+      ],
+    },
+    qualityFactor: {
+      control: { type: 'range', min: 0, max: 1, step: 0.1 },
+    },
+  },
+} as Meta;
--- a/apps/storybook/src/TiledHeatmap.stories.tsx
+++ b/apps/storybook/src/TiledHeatmap.stories.tsx
@@ -60,7 +60,7 @@ function mandelbrot(
 }
 
 interface TileParams {
-  lod: number;
+  layer: number;
   offset: Vector2;
 }
 
@@ -81,8 +81,8 @@ class MandelbrotTilesApi extends TilesApi {
 
     this.store = createFetchStore(
       async (tile: TileParams) => {
-        const { lod, offset } = tile;
-        const layerSize = this.layerSizes[lod];
+        const { layer, offset } = tile;
+        const layerSize = this.layerSizes[layer];
         // Clip slice to size of the level
         const width = clamp(layerSize.width - offset.x, 0, this.tileSize.width);
         const height = clamp(
@@ -107,13 +107,13 @@ class MandelbrotTilesApi extends TilesApi {
       {
         type: 'Map',
         areEqual: (a: TileParams, b: TileParams) =>
-          a.lod === b.lod && a.offset.equals(b.offset),
+          a.layer === b.layer && a.offset.equals(b.offset),
       }
     );
   }
 
-  public get(lod: number, offset: Vector2): NdArray<Float32Array> {
-    return this.store.get({ lod, offset });
+  public get(layer: number, offset: Vector2): NdArray<Float32Array> {
+    return this.store.get({ layer, offset });
   }
 }
 const Template: Story<TiledHeatmapProps> = (args) => {
--- a/apps/storybook/src/TiledHeatmap.stories.tsx
+++ b/apps/storybook/src/TiledHeatmap.stories.tsx
@@ -118,7 +118,7 @@ class MandelbrotTilesApi extends TilesApi {
 }
 const Template: Story<TiledHeatmapProps> = (args) => {
   const { api, ...tiledHeatmapProps } = args;
-  const { width, height } = api.imageSize;
+  const { width, height } = api.baseLayerSize;
 
   return (
     <VisCanvas
--- a/packages/lib/src/vis/heatmap/HeatmapMesh.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapMesh.tsx
@@ -7,7 +7,7 @@ import type { TextureFilter } from 'three';
 import { DataTexture, RGBAFormat, UnsignedByteType } from 'three';
 
 import { useAxisSystemContext } from '../..';
-import type { VisScaleType } from '../models';
+import type { Size, VisScaleType } from '../models';
 import VisMesh from '../shared/VisMesh';
 import { DEFAULT_DOMAIN, getUniforms, VERTEX_SHADER } from '../utils';
 import type { ColorMap, TextureSafeTypedArray } from './models';
@@ -22,6 +22,7 @@ interface Props {
   magFilter?: TextureFilter;
   alphaValues?: NdArray<TextureSafeTypedArray | Uint16Array>; // uint16 values are treated as half floats
   alphaDomain?: Domain;
+  size?: Size;
 }
 
 const CMAP_SIZE = 256;
@@ -47,6 +48,7 @@ function HeatmapMesh(props: Props) {
     magFilter,
     alphaValues,
     alphaDomain = DEFAULT_DOMAIN,
+    size,
   } = props;
 
   const { ordinateConfig } = useAxisSystemContext();
@@ -141,7 +143,7 @@ function HeatmapMesh(props: Props) {
   };
 
   return (
-    <VisMesh scale={[1, ordinateConfig.flip ? -1 : 1, 1]}>
+    <VisMesh scale={[1, ordinateConfig.flip ? -1 : 1, 1]} size={size}>
       <shaderMaterial args={[shader]} />
     </VisMesh>
   );
--- a/packages/lib/src/vis/shared/VisMesh.tsx
+++ b/packages/lib/src/vis/shared/VisMesh.tsx
@@ -1,18 +1,25 @@
 import type { MeshProps } from '@react-three/fiber';
 
+import type { Size } from '../models';
 import { useAxisSystemContext } from './AxisSystemContext';
 
-function VisMesh(props: MeshProps) {
-  const { children, ...meshProps } = props;
+interface Props extends MeshProps {
+  size?: Size;
+}
+
+function VisMesh(props: Props) {
+  const { children, size, ...meshProps } = props;
 
   const { visSize } = useAxisSystemContext();
+  const { width, height } = size ?? visSize;
 
   return (
     <mesh {...meshProps}>
-      <planeGeometry args={[visSize.width, visSize.height]} />
+      <planeGeometry args={[width, height]} />
       {children}
     </mesh>
   );
 }
 
+export type { Props as VisMeshProps };
 export default VisMesh;
--- a/packages/lib/src/vis/tiles/Tile.tsx
+++ b/packages/lib/src/vis/tiles/Tile.tsx
@@ -1,0 +1,36 @@
+import { memo } from 'react';
+import { Vector2 } from 'three';
+import type { TextureFilter } from 'three';
+
+import HeatmapMesh from '../heatmap/HeatmapMesh';
+import { useAxisSystemContext } from '../shared/AxisSystemContext';
+import type { TilesApi } from './api';
+import type { ColorMapProps } from './models';
+
+interface Props extends ColorMapProps {
+  api: TilesApi;
+  lod: number;
+  x: number;
+  y: number;
+  magFilter: TextureFilter;
+}
+
+function Tile(props: Props) {
+  const { api, lod, x, y, magFilter, ...colorMapProps } = props;
+
+  const { visSize } = useAxisSystemContext();
+
+  const array = api.get(lod, new Vector2(x, y));
+  const [height, width] = array.shape;
+
+  return (
+    <group
+      position={[x + width / 2, y + height / 2, 0]} // Tile center
+      scale={[width / visSize.width, height / visSize.height, 1]}
+    >
+      <HeatmapMesh values={array} {...colorMapProps} magFilter={magFilter} />
+    </group>
+  );
+}
+
+export default memo(Tile);
--- a/packages/lib/src/vis/tiles/Tile.tsx
+++ b/packages/lib/src/vis/tiles/Tile.tsx
@@ -8,16 +8,16 @@ import type { ColorMapProps } from './models';
 
 interface Props extends ColorMapProps {
   api: TilesApi;
-  lod: number;
+  layer: number;
   x: number;
   y: number;
   magFilter: TextureFilter;
 }
 
 function Tile(props: Props) {
-  const { api, lod, x, y, magFilter, ...colorMapProps } = props;
+  const { api, layer, x, y, magFilter, ...colorMapProps } = props;
 
-  const array = api.get(lod, new Vector2(x, y));
+  const array = api.get(layer, new Vector2(x, y));
   const [height, width] = array.shape;
 
   return (
--- a/packages/lib/src/vis/tiles/Tile.tsx
+++ b/packages/lib/src/vis/tiles/Tile.tsx
@@ -3,7 +3,6 @@ import { Vector2 } from 'three';
 import type { TextureFilter } from 'three';
 
 import HeatmapMesh from '../heatmap/HeatmapMesh';
-import { useAxisSystemContext } from '../shared/AxisSystemContext';
 import type { TilesApi } from './api';
 import type { ColorMapProps } from './models';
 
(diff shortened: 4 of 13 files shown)


apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -0,0 +1,322 @@
...
+      } else {
+        levelsSize.push({
+          width: 1 + Math.floor((size.width - 1) / 2 ** level),
+          height: 1 + Math.floor((size.height - 1) / 2 ** level),
+        });
+      }
+    } while (
+      levelsSize[levelsSize.length - 1].width > tileSize.width &&
+      levelsSize[levelsSize.length - 1].height > tileSize.height
+    );
```
t20100, line 39: To be moved in `TilesApi` as a default init of sizes

t20100: done

apps/storybook/src/TiledHeatmap.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+    size: Size,
+    tileSize: Size,
+    xDomain: Domain,
+    yDomain: Domain
+  ) {
+    super(tileSize, getLevelsSize(size, tileSize));
+    this.xDomain = xDomain;
+    this.yDomain = yDomain;
+
+    this.store = createFetchStore(
```
t20100, line 80: I'm wondering whether the store should be part of the `api` or integrated in the `TiledHeatmap` component.

packages/lib/src/vis/tiles/models.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import type { ColorMap } from '../heatmap/models';
+import type { VisScaleType } from '../models';
+
+export interface ColorMapAttributes {
+  domain: Domain;
+  scaleType: VisScaleType;
+  colorMap: ColorMap;
+  invertColorMap?: boolean;
+}
```
t20100, line 11: Should this be shared at a large scope?

t20100:
Only the mandelbrot story remains and it now uses a store which changes the game!

The main limitation of the current implementation is that the level of details is computed from the axes rather than from the array directly, so axes range must be x: [0, width] and y: [0, height] for it to work.

apps/storybook/src/TiledHeatmap.stories.tsx
axelboc: Hmm yeah, I see your point. That's why, in `@h5web/app`, we have a `Provider` that abstracts the creation of the stores with the methods of the `ProviderAPI` implementations. I'm not too keen on using `react-suspense-fetch` in @h5web/lib and creating the stores in `TiledHeatmap`, though, as I feel like caching should remain the responsibility of the consumer (or another package maybe).

Let's leave the store creation in the Storybook for now.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,76 @@
...
+      scale={[
+        visSize.width / level0Size.width,
+        visSize.height / level0Size.height,
+        0,
+      ]}
+    >
+      <mesh position={[level0Size.width / 2, level0Size.height / 2, -0.1]}>
+        <planeGeometry args={[level0Size.width, level0Size.height]} />
+        <meshBasicMaterial
+          color={new Color(getInterpolator(colorMap, invertColorMap)(0))}
```
loichuder, line 59: Correct me if I am wrong, but I don't think the `Color` constructor is needed ?

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,76 @@
...
+        />
+      </mesh>
```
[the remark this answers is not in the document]
t20100: Indeed

packages/lib/src/vis/tiles/TiledHeatmap.tsx
t20100: typescript complains if `new` is removed

t20100:
> I will look into the algorithm next: I am not too convinced about the nested position/scaling but I need time to think about it

I reworked the group's scale/position so it is now managed in `TiledHeatmap` but for the positioning of `Tile`. Hopefully simpler...

I'm wondering whether TileHeatmapLayer should better be merged into `TileHeatmap`?

packages/lib/src/vis/tiles/Tile.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  const { visSize } = useAxisSystemContext();
+
+  const array = api.get(lod, new Vector2(x, y));
+  const [height, width] = array.shape;
+
+  return (
+    <group
+      position={[x + width / 2, y + height / 2, 0]}
+      scale={[width / visSize.width, height / visSize.height, 1]}
```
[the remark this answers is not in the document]
loichuder: If @axelboc agress, I think it would be fine to integrate it to this PR: it is not about adding a feature but streamlining your changes.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
loichuder: Sorry I wasn't clear: I mean that doing `color={getInterpolator(colorMap, invertColorMap)(0)}` might be enough (no need to wrap it in a `Color` object)

packages/lib/src/vis/tiles/Tile.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { useAxisSystemContext } from '../shared/AxisSystemContext';
+import type { TilesApi } from './api';
+import type { ColorMapAttributes } from './models';
+
+interface Props extends ColorMapAttributes {
+  api: TilesApi;
+  lod: number;
+  x: number;
+  y: number;
+  magFilter?: TextureFilter;
```
loichuder, line 14: As you specify `magFilter` for each call of `Tile`, I would not bother to make it an optional with a default value that is never used.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const nLevels = api.levelsSize.length;
+  const [level0Size] = api.levelsSize;
+
+  const quality = 1 - clamp(qualityFactor, 0, 1);
+  const levelOfDetail = Math.min(
+    Math.floor(Math.log2(Math.max(1, xDataPerPixel, yDataPerPixel)) + quality),
+    Math.max(nLevels - 1, 0)
+  );
+
+  const lods = progressive
```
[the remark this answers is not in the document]
loichuder: Nice comment ! Reading it, I wonder if we could remove it by renaming `progressive` into `displayLowerResolutions` ? What do you think ?

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,82 @@
...
+    ? range(levelOfDetail, nLevels).reverse()
+    : [levelOfDetail];
+
+  const { colorMap, invertColorMap = false } = colorMapAttrs;
+
+  useFrameRendering();
+
+  return (
+    <group>
```
loichuder, line 51: You could get away with `<></>` instead of `<group>` here but up to you

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,82 @@
...
+            -visSize.width / 2,
+            -visSize.height / 2,
+            (nLevels - lod) / (nLevels + 1),
+          ]}
+          scale={[
+            visSize.width / levelsSize[lod].width,
+            visSize.height / levelsSize[lod].height,
+            0,
+          ]}
+        >
```
loichuder, line 73: To me, it would be clearer if the `position` and `scale` were set into `TiledHeatmapLayer`, closer to where you really need it. 

Removing the group would also allow to simplify the `map` into
```jsx
{lods.map((lod) => (<TiledHeatmapLayer key={lod} api={api} lod={lod} {...colorMapAttrs} />)}
```

packages/lib/src/vis/tiles/TiledHeatmapLayer.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const center = new Vector2(0.5 * (xOrigin + xEnd), 0.5 * (yOrigin + yEnd));
+  tileOffsets.sort((a, b) => {
+    return (
+      tileToCenterSquareDistance(a, center, tileSize) -
+      tileToCenterSquareDistance(b, center, tileSize)
+    );
+  });
+
+  return (
+    <group>
```
loichuder, line 44: To be clear, this is where I propose to set `position` and `scale` in my previous comment

packages/lib/src/vis/tiles/TiledHeatmapLayer.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+interface Props extends ColorMapAttributes {
+  api: TilesApi;
+  lod: number;
+}
+
+function TiledHeatmapLayer(props: Props) {
+  const { api, lod, ...colorMapAttrs } = props;
+
+  const scale = 2 ** lod;
```
loichuder, line 18: This is coupled on how you compute `levelSizes` in `getLevelsSize` **but is in no way enforced**. This is quite problematic as the API and this component could end up computing `scale` differently.

One solution would be to make the scale computation a method of the `api`. Perhaps do you have a better idea ?

packages/lib/src/vis/tiles/TiledHeatmapLayer.tsx
```diff
@@ -0,0 +1,61 @@
...
+  );
+
+  // Sort tiles from closest to vis center to farthest away
+  const center = new Vector2(0.5 * (xOrigin + xEnd), 0.5 * (yOrigin + yEnd));
+  tileOffsets.sort((a, b) => {
+    return (
+      tileToCenterSquareDistance(a, center, tileSize) -
+      tileToCenterSquareDistance(b, center, tileSize)
+    );
+  });
```
loichuder, line 41: You could make a function `sortByDistanceFromCenter` to inline this (including the comment) but no pressure :smile:

packages/lib/src/vis/tiles/api.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import type { TextureTypedArray } from '../heatmap/models';
+import type { Size } from '../models';
+
+export function getLevelsSize(
+  size: Size,
+  tileSize: Size,
+  evenLevelSize = false
+): Size[] {
+  const levelsSize = [] as Size[];
```
loichuder, line 12:
```suggestion
  const levelsSize: Size[] = [];
```

(replace cast by type declaration)

packages/lib/src/vis/tiles/hooks.ts
```diff
@@ -0,0 +1,13 @@
+import { useThree } from '@react-three/fiber';
+
+import { useVisibleDomains } from '../hooks';
+
+export function useDataPerPixel() {
+  const { xVisibleDomain, yVisibleDomain } = useVisibleDomains();
```
loichuder, line 6: We will discuss in the next PR but adding support for non-index coordinates would only need a call to `useValueToIndexScale` here to transform visible _data_ domains into visible _index_ domains.

But as I said, for later !

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  return tilesOffset;
+}
+
+export function tileToCenterSquareDistance(
+  offset: Vector2,
+  center: Vector2,
+  tileSize: Size
+): number {
+  return (
+    (offset.x + tileSize.width / 2 - center.x) ** 2 +
```
loichuder, line 41: Just an idea: would it make things easier to work with `offset` being the center of the tile rather than the corner ? 

This would avoid the `+tileSize.width / 2` at several places.

loichuder (approved):
Second round ! :boxing_glove: 

A few improvement proposals but nothing preventing merging for me

packages/lib/src/vis/tiles/utils.ts
t20100: Probably, I'll try it.
I'm used to place images with the lower-left corner, that's why I did that, but since three uses the center, it should indeed be better.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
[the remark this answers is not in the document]
t20100: Sounds good! I left the comment for now.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
t20100: Good to know!

packages/lib/src/vis/tiles/TiledHeatmap.tsx
t20100: OK, done!

packages/lib/src/vis/tiles/TiledHeatmapLayer.tsx
t20100: Indeed.
By moving group's scale and position in `TiledHeatmapLayer`, it is possible to use the right scale here computed from the level's size.

packages/lib/src/vis/tiles/utils.ts
t20100: There are some issues with it:
- The API `get` function works with an offset and to me it is best this way because it maps easily to array slicing... so back to `-tileSize.width/2`
- The tiles on the edges are smaller than `tileSize`, so, while the `offset` remains the same, the `center` changes. This specific case would need to be handled when generating the tile centers. The current way to handle this is to use the size of the received array rather than the tile size.

In the end, it removes some offsets at some places, but adds them at other places.

packages/lib/src/vis/tiles/api.ts
```diff
@@ -0,0 +1,50 @@
+import type { NdArray } from 'ndarray';
+import type { Vector2 } from 'three';
+
+import type { TextureSafeTypedArray } from '../heatmap/models';
+import type { Size } from '../models';
+
+export function getLevelsSize(
+  size: Size,
+  tileSize: Size,
+  evenLevelSize = false
```
t20100, line 10: This would better be called `roundToEven` since level size can be odd when `evenLevelSize = true`

apps/storybook/src/TiledHeatmap.stories.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+import FillHeight from './decorators/FillHeight';
+
+// See https://en.wikipedia.org/wiki/Mandelbrot_set
+function mandelbrot(
+  iterations: number,
+  xDomain: Domain,
+  yDomain: Domain,
+  size: Size
+) {
```
axelboc, line 26: When it's easy-enough to do, it's usually good to add an explicit return type to help TS (and readability).

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <>
+      <mesh position={[0, 0, -0.1]}>
+        <planeGeometry args={[visSize.width, visSize.height]} />
+        <meshBasicMaterial
+          color={getInterpolator(colorMap, invertColorMap)(0)}
+        />
+      </mesh>
+      {lods.map((lod) => (
+        <TiledHeatmapLayer
+          key={`${lod}`}
```
loichuder, line 57:
```suggestion
          key={lod}
```
is enough. No need to stringify it.

packages/lib/src/vis/tiles/TiledHeatmapLayer.tsx
```diff
@@ -0,0 +1,72 @@
...
+  sortTilesByDistanceTo(tileOffsets, tileSize, center);
+
+  return (
+    <group
+      position={[
+        -visSize.width / 2,
+        -visSize.height / 2,
+        (nLevels - lod) / (nLevels + 1),
+      ]}
+      scale={[visSize.width / width, visSize.height / height, 1]}
```
loichuder, line 54: I think it is worth keeping the comment saying that this is needed to work in lod coordinates

apps/storybook/src/TiledHeatmap.stories.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const { width, height } = size;
+
+  const array = ndarray(new Float32Array(height * width), [height, width]);
+  array.data.fill(0);
```
axelboc, line 30: To avoid looping through the array twice, could `0` be set as part of the `for` loops below? Maybe before the third nested `for` loop?

You could also put the two domain ranges in constants to save some subtractions.

apps/storybook/src/TiledHeatmap.stories.tsx
```diff
@@ -0,0 +1,208 @@
...
+      },
+      {
+        type: 'Map',
+        areEqual: (a: TileKey, b: TileKey) =>
+          a.lod === b.lod && a.offset.equals(b.offset),
+      }
+    );
+  }
+
+  public get(lod: number, offset: Vector2) {
```
axelboc, line 111: return type here as well

packages/lib/src/vis/tiles/utils.ts
loichuder: Nevermind then. Do as you see fit

apps/storybook/src/TiledHeatmap.stories.tsx
```diff
@@ -0,0 +1,208 @@
...
+        invertColorMap={invertColorMap}
+        displayLowerResolutions={displayLowerResolutions}
+        qualityFactor={qualityFactor}
+      />
+    </VisCanvas>
+  );
+}
+
+const Template: Story<TiledHeatmapVisProps> = (args) => {
+  return <TiledHeatmapVis {...args} />;
```
axelboc, line 168: You can just inline the `TiledHeatmapVis` component above in here. No need for the extra layer.

You can also import `TiledHeatmapProps` from the `TileHeatmap` component instead of duplicating the interface here.

apps/storybook/src/TiledHeatmap.stories.tsx
```diff
@@ -0,0 +1,208 @@
+import {
+  getLevelsSize,
```
axelboc, line 2: As per our earlier discussion, I feel like this should be called `getLevelSizes`, since it returns `Size[]`. Same with the `levelsSize` instance variables/props/... throughout.

packages/lib/src/vis/tiles/api.ts
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      levelsSize.push({
+        width: 1 + Math.floor((size.width - 1) / scale),
+        height: 1 + Math.floor((size.height - 1) / scale),
+      });
+    }
+  } while (
+    levelsSize[levelsSize.length - 1].width > tileSize.width &&
+    levelsSize[levelsSize.length - 1].height > tileSize.height
+  );
```
axelboc, line 29: It'd be nice to do this recursively, just sayin' :grin: -- it can wait though :wink:

packages/lib/src/vis/tiles/api.ts
```diff
@@ -0,0 +1,50 @@
+import type { NdArray } from 'ndarray';
+import type { Vector2 } from 'three';
+
+import type { TextureSafeTypedArray } from '../heatmap/models';
+import type { Size } from '../models';
+
+export function getLevelsSize(
+  size: Size,
```
axelboc, line 8: Since we're dealing with two different sizes, I would maybe clarify that this is the `imageSize` (or `rootSize` or something if you end up doing it recursively).

packages/lib/src/vis/tiles/api.ts
```diff
@@ -0,0 +1,50 @@
...
+  public readonly tileSize: Size;
+  protected levelsSizeArray: Size[];
+
+  public constructor(tileSize: Size, levelsSize: Size[]) {
+    this.tileSize = tileSize;
+    this.levelsSizeArray = [...levelsSize];
+  }
+
+  public get levelsSize() {
+    return [...this.levelsSizeArray];
```
axelboc, line 43: I wouldn't bother with this getter and with copying the array in the constructor. Just mark it as `readonly` and TS should prevent anyone from modifying it (I forgot about this feature when you asked me about freezing objects a while back).

apps/storybook/src/TiledHeatmap.stories.tsx
axelboc: Now that I think of it, I find the term "level" ambiguous. I assume it comes from "level of detail"? Here, I feel like the term "layer" would make more sense, so `getLayerSizes` and `layerSizes`.

apps/storybook/src/TiledHeatmap.stories.tsx
axelboc: (You can keep destructuring the story args, though: `{...args}`)

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import { getInterpolator } from '../heatmap/utils';
+import { useFrameRendering } from '../hooks';
+import { useAxisSystemContext } from '../shared/AxisSystemContext';
+import TiledHeatmapLayer from './TiledHeatmapLayer';
+import type { TilesApi } from './api';
+import { useDataPerPixel } from './hooks';
+import type { ColorMapAttributes } from './models';
+
+interface Props extends ColorMapAttributes {
```
axelboc, line 11: Maybe just `ColorMapProps` and `...colorMapProps` below? The term "attributes" already refers to HDF5 and HTML attributes.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const { visSize } = useAxisSystemContext();
+  const { xDataPerPixel, yDataPerPixel } = useDataPerPixel();
+
+  const nLevels = api.levelsSize.length;
+
+  const quality = 1 - clamp(qualityFactor, 0, 1);
+  const levelOfDetail = Math.min(
+    Math.floor(Math.log2(Math.max(1, xDataPerPixel, yDataPerPixel)) + quality),
+    Math.max(nLevels - 1, 0)
+  );
```
axelboc, line 34: This computation goes a little bit over my head. Any way to split it with additional, self-descriptive variables and/or explain what's going on with a comment or two?

Can we assume that `nLevels` is greater than 0 to avoid the second `Math.max`? You could put a little assertion in the constructor of `TileApi` to check for a minimum length (`if() { throw new Error('Expected `levelSizes` to have at least one size') }`). Alternatively, the constructor could accept the image size as a separate parameter, and could prepend it to the `levelSizes` array -- this would guarantee a min length of 1. Or you could keep the image size separate from the other levels and just do `[api.imageSize, ...levelSizes]` where you need to loop over all the sizes.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  // displayLowerResolutions selects which levels of detail layers are displayed:
+  // true: lower resolution layers displayed behind the current one
+  // false: only current level of detail layer is displayed
+  const lods = displayLowerResolutions
+    ? range(levelOfDetail, nLevels).reverse()
+    : [levelOfDetail];
```
axelboc, line 41: As discussed, I wonder if storing the layer sizes in the opposite order in the first place would simplify a few things. Conceptually, I find that when I'm zoomed out, the level of detail should be 0 and should increase as I zoom in. It would then make even more sense to keep the `imageSize` out of `layerSizes`, since it would no longer be the first size in the array but the last.

packages/lib/src/vis/tiles/TiledHeatmapLayer.tsx
```diff
@@ -0,0 +1,72 @@
...
+import type { TilesApi } from './api';
+import type { ColorMapAttributes } from './models';
+import { getTileOffsets, sortTilesByDistanceTo } from './utils';
+
+interface Props extends ColorMapAttributes {
+  api: TilesApi;
+  lod: number;
+}
+
+function TiledHeatmapLayer(props: Props) {
```
axelboc, line 16: I would maybe rename this to `TiledLayer` so it's more easily distinguishable from the root component name when searching for files.

packages/lib/src/vis/tiles/TiledHeatmapLayer.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const end = new Vector2(
+    Math.min(width, xVisibleDomain[1] * xScale),
+    Math.min(height, yVisibleDomain[1] * yScale)
+  );
+  const tileOffsets = getTileOffsets(origin, end, tileSize);
+
+  // Sort tiles from closest to vis center to farthest away
+  const center = new Vector2(
+    0.5 * (origin.x + end.x),
+    0.5 * (origin.y + end.y)
```
axelboc, line 43: Is it just me or ... `(origin.x + end.x) / 2`? :grin:

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// Sort tiles from closest to farthest away from ref
+export function sortTilesByDistanceTo(
+  offsets: Vector2[],
+  tileSize: Size,
+  ref: Vector2
+) {
+  offsets.sort((a, b) => {
+    const aCenter = new Vector2(
+      a.x + tileSize.width / 2,
```
axelboc, line 38: Maybe a quick destructuring of `width` / `height` here as well?

axelboc (approved):
Awesome! Feel free to implement any of these suggestions in another PR. I think it can be merged as is.

apps/storybook/src/TiledHeatmap.stories.tsx
```diff
@@ -0,0 +1,208 @@
...
+          array.set(row, col, index / iterations);
+          break;
+        }
+      }
+    }
+  }
+  return array;
+}
+
+interface TileKey {
```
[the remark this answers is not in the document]
t20100: up to you ;)

apps/storybook/src/TiledHeatmap.stories.tsx
t20100: Great, that removes a lot of boiler plate!

apps/storybook/src/TiledHeatmap.stories.tsx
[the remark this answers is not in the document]
t20100: I leave `TileKey` for now.

t20100:
Many thanks for the fantastic reviews!!!

I'll address the remaining comments when I'll be back.

apps/storybook/src/TiledHeatmap.stories.tsx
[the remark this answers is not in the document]
t20100: Finally renamed to `TileParams` to be consistent with current naming.

packages/lib/src/vis/tiles/api.ts
```diff
@@ -0,0 +1,50 @@
...
+      });
+    }
+  } while (
+    levelsSize[levelsSize.length - 1].width > tileSize.width &&
+    levelsSize[levelsSize.length - 1].height > tileSize.height
+  );
+  return levelsSize;
+}
+
+export abstract class TilesApi {
```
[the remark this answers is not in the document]
t20100: Indeed, this simplifies the code here and there

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { xDataPerPixel, yDataPerPixel } = useDataPerPixel();
+
+  const { numLayers } = api;
+
+  const dataPerPixel = Math.max(1, xDataPerPixel, yDataPerPixel);
+  const roundingOffset = 1 - clamp(qualityFactor, 0, 1);
+  const subsamplingLevel = Math.min(
+    Math.floor(Math.log2(dataPerPixel) + roundingOffset),
+    numLayers - 1
+  );
```
t20100, line 35: I hope this clarifies the computation?

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -22,6 +22,7 @@ interface Props {
   magFilter?: TextureFilter;
   alphaValues?: NdArray<TextureSafeTypedArray | Uint16Array>; // uint16 values are treated as half floats
   alphaDomain?: Domain;
+  size?: Size;
```
t20100, line 25: I wasn't too sure what to add as props: `size`, `width` and `height`, `geometrySize`... ?
I also hesitated to `extend VisMeshProps`, but I don't need it and `scale` is used by `HeatmapMesh`.

packages/lib/src/vis/shared/VisMesh.tsx
```diff
@@ -1,18 +1,25 @@
 import type { MeshProps } from '@react-three/fiber';
 
+import type { Size } from '../models';
 import { useAxisSystemContext } from './AxisSystemContext';
 
-function VisMesh(props: MeshProps) {
-  const { children, ...meshProps } = props;
+interface Props extends MeshProps {
+  size?: Size;
```
t20100, line 7: Same here: `width` and `height` ?

packages/lib/src/vis/tiles/Tile.tsx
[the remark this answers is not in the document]
t20100: Done, but it raises some questions on the props to expose.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
t20100: Hopefully made more explicit and checking for empty array.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
t20100: I inverted the order for storing layer sizes, but I kept all sizes in the same array and added a getter for the base one.

t20100:
I squashed previous commits and addressed the remaining comments each in a separate commit to hopefully make it simpler.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
axelboc: Very much so, thanks!

packages/lib/src/vis/shared/VisMesh.tsx
axelboc: Since it can be `undefined`, better keep a single object prop :+1:

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
axelboc: Let's stick with this for now, I reckon, it's nice and simple. We can refactor later.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -28,18 +28,19 @@ function TiledHeatmap(props: Props) {
...
   const dataPerPixel = Math.max(1, xDataPerPixel, yDataPerPixel);
-  const quality = 1 - clamp(qualityFactor, 0, 1);
-  const levelOfDetail = Math.min(
-    Math.floor(Math.log2(dataPerPixel) + quality),
+  const roundingOffset = 1 - clamp(qualityFactor, 0, 1);
+  const subsamplingLevel = Math.min(
+    Math.floor(Math.log2(dataPerPixel) + roundingOffset),
     numLayers - 1
   );
+  const currentLayerIndex = numLayers - 1 - subsamplingLevel;
```
axelboc, line 36: Looks like reversing the order of the layers simplified a couple things, that's good! You could maybe just add another getter for `numLayers - 1`, like `lastLayerIndex` or `topLayerIndex` or `maxLayerIndex` ... something like that? Cause it seems you need it in a few places.

axelboc (approved):
Thanks for bearing with use :joy:

packages/lib/src/vis/tiles/TiledHeatmap.tsx
t20100: Right!

t20100: I used `baseLayerIndex` and renamed `imageSize` getter to `baseLayerSize` for consistency.

Issue #962, opened 2022-02-14
RGBFormat is deprecated


axelboc:
https://github.com/mrdoob/three.js/blob/master/src/renderers/webgl/WebGLUtils.js#L50

loichuder:
#969 removed the use of `RGBFormat` in `RgbVis` but `HeatmapVis` still makes use of it (through `HeatmapMesh`)

PR #969, opened 2022-02-15
Refactor `RgbVis` with shader

axelboc:
Driven by the deprecation of `RGBFormat` #962.

An RGB image is represented as an array of size `[rows, cols, 3]`, where the three values in the last dimension are for the red, green and blue channels (or blue, green and red). There's no alpha values, so if we want to use the typed array returned by the back-end provider as is, without looping through it (for performance reasons), we can't just switch to `RGBAFormat`.

So instead, we use a 3D texture with `REDFormat` and access each color component in a custom fragment shader with the `texture()` function. Three's `DataTexture3D` is still under development, so we have to set the format and type manually: https://github.com/mrdoob/three.js/blob/master/src/textures/DataTexture3D.js

The cast/conversion to a compatible texture type remains the same as before.


packages/lib/src/vis/rgb/RgbMesh.tsx
```diff
@@ -0,0 +1,48 @@
...
+function RgbMesh(props: Props) {
+  const { values, bgr = false } = props;
+
+  const dataTexture = useMemo(() => getDataTexture3D(values), [values]);
+
+  const shader = {
+    uniforms: getUniforms({ data: dataTexture, bgr }),
+    vertexShader: VERTEX_SHADER,
+    fragmentShader: `
+      uniform highp sampler3D data;
```
axelboc, line 22: WebGL complains if I don't specify the `highp` precision for this `sampler3D`.

packages/lib/src/vis/rgb/RgbMesh.tsx
```diff
@@ -0,0 +1,57 @@
...
+      void main() {
+        float red = texture(data, vec3(0., coords)).r;
+        float green = texture(data, vec3(0.5, coords)).r;
+        float blue = texture(data, vec3(1., coords)).r;
+
+        if (bgr) {
+          gl_FragColor = vec4(blue, green, red, 1.);
+        } else {
+          gl_FragColor = vec4(red, green, blue, 1.);
+        }
```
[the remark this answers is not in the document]
loichuder: Sounds like a skateboarding move :skateboard:

packages/lib/src/vis/rgb/utils.ts
```diff
@@ -33,12 +34,14 @@ export function toRgbSafeNdArray(
...
-  const { shape } = dataArray;
-  const steps = shape.map((_, index) => (index === shape.length - 1 ? -1 : 1));
+export function getDataTexture3D(
+  values: NdArray<Uint8Array | Uint8ClampedArray | Float32Array>
+): DataTexture3D {
+  const { rows, cols } = getDims(values);
 
-  const flippedView = dataArray.step(...steps);
-  return createArrayFromView(flippedView);
+  const texture = new DataTexture3D(values.data, 3, cols, rows);
```
loichuder, line 42: Let me sum up so I get this right. `DataTexture3D` args are supposed to be `data, width, height, depth`. Semantically, we should then put `values.data, cols, rows, 3` in this order.

_But_, you put `3` first to be able to recover the RGB channels in the shader by using:
```
        float red = texture(data, vec3(0., coords)).r;
        float green = texture(data, vec3(0.5, coords)).r;
        float blue = texture(data, vec3(1., coords)).r;
```
?

packages/lib/src/vis/rgb/RgbMesh.tsx
```diff
@@ -0,0 +1,48 @@
...
+      }
+    `,
+  };
+
+  return (
+    <VisMesh scale={[1, -1, 1]}>
```
loichuder, line 42: It would be funny to try [flipY](https://threejs.org/docs/index.html?q=DataT#api/en/textures/Texture.flipY) from the texture

packages/lib/src/vis/rgb/utils.ts
axelboc: I tried `texture(data, vec3(coords, 0.)).r` at first, with `3` passed as `depth` , but was getting the wrong colors. I guess, we already pass `cols, rows` (i.e. the ndarray's reversed shape) for `width, height`, so it kinda makes sense?

loichuder (approved):

packages/lib/src/vis/rgb/RgbMesh.tsx
axelboc: Hmm, it seems to be inverting the 3rd dimension as well (i.e. the channels), it's weird:

image

axelboc: That being said, I managed to flip Y only in the shader. Pusing now.

loichuder: In this end, I don't think it is worth doing in the shader: it is more obscure than using the `scale` as in `HeatmapMesh`.

axelboc: Ah damn, sorry, I merged it 😞

PR #973, opened 2022-02-16
Avoid copying ndarray when no slicing/mapping is performed

### LINKED ISSUE #961: Avoid calling `createArrayFromView` where possible
- From `flipLastDimension` in `RgbVis`: do the BGR flip in a shader instead. #969 
- From `applyMapping`: skip call when the base array is not sliced/mapped.

_Originally posted by @loichuder in https://github.com/silx-kit/h5web/pull/960#discussion_r804417434_

axelboc:
Fix #961

### DIFF
--- a/packages/app/src/vis-packs/core/utils.ts
+++ b/packages/app/src/vis-packs/core/utils.ts
@@ -43,6 +43,11 @@ export function applyMapping<T>(
     mapping.indexOf('x') < mapping.indexOf('y');
 
   const slicingState = mapping.map((val) => (isNumber(val) ? val : null));
+
+  if (!isXBeforeY && slicingState.every((val) => val === null)) {
+    return baseArray; // no mapping/slicing needed
+  }
+
   const slicedView = baseArray.pick(...slicingState);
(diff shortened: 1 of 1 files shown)


packages/app/src/vis-packs/core/utils.ts
```diff
@@ -37,17 +37,20 @@ export function applyMapping<T>(
...
+  const slicedView = slicingState.some((val) => val === null)
+    ? baseArray.pick(...slicingState)
+    : baseArray; // no extra dimension to slice, or back-end has already done it
+
   const isXBeforeY =
     mapping.includes('y') &&
     mapping.includes('x') &&
     mapping.indexOf('x') < mapping.indexOf('y');
 
-  const slicingState = mapping.map((val) => (isNumber(val) ? val : null));
```
loichuder, line 45: Hm, what about putting an early return of `baseArray` here if `slicingState.every((val) => val === null) && !isXBeforeY` ?

This would remove the need for the two ternaries for `slicedView` and for the return value.

loichuder (approved):
Great :+1: 

(just realized I am slightly patting myself on the back here :joy: )

axelboc:
You can, you can 😁

PR #974, opened 2022-02-21
Allow Line and RectSelectionMesh to specify color of the selection

loichuder:
And [Move Selection stories out of VisCanvas category](https://github.com/silx-kit/h5web/commit/70e35f1a867fde33f364a0f3808a2bf30a1ff6c2)

### DIFF
--- a/packages/lib/src/vis/shared/LineSelectionMesh.tsx
+++ b/packages/lib/src/vis/shared/LineSelectionMesh.tsx
@@ -1,9 +1,27 @@
+import type { Color } from '@react-three/fiber';
+
 import SelectionLine from './SelectionLine';
 import type { SelectionMeshProps } from './SelectionMesh';
 import SelectionMesh from './SelectionMesh';
 
-function LineSelectionMesh(props: SelectionMeshProps) {
-  return <SelectionMesh selectionComponent={SelectionLine} {...props} />;
+interface Props extends Omit<SelectionMeshProps, 'children'> {
+  color?: Color;
+}
+
+function LineSelectionMesh(props: Props) {
+  const { color, ...meshProps } = props;
+  return (
+    <SelectionMesh {...meshProps}>
+      {(startPoint, endPoint) => (
+        <SelectionLine
+          startPoint={startPoint}
+          endPoint={endPoint}
+          color={color}
+          {...props}
+        />
+      )}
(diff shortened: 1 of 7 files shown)


packages/lib/src/vis/shared/SelectionLine.tsx
```diff
@@ -19,7 +19,7 @@ declare global {
 interface Props {
   startPoint: Vector2;
   endPoint: Vector2;
-  color?: string;
+  color?: Color;
```
loichuder, line 22: Extend the type of `color` to accept all types accepted by `meshBasicMaterial`

packages/lib/src/vis/shared/RectSelectionMesh.tsx
```diff
@@ -1,9 +1,28 @@
...
+import type { SelectionMeshProps } from '../models';
 import SelectionMesh from './SelectionMesh';
 import SelectionRect from './SelectionRect';
 
-function RectSelectionMesh(props: SelectionMeshProps) {
-  return <SelectionMesh selectionComponent={SelectionRect} {...props} />;
+interface Props extends SelectionMeshProps {
+  color?: Color;
+  opacity?: number;
+}
```
loichuder, line 10: I added the prop `opacity` for the rectangle because why not :shrug:

packages/lib/src/vis/shared/LineSelectionMesh.tsx
```diff
@@ -1,9 +1,27 @@
+import type { Color } from '@react-three/fiber';
+
+import type { SelectionMeshProps } from '../models';
 import SelectionLine from './SelectionLine';
-import type { SelectionMeshProps } from './SelectionMesh';
 import SelectionMesh from './SelectionMesh';
 
-function LineSelectionMesh(props: SelectionMeshProps) {
-  return <SelectionMesh selectionComponent={SelectionLine} {...props} />;
+interface Props extends SelectionMeshProps {
```
axelboc, line 7: You could keep `SelectionMeshProps` in `SelectionMesh` with the `children` property (and `export type { Props as SelectionMeshProps };`, and have `Omit<SelectionMeshProps, 'children'>` here maybe. Same result, so up to you.

axelboc (approved):
:100:

PR #975, opened 2022-02-21
Debounce slicing without affecting UI responsiveness

### LINKED ISSUE #920: [DimensionMapper] Review debouncing of dim mapping state
Just so I remember to look at this: the dimension mapper UI feels slightly laggy because of the debuncing on the state. We should look for an alternative, like canceling requests when slicing, or keeping a fast, internal state and debouncing on-change events.

axelboc:
Fix #920

### DIFF
--- a/packages/app/src/dimension-mapper/DimensionMapper.tsx
+++ b/packages/app/src/dimension-mapper/DimensionMapper.tsx
@@ -48,10 +48,13 @@ function DimensionMapper(props: Props) {
             <SlicingSlider
               key={`${index}`} // eslint-disable-line react/no-array-index-key
               dimension={index}
-              slicingIndex={val}
               maxIndex={rawDims[index] - 1}
-              mapperState={mapperState}
-              onChange={onChange}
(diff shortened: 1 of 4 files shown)


packages/app/src/dimension-mapper/DimensionMapper.tsx
```diff
@@ -48,10 +48,13 @@ function DimensionMapper(props: Props) {
             <SlicingSlider
               key={`${index}`} // eslint-disable-line react/no-array-index-key
               dimension={index}
-              slicingIndex={val}
               maxIndex={rawDims[index] - 1}
-              mapperState={mapperState}
-              onChange={onChange}
+              initialValue={val}
```
axelboc, line 52: `SlicingSlider` is now "uncontrolled". It manages its own internal state (initialised with `initialValue`), which it updates instantly whenever the user interacts with the slider. Debouncing is now performed on the `onChange` callback, which is only called 250ms after the user has finished interacting with the slider.

Issue #916, opened 2022-01-04
[MatrixVis] Output is not readable for arrays of long strings.


loichuder:
### Describe the bug

In the MatrixVis, elements of arrays are displayed in cells that have a fixed width and height. These dimensions are quite small as they are appropriate for numbers.

As a consequence, arrays containing long strings make unreadable cells. This is unfortunate as MatrixVis is the only vis for arrays of strings (dtype `S`).

### To Reproduce

1. Go to https://h5web.panosc.eu/?file=epics.h5
2. Click on `A note`
3. See long strings cramped in small cells

### Expected behaviour

As far as I remember, cell dimensions are fixed to allow for virtualisation of the matrix so we cannot use flex or similar fit-to-content tricks.

Given that the length of strings is part of the dtype, we could increase the cell width accordingly when mounting the component.

### Screenshots
![image](https://user-images.githubusercontent.com/42204205/148072107-572572c3-8ff7-469d-b899-9b004fd0ac32.png)

axelboc:
> Given that the length of strings is part of the dtype, we could increase the cell width accordingly when mounting the component.

That may not always be the case, though -- the dtype could be `VLEN`.

PR #976, opened 2022-02-21
Make row header cells narrower than content cells in Matrixvis

axelboc:
Preparing for #916 -- before, row header cells had the same width as content cells, which limited out ability to significantly increase the width of content cells. I had to rethink things once more:

- I no longer increment the row/col counts by 1 to account for header cells, and the `Cell` component no longer has to return `null` when `rowIndex * colIndex = 0`.
- Instead, I just move all the cells by the width of the row header cells and the height of the column header cells by incrementing the `left` and `top` positioning values.

image


packages/app/src/vis-packs/core/matrix/MappedMatrixVis.tsx
```diff
@@ -12,35 +11,29 @@ import type { DimensionMapping } from '../../../dimension-mapper/models';
...
+import { getCellWidth, getFormatter } from './utils';
 
 interface Props {
   dataset: Dataset<ArrayShape, PrintableType>;
   selection: string | undefined;
   value: ArrayValue<PrintableType>;
   dims: number[];
   dimMapping: DimensionMapping;
-  formatter: (value: Primitive<PrintableType>) => string;
-  cellWidth: number;
```
axelboc, line 23: Thought these would be better computed in this component than in the container.

packages/lib/src/vis/matrix/MatrixVis.module.css
```diff
@@ -74,6 +74,8 @@
   align-items: center;
   justify-content: center;
   background-color: var(--h5w-matrix-cell--bgColor, transparent);
+  overflow: hidden;
+  white-space: nowrap;
```
axelboc, line 78: Added this to prevent text wrapping and overflow like on the screenshot in #916  (note that `text-overflow: ellipsis` does not work due to the use of Flexbox, and I don't think it's wise to add an extra `span` in every cell.)

axelboc (commented):
In my next PR, I'll add logic to `getCellWidth` for when the dtype is fixed-length string. I'll also add an input field to the toolbar so users can customise the `cellWidth` to their liking, especially when dealing with vlen dtypes.

cypress/specs/app.spec.ts
```diff
@@ -59,8 +59,8 @@ describe('App', () => {
     cy.findByRole('table').should('exist');
     cy.findAllByRole('cell')
       .first()
-      .should('have.attr', 'aria-rowindex', 1)
-      .and('have.attr', 'aria-colindex', 1);
+      .should('have.attr', 'aria-rowindex', 0)
+      .and('have.attr', 'aria-colindex', 0);
```
axelboc, line 63: This matches the indices in header cells.

packages/lib/src/vis/matrix/StickyGrid.tsx
```diff
@@ -27,26 +27,36 @@ const StickyGrid = forwardRef<HTMLDivElement, Props>((props, ref) => {
...
       aria-colcount={columnCount}
       data-sticky={sticky || undefined}
     >
       <div className={styles.colHeaders}>
-        <div className={styles.topLeftCell} style={cellSize} data-bg />
+        <div
+          className={styles.topLeftCell}
+          style={{ width: rowHeaderCellsWidth, height: cellSize.height }}
+          data-bg
+          aria-hidden="true"
```
axelboc, line 45: Hiding header cells from the accessibility tree (`aria-rowindex`/`aria-colindex` on content cells should suffice).

packages/lib/src/vis/matrix/HeaderCells.tsx
```diff
@@ -7,20 +7,25 @@ import { SettingsContext } from './context';
 interface Props {
   indexMin: number;
   indexMax: number;
-  transform: string /* to compensate for header cells not rendered in the range [0 indexMin] */;
+  transform: string; // to compensate for header cells not rendered in the range [0 indexMin]
```
loichuder, line 10:
```suggestion
  transform: string; // to compensate for header cells not rendered in the range [0, indexMin]
```

packages/lib/src/vis/matrix/MatrixVis.tsx
```diff
@@ -18,8 +19,7 @@ function MatrixVis(props: Props) {
   const { dataArray, formatter, cellWidth, sticky = true } = props;
   const dims = dataArray.shape;
 
-  const rowCount = dims[0] + 1; // includes IndexRow
-  const columnCount = (dims.length === 2 ? dims[1] : 1) + 1; // includes IndexColumn
+  const [rowCount, columnCount = 1] = dims;
```
loichuder, line 22: Nice :sparkles:

packages/lib/src/vis/matrix/context.tsx
```diff
@@ -23,8 +24,15 @@ export const RenderedItemsContext = createContext<
...
 
 function GridProvider(props: Props) {
-  const { rowCount, columnCount, cellSize, sticky, cellFormatter, children } =
-    props;
+  const {
+    rowCount,
+    columnCount,
+    cellSize,
+    sticky,
+    rowHeaderCellsWidth,
```
loichuder, line 32: It is a bit weird to have a specific parameter for the row header cell width while the height is given by `cellSize.height`

Perhaps we should also have a specific parameter for header cell height that defaults to `cellSize.height` ? Following the same behaviour as `rowHeaderCellsWidth` ?

packages/lib/src/vis/matrix/StickyGrid.tsx
```diff
@@ -27,26 +27,36 @@ const StickyGrid = forwardRef<HTMLDivElement, Props>((props, ref) => {
     <div
       ref={ref}
       className={styles.stickyGrid}
-      style={style}
+      style={{
+        ...style,
+        width: (style.width as number) + rowHeaderCellsWidth, // account for row header cells
+        height: (style.height as number) + cellSize.height, // account for column header cells
```
loichuder, line 33: To follow on my previous comment, I would be inclined to have a `columnHeaderCellsHeight` here for example

cypress/specs/app.spec.ts
loichuder: Much better :+1:

loichuder (approved):

packages/lib/src/vis/matrix/context.tsx
axelboc: Hmm you mean like provide a `rowHeaderCellSize` object through context? That makes sense. I'll give it a try,.

axelboc: Actually, since we never use `cellSize` as a whole but always `cellSize.width/height`, I'm actually going to go the other way and change `cellSize` to `cellWidth`, and use the global `CELL_HEIGHT` constant where needed. What do you think?

loichuder: Yep sounds good :+1:

axelboc: Not quite happy with the result. Merging as is for now, maybe we'll think of something better at some point, especially if we ever have to make the cell height configurable (it is currently hard-coded).

PR #977, opened 2022-02-22
Size SVGs explicitly in axis system to fix zoom bug in Chrome

axelboc:
Fix https://github.com/silx-kit/jupyterlab-h5web/issues/80

### DIFF
--- a/packages/lib/src/vis/shared/Axis.tsx
+++ b/packages/lib/src/vis/shared/Axis.tsx
@@ -32,11 +32,12 @@ interface Props {
   config: AxisConfig;
   domain: Domain;
   canvasSize: Size;
+  svgSize: Size;
   flipAxis?: boolean;
 }
 
 function Axis(props: Props) {
-  const { type, config, domain, canvasSize, flipAxis } = props;
+  const { type, config, domain, canvasSize, svgSize, flipAxis } = props;
 
   const { width, height } = canvasSize;
   const axisLength = type === 'abscissa' ? width : height;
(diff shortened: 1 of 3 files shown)


packages/lib/src/vis/shared/Axis.tsx
```diff
@@ -69,7 +70,7 @@ function Axis(props: Props) {
         />
       </svg>
       {showGrid && (
-        <svg className={styles.grid}>
+        <svg className={styles.grid} style={canvasSize}>
```
axelboc, line 73: Might as well give the grids their proper sizes.

packages/lib/src/vis/shared/AxisSystem.module.css
```diff
@@ -13,19 +13,15 @@
...
 }
 
 .axis[data-type='abscissa'] {
   grid-area: bottom-axis;
 }
 
 .axis[data-type='ordinate'] {
-  grid-area: axis-grid;
+  grid-area: left-axis;
+  transform: translateX(100%);
```
axelboc, line 24: Moving left axis to correct CSS grid area and adding `transform` to compensate for the positioning of the `group` inside the SVG. Feels more logical than to position the axis in the middle area, after all.

axelboc:
The screenshots have changed a bit, I'm going to double check them.

axelboc:
Looks like it's just pixel aliasing in the grid. Can't notice a change locally in both Chrome and Firefox.

loichuder (approved):
Great :+1: 

> Looks like it's just pixel aliasing in the grid. Can't notice a change locally in both Chrome and Firefox.

Yeah I couldn't find a difference either (except for the bug fix of course :smile: )

PR #979, opened 2022-02-22
Remove duplicate hooks

axelboc:
Just a bit of almost-Spring cleaning.

### DIFF
--- a/apps/storybook/src/Utilities.stories.mdx
+++ b/apps/storybook/src/Utilities.stories.mdx
@@ -75,7 +75,7 @@ getLinearGradient(interpolator: D3Interpolator, direction: 'top' | 'bottom' | 'r
 
 #### getVisDomain
 
-Determine the domain to be used for the visualization based on a user-selected custom domain. If a bound of the custom domain is `null`, it falls back to the corresponding bound of the data domain.
+Determine the domain to be used for the visualization based on a custom domain selected by the user. If a bound of the custom domain is `null`, it falls back to the corresponding bound of the data domain.
 
 ```ts
 getVisDomain(customDomain: CustomDomain, dataDomain: Domain): Domain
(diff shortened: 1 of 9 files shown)


packages/lib/src/vis/hooks.ts
```diff
@@ -18,29 +17,25 @@ import {
...
 }
 
-export function useFrameRendering(): void {
-  const [, setNum] = useState(0);
-
-  useFrame(() => {
-    setNum(Math.random());
-  });
-}
-
```
axelboc, line 39: Moving things around a little.

packages/lib/src/vis/hooks.ts
```diff
@@ -18,29 +17,25 @@ import {
...
 export function useDomains(
-  valuesArrays: (NdArray<number[]> | number[])[],
+  valuesArrays: AnyNumArray[],
```
axelboc, line 36: This hook wasn't the one that was used, so it didn't get the typed arrays changes.

apps/storybook/src/Utilities.stories.mdx
```diff
@@ -86,9 +86,9 @@ const visDomain2 = getVisDomain([null, 20], [0, 100]); // [0, 20]
 
 #### getSafeDomain
 
-Determine a domain that is safe for the visualization. This is typically called with a user-defined `customDomain`, or with a `visDomain` as returned by `getVisDomain()`.
-If the domain is determined to be unsafe, a safe domain based on `fallbackDomain` is returned along with an error object. Note that `fallbackDomain` is assumed to be safe.
-The domain is considered unsafe if it's inverted (`min > max`), or if the scale is log and at least one of the bounds is negative.
+Make a domain safe in a given scale. This is typically called with a user-defined `customDomain`, or with a `visDomain` as returned by `getVisDomain()`.
+If `domain` is determined to be unsafe, a safe domain based on `fallbackDomain` is returned along with an error object. Note that `fallbackDomain` is assumed to be safe in the given scale.
+The domain is considered unsafe if it's inverted (`min > max`), if the scale is log and at least one of the bounds is negative, etc.
```
loichuder, line 91: Not a fan of the addtion of `etc` here.

Im my opinion, we should make clear which domains are unsafe. If more cases need to be added, you can convert this in a bullet list

packages/app/src/vis-packs/core/heatmap/hooks.ts
```diff
@@ -1,5 +0,0 @@
-import { getSafeDomain, getVisDomain } from '@h5web/lib';
-import { createMemo } from 'react-use';
-
-export const useVisDomain = createMemo(getVisDomain);
-export const useSafeDomain = createMemo(getSafeDomain);
```
loichuder, line 5: Well spotted :monocle_face: !

loichuder (approved):

apps/storybook/src/Utilities.stories.mdx
axelboc: Might add some tests as well actually, and update the examples, I think they're out of date.

loichuder (approved):
Excellent :+1:

Issue #966, opened 2022-02-15
Selection-based zoom


axelboc:
- [x] Implement component to zoom by tracing a selection rectangle (via `SelectionMesh`). #981
- [ ] Bind selection zoom to `Ctrl` key in `@h5web/app` for Line, Heatmap and RGB vis.

loichuder:
I would add that #984 would be a welcome companion to this feature

PR #981, opened 2022-02-23
Add a SelectionMesh that zoom on the selected rectangle

loichuder:
First part of #966 

It made more sense to me to have a "frame" rather than a filled rectangle to select the zoom area. For that, I had to change the component displaying the selection `SelectionRect`.

Also, once zoomed in on the selection, we don't want the selection to still be displayed. So I changed the render props function as well (see below).


apps/storybook/src/Selection.stories.tsx
```diff
@@ -24,6 +27,13 @@ function vectorToStr(vec: Vector2) {
   return `(${format('.2f')(vec.x)}, ${format('.2f')(vec.y)})`;
 }
 
+const X = range(-5, 5, 0.01);
+const Y = X.map((x) => Math.exp((-x * x) / 0.2));
+
+function GaussianCurve() {
+  return <DataCurve abscissas={X} ordinates={Y} color="blue" />;
+}
```
loichuder, line 35: I made a new story to demonstrate the new component. I found it more user-friendly to have something to zoom at, hence this small component that displays a gaussian curve.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -106,6 +116,32 @@ ChangeSelectionColors.argTypes = {
   },
 };
 
+export const SelectToZoom: Story<{
```
loichuder, line 119: Please try the new story yourself and tell me how it feels !

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -11,7 +11,11 @@ import VisMesh from './VisMesh';
 interface Props extends MeshProps {
   onSelection?: (startPoint: Vector2, endPoint: Vector2) => void;
   modifierKey?: ModifierKey;
-  children: (startPoint: Vector2, endPoint: Vector2) => ReactElement;
+  children: (
+    startPoint: Vector2,
+    endPoint: Vector2,
+    isDragging: boolean
+  ) => ReactElement | null;
```
loichuder, line 18: As explained above, the render props function was changed to include the `isDragging` boolean that tells if the user is selecting/dragging the selection.

This allows to remove the selection component in `ZoomSelectionMesh` when the selection was finished.

packages/lib/src/vis/shared/SelectionRect.tsx
```diff
@@ -1,23 +1,44 @@
...
 
 function SelectionRect(props: Props) {
-  const { startPoint, endPoint, color = 'red', opacity = 0.5 } = props;
+  const {
+    startPoint,
+    endPoint,
+    color = 'red',
+    opacity = 0.5,
+    borderColor,
+  } = props;
```
loichuder, line 21: `SelectionRect` is now a bit more generic: it can display a filled rectangle **and** a border. By default, `borderColor` is `undefined` which disable the border.

packages/lib/src/vis/shared/SelectionRect.tsx
```diff
@@ -1,23 +1,44 @@
...
   const width = endPoint.x - startPoint.x;
   const height = endPoint.y - startPoint.y;
 
+  const rectGeometry = useMemo(
+    () => new PlaneGeometry(Math.abs(width), Math.abs(height)),
+    [height, width]
+  );
```
loichuder, line 28: To display the border, I use [EdgesGeometry](https://threejs.org/docs/index.html#api/en/geometries/EdgesGeometry) which takes the geometry of the filled volume as arg.

That means that unfortunately, I need to create the filled rectangle geometry imperatively to have access to it declaratively below.

packages/lib/src/vis/shared/ZoomSelectionMesh.tsx
```diff
@@ -0,0 +1,63 @@
...
+        isDragging ? (
+          <SelectionRect
+            startPoint={startPoint}
+            endPoint={endPoint}
+            opacity={0}
+            borderColor="black"
+          />
+        ) : null
+      }
+    </SelectionMesh>
```
loichuder, line 59: `RectSelectionMesh` API was too simple for what I wanted to do. So rather than expanding it, I made use of `SelectionMesh` directly.

packages/lib/src/vis/shared/ZoomSelectionMesh.tsx
```diff
@@ -0,0 +1,63 @@
...
+    camera.updateProjectionMatrix();
+    camera.updateMatrixWorld();
+
+    const midPoint = worldEndPoint
+      .clone()
+      .sub(worldStartPoint)
+      .divideScalar(2)
+      .add(worldStartPoint);
+
+    moveCameraTo(midPoint.x, midPoint.y);
```
loichuder, line 44: The zooming works in two steps:
- First, we need to set the camera scale so that the dimensions of the rectangle corresponds to the dimensions of the FOV. This is a simple proportional rule, for example for x :

| scale | FOV |
 | -- | -- |
 | 1 | `width` |
| x | `rectWidth` |

=> `x = rectWidth / width`

- Then, the camera must be moved to the center of the rectangle (called `midPoint` but `rectCenter` might be a better name). For this, I make use of the `moveCameraTo` utility that moves the camera in the vis bounds. As explained by the comment, the scale must be computed first to update camera matrices so that `moveCameraTo` computes the correct camera bounds.

packages/lib/src/vis/shared/ZoomSelectionMesh.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const { width, height } = useThree((state) => state.size);
+  const camera = useThree((state) => state.camera);
+
+  const onSelection = (startPoint: Vector2, endPoint: Vector2) => {
+    if (startPoint.x === endPoint.x && startPoint.y === endPoint.y) {
+      return;
+    }
+
+    const worldStartPoint = dataToWorld(startPoint);
+    const worldEndPoint = dataToWorld(endPoint);
```
axelboc, line 29: We work with data-space coordinates in `SelectionMesh` but in the end, we call `dataToWorld` here and in the rect/line selection components, so perhaps we should reconsider or pass coordinates in both spaces?

packages/lib/src/vis/shared/ZoomSelectionMesh.tsx
axelboc: I agree with `rectCenter` or `selectionCenter` (and `selectionWidth/Height`)

packages/lib/src/vis/shared/ZoomSelectionMesh.tsx
axelboc: As per my review's overall comment, I think with `isDragging` gone, you could forward the `SelectionRect` props. That being said, I'm wondering if `RectSelectionMesh` and `LineSelectionMesh` are really worth having at all. I'm starting to think that the API of `SelectionMesh` is clear-enough as it is thanks to `SelectionRect` and `SelectionLine`.

axelboc (changes requested):
The zoom works great! :clap:  

I'm just not sure about the `isDragging` change. The `SelectionMesh` as it is was designed with the profile feature in mind, where the selection stays on the screen once drawn. This is incompatible with the sleect-to-zoom feature where the selection has to be temporary. It will also cause issues if we ever want to trace multiple lines/rectangles on the canvas at once.

There's a fairly easy way to make `SelectionMesh` more flexible: reset the mesh's internal `startPoint` and `endPoint` state when the user stops dragging, and let the consumer component store the selection via `onSelection` and render it onto the canvas (with `SelectionRect` or `SelectionLine`).

axelboc:
Also something I found when testing the story: should the selection be cleared when the user releases the modifier key? At the moment there's now way to "cancel" the selection (apart from bringing it back to 0px).

apps/storybook/src/Selection.stories.tsx
```diff
@@ -24,6 +30,45 @@ function vectorToStr(vec: Vector2) {
...
+                endPoint: worldToData(endPoint),
+              })
+          : undefined
+      }
+      modifierKey={modifierKey}
+    >
+      {children}
+    </SelectionMesh>
+  );
+}
```
loichuder, line 70: Due to https://github.com/silx-kit/h5web/pull/981#discussion_r813019807, `onSelection` parameters are now in world coordinates.

That means that for the story that prints the selection points in data coordinates, we need for a component that has access to `AxisSystemContext` to call `worldToData`.

Perhaps it is worth having this component in `@h5web/lib` ?

apps/storybook/src/Selection.stories.tsx
```diff
@@ -24,6 +30,45 @@ function vectorToStr(vec: Vector2) {
...
+
+const modifierKeyArgType = {
+  control: { type: 'inline-radio' },
+  options: ['Alt', 'Control', 'Shift'],
+};
+
+const SelectionArgType = {
+  control: { type: 'inline-radio' },
+  options: ['line', 'rectangle'],
+};
```
loichuder, line 48: I put the arg types in constants so that `sonarjs` would stop complaining about the repetition of `inline-radio`

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -63,24 +61,22 @@ function SelectionMesh(props: Props) {
...
+    [onSelection, startPoint, isDragging]
   );
 
   return (
     <VisMesh {...{ onPointerMove, onPointerUp, onPointerDown, ...meshProps }}>
       <meshBasicMaterial opacity={0} transparent />
-      {startPoint &&
-        endPoint &&
-        children(dataToWorld(startPoint), dataToWorld(endPoint))}
+      {startPoint && endPoint ? children({ startPoint, endPoint }) : null}
```
loichuder, line 79: Unsure if setting `null` if the points are undefined should be the responsibility of `SelectionMesh` or its consumers

loichuder (commented):
I tried to rework `SelectionMesh` in accordance with your comments.

This changes the behaviour of the selection components: they are now removed when releasing the mouse. It would be the responsibility of the consumer components to persist them.

I will work on a prototype while you review the first round of changes.

apps/storybook/src/Selection.stories.tsx
axelboc: You could also just disable the rule in `stories` files in `apps/storybook/.eslintrc.js`

apps/storybook/src/Selection.stories.tsx
axelboc: Perhaps this story should live in its own file, with title `Building Blocks/ZoomSelectionMesh`.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -101,11 +151,95 @@ ChangeSelectionColors.args = {
...
+  const SelectionComponent =
+    selection === 'line' ? SelectionLine : SelectionRect;
+
+  return (
+    <VisCanvas
+      abscissaConfig={{ visDomain: [-5, 5], showGrid: true }}
+      ordinateConfig={{ visDomain: [-0.5, 1.5], showGrid: true }}
+    >
+      {/* <PanMesh />
+      <ZoomMesh /> */}
```
axelboc, line 215: In case this flew under your radar, it might be worth uncommenting these two meshes.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -101,11 +151,95 @@ ChangeSelectionColors.args = {
...
+      {selection && (
```
axelboc, line 216: I don't think `selection` needs to be optional in `TemplateProps` now that we have a separate template for `SelectionMesh` stories.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -101,11 +151,95 @@ ChangeSelectionColors.args = {
...
+        <DataSelectionMesh
+          modifierKey={modifierKey}
+          onSelectionStart={(start) => {
+            setStart(start);
+            setEnd(undefined);
+          }}
+          onSelectionEnd={setEnd}
```
axelboc, line 223: The way I had pictured it was that the consumer would persist a full `selection` object with mandatory start and end points - i.e. `useState<Selection | undefined>` with `interface Selection { startPoint: Vector2; endpoint: Vector2 }`.

So the `SelectionMesh` callbacks would have the following signatures:

```ts
onSelectionStart: () => void;
onSelectionEnd: (selection: Selection): void;
```

and would be set as follows in this story:

```tsx
onSelectionStart={() => setSelection(undefined)}
onSelectionEnd={setSelection}
```

apps/storybook/src/Selection.stories.tsx
```diff
@@ -52,19 +100,17 @@ const Template: Story<TemplateProps> = (args) => {
...
-        {selection === 'rectangle' && (
-          <RectSelectionMesh
-            onSelection={(start, end) => setSelectedVectors([start, end])}
+        {selection && (
+          <DataSelectionMesh
+            onSelectionStart={(start) => {
+              setEnd(undefined);
+              setStart(start);
+            }}
+            onSelectionEnd={setEnd}
```
axelboc, line 109: Following on my comment in story `PersistSelection`, the API I'm suggesting is not sufficient to implement this story. The goal here is to display the selection coordinates as the user is drawing the selection (i.e. from `onPointerMove`), so we need a third callback, perhaps `onSelectionChange(selection: Selection) => void`. The callbacks would then be used as follows:

```tsx
onSelectionChange={setSelection}
onSelectionEnd={() => setSelection(undefined)}
```

packages/lib/src/vis/shared/SelectionMesh.tsx
axelboc: Oh it's actually better like this, I reckon, you're right. Less to think about in the consumer.

packages/lib/src/vis/shared/ZoomSelectionMesh.tsx
```diff
@@ -0,0 +1,62 @@
...
+      .add(startPoint);
+
+    moveCameraTo(selectionCenter.x, selectionCenter.y);
+  };
+
+  return (
+    <SelectionMesh
+      modifierKey={modifierKey}
+      onSelectionStart={setStartPoint}
+      onSelectionEnd={onSelectionEnd}
```
axelboc, line 53: Following on my previous two comments in the stories, I don't think this component needs a state at all. I think the callbacks should be declared like this (with `onSelectionStart` removed):

```ts
onSelectionEnd={handleSelectionEnd}
```

with `function handleSelectionEnd(selection:Selection): void { ... }`

axelboc:
Forgot to comment on the `dataToWorld` issue. Perhaps a solution could be to add a prop to `SelectionMesh` called `worldSpace` that defaults to `false` and controls whether the coordinates of the points should be left in world space or converted to data space?

apps/storybook/src/Selection.stories.tsx
```diff
@@ -1,89 +1,137 @@
...
-        abscissaConfig={{ visDomain: [-10, 0], showGrid: true }}
-        ordinateConfig={{ visDomain: [50, 100], showGrid: true }}
+    <VisCanvas
+      title={
+        selection
+          ? `Selection from ${vectorToStr(
+              selection.startPoint
+            )} to ${vectorToStr(selection.endPoint)}`
+          : 'No selection'
+      }
```
loichuder, line 86: I changed the story according to https://github.com/silx-kit/h5web/pull/982/files#r812992731. It indeed feels better.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -106,10 +154,46 @@ ChangeSelectionColors.argTypes = {
...
+          <SelectionComponent
+            selectionType={selectionType}
+            color={color}
+            {...selection}
+          />
+        ) : null
+      }
+    </Template>
+  );
+};
```
loichuder, line 175: Not super happy about how this turned out. If you have a better idea...

Also, it seems there is a flickering of the selection component when releasing the mouse

apps/storybook/src/Selection.stories.tsx
axelboc: Would it make sense for `SelectionLine` and `SelectionRect` to also have a prop `withWorldCoordinates` so that we wouldn't have to call `dataToWorld` here? This would allow inlining this component in the story template.

For now, though, you could maybe just remove the `useWorldCoordinates` argument and inline this component in the template right away.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -1,89 +1,137 @@
...
     disablePan = true,
     disableZoom = true,
     modifierKey,
     color,
+    useWorldCoordinates,
+    children,
   } = args;
 
-  const [selectedVectors, setSelectedVectors] = useState<[Vector2, Vector2]>();
+  const [selection, setSelection] = useState<SelectionPoints>();
```
axelboc, line 71: I would maybe rename this state to `activeSelection`

apps/storybook/src/Selection.stories.tsx
axelboc: Yeah, I don't think the render prop pattern is needed here -- it's just confusing. You can move the `childrenSelection` state here (and rename it to `savedSelection` or just `selection`), forward an `onSelectionEnd` prop through the story template, and just render the selection as a simple child of the template.

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -1,27 +1,48 @@
...
 
 interface Props extends MeshProps {
-  onSelection?: (startPoint: Vector2, endPoint: Vector2) => void;
+  onSelectionStart?: () => void;
+  onSelectionChange?: (points: SelectionPoints) => void;
+  onSelectionEnd?: (points: SelectionPoints) => void;
   modifierKey?: ModifierKey;
-  children: (startPoint: Vector2, endPoint: Vector2) => ReactElement;
+  children: (points: SelectionPoints) => ReactElement;
+  useWorldCoordinates?: boolean;
```
axelboc, line 18: Maybe `withWorldCoordinates` to avoid confusion with hooks?

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -1,27 +1,48 @@
 import type { MeshProps, ThreeEvent } from '@react-three/fiber';
 import type { ReactElement } from 'react';
-import { useCallback, useState } from 'react';
-import type { Vector2 } from 'three';
+import { useCallback, useMemo, useState } from 'react';
+import type { Vector3 } from 'three';
+import { Vector2 } from 'three';
 
-import type { ModifierKey } from '../models';
+import type { ModifierKey, SelectionPoints } from '../models';
```
axelboc, line 7: Is there a problem in just calling this `Selection`?

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -1,27 +1,48 @@
...
 
+  const { worldToData } = useAxisSystemContext();
+
+  const project = useMemo(
+    () =>
+      useWorldCoordinates
+        ? (pt: Vector3) => new Vector2(pt.x, pt.y)
+        : worldToData,
+    [useWorldCoordinates, worldToData]
+  );
```
axelboc, line 44: `useCallback` to avoid the extra arrow paren?

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -35,25 +56,31 @@ function SelectionMesh(props: Props) {
...
   const onPointerMove = useCallback(
     (evt: ThreeEvent<PointerEvent>) => {
       if (!isDragging) {
         return;
       }
-      const point = worldToData(evt.unprojectedPoint);
+      const point = project(evt.unprojectedPoint);
 
       setEndPoint(point);
+      if (startPoint && onSelectionChange) {
```
axelboc, line 79: I wonder if you still need `isDragging` now? You may be able to just rely on `startPoint` being defined (in which case the early return would allow removing it from this condition).

packages/lib/src/vis/shared/ZoomSelectionMesh.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { modifierKey } = props;
+
+  const moveCameraTo = useMoveCameraTo();
+
+  const { width, height } = useThree((state) => state.size);
+  const camera = useThree((state) => state.camera);
+
+  const onSelectionEnd = (selection: SelectionPoints) => {
+    const { startPoint, endPoint } = selection;
+    if (startPoint.x === endPoint.x && startPoint.y === endPoint.y) {
```
axelboc, line 22: Not for this PR, but we could imagine calculating the width/height of the selection and returning when it's below a threshold -- to avoid zooming in on a tiny area by mistake.

axelboc (approved):

axelboc:
If the flicker still remains, we can investigate later.

packages/lib/src/vis/shared/SelectionMesh.tsx
loichuder: Excellent analysis :+1:

apps/storybook/src/Selection.stories.tsx
loichuder: We could.

But then, we must be aware that we cannot use components with `withWorldCoordinates` to persist selections at the risk of falling back on the issue solved by https://github.com/silx-kit/h5web/pull/935

loichuder: `SelectionLine` and `SelectionRect` now consider that props are in data coordinates to avoid any resizing problems

apps/storybook/src/Selection.stories.tsx
loichuder: Implemented this. I just had to put `onSelectionStart` as well to reset the saved selection

PR #982, opened 2022-02-23
Upgrade deps and remove CRA preset from Storybook

axelboc:
I ran into issues upgrading Storybook, so I decided to try removing CRA again and I finally got there.

Upgrading Three.js still breaks the shader-based visualizations: `texture2D` seems to always return `0`.


apps/storybook/.storybook/main.js
```diff
@@ -6,23 +6,8 @@ module.exports = {
   },
   stories: ['../src/**/*.stories.@(tsx|mdx)'],
   addons: [
-    {
-      name: '@storybook/preset-create-react-app',
-      // Override react-scripts package path as it resolves to the wrong path on Linux
-      options: { scriptsPackageName: '../node_modules/react-scripts' },
-    },
+    'storybook-css-modules-preset',
```
axelboc, line 9: Storybook doesn't support CSS modules by default... :disappointed:

apps/storybook/.storybook/preview.js
```diff
@@ -1,5 +1,5 @@
 import 'normalize.css';
-import '@h5web/lib/src/styles.css';
+import '@h5web/shared/src/styles.css';
```
axelboc: Storybook doesn't support importing modules with `@import '<module-name>`, so the import of shared package's styles in `lib/src/styles.css` fails. It supports `@import '~<module-name>`, which is webpack's `css-loader` module import syntax, but Vite doesn't support it.

apps/storybook/.storybook/preview.js
```diff
@@ -1,5 +1,5 @@
 import 'normalize.css';
-import '@h5web/lib/src/styles.css';
+import '@h5web/shared/src/styles.css'; // importing via `@h5web/lib/src/styles.css` styles fails due ot lack of `~` prefix (as it is not supported by Vite)
```
axelboc, line 2: Storybook doesn't support importing modules with @import '<module-name>, so the import of shared package's styles in lib/src/styles.css fails. It supports @import '~<module-name>, which is webpack's css-loader module import syntax, but Vite doesn't support it.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -36,38 +36,34 @@ const Template: Story<TemplateProps> = (args) => {
...
   return (
-    <>
-      {selection && (
-        <p style={{ textAlign: 'center' }}>
-          {selectedVectors
-            ? `Selection from ${vectorToStr(
-                selectedVectors[0]
-              )} to ${vectorToStr(selectedVectors[1])}`
-            : 'No selection'}
-        </p>
```
axelboc, line 47: I noticed that the canvas would grow when opening the story, probably due to `display: grid` on the parent `FillHeight` decorator. So I now pass the current selection via the `title` prop of `VisCanvas`.

eslint.shared.js
```diff
@@ -76,6 +77,7 @@ const overrides = [
       'jest/prefer-strict-equal': 'off', // `toEqual` is shorter and sufficient in most cases
       'jest-formatting/padding-around-all': 'off', // allow writing concise two-line tests
       'jest/require-top-level-describe': 'off', // filename should already be meaningful, extra nesting is unnecessary
+      'jest/no-conditional-in-test': 'off', // false positives in E2E tests (snapshots), and too strict (disallows using `||` for convenience)
```
axelboc, line 80: Disabling some new troublesome linting rules.

axelboc:
I'll rebase and fix the conflicts if you want to merge first :wink:

apps/storybook/.storybook/main.js
loichuder: Well, it could be worse than just an addon to add :smile:

loichuder (approved):

PR #983, opened 2022-02-28
Bring ScatterVis parameters in line with other vis

loichuder:
Now accepts:
- a `ndarray` instead of `number[]`. The `ndarray` can contain a `number[]` or a `TypedArray`
- a `size` parameter
- a `title` parameter
- Labels for abscissas and ordinates
- `children`


axelboc (approved):
Excellent!

PR #990, opened 2022-03-03
Add ZoomSelectionMesh to Heatmap, Line, Rgb and ScatterVis

### LINKED ISSUE #966: Selection-based zoom

loichuder:
Peek 2022-03-03 08-55

Fix #966. I still need to fix the behaviour when the pointer goes out of the window. Probably related to pointer capture


axelboc (approved):
Isn't `Control` the default?

loichuder:
> Isn't `Control` the default?

No but we can set this as such :slightly_smiling_face:

axelboc:
It would make sense, I think, but up to you

PR #991, opened 2022-03-03
Add a Reset zoom button to Heatmap, Line, Rgb and ScatterVis

### LINKED ISSUE #984: Ability to reset zoom level
This has been a long-standing request. It can take time to zoom out with the mouse wheel so we should provide a way to reset the zoom.

loichuder:
Fix #984 

image

The button only appears when the zoom is not null.


axelboc (commented):
Good idea putting the button inside the heatmap. However, I feel like it should maybe be in the bottom right corner and look a bit more like a button -- a bit like on Google Maps:

image

I can do some CSS tweaking myself if you'd like?

axelboc:
image

If you can please just double check with dark mode on, that'd be great.

loichuder:
image

Looks fine albeit a bit discreet. What do you think ?

PR #992, opened 2022-03-03
Bound selection mesh to the current FOV

loichuder:


packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -59,7 +63,10 @@ function SelectionMesh(props: Props) {
       if (!startPoint) {
         return;
       }
-      const point = worldToData(evt.unprojectedPoint);
+      const boundedX = clamp(evt.unprojectedPoint.x, fov.left, fov.right);
+      const boundedY = clamp(evt.unprojectedPoint.y, fov.bottom, fov.top);
```
axelboc, line 67: `bound` is already past tense :wink: - maybe just use the same names as in `useMoveCameraTo`: `xBound` and `yBound`?

packages/lib/src/vis/shared/hooks.ts
```diff
@@ -84,3 +84,12 @@ export function useZoomOnWheel(
...
   return onWheel;
 }
+
+export function useCameraFOV() {
+  const camera = useThree((state) => state.camera);
+
+  const { x: right, y: top } = CAMERA_TOP_RIGHT.clone().unproject(camera);
+  const { x: left, y: bottom } = CAMERA_BOTTOM_LEFT.clone().unproject(camera);
+
+  return { left, right, top, bottom };
```
axelboc, line 94: If the hook is used in a component that doesn't also call `useFrameRendering`, then the FOV will be stale pretty quickly. Perhaps it would be better to return a function created with `useCallback` - i.e `useGetCameraFOV`, thus following the same pattern as `useMoveCameraTo`?

Alternatively, you could make it a normal function with a single `camera` parameter and call it from inside your `useCallback` in `SelectionMesh`.

axelboc: Might be worth using this new utility in `PanMesh`, by the way!

loichuder:
>  I assume this PR comes in preparation for the pointer capture fix but does not yet fix it, right?

Not really as preparation. I rather wanted to fix an unrelated problem.

packages/lib/src/vis/shared/SelectionMesh.tsx
loichuder: Can't we say that the FOV bounds the coordinates ? Therefore, the coordinate is bounded ? :thinking: 

To me, `xBound` means the value is the limit in x whereas here, it stores the x value limited by the FOV.

So, the meanings are not the same but I am open to other namings !

packages/lib/src/vis/shared/hooks.ts
loichuder: > If the hook is used in a component that doesn't also call useFrameRendering, then the FOV will be stale pretty quickly. 

Good point ! I will change this

packages/lib/src/vis/shared/SelectionMesh.tsx
axelboc: Oh apologies, `bounded` is indeed the paste tense of "to bound" 😅 and I get what you mean about the two different meanings of the variables. 👍

axelboc (approved):

PR #995, opened 2022-03-04
Rework InteractionMesh to attach events to canvas

### LINKED ISSUE #994: The selection interaction should keep working when the cursor leaves the canvas.

loichuder:
...to make the events independent of intersections with the mesh.

Fix #994

As we opt out from R3F event machinery, we have to compute the world coordinates of the event location by ourselves (called `unprojectedPoint` in R3F). I added a utility to this purpose.


packages/lib/src/vis/shared/InteractionMesh.tsx
```diff
@@ -1,15 +1,51 @@
...
-  const { width, height } = useThree((state) => state.size);
+interface Props {
+  onPointerDown?: (evt: PointerEvent) => void;
+  onPointerMove?: (evt: PointerEvent) => void;
+  onPointerUp?: (evt: PointerEvent) => void;
+  onWheel?: (evt: WheelEvent) => void;
+  children?: ReactNode;
+}
+
+function InteractionMesh(props: Props) {
```
loichuder, line 13: Perhaps it should be renamed ?

packages/lib/src/vis/shared/InteractionMesh.tsx
```diff
@@ -1,15 +1,51 @@
...
+    }
+    if (onPointerMove) {
+      domElement.addEventListener('pointermove', onPointerMove);
+    }
+    if (onPointerUp) {
+      domElement.addEventListener('pointerup', onPointerUp);
+    }
+    if (onWheel) {
+      domElement.addEventListener('wheel', onWheel);
+    }
```
loichuder, line 30: Not sure if there is a better of doing this

packages/lib/src/vis/shared/PanMesh.tsx
```diff
@@ -17,14 +16,15 @@ function PanMesh(props: Props) {
...
 
   const startOffsetPosition = useRef<Vector3>(); // `useRef` to avoid re-renders
 
   const moveCameraTo = useMoveCameraTo();
 
   const onPointerDown = useCallback(
-    (evt: ThreeEvent<PointerEvent>) => {
-      const { sourceEvent, unprojectedPoint } = evt;
+    (sourceEvent: PointerEvent) => {
+      const unprojectedPoint = getUnprojectedPoint(sourceEvent, size, camera);
```
loichuder, line 27: `unprojectedPoint` is computed through a function that takes the `size` and the `camera` as arg. 

We could create a callback out of a hook to avoid passing such arg but most callers already get the `camera` via `useThree`. And I am not convinced about the need for `size` (see the discussion on `getUnprojectedPoint` below)

packages/lib/src/vis/shared/SelectionMesh.tsx
```diff
@@ -100,14 +104,13 @@ function SelectionMesh(props: Props) {
...
+    [startPoint, size, camera, worldToData, onSelectionEnd]
   );
 
   return (
-    <VisMesh {...{ onPointerMove, onPointerUp, onPointerDown, ...meshProps }}>
-      <meshBasicMaterial opacity={0} transparent />
+    <InteractionMesh {...{ onPointerDown, onPointerMove, onPointerUp }}>
       {startPoint && endPoint ? children({ startPoint, endPoint }) : null}
-    </VisMesh>
+    </InteractionMesh>
```
loichuder, line 113: By adding a `children` prop to `InteractionMesh`, I was able to use in `SelectionMesh`

packages/lib/src/vis/utils.ts
```diff
@@ -342,3 +342,16 @@ export function getCameraFOV(camera: Camera): {
...
   return { topRight, bottomLeft };
 }
+
+export function getUnprojectedPoint(
+  event: PointerEvent | WheelEvent,
+  size: Size,
+  camera: Camera
+): Vector3 {
+  const { offsetX: x, offsetY: y } = event;
+  const { width, height } = size;
```
loichuder, line 352: The size is supposed to be retrieved from `useThree`. 

Would it be possible to retrieve it from the `event` ? It contains the `innerWidth` and `innerHeight` of the element on which the event is attached, right (hence the canvas) ?

packages/lib/src/vis/shared/InteractionMesh.tsx
```diff
@@ -1,15 +1,51 @@
...
+    };
+  });
 
-  return (
-    <mesh {...props}>
-      <meshBasicMaterial opacity={0} transparent />
-      <planeGeometry args={[width, height]} />
-    </mesh>
-  );
+  return <group>{children}</group>;
```
axelboc, line 48: Does it even need to render a group or `children`? We could just render it separately - e.g. 

```ts
<>
  <CanvasEvents {...eventHandlers} /> // suggested name (also thought of `EventsHelper`, `EventsProxy`, etc.)
  <PanMesh />
</>
```

axelboc (commented):
I think it would be nice if consumers of `InteractionsMesh` didn't have to worry about calling `getUnprojectedPoint`. The code of `InteractionsMesh` would then look something like this:

```ts
const size = useThree((state) => state.size);

const getUnprojectedPoint = useCallback((evt: PointerEvent | WheelEvent) => {
   ...
})

const handlePointerDown = useCallback((evt: PointerEvent) => {
  if (onPointerDown) {
    onPointerDown(getUnprojectedPoint(evt))
  }
});

...

useEffect(() => {
  domElement.addEventListener('pointerdown', onPointerDown); // always registered but noop if `onPointerDown` is not provided
  ...
  return () => {
    domElement.removeEventListener('pointerdown', onPointerDown);
  };
});
```

packages/lib/src/vis/shared/InteractionMesh.tsx
axelboc: We should also reconsider renaming `PanMesh` and the like later, since they will no longer contain `mesh`es.

packages/lib/src/vis/utils.ts
```diff
@@ -342,3 +342,16 @@ export function getCameraFOV(camera: Camera): {
...
+  const normX = (x - width / 2) / (width / 2);
+  const normY = -(y - height / 2) / (height / 2);
+
+  return new Vector3(normX, normY, 0).unproject(camera);
+}
```
axelboc, line 357: You can remove this utility now, I think.

axelboc (approved):
This feels right :beers: -- next up, custom hooks!

PR #996, opened 2022-03-08
Make EventsHelper a custom hook useCanvasEvents

loichuder:
Follow-up of #995


packages/lib/src/vis/utils.ts
```diff
@@ -342,3 +339,24 @@ export function getCameraFOV(camera: Camera): {
...
+export function checkModifierKey(
+  modifierKey: ModifierKey | undefined,
+  event: PointerEvent | WheelEvent
+) {
+  if (!modifierKey) {
+    return !event.altKey && !event.ctrlKey && !event.shiftKey;
+  }
+
+  return event.getModifierState(modifierKey);
+}
```
loichuder, line 362: Two new utilities to reduce the "cognitive complexity" of `Selection`. `noModifierKeyPressed` was removed as it was used only there.

Not too happy with the naming `checkModifierKey`: if you have a better name...

axelboc (approved):
Great, it's actually quite nice to keep a declarative JSX API for interactivity.

Just a couple of naming suggestions:
- `<Selection />` => `<SelectionTool />` (which allows keeping `Selection` to represent an actual selection)
- `<ZoomSelection />` => `<SelectToZoom />`

Issue #993, opened 2022-03-03
Improvements for selection-based zoom


axelboc:
- [x] #999
- [x] #998
- [x] #997
- [x] #1003
- [x] #994
- [x] #1025
- [x] #1055
- [x] #1057
- [x] #1076

loichuder:
> On the heatmap vis, it is sometimes difficult to see the selection rectangle due to poor contrast. We could imagine giving it a subtle white background?

On this, I will add that there is an inconsistent behaviour on dark mode: the rectangle color gets inverted for the `LineVis` and `Scatter` but not for the `Heatmap` nor the `RgbVis`.

axelboc:
@t20100 mentioned that in Silx and other software, when the ratio has to be kept, the area that the user selects must be fully visible once the zoom is applied. This means that the rectangle that shows the area that will be zoomed on would be the outter rectangle, not the inner rectangle. This behaviour also means that the outter selection rectangle must be properly constrained to the bounds of the visualization. Here's a mock-up for a selection where the pointer leaves the bounds of the heatmap -- the outter, continuous rectangle would be the area that is actually zoomed on:

![selection](https://user-images.githubusercontent.com/2936402/156601405-82d1cbd3-fa5c-4972-9019-692d2738d243.png)

loichuder:
:fireworks:

PR #1001, opened 2022-03-08
Draw SelectionLine and Rect with svg rather than WebGL

### LINKED ISSUE #997: The selection rectangle seems to appear underneath the tooltip guides on the heatmap vis -- it should be moved above them.

### LINKED ISSUE #998: [SelectToZoom] Fix contrast issues of the selection rectangle
On the heatmap vis, it is sometimes difficult to see the selection rectangle due to poor contrast. We could imagine giving it a subtle white background?

loichuder:
Fix #997 and https://github.com/silx-kit/h5web/issues/993#issuecomment-1058018044.

Fixes #998 as well.

### DIFF
--- a/packages/lib/src/vis/shared/SelectionLine.tsx
+++ b/packages/lib/src/vis/shared/SelectionLine.tsx
@@ -1,27 +1,13 @@
-import type { Color, Object3DNode } from '@react-three/fiber';
-import { extend, useThree } from '@react-three/fiber';
-import { useLayoutEffect, useState } from 'react';
-import type { Vector2 } from 'three';
-import { BufferGeometry, Line } from 'three';
+import { useThree } from '@react-three/fiber';
+import { Vector3 } from 'three';
 
+import type { Selection } from '../models';
 import { useAxisSystemContext } from './AxisSystemContext';
+import Html from './Html';
+import { useWorldToHtml } from './hooks';
 
-extend({ Line_: Line });
-
-// https://github.com/pmndrs/react-three-fiber/issues/1152
-declare global {
-  // eslint-disable-next-line @typescript-eslint/no-namespace
-  namespace JSX {
-    interface IntrinsicElements {
-      line_: Object3DNode<Line, typeof Line>;
-    }
-  }
-}
-
-interface Props {
-  startPoint: Vector2;
-  endPoint: Vector2;
-  color?: Color;
+interface Props extends Selection {
+  color?: string;
 }
 
 function SelectionLine(props: Props) {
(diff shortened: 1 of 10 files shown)


packages/lib/src/vis/shared/SelectionLine.tsx
```diff
@@ -31,23 +17,32 @@ function SelectionLine(props: Props) {
...
-    invalidate();
-  }, [dataGeometry, endPoint, invalidate, startPoint]);
+  const worldStartPoint = dataToWorld(dataStartPoint);
+  const htmlStartPt = worldToHtml(
+    new Vector3(worldStartPoint.x, worldStartPoint.y, 0)
+  );
+  const worldEndPoint = dataToWorld(dataEndPoint);
+  const htmlEndPt = worldToHtml(
+    new Vector3(worldEndPoint.x, worldEndPoint.y, 0)
+  );
```
loichuder, line 32: I think we could improve this. For example, making `dataToWorld` return a `Vector3` when given a `Vector3` would remove the need of the conversion to `Vector3` in the selection component.

Your thoughts ?

packages/lib/src/vis/shared/SelectionLine.tsx
```diff
@@ -31,23 +17,32 @@ function SelectionLine(props: Props) {
...
-      <lineBasicMaterial color={color} />
-    </line_>
+    <Html>
+      <svg width={width} height={height}>
+        <line
+          x1={htmlStartPt.x}
+          y1={htmlStartPt.y}
+          x2={htmlEndPt.x}
+          y2={htmlEndPt.y}
+          stroke={color}
```
loichuder, line 42: Maybe the props from the component should refer to the svg now (e.g. set `stroke` instead of `color` for the prop name).

packages/lib/src/vis/shared/SelectionRect.tsx
```diff
@@ -22,30 +21,34 @@ function SelectionRect(props: Props) {
...
+    <Html>
+      <svg width={width} height={height}>
+        <rect
+          x={Math.min(htmlStartPt.x, htmlEndPt.x)}
+          y={Math.min(htmlStartPt.y, htmlEndPt.y)}
+          width={Math.abs(htmlEndPt.x - htmlStartPt.x)}
+          height={Math.abs(htmlEndPt.y - htmlStartPt.y)}
+          stroke={borderColor}
+          fill={color}
+          fillOpacity={opacity}
```
loichuder, line 48: Same comment here about svg prop names

packages/lib/src/vis/shared/SelectionLine.tsx
axelboc: How about changing `worldToHtml` to accept a `Vector2` instead of a `Vector3` and adapt the code in `Html` accordingly?

packages/lib/src/vis/shared/SelectionLine.tsx
axelboc: Hmm, this would indeed be consistent with visx components.

packages/lib/src/vis/shared/hooks.ts
```diff
@@ -163,3 +163,16 @@ export function useCanvasEvents(callbacks: CanvasEventCallbacks) {
     };
   });
 }
+
+export function useWorldToHtml(): (point: Vector3) => Vector3 {
```
axelboc, line 167: Two suggestions:

- Provide this through context, like `worldToData` and `dataToWorld`.
- Inline `projctCameraToHtml` in the `useCallback` below since it should now only be used there.

packages/lib/src/vis/shared/SelectionLine.tsx
axelboc: Perhaps this is a bit tricky due to `Vector2` not having a `project` method, but I'm thinking that you could make a new `Vector3` in `worldToHtml` with `0` as `z` and convert back to `Vector2` before returning.

packages/lib/src/vis/shared/SelectionLine.tsx
loichuder: We could even have `Props` extend the svg props. That may be a step too far but would facilitate the use of a dashed line for #999 for example.

axelboc: Yep, I'm down with that.

axelboc (approved):
Awesome!!

PR #1002, opened 2022-03-09
Allow SelectToZoom rectangle to respect the FOV ratio

### LINKED ISSUE #999: Currently, the selection-based zoom doesn't respect the "keep ratio" setting of the heatmap vis. We could imagine tracing a selection with a dotted outter inner rectangle to show the user's real selection and a continuous inner outter rectangle constrained to the vis ratio and vis bounds to show the area that will actually be zoomed on.

loichuder:
Fix #999 

Following the process described in https://github.com/silx-kit/h5web/issues/993#issuecomment-1058190341

The behaviour is activated when checking `Keep ratio` on **Heatmap** and **RGB** vis.

### DIFF
--- a/packages/lib/src/vis/heatmap/HeatmapVis.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapVis.tsx
@@ -8,6 +8,7 @@ import {
 import type { NdArray } from 'ndarray';
 import type { ReactElement, ReactNode } from 'react';
 
+import { XAxisZoom, YAxisZoom } from '../..';
 import { useAxisDomain, useValueToIndexScale } from '../hooks';
 import type { AxisParams, VisScaleType } from '../models';
 import Pan from '../shared/Pan';
@@ -78,12 +79,14 @@ function HeatmapVis(props: Props) {
   const safeDataArray = useTextureSafeNdArray(dataArray);
   const safeAlphaArray = useTextureSafeNdArray(alpha?.array);
 
+  const keepRatio = layout !== 'fill';
+
   return (
     <figure className={styles.root} aria-label={title} data-keep-canvas-colors>
       <VisCanvas
         title={title}
         canvasRatio={layout === 'contain' ? cols / rows : undefined}
-        visRatio={layout !== 'fill' ? cols / rows : undefined}
+        visRatio={keepRatio ? cols / rows : undefined}
         abscissaConfig={{
           visDomain: abscissaDomain,
           showGrid,
@@ -100,7 +103,9 @@ function HeatmapVis(props: Props) {
       >
         <Pan />
         <Zoom />
-        <SelectToZoom />
+        <XAxisZoom disabled={keepRatio} />
+        <YAxisZoom disabled={keepRatio} />
+        <SelectToZoom keepRatio={keepRatio} />
         <ResetZoomButton />
         <TooltipMesh
           guides="both"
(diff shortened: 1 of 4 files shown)


packages/lib/src/vis/heatmap/HeatmapVis.tsx
```diff
@@ -100,7 +103,9 @@ function HeatmapVis(props: Props) {
       >
         <Pan />
         <Zoom />
-        <SelectToZoom />
+        <XAxisZoom disabled={keepRatio} />
+        <YAxisZoom disabled={keepRatio} />
```
loichuder, line 107: Now adding axial zooming for Heatmap and RGB vis when there is no need to keep the ratio

packages/lib/src/vis/shared/SelectToZoom.tsx
```diff
@@ -49,13 +55,25 @@ function SelectToZoom(props: Props) {
...
+          {keepRatio && (
+            <SelectionRect
+              startPoint={startPoint}
+              endPoint={getRatioEndPoint(startPoint, endPoint, width / height)}
+              fillOpacity={0.25}
+              fill="white"
+              stroke="black"
+            />
+          )}
+        </>
```
loichuder, line 76: The first `SelectionRect` shows the rectangle drawn by the user. The second after the `keepRatio` check shows the resulting FOV when the ratio is kept.

When `keepRatio` is `false`, the first rectangle already shows the resulting FOV so that we don't need to show the second.

This being said, I feel this part is not self-documenting enough. I should perhaps rename `endPoint` into `userEndPoint` or `pickedEndPoint` ?

packages/lib/src/vis/utils.ts
```diff
@@ -71,10 +71,9 @@ export function getSizeToFit(
 
   const availableRatio = width / height;
 
-  return {
-    width: availableRatio > ratioToRespect ? height * ratioToRespect : width,
-    height: availableRatio < ratioToRespect ? width / ratioToRespect : height,
-  };
+  return availableRatio > ratioToRespect
+    ? { width: height * ratioToRespect, height }
+    : { width, height: width / ratioToRespect };
```
loichuder, line 76: I found this part not very readable when coming back to it. I hope it's better now.

packages/lib/src/vis/utils.ts
```diff
@@ -348,3 +347,30 @@ export function checkModifierKey(
...
+      startPoint.x + widthSign * height * ratio,
+      startPoint.y + heightSign * height
+    );
+  }
+
+  return new Vector2(
+    startPoint.x + widthSign * width,
+    startPoint.y + (heightSign * width) / ratio
+  );
+}
```
loichuder, line 376: Contrary to `getSizeToFit` that returns a size included in the input size, this utility changes the `endPoint` so that the resulting rectangle **includes** the original rectangle.

packages/lib/src/vis/shared/SelectToZoom.tsx
axelboc: I think it's fine for now. It will perhaps be clearer once we start dealing the boundaries of the vis, as `getRatioEndPoint` may then have to accept and return a full selection rather than a single point (as the start point may also change). Depends on what we decide the behaviour should be, as we discussed via Zoom.

axelboc (approved):

PR #1004, opened 2022-03-10
Make selection disappear when modifier key is released

### LINKED ISSUE #1003: [SelectToZoom] Releasing the modifier key should cancel the selection
It should be possible to cancel a selection before releasing the mouse pointer, perhaps when the modifier key is released or when the escape key is pressed?

After discussion, we settled on the following behaviour: releasing the key make the zoom rectangle disappear. Then:
- Either the mouse is released and nothing happens.
- Either the modifier key is pressed again and the rectangle reappears: ending point attached to the mouse and starting point being the original one.

loichuder:
...and make it reappear when it is pressed again.

Fix #1003 

`SelectionTool` is becoming quite complex and `sonarjs` is complaining :scream_cat: 

But I want to get your opinion on the implementation before I start refactoring

### DIFF
--- a/packages/lib/src/vis/shared/SelectionTool.tsx
+++ b/packages/lib/src/vis/shared/SelectionTool.tsx
@@ -1,10 +1,11 @@
+import { useKeyboardEvent, useToggle } from '@react-hookz/web';
 import { useThree } from '@react-three/fiber';
 import type { ReactElement } from 'react';
 import { useCallback, useState } from 'react';
 import type { Vector2 } from 'three';
 
 import type { CanvasEvent, ModifierKey, Selection } from '../models';
-import { boundPointToFOV, checkModifierKey } from '../utils';
+import { boundPointToFOV } from '../utils';
 import { useAxisSystemContext } from './AxisSystemContext';
 import { useCanvasEvents } from './hooks';
 
@@ -27,28 +28,25 @@ function SelectionTool(props: Props) {
 
   const [startPoint, setStartPoint] = useState<Vector2>();
   const [endPoint, setEndPoint] = useState<Vector2>();
+  const [isAllowed, toggleAllowed] = useToggle(!modifierKey);
   const camera = useThree((state) => state.camera);
 
   const { worldToData } = useAxisSystemContext();
   const onPointerDown = useCallback(
     (evt: CanvasEvent<PointerEvent>) => {
-      const { unprojectedPoint, sourceEvent } = evt;
-
-      const isSelectionAllowed = checkModifierKey(modifierKey, sourceEvent);
-      if (!isSelectionAllowed) {
+      if (!isAllowed) {
         return;
       }
 
+      const { unprojectedPoint, sourceEvent } = evt;
       const { target, pointerId } = sourceEvent;
       (target as Element).setPointerCapture(pointerId);
 
       setStartPoint(worldToData(unprojectedPoint));
       setEndPoint(undefined);
-      if (onSelectionStart) {
-        onSelectionStart();
-      }
+      onSelectionStart?.();
     },
-    [modifierKey, worldToData, onSelectionStart]
+    [isAllowed, worldToData, onSelectionStart]
   );
 
   const onPointerMove = useCallback(
@@ -57,16 +55,16 @@ function SelectionTool(props: Props) {
         return;
       }
       const point = worldToData(boundPointToFOV(evt.unprojectedPoint, camera));
-
       setEndPoint(point);
-      if (onSelectionChange) {
+
+      if (isAllowed && onSelectionChange) {
         onSelectionChange({
           startPoint,
           endPoint: point,
         });
       }
     },
-    [camera, onSelectionChange, startPoint, worldToData]
+    [camera, isAllowed, onSelectionChange, startPoint, worldToData]
   );
 
   const onPointerUp = useCallback(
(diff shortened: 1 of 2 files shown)


packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -27,15 +28,15 @@ function SelectionTool(props: Props) {
 
   const [startPoint, setStartPoint] = useState<Vector2>();
   const [endPoint, setEndPoint] = useState<Vector2>();
+  const [modifierKeyPressed, setModifierPressed] = useState(!modifierKey);
```
loichuder, line 31: `modifierKeyPressed` is initialized to `true` if no `modifierKey` is given. Otherwise, we can't select at first render.  

When `modifierKey` is given, I initialize to `false` so that it can be set to `true` when the appropriate `keyDown` event is fired.

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -57,16 +58,20 @@ function SelectionTool(props: Props) {
         return;
       }
       const point = worldToData(boundPointToFOV(evt.unprojectedPoint, camera));
-
       setEndPoint(point);
+
+      if (!modifierKeyPressed) {
+        return;
+      }
```
loichuder, line 65: Prevent the call of `onSelectionChange` if the modifier key is not pressed but still change the endpoint thanks to `setEndpoint` above.

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -77,24 +82,49 @@ function SelectionTool(props: Props) {
       const { sourceEvent, unprojectedPoint } = evt;
       const { target, pointerId } = sourceEvent;
       (target as Element).releasePointerCapture(pointerId);
+      setStartPoint(undefined);
+      setEndPoint(undefined);
```
loichuder, line 86: Releasing the mouse should reset the selection whether the modifier key is pressed or not

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -77,24 +82,49 @@ function SelectionTool(props: Props) {
...
-      const point = worldToData(boundPointToFOV(unprojectedPoint, camera));
+      if (!modifierKeyPressed) {
+        return;
+      }
 
       if (onSelectionEnd) {
         onSelectionEnd({
           startPoint,
-          endPoint: point,
+          endPoint: worldToData(boundPointToFOV(unprojectedPoint, camera)),
```
loichuder, line 95: Inlining `point`

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -77,24 +82,49 @@ function SelectionTool(props: Props) {
...
-  if (!startPoint || !endPoint) {
+  useEffect(() => {
+    document.addEventListener('keydown', onKeyDown);
+    document.addEventListener('keyup', onKeyUp);
+
+    return () => {
+      document.removeEventListener('keydown', onKeyDown);
+      document.removeEventListener('keyup', onKeyUp);
+    };
+  });
```
loichuder, line 125: I attached key events on `document` rather than on the `canvas` because the `canvas` is not focusable and therefore does not trigger key events.

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -77,24 +82,49 @@ function SelectionTool(props: Props) {
       const { sourceEvent, unprojectedPoint } = evt;
       const { target, pointerId } = sourceEvent;
       (target as Element).releasePointerCapture(pointerId);
+      setStartPoint(undefined);
+      setEndPoint(undefined);
 
-      const point = worldToData(boundPointToFOV(unprojectedPoint, camera));
+      if (!modifierKeyPressed) {
+        return;
```
axelboc, line 89: Might be more concise to merge this with the next conditional (same in `onPointerMove`).

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -27,15 +28,15 @@ function SelectionTool(props: Props) {
 
   const [startPoint, setStartPoint] = useState<Vector2>();
   const [endPoint, setEndPoint] = useState<Vector2>();
+  const [modifierKeyPressed, setModifierPressed] = useState(!modifierKey);
   const camera = useThree((state) => state.camera);
 
   const { worldToData } = useAxisSystemContext();
   const onPointerDown = useCallback(
     (evt: CanvasEvent<PointerEvent>) => {
       const { unprojectedPoint, sourceEvent } = evt;
```
axelboc, line 37: Maybe move this destructuring below the early return?

packages/lib/src/vis/shared/SelectionTool.tsx
axelboc: I would maybe rename this state `isAllowed` - i.e. `const [isAllowed, toggleAllowed] = useToggle(!modifierKey)`, or `isSelectionAllowed` if it's too vague, to better cover the cases where no modifier key is passed and where the wrong modifier key is pressed.

packages/lib/src/vis/shared/SelectionTool.tsx
axelboc: `window` would be an even better choice. That being said, I just remembered that react-hookz provides a hook for this! https://react-hookz.github.io/web/?path=/docs/dom-usekeyboardevent--example

Can't believe I didn't think of it, but you could also use https://react-hookz.github.io/web/?path=/docs/dom-useeventlistener--example in `useCanvasEvents`!

packages/lib/src/vis/utils.ts
```diff
@@ -339,7 +339,7 @@ export function boundPointToFOV(
 
 export function checkModifierKey(
   modifierKey: ModifierKey | undefined,
-  event: PointerEvent | WheelEvent
+  event: PointerEvent | WheelEvent | KeyboardEvent
```
axelboc, line 342: Might be worth broadening the type to `MouseEvent | KeyboardEvent`

packages/lib/src/vis/shared/SelectionTool.tsx
axelboc: Second part of my comment done in #1007

packages/lib/src/vis/shared/SelectionTool.tsx
loichuder: I renamed to `isAllowed` but wasn't too sure if the implementation using `toggle` was going to work. I did it in a separate commit and it appears to work well.

Tell me what you prefer

axelboc (approved):
Looks great to me!

PR #1005, opened 2022-03-10
Add a help button to show interaction possiblities

### LINKED ISSUE #985: Make interactivity features discoverable
Features behind modifier keys like axial zoom are particularly difficult to discover. It would be good to have a help button in the breadcrumbs bar, for instance, to document mouse/keyboard interactions in a modal.

loichuder:
Fix #985 

**Line and Heatmap without `Keep Ratio` (light & dark)**

image
image

**Heatmap with `Keep Ratio` (dark)**

image

**Collapsed menu (dark)**

image


axelboc:
This is awesome. A couple of initial suggestions:

- Swap the labels and the shortcuts so that the shortcuts are on the right-hand side, like on desktop application menus. I would also use `justify-content: space-between` so the shortcuts are aligned to the right of the popup:

    image

- Compose the `popup` and `popupInner` toolbar styles to get the exact same look and feel as other toolbar popups.
- Maybe move the help button to the right of the overflow menu button so it's always visible? If you do so, the responsive behaviour will likely still work, but you'll have to move the `Separator` out of `OverflowMenu` and into `Toolbar`.

axelboc:
Btw, good call putting it in the toolbar. The breadcrumbs bar would have indeed been problematic since the interactions depend on the visualization.

packages/lib/src/toolbar/controls/InteractionHelp.module.css
```diff
@@ -0,0 +1,27 @@
...
+.menuList > li {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin: 0.25rem 0.25rem;
+  white-space: nowrap;
+}
+
+.menuList > li > kbd {
+  background-color: var(--primary-light);
```
axelboc, line 22: Since it's a lib component, you should use a customizable CSS property starting with `--h5w` with a default value that doesn't rely on `@h5web/app`'s default theme properties. Something like `var(--h5w-interactionsHelp-shortcut-bgColor, lightgray)`, and in `App.module.css`: `--h5w-interactionsHelp-shortcut-bgColor: var(--primary-light)` (not forgetting to document it in Storybook of course).

packages/lib/src/toolbar/controls/InteractionHelp.module.css
```diff
@@ -0,0 +1,27 @@
...
+
+.menuList > li {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin: 0.25rem 0.25rem;
+  white-space: nowrap;
+}
+
+.menuList > li > kbd {
```
axelboc, line 21: Not worth having `> li >` in the middle, I reckon.

packages/lib/src/index.ts
```diff
@@ -24,6 +24,7 @@ export { default as FlipYAxisToggler } from './toolbar/controls/FlipYAxisToggler
 export { default as Selector } from './toolbar/controls/Selector/Selector';
 export { default as ExportMenu } from './toolbar/controls/ExportMenu';
 export { default as CellWidthInput } from './toolbar/controls/CellWidthInput';
+export { default as InteractionHelp } from './toolbar/controls/InteractionHelp';
```
axelboc, line 27: Maybe add an `InteractionHelp` story between the `Default` and `Responsive` toolbar stories? Especially if you refactor by adding an `interactions` prop to the `Toolbar`, which gets forwarded to `InteractionHelp`. Then, you can also pass an `interactions` object to the `Responsive` story to test that the button doesn't move to the overflow menu when space is limited.

packages/lib/src/toolbar/controls/InteractionHelp.tsx
```diff
@@ -0,0 +1,61 @@
...
+      toggleHelpMenu(false);
+    }
+  });
+
+  return (
+    <div className={styles.root} ref={rootRef}>
+      <ToggleBtn
+        icon={FiHelpCircle}
+        iconOnly
+        onToggle={() => toggleHelpMenu()}
```
axelboc, line 44: `onToggle` doesn't receive an argument, so it's safe to just write `onToggle={toggleHelpMenu}`

axelboc (approved):

apps/storybook/src/Toolbar.stories.tsx
```diff
@@ -92,6 +98,16 @@ export const Default = Template.bind({});
 export const Responsive = Template.bind({});
 Responsive.args = { narrow: true };
 
+export const DocumentInteractions = Template.bind({});
+DocumentInteractions.args = {
+  narrow: true,
+  interactions: [
+    { shortcut: 'Wheel', description: 'Turn' },
+    { shortcut: 'Space', description: 'Accelerate' },
+    { shortcut: 'Ctrl+Alt', description: 'Do a backflip' },
```
axelboc, line 107: I didn't mind your previous syntax: `{ 'Wheel': 'Turn', ... }`, I thought it was more concise, but I don't have strong feelings either way.

axelboc (approved):
Looks really good :star_struck:

apps/storybook/src/Toolbar.stories.tsx
loichuder: I find that sometimes `Record` can be a bit awkward to go through and understand. 

While this is longer, I find it more explicit and it does not need `Object.values/entries` fiddling

PR #1007, opened 2022-03-10
Expose and document hook `useCanvasEvents` + misc documentation improvements

axelboc:
See individual commits.

### DIFF
--- a/apps/storybook/src/Utilities.stories.mdx
+++ b/apps/storybook/src/Utilities.stories.mdx
@@ -195,6 +195,7 @@ const {
   ordinateScale,
   dataToWorld,
   worldToData,
+  worldToHtml
 } = useAxisSystemContext();
 ```
 
--- a/apps/storybook/src/Utilities.stories.mdx
+++ b/apps/storybook/src/Utilities.stories.mdx
@@ -153,7 +153,7 @@ scaleWithMid(3); // Returns 3
 
 #### useVisibleDomains
 
-Compute the x and y-axis domains at the current camera position and zoom level.
+Compute the x and y-axis domains at the current camera position and zoom level. Must be called from a child component of `VisCanvas`.
 
 ```ts
 useVisibleDomains(): { xVisibleDomain: Domain; yVisibleDomain: Domain; }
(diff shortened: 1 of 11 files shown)


apps/storybook/src/Utilities.stories.mdx
```diff
@@ -163,12 +163,23 @@ const { xVisibleDomain, yVisibleDomain } = useVisibleDomains();
...
+```ts
+useCanvasEvents(callbacks: CanvasEventCallbacks): void
+
+const handlePointerDown = useCallback((evt: PointerEvent) => { ... }, []);
+useCanvasEvents({ onPointerDown: handlePointerDown }); // also supported: `onPointerMove`, `onPointerUp` and `onWheel`
+```
+
 ### AxisSystemContext
 
 Children of `VisCanvas` have access to the `AxisSystemContext`, which exposes the visualization's size, the axes' configurations and scales, as well as two utilities to transform a vector from world to data space and from data to world space:
```
loichuder, line 185: We could add a part about the new utility to transform a vector from world position to HTML position (top, left).

Given that the list is becoming bigger, we could make a table like https://github.com/pmndrs/react-three-fiber/blob/78046f254f6bbacec6c2d3a62f0d0f673ed5fc52/docs/API/hooks.mdx#state-properties

loichuder (approved):
Great :+1: :slightly_smiling_face:

apps/storybook/src/Utilities.stories.mdx
axelboc: agreed!

apps/storybook/src/Utilities.stories.mdx
```diff
@@ -184,9 +195,21 @@ const {
...
+| `ordinateConfig` | Ordinate configuration object passed to `VisCanvas`                                                                   | <code>AxisConfig</code>                               |
+| `abscissaScale`  | Computes the X coordinate in the HTML space of a given abscissa value (or the opposite with `abscissaScale.invert()`) | <code>AxisScale</code>                                |
+| `ordinateScale`  | Computes the Y coordinate in the HTML space of a given ordinate value (or the opposite with `ordinateScale.invert()`) | <code>AxisScale</code>                                |
+| `dataToWorld`    | Converts a vector from data space to world space                                                                      | <code>(vec: Vector2 &#124; Vector3) => Vector2</code> |
+| `worldToData`    | Converts a vector from world space to data space                                                                      | <code>(vec: Vector2 &#124; Vector3) => Vector2</code> |
+| `worldToHtml`    | Converts a vector from world space to HTML space                                                                      | <code>(vec: Vector2 &#124; Vector3) => Vector2</code> |
```
axelboc, line 211: Even with backticks, I had to use this `&#124;` hack to prevent the pipes from being interpreted as column delimiters... Then I had to change the backticks to `<code>` to actually render the pipes and not `&#124`... And then I had to use `<code>` for the other types as well, because Storybook renders them differently than code wrapped in backticks... Go figure :disappointed:

PR #1008, opened 2022-03-11
Use default line colors for better contrast and distinction of auxiliary plots

axelboc:
Part of https://github.com/silx-kit/jupyterlab-h5web/issues/82

The `--secondary-dark` color gets a bit lost if I try to give auxiliary plots more distinct colours. Changing to the default colour for the main plot, `midnightblue`, makes it pop out more, which means I can also use the default aux colours: `orangered, forestgreen, crimson, mediumslateblue, sienna`.

We've had that green for a long time, so it's going to take some getting used to, but it's not shocking. It's also closer to Matplotlib I think (that's where we got the default colours from, if I recall).

image

image


loichuder (approved):
Me likes it :sparkles:

loichuder:
Here is the aspect on the dark mode for the record:

image

I find the colours less distinguishable than on the light mode but maybe that is just my deuteranomaly :see_no_evil:

packages/lib/src/vis/hooks.ts
```diff
@@ -83,22 +83,30 @@ export function useWheelCapture() {
...
+
+  const colors = Object.entries(properties).map(([name, defaultColors]) => {
+    const [light, dark] = Array.isArray(defaultColors)
+      ? defaultColors
+      : [defaultColors, defaultColors];
+
+    return styles.getPropertyValue(name).trim() || (isDark ? dark : light);
+  });
+
+  return [colors, refCallback];
```
loichuder, line 111: What is the added value of this utility compared to our previous use of `useCSSCustomProperties` with defaults set in `LineVis` ?

If it is the change of default colors depending on the light/dark color scheme, then this can be done by changing the defaults in `LineVis` rather than encapsulating in an utility ?

axelboc: One problem was that `useCSSCustomProperties` was returning empty strings, which I thought was confusing as a way to indicate that the properties were not set. It was also forcing us to fall back to default values on separate lines instead of in the destructuring (since `''` is defined).

I also saw an opportunity to be more declarative, more like `var(--h5w-my-var, myFallback)`. With dark mode in the picture, I think this is all the more important. Detecting dark mode directly in `LineVis` would add unnecessary clutter.

But I can clean things up a bit. Hang on a sec.

packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -28,12 +28,18 @@ import styles from './LineVis.module.css';
...
+  {
+    property: '--h5w-line--color',
+    fallback: 'darkblue',
+    darkFallback: 'deepskyblue',
+  },
+  {
+    property: '--h5w-line--colorAux',
+    fallback: 'orangered, forestgreen, red, mediumorchid, olive',
+    darkFallback: 'orange, lightgreen, red, violet, gold',
+  },
```
axelboc, line 42: I've made the API of `useCustomColors` a lot less obscure. Let me know what you think.

packages/lib/src/vis/models.ts
```diff
@@ -111,3 +111,9 @@ export interface CanvasEventCallbacks {
   onPointerUp?: (evt: CanvasEvent<PointerEvent>) => void;
   onWheel?: (evt: CanvasEvent<WheelEvent>) => void;
 }
+
+export interface CustomColor {
```
loichuder, line 115: I get the impression this describes a `FallbackColor` rather than a `CustomColor`.

packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -99,18 +105,16 @@ function LineVis(props: Props) {
...
-  const {
-    colors: [curveColor, rawAuxColor],
-    refCallback: rootRef,
-  } = useCSSCustomProperties('--h5w-line--color', '--h5w-line--colorAux');
-
-  // Support comma-separated list of auxiliary colors
-  const auxColors = (rawAuxColor || DEFAULT_AUX_COLORS)
-    .split(',')
-    .map((col) => col.trim());
+  const [[curveColor, auxColorList], rootRef] = useCustomColors(COLORS);
```
loichuder, line 108:
```suggestion
  const [[curveColor, auxColorList], rootRef] = useCustomColors(FALLBACK_COLORS);
```

loichuder (approved):

packages/lib/src/vis/models.ts
axelboc: Hmm not only since it tells `useCustomColor` which color property to retrieve. I hesitated with `CustomColorDef` and `COLOR_DEFS`, is that better?

loichuder: I see. Then `CustomColor` is fine

PR #1011, opened 2022-03-14
Fix first selection not starting and flickering on drag

axelboc:
- The flicker was caused by `keydown` being triggered continuously while a key is held down, which led to `toggleAllowed` being called repeatedly. I fixed this with a simple `if` condition.
- The first selection not starting was due to `keydown` events not being fired when the window doesn't have focus. This was more challenging to fix. I had to bring back `checkModifierKey` in the event handlers so that the result of the check is always up to date. The `keydown` and `keyup` listeners now control a state called `isVisible` that defaults to `true` and is used only to decide whether or not to render the current selection. So this `isVisible` state cannot become `false` (when the modifier key is released) until the user has actually started a selection and therefore gave focus to the window. I hope this makes sense, but basically I'm isolating the behaviour of the modifier key being released during a selection from the rest of the logic.


apps/storybook/package.json
```diff
@@ -23,7 +23,7 @@
   "dependencies": {
     "@h5web/lib": "workspace:*",
     "@h5web/shared": "workspace:*",
-    "@react-hookz/web": "12.3.0",
+    "@react-hookz/web": "13.1.0",
```
axelboc, line 26: Upgrade for the new `useRafHook`, which avoids triggering more renders than the browser can actually process from `onPointerMove`: https://github.com/react-hookz/web/releases/tag/v13.1.0

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -17,6 +18,7 @@ interface Props {
   children: (points: Selection) => ReactElement;
 }
 
+// eslint-disable-next-line sonarjs/cognitive-complexity
```
axelboc, line 21: As mentioned in the chat, I think we'll be able to remove this by abstracting the drag-with-modifier-key logic.

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -26,78 +28,96 @@ function SelectionTool(props: Props) {
...
       const { target, pointerId } = sourceEvent;
       (target as Element).setPointerCapture(pointerId);
-
       setStartPoint(worldToData(unprojectedPoint));
-      setEndPoint(undefined);
-      onSelectionStart?.();
+
+      if (onSelectionStart) {
+        onSelectionStart();
+      }
```
axelboc, line 51: Reverting to a clearer syntax, since I'm going over the complexity limit anyway...

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -26,78 +28,96 @@ function SelectionTool(props: Props) {
...
+      if (!checkModifierKey(modifierKey, sourceEvent)) {
         return;
       }
 
-      const { unprojectedPoint, sourceEvent } = evt;
       const { target, pointerId } = sourceEvent;
       (target as Element).setPointerCapture(pointerId);
-
       setStartPoint(worldToData(unprojectedPoint));
-      setEndPoint(undefined);
```
axelboc, line 46: Calling `setEndPoint` here is actually not needed, since we reset the end point in `onPointerUp`.

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -2,10 +2,11 @@ import { useKeyboardEvent, useToggle } from '@react-hookz/web';
 import { useThree } from '@react-three/fiber';
 import type { ReactElement } from 'react';
 import { useCallback, useState } from 'react';
+import { useRafState } from 'react-use';
```
loichuder, line 5: Weren't you intending to use `react-hookz` version here ?

packages/lib/src/vis/shared/SelectionTool.tsx
```diff
@@ -26,78 +28,96 @@ function SelectionTool(props: Props) {
...
         return;
       }
-      const point = worldToData(boundPointToFOV(evt.unprojectedPoint, camera));
+
+      const { unprojectedPoint, sourceEvent } = evt;
+      const point = worldToData(boundPointToFOV(unprojectedPoint, camera));
       setEndPoint(point);
 
-      if (isAllowed && onSelectionChange) {
+      if (onSelectionChange && checkModifierKey(modifierKey, sourceEvent)) {
```
loichuder, line 66: It made more sense to me to have the `checkModifierKey` check before but it does not matter.

loichuder (approved):
Thanks !

packages/lib/src/vis/shared/SelectionTool.tsx
axelboc: Ouch, well spotted!!

packages/lib/src/vis/shared/SelectionTool.tsx
axelboc: Well, it's less work to do if `onSelectionChange` is not passed (which is likely).

loichuder: True :+1:

PR #1016, opened 2022-03-14
Allow switch to `NX Spectrum` when interpretation is not specified/known

### LINKED ISSUE #1006: Allow switching to NX Spectrum visualization when possible

When the `(2+n)D` signal of an `NXdata` group is not tagged with the `spectrum` interpretation, the visualizer shows only the `NX Image` visualization and does not provide a way to visualize the `NXdata` group as a spectrum.

We should be less restrictive when determining the supported visualization of an `NXdata` group.

### Additional context

Original feedback from Diamond:

> Having a quick play, the only issue I have is (and its probably because I don't add the interpretation tagging to the datasets) that multi-dimensional datasets always plot as NX Image, even if they are better represented as spectra, and I can't plot them as lines without losing the NeXus defined axis. Being able to choose between image (which is good for time series of spectra/patterns) and lines would be awesome. But that's basically it.

axelboc:
Fix #1006 

Currently, when the interpretation is not specified or known, we look at the number of dimensions. When this number is greater than or equal to 2, we show `NX Image` as the only supported vis. Now, in this particular case, we also show `NX Spectrum`. This is the only use case that changes.

I hesitated to allow switching to other NX vis when the interpretation is known, but when dealing with `NXdata` groups with 2D `spectrum` signals, this would mean selecting the first tab (`NX Spectrum`) instead of the last tab (`NX Image`) by default, which is incompatible with how the `VisManager` works. This wasn't the use case reported by Diamond anyway, so I think it's fine as it is.

After I made the change, I noticed that if I switched to `NX Spectrum` on mock entity `nexus_entry/complex_image`, I would get an error saying that the number of abscissa values did not match the number of data values. Turns out, the `position` axis dataset had one too many values, and this wasn't reported in the `NX Image` visualization because the `HeatmapVis` did not have any assertions on the length of custom axis arrays. Both problems are now fixed.

image


packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -220,8 +225,9 @@ test('visualize NXdata group with unknown interpretation', async () => {
   await selectExplorerNode('nexus_malformed/interpretation_unknown');
 
   const tabs = await findVisSelectorTabs();
-  expect(tabs).toHaveLength(1);
-  expect(tabs[0]).toHaveTextContent(NexusVis.NxImage); // support check falls back to signal dataset dimensions
+  expect(tabs).toHaveLength(2); // support check falls back to signal dataset dimensions (4D)
```
loichuder, line 228: This comment and the one below would gain to be a bit more explicit. For example, simply add `(4D supports both Image and Spectrum)`

packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -233,8 +239,8 @@ test('visualize NXdata group with "rgb-image" interpretation but incompatible si
   await selectExplorerNode('nexus_malformed/rgb-image_incompatible');
 
   const tabs = await findVisSelectorTabs();
-  expect(tabs).toHaveLength(1);
-  expect(tabs[0]).toHaveTextContent(NexusVis.NxSpectrum); // support check falls back to signal dataset dimensions
+  expect(tabs).toHaveLength(1); // support check falls back to signal dataset dimensions (1D)
```
loichuder, line 242: And here `(1D supports only Spectrum)`

PR #1019, opened 2022-03-15
Put interaction components and utils in a dedicated folder

loichuder:
I hesitate to include `Tooltip` and `ViewportCenterer` as well :shrug:


axelboc (approved):
Thanks!

axelboc:
Hmm, definitely not `ViewportCenterer` and think we can leave `Tooltip` where it is as well. I reckon we can use the help popup as reference for what is considered an interaction.

Issue #1013, opened 2022-03-14
Allow configuring interactivity features of visualization components


axelboc:
It should be possible, for instance, to change the modifier key of the select-to-zoom feature when using the `HeatmapVis`, or to disable axial zoom when using the `LineVis`.

loichuder:
Fixed by #1023

PR #1020, opened 2022-03-17
Handle interaction key conflicts by setting them in VisCanvas

### LINKED ISSUE #1015: [SelectToZoom, XAxisZoom, YAxisZoom] Default modifier key cannot be unset
These three components cannot be used without a modifier key due to the default values given to theirs `modifierKey` props.

loichuder:
Stepping stone for #1013 

Following our discussion, here is a first prototype to have interaction configuration stored at a single place (here `AxisSystemContext`) to have a place where we can handle interactions between interactions (do you follow?).

Also, this allows to have a more permissive behaviour for interactions with a `modifierKey` set to `undefined` (`Pan` in particular). They will now only be prevented if the pressed modifier key is bound to another interaction.

This fixes #1015 as default modifier keys values in these components were removed, allowing modifier keys to be set to `undefined`.

### DIFF
--- a/packages/lib/src/vis/heatmap/HeatmapVis.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapVis.tsx
@@ -9,16 +9,21 @@ import {
 import type { NdArray } from 'ndarray';
 import type { ReactElement, ReactNode } from 'react';
 
-import { XAxisZoom, YAxisZoom } from '../..';
 import Pan from '../../interactions/Pan';
 import ResetZoomButton from '../../interactions/ResetZoomButton';
 import SelectToZoom from '../../interactions/SelectToZoom';
+import XAxisZoom from '../../interactions/XAxisZoom';
+import YAxisZoom from '../../interactions/YAxisZoom';
 import Zoom from '../../interactions/Zoom';
 import { useAxisDomain, useValueToIndexScale } from '../hooks';
 import type { AxisParams, VisScaleType } from '../models';
 import TooltipMesh from '../shared/TooltipMesh';
 import VisCanvas from '../shared/VisCanvas';
-import { DEFAULT_DOMAIN, formatNumType } from '../utils';
+import {
+  DEFAULT_DOMAIN,
+  DEFAULT_INTERACTIONS_KEYS,
+  formatNumType,
+} from '../utils';
 import ColorBar from './ColorBar';
 import HeatmapMesh from './HeatmapMesh';
 import styles from './HeatmapVis.module.css';
@@ -103,6 +108,7 @@ function HeatmapVis(props: Props) {
           label: ordinateLabel,
           flip: flipYAxis,
         }}
+        interactionKeys={DEFAULT_INTERACTIONS_KEYS}
       >
         <Pan />
         <Zoom />
--- a/packages/lib/src/vis/shared/AxisSystemProvider.tsx
+++ b/packages/lib/src/vis/shared/AxisSystemProvider.tsx
@@ -3,6 +3,7 @@ import type { PropsWithChildren } from 'react';
 import { useCallback, useMemo } from 'react';
 import { Matrix4, Vector2, Vector3 } from 'three';
 
+import type { InteractionKeys, ModifierKey } from '../../interactions/models';
 import type { AxisConfig } from '../models';
 import { getCanvasScale, getSizeToFit } from '../utils';
 import { AxisSystemContext } from './AxisSystemContext';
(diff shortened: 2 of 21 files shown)


packages/lib/src/vis/heatmap/HeatmapVis.tsx
```diff
@@ -8,16 +8,21 @@ import {
 import type { NdArray } from 'ndarray';
 import type { ReactElement, ReactNode } from 'react';
 
-import { XAxisZoom, YAxisZoom } from '../..';
 import Pan from '../../interactions/Pan';
 import ResetZoomButton from '../../interactions/ResetZoomButton';
 import SelectToZoom from '../../interactions/SelectToZoom';
+import XAxisZoom from '../../interactions/XAxisZoom';
+import YAxisZoom from '../../interactions/YAxisZoom';
```
loichuder, line 15: Forgot to put proper imports there

packages/lib/src/vis/shared/AxisSystemProvider.tsx
```diff
@@ -46,6 +54,20 @@ function AxisSystemProvider(props: PropsWithChildren<Props>) {
     [camera, cameraToHtmlMatrix]
   );
 
+  const shouldInteract = useCallback(
+    (id: string, event: MouseEvent | WheelEvent) => {
```
axelboc, line 58: Just `MouseEvent` here, I think?

packages/lib/src/vis/shared/AxisSystemProvider.tsx
```diff
@@ -46,6 +54,20 @@ function AxisSystemProvider(props: PropsWithChildren<Props>) {
...
+      if (!Object.keys(interactionKeys).includes(id)) {
+        throw new Error(`Interaction ${id} was not registered in VisCanvas.`);
+      }
+      const modifierKeyPressed = Object.values(interactionKeys).find(
+        (key) => !!key && event.getModifierState(key)
+      );
+
+      return modifierKeyPressed === interactionKeys[id];
```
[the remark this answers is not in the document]
axelboc: Although I get why this works, I feel like it would be clearer to treat the two cases (whether `interactionKeys[id]` is defined or not) separately: when defined, you just return `event.getModifierKey(interactionKeys[id])`; otherwise, you return `Object.values(interactionKeys).find(...)`. Makes sense?

To deal with the case where two interactions are bound to the same key, you could, before the `useCallback`, initialise an array of all the modifier keys in the `interactionKeys` object. Then, you put that in a `Set` and compare the length of the array with the size of the set and throw if they're different. Then, you can loop through the set where you need to inside the `useCallback`.

When it comes to pressing multiple keys, I don't see it as a problem - e.g. `Alt + Shit + Wheel` is just equivalent to normal zoom. If you put the normal zoom on `Alt` and keep the Y-axis zoom on `Shift`, then you just do a normal zoom (I think?) I guess it becomes a bit more problematic with drag interactions? Not sure. Either way, I'm more for waiting to see if a problematic combination comes up.

packages/lib/src/vis/utils.ts
```diff
@@ -34,6 +35,14 @@ export const CAMERA_TOP_RIGHT = new Vector3(1, 1, 0);
 
 const AXIS_OFFSETS = { vertical: 72, horizontal: 40, fallback: 16 };
 
+export const DEFAULT_INTERACTIONS_KEYS: InteractionKeys = {
+  Pan: undefined,
+  Zoom: undefined,
+  XAxisZoom: 'Alt',
+  YAxisZoom: 'Shift',
+  SelectToZoom: 'Control',
+};
```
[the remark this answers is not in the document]
axelboc: Just a thought: I wonder whether `{ Pan: true }` would feel like a more natural API. I'm not a huge fan of relying on the fact that the property `Pan` is present in the array even though it is set to `undefined`. When we start passing objects with `modifierKey` and `disabled`, it may still be convenient to allow passing `Pan: true` as a shortcut for `Pan: {}` (i.e. enabled with no modifier key).

axelboc (approved):

packages/lib/src/vis/shared/AxisSystemProvider.tsx
```diff
@@ -46,6 +55,50 @@ function AxisSystemProvider(props: PropsWithChildren<Props>) {
...
+  }
+
+  const shouldInteract = useCallback(
+    (id: string, event: MouseEvent) => {
+      if (!Object.keys(interactionKeys).includes(id)) {
+        throw new Error(`Interaction ${id} was not registered in VisCanvas.`);
+      }
+
+      const interactionKey = interactionKeys[id];
+      if (isBoolean(interactionKey)) {
```
axelboc, line 73: `interactionKey === true` (see comment in `models`)

axelboc (approved):

PR #1021, opened 2022-03-21
Disable interactions via VisCanvas

loichuder:
Work towards #1020 

The last step will be to allow interactions to be changed through a prop of `...Vis` components that will be forwarded to `VisCanvas`

### DIFF
--- a/packages/lib/src/interactions/Pan.tsx
+++ b/packages/lib/src/interactions/Pan.tsx
@@ -6,12 +6,7 @@ import { useAxisSystemContext } from '../vis/shared/AxisSystemContext';
 import { useCanvasEvents, useMoveCameraTo } from './hooks';
 import type { CanvasEvent } from './models';
 
-interface Props {
-  disabled?: boolean;
-}
-
-function Pan(props: Props) {
-  const { disabled } = props;
+function Pan() {
   const { shouldInteract } = useAxisSystemContext();
 
   const camera = useThree((state) => state.camera);
@@ -25,16 +20,12 @@ function Pan(props: Props) {
       const { unprojectedPoint, sourceEvent } = evt;
       const { target, pointerId } = sourceEvent;
 
-      if (disabled) {
-        return;
-      }
-
       if (shouldInteract('Pan', sourceEvent)) {
         (target as Element).setPointerCapture(pointerId); // https://stackoverflow.com/q/28900077/758806
         startOffsetPosition.current = unprojectedPoint.clone();
       }
     },
-    [shouldInteract, disabled]
+    [shouldInteract]
   );
 
   const onPointerUp = useCallback((evt: CanvasEvent<PointerEvent>) => {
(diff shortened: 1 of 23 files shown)


packages/lib/src/vis/rgb/RgbVis.tsx
```diff
@@ -59,12 +58,18 @@ function RgbVis(props: Props) {
...
           flip: true,
         }}
-        interactionKeys={DEFAULT_INTERACTIONS_KEYS}
+        interactionKeys={{
+          Pan: true,
+          Zoom: true,
+          XAxisZoom: { modifierKey: 'Alt', disabled: keepRatio },
+          YAxisZoom: { modifierKey: 'Shift', disabled: keepRatio },
+          SelectToZoom: { modifierKey: 'Control' },
+        }}
```
loichuder, line 67: I could not use `DEFAULT_INTERACTIONS_KEYS` here due to dynamic setting of `disabled`.

Do you think it is worth putting the creating of an object in an utility ?

packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -16,7 +16,7 @@ interface Props {
   visRatio?: number | undefined;
   abscissaConfig: AxisConfig;
   ordinateConfig: AxisConfig;
-  interactionKeys?: InteractionKeys;
+  interactionKeys?: InteractionParams;
```
axelboc, line 19: What do you think of `interactions: Interactions` or `InteractionsConfig`?

axelboc (approved):

packages/lib/src/vis/shared/VisCanvas.tsx
loichuder: I would like to put `Interactions` but there is a `Interaction` interface for the interaction help button that is different. This being said, we can rename `Interaction` into `InteractionHelp` or something like that

packages/lib/src/interactions/Pan.tsx
```diff
@@ -47,7 +38,7 @@ function Pan(props: Props) {
 
   const onPointerMove = useCallback(
     (evt: CanvasEvent<PointerEvent>) => {
-      if (disabled || !startOffsetPosition.current) {
+      if (!startOffsetPosition.current) {
```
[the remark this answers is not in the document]
loichuder: Not too sure: the movement of the camera feels odd when releasing, moving the mouse and pressing again.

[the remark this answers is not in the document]
axelboc: Hmm yeah, maybe more a use case for `useKeyboardEvent` and reseting the `startOffsetPosition` when the modifier key is released. User then has to release the pointer and restart a drag gesture for the pan to start again. For another PR.

packages/lib/src/vis/shared/VisCanvas.tsx
loichuder: Renamed `InteractionParams` to `Interactions` and `Interaction` to `InteractionInfo`

PR #1023, opened 2022-03-22
Register interaction in interaction components through a custom hook

loichuder:
Interaction components have their props back (`modifierKey` and `disabled`).

The registration of interactions is no longer done by `VisCanvas` but in each component using a custom hook that registers the interaction in the `AxisSystemContext`. `VisCanvas` therefore loses its `interactions` props

I also added a `DefaultInteractions` component to deduplicate the instantiations of interaction components in our vis. Props can be given to this component to modify the default components. This brings a bit of complexity in `DefaultInteractions`.

### DIFF
--- a/packages/lib/src/interactions/DefaultInteractions.tsx
+++ b/packages/lib/src/interactions/DefaultInteractions.tsx
@@ -1,0 +1,51 @@
+import { merge } from 'lodash';
+
+import Pan from '../interactions/Pan';
+import SelectToZoom from '../interactions/SelectToZoom';
+import XAxisZoom from '../interactions/XAxisZoom';
+import YAxisZoom from '../interactions/YAxisZoom';
+import Zoom from '../interactions/Zoom';
+import type { Interactions } from './models';
+import { getDefaultInteractions } from './utils';
+
+interface Props {
+  interactions?: Interactions;
+  keepRatio?: boolean;
+}
+
+function DefaultInteractions(props: Props) {
+  const { interactions: givenInteractions = {}, keepRatio } = props;
+
+  const parsedInteractions = Object.fromEntries(
+    Object.entries(givenInteractions).map(([k, v]) => {
+      if (v === true) {
+        return [k, {}];
+      }
+
+      if (v === false) {
+        return [k, null];
+      }
+
+      return [k, v];
+    })
+  );
+
+  const interactions = merge(
+    getDefaultInteractions(keepRatio),
+    parsedInteractions
+  );
+
+  return (
+    <>
+      {interactions.Pan && <Pan {...interactions.Pan} />}
+      {interactions.Zoom && <Zoom {...interactions.Zoom} />}
+      {interactions.XAxisZoom && <XAxisZoom {...interactions.XAxisZoom} />}
+      {interactions.YAxisZoom && <YAxisZoom {...interactions.YAxisZoom} />}
+      {interactions.SelectToZoom && (
+        <SelectToZoom {...interactions.SelectToZoom} keepRatio={keepRatio} />
+      )}
+    </>
+  );
+}
+
+export default DefaultInteractions;
--- a/packages/lib/src/interactions/utils.ts
+++ b/packages/lib/src/interactions/utils.ts
@@ -4,7 +4,7 @@ import type { Vector3 } from 'three';
 import { Vector2 } from 'three';
 
 import { getCameraFOV } from '../vis/utils';
-import type { ModifierKey } from './models';
+import type { Interaction, ModifierKey } from './models';
 
 export function boundPointToFOV(
   unboundedPoint: Vector2 | Vector3,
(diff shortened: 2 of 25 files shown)


packages/lib/src/vis/shared/AxisSystemProvider.tsx
```diff
@@ -55,46 +48,43 @@ function AxisSystemProvider(props: PropsWithChildren<Props>) {
...
-
-      if (params !== true && params.disabled) {
+      const registeredKeys = [...interactionMap.values()]
+        .map((params: Interaction) => params.modifierKey)
+        .filter(isDefined);
+
+      const params = interactionMap.get(id);
+      assertDefined(
+        params,
+        `Interaction ${id} was not registered in VisCanvas.`
```
axelboc, line 62: I would use present tense here and drop `in VisCanvas`: `Interaction ${id} is not registered`.

packages/lib/src/vis/shared/AxisSystemProvider.tsx
```diff
@@ -55,46 +48,43 @@ function AxisSystemProvider(props: PropsWithChildren<Props>) {
...
+    [interactionMap]
   );
 
+  const registerInteraction = (id: string, value: Interaction) => {
+    interactionMap.set(id, value);
+  };
+
+  const unregisterInteraction = (id: string) => {
+    interactionMap.delete(id);
+  };
```
axelboc, line 86: Maybe wrap those with `useCallback` as well.

axelboc (changes requested):
It definitely feels right, now.

I just think the code in `DefaultInteractions` could be refactored a little thanks to loadsh, but don't worry about it if TS makes it too difficult.

Also, just wondering if we could add a little story to demonstrate how to implement a custom interaction with the new API, like, I don't know, a story that turns off panning and instead renders a component that draws a small rectangle around the mouse pointer during a drag. Just a thought, maybe for another PR later on.

packages/lib/src/interactions/DefaultInteractions.tsx
```diff
@@ -0,0 +1,85 @@
...
+  interactions?: {
+    Pan?: Interaction | boolean;
+    Zoom?: Interaction | boolean;
+    XAxisZoom?: Interaction | boolean;
+    YAxisZoom?: Interaction | boolean;
+    SelectToZoom?:
+      | {
+          modifierKey?: ModifierKey;
+          disabled?: boolean;
+          keepRatio?: boolean;
```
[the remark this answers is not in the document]
loichuder: I made `keepRatio` a separate prop. It made more sense to me.

packages/lib/src/vis/utils.ts
```diff
@@ -35,14 +34,6 @@ export const CAMERA_TOP_RIGHT = new Vector3(1, 1, 0);
 
 const AXIS_OFFSETS = { vertical: 72, horizontal: 40, fallback: 16 };
 
-export const DEFAULT_INTERACTIONS: Interactions = {
-  Pan: true,
-  Zoom: true,
-  XAxisZoom: { modifierKey: 'Alt' },
-  YAxisZoom: { modifierKey: 'Shift' },
-  SelectToZoom: { modifierKey: 'Control' },
-};
```
loichuder, line 44: Moved to `interactions/utils` as it is used in `interactions/DefaultInteractions` now

packages/lib/src/interactions/utils.ts
```diff
@@ -53,3 +53,15 @@ export function getRatioEndPoint(
...
+  keepRatio?: boolean
+): Record<string, Interaction> {
+  return {
+    Pan: {},
+    Zoom: {},
+    XAxisZoom: { modifierKey: 'Alt', disabled: keepRatio },
+    YAxisZoom: { modifierKey: 'Shift', disabled: keepRatio },
+    SelectToZoom: { modifierKey: 'Control' },
+  };
+}
```
loichuder, line 67: I made `DEFAULT_INTERACTIONS` a function for two reasons:
- It can disable dynamically `XAxisZoom` and `YAxisZoom` depending on `keepRatio`
- It returns a new object at each call which avoid mutation by lodash's `merge` in `DefaultInteractions`

packages/lib/src/interactions/DefaultInteractions.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { interactions: givenInteractions = {}, keepRatio } = props;
+
+  const parsedInteractions = Object.fromEntries(
+    Object.entries(givenInteractions).map(([k, v]) => {
+      if (v === true) {
+        return [k, {}];
+      }
+
+      if (v === false) {
+        return [k, null];
```
loichuder, line 26: I put `null` instead of `undefined` as fields with `undefined` will not overwrite the default value:

> Source properties that resolve to undefined are skipped if a destination value exists.
>
> https://lodash.com/docs/4.17.15#merge

axelboc (approved):

PR #1026, opened 2022-03-23
Fix ratio of SelectToZoom for NXImage

loichuder:
While working on https://github.com/silx-kit/h5web/issues/1025, I noticed a small bug for NXImage with axes.
I captured it below: the ratio of the rectangle is not computed properly while it is for the image without axes.

Peek 2022-03-23 09-53

The cause is quite simple: the part that computes the endpoint to respect the ratio works in the data space but we gave the ratio in the world space (`width / height`). 

Most of the time, this works fine because the ratio in the world space is the same as the ratio in the data space (i.e. the changes in width/height are proportional to the changes in x/y). This obviously changes when using axes in NeXus visualisations where the ratio between abscissas and ordinates values can be different than the ratio between width and `height`.

So two ways of solving this:
- Keep the ratio in the world space but do the computation with vectors in the world space
- Do the computation with the ratio in the data space

I chose the latter solution

### DIFF
--- a/packages/lib/src/interactions/SelectToZoom.tsx
+++ b/packages/lib/src/interactions/SelectToZoom.tsx
@@ -1,5 +1,6 @@
 import { useThree } from '@react-three/fiber';
 
+import { useVisibleDomains } from '../vis/hooks';
 import { useAxisSystemContext } from '../vis/shared/AxisSystemContext';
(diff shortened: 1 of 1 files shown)


axelboc (approved):
Nice, self-contained fix 👏

PR #1030, opened 2022-03-24
Refactor `ResetZoomButton` with new `FloatingControl` component

axelboc:
The `FloatingControl` component uses a new, barebone version of `Html`. The old `Html` is now renamed `Overlay` for now.  `FloatingControl` renders its children with a portal in the `floatingToolbar` element rendered by `VisCanvas`. This allows floating controls to all end up in the same container, which means we can lay them out next to one another with flexbox, for instance.

Of course, this is only the first step; I haven't yet abstracted floating button styles into a `FloatingBtn` component, for instance. I did add a story, though, just to demonstrate how a consumer might create a custom floating control.

I've also added a story to document the new `Html` component. Next step will be to refactor components that use the old `Html`/`Overlay` component.


packages/lib/src/vis/shared/Html.tsx
```diff
@@ -8,14 +8,15 @@ import {
...
-import type { HTMLAttributes } from 'react';
+import type { HTMLAttributes, RefCallback } from 'react';
 import ReactDOM from 'react-dom';
 import type { Group } from 'three';
 import { Vector3 } from 'three';
 
 import { useAxisSystemContext } from './AxisSystemContext';
 
 interface Props extends HTMLAttributes<HTMLDivElement> {
+  refCallback?: RefCallback<HTMLDivElement>;
```
axelboc, line 19: I need to pass a `ref` to `Html` so I can get a reference to the DOM element of the floating toolbar.

However, the vis canvas is rendered dynamically into the DOM, so the ref is `null` on first render. `FloatingControl` needs to know when this ref has been set, so the only way to do that is to use a ref callback. To simplify typings, I decided to use a prop, as `forwardRef` allows for a lot more types of ref to be passed. This technique is mentioned here: https://reactjs.org/docs/refs-and-the-dom.html#exposing-dom-refs-to-parent-components

- About callback refs: https://reactjs.org/docs/refs-and-the-dom.html#callback-refs
- About re-rendering when ref gets set: https://reactjs.org/docs/hooks-reference.html#useref (cf. note at bottom of section)

packages/lib/src/vis/shared/Html.tsx
```diff
@@ -93,11 +95,7 @@ function Html(props: Props) {
     }
 
     return () => {
-      if (parentElement) {
-        el.remove();
-      }
-
-      ReactDOM.unmountComponentAtNode(el);
+      el.remove();
```
axelboc, line 98: This code initially came [from drei](https://github.com/pmndrs/drei/blob/master/src/web/Html.tsx#L174), but doing the unmount in the clean-up of this effect was problematic for me in `FloatingControl`, where the `floatingToolbar` is a dependency of the `useLayoutEffect`. When `floatingToolbar` becomes defined and everything re-renders, the effect clean-up function gets run, which unmounts the control from the DOM. Since the effect that mounts the control doesn't run again during this re-render, the control remains unmounted...

This works in drei, because the ref is set on initial mount, not on subsequent re-render. Anyway, so I moved the unmount in the clean-up function of the `useEffect` that does the mount... Seems more logical.

Note also that `if (parentElement)` is unnecessary, since calling `remove()` on an already-removed element has no effect.

packages/lib/src/vis/shared/Html.tsx
```diff
@@ -114,7 +112,10 @@ function Html(props: Props) {
 
     ReactDOM.render(
       <div
-        ref={innerRef}
+        ref={(elem) => {
+          refCallback?.(elem);
+          innerRef.current = elem || undefined;
```
axelboc, line 117: We already have an internal ref for this element, so here I'm passing a callback to `ref` that sets both the external ref and the internal ref.

apps/storybook/src/FloatingToolbar.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+    >
+      <FloatingControl>
+        <button
+          type="button"
+          style={{ fontWeight: isToggled ? 'bold' : undefined }}
+          onClick={() => toggle()}
+        >
+          Toggle me!
+        </button>
+      </FloatingControl>
```
loichuder, line 31: Could we add several controls to check the toolbar behaviour more thoroughly ?

packages/lib/src/toolbar/floating/FloatingControl.module.css
```diff
@@ -0,0 +1,17 @@
...
+  display: none;
+}
+
+.btnLike {
+  composes: btnLike from '../Toolbar.module.css';
+  height: 1.625rem !important;
+  background-color: rgba(255, 255, 255, 0.7);
+  box-shadow: rgba(0, 0, 0, 0.1) 0px 0px 0px 1px,
+    rgba(0, 0, 0, 0.1) 0px 4px 11px;
+}
```
loichuder, line 17: I am not convinced we will want to add these styles to all floating buttons. We can keep it like this for now though.

packages/lib/src/toolbar/floating/FloatingControl.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+function FloatingControl(props: Props) {
+  const { children } = props;
+  const { floatingToolbar } = useContext(AxisSystemContext);
+
+  // Container `div` for ReactDOM to render the control into, appended to floating toolbar
+  const [controlContainer] = useState(() => {
+    const el = document.createElement('div');
+    el.style.cssText = 'pointer-events: auto;'; // toolbar uses `Html` which disables `pointer-events`
```
loichuder, line 18: Remind me why `Html` needs to disable `pointer-events` ?

packages/lib/src/toolbar/floating/FloatingControl.tsx
```diff
@@ -0,0 +1,39 @@
...
+  }, [controlContainer, floatingToolbar]);
+
+  useEffect(() => {
+    ReactDOM.render(<>{children}</>, controlContainer); // eslint-disable-line react/jsx-no-useless-fragment
+    return () => {
+      ReactDOM.unmountComponentAtNode(controlContainer);
+    };
+  }, [children, controlContainer]);
+
+  return <group />;
```
loichuder, line 36: Why is it needed to return a `group` here ?

loichuder (approved):

packages/lib/src/toolbar/floating/FloatingControl.tsx
axelboc: Because it adds a `div` on top of the whole canvas. `Html` does a lot more than needed in some cases, so it may need to be split at some point.

axelboc:
I'm getting an infinite loop in `Html` in the Storybook on Linux caused by the move of `unmountComponentAtNode` into the other `useEffect`. I need to try to understand what's going on a bit more.

packages/lib/src/vis/shared/Html.tsx
```diff
@@ -1,145 +1,38 @@
...
-
-  const getGroupPosition = useCallback(() => {
-    const worldPos = new Vector3().setFromMatrixPosition(
-      group.current.matrixWorld
-    );
-    const htmlPos = worldToHtml(worldPos);
-
-    return [htmlPos.x, htmlPos.y];
-  }, [worldToHtml]);
+  const { container = gl.domElement.parentElement, children } = props;
```
axelboc, line 13: I recommend looking at the diff side by side. So now `Html` can be rendered outside of `VisCanvas`, which opens the door to some interesting refactorings in `AxisSystem`.

apps/storybook/package.json
```diff
@@ -31,6 +31,7 @@
     "react-dom": "17.0.2",
     "react-icons": "4.3.1",
     "react-suspense-fetch": "0.4.0",
+    "react-three-fiber": "6.0.13",
```
loichuder, line 34: `@react-three/fiber` is already installed so I think this is the old package ?

apps/storybook/src/Utilities.stories.mdx
```diff
@@ -199,16 +199,16 @@ const {
...
+| Name             | Description                                                                                                       | Type                                                  |
+| ---------------- | ----------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------- |
+| `visSize`        | Visualization size (different from canvas size on ratio-constrained visualizations, like `HeatmapVis`)            | <code>Size</code>                                     |
+| `abscissaConfig` | Abscissa configuration object passed to `VisCanvas`                                                               | <code>AxisConfig</code>                               |
+| `ordinateConfig` | Ordinate configuration object passed to `VisCanvas`                                                               | <code>AxisConfig</code>                               |
+| `abscissaScale`  | Computes the X coordinate in data space of a given abscissa value (or the opposite with `abscissaScale.invert()`) | <code>AxisScale</code>                                |
```
loichuder, line 207: This computes the X coordinate in **world** space of a abscissa value (therefore in the **data** space).

And same below.

packages/lib/src/vis/shared/Overlay.tsx
```diff
@@ -0,0 +1,153 @@
...
+interface Props extends HTMLAttributes<HTMLDivElement> {
+  refCallback?: RefCallback<HTMLDivElement>;
+  groupProps?: GroupProps;
+  followCamera?: boolean;
+  scaleOnZoom?: boolean;
+}
+
+// Customised version of drei's `<Html>` component
+// https://github.com/pmndrs/drei/blob/v6.0.3/src/web/Html.tsx
+function Overlay(props: Props) {
```
loichuder, line 27: I guess this is a copy of the old `Html` and that you will use the new `Html` in a future PR.

packages/lib/src/vis/shared/Html.tsx
```diff
@@ -1,145 +1,38 @@
...
-      innerRef.current.style.transform = getInnerDivTransform();
-    }
-  });
+    return () => {
+      renderTarget.remove();
+    };
+  }, [container, renderTarget]);
 
-  return <group ref={group} {...groupProps} />;
+  return null;
```
loichuder, line 35: Really dumb question: I guess returning `createPortal(children, container)` doesn't work as it gets picked up by R3F renderer and not ReactDOM's ?

loichuder (approved):
Looks good to me. Let me test it on my end and it should be good to merge.

apps/storybook/package.json
axelboc: Hmm yep :scream:

apps/storybook/src/Utilities.stories.mdx
axelboc: :expressionless:

packages/lib/src/vis/shared/Overlay.tsx
axelboc: Indeed.

packages/lib/src/vis/shared/Html.tsx
axelboc: Exactly! R3F does have its own `createPortal`, but the problem is also with the `children`, which contain HTML markup instead of R3F markup.

PR #1033, opened 2022-03-25
Make selection rectangle correspond to zoom window when respecting ratio

### LINKED ISSUE #1025: When a vis ratio is set and must be respected, the actual zoom rectangle should be bound by the vis limits (and not the canvas limits)

loichuder:
Fix #1025 

I also put the ratio-respecting `SelectionRect` in its own component. This is a separate commit if you want to read the algorithmic changes first.

As discussed, I could not reuse directly the code in `moveCameraTo` as it relies on the camera FOV which does not change when drawing rectangles. Instead, I made new utilities to clamp a rectangle in an area.

Given a starting and an end point of a rectangle, the utility will compute the new center of the rectangle so that the rectangle with its actual height/width stays inside the area (the vis in our case). For the selection rectangle, I then compute the shift between the new center and the old one and apply it to the starting and end point. As this is a simple shift, it conserves the ratio of the rectangle.

I could refactor the code in `moveCameraTo` to make use of the new utilities but I wasn't sure it was worth it.


packages/lib/src/interactions/utils.ts
```diff
@@ -16,17 +17,6 @@ export function boundPointToFOV(
...
-  modifierKey: ModifierKey | undefined,
-  event: MouseEvent | KeyboardEvent
-) {
-  if (!modifierKey) {
-    return !event.altKey && !event.ctrlKey && !event.shiftKey;
-  }
-
-  return event.getModifierState(modifierKey);
-}
-
```
loichuder, line 29: This isn't used anymore after #1023

packages/lib/src/interactions/utils.ts
```diff
@@ -65,3 +55,45 @@ export function getDefaultInteractions(
...
+  endPoint: Vector2,
+  visSize: Size
+) {
+  const { center, ...rectSize } = getEnclosedRectangle(startPoint, endPoint);
+
+  const newCenter = clampRectangleToArea(center, rectSize, visSize);
+
+  const shift = newCenter.clone().sub(center);
+  return [startPoint, endPoint].map((v) => v.clone().add(shift));
+}
```
loichuder, line 99: Not really convinced on the function separation here.

I would like to keep `getEnclosedRectangle` to reuse it in `SelectToZoom` but  `clampRectangleToArea` can be inlined if we don't plan on reusing it in `moveCameraTo`.

packages/shared/src/mock/metadata.ts
```diff
@@ -159,6 +159,14 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
...
           ),
         }),
+        makeNxDataGroup('descending-axes', {
+          signal: makeDataset('twoD', intType, [20, 41]),
+          axes: {
+            X: makeNxDataset('X', intType, [41], { valueId: 'X_desc' }),
+            Y: makeNxDataset('Y', intType, [20], { valueId: 'Y_desc' }),
+          },
+          axesAttr: ['Y', 'X'],
+        }),
```
loichuder, line 169: Added a NXData group with descending axes in X and Y for more thorough checks

packages/lib/src/interactions/utils.ts
axelboc: I'm curious as to how you would use `clampRectangleToArea` in `moveCameraTo`. The code is so similar, it'd be a shame to duplicate it. But if you think it's not worth it, then inlining would indeed make sense here.

axelboc (approved):
Nicely done :clap:

packages/lib/src/interactions/utils.ts
loichuder: My first implementation felt too convoluted which why I did not include it in the PR.

I tried again this morning and it went smoothly. I guess that's why I should not try too hard Friday afternoon :joy:

packages/lib/src/interactions/hooks.ts
```diff
@@ -23,22 +22,22 @@ export function useMoveCameraTo() {
...
 
-      const xBound = Math.max(visWidth / 2 - cameraLocalBounds.x, 0);
-      const yBound = Math.max(visHeight / 2 - cameraLocalBounds.y, 0);
-
-      position.set(
-        clamp(x, -xBound, xBound),
-        clamp(y, -yBound, yBound),
-        position.z
+      const clampedPosition = clampRectangleToArea(
+        new Vector2(x, y),
```
axelboc, line 30: Awesome! I think this even opens the door, in a future PR, to let `moveCameraTo` and `clampRectangleToArea` accept `Vector2 | Vector3`.

Only minor thing that bugs me is the naming of `clampRectangleToArea`, which doesn't fit the return type (`Vector2`). Perhaps `clampPositionToArea` would make more sense?

axelboc (approved):

PR #1035, opened 2022-03-28
Allow to set axes config `visDomain` to any Domain 

### LINKED ISSUE #987: Enable non-index coordinates with TiledHeatmap

As it is, the TiledHeatmap only works when the axes coordinates matches the base image layer size since the subsampled layer is choosen according to the current axes coordinates per pixel resolution

I would need to use different coordinates, i.e., to use meaningful coordinates rather than the image pixel coordinates.

From @loichuder (https://github.com/silx-kit/h5web/pull/967#discussion_r808743511):
> We will discuss in the next PR but adding support for non-index coordinates would only need a call to useValueToIndexScale here to transform visible data domains into visible index domains.

I also considered retrieving the transforms from the image to world coordinates from three.js to figure out the current image resolution in an independent manner thus not relying on the axes coordinates.

t20100:
This PR allows to set the `visDomain` freely.
Beforehand, the `visDomain`s needed to be `[0, number of image pixel]`

This is done by changing the hooks used to retrieve the visible domains and resolution.

closes #987


apps/storybook/src/TiledHeatmap.stories.tsx
```diff
@@ -116,23 +116,28 @@ class MandelbrotTilesApi extends TilesApi {
     return this.store.get({ layer, offset });
   }
 }
+
+const X_DOMAIN: Domain = [-2, 1];
+const Y_DOMAIN: Domain = [-1.5, 1.5];
```
t20100, line 121: Alternative was to use `api.xDomain` and `api.yDomain` from the `MandelbrotTilesApi`, but it requires to override the props interface provided for the `Template`.

packages/lib/src/vis/scaleGamma.ts
```diff
@@ -10,11 +10,7 @@ import type {
 import { isNumber } from 'lodash';
 
 import type { ScaleGammaConfig } from './models';
-
-function normalize(a: number, b: number) {
-  const extent = b - a;
-  return (x: number) => (extent !== 0 ? (x - a) / extent : 0.5);
-}
+import { normalize } from './utils';
```
t20100, line 13: Moved so that I can also use it.

packages/lib/src/vis/tiles/TiledHeatmap.tsx
```diff
@@ -23,14 +23,20 @@ function TiledHeatmap(props: Props) {
...
 
   const { visSize } = useAxisSystemContext();
-  const { xDataPerPixel, yDataPerPixel } = useDataPerPixel();
+  const { xNormPerPixel, yNormPerPixel } = useNormalizedResolution();
 
-  const { baseLayerIndex } = api;
+  const { baseLayerIndex, baseLayerSize } = api;
+  const { width, height } = baseLayerSize;
+
+  const itemPerPixel = Math.max(
```
t20100, line 31: This is now the resolution given as the number of image array items per screen pixel

apps/storybook/src/TiledHeatmap.stories.tsx
loichuder: I don't understand. You can use `api.xDomain` instead of `X_DOMAIN` inside the `Template` without changing the `props`, right?

Also, perhaps it is worth having a different story for non-pixel domain.

packages/lib/src/vis/tiles/hooks.ts
```diff
@@ -1,13 +1,34 @@
...
+export function useNormalizedVisibleDomains() {
   const { xVisibleDomain, yVisibleDomain } = useVisibleDomains();
+
+  const { abscissaConfig, ordinateConfig } = useAxisSystemContext();
+  const { visDomain: abscissaVisDomain } = abscissaConfig;
+  const { visDomain: ordinateVisDomain } = ordinateConfig;
+
+  return {
+    xNormVisibleDomain: xVisibleDomain.map((v) =>
+      clamp(normalize(...abscissaVisDomain)(v), 0, 1)
```
loichuder, line 17: Suggestion: you could use a `ScaleLinear` here instead of normalizing "by hand"

apps/storybook/src/TiledHeatmap.stories.tsx
t20100: > Template without changing the props, right?
Apparently no because of the typing `api.xDomain` is not part of `TilesApi`

packages/lib/src/vis/tiles/hooks.ts
t20100: Yes, why not, I used visx's scaleLinear

apps/storybook/src/TiledHeatmap.stories.tsx
loichuder: > Apparently no because of the typing `api.xDomain` is not part of `TilesApi`

Oh yes true. Didn't realize that `xDomain` and `yDomain` were implemented by `MandelbrotTilesApi`

packages/lib/src/vis/tiles/hooks.ts
```diff
@@ -1,13 +1,32 @@
-import { useThree } from '@react-three/fiber';
+import type { Domain } from '@h5web/shared';
+import { scaleLinear } from '@visx/scale';
 
 import { useVisibleDomains } from '../hooks';
+import { useAxisSystemContext } from '../shared/AxisSystemContext';
 
-export function useDataPerPixel() {
+export function useScaledVisibleDomains(
```
t20100, line 7: Following our discussion, I still kept a hook since it is needed both in `TiledHeatmap` and `TiledLayer`

packages/lib/src/vis/tiles/hooks.ts
```diff
@@ -1,13 +1,32 @@
...
+    domain: ordinateConfig.visDomain,
+    range: yRange,
+    clamp: true,
+  });
 
   return {
-    xDataPerPixel: (xVisibleDomain[1] - xVisibleDomain[0]) / width,
-    yDataPerPixel: (yVisibleDomain[1] - yVisibleDomain[0]) / height,
+    xVisibleDomain: xVisibleDomain.map(xScale) as Domain,
+    yVisibleDomain: yVisibleDomain.map(yScale) as Domain,
```
t20100, line 30: Is using same naming for output as `useVisibleDomains` ok for you or best use a different naming (e.g., `xScaledVisibleDomain`... but that starts to be long)?

t20100:
I updated and use a single `useScaledVisibleDomains` hook (use from 2 components).
I also changed the parameters of `getTileOffsets` from rectangle corners to ranges to spare the conversion from domain to corners in `TiledLayer`: commit dbd1c11c12b6baa54ebb160ed19c852b76a1ad27. If you don't think it simplifies the reading, I'll remove it.

packages/lib/src/vis/tiles/hooks.ts
axelboc: I think it's fine.

axelboc (approved):

apps/storybook/src/TiledHeatmap.stories.tsx
```diff
@@ -116,23 +116,29 @@ class MandelbrotTilesApi extends TilesApi {
...
-        visDomain: [0, height],
-        isIndexAxis: true,
+        visDomain: ordinateDomain,
         showGrid: false,
       }}
-      visRatio={width / height}
+      visRatio={Math.abs(
+        (abscissaDomain[1] - abscissaDomain[0]) /
+          (ordinateDomain[1] - ordinateDomain[0])
+      )}
```
loichuder, line 141: The ratio you want to keep should not be `api.baseLayerSize.width / api.baseLayerSize.height` ?

In any case, I think the story would allow to catch more regressions if the span of the abscissa domain was different from the span of the ordinate domain (so far, both span 3).

packages/lib/src/vis/tiles/TiledLayer.tsx
```diff
@@ -16,28 +17,21 @@ interface Props extends ColorMapProps {
...
+  const { baseLayerIndex, numLayers } = api;
   const { width, height } = api.layerSizes[layer];
   const { tileSize } = api;
 
-  const { xVisibleDomain, yVisibleDomain } = useVisibleDomains();
   const { visSize } = useAxisSystemContext();
+  const { xVisibleDomain, yVisibleDomain } = useScaledVisibleDomains(
+    [0, width],
+    [0, height]
+  );
```
loichuder, line 28: Same here, you could pass `api.layerSizes[layer]` directly

packages/lib/src/vis/tiles/hooks.ts
loichuder: Agreed.

loichuder (approved):
This is way clearer in my opinion. Great job :+1:

apps/storybook/src/TiledHeatmap.stories.tsx
t20100: > The ratio you want to keep should not be api.baseLayerSize.width / api.baseLayerSize.height ?

I expect most of the time, but maybe not... But it behaves poorly if there is a large difference between the axes ratio and the ratio of the image since the level of details to then differs between the axes.

> I think the story would allow to catch more regressions if the span of the abscissa domain was different from the span of the ordinate domain

:+1:

PR #1036, opened 2022-03-29
Make Overlay and Annotation use new Html component

loichuder:


packages/lib/src/vis/hooks.ts
```diff
@@ -101,3 +102,28 @@ export function useCustomColors(
...
+  useFrameRendering();
+
+  return useCallback(
+    (point: Vector2 | Vector3) => {
+      const cameraPoint = new Vector3(point.x, point.y, 0).project(camera);
+      const htmlPoint = cameraPoint.clone().applyMatrix4(cameraToHtmlMatrix);
+      return new Vector2(htmlPoint.x, htmlPoint.y);
+    },
+    [camera, cameraToHtmlMatrix]
+  );
```
axelboc, line 128: I don't think there's a need for all this memoisation, since we use `useFrameRendering` anyway -- memoisation of functions like this is more to prevent triggering extra renders -- but here we already re-render on every frame.

packages/lib/src/vis/shared/Annotation.tsx
```diff
@@ -1,28 +1,39 @@
...
+import { useWorldToHtml } from '../hooks';
 import { useAxisSystemContext } from './AxisSystemContext';
 import Overlay from './Overlay';
 
 interface Props extends HTMLAttributes<HTMLDivElement> {
   x: number;
   y: number;
-  z?: number;
   scaleOnZoom?: boolean;
   style?: CSSProperties;
```
axelboc, line 13: Don't think the `style` prop is needed since it is already included by the base interface. You may want to spread it on the inner `div` styles, though (as per my other comment).

packages/lib/src/vis/shared/Overlay.tsx
```diff
@@ -1,153 +1,32 @@
...
-          innerRef.current = elem || undefined;
-        }}
         className={className}
         style={{
           position: 'absolute',
           top: 0,
           left: 0,
-          transformOrigin: 'left top',
-          transform: getInnerDivTransform(),
+          pointerEvents: 'none',
```
axelboc, line 22: Doesn't it need a `width`/`height` as well?

packages/lib/src/vis/shared/Annotation.tsx
```diff
@@ -1,28 +1,39 @@
...
+
+  const camera = useThree((state) => state.camera);
+
+  const { dataToWorld } = useAxisSystemContext();
+  const worldToHtml = useWorldToHtml();
+
+  const htmlPt = worldToHtml(dataToWorld(new Vector2(x, y)));
 
   return (
     <Overlay
```
axelboc, line 27: I think you could use `Html` directly, and just set `position: absolute` and `pointer-events: none` yourself (and move those styles to the inner `div`). This way, the overlay is really just to put a full-size `div` on top of the canvas (or another element).

packages/lib/src/vis/shared/TooltipMesh.tsx
```diff
@@ -22,7 +23,8 @@ function TooltipMesh(props: Props) {
   const { width, height } = useThree((state) => state.size);
 
   // Scales to compute data coordinates from unprojected mesh coordinates
-  const { worldToData, worldToHtml } = useAxisSystemContext();
+  const { worldToData } = useAxisSystemContext();
+  const worldToHtml = useWorldToHtml();
```
axelboc, line 27: Damn, this isn't great. We're now calling `useFrameRendering` in the tooltip... I hadn't thought of the cases where `worldToHtml` is invoked from inside a callback like an event handler. :disappointed:

axelboc (approved):
Nice simplification in the tooltip :clap:

PR #1037, opened 2022-03-30
Upgrade react-slider

axelboc:
Two [main changes](https://github.com/zillow/react-slider/releases/tag/v2.0.0) that concern us:

- `prop-types` is no longer a peer dependency, which means we won't get the warnings any more
- `react-slider` now re-renders itself when it is resized thanks to `ResizeObserver` (before, it used to just listen to the window's `resize` event), which means we no longer have to take care of this ourselves in `SlicingSlider` in order for the sliders to re-render when the H5Web notebook cell is resized in JLab.

### DIFF
--- a/packages/app/src/dimension-mapper/SlicingSlider.tsx
+++ b/packages/app/src/dimension-mapper/SlicingSlider.tsx
@@ -1,5 +1,5 @@
 import { useDebouncedCallback, useMeasure } from '@react-hookz/web';
-import { useEffect, useRef, useState } from 'react';
+import { useState } from 'react';
 import ReactSlider from 'react-slider';
 
 import styles from './SlicingSlider.module.css';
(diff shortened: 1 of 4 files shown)


packages/app/src/dimension-mapper/SlicingSlider.tsx
```diff
@@ -17,23 +17,16 @@ function SlicingSlider(props: Props) {
...
-
-  const [containerSize, containerRef] = useMeasure<HTMLDivElement>();
-  const sliderRef = useRef<ReactSlider>(null);
-
   const onDebouncedChange = useDebouncedCallback(onChange, [onChange], 250);
 
-  useEffect(() => {
-    sliderRef.current?.resize();
-  }, [containerSize]);
+  const [containerSize, containerRef] = useMeasure<HTMLDivElement>();
```
axelboc, line 22: We still need to measure the container to know whether to render marks on the slider.

loichuder (approved):

PR #1038, opened 2022-03-30
Upgrade @testing-library/user-event

axelboc:
[Major release](https://github.com/testing-library/user-event/releases/tag/v14.0.0) with a [major change](https://testing-library.com/docs/user-event/intro#writing-tests-with-userevent) in API:

```ts
// BEFORE
userEvent.click();

// AFTER
const user = userEvent.setup()
await user.click();
```

To make things easier, I've put `userEvent.setup()` in our `renderApp` function.

One thing to note is that `user.click()` now calls `setTimeout` internally between events (`hover`, `mousedown`, `pointerdown`, `click` etc.) which is [complex to deal with](https://github.com/testing-library/user-event/issues/833) in tests that user fake Jest timers. As a workaround until user-event provides [a better way](https://github.com/testing-library/user-event/issues/585) to deal with this, I've disabled this new `setTimeout` behaviour with `userEvent.setup({ delay: null })` as recommended.

Note also that a couple of our utilities like `selectExplorerNode` need access to the `user` object now, so instead of passing it as parameter, which would have been a bit tedious, I decided to return the utilities from `renderApp` so they have direct access to the `user` object in their closures.

```ts
// BEFORE
await renderApp();
await selectExplorerNode('...');

// AFTER
const { selectExplorerNode } = await renderApp();
await selectExplorerNode('...');
```


packages/app/src/test-utils.tsx
```diff
@@ -1,28 +1,47 @@
...
+export async function renderApp(): Promise<RenderAppResult> {
+  const user = userEvent.setup({ delay: null }); // https://github.com/testing-library/user-event/issues/833
+  const renderResult = render(
+    <MockProvider>
+      <App />
+    </MockProvider>
+  );
+
+  await screen.findByLabelText('Loading root metadata...');
+  await screen.findByRole('treeitem', { name: 'entities' });
```
axelboc, line 25: The weird async workaround we had in `renderApp`: `act(async () => { render(...) })`, was no longer working. I managed to fix the tests with the above, which is no longer a workaround, really -- we just wait for the initial metadata to finish loading, which makes sense.

loichuder (approved):

PR #1039, opened 2022-03-31
Fix tooltip not hiding during a drag

axelboc:


packages/lib/src/vis/shared/AxisSystem.module.css
```diff
@@ -4,10 +4,6 @@
     '. top-axis .'
     'left-axis grid right-axis'
     '. bottom-axis .';
-
-  /* - Prevent panning/zooming from axis when zoomed in.
-   * - Prevent text selection of tick labels when panning with touchpad. */
-  pointer-events: none;
```
axelboc, line 10: Already taken care of by `Overlay`.

packages/lib/src/vis/shared/TooltipMesh.tsx
```diff
@@ -49,11 +53,19 @@ function TooltipMesh(props: Props) {
...
+    if (tooltipOpen) {
+      hideTooltip();
+    }
+  }, [hideTooltip, tooltipOpen]);
+
+  // Hide tooltip when user starts panning
+  const onPointerDown = useCallback(() => hideTooltip(), [hideTooltip]);
 
-  // Show tooltip after dragging unless pointer has left canvas
+  // Show tooltip after dragging, if pointer is released inside the vis viewport
```
axelboc, line 68: `canvas` is the wrong term here. On a heatmap with a ratio, the tooltip remains hidden if the pointer is released outside of the vis (i.e. outside of the heatmap mesh). The term `vis` would also be incorrect as we want the tooltip to remain hidden when the vis is zoomed in and the pointer is released outside of the canvas (whereas the vis technically extends past the canvas). `vis viewport` seems more accurate - i.e. the area of the vis located inside the canvas (and therefore visible to the user).

loichuder: or `vis visible area`. `viewport` is fine though

loichuder (approved):
:fishing_pole_and_fish:

Issue #141, opened 2020-05-18
Scatterplot


axelboc:
If I'm not mistaken, this visualization will only be useful once we start supporting Nexus. I'm opening an epic nonetheless for reference and any preliminary discussion.

loichuder:
Could you put explicitly what do you mean by "scatterplot" ?

axelboc:
Something like this, where each point represents a country:

![image](https://user-images.githubusercontent.com/2936402/82324202-8d8cf680-99d9-11ea-9a92-6486d81fd7f4.png)

Each data point is positioned according to two values, in this case "life expectancy" and "GDP per capita". Once positioned, a point can be represented differently (size, shape, colour, etc.) based on additional values, in this case the colour of each point represents the country's continent.

In our case, this is different from a line plot in the sense that both axis are data axis, which cannot happen without Nexus support, I think.

loichuder:
Request from @vallsv

Could be point-based:
![image](https://user-images.githubusercontent.com/42204205/144569886-761df752-4060-4606-9876-a30aa4286168.png)

Or using a regular grid:
![image](https://user-images.githubusercontent.com/42204205/144570049-fca33b17-289c-4cdd-81f2-c5ef874c1a89.png)

vallsv:
The data format is
```
x: [...]
y: [...]
value: [...]
```

The solid rendering is used to be called hitmap i think (on silx it's histogram2D)

For solid, at one point we will probably also have to deal with meta like x/y-pixel-size, x/y-min, x/y-max to properly place the grid on the axis. See https://github.com/silx-kit/silx/blob/058b479031b450846c792aa91c63a161c9f8b778/src/silx/gui/plot/items/core.py#L1181 , but for now displaying anything would be very helpful.

loichuder:
First implementation in #913

loichuder:
After discussion with @t20100, we could have a NXScatter if a NXData group has
- a 1D `signal`
- two 1D `axes`
- `signal` and `axes` datasets of the same length

This would be a first basic support

axelboc:
I feel like this issue is now resolved:

- We have `ScatterVis`, which comes with a full-featured API, which is pretty much in line with other visualizations (axes scale types, `invertColorMap`, `interactions`, etc.)
- We also have `NxScatter`, which brings scatter visualizations to `@h5web/app`.

The solid/grid-based scatter plot is now covered by #1061. Feel free to re-open if needed.

PR #1040, opened 2022-03-31
Add a NxScatter vis for 1D data with axes of same length

loichuder:
image

Linked to #141


packages/app/src/vis-packs/core/scatter/config.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  showGrid: boolean;
+  toggleGrid: () => void;
+}
+
+function createStore() {
+  return create<ScatterConfig>(
+    persist(
+      // https://github.com/pmndrs/zustand/issues/701
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
loichuder, line 31: I tried updating zustand to 3.7.1 but it looks this is still an issue

packages/app/src/visualizer/utils.ts
```diff
@@ -88,16 +92,17 @@ function getSupportedCoreVis(
...
 ): VisDef[] {
-  if (!isGroup(entity) || !isNxDataGroup(entity, attrValueStore)) {
+  if (!isGroup(entity) || !isNxDataGroup(entity, attrValuesStore)) {
     return [];
   }
 
   assertGroupWithChildren(entity);
-  const dataset = findSignalDataset(entity, attrValueStore);
+  const dataset = findSignalDataset(entity, attrValuesStore);
+  const axisDatasets = findAssociatedDatasets(entity, 'axes', attrValuesStore);
```
loichuder, line 103: The particularity of NxScatter is that the two axes datasets must be defined. Therefore, I need to check for them here.

packages/lib/src/vis/scatter/ScatterVis.tsx
```diff
@@ -57,7 +57,7 @@ function ScatterVis(props: Props) {
   assertDefined(ordinateDomain, 'Ordinates have undefined domain');
 
   return (
-    <figure className={styles.root}>
+    <figure className={styles.root} aria-label={title} data-keep-canvas-colors>
```
loichuder, line 60: Fixing some overlooked attributes...

The `aria-label` is mainly to make tests work. The `data-keep-canvas-colors` fixes the fact that scatter vis were inverted in dark mode up to now.

cypress/specs/app.spec.ts
```diff
@@ -369,5 +369,21 @@ describe('App', () => {
         cy.matchImageSnapshot('nxrgb');
       }
     });
+
+    it('visualize dataset with 1D signal and 1D axes of same length as NxScatter', () => {
```
axelboc, line 373:
```suggestion
    it('visualize dataset with 1D signal and two 1D axes of same length as NxScatter', () => {
```

packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -47,6 +47,19 @@ test('visualize NXdata group with 2D signal', async () => {
   ).resolves.toBeVisible();
 });
 
+test('visualize NXdata group with 1D signal and 1D axes of same length', async () => {
```
axelboc, line 50:
```suggestion
test('visualize NXdata group with 1D signal and two 1D axes of same length', async () => {
```

packages/app/src/visualizer/utils.ts
axelboc: This doesn't trigger any requests, right? Because the attribute and the metadata of the children should already be fetched at this point.

That being said, it might still be worth moving this line inside the new `if` below, and splitting the `if` in two: a first check on the dimension of `dataset` and then on the axis datasets:

```ts
  if (hasNumDims(dataset, 1)) {
    const axisDatasets = findAssociatedDatasets(entity, 'axes', attrValuesStore);

    if (axisDatasets.length === 2 && axisDatasets.every(d => d && hasNumDims(d, 1)) {
      return [NEXUS_VIS[NexusVis.NxScatter]];
    }
  }
```

I would maybe also move this check below the interpretation checks so `findAssociatedDatasets` is not called for simple `spectrum` NXdata groups.

packages/app/src/vis-packs/nexus/containers/NxScatterContainer.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { isEqual } from 'lodash';
+
+import VisBoundary from '../../VisBoundary';
+import MappedScatterVis from '../../core/scatter/MappedScatterVis';
+import type { VisContainerProps } from '../../models';
+import NxValuesFetcher from '../NxValuesFetcher';
+import { assertNumericNxData } from '../guards';
+import { useNxData } from '../hooks';
+
+function NxSpectrumContainer(props: VisContainerProps) {
```
axelboc, line 15: Oops ^^

axelboc (approved):
Looks good.

Just a thought for a future refactoring: I feel like there's potential for making things a bit clearer by passing a more scatter-specific model than `AxisMapping` to `MappedScatterVis`. Part of this more specific model could then be used in `ScatterVis` to avoid having separate props for axis arrays and labels.

loichuder:
> Just a thought for a future refactoring: I feel like there's potential for making things a bit clearer by passing a more scatter-specific model than AxisMapping to MappedScatterVis. Part of this more specific model could then be used in ScatterVis to avoid having separate props for axis arrays and labels (similarly to AxisParams).

I gave this a spin and it truly improves things, good call ! I will do this in another PR though as I took advantage of this to bring support of axis scale types

Issue #970, opened 2022-02-16
[NX Spectrum] Legend and tooltip values for auxiliary signals


tomio13:
## Plotting multiple curves
based on the auxiliary signal in data file has no indication of which signal is which. Moving the cursor around also shows only data related to one curve, thus no help.

It would be great to have a **legend** which could be called up on the side indicating which color / curve type belongs to which signal.

axelboc:
Hi @tomio13, we now show the names and values of the auxiliary signals in the tooltip:

![image](https://user-images.githubusercontent.com/2936402/161081589-3508ce98-c420-455a-beb2-cad02b5da4d9.png)

Next, I was thinking of adding the colours of the auxiliary signals (and of the main plot) to the tooltip as well, which, in a way, makes the tooltip acts as a legend. So my question is: would that be sufficient, do you think, or would you still need a separate legend in the bottom right corner?

axelboc:
PR open to add plot colors to the tooltip:

![image](https://user-images.githubusercontent.com/2936402/161242604-42aa6133-7a6d-4888-b429-40f11aea1e80.png)

I'll wait to hear back from you and your team, @tomio13, before implementing a separate legend, as I feel like this version of the tooltip solves the initial issue of identifying which plot is which.

tomio13:
@axelboc : Thank you, this looks quite great.
I think, this is excellent for live view and handling of the data.
A legend is necessary then whenever the user tries to export / print the plot.

axelboc:
Great! I'll close this issue then. I've opened #1050 to deal with exporting visualisations as images, as this is not something that is currently implemented at all.

tomio13:
Thanks!

Issue #1029, opened 2022-03-24
[NX Spectrum] Errors on auxiliary signals


woutdenolf:
Auxiliary signals can have errors associated with them. Currently only the errors in the `signal` are shown.

In the example below, `area` is the signal and `area_errors` are the associated errors which are shown (blue). But you also have for example the auxiliary signal `FWHM_left` with errors `FWHM_left_errors`.

![image](https://user-images.githubusercontent.com/7264703/159944961-1868cc2c-73e7-49dd-b6b4-37f2e4f16d0d.png)

PR #1041, opened 2022-03-31
Show auxiliary values in tooltip

axelboc:
First part of #970 

In order to display the auxiliaries in the tooltip and the legend, we need to know their names. So I change the API of `LineVis` as follows: the `auxArray: NdArray<NumArray>[]` prop becomes `auxiliaries: Auxiliary[]`, where `Auxiliary` is an object with a `name`, `label` and `array`.

Since this was quite the refactoring already, I decided to implement the legend in a separate PR. So for now, I use the `name` in the tooltip. I plan to use the `label` (either `long_name`, or dataset name + `units`, or just dataset name) in the legend, since I'll have more space to work with.

image

### DIFF
--- a/packages/app/src/vis-packs/core/line/MappedLineVis.tsx
+++ b/packages/app/src/vis-packs/core/line/MappedLineVis.tsx
@@ -10,6 +10,7 @@ import { createPortal } from 'react-dom';
 import shallow from 'zustand/shallow';
 
 import type { DimensionMapping } from '../../../dimension-mapper/models';
+import type { Auxiliary } from '../../nexus/models';
 import {
   useMappedArrays,
   useMappedArray,
@@ -28,7 +29,7 @@ interface Props {
   valueLabel?: string;
   valueScaleType?: ScaleType;
   errors?: NumArray;
-  auxiliaries?: NumArray[];
+  auxiliaries?: Auxiliary[];
   dims: number[];
   dimMapping: DimensionMapping;
   axisMapping?: AxisMapping;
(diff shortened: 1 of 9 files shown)


packages/app/src/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -28,7 +29,7 @@ interface Props {
   valueLabel?: string;
   valueScaleType?: ScaleType;
   errors?: NumArray;
-  auxiliaries?: NumArray[];
+  auxiliaries?: Auxiliary[];
```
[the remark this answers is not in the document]
loichuder: This and #1029 makes me wonder if it would not be time to review the API of `LineVis` to have auxiliary signals first-class data arrays with the same treatment.

But I think it will be easier once #1029 is dealt with. So for now, let's keep it even though it is not ideal. Perhaps rename one `AuxParams` as it is close to `AxisParams` ?

packages/app/src/vis-packs/nexus/hooks.ts
```diff
@@ -56,3 +60,18 @@ export function useAxisMapping(
...
+  selection: string | undefined
+): Auxiliary[] {
+  const { attrValuesStore } = useContext(ProviderContext);
+
+  const auxValues = useDatasetValues(auxDatasets, selection);
+
+  return auxDatasets.map((dataset) => ({
+    name: dataset.name,
+    label: getDatasetLabel(dataset, attrValuesStore),
+    value: auxValues[dataset.name],
```
loichuder, line 75: I would remove `name` and put it as a fallback of domain. I don't think it is worth tossing `name` around.

Sure, `label` can be long but we already include the labels of axes in the tooltip so why not those of signals ?

packages/app/src/vis-packs/nexus/models.ts
```diff
@@ -50,3 +50,9 @@ export interface SilxStyle {
   signalScaleType?: ScaleType;
   axisScaleTypes?: ScaleType[];
 }
+
+export interface Auxiliary {
+  name: string;
+  label: string;
+  value: NumArray;
+}
```
loichuder, line 58: Same comment here about `name` an `label`

loichuder (approved):

PR #1047, opened 2022-04-01
Bring support for axis scale type in ScatterVis

loichuder:
With the refactoring suggested in https://github.com/silx-kit/h5web/pull/1040#pullrequestreview-927545065


apps/storybook/src/ScatterVis.stories.tsx
```diff
@@ -18,14 +18,16 @@ const abscissas = [
...
+  10.13, 10.04, 10.05, 10.14, 10.17, 10.2, 10.12, 10.06, 10.15, 10.19, 11,
+  11.08, 11.17, 11.09, 11.15, 11.01, 11.04, 11.09, 11.18, 11.15, 12.18, 12.17,
+  12.08, 12.07, 12.08, 12.06, 12.06, 12.04, 12.04, 12.06, 13.1, 13.06, 13.18,
+  13, 13.01, 13.18, 13.12, 13.06, 13.19, 13.15, 14.14, 14.18, 14.11, 14.02,
+  14.16, 14.05, 14.02, 14.01, 14.18, 14.11, 15.16, 15.05, 15.15, 15.16, 15.14,
+  15.09, 15.09, 15.15, 15.19, 15.11, 16.02, 16.04, 16.04, 16.12, 16.2, 16.17,
+  16.01, 16.05, 16.18, 16.12, 17.18, 17, 17.12, 17.1, 17.18, 17.01, 17.16,
+  17.17, 17.1, 17.02, 18.12, 18.1, 18.13, 18.19, 18.16, 18.03, 18.13, 18.17,
+  18.15, 18.18, 19.16, 19.12, 19.1, 19.17, 19.12, 19.05, 19.08, 19.17, 19.02,
+  19.13,
```
loichuder, line 30: I add `10` to all values once for all to avoid doing a `map` below

packages/app/src/vis-packs/core/scatter/MappedScatterVis.tsx
```diff
@@ -1,34 +1,42 @@
...
+    ordinateParams,
+    title,
+    toolbarContainer,
+    colorScaleType,
+  } = props;
 
-  const [xAxis, yAxis] = axisMapping;
-
-  assertDefined(xAxis?.value);
-  assertDefined(yAxis?.value);
```
loichuder, line 28: Assertions should now done in the HOC as enforced by the change in typings of this component

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -181,3 +182,9 @@ export function assertComplexSignal(
 ): asserts nxData is NxData<ComplexType> {
   assertComplexType(nxData.signalDataset);
 }
+
+export function assertScatterAxisParams(
+  params: AxisParams | undefined
+): asserts params is ScatterAxisParams {
+  assertDefined(params?.value);
+}
```
loichuder, line 190: This is not really a `nexus/utils` but it is used only in `NxScatterContainer` so I did not find a better place for it

packages/lib/src/vis/scatter/models.ts
```diff
@@ -0,0 +1,7 @@
+import type { NumArray, ScaleType } from '@h5web/shared';
+
+export interface ScatterAxisParams {
+  label?: string;
+  value: NumArray;
+  scaleType?: ScaleType;
+}
```
loichuder, line 7: This type is simply `AxisParams` with a required `value`. I could use `extend` but given the few fields, I thought it was clearer to rewrite it.

axelboc (approved):
Nicely done, this really makes the `ScatterVis` more consistent with the other vis.

PR #1049, opened 2022-04-01
Show auxiliary plot colors in line vis tooltip

### LINKED ISSUE #970: [NX Spectrum] Legend and tooltip values for auxiliary signals

axelboc:
Fix #970 

The tooltip remains unchanged when no auxiliaries are passed (i.e. the color of the main plot is not shown unless there are auxiliary plots).

image

The tooltip code of the line vis is starting to take quite a bit of space, but I wasn't sure if it was worth extracting the whole thing into a separate component, since we need access to so many props/variables from `LineVis`. Perhaps there's a way to abstract things but I'm not seeing it yet, especially since everything in there is so specific to the line vis at the moment.

### DIFF
--- a/packages/lib/src/vis/line/LineVis.module.css
+++ b/packages/lib/src/vis/line/LineVis.module.css
@@ -6,15 +6,33 @@
   margin: 1rem;
 }
 
-.tooltipValue {
+.tooltipItem {
+  display: flex;
+  align-items: baseline;
+}
+
+.tooltipValue,
+.tooltipAux {
+  composes: tooltipItem;
+}
+
+.tooltipValue,
+.tooltipValue + .tooltipAux {
   margin-top: 0.25rem;
 }
 
-.tooltipValue:not(:last-child) {
-  margin-bottom: 0.25rem;
-}
-
-.tooltipValue > strong {
+.tooltipValue strong {
   font-weight: 600;
   font-size: 1.125em;
+}
+
+.mark {
+  content: '';
+  display: block;
+  width: 0.75em;
+  height: 0.375rem;
+  margin-right: 0.375rem;
+  padding-bottom: 0.125rem;
+  background-color: currentColor;
+  transform: translateY(-1px);
 }
--- a/packages/lib/src/vis/line/LineVis.tsx
+++ b/packages/lib/src/vis/line/LineVis.tsx
@@ -152,15 +152,28 @@ function LineVis(props: Props) {
                 {`${abscissaLabel ?? 'x'} = ${formatTooltipVal(abscissa)}`}
 
                 <div className={styles.tooltipValue}>
-                  <strong>{formatTooltipVal(value)}</strong>
-                  {error && ` ±${formatTooltipErr(error)}`}
-                  {dtype && <em>{` (${formatNumType(dtype)})`}</em>}
+                  {auxiliaries.length > 0 && (
+                    <span
+                      className={styles.mark}
+                      style={{ color: curveColor }}
+                    />
+                  )}
+                  <span>
+                    <strong>{formatTooltipVal(value)}</strong>
+                    {error !== undefined && ` ±${formatTooltipErr(error)}`}
+                    {dtype && <em>{` (${formatNumType(dtype)})`}</em>}
+                  </span>
                 </div>
 
-                {auxiliaries.map(({ label, array }) => {
-                  const val = formatTooltipVal(array.get(xi));
-                  return <div key={label}>{`${label} = ${val}`}</div>;
-                })}
+                {auxiliaries.map(({ label, array }, index) => (
+                  <div className={styles.tooltipAux} key={label}>
+                    <span
+                      className={styles.mark}
+                      style={{ color: auxColors[index % auxColors.length] }}
+                    />
+                    {label} = {formatTooltipVal(array.get(xi))}
+                  </div>
+                ))}
               </>
             );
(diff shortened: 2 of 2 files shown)


packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -152,14 +153,33 @@ function LineVis(props: Props) {
...
+                    {error && ` ±${formatTooltipErr(error)}`}
+                    {dtype && <em>{` (${formatNumType(dtype)})`}</em>}
+                  </span>
                 </div>
 
-                {auxiliaries.map(({ label, array }) => {
+                {auxiliaries.map(({ label, array }, index) => {
                   const val = formatTooltipVal(array.get(xi));
-                  return <div key={label}>{`${label} = ${val}`}</div>;
+                  const color = auxColors[index] || lastAuxColor;
```
loichuder, line 171: You could instead cycle though `auxColors` with `auxColors[index % auxColors.length]`

packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -152,14 +153,33 @@ function LineVis(props: Props) {
...
+
+                  return (
+                    <div
+                      className={styles.tooltipAux}
+                      key={label}
+                      data-color={color}
+                    >
+                      <span className={styles.mark} style={{ color }} />
+                      {label} = {val}
```
loichuder, line 180: It makes sense to me to format auxiliary values as the signal value
```suggestion
                      {label} = {formatTooltipVal(val)}
```

packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -152,14 +153,33 @@ function LineVis(props: Props) {
...
+                {auxiliaries.map(({ label, array }, index) => {
                   const val = formatTooltipVal(array.get(xi));
-                  return <div key={label}>{`${label} = ${val}`}</div>;
+                  const color = auxColors[index] || lastAuxColor;
+
+                  return (
+                    <div
+                      className={styles.tooltipAux}
+                      key={label}
+                      data-color={color}
```
loichuder, line 177: Why do you need this attribute ?

packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -152,14 +153,33 @@ function LineVis(props: Props) {
...
-                  {dtype && <em>{` (${formatNumType(dtype)})`}</em>}
+                  {auxiliaries.length > 0 && (
+                    <span
+                      className={styles.mark}
+                      style={{ color: curveColor }}
+                    />
+                  )}
+                  <span>
+                    <strong>{formatTooltipVal(value)}</strong>
+                    {error && ` ±${formatTooltipErr(error)}`}
```
loichuder, line 164: This displays `0` if `error` is 0 which is unfortunate.

It should display either `±0` or nothing at all.

packages/lib/src/vis/line/LineVis.module.css
```diff
@@ -6,15 +6,33 @@
...
+.tooltipValue,
+.tooltipAux {
+  composes: tooltipItem;
 }
 
-.tooltipValue > strong {
+.tooltipValue,
+.tooltipValue + .tooltipAux {
+  margin-top: 0.25rem;
+}
```
loichuder, line 22: I get the impression it displays fine without this `margin-top` but up to you.

loichuder (approved):
Great, a few nitpicks as always :smiley_cat: 

I wouldn't worry too much about abstracting things right now. The work on #1029 will bring new constraints and it will be easier to work on without unfit abstractions

packages/lib/src/vis/line/LineVis.module.css
axelboc: Hmm, I find it aerates things a bit more.

packages/lib/src/vis/line/LineVis.tsx
axelboc: Good catch!

packages/lib/src/vis/line/LineVis.tsx
axelboc: I don't, thanks. Left-over from another attempt of using `background-color: attr(data-color)` in CSS, but this doesn't currently work across browsers -- `attr()` only works with the `content` CSS property. https://devdocs.io/css/attr()

packages/lib/src/vis/line/LineVis.tsx
axelboc: This is already the case, but I've removed the extra `val` variable for simplicity.

packages/lib/src/vis/line/LineVis.tsx
axelboc: Smart 🧠

axelboc: I've applied the same logic when passing the `color` prop to the aux curves.

packages/lib/src/vis/line/LineVis.tsx
loichuder: Oh, indeed, sorry for the false alarm

PR #1051, opened 2022-04-04
Add onClick callback to ScatterVis

loichuder:
For daiquiri, I need to be able to trigger events when clicking on point of a `ScatterVis`. 

In my early tries, I noticed that when attaching `onClick` to `points`, the handler was only called when clicking on the very center of the point. I figured this was because the handler was unaware of the shader rendering of the point that gives a point of bigger size.

I was able to solve this by increasing [the threshold of the raycaster](https://threejs.org/docs/index.html?q=raycas#api/en/core/Raycaster.params) to the size of the point so that events are caught for the full point.

Note that there is a limitation: the threshold is given in world units but point size stay constant when zooming in the scene. This means that the hitbox does not match the point when zooming in.

### DIFF
--- a/packages/lib/src/vis/scatter/ScatterPoints.tsx
+++ b/packages/lib/src/vis/scatter/ScatterPoints.tsx
@@ -1,7 +1,8 @@
 import type { Domain, NumArray, ScaleType } from '@h5web/shared';
+import type { ThreeEvent } from '@react-three/fiber';
 import { useThree } from '@react-three/fiber';
 import { rgb } from 'd3-color';
-import { useLayoutEffect, useMemo, useState } from 'react';
+import { useEffect, useLayoutEffect, useMemo, useState } from 'react';
 import { BufferAttribute, BufferGeometry } from 'three';
 
 import type { ColorMap } from '../heatmap/models';
@@ -20,6 +21,7 @@ interface Props {
   colorMap: ColorMap;
   invertColorMap: boolean;
   size: number;
+  onClick?: (index: number | undefined, evt: ThreeEvent<MouseEvent>) => void;
 }
 
 function ScatterPoints(props: Props) {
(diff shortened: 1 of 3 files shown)


axelboc:
> Note that there is a limitation: the threshold is given in world units but point size stay constant when zooming in the scene. This means that the hitbox does not match the point when zooming in.

Can the threshold be adjusted on zoom in the scatter vis? Maybe you're already planning to do this later.

packages/lib/src/vis/scatter/ScatterPoints.tsx
```diff
@@ -32,8 +34,23 @@ function ScatterPoints(props: Props) {
...
+  const raycaster = useThree((state) => state.raycaster);
+  useEffect(() => {
+    const oldThreshold = raycaster.params.Points?.threshold || 1;
+    if (raycaster.params.Points) {
+      raycaster.params.Points.threshold = size / 2;
+    }
+
+    return () => {
+      if (raycaster.params.Points) {
+        raycaster.params.Points.threshold = oldThreshold;
```
axelboc, line 49: Ooh wouldn't have thought to restore the initial value, smart :clap:

packages/lib/src/vis/scatter/ScatterPoints.tsx
```diff
@@ -20,6 +21,7 @@ interface Props {
   colorMap: ColorMap;
   invertColorMap: boolean;
   size: number;
+  onClick?: (evt: ThreeEvent<MouseEvent>) => void;
```
axelboc, line 24: Might be worth having a `handleClick` handler in this component that calls `onClick(index, evt)`, so consumers can access the index of the point more easily.

axelboc: We could also imagine having a `handleClick` handler in `ScatterVis` to add the point's value/abscissa/ordinate. Not as crucial, though, as the consumer can easily access those with the `index`.

axelboc (approved):
Great find :man_student:

PR #1052, opened 2022-04-04
Fixes TiledHeatmap display with flipped axes

t20100:
This PR fixes display issues of tiles heatmap when an axis is flipped.
It adds a story with both axes flipped.
It supports flipped axes by handling the axes flip early with a `<group>` in `TiledLayer` instead of late in `HeatmapMesh` and by supporting reverted domains.

### DIFF
--- a/packages/lib/src/vis/heatmap/HeatmapMesh.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapMesh.tsx
@@ -6,14 +6,14 @@ import { memo, useMemo } from 'react';
 import type { TextureFilter } from 'three';
 import { DataTexture, RGBAFormat, UnsignedByteType } from 'three';
 
-import { useAxisSystemContext } from '../..';
-import type { Size, VisScaleType } from '../models';
+import type { VisScaleType } from '../models';
+import type { VisMeshProps } from '../shared/VisMesh';
 import VisMesh from '../shared/VisMesh';
 import { DEFAULT_DOMAIN, getUniforms, VERTEX_SHADER } from '../utils';
 import type { ColorMap, TextureSafeTypedArray } from './models';
 import { getDataTexture, getInterpolator, scaleDomain } from './utils';
 
-interface Props {
+interface Props extends VisMeshProps {
   values: NdArray<TextureSafeTypedArray | Uint16Array>; // uint16 values are treated as half floats
   domain: Domain;
   scaleType: VisScaleType;
@@ -22,7 +22,6 @@ interface Props {
   magFilter?: TextureFilter;
   alphaValues?: NdArray<TextureSafeTypedArray | Uint16Array>; // uint16 values are treated as half floats
   alphaDomain?: Domain;
-  size?: Size;
 }
 
 const CMAP_SIZE = 256;
@@ -48,10 +47,8 @@ function HeatmapMesh(props: Props) {
     magFilter,
     alphaValues,
     alphaDomain = DEFAULT_DOMAIN,
-    size,
+    ...visMeshProps
   } = props;
-
-  const { ordinateConfig } = useAxisSystemContext();
 
   const dataTexture = useMemo(
     () => getDataTexture(values, magFilter),
@@ -143,7 +140,7 @@ function HeatmapMesh(props: Props) {
   };
 
   return (
-    <VisMesh scale={[1, ordinateConfig.flip ? -1 : 1, 1]} size={size}>
+    <VisMesh {...visMeshProps}>
       <shaderMaterial args={[shader]} />
     </VisMesh>
   );
--- a/packages/lib/src/vis/tiles/TiledLayer.tsx
+++ b/packages/lib/src/vis/tiles/TiledLayer.tsx
@@ -20,10 +20,9 @@ function TiledLayer(props: Props) {
   const { baseLayerIndex, numLayers, tileSize } = api;
   const layerSize = api.layerSizes[layer];
 
-  const { visSize } = useAxisSystemContext();
+  const { abscissaConfig, ordinateConfig, visSize } = useAxisSystemContext();
   const { xVisibleDomain, yVisibleDomain } = useScaledVisibleDomains(layerSize);
 
-  // Transform visible domain to current level-of-detail array coordinates
   const tileOffsets = getTileOffsets(xVisibleDomain, yVisibleDomain, tileSize);
 
   // Sort tiles from closest to vis center to farthest away
@@ -31,27 +30,33 @@ function TiledLayer(props: Props) {
   sortTilesByDistanceTo(tileOffsets, tileSize, center);
 
   return (
-    // Tranforms to use level of details layer array coordinates
+    // Transforms to handle axes flip and use level of details layer array coordinates
     <group
-      position={[-visSize.width / 2, -visSize.height / 2, layer / numLayers]}
-      scale={[
-        visSize.width / layerSize.width,
-        visSize.height / layerSize.height,
-        1,
-      ]}
+      scale={[abscissaConfig.flip ? -1 : 1, ordinateConfig.flip ? -1 : 1, 1]}
     >
-      {tileOffsets.map((offset) => (
-        <Suspense key={`${offset.x},${offset.y}`} fallback={null}>
-          <Tile
-            api={api}
-            layer={layer}
-            x={offset.x}
-            y={offset.y}
-            {...colorMapProps}
-            magFilter={layer === baseLayerIndex ? NearestFilter : LinearFilter}
-          />
-        </Suspense>
-      ))}
+      <group
+        position={[-visSize.width / 2, -visSize.height / 2, layer / numLayers]}
+        scale={[
+          visSize.width / layerSize.width,
+          visSize.height / layerSize.height,
+          1,
+        ]}
+      >
+        {tileOffsets.map((offset) => (
+          <Suspense key={`${offset.x},${offset.y}`} fallback={null}>
+            <Tile
+              api={api}
+              layer={layer}
+              x={offset.x}
+              y={offset.y}
+              {...colorMapProps}
+              magFilter={
+                layer === baseLayerIndex ? NearestFilter : LinearFilter
+              }
+            />
+          </Suspense>
+        ))}
+      </group>
     </group>
   );
 }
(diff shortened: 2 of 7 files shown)


packages/lib/src/vis/heatmap/HeatmapMesh.tsx
```diff
@@ -143,7 +145,7 @@ function HeatmapMesh(props: Props) {
   };
 
   return (
-    <VisMesh scale={[1, ordinateConfig.flip ? -1 : 1, 1]} size={size}>
+    <VisMesh scale={scale ?? [1, ordinateConfig.flip ? -1 : 1, 1]} size={size}>
```
t20100, line 148: I did it this way to minimize the changes to `HeatmapMesh`, but I have the feeling that the axes flip would be best handle in the vis (no matter how), so that the HeatmapMesh` is as close as possible to a basic three.js component (i.e., not relying on the axes context).

packages/lib/src/vis/tiles/Tile.tsx
```diff
@@ -29,6 +29,7 @@ function Tile(props: Props) {
         {...colorMapProps}
         magFilter={magFilter}
         size={{ width, height }}
+        scale={new Vector3(1, 1, 1)}
```
t20100, line 32: Override default `HeatmapMesh` scale.

packages/lib/src/vis/tiles/TiledLayer.tsx
```diff
@@ -20,38 +20,43 @@ function TiledLayer(props: Props) {
...
-    // Tranforms to use level of details layer array coordinates
+    // Tranforms to handle axes flip and use level of details layer array coordinates
     <group
-      position={[-visSize.width / 2, -visSize.height / 2, layer / numLayers]}
-      scale={[
-        visSize.width / layerSize.width,
-        visSize.height / layerSize.height,
-        1,
-      ]}
+      scale={[abscissaConfig.flip ? -1 : 1, ordinateConfig.flip ? -1 : 1, 1]}
```
t20100, line 35: Handle axes flip here.

packages/lib/src/vis/tiles/hooks.ts
```diff
@@ -14,12 +14,17 @@ export function useScaledVisibleDomains(size: Size): {
   const { abscissaConfig, ordinateConfig } = useAxisSystemContext();
 
   const xScale = scaleLinear({
-    domain: abscissaConfig.visDomain,
+    domain: abscissaConfig.flip
+      ? abscissaConfig.visDomain.reverse()
+      : abscissaConfig.visDomain,
```
t20100, line 19: I hesitated to use `dataToWorld` (which handles the flip) and add a "worldToArray" scale then instead of taking care myself of the flip... but it doesn't sounds simpler.

packages/lib/src/vis/heatmap/HeatmapMesh.tsx
loichuder: Agreed. Given your changes, I think it is best to pass the `[1, ordinateConfig.flip ? -1 : 1, 1]` in `HeatmapVis` through the  `scale` prop.

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -10,15 +10,17 @@ export function getTileOffsets(
 ): Vector2[] {
   const { width, height } = tileSize;
 
-  const [xOrigin, xEnd] = xDomain;
-  const [yOrigin, yEnd] = yDomain;
+  const xMin = Math.min(...xDomain);
+  const xMax = Math.max(...xDomain);
```
loichuder, line 14: You could use this instead
```
const [xMin, xMax] = isDescending(xDomain) ? xDomain.reverse() : xDomain;
```

Up to you

packages/lib/src/vis/tiles/TiledLayer.tsx
loichuder: Is it necessary to have it separate from the `group` setting the `position` ?

packages/lib/src/vis/tiles/Tile.tsx
loichuder: That won't be needed if you implement https://github.com/silx-kit/h5web/pull/1052/files#r841521948

packages/lib/src/vis/tiles/TiledLayer.tsx
t20100: Not really, This can also be applied in `scale` and `position` in the other `group`

packages/lib/src/vis/tiles/hooks.ts
t20100: yes, it should be equivalent to `getCanvasScale(config, canvasSize) +/- canvasSize/2`

packages/lib/src/vis/utils.ts
```diff
@@ -177,19 +177,22 @@ export function getValueToIndexScale(
   );
 }
 
-export function getCanvasScale(
-  config: AxisConfig,
-  canvasSize: number
-): AxisScale {
+export function getDomainScale(config: AxisConfig, range: Domain): AxisScale {
```
t20100, line 180: Added function to share the code between the hook used in tiling and `getCanvasSize`.
Any better idea for the name?

packages/lib/src/vis/utils.ts
```diff
@@ -177,19 +177,22 @@ export function getValueToIndexScale(
...
   const { scaleType, visDomain, flip } = config;
-  const range = [-canvasSize / 2, canvasSize / 2];
 
   return createAxisScale(scaleType ?? ScaleType.Linear, {
     domain: visDomain,
-    range: flip ? range.reverse() : range,
+    range: flip ? [...range].reverse() : range,
```
t20100, line 185: Copy here since `reverse` is in-place and modifies the argument

packages/lib/src/vis/tiles/hooks.ts
```diff
@@ -13,19 +14,18 @@ export function useScaledVisibleDomains(size: Size): {
...
+  const yDomain: Domain = [0, height];
+
+  const xScale = getDomainScale(abscissaConfig, xDomain);
+  const yScale = getDomainScale(ordinateConfig, yDomain);
 
   return {
-    xVisibleDomain: xVisibleDomain.map(xScale) as Domain,
-    yVisibleDomain: yVisibleDomain.map(yScale) as Domain,
+    xVisibleDomain: xVisibleDomain.map((v) =>
+      clamp(xScale(v), ...xDomain)
```
t20100, line 25: `clamp` is no longer part of the scale

packages/lib/src/vis/tiles/hooks.ts
t20100: Here is a way to share code with `getCanvasScale`: https://github.com/silx-kit/h5web/pull/1052/commits/bc1d1a526d6ba0fe49cafeb11dc40192c3ce9a0c
What do you think?

loichuder: Looks good ! I would add `clamp` as an optional parameter to `getDomainScale` to solve https://github.com/silx-kit/h5web/pull/1052#discussion_r841662636

packages/lib/src/vis/tiles/TiledLayer.tsx
```diff
@@ -20,23 +20,29 @@ function TiledLayer(props: Props) {
...
   // Sort tiles from closest to vis center to farthest away
   const center = new Vector2(sum(xVisibleDomain) / 2, sum(yVisibleDomain) / 2);
   sortTilesByDistanceTo(tileOffsets, tileSize, center);
 
+  const abscissaFlip = abscissaConfig.flip ? -1 : 1;
+  const ordinateFlip = ordinateConfig.flip ? -1 : 1;
+
   return (
-    // Tranforms to use level of details layer array coordinates
+    // Tranforms to handle axes flip and use level of details layer array coordinates
```
loichuder, line 36:
```suggestion
    // Transforms to handle axes flip and use level of details layer array coordinates
```

packages/lib/src/vis/tiles/TiledLayer.tsx
```diff
@@ -20,23 +20,29 @@ function TiledLayer(props: Props) {
...
+      position={[
+        abscissaFlip * (-visSize.width / 2),
+        ordinateFlip * (-visSize.height / 2),
+        layer / numLayers,
+      ]}
       scale={[
-        visSize.width / layerSize.width,
-        visSize.height / layerSize.height,
+        (abscissaFlip * visSize.width) / layerSize.width,
+        (ordinateFlip * visSize.height) / layerSize.height,
```
loichuder, line 45: That's what I suspected: the flip must be applied on both the position and the scale.

In this case, I think it makes more sense to have a separated group that flips everything around as you did in the first place. Sorry !

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -10,15 +11,15 @@ export function getTileOffsets(
 ): Vector2[] {
   const { width, height } = tileSize;
 
-  const [xOrigin, xEnd] = xDomain;
-  const [yOrigin, yEnd] = yDomain;
+  const [xMin, xMax] = isDescending(xDomain) ? xDomain.reverse() : xDomain;
+  const [yMin, yMax] = isDescending(yDomain) ? yDomain.reverse() : yDomain;
```
loichuder, line 15: Should you make a copy here according to [your comment](https://github.com/silx-kit/h5web/pull/1052/files#r841662024) ?

packages/lib/src/vis/utils.ts
loichuder: `getAxisScaleFromConfig` ?

loichuder (approved):

packages/lib/src/vis/utils.ts
t20100: Go for it

packages/lib/src/vis/tiles/TiledLayer.tsx
t20100: :fishing_pole_and_fish:

packages/lib/src/vis/tiles/hooks.ts
t20100: I reworked this part because there was an issue with the flip (not needed here since it is handled at the top in a group).
It now calls `createAxisScale` directly, rather than sharing code with `getCanvasScale`

axelboc (approved):
Nice, so there's a fix for descending axis domains as well, right? Is it worth adding a story for that too?

t20100:
It works with descending axis values, I added a story for it.

PR #1056, opened 2022-04-05
Cancel selection in SelectionTool on Escape

### LINKED ISSUE #1055: When no modifier key is passed to `SelectionTool`, it should be possible to cancel the selection by pressing `Escape`

loichuder:
Fix #1055

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -125,6 +125,15 @@ function SelectionTool(props: Props) {
     [],
     { event: 'keydown' }
   );
+  useKeyboardEvent(
+    'Escape',
+    () => {
+      setStartPoint(undefined);
+      setEndPoint(undefined);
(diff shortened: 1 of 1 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -125,6 +125,17 @@ function SelectionTool(props: Props) {
     [],
     { event: 'keydown' }
   );
+  useKeyboardEvent(
+    'Escape',
+    () => {
+      if (!modifierKey) {
```
axelboc, line 131: Does this condition matter? Just wondering.

axelboc (approved):

packages/lib/src/interactions/SelectionTool.tsx
loichuder: Not really. We could have the disable by `Escape` even if `modifierKey` is defined.

PR #1058, opened 2022-04-05
Allow zoom selection rectangle to be centred around user click

### LINKED ISSUE #1057: Selection area to be zoomed on should be centred around user selection

loichuder:
Fix #1057

### DIFF
--- a/apps/storybook/src/SelectToZoom.stories.tsx
+++ b/apps/storybook/src/SelectToZoom.stories.tsx
@@ -1,39 +1,54 @@
 import {
-  DataCurve,
   Pan,
   VisCanvas,
   Zoom,
   SelectToZoom,
   ResetZoomButton,
+  HeatmapMesh,
+  getDomain,
 } from '@h5web/lib';
 import type { ModifierKey as ModifierKeyType } from '@h5web/lib';
+import { mockValues, ScaleType } from '@h5web/shared';
 import type { Meta, Story } from '@storybook/react';
-import { range } from 'lodash';
+import ndarray from 'ndarray';
 
 import FillHeight from './decorators/FillHeight';
 
-const X = range(-5, 5, 0.01);
-const Y = X.map((x) => Math.exp((-x * x) / 0.2));
+const cols = 41;
+const rows = 20;
+const values = ndarray(Float32Array.from(mockValues.twoD.flat()), [rows, cols]);
+const domain = getDomain(values);
 
-function GaussianCurve() {
-  return <DataCurve abscissas={X} ordinates={Y} color="blue" />;
+function Image() {
+  return (
+    <HeatmapMesh
+      values={values}
+      domain={domain || [0, 1]}
+      colorMap="Viridis"
+      scaleType={ScaleType.Linear}
+    />
+  );
 }
 
 interface TemplateProps {
   modifierKey: ModifierKeyType;
+  keepRatio?: boolean;
+  clampCenter?: boolean;
 }
 
 const Template: Story<TemplateProps> = (args) => {
+  const { keepRatio } = args;
   return (
     <VisCanvas
       abscissaConfig={{ visDomain: [-5, 5], showGrid: true }}
       ordinateConfig={{ visDomain: [-0.5, 1.5], showGrid: true }}
+      visRatio={keepRatio ? cols / rows : undefined}
     >
       <Pan />
       <Zoom />
       <SelectToZoom {...args} />
       <ResetZoomButton />
-      <GaussianCurve />
+      <Image />
     </VisCanvas>
   );
 };
@@ -42,11 +57,24 @@ export const Default = Template.bind({});
 Default.args = {
   modifierKey: 'Control',
 };
-Default.argTypes = {
-  modifierKey: {
-    control: { type: 'inline-radio' },
-    options: ['Alt', 'Control', 'Shift'],
-  },
+
+export const KeepRatio = Template.bind({});
+KeepRatio.args = {
+  modifierKey: 'Control',
+  keepRatio: true,
+};
+
+export const ClampCenter = Template.bind({});
+ClampCenter.args = {
+  ...Default.args,
+  clampCenter: true,
+};
+
+export const KeepRatioAndClampCenter = Template.bind({});
+KeepRatioAndClampCenter.args = {
+  ...Default.args,
+  keepRatio: true,
+  clampCenter: true,
 };
 
 export default {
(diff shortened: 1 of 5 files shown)


apps/storybook/src/SelectToZoom.stories.tsx
```diff
@@ -1,39 +1,54 @@
...
+function Image() {
+  return (
+    <HeatmapMesh
+      values={values}
+      domain={domain || [0, 1]}
+      colorMap="Viridis"
+      scaleType={ScaleType.Linear}
+    />
+  );
 }
```
loichuder, line 31: I now display an `Image` in the story rather than a curve to demonstrate the `keepRatio` behaviour

packages/lib/src/interactions/RatioSelectionRect.tsx
```diff
@@ -3,22 +3,28 @@ import type { Vector2 } from 'three';
...
 
   const { dataToWorld, worldToData, visSize } = useAxisSystemContext();
 
-  const ratioEndPoint = getRatioEndPoint(startPoint, endPoint, ratio);
+  const [ratioStartPoint, ratioEndPoint] = getRatioRespectingRectangle(
+    startPoint,
+    endPoint,
+    ratio,
+    clampCenter
+  );
```
loichuder, line 25: If the center stays at the same place, it means the `startPoint` must move. This means that the code I wrote by assuming `endPoint` is the only vector to move must be refactored.

In the case of the ratio, this is not too difficult: I change the function returning the modified `endPoint` by a function that can modify `startPoint` and `endPoint`. `startPoint` is only modified if `clampCenter` is set to `true`.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -27,16 +28,31 @@ function SelectionTool(props: Props) {
...
   const camera = useThree((state) => state.camera);
   const { worldToData } = useAxisSystemContext();
   const shouldInteract = useRegisterInteraction(id, { modifierKey, disabled });
 
-  const [startPoint, setStartPoint] = useState<Vector2>();
-  const [endPoint, setEndPoint] = useRafState<Vector2 | undefined>(undefined);
+  const [originPoint, setOriginPoint] = useState<Vector2>();
+  const [selection, setSelection] = useRafState<Selection | undefined>(
+    undefined
+  );
```
loichuder, line 41: As `startPoint` and `endPoint` are updated together, it is more handy to work with a single state. It also simplifies the syntax at several places below.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -27,16 +28,31 @@ function SelectionTool(props: Props) {
...
+    clampCenter,
   } = props;
 
   const camera = useThree((state) => state.camera);
   const { worldToData } = useAxisSystemContext();
   const shouldInteract = useRegisterInteraction(id, { modifierKey, disabled });
 
-  const [startPoint, setStartPoint] = useState<Vector2>();
-  const [endPoint, setEndPoint] = useRafState<Vector2 | undefined>(undefined);
+  const [originPoint, setOriginPoint] = useState<Vector2>();
```
loichuder, line 38: As the `originPoint` (that is the point where the user clicked) can be different from the `startPoint`, it must be stored separately. I also store it in world coordinates for reasons exposed below.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -46,69 +62,81 @@ function SelectionTool(props: Props) {
...
       }
 
       const { unprojectedPoint, sourceEvent } = evt;
-      const point = worldToData(boundPointToFOV(unprojectedPoint, camera));
-      setEndPoint(point);
+      const worldEndPoint = new Vector2(unprojectedPoint.x, unprojectedPoint.y);
+      const worldStartPoint = getWorldStartPoint(originPoint, worldEndPoint);
+      const selection = {
+        startPoint: worldToData(boundPointToFOV(worldStartPoint, camera)),
+        endPoint: worldToData(boundPointToFOV(worldEndPoint, camera)),
```
loichuder, line 89: The `boundPointToFOV` must now be applied to both `startPoint` and `endPoint`. As it works in world coordinates, it is easier to deduce the `startPoint` from `originPoint` (termed `worldStartPoint`) in the world space, which is why I store `originPoint` in world space.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -46,69 +62,81 @@ function SelectionTool(props: Props) {
...
       (target as Element).releasePointerCapture(pointerId);
 
-      setStartPoint(undefined);
-      setEndPoint(undefined);
+      const worldEndPoint = new Vector2(unprojectedPoint.x, unprojectedPoint.y);
+      const worldStartPoint = getWorldStartPoint(originPoint, worldEndPoint);
+      const selection = {
+        startPoint: worldToData(boundPointToFOV(worldStartPoint, camera)),
+        endPoint: worldToData(boundPointToFOV(worldEndPoint, camera)),
+      };
```
loichuder, line 123: This part is the same as the one in `onPointerMove`. It might be a way to reduce the complexity of the component (which starts to be quite bloated).

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -46,69 +62,81 @@ function SelectionTool(props: Props) {
...
+
+      setSelection(undefined);
```
loichuder, line 125: Quite handy to reset the selection in one line

packages/lib/src/interactions/utils.ts
```diff
@@ -31,17 +32,26 @@ export function getRatioEndPoint(
...
-      startPoint.y + heightSign * height
-    );
+  const shiftX = widthSign * (originalRatio < ratio ? height * ratio : width);
+  const shiftY = heightSign * (originalRatio < ratio ? height : width / ratio);
+
+  if (!clampCenter) {
+    return [
+      startPoint,
+      new Vector2(startPoint.x + shiftX, startPoint.y + shiftY),
+    ];
```
loichuder, line 42: This is equivalent to the previous algorithm: the shift is applied fully to the `endPoint` while the `startPoint` stays at the same position ...

packages/lib/src/interactions/utils.ts
```diff
@@ -31,17 +32,26 @@ export function getRatioEndPoint(
...
+  const centerPoint = endPoint
+    .clone()
+    .sub(startPoint)
+    .divideScalar(2)
+    .add(startPoint);
+
+  return [
+    new Vector2(centerPoint.x - shiftX / 2, centerPoint.y - shiftY / 2),
+    new Vector2(centerPoint.x + shiftX / 2, centerPoint.y + shiftY / 2),
+  ];
```
loichuder, line 54: .. but in the case where we want to keep the center at the same place, we distribute the shift between the `startPoint` and the `endPoint`

loichuder (commented):
We might want to review our naming `startPoint`/`endPoint` as the `startPoint` could no longer be the point where the selection was started

PR #1059, opened 2022-04-06
Center ratio-respecting rectangle around drawn rectangle

### LINKED ISSUE #1057: Selection area to be zoomed on should be centred around user selection

loichuder:
Fix #1057 

Peek 2022-04-06 10-47

### DIFF
--- a/apps/storybook/src/SelectToZoom.stories.tsx
+++ b/apps/storybook/src/SelectToZoom.stories.tsx
@@ -1,39 +1,53 @@
 import {
-  DataCurve,
   Pan,
   VisCanvas,
   Zoom,
   SelectToZoom,
   ResetZoomButton,
+  HeatmapMesh,
+  getDomain,
 } from '@h5web/lib';
 import type { ModifierKey as ModifierKeyType } from '@h5web/lib';
+import { mockValues, ScaleType } from '@h5web/shared';
 import type { Meta, Story } from '@storybook/react';
-import { range } from 'lodash';
+import ndarray from 'ndarray';
 
 import FillHeight from './decorators/FillHeight';
 
-const X = range(-5, 5, 0.01);
-const Y = X.map((x) => Math.exp((-x * x) / 0.2));
+const cols = 41;
+const rows = 20;
+const values = ndarray(Float32Array.from(mockValues.twoD.flat()), [rows, cols]);
+const domain = getDomain(values);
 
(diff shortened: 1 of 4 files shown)


apps/storybook/src/SelectToZoom.stories.tsx
```diff
@@ -1,39 +1,53 @@
...
-  return <DataCurve abscissas={X} ordinates={Y} color="blue" />;
+function Image() {
+  return (
+    <HeatmapMesh
+      values={values}
+      domain={domain || [0, 1]}
+      colorMap="Viridis"
+      scaleType={ScaleType.Linear}
+    />
+  );
```
loichuder, line 30: I now display an `Image` in the story rather than a curve to demonstrate the `keepRatio` behaviour

packages/lib/src/interactions/RatioSelectionRect.tsx
```diff
@@ -16,9 +16,13 @@ function RatioSelectionRect(props: Props) {
 
   const { dataToWorld, worldToData, visSize } = useAxisSystemContext();
 
-  const ratioEndPoint = getRatioEndPoint(startPoint, endPoint, ratio);
+  const [ratioStartPoint, ratioEndPoint] = getRatioRespectingRectangle(
+    startPoint,
+    endPoint,
+    ratio
+  );
```
loichuder, line 23: If the center stays at the same place, it means the `startPoint` must move. This means that the code I wrote by assuming `endPoint` is the only vector to move must be refactored.

Therefore, I change the function returning the modified `endPoint` by a function that can modify `startPoint` and `endPoint`.

packages/lib/src/interactions/utils.ts
```diff
@@ -31,17 +31,19 @@ export function getRatioEndPoint(
...
+  const centerPoint = endPoint
+    .clone()
+    .sub(startPoint)
+    .divideScalar(2)
+    .add(startPoint);
+
+  return [
+    new Vector2(centerPoint.x - shiftX / 2, centerPoint.y - shiftY / 2),
+    new Vector2(centerPoint.x + shiftX / 2, centerPoint.y + shiftY / 2),
+  ];
```
loichuder, line 46: To keep the center at the same place, we distribute the shift between the `startPoint` and the `endPoint`

axelboc (approved):

PR #1060, opened 2022-04-06
implement resolution of old-style signal on NXdata

bmaranville:
For older NeXus files, the plottable signal is identified by a ```@signal``` attribute on the dataset itself, rather than the NXdata group (see Version 1 and Version 2 at https://manual.nexusformat.org/datarules.html#find-plottable-data). 

This PR implements a fallback strategy for finding these plottable datasets by searching the children of the NXdata group to find a dataset with the ```signal``` attribute.

It does not implement the identification of axes using those older lookup algorithms, and that is still to-do.

---

Part of #1062

### DIFF
--- a/packages/app/src/__tests__/NexusPack.test.tsx
+++ b/packages/app/src/__tests__/NexusPack.test.tsx
@@ -88,6 +88,20 @@ test('visualize NXentry group with absolute path to 2D default signal', async ()
   ).resolves.toBeVisible();
 });
 
+test('visualize NXentry group with old-style signal', async () => {
+  const { selectExplorerNode } = await renderApp();
+  await selectExplorerNode('nexus_entry/nx_process/old-style_signal');
+
+  const tabs = await findVisSelectorTabs();
+  expect(tabs).toHaveLength(2);
+  expect(tabs[0]).toHaveTextContent(NexusVis.NxSpectrum);
+  expect(tabs[1]).toHaveTextContent(NexusVis.NxImage);
+
+  await expect(
+    screen.findByRole('figure', { name: 'twoD' }) // name of dataset with `signal` attribute
+  ).resolves.toBeVisible();
+});
+
 test('visualize NXroot group with 2D default signal', async () => {
   await renderApp();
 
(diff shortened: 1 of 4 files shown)


axelboc:
Hi @bmaranville, thanks for the PR.

Unfortunately, early on in the project, we decided not to support this use case and, more generally, old versions of the NeXus spec. If I recall, one the justifications was that supporting them would somewhat hide the fact that a file is using an old version of the spec, which may not incite people to update to the latest version. I know this is not always practical, but it is important to us to limit the complexity of the code as much as possible to make adding new features in the future easier.

If support for older NeXus versions is crucial for you, then we recommend dealing with it in your back-end infrastructure, perhaps by implementing a converter dedicated to upgrading legacy NeXus files.

If you've got other contributions planned (awesome!), it might be a good idea to open issues first, so we can all get on the same page with regards to user needs and implementation.

bmaranville:
Sure - I can understand wanting to encourage people to update to the latest version of the spec.

That said - we have an immutable (by policy) archive of 10 years of facility NeXus files in the older spec, and rewriting them is not an option.  We will have to think of something else if we want to support easy visualization of the signal in these files.

We absolutely need to be able to read files directly in the browser in order to use h5web at the facility ( as in #1031 ), but some of these other features are in the "nice to have" category and I'm happy to discuss them in an issue.

axelboc:
I was more talking about "patching" the files when they are requested. This would be relatively easy to do with a server based on H5Grove, for instance: whenever an `NXdata` group is requested, you check if it contains a dataset with a `signal` attribute, and if it does, you add a `signal` attribute to the response metadata object of the group. Granted, it is more involved than a static file server, but it would really give you a lot more flexibility.

bmaranville:
We could patch the files in the browser when they are read, and then the added complexity moves into a special version of the h5wasm library that we would maintain instead of having it in h5web, but this seems more awkward.  Running an H5Grove server is not an option at this time.

loichuder (approved):
Works for me.

Perhaps we could put the search for a signal dataset name in a function `findSignalDatasetName` that would return `undefined` if no signal is found. This function could be used both in `isNxDataGroup` and `findSignalDataset`.

Thoughts @axelboc ?

loichuder:
Oh, and we should add "old-style" NXdata to our mock data ! This can be done later.

axelboc (commented):
I took the liberty of rebasing and pushing a refactor. I've also added a couple of mock `NXdata` groups for testing purposes.

axelboc:
Sorry for the spam, just had to close/reopen to trigger the CI workflow (cf. c6938dab4ed58cac6a1e7d1ea7ed76a0d72f4d6d).

loichuder (approved):

PR #1064, opened 2022-04-11
Scale raycast threshold according to zoom in ScatterPoints

loichuder:
Implement https://github.com/silx-kit/h5web/pull/1051#issuecomment-1091648466


packages/lib/src/vis/scatter/ScatterPoints.tsx
```diff
@@ -88,6 +90,15 @@ function ScatterPoints(props: Props) {
...
   }, [color, dataGeometry, invalidate, position]);
 
+  const zoomVector = useThree((state) => state.camera.scale);
+
+  useFrame(() => {
+    const zoom =
+      Math.sqrt(zoomVector.x * zoomVector.x + zoomVector.y * zoomVector.y) /
+      Math.SQRT2;
+    setThreshold((size * zoom) / 2);
+  });
```
loichuder, line 100: There is maybe a way to combine the `useFrame` and the `useEffect` in a single hook ? What do you think ?

packages/lib/src/vis/scatter/ScatterPoints.tsx
```diff
@@ -88,6 +90,15 @@ function ScatterPoints(props: Props) {
     invalidate();
   }, [color, dataGeometry, invalidate, position]);
 
+  const zoomVector = useThree((state) => state.camera.scale);
+
+  useFrame(() => {
+    const zoom =
+      Math.sqrt(zoomVector.x * zoomVector.x + zoomVector.y * zoomVector.y) /
```
axelboc, line 97: Isn't this just `zoomVector.length()`? On the other hand, you could just take `zoomVector.x` as the threshold, since axial zoom is disabled (and doesn't make sense, since we can't have oval points and click targets).

axelboc (approved):

PR #1065, opened 2022-04-11
Move `worldToHtml` to new `CameraContext`

axelboc:
The end goal is to never have to use the `camera` directly in the components, so as to not worry about re-rendering on every R3F frame. For now, I introduce the new `CameraContext` and provide only `worldToHtml` through it. In my next PR, I will try to refactor all the places where we use `useThree` to retrieve the `camera` object. For instance, I'll provide getters for the `scale` and `position` that will return already-cloned, always-up-to-date vectors. Hopefully this will solve all our frame-related challenges and allow us to stop exposing `useFrameRendering`.

### DIFF
--- a/apps/storybook/src/Contexts.stories.mdx
+++ b/apps/storybook/src/Contexts.stories.mdx
@@ -1,0 +1,48 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Contexts" />
+
+## Contexts
+
+Children of `VisCanvas` can access two React contexts: `AxisSystemContext` and `CameraContext`.
+
+### AxisSystemContext
+
+`AxisSystemContext` includes helpful utilities to convert vectors from data space to world space and back.
+It exposes the size of the visualization, as well as the axis configs passed to `VisCanvas` for convenience.
+
+This context re-renders **whenever the size of the canvas changes**.
+
+```ts
+useAxisSystemContext(): AxisSystemParams
+
+const { visSize, dataToWorld, worldToData } = useAxisSystemContext();
+```
+
+| Name             | Description                                                                                                        | Type                                                  |
+| ---------------- | ------------------------------------------------------------------------------------------------------------------ | ----------------------------------------------------- |
+| `visSize`        | Visualization size (different from canvas size on ratio-constrained visualizations, like `HeatmapVis`)             | <code>Size</code>                                     |
+| `abscissaConfig` | Abscissa configuration object passed to `VisCanvas`                                                                | <code>AxisConfig</code>                               |
+| `ordinateConfig` | Ordinate configuration object passed to `VisCanvas`                                                                | <code>AxisConfig</code>                               |
+| `abscissaScale`  | Computes the X coordinate in world space of a given abscissa value (or the opposite with `abscissaScale.invert()`) | <code>AxisScale</code>                                |
+| `ordinateScale`  | Computes the Y coordinate in world space of a given ordinate value (or the opposite with `ordinateScale.invert()`) | <code>AxisScale</code>                                |
+| `dataToWorld`    | Converts a vector from data space to world space (calls `abscissaScale()` and `ordinateScale`)                     | <code>(vec: Vector2 &#124; Vector3) => Vector2</code> |
+| `worldToData`    | Converts a vector from world space to data space (calls `abscissaScale.invert()` and `ordinateScale.invert`)       | <code>(vec: Vector2 &#124; Vector3) => Vector2</code> |
+
+> This table is not exhaustive. Please consider any undocumented property as experimental or meant for internal use only.
+
+### CameraContext
+
+`CameraContext` provides utilities that make use of the position and scale (i.e. zoom level) of the camera.
+
+This context should be used sparingly and as far down as possible in the rendering tree, as it re-renders **every time the position and/or scale of the camera changes** (i.e. on every React Three Fiber frame).
+
+```ts
+useCameraContext(): CameraContextValue
+
+const { worldToHtml } = useCameraContext();
+```
+
+| Name          | Description                                      | Type                                                  |
+| ------------- | ------------------------------------------------ | ----------------------------------------------------- |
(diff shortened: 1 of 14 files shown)


apps/storybook/src/Contexts.stories.mdx
```diff
@@ -0,0 +1,52 @@
...
+`AxisSystemContext` includes helpful utilities to convert vectors from data space to world space and back.
+It exposes the size of the visualization, as well as the axis configs passed to `VisCanvas` for convenience.
+It also provides an API to implement custom interactions.
+
+This context re-renders **whenever the size of the canvas changes**.
+
+```ts
+useAxisSystemContext(): AxisSystemParams
+
+const { visSize, dataToWorld, worldToData } = useAxisSystemContext();
```
axelboc, line 20: I no longer list all the available properties. The point is to give an example of use. The full list is already in the table below.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -17,8 +17,6 @@ import FillHeight from './decorators/FillHeight';
 
 interface TemplateProps {
   selectionType: 'line' | 'rectangle';
-  disablePan?: boolean;
-  disableZoom?: boolean;
```
axelboc, line 21: I've chosen to always enable pan/zoom in the selection stories after all. It's annoying to have to enable them to test that it still works regardless of zoom level. I've also renamed the stories a bit.

packages/lib/src/vis/shared/AxisSystemProvider.tsx
```diff
@@ -37,8 +37,27 @@ function AxisSystemProvider(props: PropsWithChildren<Props>) {
...
+    [interactionMap]
+  );
+
+  const unregisterInteraction = useCallback(
+    (id: string) => {
+      interactionMap.delete(id);
+    },
+    [interactionMap]
+  );
+
```
axelboc, line 60: Just moving `shouldInteract` below the other two functions, to match the order in which the API is used.

packages/lib/src/vis/shared/CameraProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Vector2, Vector3 } from 'three';
+
+import { useFrameRendering } from '../hooks';
+import { useAxisSystemContext } from './AxisSystemContext';
+
+export interface CameraContextValue {
+  worldToHtml: (point: Vector2 | Vector3) => Vector2;
+}
+
+const CameraContext = createContext({} as CameraContextValue);
```
axelboc, line 13: I've put the context in the same file for simplicity. I'll do the same with `AxisSystemProvider` so that it's easier to get to the provider from a consumer component in VSCode.

apps/storybook/src/Contexts.stories.mdx
```diff
@@ -0,0 +1,54 @@
...
+| `cameraToHtmlMatrix`    | Transformation matrix from camera space to HTML space (cf. `CameraProvider` for an example of use)                                                  | <code>Matrix4</code>                                    |
+| `registerInteraction`   | Registers a custom interaction with a given ID and an optional `modifierKey` and `disabled` state                                                   | <code>(id: string, value: Interaction) => void</code>   |
+| `unregisterInteraction` | Unregisters the custom interaction with the given ID                                                                                                | <code>(id: string) => void</code>                       |
+| `shouldInteract`        | Checks if the interaction with the given ID is allowed to proceed - i.e. enabled, `modifierKey` pressed if any, no conflict with other interactions | <code>(id: string, event: MouseEvent) => boolean</code> |
+| `floatingToolbar`       | Floating toolbar container, positioned in the bottom-right corner of the canvas. Do not use directly; use `FloatingControl` instead.                | <code>HTMLDivElement &#124; undefined</code>            |
```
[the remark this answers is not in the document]
loichuder: Is it worth documenting `floatingToolbar` if it is not meant to be used ?

apps/storybook/src/HeatmapVis.stories.tsx
```diff
@@ -57,13 +57,13 @@ AxisValues.args = {
...
     value: Array.from(
-      { length: dataArray.shape[1] }, // works even when right edge of last pixel is not provided
+      { length: dataArray.shape[1] }, // far edge of last pixel must not be included
       (_, i) => 100 + 10 * i
     ),
   },
   ordinateParams: {
     value: Array.from(
-      { length: dataArray.shape[0] + 1 },
+      { length: dataArray.shape[0] }, // far edge of last pixel must not be included
```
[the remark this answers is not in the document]
loichuder: Odd. I will have a second look to check that this is the expected behaviour

apps/storybook/src/Selection.stories.tsx
loichuder: Doesn't it raise an issue when having no `modifierKey` for the selection ? Dragging when zoomed in could then trigger the selection and the pan.

Thanks for the renaming, that's way better !

packages/lib/src/vis/shared/CameraProvider.tsx
loichuder: THANK YOU

loichuder (approved):
Great stepping stone :+1:

apps/storybook/src/Selection.stories.tsx
axelboc: Arf, forgot about this, sorry.

apps/storybook/src/Contexts.stories.mdx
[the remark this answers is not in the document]
axelboc: I've reverted this and left all those properties undocumented. I've added a comment in `AxisSystemContext.ts` to clearly identify those properties as "internal". I've also added a note below the table in the Storybook explaining that the table is not exhaustive and that any property not listed should be considered internal or experimental.

axelboc:
In the end, I don't think this is the way to go. It just moves `useFrameRendering` higher up without resolving the underlying problem it poses. I'm going to close this and re-open a PR with some of the changes unrelated to `CameraContext`. I'll also open a PR with a revert of the `useWorldToHtml` refactoring so as to avoid having to document a breaking change in the release notes (especially since this would likely change again).

PR #1067, opened 2022-04-12
Restore `worldToHtml` for now to avoid breaking change

axelboc:
As mentioned in https://github.com/silx-kit/h5web/pull/1065#issuecomment-1096656798, the way we respond to camera changes needs more work. Instead of releasing a breaking change that will be changed again later, we go back to what we had before (while still fixing the missing `useFrameRendering` in `SelectionRect` and `SelectionLine`).

Note that I also removed calls to `camera.updateProjectionMatrix` after setting `camera.scale`, as the projection matrix stays the same in this case -- it changes only when the size of the canvas changes (and R3F already takes care of calling `updateProjectionMatrix` in this case).


packages/lib/src/vis/shared/Annotation.tsx
```diff
@@ -2,7 +2,7 @@ import { useThree } from '@react-three/fiber';
 import type { HTMLAttributes } from 'react';
 import { Vector2 } from 'three';
 
-import { useWorldToHtml } from '../hooks';
+import { useFrameRendering } from '../..';
```
loichuder, line 5:
```suggestion
import { useFrameRendering } from '../hooks';
```

loichuder (approved):

PR #1068, opened 2022-04-12
Misc docs improvements

axelboc:
Repurposed from #1065. See individual commits.

### DIFF
--- a/packages/lib/src/vis/shared/AxisSystemContext.tsx
+++ b/packages/lib/src/vis/shared/AxisSystemContext.tsx
@@ -12,15 +12,15 @@ export interface AxisSystemParams {
   ordinateScale: AxisScale;
   dataToWorld: (vec: Vector2 | Vector3) => Vector2;
   worldToData: (vec: Vector2 | Vector3) => Vector2;
+
+  // For internal use only
   floatingToolbar: HTMLDivElement | undefined;
-  shouldInteract: (id: string, event: MouseEvent) => boolean;
   registerInteraction: (id: string, value: Interaction) => void;
   unregisterInteraction: (id: string) => void;
+  shouldInteract: (id: string, event: MouseEvent) => boolean;
 }
 
-export const AxisSystemContext = createContext<AxisSystemParams>(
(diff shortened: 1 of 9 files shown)


packages/lib/src/vis/shared/AxisSystemContext.tsx
```diff
@@ -12,15 +12,15 @@ export interface AxisSystemParams {
   ordinateScale: AxisScale;
   dataToWorld: (vec: Vector2 | Vector3) => Vector2;
   worldToData: (vec: Vector2 | Vector3) => Vector2;
+
+  // For internal use only
```
axelboc, line 16: And this comment makes it clear to us that anything underneath is not documented and can be broken without worry :smile:

apps/storybook/src/HeatmapVis.stories.tsx
```diff
@@ -57,13 +57,13 @@ AxisValues.args = {
...
     value: Array.from(
-      { length: dataArray.shape[1] }, // works even when right edge of last pixel is not provided
+      { length: dataArray.shape[1] }, // far edge of last pixel must not be included
       (_, i) => 100 + 10 * i
     ),
   },
   ordinateParams: {
     value: Array.from(
-      { length: dataArray.shape[0] + 1 },
+      { length: dataArray.shape[0] }, // far edge of last pixel must not be included
```
loichuder, line 66: The problem lies elsewhere: asserting length of `abscissaValue` and `ordinateValue` removed the possibility to give pixel edges as values which is something we supported earlier.

We should leave the stories as unchanged and instead remove the [`assertLength`](https://github.com/silx-kit/h5web/blob/main/packages/lib/src/vis/heatmap/HeatmapVis.tsx#L68_) as [`getAxisValues`](https://github.com/silx-kit/h5web/blob/main/packages/lib/src/vis/heatmap/utils.ts#L124) already asserts the length of arrays.

loichuder (approved):

axelboc (commented):
Arf, hadn't submitted my review...

PR #1069, opened 2022-04-12
Clarify what is documented and undocumented in lib index

axelboc:
Also, rename `TiledHeatmap` to `TiledHeatmapMesh` for consistency with `HeatmapMesh`. (Note that we could also have renamed `HeatmapMesh` to `Heatmap` instead, but I think `Mesh` makes it clearer that it is a building block meant to be used inside `VisCanvas`.)

New story for `Overlay`:

image

New stories for `DataCurve`:

image

### DIFF
--- a/apps/storybook/src/DataCurve.stories.tsx
+++ b/apps/storybook/src/DataCurve.stories.tsx
@@ -69,4 +69,7 @@ export default {
     curveType: CurveType.LineOnly,
     visible: true,
   },
+  argTypes: {
+    color: { control: { type: 'color' } },
+  },
 } as Meta;
--- a/apps/storybook/src/DataCurve.stories.tsx
+++ b/apps/storybook/src/DataCurve.stories.tsx
@@ -1,0 +1,72 @@
+import type { DataCurveProps } from '@h5web/lib';
+import { CurveType, DataCurve, useDomain, VisCanvas } from '@h5web/lib';
+import { assertDefined, mockValues } from '@h5web/shared';
+import type { Meta, Story } from '@storybook/react/types-6-0';
+import { range } from 'lodash';
+
+import FillHeight from './decorators/FillHeight';
+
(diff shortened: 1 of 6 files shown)


apps/storybook/src/DataCurve.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const Glyphs = Template.bind({});
+Glyphs.args = {
+  abscissas: range(0, mockValues.oneD.length),
+  ordinates: mockValues.oneD,
+  color: 'black',
+  curveType: CurveType.GlyphsOnly,
+};
+Glyphs.argTypes = {
+  showErrors: { control: false },
+};
```
loichuder, line 46: It would perhaps make more sense to set this in an `argTypes` field in the `export` below and activate the control only for `WithErrors`.

apps/storybook/src/DataCurve.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+    layout: 'fullscreen',
+    controls: {
+      sort: 'requiredFirst',
+      exclude: ['abscissas', 'ordinates', 'errors'],
+    },
+  },
+  args: {
+    curveType: CurveType.LineOnly,
+    visible: true,
+  },
```
loichuder, line 71:
```suggestion
  },
  argTypes: {
    color: { control: { type: 'color' } }
 }
```

Not mandatory but does not cost much :money_with_wings:

loichuder (approved):
Thanks :slightly_smiling_face:

apps/storybook/src/DataCurve.stories.tsx
axelboc: I tried, but it didn't work for some reason 🤷

loichuder: Nvm, go ahead then :+1:

PR #1070, opened 2022-04-12
Add a tooltip to ScatterVis

loichuder:
Peek 2022-04-12 15-54


packages/lib/src/vis/scatter/ScatterVis.tsx
```diff
@@ -106,7 +137,25 @@ function ScatterVis(props: Props) {
...
+                )}, ${ordinateLabel ?? 'y'} = ${formatTooltipVal(
+                  ordinates[tooltipIndex]
+                )}`}
+              </span>
+              <div className={styles.tooltipValue}>
+                <strong>{formatTooltipVal(dataArray.get(tooltipIndex))}</strong>
+              </div>
+            </>
+          )}
+        </TooltipOverlay>
```
loichuder, line 158: `LineVis` and `HeatmapVis` have a `renderTooltip` prop to tune this default tooltip. Do you think we should add it to `ScatterVis` as well ?

If yes, what arguments should it take ? Coordinates of the mouse, data abscissa, data ordinate, data value and index ?

packages/lib/src/vis/shared/TooltipMesh.tsx
```diff
@@ -84,38 +82,14 @@ function TooltipMesh(props: Props) {
...
-        )}
-      </Overlay>
+      <TooltipOverlay
+        tooltipLeft={tooltipLeft}
+        tooltipTop={tooltipTop}
+        tooltipOpen={tooltipOpen}
+        guides={guides}
+      >
+        {content}
+      </TooltipOverlay>
```
loichuder, line 92: New component to reuse it in `ScatterVis`

packages/lib/src/vis/scatter/ScatterPoints.tsx
```diff
@@ -68,6 +78,22 @@ function ScatterPoints(props: Props) {
...
+    }
+  }
+
+  function handlePointerOut(evt: ThreeEvent<PointerEvent>) {
+    const { index } = evt;
+
+    if (onPointerOut) {
+      onPointerOut(index, evt);
+    }
+  }
```
axelboc, line 95: Maybe wrap those with `useCallback`?

packages/lib/src/vis/scatter/ScatterVis.tsx
```diff
@@ -74,6 +81,30 @@ function ScatterVis(props: Props) {
...
+  const {
+    tooltipOpen,
+    tooltipTop,
+    tooltipLeft,
+    tooltipData: tooltipIndex,
+    showTooltip,
+    hideTooltip,
+  } = useTooltip<number>();
+
+  function onPointerEnter(
```
axelboc, line 93: Might be worth inlining this once you remove the early return.

packages/lib/src/vis/scatter/ScatterVis.tsx
axelboc: Hmm neh, no point until we find a use case (which will then help determine what arguments are needed).

packages/lib/src/vis/shared/TooltipMesh.tsx
```diff
@@ -84,38 +82,14 @@ function TooltipMesh(props: Props) {
...
-                )}
-              </svg>
-            )}
-          </>
-        )}
-      </Overlay>
+      <TooltipOverlay
+        tooltipLeft={tooltipLeft}
+        tooltipTop={tooltipTop}
+        tooltipOpen={tooltipOpen}
```
axelboc, line 88: I'd move `tooltipOpen` first for consistency with the order of the props.

axelboc (approved):
Nice! We can export/document this new `TooltipOverlay` component in due time.

PR #1071, opened 2022-04-13
Remove outline for circle glyph of GlyphMaterial

loichuder:
Fixes an issue reported by @vallsv : 
image

The outline was particularly visible when points were overlapping.

**Before**:
image

**With this PR**:
image

Note that the shader was originally from silx so pinging @t20100 here: is this outline expected ?


axelboc (approved):

packages/lib/src/vis/line/GlyphMaterial.tsx
```diff
@@ -33,9 +33,12 @@ const GLYPH_SHADERS = {
   `,
   [GlyphType.Circle]: `
     float alphaSymbol(vec2 coord, float size) {
-      float radius = 0.5;
       float r = distance(coord, vec2(0.5, 0.5));
-      return clamp(size * (radius - r), 0.0, 1.0);
+      if( r < 0.5) {
+        return 1.0;
+      } else {
```
axelboc, line 39: You don't need the `else` technically.

loichuder: True, This is more to comply with the syntax of other shaders from silx.

PR #1074, opened 2022-04-14
Center annotation automatically when using `scaleOnZoom`

axelboc:
With `scaleOnZoom`, if the annotation is not centered around its coordinates, it shifts once the user starts zooming. The annotation in the story is currently positioned at `(30, 18)`:

image

But loses its position when zooming:

image

With this PR, I propose that we always center annotations when `scaleOnZoom` is `true`, with `translate(-50%, -50%)`.

I take the opportunity to improve the annotation stories a little, notably by showing how one might go about drawing an SVG rectangle:

image

image

### DIFF
--- a/apps/storybook/src/HeatmapVisDisplay.stories.tsx
+++ b/apps/storybook/src/HeatmapVisDisplay.stories.tsx
@@ -124,17 +124,57 @@ WithAnnotation.args = {
 export const WithAnnotationZoom: Story<HeatmapVisProps> = (args) => (
   <HeatmapVis {...args}>
     <Annotation
-      x={30}
-      y={18}
-      style={{
-        width: '100px',
-        color: 'white',
-        textAlign: 'center',
-        fontSize: '0.75rem',
-      }}
+      x={10}
+      y={15}
       scaleOnZoom
+      style={{
+        width: '230px',
+        color: 'white',
+        fontSize: '0.875rem',
+        textAlign: 'center',
+      }}
     >
-      An annotation scaling with zoom
+      HTML annotation at (10, 15) that scales with zoom. Note that it is now
+      centred automatically.
+    </Annotation>
+    <Annotation
+      x={25}
+      y={10}
+      scaleOnZoom
+      style={{
+        display: 'flex',
+        alignItems: 'center',
+        width: 320,
+        height: 50,
+        color: 'white',
+        fontSize: '0.875rem',
+        textAlign: 'center',
+      }}
+    >
+      <>
+        <p style={{ flex: '1 1 0%', margin: 0, padding: '0.5rem' }}>
+          Annotations don't have to contain just text. You can draw shapes with
+          SVG or CSS, for instance.
+        </p>
+        <svg
+          style={{
+            position: 'absolute',
+            top: 0,
+            left: 0,
+            width: '100%',
+            height: '100%',
+            overflow: 'visible',
+          }}
+        >
+          <rect
+            width="100%"
+            height="100%"
+            fill="none"
+            stroke="currentColor"
+            strokeWidth={5}
+          />
+        </svg>
+      </>
     </Annotation>
   </HeatmapVis>
 );
--- a/apps/storybook/src/HeatmapVisDisplay.stories.tsx
+++ b/apps/storybook/src/HeatmapVisDisplay.stories.tsx
@@ -98,8 +98,20 @@ WithAxisLabels.args = {
 
 export const WithAnnotation: Story<HeatmapVisProps> = (args) => (
   <HeatmapVis {...args}>
-    <Annotation x={30} y={18} style={{ width: '200px', color: 'white' }}>
-      An annotation
+    <Annotation x={10} y={15} style={{ color: 'white' }}>
+      HTML annotation positioned at (10, 15)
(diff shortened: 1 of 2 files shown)


packages/lib/src/vis/shared/Annotation.tsx
```diff
@@ -16,11 +16,16 @@ function Annotation(props: Props) {
   const { x, y, scaleOnZoom, children, style, ...divProps } = props;
 
   const camera = useThree((state) => state.camera);
+  useFrameRendering();
 
   const { dataToWorld, worldToHtml } = useAxisSystemContext();
   const htmlPt = worldToHtml(dataToWorld(new Vector2(x, y)));
 
-  useFrameRendering();
+  if (scaleOnZoom && style?.transform) {
```
axelboc, line 24: We only want to forbid passing a `transform` if `scaleOnZoom` is `true`. It was a bit complicated to do this with types, so I went with a runtime assertion.

loichuder:
I see which problem you try to solve here but it seems a bit odd to me to change the centering when acting on a unrelated prop `scaleOnZoom`.

Could we instead fix the positioning so that the top-left stays at the designated position whatever the position of `scaleOnZoom` ?

axelboc:
Alright, I had another crack at it, and I remembered CSS property [`transform-origin`](https://devdocs.io/css/transform-origin). It defaults to `50% 50% 0` (i.e. `center`), which explains why the top left corner of annotations with a `scale` transform would not stay in the right place on zoom.

I've now added a prop called `center` to explicitly centre the annotation (with `transform: translate(-50%, -50%)`). When `scaleOnZoom` is `true` and `center` is `false`, I set `transform-origin` to `top left` to fix the positioning on zoom in this specific case.

axelboc:
image

image

apps/storybook/src/HeatmapVisDisplay.stories.tsx
```diff
@@ -98,8 +98,65 @@ WithAxisLabels.args = {
...
+        textAlign: 'center',
+      }}
+    >
+      <>
+        <p
+          style={{
+            flex: '1 1 0%',
+            margin: 0,
+            padding: '0.5rem',
+            border: '10px solid pink',
```
loichuder, line 135: :nauseated_face:

packages/lib/src/vis/shared/Annotation.tsx
```diff
@@ -10,18 +11,30 @@ interface Props extends HTMLAttributes<HTMLDivElement> {
...
 
   const camera = useThree((state) => state.camera);
+  useFrameRendering();
 
   const { dataToWorld, worldToHtml } = useAxisSystemContext();
   const htmlPt = worldToHtml(dataToWorld(new Vector2(x, y)));
 
-  useFrameRendering();
-
+  if ((center || scaleOnZoom) && style?.transform) {
```
loichuder, line 26: Perhaps add `transformOrigin` to the check ?

packages/lib/src/vis/shared/Annotation.tsx
```diff
@@ -10,18 +11,30 @@ interface Props extends HTMLAttributes<HTMLDivElement> {
...
+      'Annotation with `center` and/or `scaleOnZoom` cannot have its own `transform`'
+    );
+  }
+
+  const transforms = [
+    center ? 'translate(-50%, -50%)' : undefined,
+    scaleOnZoom
+      ? `scale(${1 / camera.scale.x}, ${1 / camera.scale.y})`
+      : undefined,
+  ];
```
loichuder, line 37: You could use template strings instead to avoid the `filter(isDefined).join` below.
  
```ts
  const transform = `${center ? 'translate(-50%, -50%)' : ''}
  ${scaleOnZoom ? ` scale(${1 / camera.scale.x}, ${1 / camera.scale.y})` : ''}`;
```

Feel free to reject this suggestion.

loichuder (approved):
Much closer to what I would expect. Nice work !

packages/lib/src/vis/shared/Annotation.tsx
axelboc: My feeling was that one would not use one without the other. :shrug:

axelboc: If someone uses `transform-origin` on its own, it'd probably be to override our own value on purpose, and I didn't want to prevent that.

packages/lib/src/vis/shared/Annotation.tsx
axelboc: My OCD has an issue with that leading space when `scaleOnZoom` is `true` and `center` is `false` (which could also be a trailing space with the opposite conditions). :rofl: I could add a `trim()`though :thinking:

axelboc: I kept an array but replaced `undefined` with empty strings and used `join(' ').trim()`.

apps/storybook/src/HeatmapVisDisplay.stories.tsx
axelboc: You're right, `pink` and `white` did not work together. I went with `pink` and `papayawhip` instead.

image

PR #1075, opened 2022-04-14
Stop panning temporarily if modifier key is released

### LINKED ISSUE #1014: [Pan] Panning should stop when modifier key is released
This behaviour is already implemented in `SelectionTool`: if a `modifierKey` is passed to `<Pan />`, the pan should stop when the key is released, and restarted (wherever the mouse is) when the key is pressed again.

axelboc:
Fix #1014

### DIFF
--- a/packages/lib/src/interactions/Pan.tsx
+++ b/packages/lib/src/interactions/Pan.tsx
@@ -4,19 +4,21 @@ import type { Vector3 } from 'three';
 
 import {
   useCanvasEvents,
+  useModifierKeyPressed,
   useMoveCameraTo,
   useRegisterInteraction,
 } from './hooks';
 import type { CanvasEvent, Interaction } from './models';
 
 function Pan(props: Interaction) {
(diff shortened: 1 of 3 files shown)


packages/lib/src/interactions/Pan.tsx
```diff
@@ -4,19 +4,21 @@ import type { Vector3 } from 'three';
...
+  const { modifierKey } = props;
 
   const camera = useThree((state) => state.camera);
+  const shouldInteract = useRegisterInteraction('Pan', props);
+  const moveCameraTo = useMoveCameraTo();
 
   const startOffsetPosition = useRef<Vector3>(); // `useRef` to avoid re-renders
-
-  const moveCameraTo = useMoveCameraTo();
+  const isModifierKeyPressed = useModifierKeyPressed(modifierKey);
```
axelboc, line 21: I abstracted the code from `SelectionTool` into a custom hook.

loichuder (approved):
The custom hook is a great improvement :+1:

Issue #1017, opened 2022-03-15
[XAxisZoom] Pressing/releasing Alt key toggles browser menu bar


axelboc:

axelboc:
I had a go at fixing this, but unfortunately, Firefox doesn't seem to allow listening to keyboard events on the `Alt` key at all. At least this is what I gather from my own tests, as well as https://stackoverflow.com/questions/41748450/firefox-preventdefault-of-alt-key and https://bugzilla.mozilla.org/show_bug.cgi?id=380637.

You can test it by configuring the `SelectToZoom` interaction with `Alt` in Storybook: when you release the `Alt` key while drawing a selection, the selection remains visible even though we have event handlers that are supposed to toggle it.

loichuder:
> You can test it by configuring the SelectToZoom interaction with Alt in Storybook: when you release the Alt key while drawing a selection, the selection remains visible even though we have event handlers that are supposed to toggle it.

Not sure this is linked to Firefox. There is something fishy when changing the modifier key on the fly in the Storybook: sometimes changing the modifier key though the controls make the interaction unusable or buggy. You have to refresh the page to fix the problem.

So, the toggling of the Selection on `Alt` works for me when landing directly on https://h5web-docs.panosc.eu/?path=/story/building-blocks-selection--with-modifier-key&args=selectionModifierKey:Alt

axelboc:
This is fixed with the proper merge of #1153 and the v6.0.1 release, I think?

PR #1077, opened 2022-04-14
Interactions provider

axelboc:
While looking at #1017, I ended up creating an `InteractionsProvider` to extract the three interaction callbacks out of `AxisSystemContext`. I also renamed `useRegisterInteraction` to `useInteraction`. Let me know what you think.


loichuder (approved):
Great initiative :+1:

PR #1078, opened 2022-04-19
Co-locate axis system context and provider to facilitate IDE navigation

axelboc:


loichuder (approved):
:heart: 

Perhaps it is the occasion to rename it ? :shrug:

axelboc:
Haha I did consider it, but I didn't want to introduce too many changes (since my branches are all based on one another), and thought we should discuss it together again ^^

PR #1079, opened 2022-04-20
Introduce new `useCameraState` hook to replace `useFrameRendering`

axelboc:
Draft for now to get your initial feedback. I'll update the documentation once we settle on a final solution.

---

Okay, so as a reminder, `useFrameRendering` is required when a component uses the camera's state (`position`, `scale`, `matrixWorldInverse` via `someVec.project(camera)`, etc.) during rendering. The problem is that the coupling between the camera-based code and the hook is implicit (i.e. it's easy to forget to use the hook or to remove it by mistake). Also, `useFrameRendering` does not allow for performance optimisations when a component doesn't care about every R3F frame (e.g. zoom level only).

To replace `useFrameRendering`, we need something similar (i.e. a combination of `useFrame` and `useState`), but where the code that runs on each frame and the state that is stored in `useState` are configurable. Most importantly, the state needs to contain the _result_ of the camera-based code, as this is what is relevant to decide whether the component should re-render or not. By doing so, the camera-based code can run on every frame while the state can be updated only if it actually changes (to avoid unnecessary re-renders).

To solve this, I've come up with:

- a **custom hook called `useCameraState`**, which accepts a function that runs camera-based code as argument;
- a **naming convention** for functions passed to `useCameraState`, to make them stand-out in the codebase and to try to convey that they should not be used directly during the render: I went with the prefix `grab` ("s'emparer de", "se saisir de", "attraper" ...), as in `grabVisibleDomains` for instance -- this allows me to also call these functions "grabbers" (similarly to "getters").

The idea is that a _grabber_ can only be used in three places:

- as an argument to `useCameraState` - e.g."
    ```ts
    const htmlPt = useCameraState(grabHtmlPoint, [dataPt]) // => grabHtmlPoint(camera, context, dataPt)
    ```
- inside another grabber, especially when dealing with camera state composition - e.g.
    ```ts
    const { htmlPt } = useCameraState((camera, context, dataPt) => ({ htmlPt: grabHtmlPoint(dataPt) }), [myDataPt])
    ```
- inside any event handler - e.g.
    ```ts
    const camera = useThree(state => state.camera);

    function handleClick(evt) {
      ...
      const htmlPt = grabHtmlPt(camera, context, new Vector2(x, y));
      ...
    }
    ```

Internally, `useCameraState` calls the grabber:

- on first render, to compute the initial state;
- on every R3F frame;
- in a `useEffect`, when the `context`, grabber function or grabber arguments change

The return value of the grabber is what gets stored in the state. Before updating the state, `useCameraState` always performs a deep comparison between the current state and the value returned by the grabber; if nothing changes, `setState` is not called to avoid unnecessary renders (and also to prevent infinite loops when the grabber function and/or grabber arguments are recreated on every render, like in the second example above).


packages/lib/src/vis/shared/Annotation.tsx
```diff
@@ -15,12 +14,13 @@ interface Props extends HTMLAttributes<HTMLDivElement> {
...
-  const htmlPt = worldToHtml(dataToWorld(new Vector2(x, y)));
-
-  useFrameRendering();
+  const { htmlPt, cameraScale } = useCameraState(
+    (camera, context, x, y) => ({
+      htmlPt: grabHtmlPoint(camera, context, new Vector2(x, y)),
+      cameraScale: grabCameraScale(camera),
+    }),
+    [x, y]
+  );
```
axelboc, line 23: In this case, the grabber is declared inline in order to compose the result of two grabbers into a single camera state. The reason for this is that we want to avoid multiple calls to `useCameraState` in the same component, as this would most likely lead to two consecutive renders on (potentially) every frame.

Because the grabber is recreated on every frame, the `useEffect` inside `useCameraState` ends up being called on every render, which basically means that the grabber is called twice in a row on every frame. The second time it's called, though, the result is the same, so the deep comparison check ensures that we don't re-render again (and therefore get stuck in a render loop).

To work around this unnecessary deep comparison, we could declare the grabber outside of the component and import it (like we do with other grabbers like `grabHtmlSelection`). This would work in the example above, but the resulting grabber would be used only by `Annotation`, so it feels a little unnecessary... Especially since the grabber is not computationally expensive.

If the extra deep comparison check is a problem and moving the grabber outside of the component is not possible (e.g. because it makes use of a variable that is declared inside the component), then it's always possible to wrap it in a `useCallback`:

```ts
const { x, y } = props;

const grabber = useCallback((camera, context) => ({
  htmlPt: grabHtmlPoint(camera, context, new Vector2(x, y)),
  cameraScale: grabCameraScale(camera),
}), [x, y]); // `x` and `y` are referenced directly instead of via the grabber arguments array of `useCameraState`

const { htmlPt, cameraScale } = useCameraState(grabber);
```

Now, you may be wondering why I chose to put the `grabber` function as dependency of the `useEffect` in `useCameraState`. Well, it's simply to avoid ending up with obscure bugs when misusing `useCameraState`. Typically, the code below works perfectly fine even though the grabber is declared inline and references variables declared in the component instead of passing them as grabber arguments:

```ts
const { x, y } = props;

const { htmlPt, cameraScale } = useCameraState(((camera, context) => ({
  htmlPt: grabHtmlPoint(camera, context, new Vector2(x, y)),
  cameraScale: grabCameraScale(camera),
}); // no grabber arguments array
```

If I remove the `grabber` function from the `useEffect`'s dependencies, then any changes to the `x` and `y` props will not be detected and the camera state will not be updated. We could try to prevent code like this from being written by configuring the `exhaustive-deps` linting rule to recognise `useCameraState` (https://www.npmjs.com/package/eslint-plugin-react-hooks), but this would only work in this project -- consumers of @h5web/lib would have to configure the linting rule in their own projects as well...

In summary:

- `useCameraState` is very robust and can be used in many ways; 
- some of these ways are not quite optimal, but their performance impacts are typically minimal;
- if performance is a worry (i.e. if the grabber is computationally expensive), an optimal solution is available.

packages/lib/src/vis/shared/AxisSystemProvider.tsx
```diff
@@ -60,14 +60,13 @@ function AxisSystemProvider(props: PropsWithChildren<Props>) {
       .setPosition(width / 2, height / 2); // account for shift of (0,0) position (center for camera, top-left for HTML)
   }, [availableSize]);
 
-  const camera = useThree((state) => state.camera);
-  const worldToHtml = useCallback(
```
axelboc, line 64: Converting from world space to camera space with `someVec3.project(camera)`, and back with `unproject`, must now be done inside grabbers. We could imagine providing two grabbers called `grabCameraPtFromWorld` and `grabWorldPointFromCamera` or something like that, but I don't see a use case for them at the moment. I feel like camera space is never a source/target space; only an in-between space. That's why I went with the more specific `grabHtmlPoint` instead, for data-to-HTML conversions (and we could later add `grabHtmlPointFromWorld` for world-to-HTML conversions).

packages/lib/src/vis/utils.ts
```diff
@@ -15,8 +15,10 @@ import { tickStep, range } from 'd3-array';
 import type { ScaleLinear, ScaleThreshold } from 'd3-scale';
```
axelboc, line 15: Eventually, we could move the grabbers into a file called `grabbers.ts`.

packages/lib/src/vis/utils.ts
```diff
@@ -316,17 +318,64 @@ export function getUniforms(
...
+    endPoint: grabHtmlPoint(camera, context, endPoint),
+  };
+}
+
+export function grabWorldFOV(camera: Camera): {
   topRight: Vector3;
   bottomLeft: Vector3;
 } {
-  // Unproject from normalized camera space (-1, -1) to (1, 1) to world space
+  // Unproject normalized camera viewport [(-1, -1), (1, 1)] to world space
```
t20100, line 357: In OpenGL the viewport is the defined in pixels (https://docs.gl/gl4/glViewport), the [-1, 1] space is called normalized device coordinates
```suggestion
  // Unproject normalized device coordinates [(-1, -1), (1, 1)] to world space
```

t20100:
It seems simple to use, at least from the changes in the code. I would need to use it to provide more feedbacks.

About the naming convention, the `grab` prefix sounds too generic to me but the good point is that it is short, alternatives I can think of  `grabCamera...` (but long), `getR3f...`, `r3f...`

packages/lib/src/vis/utils.ts
axelboc: Hmm, I was trying to explain a bit more clearly that we're computing the world coordinates of the area that the camera is pointing at. Unprojecting from one space to another does not explain why we unproject specifically `CAMERA_TOP_RIGHT` and `CAMERA_BOTTOM_LEFT`, but perhaps the name of the function is clear enough now and I can leave the initial comment as is? I see why the term viewport might be confusing, for sure.

axelboc:
`r3f` is good because it would allow us to use the prefix for functions that do things, like `r3fMoveCameraTo`. However, technically, it should be `threeMoveCameraTo`, `threeVisibleDomain`, etc. 😄

axelboc:
Perhaps I was too worried about the issue of having to configure ESLint to detect missing dependencies to `useCameraState`, and that forced me to make choices that made the API a bit too obscure. Here is the solution I had previously rejected but that may be a lot simpler in the end:

- `useCameraState` become like `useEffect` and the like: it receives an inline function (that doesn't have to be memoised) and an array of dependencies.
- The dependencies are not passed as arguments to the function by `useCameraState`; they are just passed as is to the hook's internal `useEffect`. This requires configure ESLint to detect missing dependencies to `useCameraState`, and explaining to consumers of the lib that they should do so as well once we expose the hook.
- We replace the deep comparison check before calling `setState` with a shallow comparison check. We can consider making the deep check an opt-in later by accepting a third argument: either an `options` object with `{ deep: true }` or a custom `comparator` function like in zustand.

```ts
const { xVisibleDomain, yVisibleDomain } = useCameraState(grabVisibleDomains, []);
```

```ts
const { htmlPt, cameraScale } = useCameraState(
  (camera, context) => ({
    htmlPt: grabHtmlPoint(camera, context, new Vector2(x, y)),
    cameraScale: grabCameraScale(camera),
  }),
  [x, y] // these are now classic effect dependencies, not grabber arguments
);
```

We can also forget about the `grab` naming convention for now, as it was attempting to solve a separate problem (i.e. to avoid calling camera-based functions during render).

Does this all sound like a better way to go?

loichuder (commented):
> Does this all sound like a better way to go?

It does :+1:

packages/lib/src/index.ts
```diff
@@ -67,15 +67,15 @@ export {
...
   extendDomain,
   getValueToIndexScale,
+  grabVisibleDomains,
 } from './vis/utils';
 
 export {
   useDomain,
   useDomains,
   useCombinedDomain,
-  useVisibleDomains,
```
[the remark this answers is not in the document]
axelboc: Totally, though by hiding `useCameraState` inside a custom hook, we risk that a consumer will use `useVisibleDomains` and `useCameraState` (or another custom hook with `useCameraState` inside) in the same component. This leads to potentially two re-renders on every frame, which could be avoided by composing the states in a single `useCameraState` call.

.eslintrc.js
```diff
@@ -6,7 +6,7 @@ const config = createConfig({
   cwd: __dirname,
   rules,
   overrides: [
-    ...overrides,
+    ...overrides.slice(1), // exclude shared React overrides (`react` is not installed at the root of the repo)
```
axelboc, line 9: Got an error because I turn on a rule from `eslint-plugin-react-hooks`, and this plugin is only enabled by Galex when it detects that `react` is installed.

eslint.shared.js
```diff
@@ -31,6 +31,13 @@ const overrides = [
     files: reactFiles,
     rules: {
       'react/jsx-no-constructed-context-values': 'off', // too strict
+
+      // `useCameraState` accepts an array of deps like `useEffect`
+      // https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks#advanced-configuration
+      'react-hooks/exhaustive-deps': [
+        'error',
+        { additionalHooks: '(useCameraState)' },
```
axelboc, line 39: this is a regex, so we can add more hooks with `|` later if needed

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -22,14 +23,15 @@ function SelectToZoom(props: Props) {
   const { width, height } = useThree((state) => state.size);
   const camera = useThree((state) => state.camera);
 
-  const { xVisibleDomain, yVisibleDomain } = useVisibleDomains();
+  const { xVisibleDomain, yVisibleDomain } = useCameraState(
+    getVisibleDomains,
+    []
```
axelboc, line 28: If I default the `deps` parameter to `[]` in `useCameraState`, then the linting rule `exhaustive-deps` no longer warns you if you don't pass a dependency array even though you need one.

axelboc (commented):
I've refactored the hook, configured ESLint and reverted the `grab` naming convention.

packages/lib/src/interactions/SelectToZoom.tsx
t20100: It's the same for `useCallback`, right? So it's inline with react.

loichuder (approved):

t20100 (approved):

PR #1080, opened 2022-04-22
Implement new `H5WasmProvider`

axelboc:
The provider code is ported from #1031 and https://github.com/bmaranville/h5web/pull/1/ but limited to the low-level `H5WasmProvider` first discussed in https://github.com/silx-kit/h5web/pull/1031#pullrequestreview-924227264. The provider just accepts an `ArrayBuffer` as prop and reads it on demand with `h5wasm`; it doesn't do any fetching and doesn't provide any UI for picking files.

To demonstrate the use of the provider, I implemented a demo with a file picker/drag-and-drop interface. It's available at http://localhost:3000/h5wasm if you want to test it.

Peek 2022-04-22 14-24

The focus for now is to bootstrap the low-level provider and deal with the BigInt issue. I didn't port the slicing code, didn't worry about broken and external links, and didn't fix some of the dtype inconsistencies with H5Grove that I noticed along the way. These things will come in subsequent PRs.

To work around the BigInt issue, I put the new provider in a separate package called `@h5web/h5wasm`. This way, people can still install `@h5web/app` in their apps without having to target browsers that support BigInt. I didn't bother creating a separate demo app, though, so the demo now only works in browsers that support BigInt (thanks to `build: { target: 'es2020' }`) in `demo/vite.config.js`. I don't think it's a big deal; what's important is that the JLab extension and Hibou still support older browsers. I've explained the browser support caveat in the new package's README.

### DIFF
--- a/packages/h5wasm/src/h5wasm-api.ts
+++ b/packages/h5wasm/src/h5wasm-api.ts
@@ -1,0 +1,140 @@
+import type { ValuesStoreParams } from '@h5web/app';
+import { convertDtype, ProviderApi } from '@h5web/app';
+import { getNameFromPath } from '@h5web/app/src/providers/utils';
+import type {
+  AttributeValues,
+  Dataset,
+  Entity,
+  Group,
+  GroupWithChildren,
+  UnresolvedEntity,
+} from '@h5web/shared';
+import {
+  buildEntityPath,
+  DTypeClass,
+  EntityKind,
+  hasScalarShape,
+} from '@h5web/shared';
+import { File as H5WasmFile, FS, ready } from 'h5wasm';
+
+import type { H5WasmAttribute, H5WasmEntity } from './models';
+import {
+  assertH5WasmDataset,
+  assertH5WasmEntityWithAttrs,
+  isH5WasmDataset,
+  isH5WasmGroup,
+  isHDF5,
+} from './utils';
+
+export class H5WasmApi extends ProviderApi {
+  private readonly file: Promise<H5WasmFile>;
+
+  public constructor(filename: string, buffer: ArrayBuffer) {
+    super(filename);
+
+    if (!isHDF5(buffer)) {
+      throw new Error('Expected valid HDF5 file');
+    }
+
+    this.file = this.initFile(buffer);
+  }
+
+  public async getEntity(path: string): Promise<Entity> {
+    const file = await this.file;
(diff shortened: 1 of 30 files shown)


packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    this.file = this.initFile(buffer);
+  }
+
+  public async getEntity(path: string): Promise<Entity> {
+    const file = await this.file;
+    const h5wEntity = file.get(path);
+
+    const segments = path.slice(1).split('/');
```
loichuder, line 45: You could use an util `getName`:

```ts
function getName(path) {
  const segments = path.slice(1).split('/');
  return segments[segments.length - 1]   
}
```

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      return {
+        ...baseEntity,
+        kind: EntityKind.Dataset,
+        attributes: this.processH5WasmAttrs(h5wEntity.attrs),
+        shape,
+        type:
+          typeof dtype === 'string'
+            ? convertDtype(dtype)
+            : { class: DTypeClass.Compound, fields: dtype.compound },
```
loichuder, line 127: h5wasm dtypes masy need specific conversions so we will have to revisit this later.

loichuder (approved):
The drag'n'drop demo works great (expect some dtype problems)

axelboc:
Forgot to submit my review comments again, sorry :disappointed: -- oh well, they were not critical.

packages/h5wasm/src/h5wasm-api.ts
bmaranville: These conversions were implemented in https://github.com/silx-kit/h5web/pull/1031/commits/f202aa8c604afd8cd2a029613e48429ea790f661 if you want to port them in another PR.

bmaranville:
I couldn't tell from your PR description whether you noticed that the slice functions, dtype mappings and broken/external links handling were all implemented in the original PR after it was cloned to this PR.  It should be straightforward to adapt them to this version.

axelboc:
Yes, I purposely decided not to port these features from your PR to limit the scope of this one. @loichuder is looking into bringing the features back and tracking the work in #1086.

PR #1081, opened 2022-04-25
Repair zoom if needed when "keep ratio" becomes enabled

### LINKED ISSUE #1076: Reset zoom when toggling on "keep aspect ratio"

axelboc:
Fix #1076

This issue is part of #993, but technically, the same problem can occur with the axial zoom components - i.e. you start zooming on one axis because `keepRatio` is `false` then toggle the flag to `true` and the visualisation is broken. Therefore, instead of putting the code in `SelectToZoom`, I decided to create a specialised `VisCanvas` component similarly to `ViewportCenterer`.

Also, instead of resetting the zoom to `(1, 1)` when `visRatio` becomes defined, I reset it to the scale of the axis that is most zoomed out - e.g. `(0.8, 0.6)` becomes `(0.8, 0.8)`. I also make sure that the scales of the two axes are actually different before resetting, to avoid triggering a frame unnecessarily.

Peek 2022-04-25 14-22

### DIFF
--- a/packages/lib/src/interactions/SelectToZoom.tsx
+++ b/packages/lib/src/interactions/SelectToZoom.tsx
@@ -1,6 +1,6 @@
 import { useThree } from '@react-three/fiber';
 
-import { useVisibleDomains } from '../vis/hooks';
+import { useVisibleDomains, useFrameRendering } from '../vis/hooks';
 import { useAxisSystemContext } from '../vis/shared/AxisSystemProvider';
 import RatioSelectionRect from './RatioSelectionRect';
 import SelectionRect from './SelectionRect';
@@ -28,7 +28,9 @@ function SelectToZoom(props: Props) {
       (yVisibleDomain[1] - yVisibleDomain[0])
   );
 
-  const onSelectionEnd = (selection: Selection) => {
+  useFrameRendering();
+
+  function onSelectionEnd(selection: Selection) {
     const { startPoint: dataStartPoint, endPoint: dataEndPoint } = selection;
 
     // Work in world coordinates as we need to act on the world camera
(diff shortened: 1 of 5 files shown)


axelboc:
Haha so, it's because we use `useVisibleDomains` in `SelectToZoom` without `useFrameRendering()`... :joy:

packages/lib/src/toolbar/floating/ResetZoomButton.tsx
```diff
@@ -21,9 +20,10 @@ function ResetZoomButton() {
   function resetZoom() {
     camera.scale.x = 1;
     camera.scale.y = 1;
+    camera.position.x = 0;
+    camera.position.y = 0;
     camera.updateMatrixWorld();
-
-    moveCameraTo(0, 0);
+    invalidate();
```
axelboc, line 26: I figure this avoids a few unnecessary calculations...

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -28,7 +28,9 @@ function SelectToZoom(props: Props) {
       (yVisibleDomain[1] - yVisibleDomain[0])
   );
 
-  const onSelectionEnd = (selection: Selection) => {
+  useFrameRendering();
```
axelboc, line 31: Now `dataRatio` remains up to date.

axelboc:
> Why did you decide against resetting the zoom to `(1, 1)` ? I thought it was fine.

Just seemed like a good idea to keep users as close to where they are as possible. Did you try the full reset? It's quite jarring.

loichuder (approved):
> Did you try the full reset? It's quite jarring.

Yes I tried and it seemed okay to me :shrug: 

Anyway, I am fine with your implementation. You can merge it whenever :octopus:

axelboc:
> Yes I tried and it seemed okay to me shrug

As in you preferred the full reset, or you don't care either way? I'm not all that attached to the partial reset, so if you prefer the full reset, I don't mind simplifying the code.

loichuder:
> you don't care either way?

I don't really mind but I don't think anyone does :grin: . Now that you have a working version with the partial reset, we may as well roll with it

PR #1082, opened 2022-04-26
Upgrade h5wasm and axios

axelboc:
- https://github.com/usnistgov/h5wasm/releases/tag/v0.3.2 - `group.get(path)` can return `null` if no entity is found at the given path
- https://github.com/axios/axios/releases/tag/v0.27.0 - new `AxiosError` class that can be used with `instanceof` to check for axios errors

### DIFF
--- a/packages/app/src/types.d.ts
+++ b/packages/app/src/types.d.ts
@@ -1,4 +1,3 @@
 /* eslint-disable spaced-comment */
 
 /// <reference types="vite/client" />
-/// <reference types="@h5web/shared/src/types" />
--- a/packages/h5wasm/src/h5wasm-api.ts
+++ b/packages/h5wasm/src/h5wasm-api.ts
@@ -10,12 +10,12 @@ import type {
   UnresolvedEntity,
 } from '@h5web/shared';
 import {
+  assertNonNull,
   buildEntityPath,
   DTypeClass,
   EntityKind,
-  hasScalarShape,
 } from '@h5web/shared';
-import { File as H5WasmFile, FS, ready } from 'h5wasm';
+import { File as H5WasmFile, FS, ready as h5wasmReady } from 'h5wasm';
 
 import type { H5WasmAttribute, H5WasmEntity } from './models';
 import {
@@ -40,26 +40,21 @@ export class H5WasmApi extends ProviderApi {
   }
 
   public async getEntity(path: string): Promise<Entity> {
-    const file = await this.file;
-    const h5wEntity = file.get(path);
-
+    const h5wEntity = await this.getH5WasmEntity(path);
     return this.processH5WasmEntity(getNameFromPath(path), path, h5wEntity);
   }
 
   public async getValue(params: ValuesStoreParams): Promise<unknown> {
     const { dataset } = params;
 
-    const file = await this.file;
-    const h5wDataset = file.get(dataset.path);
+    const h5wDataset = await this.getH5WasmEntity(dataset.path);
     assertH5WasmDataset(h5wDataset);
 
-    const array = h5wDataset.value;
-    return hasScalarShape(dataset) ? array[0] : array;
+    return h5wDataset.value;
   }
 
   public async getAttrValues(entity: Entity): Promise<AttributeValues> {
-    const file = await this.file;
-    const h5wEntity = file.get(entity.path);
+    const h5wEntity = await this.getH5WasmEntity(entity.path);
     assertH5WasmEntityWithAttrs(h5wEntity);
 
     return Object.fromEntries(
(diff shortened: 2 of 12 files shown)


packages/app/src/types.d.ts
```diff
@@ -1,4 +1,3 @@
 /* eslint-disable spaced-comment */
 
 /// <reference types="vite/client" />
-/// <reference types="@h5web/shared/src/types" />
```
axelboc, line 4: Left over from an attempt to fix type issues during package builds.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -10,12 +10,12 @@ import type {
...
 } from '@h5web/shared';
 import {
+  assertDefined,
   buildEntityPath,
   DTypeClass,
   EntityKind,
-  hasScalarShape,
 } from '@h5web/shared';
-import { File as H5WasmFile, FS, ready } from 'h5wasm';
+import { File as H5WasmFile, FS, ready as h5wasmReady } from 'h5wasm';
```
axelboc, line 18: Small renaming so the code is more self-documenting.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -40,26 +40,21 @@ export class H5WasmApi extends ProviderApi {
   }
 
   public async getEntity(path: string): Promise<Entity> {
-    const file = await this.file;
-    const h5wEntity = file.get(path);
-
+    const h5wEntity = await this.getH5WasmEntity(path);
```
axelboc, line 43: We now need to check that the entity returned is defined and throw an entity-not-found error if it isn't, so I've moved this to a separate method.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -40,26 +40,21 @@ export class H5WasmApi extends ProviderApi {
...
     const { dataset } = params;
 
-    const file = await this.file;
-    const h5wDataset = file.get(dataset.path);
+    const h5wDataset = await this.getH5WasmEntity(dataset.path);
     assertH5WasmDataset(h5wDataset);
 
-    const array = h5wDataset.value;
-    return hasScalarShape(dataset) ? array[0] : array;
+    return h5wDataset.value;
```
axelboc, line 53: The new, stricter type of `value` allowed me to identify a bug here: this logic is not needed since h5wasm returns scalars directly.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -75,12 +70,24 @@ export class H5WasmApi extends ProviderApi {
   }
 
   private async initFile(buffer: ArrayBuffer): Promise<H5WasmFile> {
-    await ready;
+    await h5wasmReady;
+
+    // Write file to Emscripten virtual file system
+    // https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.writeFile
+    FS?.writeFile(this.filepath, new Uint8Array(buffer), { flags: 'w+' });
```
axelboc, line 77: @bmaranville any reason why `FS` can now be `null`?

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -75,12 +70,24 @@ export class H5WasmApi extends ProviderApi {
...
 
-    FS.writeFile(this.filepath, new Uint8Array(buffer), { flags: 'w+' });
     return new H5WasmFile(this.filepath, 'r');
   }
 
+  private async getH5WasmEntity(path: string): Promise<H5WasmEntity> {
+    const file = await this.file;
+
+    const h5wEntity = file.get(path) || undefined;
+    assertDefined(h5wEntity, `No entity found at ${path}`);
```
axelboc, line 86: Note that `|| undefined` is due to the fact that `assertDefined` asserts only for `undefined`, not `null` (which is a good thing, we don't want to change that). I felt creating an `assertNonNull` guard was unnecessary.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -101,7 +108,9 @@ export class H5WasmApi extends ProviderApi {
       }
 
       const children = h5wEntity.keys().map((name) => {
-        const h5wChild = h5wEntity.get(name);
+        const h5wChild = h5wEntity.get(name) || undefined;
+        assertDefined(h5wChild);
```
axelboc, line 112: Not bothering with a user-friendly error message here, since if the child doesn't exist, something's wrong with h5wasm, not with the HDF5 file.

packages/h5wasm/src/h5wasm-api.ts
axelboc: This fixes the issue with H5Web showing only the first character for scalar datasets, like `{` for datasets containing JSON.

packages/h5wasm/src/h5wasm-api.ts
bmaranville: It is exported at the top level, and was previously initialized as null, but typed as Emscripten.FileSystemType.  This was caught by strict type checking and I updated the type.

This one is a little tricky, as the value of FS is replaced by the wasm Module.FS when the wasm is initialized, so if you await h5wasm.ready before using FS it will always be the right thing.  It's possible FS should be presented as a promise, instead.

axelboc: Right, I see. I'll replace `?.` with `!.` for now, then.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -102,6 +113,8 @@ export class H5WasmApi extends ProviderApi {
 
       const children = h5wEntity.keys().map((name) => {
         const h5wChild = h5wEntity.get(name);
+        assertNonNull(h5wChild);
```
loichuder, line 116: Perhaps worth putting a more meaningful error message as in `getH5WasmEntity` ?

loichuder (approved):
Nice !

packages/h5wasm/src/h5wasm-api.ts
loichuder: Arf just saw your comment https://github.com/silx-kit/h5web/pull/1082#discussion_r858522417. Discard this comment then.

PR #1084, opened 2022-04-27
Display chunks and compression info in dataset metadata

### LINKED ISSUE #311: [Inspect] Give info about compression and chunking of datasets
This would require the backend to return the compression information to be displayed by h5web

loichuder:
Fix #311 

Requires h5grove [v0.0.16](https://github.com/silx-kit/h5grove/releases/tag/v0.0.16)

**Updated** example on the compressed dataset in **water_224.h5**:
image

You can also look at [compressed.h5](http://localhost:3000/h5grove?file=compressed.h5) for more examples

### DIFF
--- a/packages/app/src/metadata-viewer/FiltersInfo.tsx
+++ b/packages/app/src/metadata-viewer/FiltersInfo.tsx
@@ -1,0 +1,26 @@
+import type { Filter } from '@h5web/shared';
+
+interface Props {
+  filters: Filter[];
+}
+
+function FiltersInfo(props: Props) {
+  const { filters } = props;
+
+  return (
+    <>
+      {filters.map((filter) => {
+        const { name, id } = filter;
+
+        return (
+          <tr key={id}>
+            <th scope="row">{id}</th>
+            <td>{name}</td>
+          </tr>
+        );
+      })}
+    </>
+  );
+}
+
+export default FiltersInfo;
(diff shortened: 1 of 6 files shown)


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -172,11 +172,19 @@ export class H5GroveApi extends ProviderApi {
...
+        shape,
+        chunks,
+        filters,
+      } = response;
       const attributes = await this.processAttrsMetadata(path, attrsMetadata);
 
       return {
         ...baseEntity,
+        chunks: chunks ? chunks : undefined,
+        filters: filters ? filters : undefined,
```
loichuder, line 187: Ideally, I would prefer the fields to be absent if `null`.

I don't know if there is a better way to do this ?

packages/app/src/metadata-viewer/FiltersInfo.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  return (
+    <>
+      {filters.map((filter, i) => {
+        const { name, id } = filter;
+
+        return (
+          <tr key={id}>
+            <th scope="row">{`Filter ${i}`}</th>
+            <td>{`${name} (HDF5 id: ${id})`}</td>
```
axelboc, line 18: Not sure showing the index of the filter is worth it. Isn't the order sufficient? You could remove the `th` and have `colspan="2"` on the `td`, or you could move the filters to the dataset's core metadata table, on a single row (maybe with an `ol` inside the `td`).

axelboc: Or you could put the HDF5 ID in the `th`, maybe people will know what it's about without further clue?

PR #1085, opened 2022-04-28
Revert CI triggers but skip double runs

axelboc:
So I've looked a bit more into it and it really sucks... Only way I could find is to add a condition to each job to skip it if it's triggered by a local PR. Downside is that the skipped jobs still appear in the UI.

Running the CI on `push` to any branch is quite convenient in my opinion for a few reasons:

- By the time you open the PR, the CI is likely to have already passed, which means you can be sure it's ready to be reviewed. This saves us a lot of time on a day-to-day basis.
- If you push a new branch and you know you forgot to fix the linting or the tests, you can cancel the CI run or force-push the fixes before opening the PR.
- If you work on something that might affect the screenshots or the CI workflows themselves, it can be useful have a branch to push to before you can be sure that the solution you settle on is correct and worthy of a PR. Arguable, since you can always open a draft PR.

Let me know what you prefer: skipped jobs in the UI vs the points above.

image

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -1,14 +1,16 @@
 name: Lint & Test
 
 on:
+  pull_request: # for PRs from forks
   push:
-    branches: [main]
-  pull_request:
-    branches: [main]
+    branches-ignore:
+      - 'update-snapshots'
 
 jobs:
   lint:
     runs-on: ubuntu-latest
+    # Skip `pull_request` runs on local PRs for which `push` runs are already triggered
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     steps:
       - name: Checkout 🏷️
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-test.yml
```diff
@@ -1,14 +1,16 @@
...
   pull_request:
-    branches: [main]
+  push:
+    branches-ignore:
+      - 'update-snapshots'
 
 jobs:
   lint:
     runs-on: ubuntu-latest
+    # Skip `pull_request` runs on local PRs where `push` is already triggered
```
loichuder, line 12:
```suggestion
    # Run only on local `push` events or `pull_request` events from forks
```

loichuder (approved):
I can live with skipped jobs

.github/workflows/lint-test.yml
axelboc: I find my comment makes the actual purpose of the condition clearer (i.e. to skip some runs). I've added a comment next to the `pull_request` trigger to mention that it's to cover PRs created from forked repos.

axelboc:
I also replaced the hardcoded repo name in the conditions with `github.repository` as you suggested, so that the jobs don't get skipped for local PRs in forked repositories.

PR #1087, opened 2022-04-28
Try to find axes on the signal when not present in the NXData

### LINKED ISSUE #1062: Reconsider support for older NeXus files (finding signal and axes)

There is a considerable body of existing NeXus files that are written to an earlier version of the standard.  The h5web project does not support features like ```@signal``` on datasets (instead of on the NXdata group) pre-NIAC2014.

As I understand it was decided by the h5web maintainers to not support older files as a matter of policy, in order to encourage users to update their NeXus implementations to a newer version.  I would ask that this policy be reconsidered, as the complexity and performance cost of adding some key features (finding the signal and axes in versions 1 and 2 of the NeXus standard) does not seem very high, and the end-users that are impacted are not able to upgrade their old NeXus files themselves.

We will write future NeXus files according to the current standard, but we would like to be able to easily visualize the large existing (immutable) corpus of archived data.  Adding transformer middleware on the server side (through a provider like H5Grove) is not possible.  We could add middleware in the browser (service worker?), but this is much more complex.

### Additional context

The standard is not fixed and will change again, and maintaining backward compatibility seems important.

loichuder:
Fix #1062


packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -124,6 +124,53 @@ export function findAssociatedDatasets(
...
+  return [dsetList];
+}
+
+export function findOldStyleAxesDatasets(
+  group: GroupWithChildren,
+  signal: Dataset,
+  attrValuesStore: AttrValuesStore
+): NumArrayDataset[] {
+  const axesList = attrValuesStore.getSingle(signal, 'axes') || [];
+  const axesNames = parseAxesList(axesList);
```
axelboc, line 150: I don't believe the spec mentions that `axes` can be an array. To simplify things, I would just return an empty array if the value of the attribute is not a string (with a type guard, not an assertion, so it doesn't crash). This would ensure that we always end up with an array of strings without having to double check the type of the `axes` attribute.

packages/shared/src/mock/metadata.ts
```diff
@@ -89,6 +89,21 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
                 }),
               ],
             }),
+            makeNxGroup('old-style_signal_axes', 'NXdata', {
```
axelboc, line 92: should we just have one entity called `old-style` to test everything at once?

packages/shared/src/mock/metadata.ts
```diff
@@ -89,6 +89,21 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
...
+              children: [
+                makeDataset('twoD', intType, [20, 41], {
+                  attributes: [
+                    makeIntAttr('signal', 1),
+                    makeStrAttr('axes', 'Y,X'),
```
axelboc, line 97: I would maybe use a colon here, as the comma is an alternative proposed by the spec and not the main suggested delimiter. Testing the comma seems less important.

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -124,6 +124,53 @@ export function findAssociatedDatasets(
...
+    const dataset = getChildEntity(group, name);
+    assertDefined(dataset);
+    assertDataset(dataset);
+    assertArrayShape(dataset);
+    assertNumericType(dataset);
+    return dataset;
+  });
+}
+
+export function findAxesDatasets(
```
axelboc, line 162: :+1:

axelboc (changes requested):

packages/app/src/vis-packs/nexus/utils.ts
loichuder: Good catch. I think I was still considering treating the `axes` of version 3 but it is not relevant here.

axelboc (approved):

PR #1090, opened 2022-05-02
Let consumers of data providers access context

axelboc:
- `Provider` becomes `DataProvider`
- `ProviderContext` becomes `DataContext`
- `Context` type becomes `DataContextValue`
- The context and its type are now located in `DataProvider.tsx`
- Add hook `useDataContext` and expose in @h5web/app.
- Document API of @h5web/app in README (built-in providers and context value).


packages/app/README.md
```diff
@@ -129,3 +210,51 @@ import { getFeedbackMailto } from '@h5web/app';
...
+}
+```
+
+The three stores are created with the
+[react-suspense-fetch](https://github.com/dai-shi/react-suspense-fetch) library,
+which relies on
+[React Suspense](https://reactjs.org/docs/react-api.html#reactsuspense). A
+component that uses one of these stores (e.g.
+`entitiesStore.get('/path/to/entity')`) must have a `Suspense` ancestor to
+manage the loading state.
```
loichuder, line 260: Suggestion: you could add a simple example here.

packages/app/README.md
```diff
@@ -129,3 +210,51 @@ import { getFeedbackMailto } from '@h5web/app';
...
+
+  getExportURL?: (
+    dataset: Dataset<ArrayShape, NumericType>,
+    selection: string | undefined,
+    format: ExportFormat
+  ) => string | undefined;
+
+  addProgressListener: (cb: ProgressCallback) => void;
+  removeProgressListener: (cb: ProgressCallback) => void;
+}
```
loichuder, line 251: It is worth mentioning those ?

packages/app/src/providers/mock/MockProvider.tsx
```diff
@@ -1,18 +1,14 @@
...
-import Provider from '../Provider';
+import DataProvider from '../DataProvider';
 import { MockApi } from './mock-api';
 
-interface Props {
-  children: ReactNode;
-}
-
-function MockProvider(props: Props) {
+function MockProvider(props: PropsWithChildren<Record<never, never>>) {
```
loichuder, line 7: Don't you think it would be clearer to keep the `Props` interface here ?

loichuder (approved):
:+1:

packages/app/src/providers/mock/MockProvider.tsx
axelboc: I put back an empty `interface Props {}`

packages/app/README.md
axelboc: Yeah no, you're right.

axelboc:
Thanks for the great suggestions 👍

PR #1091, opened 2022-05-02
Update h5wasm to 0.4.0

loichuder:
https://github.com/usnistgov/h5wasm/releases/tag/v0.4.0

The typing of attributes gave me a headache. See below.

### DIFF
--- a/packages/h5wasm/src/h5wasm-api.ts
+++ b/packages/h5wasm/src/h5wasm-api.ts
@@ -2,11 +2,14 @@ import type { ValuesStoreParams } from '@h5web/app';
 import { convertDtype, ProviderApi } from '@h5web/app';
 import { getNameFromPath } from '@h5web/app/src/providers/utils';
 import type {
+  Attribute,
   AttributeValues,
   Dataset,
+  DType,
   Entity,
   Group,
   GroupWithChildren,
+  Shape,
   UnresolvedEntity,
 } from '@h5web/shared';
 import {
@@ -15,9 +18,13 @@ import {
   DTypeClass,
   EntityKind,
 } from '@h5web/shared';
-import { File as H5WasmFile, FS, ready as h5wasmReady } from 'h5wasm';
+import type {
+  Attribute as H5WasmAttribute,
+  Dtype as H5WasmDtype,
+} from 'h5wasm';
+import { File as H5WasmFile, ready as h5wasmReady } from 'h5wasm';
 
-import type { H5WasmAttribute, H5WasmEntity } from './models';
+import type { H5WasmAttributes, H5WasmEntity } from './models';
 import {
   assertH5WasmDataset,
   assertH5WasmEntityWithAttrs,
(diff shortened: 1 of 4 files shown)


packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -59,7 +64,9 @@ export class H5WasmApi extends ProviderApi {
 
     return Object.fromEntries(
       Object.entries(h5wEntity.attrs).map(([name, attr]) => {
-        return [name, (attr as H5WasmAttribute).value];
+        assertH5WasmAttribute(attr);
```
loichuder, line 67: So `h5wEntity.attrs` is now typed as `{[key: string]: {get(): Attribute, enumerable: true}}`.

As a result, `attr` type gets infered as `{get(): Attribute, enumerable: true}` for some reason while the actual runtime object `attr` has the signature of `H5WasmAttribute`. 

There I cannot access `dtype`, `shape` or `value` without asserting first the type. I tried to find a more elegant workaround but could not.

Some relevant issues that I came across that could help understand:
- https://github.com/microsoft/TypeScript/issues/28694
- https://github.com/microsoft/TypeScript/issues/38115
- https://github.com/microsoft/TypeScript/issues/42919

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -144,10 +148,23 @@ export class H5WasmApi extends ProviderApi {
...
+  private processH5WasmAttrs(h5wAttrs: H5WasmAttributes): Attribute[] {
     return Object.entries(h5wAttrs).map(([name, attr]) => {
-      const { shape, dtype } = attr as H5WasmAttribute;
-      return { name, shape, type: convertDtype(dtype) };
+      assertH5WasmAttribute(attr);
+      const { shape, dtype } = attr;
+      return {
+        name,
+        shape: shape as Shape,
+        type: this.processH5WasmDtype(dtype),
```
loichuder, line 158: I plan to replace this by a processing from `metadata` in a future PR

packages/h5wasm/src/utils.ts
```diff
@@ -19,6 +20,15 @@ export function isH5WasmDataset(entity: H5WasmEntity): entity is H5WasmDataset {
...
 }
 
+export function isH5WasmAttribute(attr: unknown): attr is H5WasmAttribute {
+  const propertyNames = Object.getOwnPropertyNames(attr);
+  return (
+    propertyNames.includes('value') &&
+    propertyNames.includes('shape') &&
+    propertyNames.includes('metadata')
+  );
+}
```
loichuder, line 30: Happy to change this if you have a better idea

packages/h5wasm/src/h5wasm-api.ts
axelboc: So is the type cast `as H5WasmAttribute` no longer working because TS thinks the types are too different? If so, then maybe `as unknown as H5WasmAttribute` would work? It's not very elegant, but it has the advantage of clearly pointing out the hack (as opposed to an assertion, which has a valid purpose - i.e. typically to narrow down something from a broader type). I'm not opposed to the assertion if you prefer it to a type cast, though.

axelboc (approved):

PR #1092, opened 2022-05-02
Upgrade pnpm to v7

axelboc:
Lots of breaking changes: https://github.com/pnpm/pnpm/releases/tag/v7.0.0

I'll try to explain them as I go.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 - [Quick start](#quick-start-)
 - [Development](#development)
-  - [`pnpm` cheat sheet](#pnpm-cheat-sheet)
+  - [`pnpm` v7 cheat sheet](#pnpm-v7-cheat-sheet)
   - [Dependency management](#dependency-management)
   - [Workspace dependencies](#workspace-dependencies)
   - [Icon set](#icon-set)
@@ -30,20 +30,20 @@ pnpm start
   [Storybook](https://storybook.js.org/docs/react/get-started/introduction)
   documentation site at http://localhost:6006
 
-### `pnpm` cheat sheet
+### `pnpm` v7 cheat sheet
 
 - `pnpm install` - install the dependencies of every project in the workspace
   and of the workspace itself
-- `pnpm add [-D] <pkg-name> --filter <project-name>` -
+- `pnpm --filter <project-name> add [-D] <pkg-name>` -
   [add a dependency](https://pnpm.io/cli/add) to a project in the workspace
-- `pnpm [script] [-- --<arg>]` - run a workspace script
-- `pnpm [script] [--parallel] --filter {packages} [-- --<arg>]` -
+- `pnpm [run] <script> [--<arg>]` - run a workspace script
+- `pnpm [run] --filter {packages/*} [--parallel] <script> [--<arg>]` -
   [run a script](https://pnpm.io/cli/run) in every project in the `packages`
   folder
-- `pnpm dlx <pkg-name>` - fetch a package from the registry and run its default
+- `pnpm [exec] <binary>` - run a binary located in `node_modules/.bin`
+  (equivalent to `npx <pkg-name>` for a package installed in the workspace)
+- `pnpx <pkg-name>` - fetch a package from the registry and run its default
   command binary (equivalent to `npx <pkg-name>`)
-- `pnpm exec <binary>` - run a binary located in `node_modules/.bin` (equivalent
-  to `npx <pkg-name>` for a package installed in the workspace)
 - `pnpm why -r <pkg-name>` - show all project and packages that depend on the
   specified package
 - `pnpm outdated -r` - list outdated dependencies in the workspace
@@ -56,9 +56,9 @@ pnpm start
 1. Read the changelogs and release notes of the dependencies you'd like to
    upgrade. Look for potential breaking changes, and for bug fixes and new
    features that may help improve the codebase.
-1. Run `pnpm up -r -L <pkg-name>` to update a dependency to the latest version
-   in all projects. Alternatively, you can either replace `-r` with `--filter`
-   to target specific projects, or edit the relevant `package.json` file(s)
+1. Run `pnpm up -rL <pkg-name>` to update a dependency to the latest version in
+   all projects. Alternatively, you can either replace `-r` with `--filter` to
+   target specific projects, or edit the relevant `package.json` file(s)
    manually and run `pnpm install` (but make sure to specify an exact dependency
    version rather than a range - i.e. don't prefix the version with a caret or a
    tilde).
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -11,21 +11,21 @@
...
+    "build": "pnpm --filter demo build",
+    "build:storybook": "pnpm --filter storybook build",
+    "serve": "pnpm --filter demo serve",
+    "serve:storybook": "pnpm --filter storybook serve",
+    "packages": "pnpm --filter {packages/*} build",
+    "lint": "run-p lint:**",
+    "lint:eslint": "pnpm -r --parallel lint:eslint",
+    "lint:tsc": "pnpm -r --parallel lint:tsc",
+    "lint:cypress:tsc": "tsc --project cypress/tsconfig.json",
+    "lint:root:eslint": "eslint \"**/*.{js,cjs,ts,tsx}\" --max-warnings=0",
```
axelboc, line 28: `lint:all:<eslint|tsc>` (`pnpm -r --parallel lint:<eslint|tsc>`) no longer runs the root `lint:<eslint|tsc>` script, which means that I was able to rename some of the root scripts:

- `lint:eslint` becomes `lint:root:eslint`
- `lint:tsc` becomes `lint:cypress:tsc`
- `lint:all:eslint` becomes `lint:eslint` 
- `lint:all:tsc` becomes `lint:tsc`

A couple of notes:

- The `lint` script now runs all the `lint:` scripts, including `lint:root:eslint` and `lint:cypress:tsc`
- I realised that running `tsc` on the whole workspace made no sense since we don't have `ts` files outside of the projects' `src` folders and `cypress` folder, which are all already covered by a `tsconfing` and `lint:tsc` script. We were actually running `tsc` twice on every file... Anyway, so that's why `lint:cypress:tsc` is called that and not `lint:root:tsc`, and that's why I removed the root `tsconfig.lint.ts`.

package.json
```diff
@@ -55,6 +55,12 @@
       "react-scripts>eslint-plugin-testing-library": ">=5",
       "react-scripts>@typescript-eslint/parser": ">=5",
       "react-scripts>@typescript-eslint/eslint-plugin": ">=5"
+    },
+    "peerDependencyRules": {
+      "allowedVersions": {
+        "cypress": "*",
+        "jest": "*"
+      }
```
axelboc, line 63: pnpm enables [`strict-peer-dependencies`](https://pnpm.io/npmrc#strict-peer-dependencies) by default, now, so I was getting some errors due to `cypress-image-snapshots`. This is the only way I managed to solve them.

CONTRIBUTING.md
```diff
@@ -30,20 +30,20 @@ pnpm start
...
 
 - `pnpm install` - install the dependencies of every project in the workspace
   and of the workspace itself
-- `pnpm add [-D] <pkg-name> --filter <project-name>` -
+- `pnpm --filter <project-name> add [-D] <pkg-name>` -
   [add a dependency](https://pnpm.io/cli/add) to a project in the workspace
-- `pnpm [script] [-- --<arg>]` - run a workspace script
-- `pnpm [script] [--parallel] --filter {packages} [-- --<arg>]` -
+- `pnpm [run] <script> [--<arg>]` - run a workspace script
+- `pnpm [run] --filter {packages/*} [--parallel] <script> [--<arg>]` -
```
axelboc, line 40: Note the new glob syntax for filtering folders.

packages/app/jest.config.js
```diff
@@ -17,7 +17,7 @@ const config = {
   preset: 'ts-jest/presets/js-with-ts',
   resetMocks: true,
   roots: ['<rootDir>/src'],
-  setupFilesAfterEnv: ['<rootDir>/src/setupTests.js'],
+  setupFilesAfterEnv: ['<rootDir>/src/setupTests.ts'],
```
axelboc, line 20: I had some typing issues in tests and thought this might fix it: https://github.com/testing-library/jest-dom#usage. Turns out I had to install the `@types` package explicitly to ensure TypeScript could find it. Since converting the file to TS caught some useful issues, I decided to keep the change.

CONTRIBUTING.md
```diff
@@ -30,20 +30,20 @@ pnpm start
   [Storybook](https://storybook.js.org/docs/react/get-started/introduction)
   documentation site at http://localhost:6006
 
-### `pnpm` cheat sheet
+### `pnpm` v7 cheat sheet
 
 - `pnpm install` - install the dependencies of every project in the workspace
   and of the workspace itself
-- `pnpm add [-D] <pkg-name> --filter <project-name>` -
+- `pnpm --filter <project-name> add [-D] <pkg-name>` -
```
loichuder, line 37: Arg placement needs to be changed only for `pnpm run` no ?

package.json
loichuder: > I realised that running tsc on the whole workspace made no sense since we don't have ts files outside of the projects' src folders and cypress folder, which are all already covered by a tsconfing and lint:tsc script.

Looking back at https://github.com/silx-kit/h5web/pull/911, I think it was to ensure also that the storybook was linted ? Is it still the case ?

loichuder (changes requested):
I think the caching of `pnpm` in the CI has to be reviewed as the path to the pnpm store changed from `~/.pnpm-store` to `~/.local/share/pnpm/store`.

CONTRIBUTING.md
axelboc: I hesitated, but It matches what is recommended in the documentation: https://pnpm.io/filtering -- filtering is a global feature of the CLI, so it makes sense. Also, I think it's good to get into the habit of typing it first to avoid making mistakes with the `run` command.

package.json
axelboc: I don't think adding linting to the root was related to the Storybook, as the `storybook` project does have its own `lint:tsc` and `lint:eslint` scripts.

axelboc:
Good catch!!

axelboc:
Looks like it's working:

image

CONTRIBUTING.md
loichuder: Sensible.

package.json
loichuder: Yes but these weren't called for some reasons.

It appears to be fine with the new setup though.

loichuder (approved):

netlify.toml
```diff
@@ -7,4 +7,4 @@
 [build.environment]
   NPM_FLAGS = "--version"
 [build]
-  command = "npx pnpm install --store=node_modules/.pnpm-store && npx pnpm build"
+  command = "npx pnpm@7 install --store=node_modules/.pnpm-store && npx pnpm@7 build"
```
axelboc, line 10: I'm setting an explicit version here so our builds don't fail again on the next breaking change...

axelboc:
Arf... https://github.com/silx-kit/h5web/runs/6269954689?check_suite_focus=true

I'm going to try to upgrade the GitHub actions -- I think the versions we're using may not be compatible with pnpm v7.

axelboc:
Looks like we're gonna have to wait a bit for pnpm v7 support in the CI: https://github.com/pnpm/action-setup/issues/38

I'm going to switch this to draft for now and push a fix on `main` for the Netlify builds.

axelboc:
Turns out I was doing things wrong in the CI. The order of the actions has to be as follows:

1. `pnpm/action-setup` to install pnpm v7 ... but without running `pnpm install` (i.e. the `run-install` option must not be specified)
2. `actions/setup-node` to install Node and cache/restore pnpm packages (for the caching to work, pnpm has to already be installed, which is why it must be run after `pnpm/action-setup`)
3. Install dependencies with `pnpm install --frozen-lockfile`

loichuder (approved):

PR #1096, opened 2022-05-03
TiledHeatmapMesh: added supports of scene transforms

t20100:
This PR renders usage of TiledHeatmapMesh more generic.
The computation of requested tiles no longer relies on the axes (thanks @axelboc!) and should also handle almost any transforms.

Screen Shot 2022-05-03 at 18 27 19

### DIFF
--- a/apps/storybook/src/TiledHeatmapMesh.stories.tsx
+++ b/apps/storybook/src/TiledHeatmapMesh.stories.tsx
@@ -132,7 +132,7 @@ interface TiledHeatmapStoryProps extends TiledHeatmapMeshProps {
 }
 
 const Template: Story<TiledHeatmapStoryProps> = (args) => {
-  const { abscissaConfig, api, ordinateConfig, ...tiledHeatmapProps } = args;
+  const { abscissaConfig, ordinateConfig, ...tiledHeatmapProps } = args;
 
   return (
     <VisCanvas
@@ -150,7 +150,7 @@ const Template: Story<TiledHeatmapStoryProps> = (args) => {
       <group
         scale={[abscissaConfig.flip ? -1 : 1, ordinateConfig.flip ? -1 : 1, 1]}
       >
-        <TiledHeatmapMesh api={api} {...tiledHeatmapProps} />
+        <TiledHeatmapMesh {...tiledHeatmapProps} />
       </group>
     </VisCanvas>
   );
@@ -224,7 +224,7 @@ FlippedAxes.args = {
 
 function LinearAxesGroup(props: { children: ReactNode }) {
   const { children } = props;
-  const { visSize, abscissaConfig, ordinateConfig } = useAxisSystemContext();
+  const { abscissaConfig, ordinateConfig, visSize } = useAxisSystemContext();
   const { width, height } = visSize;
   const sx =
     ((abscissaConfig.flip ? -1 : 1) * width) /
@@ -236,7 +236,7 @@ function LinearAxesGroup(props: { children: ReactNode }) {
   const y = 0.5 * (ordinateConfig.visDomain[0] + ordinateConfig.visDomain[1]);
 
   return (
-    <group scale={[sx, sy, 1]} position={[-x * sx, -y * sy, 0]}>
+    <group position={[-x * sx, -y * sy, 0]} scale={[sx, sy, 1]}>
       {children}
     </group>
   );
@@ -283,7 +283,7 @@ WithTransforms.args = {
     visDomain: [0, 2],
     isIndexAxis: true,
     showGrid: false,
-    flip: true,
+    flip: false,
   },
 };
 
--- a/apps/storybook/src/TiledHeatmapMesh.stories.tsx
+++ b/apps/storybook/src/TiledHeatmapMesh.stories.tsx
@@ -147,7 +147,11 @@ const Template: Story<TiledHeatmapStoryProps> = (args) => {
       <Zoom />
       <SelectToZoom keepRatio modifierKey="Control" />
       <ResetZoomButton />
-      <TiledHeatmapMesh api={api} {...tiledHeatmapProps} />
+      <group
+        scale={[abscissaConfig.flip ? -1 : 1, ordinateConfig.flip ? -1 : 1, 1]}
+      >
+        <TiledHeatmapMesh api={api} {...tiledHeatmapProps} />
+      </group>
     </VisCanvas>
   );
 };
@@ -223,16 +227,25 @@ function LinearAxesGroup(props: { children: ReactNode }) {
   const { visSize, abscissaConfig, ordinateConfig } = useAxisSystemContext();
   const { width, height } = visSize;
   const sx =
-    width / (abscissaConfig.visDomain[1] - abscissaConfig.visDomain[0]);
+    ((abscissaConfig.flip ? -1 : 1) * width) /
+    (abscissaConfig.visDomain[1] - abscissaConfig.visDomain[0]);
   const sy =
-    height / (ordinateConfig.visDomain[1] - ordinateConfig.visDomain[0]);
+    ((ordinateConfig.flip ? -1 : 1) * height) /
+    (ordinateConfig.visDomain[1] - ordinateConfig.visDomain[0]);
+  const x = 0.5 * (abscissaConfig.visDomain[0] + abscissaConfig.visDomain[1]);
+  const y = 0.5 * (ordinateConfig.visDomain[0] + ordinateConfig.visDomain[1]);
 
-  return <group scale={[sx, sy, 1]}>{children}</group>;
+  return (
+    <group scale={[sx, sy, 1]} position={[-x * sx, -y * sy, 0]}>
+      {children}
+    </group>
+  );
 }
 
 export const WithTransforms: Story<TiledHeatmapStoryProps> = (args) => {
   const { abscissaConfig, api, ordinateConfig, ...tiledHeatmapProps } = args;
-  const size = { width: 1, height: 1 };
+  const { baseLayerSize } = api;
+  const size = { width: 1, height: baseLayerSize.height / baseLayerSize.width };
 
   return (
     <VisCanvas
@@ -248,7 +261,7 @@ export const WithTransforms: Story<TiledHeatmapStoryProps> = (args) => {
       <SelectToZoom keepRatio modifierKey="Control" />
       <ResetZoomButton />
       <LinearAxesGroup>
-        <group rotation={[0, 0, Math.PI / 4]} position={[1, -1, 0]}>
+        <group position={[1, 1, 0]} rotation={[0, 0, Math.PI / 4]}>
           <TiledHeatmapMesh api={api} {...tiledHeatmapProps} size={size} />
         </group>
         <group position={[-1, 1, 0]} scale={[2, 2, 1]}>
@@ -259,16 +272,18 @@ export const WithTransforms: Story<TiledHeatmapStoryProps> = (args) => {
   );
 };
 WithTransforms.args = {
-  api: defaultApi,
+  api: halfMandelbrotApi,
   abscissaConfig: {
-    visDomain: [-2, 2],
+    visDomain: [-2, 1.5],
     isIndexAxis: true,
     showGrid: false,
+    flip: false,
   },
   ordinateConfig: {
-    visDomain: [-2, 2],
+    visDomain: [0, 2],
     isIndexAxis: true,
     showGrid: false,
+    flip: true,
   },
 };
 
--- a/apps/storybook/src/TiledHeatmapMesh.stories.tsx
+++ b/apps/storybook/src/TiledHeatmapMesh.stories.tsx
@@ -7,6 +7,7 @@ import {
   TilesApi,
   ResetZoomButton,
   SelectToZoom,
+  useAxisSystemContext,
 } from '@h5web/lib';
 import type {
   Domain,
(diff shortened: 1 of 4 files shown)


t20100:
There's still a few things I need to test, like transforms + flipped axes...

apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -216,6 +218,60 @@ FlippedAxes.args = {
   },
 };
 
+function LinearAxesGroup(props: { children: ReactNode }) {
```
t20100, line 221: Wouldn't that make sense to have this in the lib at some point?
(I'll something like this soon)

apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -216,6 +218,60 @@ FlippedAxes.args = {
...
+  const { width, height } = visSize;
+  const sx =
+    width / (abscissaConfig.visDomain[1] - abscissaConfig.visDomain[0]);
+  const sy =
+    height / (ordinateConfig.visDomain[1] - ordinateConfig.visDomain[0]);
+
+  return <group scale={[sx, sy, 1]}>{children}</group>;
+}
+
+export const WithTransforms: Story<TiledHeatmapStoryProps> = (args) => {
```
t20100, line 233: Need to test with flipped axes

apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -216,6 +218,60 @@ FlippedAxes.args = {
...
+function LinearAxesGroup(props: { children: ReactNode }) {
+  const { children } = props;
+  const { visSize, abscissaConfig, ordinateConfig } = useAxisSystemContext();
+  const { width, height } = visSize;
+  const sx =
+    width / (abscissaConfig.visDomain[1] - abscissaConfig.visDomain[0]);
+  const sy =
+    height / (ordinateConfig.visDomain[1] - ordinateConfig.visDomain[0]);
+
+  return <group scale={[sx, sy, 1]}>{children}</group>;
```
t20100, line 230: Need to handle the `position` as well

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -1,28 +1,25 @@
...
 import type { AxisSystemContextValue } from '../shared/AxisSystemProvider';
-import { createAxisScale, getVisibleDomains, isDescending } from '../utils';
+import { createAxisScale } from '../utils';
 
-export function getTileOffsets(
-  xDomain: Domain,
-  yDomain: Domain,
-  tileSize: Size
-): Vector2[] {
+export function getTileOffsets(box: Box2, tileSize: Size): Vector2[] {
```
t20100, line 11: Using `Box2` ended-up being simpler than `Domain`s

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -50,32 +47,51 @@ export function sortTilesByDistanceTo(
   });
 }
 
-export function getScaledVisibleDomains(
+const NDC_BOX = new Box3(new Vector3(-1, -1, -1), new Vector3(1, 1, 1));
+
+function getObject3DVisibleBox(
```
t20100, line 52: This returns the bounding box in the object's coordinates of the threejs displayed box

packages/lib/src/vis/tiles/TiledLayer.tsx
```diff
@@ -1,51 +1,61 @@
...
 
-  const { abscissaConfig, ordinateConfig, visSize } = useAxisSystemContext();
-  const { xVisibleDomain, yVisibleDomain } = useCameraState(
-    (...args) => getScaledVisibleDomains(...args, layerSize),
-    [layerSize]
+  const groupRef = useRef<Group>(null);
+  const { abscissaConfig, ordinateConfig } = useAxisSystemContext();
+  const box = useCameraState(
+    (...args) => getScaledVisibleBox(...args, meshSize, layerSize, groupRef),
+    [meshSize, layerSize, groupRef]
```
t20100, line 33: Should I pass the `ref` here or `ref.current`?

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -50,32 +47,51 @@ export function sortTilesByDistanceTo(
...
+  camera: Camera,
+  context: AxisSystemContextValue,
```
t20100, line 54: This is not used, but I kept it to be compatible with `useCameraState`... but not sure it is useful.

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
```diff
@@ -1,71 +1,89 @@
...
 
   const canvasSize = useThree((state) => state.size);
   const { visSize } = useAxisSystemContext();
+  const meshSize = size ?? visSize;
 
-  const { xVisibleDomain, yVisibleDomain } = useCameraState(
-    (...args) => getScaledVisibleDomains(...args, baseLayerSize),
-    [baseLayerSize]
-  );
+  const groupRef = useRef<Group>(null);
```
axelboc, line 36: Just wondering: could we initialize an empty `Object3D` here to somehow end up with no layers without having to deal with an undefined scaled `box`?

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
```diff
@@ -1,71 +1,89 @@
...
 
-  const roundingOffset = 1 - clamp(qualityFactor, 0, 1);
-  const subsamplingLevel = Math.min(
-    Math.floor(Math.log2(itemsPerPixel) + roundingOffset),
-    baseLayerIndex
-  );
-  const currentLayerIndex = baseLayerIndex - subsamplingLevel;
+  let layers: number[] = [];
+  if (box) {
+    const itemsPerPixel = Math.max(
```
axelboc, line 46: Seeing this variable name again, I wonder if `dataPointsPerPixel` would be clearer?

packages/lib/src/vis/tiles/TiledLayer.tsx
```diff
@@ -1,51 +1,61 @@
...
 
 interface Props extends ColorMapProps {
   api: TilesApi;
   layer: number;
+  size: Size;
 }
 
 function TiledLayer(props: Props) {
-  const { api, layer, ...colorMapProps } = props;
+  const { api, layer, size: meshSize, ...colorMapProps } = props;
```
axelboc, line 24: Why not keep the name `meshSize` for the prop?

packages/lib/src/vis/tiles/TiledLayer.tsx
```diff
@@ -1,51 +1,61 @@
...
-  const { abscissaConfig, ordinateConfig, visSize } = useAxisSystemContext();
-  const { xVisibleDomain, yVisibleDomain } = useCameraState(
-    (...args) => getScaledVisibleDomains(...args, layerSize),
-    [layerSize]
+  const groupRef = useRef<Group>(null);
+  const { abscissaConfig, ordinateConfig } = useAxisSystemContext();
+  const box = useCameraState(
+    (...args) => getScaledVisibleBox(...args, meshSize, layerSize, groupRef),
+    [meshSize, layerSize, groupRef]
   );
```
axelboc, line 35: If `useCameraState` returns the group's visible box (unscaled) like I suggested in another comment, am I wrong in thinking that the box of the group rendered by `TiledHeatmapMesh` would always be the same as the box of the groups rendered by the layers? If so, then maybe the visible box can be passed to `TiledLayer` as a prop and this component would then just have to call `getScaledVisibleBox` (or `scaleBoxToLayer` as per my other comment) with the right `layerSize`? Apart from the obvious performance improvement of one fewer `useCameraState`, it would save you from dealing with an undefined `box` below.

axelboc (changes requested):
Love your work!! 🤩

packages/lib/src/vis/tiles/TiledLayer.tsx
axelboc: The ref (assuming ESLint complains if you don't pass it), as when `ref.current` changes, it does not trigger a re-render anyway.

packages/lib/src/vis/tiles/utils.ts
axelboc: Neh, you can remove it, especially with regard to my other comments.

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -50,32 +47,51 @@ export function sortTilesByDistanceTo(
...
+    domain: [-meshSize.height / 2, meshSize.height / 2],
+    range: [0, arraySize.height],
     clamp: true,
   });
 
-  return {
-    xVisibleDomain: xVisibleDomain.map(xScale) as Domain,
-    yVisibleDomain: yVisibleDomain.map(yScale) as Domain,
-  };
+  return new Box2().setFromPoints([
```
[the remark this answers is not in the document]
t20100: It's more generic because the constructor requires `min`, `max` while `setFromPoints` computes the min/max... but you're right, here that is not needed.

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
t20100: Not sure what the resulting box would be then.
Returning an empty `Box` makes more sense to me, I'll give it a try and you'll tell what you think of it.

axelboc: Good call, seems the resulting box covers the entire FOV of the camera, because `Object3D` is initialized with a `scale` of (1,1,1).

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -50,32 +47,51 @@ export function sortTilesByDistanceTo(
   });
 }
 
-export function getScaledVisibleDomains(
+const NDC_BOX = new Box3(new Vector3(-1, -1, -1), new Vector3(1, 1, 1));
```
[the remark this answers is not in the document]
loichuder: Indeed, there is even a [getCenter](https://threejs.org/docs/index.html?q=Box2#api/en/math/Box2.getCenter) method :heart:

apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -216,6 +218,60 @@ FlippedAxes.args = {
   },
 };
 
+function LinearAxesGroup(props: { children: ReactNode }) {
+  const { children } = props;
+  const { visSize, abscissaConfig, ordinateConfig } = useAxisSystemContext();
+  const { width, height } = visSize;
+  const sx =
+    width / (abscissaConfig.visDomain[1] - abscissaConfig.visDomain[0]);
```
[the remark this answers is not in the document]
axelboc: Ooh `extent` is a great word, but on second thoughts, it's used only in the tiled heatmap/stories so it's not that big a deal for now. Looking at where it's used to compute the vis ratio, I wonder whether we could do the ratio computation from the domains inside `VisCanvas`. Anyway, something for another day.

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -50,32 +47,51 @@ export function sortTilesByDistanceTo(
...
+  const object3D = Object3DRef.current;
+  if (!object3D) {
+    return undefined;
+  }
+
+  // Convert view box: Normalized Device Coordinates -> camera -> world -> local
+  const matrix = new Matrix4()
+    .multiplyMatrices(object3D.matrixWorld.invert(), camera.matrixWorld)
+    .multiply(camera.projectionMatrixInverse);
+  return NDC_BOX.clone().applyMatrix4(matrix);
```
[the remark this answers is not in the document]
t20100: I'd prefer to do that in a separate PR.

apps/storybook/src/TiledHeatmapMesh.stories.tsx
t20100: This is now handling position, scale and flipped axes.
What's eventually missing is that it does not "cancel" transforms applied by parents of this group (that should be possible if needed), so it's meant to be used as a "root" group.

apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -216,6 +218,60 @@ FlippedAxes.args = {
...
+          (ordinateConfig.visDomain[1] - ordinateConfig.visDomain[0])
+      )}
+    >
+      <Pan />
+      <Zoom />
+      <SelectToZoom keepRatio modifierKey="Control" />
+      <ResetZoomButton />
+      <LinearAxesGroup>
+        <group rotation={[0, 0, Math.PI / 4]} position={[1, -1, 0]}>
+          <TiledHeatmapMesh api={api} {...tiledHeatmapProps} size={size} />
```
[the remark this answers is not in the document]
t20100: Not any longer here, but I fixed it in the other example

packages/lib/src/vis/tiles/utils.ts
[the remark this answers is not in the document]
t20100: I simplified it

t20100:
I did quite a few changes because the way to compute the dataPointsPerPixel was no longer correct.
It now projects the box of a pixel at the center of the screen into the coordinates of the object.
I kept the code in `useCameraState` minimal, i.e., it only recover the necessary matrix.

t20100:
The story now uses the `VisGroup` instead of defining its own.

axelboc (approved):
👍

PR #1101, opened 2022-05-06
Include link metadata for h5wasm and fix unresolved entities

### LINKED ISSUE #1100: [h5wasm] Show more info for links (not only `Unresolved`)

loichuder:
Fix #1100 

I also fixed an issue with `Unresolved` entities (such as datatypes): clicking on the `Inspect` page would show `attributes are undefined`.

### DIFF
--- a/packages/app/src/providers/h5grove/h5grove-api.ts
+++ b/packages/app/src/providers/h5grove/h5grove-api.ts
@@ -7,7 +7,6 @@ import type {
   Group,
   GroupWithChildren,
   NumericType,
-  UnresolvedEntity,
 } from '@h5web/shared';
 import { hasScalarShape, buildEntityPath, EntityKind } from '@h5web/shared';
 import { isString } from 'lodash';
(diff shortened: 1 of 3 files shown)


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -221,8 +220,9 @@ export class H5GroveApi extends DataProviderApi {
     // Treat 'other' entities as unresolved
     return {
       ...baseEntity,
+      attributes: [],
       kind: EntityKind.Unresolved,
-    } as UnresolvedEntity;
+    };
```
loichuder, line 225: `attributes` were undefined for "other" entities.

It was not caught by TS due to the type assertion `as UnresolvedEntity` that allows missing/additionnal field in an object.

To avoid such issues in the future, we should perhaps use type declaration rather than type assertion. See [this simple example](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgGJQK7DMg3gKGWQAs4BnABWAAcyAuZAIwHtmAbCOEAbkOQXbMoDMmCigA5rwC++fKEixEKAILVqHZBAAekEABMyaTNjx8AbnHEQwATxFjJMufhgYQCMMGYh+UTpAAQnD66FhgABTAZGoaEAws7JwgAJQMYaYERAIgokzkEBk4ALx4AmxCDADkEv4QIFUANCTkVLQMYhgQsnzAMFEx6hwpZkREAPTjyACy0WSSyDDAEGz6TBg4ACoAysj6zBBGIMw4AgC2GnCgfET+YBhQvlljRAB074wFRY03yNLI5GQsQ4fB6txsD18nzIhRMYHwPTcHi8Pj8AQgAHFWKE4QNgfEmKwOFw0sZwqN+D48tDYeTSrhypVkDU6g1mqRKDR6MhOt05EQ+nihhARs8JlNNrZqCh9BAEGwrHAUb4znAANaHZA7SkXBWgZB9AGLZarA1GM5zSTjEKy-S-HJ5ODChj45D096vGnfZCWax2aoSdiyhqyF53SEA4W8Ihg5Dhx75GFFBH4IA)

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -134,10 +135,36 @@ export class H5WasmApi extends ProviderApi {
...
+        },
+      };
+    }
+
     return {
       ...baseEntity,
+      attributes: [],
       kind: EntityKind.Unresolved,
-    } as UnresolvedEntity;
+    };
```
loichuder, line 167: Same here. I removed the unneeded assertion

packages/app/src/providers/h5grove/h5grove-api.ts
axelboc: Good point, the extra variable is annoying, but if it detects this sort of errors, it is indeed worth it.

axelboc (approved):

PR #1103, opened 2022-05-06
Implement slicing for h5wasm provider

### LINKED ISSUE #1102: [h5wasm] Implement slicing of datasets

loichuder:
Fix #1102

### DIFF
--- a/packages/h5wasm/src/utils.ts
+++ b/packages/h5wasm/src/utils.ts
@@ -70,3 +70,19 @@ export function convertMetadataToDType(metadata: Metadata): DType {
       };
   }
 }
+
+export function convertSelectionToRanges(
+  dataset: H5WasmDataset,
+  selection: string
+): number[][] {
+  const { shape } = dataset;
+  const selectionMembers = selection.split(',');
+
+  return selectionMembers.map((member, i) => {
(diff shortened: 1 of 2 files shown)


packages/h5wasm/src/utils.ts
```diff
@@ -70,3 +70,22 @@ export function convertMetadataToDType(metadata: Metadata): DType {
...
+  const { shape } = dataset;
+  const selectionMembers = selection.split(',');
+
+  return selectionMembers.map((member, i) => {
+    // Index (ex: 5)
+    if (!member.includes(':')) {
+      return [Number(member), Number(member) + 1];
+    }
+    // Range (ex: 5:10)
+    const [start, end] = member.split(':');
```
loichuder, line 87: I am honestly not sure this part is needed.

As far as I remember, `selection` only contain indices and no ranges. So we could get rid for of the "range" conversion.

axelboc: Yeah it's either a number or `:` on its own, so might as well simplify (at least until we implement subsampling, in which case the logic would be different anyway).

axelboc (approved):
Nice!

PR #1104, opened 2022-05-06
Add a VisGroup (for linear axes only)

t20100:
This PR proposes to have a threejs group that uses the coordinates defined by the axes system.
There is a huge obvious limitation in that it only supports linear axes (I was wondering whether to raise an error or not to display anything in this case).
But when using linear axes, it helps organising multiple meshes in a "data" coordinate system.

What do you think?

Extracted from story in #1096

### DIFF
--- a/apps/storybook/src/VisGroup.stories.tsx
+++ b/apps/storybook/src/VisGroup.stories.tsx
@@ -33,12 +33,12 @@ const Template: Story<VisGroupStoryProps> = (args) => {
           <planeGeometry args={[20, 20]} />
           <meshBasicMaterial color="rgb(255, 0, 0)" />
         </mesh>
-        {points.map((pt) => (
-          <Annotation key={`${pt.x},${pt.y}`} center x={pt.x} y={pt.y}>
-            ({pt.x}, {pt.y})
-          </Annotation>
-        ))}
       </VisGroup>
+      {points.map((pt) => (
+        <Annotation key={`${pt.x},${pt.y}`} center x={pt.x} y={pt.y}>
+          ({pt.x}, {pt.y})
+        </Annotation>
+      ))}
     </VisCanvas>
   );
 };
@@ -52,7 +52,7 @@ Default.args = {
     flip: false,
   },
   ordinateConfig: {
-    visDomain: [0, 40],
+    visDomain: [40, 0],
     isIndexAxis: true,
     showGrid: false,
     flip: false,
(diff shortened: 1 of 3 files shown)


packages/lib/src/vis/shared/VisGroup.tsx
```diff
@@ -0,0 +1,31 @@
+import type { ReactNode } from 'react';
+
+import { useAxisSystemContext } from './AxisSystemProvider';
+
+interface Props {
+  children: ReactNode;
+}
+
+function VisGroup(props: Props) {
```
axelboc, line 9: I've started using `interface Props {}` and then `props: PropsWithChildren<Props>` for cases like this.

packages/lib/src/vis/shared/VisGroup.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { width, height } = visSize;
+
+  const sx =
+    ((abscissaConfig.flip ? -1 : 1) * width) /
+    (abscissaConfig.visDomain[1] - abscissaConfig.visDomain[0]);
+  const sy =
+    ((ordinateConfig.flip ? -1 : 1) * height) /
+    (ordinateConfig.visDomain[1] - ordinateConfig.visDomain[0]);
+  const x = 0.5 * (abscissaConfig.visDomain[0] + abscissaConfig.visDomain[1]);
+  const y = 0.5 * (ordinateConfig.visDomain[0] + ordinateConfig.visDomain[1]);
```
axelboc, line 21: I would either include the full computation of the position coordinates in `x` and `y`, or rename these `centerX` and `centerY`.

axelboc (approved):
Awesome stuff. Yeah, I think it would be wise to check the scales and throw. Something like:

```ts
if (abscissaConfig.scaleType !== ScaleType.Linear || ordinateConfig.scaleType !== ScaleType.Linear) {
  throw new Error('VisGroup supports only linear axes');
}
```

apps/storybook/src/VisGroup.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <ResetZoomButton />
+      <VisGroup>
+        <mesh position={[20, 20, 0]}>
+          <planeGeometry args={[20, 20]} />
+          <meshBasicMaterial color="rgb(255, 0, 0)" />
+        </mesh>
+        {points.map((pt) => (
+          <Annotation key={`${pt.x},${pt.y}`} center x={pt.x} y={pt.y}>
+            ({pt.x}, {pt.y})
+          </Annotation>
```
loichuder, line 39: Enclosing groups do not affect `Annotation`. It can be moved out of `VisGroup`.

PR #1108, opened 2022-05-09
Prevent annotations and other `Html` elements from overflowing canvas

### LINKED ISSUE #1105: Annotation rendering is not clamped inside the plot
The rendering of the annotations can be ouside of the plot.

image

Tested with the actual story book https://h5web-docs.panosc.eu/?path=/story/visualizations-heatmapvis-display--with-annotation

axelboc:
Fix #1105

This was a regression caused by the refactoring of `Html` in #1030. I fixed it by removing the `overflow: visible !important` override on `r3fRoot`. This override was necessary for `AxisSystem`, but I realised that the `Html` element in `AxisSystem` can be now appended to `canvasWrapper` instead of `r3fRoot` thanks to the `container` prop! By doing this, we can let `r3FRoot` hide its overflow as it does by default, which avoids having to turn the `Html` element back into an overlay.

image

### DIFF
--- a/packages/lib/src/vis/shared/AxisSystem.module.css
+++ b/packages/lib/src/vis/shared/AxisSystem.module.css
@@ -1,4 +1,6 @@
 .axisSystem {
+  position: absolute;
+  pointer-events: none;
   display: grid;
   grid-template-areas:
     '. top-axis .'
--- a/packages/lib/src/vis/shared/AxisSystem.tsx
+++ b/packages/lib/src/vis/shared/AxisSystem.tsx
@@ -6,7 +6,7 @@ import { getVisibleDomains } from '../utils';
 import Axis from './Axis';
 import styles from './AxisSystem.module.css';
 import { useAxisSystemContext } from './AxisSystemProvider';
-import Overlay from './Overlay';
+import Html from './Html';
 
 interface Props {
   axisOffsets: AxisOffsets;
@@ -16,6 +16,7 @@ interface Props {
 function AxisSystem(props: Props) {
   const { axisOffsets, title } = props;
 
+  const gl = useThree((state) => state.gl);
   const canvasSize = useThree((state) => state.size);
   const { width, height } = canvasSize;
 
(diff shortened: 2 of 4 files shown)


packages/lib/src/vis/shared/AxisSystem.module.css
```diff
@@ -1,4 +1,6 @@
 .axisSystem {
+  position: absolute;
+  pointer-events: none;
```
axelboc, line 3: `AxisSystem` no longer uses `Overlay`. I could have forwarded the `container` prop but this seemed overkill given that `AxisSystem` already declares most of `Overlay`'s CSS properties (`top`, `width`, etc.)

packages/lib/src/vis/shared/AxisSystem.tsx
```diff
@@ -26,35 +27,38 @@ function AxisSystem(props: Props) {
...
-        type="ordinate"
-        config={ordinateConfig}
-        domain={yVisibleDomain}
-        canvasSize={canvasSize}
-        svgSize={{ width: axisOffsets.left, height }}
-        flipAxis
-      />
-    </Overlay>
+    // Append to `canvasWrapper` instead of default container `r3fRoot`, which hides overflow
+    <Html container={gl.domElement.parentElement?.parentElement || undefined}>
```
axelboc, line 31: This is the key change. We go up one more element in the DOM to bypass the hidden overflow.

packages/lib/src/vis/shared/VisCanvas.module.css
```diff
@@ -5,21 +5,20 @@
...
 
 .canvasWrapper {
+  position: relative; /* for axis system */
   width: 100%; /* fill container by default (unless size is passed in JSX) */
   height: 100%;
   margin: 0 auto;
 }
 
 .r3fRoot {
-  overflow: visible !important; /* show child axis grid, which is bigger than canvas */
```
axelboc, line 14: This re-hides the overflow for other `Html` elements like `Annotation`s

packages/lib/src/vis/shared/VisCanvas.module.css
```diff
@@ -5,21 +5,20 @@
...
 }
 
 .floatingToolbar {
   position: absolute;
-  top: auto !important;
-  left: auto !important;
   right: 0;
   bottom: 0;
   display: flex;
+  z-index: 1; /* above other HTML elements */
```
axelboc, line 23: Quick fix so the floating controls always appear above other `Html` elements.

packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -81,15 +81,14 @@ function VisCanvas(props: PropsWithChildren<Props>) {
...
-                  />
-                </Html>
               </InteractionsProvider>
             </AxisSystemProvider>
           </Canvas>
+
+          <div
+            ref={(elem) => setFloatingToolbar(elem || undefined)}
+            className={styles.floatingToolbar}
+          />
```
axelboc, line 91: Now that `canvasWrapper` has `position: relative`, I can actually move the floating toolbar directly inside it to avoid rendering an `Html` element.

loichuder (approved):
Thanks !

PR #1111, opened 2022-05-11
Add support for complex dataset values in h5wasm

loichuder:
Thanks to [v0.4.2](https://github.com/usnistgov/h5wasm/releases/tag/v0.4.2) of h5wasm

I still have some issues with boolean which is why it is not included in this PR

### DIFF
--- a/packages/h5wasm/src/utils.ts
+++ b/packages/h5wasm/src/utils.ts
@@ -1,14 +1,15 @@
-import type { DType } from '@h5web/shared';
+import type { DType, NumericType } from '@h5web/shared';
 import { DTypeClass, Endianness } from '@h5web/shared';
-import {
-  BrokenSoftLink as H5WasmSoftLink,
-  Dataset as H5WasmDataset,
-  ExternalLink as H5WasmExternalLink,
-  Group as H5WasmGroup,
-} from 'h5wasm';
-import type { Metadata } from 'h5wasm/src/hdf5_util_helpers';
+import type { Dataset as H5WasmDataset, Metadata } from 'h5wasm';
 
-import type { H5WasmEntity } from './models';
+import {
+  assertNumericMetadata,
+  isCompoundMetadata,
+  isFloatMetadata,
+  isIntegerMetadata,
+  isStringMetadata,
+} from './guards';
+import type { NumericMetadata } from './models';
 
 // https://www.loc.gov/preservation/digital/formats/fdd/fdd000229.shtml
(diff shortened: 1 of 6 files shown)


packages/h5wasm/src/utils.ts
```diff
@@ -19,73 +20,66 @@ export function isHDF5(buffer: ArrayBuffer): boolean {
...
-
-export function isH5WasmSoftLink(
-  entity: H5WasmEntity
-): entity is H5WasmSoftLink {
-  return entity instanceof H5WasmSoftLink;
-}
+export function convertMetadataToDType(metadata: NumericMetadata): NumericType;
+export function convertMetadataToDType(metadata: Metadata): DType;
+export function convertMetadataToDType(metadata: Metadata): DType {
+  if (isIntegerMetadata(metadata)) {
```
loichuder, line 26: I reworked the switch to use dedicated guards to allow TS to infer the typing of metadata.

It allows more specific destructurings for Integer, Float and String but the important part is that compound metadata is correctly inferred.

packages/h5wasm/src/utils.ts
```diff
@@ -19,73 +20,66 @@ export function isHDF5(buffer: ArrayBuffer): boolean {
...
-    case 1:
-      return {
-        class: DTypeClass.Float,
-        size: length * 8,
-        endianness: littleEndian ? Endianness.LE : Endianness.BE,
-      };
+    if (nmembers === 2 && members[0].name === 'r' && members[1].name === 'i') {
+      const [realTypeMetadata, imagTypeMetadata] = members;
+      assertNumericMetadata(realTypeMetadata);
+      assertNumericMetadata(imagTypeMetadata);
```
loichuder, line 63: Thanks to this assertion and the overload of `convertMetadataToDType` above, TS knows that converting this will yield a `NumericType` removing the need to assert after conversion.

packages/h5wasm/src/utils.ts
```diff
@@ -19,73 +20,66 @@ export function isHDF5(buffer: ArrayBuffer): boolean {
...
 
-    default:
-      return {
-        class: DTypeClass.Unknown,
-      };
+    return {
+      class: DTypeClass.Compound,
+      fields: Object.fromEntries(
+        members.map((member) => [member.name, convertMetadataToDType(member)])
+      ),
```
loichuder, line 76: Following our line of work in HSDS, we try to be as specific as possible even if it is not complex and that we don't provide visualizations for such types.

axelboc (approved):
Nicely done 👍

PR #1112, opened 2022-05-11
Add support for bool dataset and attributes in h5wasm

loichuder:
Drafting for discussion about the value conversion

**EDIT: Now ready for review. `h5wasm` 0.4.4 brings `to_array` that handles the conversion for us**

### DIFF
--- a/packages/h5wasm/src/h5wasm-api.ts
+++ b/packages/h5wasm/src/h5wasm-api.ts
@@ -1,5 +1,10 @@
 import type { ValuesStoreParams } from '@h5web/app';
-import { getNameFromPath, ProviderApi } from '@h5web/app';
+import {
+  flattenValue,
+  getNameFromPath,
+  ProviderApi,
+  sliceValue,
+} from '@h5web/app';
 import type {
   Attribute,
   AttributeValues,
@@ -9,7 +14,13 @@ import type {
   GroupWithChildren,
   Shape,
 } from '@h5web/shared';
-import { assertNonNull, buildEntityPath, EntityKind } from '@h5web/shared';
+import {
+  assertNonNull,
+  buildEntityPath,
+  DTypeClass,
+  EntityKind,
+  hasBoolType,
+} from '@h5web/shared';
 import type { Attribute as H5WasmAttribute } from 'h5wasm';
 import { File as H5WasmFile, ready as h5wasmReady } from 'h5wasm';
 
@@ -52,6 +63,14 @@ export class H5WasmApi extends ProviderApi {
     const h5wDataset = await this.getH5WasmEntity(dataset.path);
     assertH5WasmDataset(h5wDataset);
 
+    if (hasBoolType(dataset)) {
+      const value = h5wDataset.to_array();
+
+      return selection
+        ? sliceValue(flattenValue(value, dataset), dataset, selection)
+        : value;
+    }
+
     if (selection) {
       const ranges = convertSelectionToRanges(h5wDataset, selection);
       return h5wDataset.slice(ranges);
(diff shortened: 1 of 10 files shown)


packages/h5wasm/src/utils.ts
```diff
@@ -43,6 +46,14 @@ export function convertMetadataToDType(metadata: Metadata): DType {
     };
   }
 
+  throw new Error('Expected numeric metadata');
+}
+
+export function convertMetadataToDType(metadata: Metadata): DType {
```
loichuder, line 52: The addition of an another `if` clause trigged a cognitive complexity warning from sonarjs.

I therefore reworked `convertMetadataToDType` to separate the treatement of numerical dtypes in another `convertNumericMetadataToDType` function. This allows to remove the override signatures of `convertMetadataToDType`.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -60,14 +67,29 @@ export class H5WasmApi extends ProviderApi {
...
+  public async getValue(params: ValuesStoreParams): Promise<unknown> {
+    const value = await this.fetchValue(params);
+
+    const { dataset } = params;
+    if (!hasBoolType(dataset)) {
+      return value;
+    }
+
+    return convertToBool(value);
+  }
```
loichuder, line 79: This is the issue I mentioned in #1111: boolean arrays get returned as `Int8Array`. 

We therefore need to convert both dataset and attribute boolean values. Perhaps this should be done in `h5wasm` instead ?

packages/h5wasm/src/models.ts
```diff
@@ -1,23 +1,34 @@
...
 
 export interface NumericMetadata extends Metadata {
   type: 0 | 1;
 }
+
+export interface EnumMetadata extends Metadata {
+  enum_type: EnumTypeMetadata;
+}
+
+export type TypedArray =
```
loichuder, line 23: Copied from `ndarray` types. I didn't think it was worth adding the dep

packages/h5wasm/src/utils.ts
```diff
@@ -97,3 +125,15 @@ export function convertSelectionToRanges(
...
   });
 }
+
+export function convertToBool(value: unknown): boolean | BoolArray {
+  if (Array.isArray(value)) {
+    return value.map(convertToBool);
+  }
+
+  if (isTypedArrayLodash(value)) {
+    return [...(value as TypedArray)].map(convertToBool);
```
loichuder, line 135: Actually, this poses an issue for attribute display as the `Int8Array` returned are always 1D. even if the underlying dataset has more dimensions. This means nD boolean arrays as attributes will be displayed as 1D arrays.

I didn't find an easy way to fix this.

packages/h5wasm/src/h5wasm-api.ts
axelboc: I can't find the reference to this problem in #1111, but it seems to me that this should indeed be fixed in `h5wasm`.

loichuder: I was talking about my comment:

> I still have some issues with boolean which is why it is not included in this PR

packages/h5wasm/src/utils.ts
bmaranville: Is there an expectation that a library will return something with more dimensions?  Given that JS/TS only has 1d typed arrays, how would this look?

loichuder: It could return regular (not typed) arrays so that multiple dimensions are accounted by nested arrays

bmaranville: There are performance (and other) issues with using nested arrays instead of typed arrays.  I still don't understand the scope of the request - is it preferred in your use case that all dataset and attribute values be converted to nested arrays, or just attribute values, or just "boolean" attribute values?

loichuder: Let's continue the discussion on https://github.com/usnistgov/h5wasm/issues/25

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -1,5 +1,6 @@
 import type { ValuesStoreParams } from '@h5web/app';
 import { getNameFromPath, ProviderApi } from '@h5web/app';
+import { flattenValue, sliceValue } from '@h5web/app/src/providers/utils';
```
axelboc, line 3: import to fix here

axelboc (approved):

PR #1113, opened 2022-05-12
Add prop `overflowCanvas` to `Html` and `Annotation`

axelboc:
This brings:

- a way for `Annotation` to overflow the canvas, following #1105, as requested by @vallsv 
- a more convenient way to move an `Html` component to `canvasWrapper` to show the overflow (than setting the `container` prop manually). The `container` prop can still be set and takes priority over `overflowCanvas`.

### DIFF
--- a/packages/lib/src/vis/shared/Html.tsx
+++ b/packages/lib/src/vis/shared/Html.tsx
@@ -4,13 +4,25 @@ import type { ReactNode } from 'react';
 import ReactDOM from 'react-dom';
 
 interface Props {
+  overflowCanvas?: boolean;
   container?: HTMLElement;
   children?: ReactNode;
 }
 
 function Html(props: Props) {
-  const gl = useThree((state) => state.gl);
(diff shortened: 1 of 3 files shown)


packages/lib/src/vis/shared/Html.tsx
```diff
@@ -4,13 +4,21 @@ import type { ReactNode } from 'react';
...
-  const { container = gl.domElement.parentElement, children } = props;
+  const {
+    overflowCanvas = false,
+    container: customContainer,
+    children,
+  } = props;
+
+  const r3fRoot = useThree((state) => state.gl.domElement.parentElement);
+  const container =
+    customContainer || (overflowCanvas ? r3fRoot?.parentElement : r3fRoot);
```
loichuder, line 21: I think a comment is due here to explain why this change of containers makes the overflow work differently

loichuder (approved):

PR #1116, opened 2022-05-12
Provide point in all coordinate spaces in `useCanvasEvents` callbacks

axelboc:
@vallsv this may be of interest to you to simplify some code in daiquiri (mainly https://gitlab.esrf.fr/ui/daiquiri-ui/-/blob/master/src/components/tomo/tiling/SelectionPoint.tsx). In case you weren't aware of it, we provide (since [v3.0.0](https://github.com/silx-kit/h5web/releases/tag/v3.0.0)) a hook called [`useCanvasEvents`](https://h5web-docs.panosc.eu/?path=/story/utilities--page#usecanvasevents) that can save you from having to call `useEventListener` and unprojecting the points yourself.

With this PR, `useCanvasEvents` also takes care of the conversion from world to data space and exposes a `dataPt` to the callbacks. It also exposes the initial `htmlPt` and the intermediary `cameraPt`, because why not. 🤷 (FYI, I've also renamed `unprojectedPoint` to `worldPt` as I find this naming clearer and more consistent).


axelboc:
@vallsv for your information also, `useCanvasEvents` registers events on the entire canvas. If you want to register events only on a specific part of the visualization, you can go with an R3F `mesh` instead (like in the `VisMesh` used by `TooltipMesh`, as discussed in https://github.com/silx-kit/h5web/discussions/1110). With this solution, you get unprojected/world coordinates directly, so you're left with only having to call `worldToData` in your handlers. Perhaps we could provide a hook called `useMeshEvents` to make this easier. It might be more flexible than the solution I suggested in https://github.com/silx-kit/h5web/discussions/1110#discussioncomment-2729024.

packages/lib/src/interactions/hooks.ts
```diff
@@ -115,66 +115,56 @@ export function useCanvasEvents(callbacks: CanvasEventCallbacks): void {
...
-      const { width, height } = size;
-      const normX = (x - width / 2) / (width / 2);
-      const normY = -(y - height / 2) / (height / 2);
+  const processEvent = useCallback(
+    <T extends PointerEvent | WheelEvent>(sourceEvent: T): CanvasEvent<T> => {
+      const { offsetX, offsetY } = sourceEvent;
 
-      return new Vector3(normX, normY, 0).unproject(camera);
+      const htmlPt = new Vector3(offsetX, offsetY, 0);
+      const cameraPt = htmlPt.clone().applyMatrix4(cameraToHtmlMatrixInverse);
```
axelboc, line 125: This replaces the manual html-to-camera transformation that we were doing previously. `cameraToHtmlMatrixInverse` is provided through context.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -65,15 +65,12 @@ function SelectionTool(props: Props) {
         return;
       }
 
-      const { unprojectedPoint, sourceEvent } = evt;
-      const point = worldToData(boundPointToFOV(unprojectedPoint, camera));
-      setEndPoint(point);
+      const { worldPt, sourceEvent } = evt;
+      const endPoint = worldToData(boundPointToFOV(worldPt, camera));
```
loichuder, line 69: I think it was called `point` to avoid name-collision with the state value `endPoint`

loichuder (approved):
Very nice :clap:

packages/lib/src/interactions/SelectionTool.tsx
axelboc: Oh well spotted, I'll revert the renaming.

axelboc: I've pushed a commit to enable the `no-shadow` ESLint rule. It's disabled in eslint-config-galex because it is considered opinionated. I think it's worth enabling, as I've explained in a comment. It led me to fixing a number of offences.

packages/app/src/vis-packs/core/complex/MappedComplexVis.tsx
```diff
@@ -49,12 +59,16 @@ function MappedComplexVis(props: Props) {
...
+  const { phaseValues, phaseBounds, amplitudeValues, amplitudeBounds } =
+    useMemo(() => getPhaseAmplitudeValues(mappedArray.data), [mappedArray]);
+
+  const phaseArray = useBaseArray(phaseValues, mappedArray.shape);
+  const amplitudeArray = useBaseArray(amplitudeValues, mappedArray.shape);
+
+  const phaseDomain =
+    useValidDomainForScale(phaseBounds, scaleType) || DEFAULT_DOMAIN;
+  const amplitudeDomain =
+    useValidDomainForScale(amplitudeBounds, scaleType) || DEFAULT_DOMAIN;
```
axelboc, line 71: To solve a shadowing issue, I decided to refactor `usePhaseAmplitudeArrays` with some other hooks that we have. In the end, it was short enough to inline in the component. I didn't use `usePhaseAmplitudeValues` based on `createMemo`, because I preferred to pass `mappedArray` instead of `mappedArray.data` in the dependencies array. After that, since `usePhaseAmplitudeValues` was only used in one place and alone in a `hooks` file, I decided to remove it altogether (see `MappedComplexeLineVis` above)

packages/lib/src/vis/shared/AxisSystemProvider.tsx
```diff
@@ -57,9 +59,13 @@ function AxisSystemProvider(props: PropsWithChildren<Props>) {
     const { width, height } = availableSize;
     return new Matrix4()
       .makeScale(width / 2, -height / 2, 1) // scale from normalized camera space to HTML space
-      .setPosition(width / 2, height / 2); // account for shift of (0,0) position (center for camera, top-left for HTML)
+      .setPosition(width / 2, height / 2, 0); // account for shift of (0,0) position (center for camera, top-left for HTML)
```
axelboc, line 62: BTW, this was giving an `undefined` value in the matrix, which led to `NaN` values when inverting the matrix.

eslint.shared.js
```diff
@@ -75,6 +75,10 @@ const overrides = [
           objectLiteralTypeAssertions: 'allow', // `never` is too strict
         },
       ],
+
+      // Disallow shadowing variables for an outter scope, as this can cause bugs
```
loichuder, line 79:
```suggestion
      // Disallow shadowing variables for an outer scope, as this can cause bugs
```

packages/lib/src/vis/shared/AxisSystemProvider.tsx
loichuder: Well spotted :monocle_face:

vallsv:
:+1:

PR #1118, opened 2022-05-13
Upgrade dependencies

axelboc:
First round of "easy", mostly low-risk upgrades:

### With relevant changes

- https://github.com/pmndrs/zustand/releases - from 3.7.0 to 4.0.0-rc.1 as this is marked as latest, and it's worth testing in case we need to provide feedback before stable v4, because they've finally figured out the types!
  - https://github.com/pmndrs/zustand/blob/main/docs/typescript.md#basic-usage
  - https://github.com/pmndrs/zustand/blob/main/docs/v4-migration.md
- https://github.com/react-dropzone/react-dropzone/releases - from 12.0.5 to 14.2.1, compatible with React 18 and small breaking change when specifying allowed extensions

### React 18 compatibility

- https://github.com/airbnb/visx/releases/tag/v2.10.0 - compatible with React 18 :tada:
- https://github.com/dai-shi/react-suspense-fetch/blob/main/CHANGELOG.md#041---2022-05-09 - compatible with React 18 :tada:
- https://github.com/bvaughn/react-window/blob/master/CHANGELOG.md#187 - compatible with React 18 :tada:

### Without relevant changes

- https://github.com/react-hookz/web/releases - from 13.1.0 to 14.2.2
  - [new hooks](https://github.com/react-hookz/web#hooks-list) that we don't need: `useRafEffect`, `useRerender` (we could have used this in `useFrameRendering()`, but since we're going to remove it...), `useLifecycleLogger` (for debugging, though the demo doesn't seem to work), `useCounter` (to inc/dec a `number` state, which we don't seem to need), `useVibration` 
- https://github.com/testing-library/user-event/releases - from 14.0.0 to 14.2.0, I've tried using the new `advanceTimers` option to revert the `delay: null` introduced in #1038, but it's too complicated to be worth it; applying pointer and keyboard actions synchronously by removing the `delay` entirely may not match the behaviour of web browsers 100% but it's fine for our tests
- https://github.com/testing-library/react-testing-library/releases - from 12.1.3 to 12.1.5, v13 requires React 18 so I'm not upgrading to it yet
- https://github.com/d3/d3-array/releases - from 3.1.1 to 3.1.6, nothing relevant to us
- https://github.com/d3/d3-color/releases/tag/v3.1.0 - critical bug fix in colour parsing regexes (I doubt we were affected, though)
- https://github.com/axios/axios/releases/tag/v0.27.2 - nothing relevant
- https://github.com/remix-run/react-router/releases/tag/v6.3.0 - they've added a package with a router that you can use to migrate from v5 to v6, which we've already done easily
- https://github.com/testing-library/jest-dom/releases - from 5.16.2 to 5.16.4, nothing relevant
- https://github.com/testing-library/dom-testing-library/releases - from 8.11.3 to 8.13.0, nothing relevant


loichuder (approved):
Thanks for compiling the list of changes :heart:

PR #1119, opened 2022-05-13
Upgrade React to v18 and R3F to v8

axelboc:
The big change with [React 18](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html), obviously, is the `createRoot` API that replaces `ReactDOM.render`. In terms of code, this affects the demo but most importantly the `Html` component (I think the new API makes the code clearer, actually). In terms of performance, this means we can now opt-in to concurrent rendering features (see below). That being said, there should already be some performance improvements under the hood, especially since R3F already makes use of such features internally.

> **ALL FIXED** There's a bunch of peer dependency warnings that should go away little by little, as usual, notably with ~Storybook~, ~`react-slider`, `react-reflex`~, etc. Talking about `react-reflex`, this library is affected by the breaking change in `@types/react` that consists in the removal of the implicit `children` prop from component types, like `Component`. So for instance, `const ReflexContainer: React.Component<Props, any>` is supposed to now be written as `const ReflexContainer: React.Component<PropsWithChildren<Props>, any>`. ~I've added `@ts-expect-error` comments for now.~ **This is now fixed**.

In [R3F v8](https://docs.pmnd.rs/react-three-fiber/tutorials/v8-migration-guide), the breaking change that affects our codebase is the `sourceEvent` key in event objects being renamed to `nativeEvent`.

Obviously there's a lot more to these major releases that we can benefit from or should watch out for (I will put this list in a discussion thread once merged):

- [`useTransition`](https://reactjs.org/docs/hooks-reference.html#usetransition) to improve the way we use `Suspense` (e.g. by not hiding the current slice while the next one is loading)
- [`useDeferredValue`](https://reactjs.org/docs/hooks-reference.html#usedeferredvalue) to improve performance on a case-by-case basis by deferring heavy, low-priority state updates
- [x] [changes to colour management](https://docs.pmnd.rs/react-three-fiber/tutorials/v8-migration-guide#color-management) in Three v0.139 (~we're currently blocked at v0.135 due to a breaking change~ #1170), thanks to a change in Three 0,139, we no longer have to set the `linear` prop on R3F's `Canvas`. I think we end up with fewer conversions of colours under the hood.
- [updating the R3F camera manually on resize](https://docs.pmnd.rs/react-three-fiber/tutorials/v8-migration-guide#manual-camera-manipulation) to (potentially) fix the pixel resizing issue 🎉 
- [using the `attach` API](https://docs.pmnd.rs/react-three-fiber/tutorials/v8-migration-guide#unified-attach-api) for declaring buffer attributes
- [testing R3F components](https://docs.pmnd.rs/react-three-fiber/tutorials/testing) (the R3F test renderer has been available for a while, but it's worth keeping in mind)
- [`suspend-react`](https://github.com/pmndrs/suspend-react), a lower-level library for `Suspense` than `react-suspense-fetch`; it's developed by the R3F team and used internally in R3F itself


packages/lib/src/vis/shared/Html.tsx
```diff
@@ -27,15 +30,18 @@ function Html(props: Props) {
   const [renderTarget] = useState(() => document.createElement('div'));
 
   useLayoutEffect(() => {
-    ReactDOM.render(<>{children}</>, renderTarget); // eslint-disable-line react/jsx-no-useless-fragment
```
axelboc, line 30: This is now a two step process: first we create the rendering root, then we render the `children` into it. Because `createRoot` does not need the `children` right away, we can move the `render(children)` step into a separate effect in order to remove `children` from the dependency array of this effect. This allows us to bring the clean up function that unmounts the rendering root into this effect.

packages/lib/src/vis/shared/Html.tsx
```diff
@@ -27,15 +30,18 @@ function Html(props: Props) {
...
-  }, [children, renderTarget]);
+    const root = createRoot(renderTarget);
+    setReactRoot(root);
```
axelboc, line 34: Since `root.render(<>children</>)` is now in a separate effect, we need a state to pass the `root` over. This leads to a double render on mount, which is not a big deal. I could probably use a ref to avoid this, since the order of the effects is guaranteed, but it felt like a premature optimisation.

apps/demo/src/index.tsx
```diff
@@ -1,13 +1,16 @@
...
 
+import { assertNonNull } from '@h5web/app';
+import { StrictMode } from 'react';
+import { createRoot } from 'react-dom/client';
+
 import DemoApp from './DemoApp';
 
-ReactDOM.render(
+const rootElem = document.querySelector('#root');
+assertNonNull(rootElem);
```
axelboc, line 10: `rootElem` can be `null`... React recommends doing `createRoot(rootElem!)` but this gives an ESLint warning. I think an assertion is cleaner. I re-exported the one we have in both `lib` and `app` to avoid duplication.

axelboc (commented):
This will need to be tested in a CodeSandbox with React 17 once merged. If it doesn't work, or if we start using concurrent rendering hooks, I may have to bump the `peerDependencies` range.

axelboc:
### TODO

- [x] Fix feature tests (lots of `act` warnings)
- [ ] Fix H5Wasm demo
- [ ] Backwards compatibility with React 17: `Html` element should fallback to legacy API

axelboc:
I've reverted the switch to `createRoot` in `Html`. It seems to work, but the downside is that we'll get lots of warnings in the console on apps running React 18. I guess we can live with it until we've migrated everything, and then we'll drop support for React 17 in a new major version.

axelboc:
Change of plan: as discussed, it might be best to consider maintaining two release branches in case some apps struggle to update to React 18. While we decide and set this up, I'll keep this PR open as draft. I'll, again, remove support for React 17 in this branch and backport all the improvements to the feature tests into the `main` branch.

bmaranville:
> @bmaranville I'm getting some warnings in the H5Wasm demo when I select a simple HDF5 file.
> 
> image
> 
> After the warnings, I get a `name not defined` error followed by a white screen. The error seems to be coming from https://github.com/usnistgov/h5wasm/blob/main/src/hdf5_util.cc
> 
> Not sure how this is linked to upgrading to React 18... Any clue?

I will try to build the react 18 branch and see what is causing the error.

bmaranville:
It seems to be calling file.close in the api.cleanUp method when the file is already closed... I suspect it is because of a change in how useEffect is handled in react 18 (it is apparently called twice sometimes) - see https://www.techiediaries.com/react-18-useeffect/

bmaranville:
I can verify that the h5wasm loader works for a pre-built version of the demo, with react 18.  So it does seem to be related to the change in useeffect in react 18, when running in development mode.

axelboc:
Oh it must be `<Strict>` mode!

axelboc:
- https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-strict-mode
- https://github.com/reactwg/react-18/discussions/18

Not quite sure how to work around it to still get a proper clean up. I'll investigate. Thanks!

axelboc:
All good, by putting the `api` instance in a state, strict mode doesn't call the effect twice, so doesn't call the clean-up. I've also confirmed that changing the props of `H5WasmProvider` after a timeout works.

axelboc:
Attn @PeterC-DLS: I've just rebased this branch on v7. From now on, I'll try to rebase it only after stable releases.

I'm just doing a basic sanity check locally with `pnpm start` before force-pushing, and then the CI runs as usual—I'm not testing this branch in production projects yet—so do let me know if you notice issues.

axelboc:
I've rebased and fixed a few things, but I had to skip a couple of tests for the CI to pass. I'll see if upgrading Testing Library helps and investigate a bit more.

The plan is to merge this and release a new major asap.

packages/app/src/test-utils.tsx
```diff
@@ -71,21 +76,27 @@ export async function renderApp(
   };
 }
 
-async function findVisSelector(): Promise<HTMLElement> {
-  return screen.findByRole('tablist', { name: 'Visualization' });
+export async function waitForAllLoaders(): Promise<void> {
+  await waitFor(() => {
+    expect(screen.queryAllByTestId(/^Loading/u)).toHaveLength(0);
```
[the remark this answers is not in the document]
axelboc: Forgot to mention that the challenge here was that suspense fallbacks can "resolve" in different orders, so I couldn't wait for a specific loader to disappear _and then_ another one — I have to wait for all loaders to disappear at once.

packages/app/src/__tests__/Explorer.test.tsx
```diff
@@ -92,13 +84,21 @@ test('show spinner when group metadata is slow to fetch', async () => {
...
+  expect(screen.getByLabelText(/Loading group/)).toBeVisible();
+
+  // Wait for fetch of group metadata to succeed
+  await expect(
+    screen.findByText(/No visualization available/, undefined, {
+      timeout: SLOW_TIMEOUT,
+    }),
+  ).resolves.toBeVisible();
+
+  // Spinner has been removed
```
loichuder, line 102:
```suggestion
  // Check that spinner has been removed
```

loichuder (approved):
Looks all right to me. 

Congrats for soldiering through the timer issues. It should not have been easy.

packages/app/src/dimension-mapper/SlicingSlider.tsx
```diff
@@ -53,6 +53,15 @@ function SlicingSlider(props: Props) {
             setValue(newValue);
             onDebouncedChange(newValue);
           }}
+          /* When slicing in E2E tests, `onChange` is called with the old value.
+             A `setState` callback in `ReactSlider` that is supposed to be called
+             after the state is updated, is in fact called before.
+             https://github.com/zillow/react-slider/blob/master/src/components/ReactSlider/ReactSlider.jsx#L890
```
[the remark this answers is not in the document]
axelboc: I'm trying to upgrade `react-slider` and this workaround no longer works, so I'll have to. I came across this issue which sounds awfully similar: https://github.com/zillow/react-slider/issues/304. I might make a CodeSandbox and re-open the issue after I investigate a bit more what's going on.

PR #1123, opened 2022-05-23
Add util getAxisValues in h5web/lib

### LINKED ISSUE #1073: `getAxisValues`-like utility for line vis
In daiquiri (https://gitlab.esrf.fr/ui/daiquiri-ui/-/merge_requests/377#note_184352), we could use a utility and memoised hook that does what we currently do inline inside `LineVis`:

```tsx
const abscissas = useMemo(() => {
  if (!abscissaValue) {
    return range(dataArray.size);
  }
  
  return toArray(abscissaValue);
}, [abscissaValue, dataArray.size]);
```

We already have a utility called `getAxisValues` (and its memoised hook `useAxisValues`) dedicated to the Heatmap (although it is not exported in `@h5web/lib`), but we don't have one for the line vis (without all the extra `pixelCount` business). The heatmap's `getAxisValues` utility could in fact call the line's `getAxisValues` internally. Naming would have to be improved, obviously.

loichuder:
Fix #1073

### DIFF
--- a/apps/storybook/src/Utilities.stories.mdx
+++ b/apps/storybook/src/Utilities.stories.mdx
@@ -142,6 +142,22 @@ scaleWithMid(2.2); // Returns 1
 scaleWithMid(3); // Returns 3
 ```
 
+### getAxisValues
+
+Return an array of axis values of the desired length.
+
+- If an array is supplied, convert it to a plain JS array if needed and check that the lengths match.
+- If no array is supplied, generate a sequence of numbers from 0 to the desired length.
+
+```ts
+getAxisValues(rawValues: NumArray | undefined, axisLength: number): number[]
+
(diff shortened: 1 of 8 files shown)


axelboc (approved):
Very nice! Just need to add the new util/hook to the `Utilities` doc page.

apps/storybook/src/Utilities.stories.mdx
```diff
@@ -142,6 +142,24 @@ scaleWithMid(2.2); // Returns 1
 scaleWithMid(3); // Returns 3
 ```
 
+### getAxisValues
+
+Return a array of numbers with the desired axis length.
```
axelboc, line 147:
```suggestion
Return an array of axis values of the desired length.
```

axelboc (approved):

PR #1124, opened 2022-05-25
Move mandlebrot to web worker in tiled heatmap story

axelboc:
I've found a library called [greenlet](https://github.com/developit/greenlet), which works beautifully. It creates an inline worker with the function you give it: https://github.com/developit/greenlet and wraps it in a promise API. It's super tiny too.

The demo is a lot more fluid, though still not amazing on my machine, as we're still doing all the computations on the CPU. It can't compare exactly to a fetch-based scenario, but it's already a lot better.

### DIFF
--- a/apps/storybook/src/TiledHeatmapMesh.stories.tsx
+++ b/apps/storybook/src/TiledHeatmapMesh.stories.tsx
@@ -17,6 +17,7 @@ import type {
 } from '@h5web/lib';
 import { ScaleType } from '@h5web/shared';
 import type { Meta, Story } from '@storybook/react/types-6-0';
+import greenlet from 'greenlet';
 import { clamp } from 'lodash';
 import ndarray from 'ndarray';
 import type { NdArray } from 'ndarray';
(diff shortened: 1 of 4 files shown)


apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -27,46 +28,49 @@ import type { Vector2 } from 'three';
...
+    xDomain: Domain,
+    yDomain: Domain,
+    size: Size
+  ): Promise<Float32Array> => {
+    const { width, height } = size;
+    const xRange = xDomain[1] - xDomain[0];
+    const yRange = yDomain[1] - yDomain[0];
 
-  const array = ndarray(new Float32Array(height * width), [height, width]);
+    const array = new Float32Array(height * width);
```
axelboc, line 42: I can't use `ndarray` inside the worker because of the way `greenlet` works. It's limited to functions that don't rely at all on external scope.

loichuder (approved):
Much smoother on my end

PR #1127, opened 2022-06-08
Make E2E tests more robust

axelboc:
After working on the feature tests in #1119, it's clear that React Testing Library is much better than Cypress Testing Library at waiting for React to finish rendering. That's why we have a lot of workarounds in the E2E specs, waiting for unrelated things to appear, and why we get a lot of retry attempts in the CI.

In this PR, I'm introducing a query called `cy.waitForStableDom()` found here: https://github.com/narinluangrath/cypress-wait-for-stable-dom/, which seems to perfectly solve all our problems. It tells Cypress to wait until there are no more mutations in the DOM over a given period of time. By calling it after every user action (`click`, `type`, etc.), all the tests pass on their first attempts.

The polling interval is currently set to 300ms, a bit more than the slicing slider, just to be safe for the screenshots. In #1119, I'm introducing a way to detect when a slice selection is actually applied to the visualization, so once it is merged, I will try to refactor to reduce the delay.

image

### DIFF
--- a/cypress/specs/app.spec.ts
+++ b/cypress/specs/app.spec.ts
@@ -1,62 +1,54 @@
-const SNAPSHOT_DELAY = 500;
+// For ARIA-based checks
+const BE_SELECTED = ['have.attr', 'aria-selected', 'true'] as const;
+const BE_CHECKED = ['have.attr', 'aria-checked', 'true'] as const;
 
-describe('App', () => {
+afterEach(() => {
+  window.localStorage.clear();
+});
+
+describe('/mock', () => {
   beforeEach(() => {
     cy.visit('/mock');
-  });
-
-  afterEach(() => {
-    window.localStorage.clear();
+    cy.waitForStableDOM();
   });
 
   it('visualize 1D dataset as Line', () => {
-    cy.findByRole('treeitem', { name: 'nD_datasets' }).click();
-    cy.findByRole('treeitem', { name: 'oneD' }).click();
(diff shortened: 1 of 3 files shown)


cypress/specs/app.spec.ts
```diff
@@ -1,119 +1,103 @@
...
-  afterEach(() => {
-    window.localStorage.clear();
+    cy.waitForStableDOM();
   });
 
   it('visualize 1D dataset as Line', () => {
-    cy.findByRole('treeitem', { name: 'nD_datasets' }).click();
-    cy.findByRole('treeitem', { name: 'oneD' }).click();
+    cy.selectExplorerNode('nD_datasets');
+    cy.selectExplorerNode('oneD');
```
axelboc, line 17: I've also added a couple of custom queries like in feature tests. Internally, they wait for the DOM to be stable.

cypress/specs/app.spec.ts
```diff
@@ -127,263 +111,250 @@ describe('App', () => {
...
   });
 });
+
+describe('/mock?wide', () => {
+  beforeEach(() => {
+    cy.visit('/mock?wide');
+    cy.waitForStableDOM();
+  });
+
+  it('start with the explorer closed', () => {
```
axelboc, line 353: Separating this test in its own `describe` block, as previously, the `visit` inside the test was being displayed as an attempt in the Cypress UI, which was weird.

Issue #1121, opened 2022-05-19
[SelectToZoom] Allow to specify a mouse button


vallsv:
Hi,

I want to implement silx interaction in the plot.

It sounds like the `SelectToZoom` interaction is working for any mouse button.

It would be good to limit the interaction to a specific button.

And as it's low cost, this used button could be selected by property.

As result i will be able to create a `SelectToZoom` interaction with a left mouse button only.

This start to be an important feature for me, as the plot is about to be integrated and used in production.
https://gitlab.esrf.fr/ui/daiquiri-tomo/-/issues/115

axelboc:
A better, more flexible approach would be to capture right clicks before they reach the `SelectToZoom` handlers -- more specifically `pointerdown` events before the reach the event handler registered on the canvas by `SelectionTool`.

To do so, you can try to render a component inside `VisCanvas` and before `SelectToZoom`. In this component, register an `onPointerDown` event handler on the canvas with `useCanvasEvents`. Inside this handler, if the event is triggered by the right mouse, call [`evt.stopImmediatePropagation()`](https://devdocs.io/dom/event/stopimmediatepropagation).

---

If the need arises to draw a selection or select-to-zoom with the right mouse button, then we will consider making the mouse button configurable in `SelectionTool` and `SelectToZoom`. Though, I don't see this ever happening, as those are two highly uncommon interactions.

axelboc:
```tsx
<VisCanvas>
  <CaptureRightMouseButton />
  <SelectToZoom />
</VisCanvas>
```

```tsx
function CaptureRightMouseButton() {
  const onPointerDown = useCallback((evt: CanvasEvent<PointerEvent>) => {
    if (evt.button === 2) {
      evt.stopImmediatePropagation();
    }
  });

  useCanvasEvents({ onPointerDown });
  return null;
}
```

axelboc:
Sorry, the title of the issue confused me. You're right that drawing a selection with the right mouse button (or middle) should not be possible. We should definitely correct this. Though, I don't believe we should make the button configurable, as explained.

loichuder:
With #1129, `SelectToZoom` now only works with the left button. 

As @axelboc argued, this is not configurable.

Issue #1120, opened 2022-05-19
[Pan] Allow to specify a mouse button


vallsv:
Hi,

I want to implement silx interaction in the plot.

It sounds like the `Pan` interaction is working for any mouse button.

It would be good to limit the interaction to a specific button.

And as it's low cost this used button could be selected by property.

As result i will be able to create a `Pan` interaction with a middle mouse button only.

This start to be an important feature for me, as the plot is about to be integrated and used in production.
https://gitlab.esrf.fr/ui/daiquiri-tomo/-/issues/115

axelboc:
As per https://github.com/silx-kit/h5web/issues/1121#issuecomment-1135987700, you're right that it should definitely not be possible to pan with the right mouse button.

When it comes to the middle button, I think I agree with you that this needs to be configurable via a prop so that you can have two `Pan` components in your canvas: one bound to the left button (and enabled conditionally based on the tool selected by the user) and the other to the middle button. I think a simple boolean prop, for instance called `withMiddleBtn`, could do the trick.

One small implementation challenge comes from having to take into account mouse buttons in our `InteractionsProvider`'s `shouldInteract` function.

vallsv:
While i can do such change of the behaviour by description, it's fine to me. Anyway it's a prop, or a new component.

loichuder:
With #1129, panning with the right mouse button no longer works and a prop `button` can be passed to `Pan` to set it to `Left` (default) or `Middle`

PR #1129, opened 2022-06-09
Allow to constrain interactions to a single button

loichuder:
Looking at #1120 and #1121, I thought we could leverage our interaction registration process to be able to constrain interactions to a single button.

I tried to keep the changes minimal to give a basis of discussion so that when `button` is not given, the behaviour is the same as before (no constrain on button). Among future improvements, I can see:
- Give default values for `button` for the `Pan`
- Take `button` in account when looking at conflicting interactions in `shouldInteract`

### DIFF
--- a/packages/lib/src/interactions/Pan.tsx
+++ b/packages/lib/src/interactions/Pan.tsx
@@ -8,13 +8,22 @@ import {
   useMoveCameraTo,
   useInteraction,
 } from './hooks';
+import { MouseButton } from './models';
 import type { CanvasEvent, Interaction } from './models';
 
-function Pan(props: Interaction) {
-  const { modifierKey } = props;
+interface Props extends Omit<Interaction, 'button'> {
+  button?: MouseButton;
+}
+
+function Pan(props: Props) {
+  const { button = MouseButton.Left, modifierKey, disabled } = props;
 
   const camera = useThree((state) => state.camera);
-  const shouldInteract = useInteraction('Pan', props);
+  const shouldInteract = useInteraction('Pan', {
+    button,
+    modifierKey,
+    disabled,
+  });
   const moveCameraTo = useMoveCameraTo();
 
   const startOffsetPosition = useRef<Vector3>(); // `useRef` to avoid re-renders
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -11,9 +11,10 @@ import {
   useInteraction,
 } from './hooks';
 import type { CanvasEvent, Interaction, Selection } from './models';
+import { MouseButton } from './models';
 import { boundPointToFOV } from './utils';
 
-interface Props extends Interaction {
+interface Props extends Omit<Interaction, 'button'> {
   onSelectionStart?: () => void;
   onSelectionChange?: (points: Selection) => void;
   onSelectionEnd?: (points: Selection) => void;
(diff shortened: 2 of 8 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -31,11 +32,12 @@ function SelectionTool(props: Props) {
     id = 'Selection',
     modifierKey,
     disabled,
+    button = Button.Left,
```
loichuder, line 35: I give `Left` as the default value for `SelectionTool` meaning that it is the same for `SelectToZoom`.

packages/lib/src/interactions/utils.ts
```diff
@@ -50,11 +51,11 @@ export function getDefaultInteractions(
   keepRatio?: boolean
 ): Record<string, Interaction> {
   return {
-    Pan: {},
+    Pan: { button: Button.Left },
     Zoom: {},
     XAxisZoom: { modifierKey: 'Alt', disabled: keepRatio },
     YAxisZoom: { modifierKey: 'Shift', disabled: keepRatio },
-    SelectToZoom: { modifierKey: 'Control' },
+    SelectToZoom: { modifierKey: 'Control', button: Button.Left },
```
loichuder, line 58: If we give a default value to `Pan`, this change can be reverted

axelboc (changes requested):
I'm conflicted. As I mentioned in both issues, I don't think it should be possible to pan with the right mouse button and to select-to-zoom with the middle or right button. It just wouldn't be great for UX in a web app...

The only case of right-click drag I can remember ever encountering is on Windows: you can select then right-click drag a bunch of files/folders to open a context menu:

image

I've never seen an image editor or the like allow right-click drag to pan or draw a selection. This would be even weirder on the web. Right click is reserved for context menus, that's what users are used to.

Middle-click pan is fine, I think I've seen that before in image editors, but not middle-click selection.

So the choice comes down to either:
1. making the mouse button freely customisable for both pan and select and let consumers of H5Web components be responsible for UX, or 
2. being more restrictive to guarantee to our users that some good UX practices will be respected when they use H5Web components.

Given how I phrased the two options above, you can tell which way I'm leaning. 😄 

So my recommendation would be:

1. to pass `Mouse.Left` to `useInteraction` in `SelectToZoom`
2. to add the `button` prop only to the `Pan` component and retrict its value to `MouseButton.Left | MouseButton.Middle` (this way, you can keep the three values in the enum in case we need it in the future).

packages/lib/src/interactions/models.ts
```diff
@@ -2,6 +2,12 @@ import type { Vector2, Vector3 } from 'three';
 
 export type ModifierKey = 'Alt' | 'Control' | 'Shift';
 
+export enum Button {
```
axelboc, line 5: Maybe rename this to `MouseButton`?

packages/lib/src/interactions/Pan.tsx
```diff
@@ -8,13 +8,22 @@ import {
   useMoveCameraTo,
   useInteraction,
 } from './hooks';
+import { MouseButton } from './models';
 import type { CanvasEvent, Interaction } from './models';
 
-function Pan(props: Interaction) {
-  const { modifierKey } = props;
+interface Props extends Omit<Interaction, 'button'> {
+  button?: MouseButton;
```
axelboc, line 15: Since you've removed `Right` from `MouseButton`, the `Omit` is not needed here. Either re-add `Right` to the enum and change the prop here to `button?: MouseButton.Left | MouseButton.Right`, or leave `props: Interaction` as is.

packages/lib/src/interactions/Pan.tsx
```diff
@@ -8,13 +8,22 @@ import {
...
+function Pan(props: Props) {
+  const { button = MouseButton.Left, modifierKey, disabled } = props;
 
   const camera = useThree((state) => state.camera);
-  const shouldInteract = useInteraction('Pan', props);
+  const shouldInteract = useInteraction('Pan', {
+    button,
+    modifierKey,
+    disabled,
+  });
```
axelboc, line 26: You can just pass `props`, I think?

axelboc (approved):

packages/lib/src/interactions/Pan.tsx
loichuder: This is to have the default value of `MouseButton.Left` (set above)

PR #1131, opened 2022-06-10
Make H5Wasm demo more permissive

### LINKED ISSUE #1126: Allow loading of files with non-standard extensions in h5wasm demo

This is just a minor issue - all of our NeXus files are written with a file name matching this format: ```long_filename0001.nxs.<instrument-specific-extension>``` where the instrument-specific extension is something like ```cgd``` or ```vsans```
The way the dropzone loader is configure in the h5wasm demo, none of our files will load without changing the extension and re-saving the file.

Reconfigure dropzone configuration so that it will still allow loading when the extension doesn't match one of the items in the list:
```ts
const EXT = ['.h5', '.hdf5', '.hdf', '.nx', '.nx5', '.nexus', '.nxs', '.cxi'];
```
It is probably still preferred that those extensions are used to populate the search in the file loader dialog, but then disabling the strict check in react-dropzone that **forbids** loading files with other extensions would be useful.

### Additional context

I'd be happy to help out with this.

axelboc:
- Allow files with any extensions to be dropped - fix #1126 
- Allow files that don't start with the HDF5 magic number (called the HDF5 File Signature in the spec), as per the HDF5 spec https://web.ics.purdue.edu/~aai/HDF5/html/H5.format.html

### DIFF
--- a/apps/demo/src/h5wasm/DropZone.tsx
+++ b/apps/demo/src/h5wasm/DropZone.tsx
@@ -31,7 +31,6 @@ function DropZone(props: Props) {
 
   const { getRootProps, getInputProps, open, isDragActive, fileRejections } =
     useDropzone({
-      accept: { 'application/x-hdf5': EXT },
       multiple: false,
       noClick: true,
       noKeyboard: true,
@@ -46,7 +45,7 @@ function DropZone(props: Props) {
         'data-disabled': isReadingFile || undefined,
       })}
     >
-      <input {...getInputProps()} />
+      <input {...getInputProps()} accept={EXT.join(',')} />
       <div className={styles.inner}>
         {isDragActive ? (
           <p className={styles.dropIt}>Drop it!</p>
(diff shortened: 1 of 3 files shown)


apps/demo/src/h5wasm/DropZone.tsx
```diff
@@ -58,12 +57,11 @@ function DropZone(props: Props) {
...
             </p>
-            <p
-              className={styles.hint}
-              role={fileRejections.length > 0 ? 'alert' : undefined}
-            >
-              Extensions allowed: <strong>{EXT.join(' ')}</strong>
-            </p>
+            {fileRejections.length > 1 && (
+              <p className={styles.hint} role="alert">
+                Please drop a single file
```
axelboc, line 62: I've removed the hint, and I'm now showing an error only when multiple files are dropped.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -118,7 +109,10 @@ export class H5WasmApi extends ProviderApi {
     const file = await this.file;
 
     const h5wEntity = file.get(path);
-    assertNonNull(h5wEntity, `No entity found at ${path}`);
+    assertNonNull(
+      h5wEntity,
+      path === '/' ? `Expected valid HDF5 file` : `No entity found at ${path}`
```
axelboc, line 114: If I drop a JSON file, for instance, `file.get('/')` logs a warning and returns `null`, so I'm generalising here that when we get `null` for the root entity, it's because h5wasm had a problem reading the file (most likely because it's not an HDF5 file).

loichuder (approved):
Note that the magic number is also called `File Signature` in the format spec on HDFGroup's website: https://portal.hdfgroup.org/display/HDF5/File+Format+Specification

PR #1132, opened 2022-06-10
Fix rolled-up types

### LINKED ISSUE #1099: Enums raise error in @h5web/app declaration types
### Describe the bug

The built declaration types of `@h5web/app` give the following error at the first encountered `enum`:
> Top-level declarations in .d.ts files must start with either a 'declare' or 'export' modifier.ts(1046)

### To Reproduce

1. Go in `packages/app`
2. Run `pnpm build:dts`
3. Open `dist/index.d.ts`
4. See error `ts(1046)`

### Context

Encountered when trying to build `jupyterlab-h5web` with `h5web@4.1.0`

### Additional info

- The issue was already raised in `rollup-plugin-dts` but closed: https://github.com/Swatinem/rollup-plugin-dts/issues/178
- Apparently `enum` should not be in `d.ts` files ? https://stackoverflow.com/questions/62109542/enums-in-typescript-d-ts-file

axelboc:
Fix #1099 

Here is the trick: I build the shared types with `tsc` to `packages/shared/dist-ts`, and then in the other packages, when the time comes to merge all the types into a single file with Rollup, I tell Rollup that `@h5web/shared` is an alias for `../shared/dist-ts/index.d.ts`! :tada: 

In `dist/index.d.ts`, the enums are now properly prefixed with `declare`, the `mockMetadata` object is no longer inlined, etc.

image

This can only be good for package consumers, even when no errors were reported. If anything it will save the TypeScript engine some work in VSCode.

Also, `pnpm packages` used to output warnings like this one:

image

Now the output is clean:

image

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -13,6 +13,7 @@
   "sideEffects": false,
   "main": "src/index.ts",
   "scripts": {
+    "build": "run-p build:*",
     "build:dts": "tsc --build tsconfig.build.json",
     "lint:eslint": "eslint \"**/*.{js,cjs,ts,tsx}\" --max-warnings=0",
     "lint:tsc": "tsc --noEmit"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -144,7 +144,10 @@ The build process of `@h5web/lib` works as follows:
      consumers who use TypeScript. This is a two step process: first we generate
      type declarations for all TS files in the `dist-ts` folder with `tsc`, then
      we use Rollup to merge all the declarations into a single file:
-     `dist/index.d.ts`, which is referenced from `package.json`.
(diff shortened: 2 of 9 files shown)


packages/shared/package.json
```diff
@@ -13,6 +13,7 @@
   "sideEffects": false,
   "main": "src/index.ts",
   "scripts": {
+    "build": "run-p build:*",
```
axelboc, line 16: `pnpm packages` was not building the shared types, since Rollup was not configured to look for them until now.

Note that since `@h5web/shared` is a dependency of the other packages, pnpm knows to call its `build` script first, so the other packages' builds are guaranteed to find the shared built types under `packages/shared/dist-ts`.

loichuder (approved):
Great, thanks

PR #1134, opened 2022-06-13
Make select zoom take the full width/height if rect points are too close

### LINKED ISSUE #1066: [SelectionZoom] Make the zoom area take the full height (width) if the drawn rectangle has no height (width)
Plotly allows to drag a rectangle to zoom with the mouse (as h5web but notice that the data outside fades to grey)
image

A very interesting feature that plotly has and h5web has not is the following: if the vertical location of the mouse is very close to the start (rectangle with no height), the zoom rectangle takes the full height (similar to axial zoom in h5web)  
image

This would be good to reuse, particularly in the case of curve plots.

loichuder:
Fix #1066

Peek 2022-06-13 10-37

Must be enabled though a `fullDimIfEmpty` prop

I refactored `SelectToZoom` so that the render props and the `onSelectionEnd` use the same function to transform points selected by the user.


packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -63,24 +62,30 @@ function SelectToZoom(props: Props) {
...
               stroke="black"
-              fillOpacity={keepRatio ? 0 : 0.25}
-              strokeDasharray={keepRatio ? '4' : undefined}
+              fillOpacity={0.25}
+              clampToVis
             />
+            {/* Draw un-transformed user-drawn rectangle */}
             {keepRatio && (
-              <RatioSelectionRect
+              <SelectionRect
```
loichuder, line 83: The diff can be a bit hard to process here.

1. The logic was inverted so that the actual zoom rectangle is always displayed. The `keepRatio` conditional is used to display the user-drawn rectangle instead. This allows to simplify the `fillOpacity` and `strokeDasharray` props.
1. The `RatioSelectionRect` was replaced by a regular `SelectionRect` that takes the transformed points with the addition of a prop `clampToVis` to keep the behaviour that was only in `RatioSelectionRect`. If `keepRatio` is `false`, the vis takes the full dims so it does not hurt to keep `clampToVis` to `true`.

packages/lib/src/interactions/SelectionRect.tsx
```diff
@@ -17,17 +19,29 @@ function SelectionRect(props: Props) {
...
-    [startPoint, endPoint]
+    (camera, context) => {
+      const { visSize, dataToWorld, worldToData } = context;
+      const [newStartPoint, newEndPoint] = clampToVis
+        ? clampRectangleToVis(
+            dataToWorld(startPoint),
+            dataToWorld(endPoint),
+            visSize
+          )
+        : [dataToWorld(startPoint), dataToWorld(endPoint)];
```
loichuder, line 37: A bit fiddly here as we need the points in world coordinates to clamp to the vis :shrug:

packages/lib/src/interactions/utils.ts
```diff
@@ -20,8 +21,14 @@ export function boundPointToFOV(
 export function getRatioRespectingRectangle(
   startPoint: Vector2,
   endPoint: Vector2,
-  ratio: number
-) {
+  xVisibleDomain: Domain,
+  yVisibleDomain: Domain
```
loichuder, line 25: The signature was changed to match the signature of the new `getDistinctPoints` and make the functions interchangeable.

packages/lib/src/interactions/utils.ts
```diff
@@ -99,3 +106,52 @@ export function clampRectangleToVis(
...
+  yDomain: Domain
+): [Vector2, Vector2] {
+  const [xMin, xMax] = xDomain;
+
+  if (Math.abs(endPoint.x - startPoint.x) < Math.abs(xMax - xMin) / 200) {
+    return [new Vector2(xMin, startPoint.y), new Vector2(xMax, endPoint.y)];
+  }
+
+  const [yMin, yMax] = yDomain;
+  if (Math.abs(endPoint.y - startPoint.y) < Math.abs(yMax - yMin) / 200) {
```
loichuder, line 123: The criteria for an "empty" rectangle is to have the width less than 1/200 of the axis coordinates. 

I tweaked this a bit and found that this value gave the best UX in my opinion.

packages/lib/src/interactions/utils.ts
```diff
@@ -99,3 +106,52 @@ export function clampRectangleToVis(
...
+      endPt: Vector2,
+      xDomain: Domain,
+      yDomain: Domain
+    ) =>
+      getRatioRespectingRectangle(
+        ...getDistinctPoints(startPt, endPt, xDomain, yDomain),
+        xDomain,
+        yDomain
+      );
+  }
```
loichuder, line 146: I combine both transform if needed but this is a bit over the top. There is not much point having both at the same time. Indeed, if the user draws an empty rectangle in this case and asks for the full width, the only way to keep the ratio when having the full width is to have the full height which makes the zoom meaningless.

packages/lib/src/interactions/SelectionRect.tsx
axelboc: What about an early return for `!clampToVis` to avoid the `dataToWorld`/`worldToData` conversions?

packages/lib/src/interactions/utils.ts
axelboc: Yeah I agree, I don't think it's needed.

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -3,19 +3,21 @@ import { useThree } from '@react-three/fiber';
...
 import SelectionRect from './SelectionRect';
 import SelectionTool from './SelectionTool';
 import { useMoveCameraTo } from './hooks';
 import type { Interaction, Selection } from './models';
-import { getEnclosedRectangle, getRatioRespectingRectangle } from './utils';
+import { getEnclosedRectangle, getZoomTransform } from './utils';
 
 interface Props extends Omit<Interaction, 'button'> {
   keepRatio?: boolean;
+  fullDimIfEmpty?: boolean;
```
axelboc, line 14: Not a huge fan of the name - what about `withAxialZoom` or `fillIfEmpty`? Also, it doesn't make sense for both props to be `true`, so perhaps we should add a `throw` when this happens (as doing an XOR of two optional boolean props in TS is not trivial).

packages/lib/src/interactions/utils.ts
axelboc: Wouldn't this become much simpler if `SelectionTool`'s callbacks provided the selection in HTML space in addition to data space, like `useCanvasEvents`? `getDistinctPoints` could then define a threshold in screen pixels.

Maybe `SelectionTool` could also provide the selection in world space to simplify other computations?

packages/lib/src/interactions/utils.ts
```diff
@@ -99,3 +106,52 @@ export function clampRectangleToVis(
...
+    return [new Vector2(startPoint.x, yMin), new Vector2(endPoint.x, yMax)];
+  }
+
+  return [startPoint, endPoint];
+}
+
+export function getZoomTransform(
+  keepRatio?: boolean,
+  fullDimIfEmpty?: boolean
+): ZoomTransformFn {
```
axelboc, line 133: Not sure I really get the point of returning a function here. If it's the number of parameters, you can combine them: `dataSelection`, `visiblesDomains`. I would also rename the final function `getZoomSelection` or `getZoomArea`.

axelboc (changes requested):
Nice feature :tada: 

I find the implementation and the refactoring a bit convoluted. I've suggested a few changes. I think it stems in part from the fact that we provide the selections only in data space from the `SelectionTool` callbacks.

loichuder:
Superseded by #1145

PR #1135, opened 2022-06-13
Include world points in Selection

loichuder:
To ease #1134 

See https://github.com/silx-kit/h5web/pull/1134#discussion_r895532239

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -35,7 +35,7 @@ function SelectionTool(props: Props) {
   } = props;
 
   const camera = useThree((state) => state.camera);
-  const { worldToData } = useAxisSystemContext();
+  const { dataToWorld, worldToData } = useAxisSystemContext();
   const shouldInteract = useInteraction(id, {
     button: MouseButton.Left,
     modifierKey,
@@ -71,11 +71,19 @@ function SelectionTool(props: Props) {
       }
 
       const { worldPt, sourceEvent } = evt;
-      const newEndPoint = worldToData(boundPointToFOV(worldPt, camera));
-      setEndPoint(newEndPoint);
+      const dataEndPoint = worldToData(boundPointToFOV(worldPt, camera));
+      setEndPoint(dataEndPoint);
+
+      const worldEndPoint = dataToWorld(dataEndPoint);
+      const worldStartPoint = dataToWorld(startPoint);
 
       if (onSelectionChange && shouldInteract(sourceEvent)) {
-        onSelectionChange({ startPoint, endPoint: newEndPoint });
+        onSelectionChange({
+          startPoint,
+          endPoint: dataEndPoint,
+          worldStartPoint,
+          worldEndPoint,
+        });
       }
     },
     [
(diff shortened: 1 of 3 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -71,18 +72,37 @@ function SelectionTool(props: Props) {
...
 
       const { worldPt, sourceEvent } = evt;
-      const newEndPoint = worldToData(boundPointToFOV(worldPt, camera));
-      setEndPoint(newEndPoint);
+      const dataEndPoint = worldToData(boundPointToFOV(worldPt, camera));
+      setEndPoint(dataEndPoint);
+
+      const worldEndPoint = dataToWorld(dataEndPoint);
+      const htmlEndPoint = cameraToHtml(
+        new Vector3(worldEndPoint.x, worldEndPoint.y, 0).project(camera)
```
axelboc, line 80: A small `worldToCamera` utility could be nice, maybe? You could use it in `dataToHtml` as well.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -71,18 +72,37 @@ function SelectionTool(props: Props) {
...
+      );
+
+      const worldStartPoint = dataToWorld(startPoint);
+      const htmlStartPoint = cameraToHtml(
+        new Vector3(worldStartPoint.x, worldStartPoint.y, 0).project(camera)
+      );
```
axelboc, line 86: On one hand I was thinking that you could have a function `processSelectionEvent` to compute all this like in `useCanvasEvents`, to avoid duplication. On the other hand, I noticed that the `worldStartPoint` and `htmlStartPoint` are recomputed unnecessarily, since they are already available in the initial `pointerDown` event. But then storing the initial `pointerDown` event in the state doesn't quite feel right. If you have any idea... Otherwise it's no big deal. :+1:

axelboc (approved):

packages/lib/src/interactions/SelectionTool.tsx
loichuder: I went a step further by adding `worldToHtml` to the `AxisSystemContext`

packages/lib/src/interactions/SelectionTool.tsx
loichuder: We could compute `worldStartPoint` and `htmlStartPoint` with `useMemo`s from the state value `startPoint` outside of the handler ? :shrug:

axelboc: Yeah or even without `useMemo`, just above the event handlers so you can re-use the same variables.

packages/lib/src/interactions/SelectionTool.tsx
axelboc: Hmm, isn't that what we had done at some point and decided to move it out? Now, this sort of "camera-based" functions are in `src/vis/utils.ts` and take the `camera` as first parameter.

- We have `dataToWorld` and `cameraToHtml` in the context, because they don't actually use `camera`.
- The only missing piece is `worldToCamera`, which does use `camera` so should go in `src/vis/utils.ts`.
- We also have a higher-order utility `dataToHtml` also in `src/vis/utils.ts`.
- Now I don't have any objection adding `worldToHtml` in there as well if you think it makes sense.

PR #1136, opened 2022-06-13
Export and document `getAxisDomain`, `useCameraState` and camera-based utilities

axelboc:
Also, document the `Annotation` component in its own stories.

### DIFF
--- a/apps/storybook/src/HeatmapVisDisplay.stories.tsx
+++ b/apps/storybook/src/HeatmapVisDisplay.stories.tsx
@@ -98,101 +98,21 @@ WithAxisLabels.args = {
 
 export const WithAnnotation: Story<HeatmapVisProps> = (args) => (
   <HeatmapVis {...args}>
-    <Annotation x={10} y={15} style={{ color: 'white' }}>
-      HTML annotation positioned at (10, 15)
-    </Annotation>
-    <Annotation
-      x={10}
-      y={5}
-      center
-      style={{
-        width: 180,
-        color: 'white',
-        textAlign: 'center',
-      }}
-    >
-      Another annotation, <strong>centred</strong> on (10, 5)
+    <Annotation x={10} y={16} style={{ color: 'white' }}>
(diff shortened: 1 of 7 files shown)


apps/storybook/src/HeatmapVisDisplay.stories.tsx
```diff
@@ -98,65 +98,16 @@ WithAxisLabels.args = {
 
 export const WithAnnotation: Story<HeatmapVisProps> = (args) => (
```
axelboc, line 99: I still keep one, very simple annotation story for the Heatmap and another for the Line vis.

apps/storybook/src/Html.stories.tsx
```diff
@@ -46,7 +46,7 @@ export const Container: Story<{
   const [portalContainer, setPortalContainer] = useState<HTMLDivElement>();
 
   return (
-    <>
+    <div style={{ display: 'flex' }}>
```
axelboc, line 49: The canvas was progressively growing to fill the height on first render because of the empty `div` below (which was taking up half of the grid on first render and then shrinking).

apps/storybook/src/Utilities.stories.mdx
```diff
@@ -183,6 +200,70 @@ const handlePointerDown = useCallback((evt: CanvasEvent<PointerEvent>) => {
...
+```ts
+const htmlPt = useCameraState(
+  (camera, context) => dataToHtml(camera, context, new Vector2(x, y)),
+  [x, y]
+);
+```
+
+Similarly to `useCallback`, `useCameraState` accepts a dependency array as second parameter. Make sure to configure ESLint's
+[`react-hooks/exhaustive-deps` rule](https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks#advanced-configuration)
+in your app to report missing dependencies:
```
axelboc, line 222: We should do this in Braggy, by the way.

loichuder (approved):
Great call to add some tests :+1:

Issue #1130, opened 2022-06-09
[Matrix] Support compound datasets with 1D fields


loichuder:
### Is your feature request related to a problem?

Compound datasets can only visualized by `Raw` vis which does not allow to explore the data

### Requested solution or feature

Compound datasets with 1D fields could be displayed by the Matrix vis (column: field, row: indices)

PR #1137, opened 2022-06-15
Improve compound dtype parsing in H5GroveProvider

loichuder:
Following https://github.com/silx-kit/h5grove/pull/69 and early step for #1130 

Previously, compound dtypes were interpreted by `H5GroveProvider` as `Unknown`. Now, there are rightly shown as `Compound` and the raw inspection can show the fields

image

### DIFF
--- a/packages/app/src/providers/h5grove/h5grove-api.ts
+++ b/packages/app/src/providers/h5grove/h5grove-api.ts
@@ -13,7 +13,7 @@ import { isString } from 'lodash';
 
 import { DataProviderApi } from '../api';
 import type { ExportFormat, ValuesStoreParams } from '../models';
-import { convertDtype, handleAxiosError } from '../utils';
+import { handleAxiosError } from '../utils';
 import type {
   H5GroveAttribute,
   H5GroveAttrValuesResponse,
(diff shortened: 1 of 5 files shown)


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -26,10 +26,11 @@ import {
   isGroupResponse,
   isSoftLinkResponse,
   typedArrayFromDType,
+  convertH5GroveDtype,
 } from './utils';
 
 export class H5GroveApi extends DataProviderApi {
-  /* API compatible with h5grove@1.0.0 */
+  /* API compatible with h5grove@1.1.0b0 */
```
loichuder, line 33: I will make a proper release before merging

packages/app/src/providers/h5grove/utils.ts
```diff
@@ -80,3 +92,87 @@ export function typedArrayFromDType(dtype: DType) {
...
+
+  return {
+    class: DTypeClass.Compound,
+    fields: Object.fromEntries(
+      Object.entries(dtype).map(([k, v]) => [k, convertH5GroveDtype(v)])
+    ),
+  };
+}
+
+function convertDtypeString(dtype: string): DType {
```
loichuder, line 109: This is the old `convertDtype` function that is unchanged

axelboc (approved):
Awesome

PR #1144, opened 2022-06-20
Allow to specify several modifier keys for interaction

loichuder:
This enhances the flexibility of interactions: one could specify `Ctrl`+`Alt` for the zoom selection for example


packages/lib/src/interactions/models.ts
```diff
@@ -36,7 +36,7 @@ export interface InteractionInfo {
 
 export interface Interaction {
   button?: MouseButton;
-  modifierKey?: ModifierKey;
+  modifierKeys?: ModifierKey[];
```
loichuder, line 39: The change in essence: `Interaction` can now specify multiple `modifierKeys`

The rest of the PR is adaptation to this change

packages/lib/src/interactions/hooks.ts
```diff
@@ -189,22 +185,23 @@ export function useInteraction(id: string, value: Interaction) {
...
-
-  return isPressed;
+  const map = useMap<ModifierKey, boolean>(modifierKeys.map((k) => [k, true]));
+
+  useEventListener(window, 'keyup', (event: KeyboardEvent) => {
+    const { key: pressedKey } = event;
+    if (isModifierKey(pressedKey) && map.has(pressedKey)) {
+      map.set(pressedKey, false);
+    }
+  });
```
loichuder, line 197: My first thought was to use a loop to call `useKeyboardEvent` for each `modifierKey` but this breaks the rule of hooks.

I therefore had to reimplement this with a `useEventListener`. Perhaps I should attach it to the `Canvas` rather than the `window` ?

apps/storybook/src/Zoom.stories.tsx
```diff
@@ -10,13 +10,15 @@ interface TemplateProps {
 }
 
 const Template: Story<TemplateProps> = (args) => {
+  const { disabled, modifierKey } = args;
```
loichuder, line 13: For convenience, I choose the keep the stories unchanged (only one `ModifierKey` can be selected with the radio button).

But the API of all interactions components (`Pan`, `Zoom`, `SelectToZoom` ...) changed to accept several modifier keys. This can be up to discussion.

packages/lib/src/interactions/InteractionsProvider.tsx
```diff
@@ -36,29 +37,63 @@ function InteractionsProvider(props: { children: ReactNode }) {
...
+        ([id, interaction]) => ({
+          id,
+          ...interaction,
+        })
+      );
+
+      function isButtonPressed(button?: MouseButton) {
+        if (event instanceof WheelEvent) {
+          return button === undefined;
+        }
```
loichuder, line 51: Due to the change of algorithm below and because `WheelEvent` have a `button` equal to 0, we need a way to distinguish `WheelEvent` from `MouseEvent`. 

Otherwise, we have multiple matching interactions (with one for wheel and the other for click) that our algorithm cannot distinguish.

packages/lib/src/interactions/InteractionsProvider.tsx
```diff
@@ -36,29 +37,63 @@ function InteractionsProvider(props: { children: ReactNode }) {
...
-    (id: string, event: MouseEvent) => {
-      const registeredKeys = [...interactionMap.values()]
-        .map((params: Interaction) => params.modifierKey)
-        .filter(isDefined);
+    (interactionId: string, event: MouseEvent | WheelEvent) => {
+      const registeredInteractions = [...interactionMap.entries()].map(
+        ([id, interaction]) => ({
+          id,
+          ...interaction,
+        })
```
axelboc, line 45: Perhaps this is the data structure we should use for the `interactionMap` state? It'd be better to loop through the array when registering/unregistering interactions rather than on every interaction event.

packages/lib/src/interactions/InteractionsProvider.tsx
```diff
@@ -36,29 +37,63 @@ function InteractionsProvider(props: { children: ReactNode }) {
...
+        (acc, { id, modifierKeys: keys }) => {
+          if (keys.length > acc.numberOfKeys) {
+            return { id, numberOfKeys: keys.length };
+          }
+          return acc;
+        },
+        { id: '', numberOfKeys: -Infinity }
+      );
+
+      return maxKeyInteraction.id === interactionId;
```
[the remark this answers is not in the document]
axelboc: Makes sense!

I feel like there has to be a simpler solution than `reduce`. Maybe a `sort` with a custom comparator? :shrug: Might be just as ugly :sweat_smile: 

> If no modifierKey, verify if no other attributed modifier key is pressed (which is essentially checking if other interaction are matching)

For the record, we had refactored this behaviour to check, for real, that no other interaction was matching: `registeredKeys.every((key) => !event.getModifierState(key));`

axelboc (changes requested):
It's more of a "Suggest changes", but if you do, I'd like to take another look.

apps/storybook/src/Zoom.stories.tsx
loichuder: Good idea :+1:

packages/lib/src/interactions/hooks.ts
```diff
@@ -58,7 +54,7 @@ function useWheelCapture() {
     evt.preventDefault();
   }, []);
 
-  // Handler must be registed as non-passive for `preventDefault` to have an effect
+  // Handler must be registed as non-passive for `preventDefault` to have an e ffect
```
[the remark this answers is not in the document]
loichuder: Yeah, prettier was not behaving well at some point during my work on this :rofl:

packages/lib/src/interactions/hooks.ts
```diff
@@ -189,22 +185,23 @@ export function useInteraction(id: string, value: Interaction) {
...
+  });
+
+  useEventListener(window, 'keydown', (event: KeyboardEvent) => {
+    const { key: pressedKey } = event;
+    if (isModifierKey(pressedKey) && map.has(pressedKey)) {
+      map.set(pressedKey, true);
+    }
+  });
+
+  return modifierKeys ? [...map.values()].every(Boolean) : true;
```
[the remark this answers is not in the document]
loichuder: `.every((v) => v)` gets replaced to `.every(Boolean)` when formatting :shrug:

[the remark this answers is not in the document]
axelboc: ah must be the auto-fixer of an ESLint rule, then :+1:

packages/lib/src/interactions/InteractionsProvider.tsx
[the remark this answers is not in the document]
axelboc: Yeah, much more concise. Good call not copying the array, don't think it's worth it :+1: 

I don't think you need the `reverse` if you just invert the comparator:

```suggestion
sort((a, b) => b.modifierKeys.length - a.modifierKeys.length)
```

packages/lib/src/interactions/utils.ts
```diff
@@ -99,3 +99,21 @@ export function clampRectangleToVis(
...
+export function getModifierKeyArray(
+  keys: ModifierKey | ModifierKey[] | undefined
+): ModifierKey[] {
+  if (!keys) {
+    return [];
+  }
+
+  if (isModifierKey(keys)) {
+    return [keys];
+  }
```
axelboc, line 116: Not sure about this -- might be cleaner to just use `Array.isArray(keys)` and leave `isModifierKey` as it was before (with a `key: string` argument, if I recall).

```ts
export function getModifierKeyArray(
  keys: ModifierKey | ModifierKey[] | undefined = []
): ModifierKey[] {
  if (Array.isArray(keys)) {
    return keys;
  }

  return [keys];
}
```

Which can be considerably shortened with a ternary, which makes me think that it might not be worth extracting into a function:

```ts
const modifierKeys = Array.isArray(modifierKey) ? modifierKey : [modifierKey];
```

packages/lib/src/interactions/InteractionsProvider.tsx
loichuder: > when event is a WheelEvent, can button ever be defined?

Hm, yes in fact when checking for matching interactions in a `WheelEvent`. With `event instanceof WheelEvent || event.button === button`, it means that every interaction is matching for any `WheelEvent`.

I think we have to treat the `Wheel` case more explicitely by having a dedicated value. I tried to add `'Wheel'` instead of `undefined` for `WheelEvent` (so that `button` is always defined for any interaction).

axelboc: Great call! So can `button` ever be `undefined` now?

loichuder: No ! It is always defined in `InteractionEntry`. Props of interaction components such as `Pan` allow `button` to be `undefined` but provide a fallback

axelboc: Oh so you can remove the `?` in the function signature. Sorry, should have been clearer.

PR #1145, opened 2022-06-21
Implement SelectToZoom for a single axis

### LINKED ISSUE #1066: [SelectionZoom] Make the zoom area take the full height (width) if the drawn rectangle has no height (width)

loichuder:
Fix #1066

Peek 2022-06-21 16-50

This PR implements a specific component `AxialSelectToZoom` for drawing a selection to zoom on a single axis. Triggering this zoom mode can be done with  a specific combination of keys (thanks to #1144) rather than relying on the size of the rectangle (which can be fiddly).

### DIFF
--- a/packages/lib/src/interactions/AxialSelectToZoom.tsx
+++ b/packages/lib/src/interactions/AxialSelectToZoom.tsx
@@ -1,0 +1,92 @@
+import { useThree } from '@react-three/fiber';
+import { Vector2 } from 'three';
+
+import { useCameraState } from '../vis/hooks';
+import { useAxisSystemContext } from '../vis/shared/AxisSystemProvider';
+import { getVisibleDomains } from '../vis/utils';
+import SelectionRect from './SelectionRect';
+import SelectionTool from './SelectionTool';
+import { useMoveCameraTo } from './hooks';
+import type { ModifierKey, Selection } from './models';
+import { getEnclosedRectangle } from './utils';
+
+interface Props {
+  axis: 'x' | 'y';
+  disabled?: boolean;
+  modifierKey?: ModifierKey[] | ModifierKey;
+}
+
+function AxialSelectToZoom(props: Props) {
+  const { axis, modifierKey, disabled } = props;
+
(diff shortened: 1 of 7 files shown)


packages/lib/src/interactions/AxialSelectToZoom.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    // Work in world coordinates as we need to act on the world camera
+    const { worldStartPoint, worldEndPoint } = getAxialSelection(
+      startPoint,
+      endPoint
+    );
+
+    if (
+      worldStartPoint.x === worldEndPoint.x &&
+      worldStartPoint.y === worldEndPoint.y
```
axelboc, line 65: Do we want to check only the coordinate that corresponds the `axis` prop here?

axelboc (approved):
Very nice :clap:

axelboc:
I reckon you should add the interaction to `DefaultInteractions` and to the help menu, while you're at it.

packages/lib/src/interactions/AxialSelectToZoom.tsx
loichuder: It would make sense but I feel this makes the condition check more complex for the same result

axelboc: Well, if you draw a rectangle that is 100px tall (i.e. the height of the vis) but 0px wide and the `axis` is `x`, doesn't this mean that the code will try to divide by 0? Maybe replacing `&&` with `||` would do the trick?

loichuder: Oh, good thinking. Well, I can replace the `&&` by `||` and call it a day :smile:

PR #1146, opened 2022-06-24
Allow to attach a handler for clicks on the DataCurve

loichuder:
For https://gitlab.esrf.fr/ui/daiquiri-ui/-/merge_requests/405#note_195040

### DIFF
--- a/apps/storybook/src/DataCurve.stories.tsx
+++ b/apps/storybook/src/DataCurve.stories.tsx
@@ -3,6 +3,7 @@ import { CurveType, DataCurve, useDomain, VisCanvas } from '@h5web/lib';
 import { assertDefined, mockValues } from '@h5web/shared';
 import type { Meta, Story } from '@storybook/react/types-6-0';
 import { range } from 'lodash';
+import { useState } from 'react';
 
 import FillHeight from './decorators/FillHeight';
 
@@ -52,6 +53,35 @@ WithErrors.args = {
   errors: mockValues.oneD_errors,
   showErrors: true,
   color: 'blue',
+};
+
+export const Click: Story<DataCurveProps> = (args) => {
+  const [index, setIndex] = useState<number>();
+  const { abscissas, ordinates } = args;
+
+  const abscissaDomain = useDomain(abscissas);
+  const ordinateDomain = useDomain(ordinates);
+  assertDefined(abscissaDomain);
+  assertDefined(ordinateDomain);
+
+  return (
+    <VisCanvas
+      abscissaConfig={{ visDomain: abscissaDomain, showGrid: true }}
+      ordinateConfig={{ visDomain: ordinateDomain, showGrid: true }}
+      title={
+        index !== undefined
+          ? `You clicked on segment ${index} going from (${abscissas[index]}, ${
+              ordinates[index]
+            }) to (${abscissas[index + 1]}, ${ordinates[index + 1]})`
+          : 'Click on the curve!'
+      }
+    >
+      <DataCurve {...args} onLineClick={(i) => setIndex(i)} />
+    </VisCanvas>
+  );
+};
+Click.args = {
+  ...Default.args,
 };
 
 export default {
--- a/packages/lib/src/vis/hooks.ts
+++ b/packages/lib/src/vis/hooks.ts
@@ -104,3 +104,38 @@ export function useCustomColors(
 
   return [colors, refCallback];
 }
+
+export function useRaycasterThreshold(
+  lineOrPoints: 'Line' | 'Points',
+  desiredThreshold: number
+) {
+  const raycaster = useThree((state) => state.raycaster);
+  const camera = useThree((state) => state.camera);
+
+  useEffect(() => {
+    const params = raycaster.params[lineOrPoints];
+    const oldThreshold = params ? params.threshold : 1;
+
+    if (params) {
+      params.threshold = desiredThreshold;
+    }
+
+    return () => {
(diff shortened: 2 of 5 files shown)


packages/lib/src/vis/hooks.ts
```diff
@@ -104,3 +104,36 @@ export function useCustomColors(
...
+  const raycaster = useThree((state) => state.raycaster);
+  const zoomVector = useThree((state) => state.camera.scale);
+
+  useEffect(() => {
+    const params = raycaster.params[lineOrPoints];
+    const oldThreshold = params ? params.threshold : 1;
+
+    if (params) {
+      params.threshold = desiredThreshold;
+    }
```
loichuder, line 121: Annoyingly, 
```
if(raycaster.params[lineOrPoints]) {
  raycaster.params[lineOrPoints].threshold
}
```
raises a TS error as TS cannot infer that `raycaster.params[lineOrPoints]` is defined in the `if` branch so I have to resort to an intermediate variable

packages/lib/src/vis/hooks.ts
```diff
@@ -104,3 +104,36 @@ export function useCustomColors(
 
   return [colors, refCallback];
 }
+
+export function useRaycasterThreshold(
+  lineOrPoints: 'Line' | 'Points',
+  desiredThreshold: number
```
loichuder, line 110: I encapsulated the work on raycaster threshold made for the interaction on Scatter points (https://github.com/silx-kit/h5web/pull/1051 and https://github.com/silx-kit/h5web/pull/1064) to reuse it for the `DataCurve`.

packages/lib/src/vis/line/DataCurve.tsx
```diff
@@ -29,6 +32,7 @@ interface Props {
   color: string;
   curveType?: CurveType;
   visible?: boolean;
+  onLineClick?: (index: number, evt: ThreeEvent<MouseEvent>) => void;
```
loichuder, line 35: I do not propagate the handler prop to `LineVis` as the only usecase we have deals with `DataCurve`

packages/lib/src/vis/hooks.ts
```diff
@@ -104,3 +104,36 @@ export function useCustomColors(
...
+    return () => {
+      if (params) {
+        params.threshold = oldThreshold;
+      }
+    };
+  }, [raycaster, lineOrPoints, desiredThreshold]);
+
+  useFrame(() => {
+    // Reduce the threshold dynamically when zooming as the size of Line and Points objects is reduced in world units
+    const zoom = zoomVector.x;
```
axelboc, line 132: Even though this is fine, since the `zoomVector` is mutated by Three, I would prefer if you used `camera.scale` here and if your retrieved the full `camera` object from `useThree`. This is more consistent with what we do in other places.

packages/lib/src/vis/hooks.ts
```diff
@@ -104,3 +104,36 @@ export function useCustomColors(
...
+
+    return () => {
+      if (params) {
+        params.threshold = oldThreshold;
+      }
+    };
+  }, [raycaster, lineOrPoints, desiredThreshold]);
+
+  useFrame(() => {
+    // Reduce the threshold dynamically when zooming as the size of Line and Points objects is reduced in world units
```
axelboc, line 131: I feel like this comment could be a bit clearer. The lines and points keep the same size in world space -- they only change in camera space and HTML space, if I'm not mistaken.

packages/lib/src/vis/line/DataCurve.tsx
```diff
@@ -52,13 +57,26 @@ function DataCurve(props: Props) {
...
+  );
+
   const showLine = visible && curveType !== CurveType.GlyphsOnly;
   const showGlyphs = visible && curveType !== CurveType.LineOnly;
 
   return (
     <>
-      <line_ visible={showLine} geometry={dataGeometry}>
-        <lineBasicMaterial color={color} linewidth={2} />
+      <line_ visible={showLine} geometry={dataGeometry} onClick={handleClick}>
```
axelboc, line 78: Just so the handler does not get registered if there's no callback:

```suggestion
      <line_ visible={showLine} geometry={dataGeometry} onClick={onLineClick && handleClick}>
```

apps/storybook/src/DataCurve.stories.tsx
```diff
@@ -54,6 +55,42 @@ WithErrors.args = {
   color: 'blue',
 };
 
+export const Click: Story<DataCurveProps> = (args) => {
+  const [title, setTitle] = useState('Click on the curve !');
```
axelboc, line 59:
```suggestion
  const [title, setTitle] = useState('Click on the curve!');
```

apps/storybook/src/DataCurve.stories.tsx
```diff
@@ -54,6 +55,42 @@ WithErrors.args = {
...
+        {...args}
+        onLineClick={(index) => {
+          setTitle(
+            index !== undefined
+              ? `You clicked on segment ${index} going from (${
+                  abscissas[index]
+                }, ${ordinates[index]}) to (${abscissas[index + 1]}, ${
+                  ordinates[index + 1]
+                })`
+              : 'Click on the curve !'
```
axelboc, line 83: I would maybe store the index in the state and compute the title in the `VisCanvas` prop above. This will avoid the duplication of the fallback title.

axelboc (approved):
Nice! Does it play well with the pan and other interactions?

PR #1147, opened 2022-06-24
Allow viewing remote files in H5Wasm demo

axelboc:
- I've added a form to the H5Wasm demo to allow specifying a URL
- When the form is submitted, the app navigates to `?url=<url>`, which fetches the file as an `ArrayBuffer`
- While fetching, the form is disabled and a spinner is shown
- Once fetched, the buffer is passed to `H5WasmProvider` and the viewer shows up
- The use of the `url` param allows refreshing (cf. GIF below), bookmarking and linking (e.g. from the H5Web homepage)

h5wasm-url

### DIFF
--- a/apps/demo/src/h5wasm/DropZone.tsx
+++ b/apps/demo/src/h5wasm/DropZone.tsx
@@ -2,16 +2,18 @@ import { useCallback, useState } from 'react';
 import { useDropzone } from 'react-dropzone';
 
 import styles from './H5WasmApp.module.css';
+import UrlForm from './UrlForm';
 import type { H5File } from './models';
 
 const EXT = ['.h5', '.hdf5', '.hdf', '.nx', '.nx5', '.nexus', '.nxs', '.cxi'];
 
 interface Props {
-  onChange: (h5File: H5File) => void;
+  onH5File: (h5File: H5File) => void;
 }
 
 function DropZone(props: Props) {
-  const { onChange } = props;
+  const { onH5File } = props;
+
   const [isReadingFile, setReadingFile] = useState(false);
 
   const onDropAccepted = useCallback(
@@ -20,13 +22,13 @@ function DropZone(props: Props) {
       reader.addEventListener('abort', () => setReadingFile(false));
       reader.addEventListener('error', () => setReadingFile(false));
       reader.addEventListener('load', () => {
-        onChange({ filename: file.name, buffer: reader.result as ArrayBuffer });
+        onH5File({ filename: file.name, buffer: reader.result as ArrayBuffer });
       });
 
       reader.readAsArrayBuffer(file);
       setReadingFile(true);
     },
-    [onChange]
+    [onH5File]
   );
 
   const { getRootProps, getInputProps, open, isDragActive, fileRejections } =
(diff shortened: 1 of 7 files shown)


apps/demo/src/h5wasm/DropZone.tsx
```diff
@@ -47,23 +49,24 @@ function DropZone(props: Props) {
...
-          <p className={styles.dropIt}>Drop it!</p>
-        ) : (
-          <>
-            <p className={styles.content}>
-              Drop an HDF5 file here{' '}
-              <button className={styles.btn} type="button" onClick={open}>
-                or use a file picker instead
-              </button>
+        {isDragActive && <p className={styles.dropIt}>Drop it!</p>}
+        <div hidden={isDragActive}>
```
axelboc, line 53: I hide the form instead of unmounting/remounting it to avoid re-triggering requests in `UrlForm`.

apps/demo/src/h5wasm/H5WasmApp.module.css
```diff
@@ -1,8 +1,10 @@
 .dropZone {
-  height: 100%;
+  display: flex;
+  min-height: 100%;
```
axelboc, line 3: `min-height` to show the scrollbar on short screens instead of cropping

apps/demo/src/h5wasm/UrlForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+    void fetchFile();
+  }, [url, fetchFile]);
+
+  function handleUrlSubmit(evt: FormEvent<HTMLFormElement>) {
+    evt.preventDefault();
+    const formData = new FormData(evt.currentTarget);
+    const newUrl = formData.get('url') as string;
+
+    if (newUrl === url) {
+      void fetchFile(); // refetch
```
axelboc, line 44: When an error occurs and the users submits the form again, navigating to the same URL would not re-invoke the `useEffect`, so I refetch manually instead.

apps/demo/src/h5wasm/UrlForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <input
+            className={styles.urlInput}
+            name="url"
+            type="url"
+            defaultValue={url}
+            placeholder="https://example.com/my-file.h5"
+            aria-labelledby="h5wasm-go"
+            required
+          />
+          <button id="h5wasm-go" className={styles.goBtn} type="submit">
```
loichuder, line 66: I would have replaced `go` by `url` to be clearer but it is fine like this too.

```suggestion
          <button id="h5wasm-url" className={styles.urlBtn} type="submit">
```

loichuder (approved):
Nice !

Minor complaint: it feels a bit odd to land on the h5wasm page briefly during the file loading when clicking on the links in the homepage. We could imagine having a spinner taking the whole page on loading but it is nitpicking.

axelboc:
Yeah, I hesitated. It would have been easy to do with suspense, but it's more in case there's a network error (which gets displayed under the URL input field) and to give an indication of which URL is being loaded -- but maybe it's not worth it :shrug: I'll merge as is and we'll reassess.

loichuder:
Yeah no worries. It is great as it is :+1:

apps/demo/src/h5wasm/UrlForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <input
+            className={styles.urlInput}
+            name="url"
+            type="url"
+            defaultValue={url}
+            placeholder="https://example.com/my-file.h5"
+            aria-labelledby="h5wasm-go"
+            required
+          />
+          <button id="h5wasm-url" className={styles.urlBtn} type="submit">
```
loichuder, line 66: Since you followed my suggestion, I think you should also change `goBtn` to `urlBtn` in the CSS and `h5wasm-go` to `h5wasm-url` in the `aria-labelledby` field above.

PR #1148, opened 2022-06-29
Enforce stricter types for default interactions configuration

axelboc:
The main goal of this PR is to make sure that the `interactions` props of the visualizations have a type that is compatible with the props of the interaction components. For instance, previously, you could pass `{ zoom: { button: MouseButton.Middle } }` even though `Zoom` doesn't accept a `button` prop.

To do this, I refactored a few things:

- `DefaultInteractions` is now written more declaratively and is more strongly typed. It accepts interaction configurations directly as props instead of inside an `interactions` prop.
- Every interaction component has its own `Props` interface that is exported and used to type the props of `DefaultInteractions`, which are in turn used to type the `interactions` props of the visualization components.
- I renamed the `Interaction` interface to `CommonInteractionProps` to make it clear that it is meant to be used to type interaction component props. The interface no longer includes the `button` prop, since it is only needed by the `Pan` component. This removes a lot of `Omit`.

I took the opportunity to document the `DefaultInteractions` component now that it feels more robust in terms of typing. Let me know what you think.

### DIFF
--- a/apps/storybook/src/DefaultInteractions.stories.tsx
+++ b/apps/storybook/src/DefaultInteractions.stories.tsx
@@ -1,0 +1,36 @@
+import { DefaultInteractions, ResetZoomButton, VisCanvas } from '@h5web/lib';
+import type { DefaultInteractionsProps } from '@h5web/lib/src/interactions/DefaultInteractions';
+import type { Meta, Story } from '@storybook/react';
+
+import FillHeight from './decorators/FillHeight';
+
+const Template: Story<DefaultInteractionsProps> = (args) => {
+  return (
+    <VisCanvas
+      abscissaConfig={{ visDomain: [-10, 0], showGrid: true }}
+      ordinateConfig={{ visDomain: [50, 100], showGrid: true }}
+    >
+      <DefaultInteractions {...args} />
+      <ResetZoomButton />
+    </VisCanvas>
+  );
+};
+
+export const Default = Template.bind({});
+
+export default {
+  title: 'Building Blocks/DefaultInteractions',
+  component: DefaultInteractions,
+  decorators: [FillHeight],
+  parameters: { layout: 'fullscreen' },
+  argTypes: {
+    keepRatio: { defaultValue: false },
+    pan: { defaultValue: {} },
+    zoom: { defaultValue: {} },
+    xAxisZoom: { defaultValue: { modifierKey: 'Alt' } },
+    yAxisZoom: { defaultValue: { modifierKey: 'Shift' } },
+    selectToZoom: { defaultValue: { modifierKey: 'Control' } },
+    xSelectToZoom: { defaultValue: { modifierKey: ['Control', 'Alt'] } },
+    ySelectToZoom: { defaultValue: { modifierKey: ['Control', 'Shift'] } },
+  },
+} as Meta<DefaultInteractionsProps>;
--- a/packages/lib/src/interactions/AxialSelectToZoom.tsx
+++ b/packages/lib/src/interactions/AxialSelectToZoom.tsx
@@ -7,13 +7,11 @@ import { getVisibleDomains } from '../vis/utils';
 import SelectionRect from './SelectionRect';
 import SelectionTool from './SelectionTool';
 import { useMoveCameraTo } from './hooks';
-import type { ModifierKey, Selection } from './models';
+import type { Selection, CommonInteractionProps } from './models';
 import { getEnclosedRectangle } from './utils';
 
-interface Props {
+interface Props extends CommonInteractionProps {
   axis: 'x' | 'y';
-  disabled?: boolean;
-  modifierKey?: ModifierKey[] | ModifierKey;
 }
 
 function AxialSelectToZoom(props: Props) {
@@ -73,7 +71,7 @@ function AxialSelectToZoom(props: Props) {
   return (
     <SelectionTool
       onSelectionEnd={onSelectionEnd}
-      id={`AxialSelectToZoom${axis}`}
+      id={`${axis.toUpperCase()}SelectToZoom`}
       modifierKey={modifierKey}
       disabled={disabled}
     >
@@ -89,4 +87,5 @@ function AxialSelectToZoom(props: Props) {
   );
 }
 
+export type { Props as AxialSelectToZoomProps };
 export default AxialSelectToZoom;
--- a/packages/lib/src/interactions/DefaultInteractions.tsx
+++ b/packages/lib/src/interactions/DefaultInteractions.tsx
@@ -1,58 +1,79 @@
-import { merge } from 'lodash';
-
 import Pan from '../interactions/Pan';
+import type { PanProps } from '../interactions/Pan';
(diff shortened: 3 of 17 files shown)


apps/storybook/src/DefaultInteractions.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+  argTypes: {
+    keepRatio: { defaultValue: false },
+    pan: { defaultValue: {} },
+    zoom: { defaultValue: {} },
+    xAxisZoom: { defaultValue: { modifierKey: 'Alt' } },
+    yAxisZoom: { defaultValue: { modifierKey: 'Shift' } },
+    selectToZoom: { defaultValue: { modifierKey: 'Control' } },
+    xSelectToZoom: { defaultValue: { modifierKey: ['Control', 'Alt'] } },
+    ySelectToZoom: { defaultValue: { modifierKey: ['Control', 'Shift'] } },
+    withAxialZoom: { defaultValue: true },
```
axelboc, line 35: Providing default values for the controls. Annoying that Storybook isn't able to access default prop values 😂

packages/lib/src/interactions/AxialSelectToZoom.tsx
```diff
@@ -73,7 +71,7 @@ function AxialSelectToZoom(props: Props) {
   return (
     <SelectionTool
       onSelectionEnd={onSelectionEnd}
-      id={`AxialSelectToZoom${axis}`}
+      id={`${axis.toUpperCase()}SelectToZoom`}
```
axelboc, line 74: Making the ID format more consistent with the name of the prop in `DefaultInteractions` (especially since we may end up allowing people to override interaction IDs).

packages/lib/src/interactions/DefaultInteractions.tsx
```diff
@@ -1,58 +1,80 @@
...
+
+export interface DefaultInteractionsConfig {
+  pan?: PanProps | false;
+  zoom?: ZoomProps | false;
+  xAxisZoom?: XAxisZoomProps | false;
+  yAxisZoom?: YAxisZoomProps | false;
+  selectToZoom?: Omit<SelectToZoomProps, 'keepRatio'> | false;
+  xSelectToZoom?: Omit<AxialSelectToZoomProps, 'axis'> | false;
+  ySelectToZoom?: Omit<AxialSelectToZoomProps, 'axis'> | false;
+  withAxialZoom?: boolean; // shortcut to turn off all axial zoom interactions
```
axelboc, line 22: `withAxialZoom` is just a convenience prop to disable axial zoom completely. This is useful for the `ScatterVis` currently (see comment).

packages/lib/src/vis/scatter/ScatterVis.tsx
```diff
@@ -104,9 +104,7 @@ function ScatterVis(props: Props) {
         }}
         title={title}
       >
-        <DefaultInteractions
-          interactions={{ XAxisZoom: false, YAxisZoom: false, ...interactions }}
-        />
+        <DefaultInteractions withAxialZoom={false} {...interactions} />
```
axelboc, line 107: Note that previously, we were turning off wheel-based axial zoom completely but not selection-based axial zoom. I wasn't sure if this was intended, so now all axial zoom interactions are off. I have no strong feeling about this and `withAxialZoom`, so let me know what you think.

packages/lib/src/interactions/DefaultInteractions.tsx
```diff
@@ -1,58 +1,80 @@
...
+import type { YAxisZoomProps } from '../interactions/YAxisZoom';
 import Zoom from '../interactions/Zoom';
+import type { ZoomProps } from '../interactions/Zoom';
 import AxialSelectToZoom from './AxialSelectToZoom';
-import type { Interactions } from './models';
-import { getDefaultInteractions } from './utils';
+import type { AxialSelectToZoomProps } from './AxialSelectToZoom';
+
+export interface DefaultInteractionsConfig {
+  pan?: PanProps | false;
```
axelboc, line 15: So now we specify the props of each interaction component explicitly (not just `Interaction`). Note also that previously, we were allowing `true` to be passed, which is kind of pointless (unless I've missed something?).

packages/lib/src/interactions/DefaultInteractions.tsx
```diff
@@ -1,58 +1,80 @@
...
-      }
-
-      return [k, v];
-    })
-  );
-
-  const interactions = merge(
-    getDefaultInteractions(keepRatio),
-    parsedInteractions
-  );
```
axelboc, line 37: At first, I tried keeping this algorithm, but I was losing a lot of type strictness and had to do a lot of castings. I went with a more declarative version where each interaction is dealt with explicitly in the JSX.

packages/lib/src/interactions/DefaultInteractions.tsx
```diff
@@ -1,58 +1,80 @@
...
-      {interactions.YAxisZoom && <YAxisZoom {...interactions.YAxisZoom} />}
-      {interactions.SelectToZoom && (
-        <SelectToZoom {...interactions.SelectToZoom} keepRatio={keepRatio} />
+      {interactions.pan !== false && <Pan {...interactions.pan} />}
+      {interactions.zoom !== false && <Zoom {...interactions.zoom} />}
+
+      {interactions.xAxisZoom !== false && withAxialZoom && (
+        <XAxisZoom
+          modifierKey="Alt"
+          disabled={keepRatio}
```
axelboc, line 40: Default config is not declared explicitly here with props.

packages/lib/src/interactions/DefaultInteractions.tsx
loichuder: If I remember correctly, `true` was meant to be a short hand for `{}` as props (no modifier key, enabled)

packages/lib/src/interactions/DefaultInteractions.tsx
loichuder: It feels a bit redundant with `keepRatio` that has the same effect. I would not add it.

packages/lib/src/vis/scatter/ScatterVis.tsx
loichuder: Not sure why I decided to remove axial zoom. I reckon that it can be enabled for `ScatterVis`.

loichuder (approved):
Great initiative and nice work :clap:

packages/lib/src/interactions/DefaultInteractions.tsx
axelboc: Which is pointless to use, since it's the same as not setting the property at all, right?

packages/lib/src/vis/scatter/ScatterVis.tsx
axelboc: Yeah alright, I'll remove the `withAxialZoom` prop as well, what do you reckon?

axelboc: Ah just saw your other comment :+1:

PR #1149, opened 2022-06-30
Let Pan accept multiple mouse buttons

### LINKED ISSUE #1143: [Pan] Support panning with both left and middle mouse buttons
Here is the interaction i want to implement.

Just to let you know that it sounds like i can't describe the interaction this way.

Sounds like using twice `Pan` interlock the interaction. And probably twice `SelectToZoom`

No idea if there is a problem somehow. Or this that have to warn or fail.

```
function DefaultMouseInteraction(props: {
  zoomMode: boolean;
  panMode: boolean;
  disabled: boolean;
}) {
  console.log(props.panMode, props.zoomMode);
  return (
    <>
      <Pan disabled={!props.panMode || props.disabled} />
      <Pan disabled={props.panMode || props.disabled} modifierKey="Control" />
      <SelectToZoom disabled={!props.zoomMode || props.disabled} keepRatio />
      <SelectToZoom
        disabled={props.zoomMode || props.disabled}
        keepRatio
        modifierKey="Control"
      />
    </>
  );
}
```

I will just rewrite my code to use a `if`.

But you maybe have a better idea.

axelboc:
Fix #1143 

```tsx
<Pan button={[MouseButton.Left, MouseButton.Middle]} />
```

With the above, panning is allowed with either the left or the middle button. Obviously, contrary to the `modifierKey` prop, the logic here is "OR", not "AND".

I've documented the new behaviour in the Storybook and took the opportunity to add a story to demonstrate setting multiple modifier keys.

### DIFF
--- a/apps/storybook/src/Pan.stories.tsx
+++ b/apps/storybook/src/Pan.stories.tsx
@@ -1,15 +1,11 @@
+import { MouseButton } from '@h5web/lib';
 import { Pan, ResetZoomButton, VisCanvas, Zoom } from '@h5web/lib';
-import type { ModifierKey as ModifierKeyType } from '@h5web/lib';
+import type { PanProps } from '@h5web/lib';
 import type { Meta, Story } from '@storybook/react';
 
-import VisCanvasStoriesConfig from './VisCanvas.stories';
+import FillHeight from './decorators/FillHeight';
 
-interface TemplateProps {
-  disabled?: boolean;
-  modifierKey?: ModifierKeyType;
-}
-
-const Template: Story<TemplateProps> = (args) => {
+const Template: Story<PanProps> = (args) => {
   return (
     <VisCanvas
       abscissaConfig={{ visDomain: [-10, 0], showGrid: true }}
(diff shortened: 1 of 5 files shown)


apps/storybook/src/Pan.stories.tsx
```diff
@@ -23,36 +19,53 @@ const Template: Story<TemplateProps> = (args) => {
...
+  args: {
+    button: [MouseButton.Left],
+    modifierKey: [],
+    disabled: false,
   },
   argTypes: {
+    button: {
+      control: {
+        type: 'inline-check',
+        labels: { [MouseButton.Left]: 'Left', [MouseButton.Middle]: 'Middle' },
```
axelboc, line 62: I have to explicitly set the labels of the options, as by default it uses the values of the enum (`0`, `1`)

apps/storybook/src/Pan.stories.tsx
```diff
@@ -23,36 +19,53 @@ const Template: Story<TemplateProps> = (args) => {
...
+      },
+      options: [MouseButton.Left, MouseButton.Middle],
+    },
     modifierKey: {
-      control: { type: 'inline-radio' },
+      control: { type: 'inline-check' },
```
axelboc, line 67: Using checkboxes for the modifier keys as well, as mentioned.

apps/storybook/src/Pan.stories.tsx
```diff
@@ -23,36 +19,53 @@ const Template: Story<TemplateProps> = (args) => {
...
       options: ['Alt', 'Control', 'Shift'],
     },
   },
-} as Meta;
+} as Meta<PanProps>;
```
axelboc, line 71: This generic allows `args` and `argsType` properties to be auto-completed.

packages/lib/src/index.ts
```diff
@@ -1,11 +1,11 @@
...
 export { default as MatrixVis } from './vis/matrix/MatrixVis';
-export type { MatrixVisProps } from './vis/matrix/MatrixVis';
 export { default as LineVis } from './vis/line/LineVis';
-export type { LineVisProps } from './vis/line/LineVis';
 export { default as HeatmapVis } from './vis/heatmap/HeatmapVis';
-export type { HeatmapVisProps } from './vis/heatmap/HeatmapVis';
 export { default as ScatterVis } from './vis/scatter/ScatterVis';
+export type { MatrixVisProps } from './vis/matrix/MatrixVis';
+export type { LineVisProps } from './vis/line/LineVis';
+export type { HeatmapVisProps } from './vis/heatmap/HeatmapVis';
```
axelboc, line 8: Just grouping the type exports together for readability.

packages/lib/src/interactions/InteractionsProvider.tsx
```diff
@@ -43,12 +43,14 @@ function InteractionsProvider(props: { children: ReactNode }) {
     (interactionId: string, event: MouseEvent | WheelEvent) => {
       const registeredInteractions = [...interactionMap.values()];
 
-      function isButtonPressed(button: MouseButton | 'Wheel') {
+      function isButtonPressed(button: MouseButton | MouseButton[] | 'Wheel') {
```
axelboc, line 46: I'm not bothering to normalise `button` into an array like with `modifierKey`, since it is only used below.

apps/storybook/src/Pan.stories.tsx
loichuder: Good to know :+1:

apps/storybook/src/Pan.stories.tsx
loichuder: :sparkles:

loichuder (approved):

PR #1151, opened 2022-06-30
Warn when trying to register two interactions with the same ID

axelboc:
Now, when this occurs, the second interaction no longer replaces the first. See https://github.com/silx-kit/h5web/issues/1143#issuecomment-1164050717


loichuder:
Warning instead of throwing is fine.

If this happens during dev, it is likely that it is because we are testing and/or adding a new interaction so we are likely to investigate if the interaction is not properly added.

PR #1153, opened 2022-06-30
Fix detection of pressed modifier keys

### LINKED ISSUE #1152: Bug when unmounting/remounting interaction components
### Describe the bug

I've confirmed that @vallsv's second example in https://github.com/silx-kit/h5web/issues/1143#issuecomment-1160240019 is not working as expected. Sure, Valentin's [third example](https://github.com/silx-kit/h5web/issues/1143#issuecomment-1160340772) works and shows how interaction components are meant to be used, but still, the second example should work in theory, since we register/unregister interactions on mount/unmount.

### To Reproduce

Drop this story in the storybook somewhere:

```tsx
export const Test: Story<{ zoomMode: boolean }> = (args) => {
  return (
    <VisCanvas
      abscissaConfig={{ visDomain: [-10, 0], showGrid: true }}
      ordinateConfig={{ visDomain: [50, 100], showGrid: true }}
    >
      {args.zoomMode ? (
        <>
          <Pan modifierKey="Control" />
          <SelectToZoom keepRatio />
        </>
      ) : (
        <>
          <Pan />
          <SelectToZoom keepRatio modifierKey="Control" />
        </>
      )}
    </VisCanvas>
  );
};
Test.args = {
  zoomMode: false,
};
```

1. Open the story
2. Press <kbd>Ctrl</kbd>, draw a box and release => select-to-zoom works as expected
3. Click and drag => pan works as expected
4. Turn on `zoomMode` in the Storybook controls panel
5. Try to draw a selection box, now without pressing any modifier key => **nothing happens, but if you release the mouse, the zoom does work**

Now if you refresh the page with `zoomMode` enabled, the problem is reversed:

1. Click and drag without pressing any modifier key => select-to-zoom works as expected
2. Press <kbd>Ctrl</kbd> and drag => pan works as expected
3. Turn off `zoomMode`
4. Try to pan without pressing any modifier key => **it doesn't work**

axelboc:
Fix #1152 

Well, this was not trivial. The bug described in #1152 came from the `useModifierKeyPressed` hook used in `Pan` and `SelectionTool`. If you recall, the goal of this hook is to stop an interaction when the user releases the required modifier key(s).

Turns out that this hook could not handle changes to the `modifierKeys` array. The internal `map` was not being reset, so you'd end up with keys in there, like `{ Control: false }`, that were no longer in the `modifierKeys` array.

I've reworked the hook to fix the issue, and I've taken the opportunity to improve optimise things a little (fewer re-renders).

I've also fixed another issue that was preventing the pan/select-to-zoom interactions with modifier key to start when the user starts the interaction while the window is out of focus.


packages/lib/src/interactions/hooks.ts
```diff
@@ -185,21 +186,40 @@ export function useInteraction(id: string, value: InteractionEntry) {
 }
 
 export function useModifierKeyPressed(modifierKeys: ModifierKey[]): boolean {
```
axelboc, line 188: I could have gone for a `useEffect` that resets the map when `modifierKeys` changes (with `useDeepCompareEffect` to avoid having to memoise the array), but this seemed overly convoluted.

Instead:

1. I initialise a `Map` that doesn't trigger re-renders when updated.
2. Then, every time a key is pressed or released:
  a. I update the map without even bothering to check whether the key pressed/released is a modifier key.
  b. I call `checkAllPressed` which loops through the `modifierKeys` array (instead of the map) and checks whether every  given modifier key is pressed (by looking in the `pressedKeys` map).
  c. I then update the `allPressed` state, but only if needed.

This solution means that we don't re-render every time a key is pressed or released but only when the `allPressed` state changes.

packages/lib/src/interactions/hooks.ts
```diff
@@ -12,11 +13,11 @@ import type {
   InteractionEntry,
   ModifierKey,
 } from './models';
-import { clampPositionToArea, isModifierKey } from './utils';
+import { clampPositionToArea } from './utils';
 
 const ZOOM_FACTOR = 0.95;
-
 const ONE_VECTOR = new Vector3(1, 1, 1);
+const MODIFIER_KEYS = ['Alt', 'Control', 'Shift'];
```
loichuder, line 20:
```suggestion
const MODIFIER_KEYS: ModifierKey[] = ['Alt', 'Control', 'Shift'];
```

loichuder (approved):
Clever :100:

PR #1155, opened 2022-07-01
Define raycaster thresholds via VisCanvas

loichuder:


packages/lib/src/vis/shared/AdaptativeThresholdEnforcer.tsx
```diff
@@ -0,0 +1,18 @@
+import { useRaycasterThreshold } from '../hooks';
+
+interface Props {
+  lineThreshold?: number;
+  pointThreshold?: number;
+}
+
+function AdaptativeThresholdEnforcer(props: Props) {
```
axelboc, line 8: What about `ThresholdAdjuster`?

packages/lib/src/vis/shared/AdaptativeThresholdEnforcer.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const { lineThreshold = 1, pointThreshold = 1 } = props;
+
+  useRaycasterThreshold('Line', lineThreshold);
+  useRaycasterThreshold('Points', pointThreshold);
```
axelboc, line 12: Would it make sense to move the content of the `useRaycastThreshold` hook in here, change `Props` to `RaycasterParams` and just re-assign the entire `raycaster.params` object?

axelboc (changes requested):

packages/lib/src/vis/shared/AdaptativeThresholdEnforcer.tsx
axelboc: And in the `useFrame`, I guess, assign the new threshold to any key that exists in the `RaycasterParams` object.

loichuder: Yeah I thought about it but this forces me to dig into the `raycaster.params` (and deal with several levels of `undefined`) to find the `desiredThreshold` values...

axelboc: Ah right, didn't realise `RaycasterParameters` had potentially more stuff than just thresholds. In that case I would maybe define an interface `Thresholds { Line?: number; Points?: number }` and work with that, still inline inside the component instead of in a custom hook. It's fine if you have to duplicate lines IMO.

loichuder: The interface is generic but you can only specify `threshold` through it from what I get: https://threejs.org/docs/#api/en/core/Raycaster.params

packages/lib/src/vis/shared/ThresholdAdjuster.tsx
```diff
@@ -0,0 +1,42 @@
...
+        params.threshold = oldThreshold;
+      }
+    };
+  }, [raycaster, value]);
+
+  useFrame(() => {
+    // Points stay the same size on the screen when zooming
+    // but as the threshold is in world units, it grows when zooming
+    // So we adapt the threshold value to the zoom
+    const zoom = (camera.scale.x + camera.scale.y) / 2;
```
loichuder, line 32: I think this formula is better in case of axial zoom

axelboc (approved):

PR #1157, opened 2022-07-05
Add new props to DataCurve to tune glyph aspect

loichuder:


packages/lib/src/vis/line/GlyphMaterial.tsx
```diff
@@ -79,22 +78,30 @@ function GlyphMaterial(props: Props) {
...
   if (color) {
     // If color is given, make it a uniform in the shader
     return (
       <shaderMaterial
-        uniforms={{ color: { value: new Color(color) }, ...shader.uniforms }}
-        vertexShader={`uniform vec3 color; ${shader.vertexShader}`}
-        fragmentShader={shader.fragmentShader}
+        args={[
+          {
+            uniforms: { ...uniforms, color: { value: new Color(color) } },
```
axelboc, line 89: I think you may be able to remove the need for the early return here by following the same approach as in `HeatmapMesh`: by always defining the uniform in the shader and always passing the uniform (either `undefined` or a fallback value, not sure what works). `getUniforms` may help simplify the code as well, by avoiding the inner `{ value }` object.

By the way, looking back at the shaders, I'm wondering why the color is passed as a `varying vec3` via the vertex shader instead of directly as a `uniform vec3` to the fragment shader. Do you remember?

loichuder: > By the way, looking back at the shaders, I'm wondering why the color is passed as a varying vec3 via the vertex shader instead of directly as a uniform vec3 to the fragment shader. Do you remember?

Yep, it is the super trick to have coexisting uniform color and `vertexColors` work.
- With `vertexColors`, `color` is defined in the `vertexShader` point-by-point, set to `vColor` and then passed to the fragment
- Without `vertexColor`, `color` needs to be defined in another way. This is the point of the `uniform vec3 color`. Then, again set to `vColor` and passed to the fragment`.

If it wasn't for the coexistence of the two cases, we might be able to set `uniform vec3 color` in the fragment shader and be done with it.

loichuder: > I think you may be able to remove the need for the early return here by following the same approach as in HeatmapMesh: by always defining the uniform in the shader and always passing the uniform (either undefined or a fallback value, not sure what works). getUniforms may help simplify the code as well, by avoiding the inner { value } object.

Not a huge fan of the result (431bacd063fd25494558af67243e52fb204c3443): the condition on `color` gets duplicated which makes the code harder to understand IMO.

Let me know what you think.

axelboc: I've played around with it and came up with this solution, which avoids passing the `color` via the vertex shader when it is provided. What do you think?

```tsx
function GlyphMaterial(props: Props) {
  const { color, size, glyphType } = props;

  // If no `color` given, use vertex colors (i.e. buffer attribute on geometry)
  const withVertexColor = !color;

  const shader = {
    uniforms: getUniforms({
      size,
      color: new Color(color),
    }),
    vertexShader: `
      uniform float size;
      varying vec3 vertexColor;

      void main() {
        gl_PointSize = size;
        gl_Position = projectionMatrix * modelViewMatrix * vec4( position, 1.0 );
        ${withVertexColor ? 'vertexColor = color;' : ''}
      }
    `,
    fragmentShader: `
      uniform float size;
      uniform vec3 color;
      varying vec3 vertexColor;

      ${GLYPH_SHADERS[glyphType]}

      void main() {
        float alpha = alphaSymbol(gl_PointCoord, size);
        if (alpha <= 0.0) {
            discard;
        } else {
            gl_FragColor = vec4(${
              withVertexColor ? 'vertexColor.rgb' : 'color.rgb'
            }, alpha);
        }
      }
    `,
  };

  return <shaderMaterial args={[shader]} vertexColors={withVertexColor} />;
}
```

PR #1161, opened 2022-07-07
Allow 1D compound dataset to be viewed in MatrixVis

### LINKED ISSUE #1130: [Matrix] Support compound datasets with 1D fields

loichuder:
Fix #1130 

image


packages/app/src/vis-packs/core/compound/CompoundMatrixVisContainer.tsx
```diff
@@ -0,0 +1,36 @@
...
+import type { VisContainerProps } from '../../models';
+import ValueFetcher from '../ValueFetcher';
+import MappedCompoundMatrixVis from './MappedCompoundMatrixVis';
+
+function CompoundMatrixVisContainer(props: VisContainerProps) {
+  const { entity, toolbarContainer } = props;
+  assertDataset(entity);
+  assertArrayShape(entity);
+  assertCompoundType(entity);
+  assertPrintableFields(entity);
```
loichuder, line 18: As a reminder, a compound dtype has a `fields` member that is a record of the fields' dtype.

With this new assertion, I check that these fields are printable (`PrintableDType`) which infers the dataset dtype to `PrintableCompoundDtype`. See in the below comments why this inference is useful. Rereading this, I could combine the two assertions (compound and printable fields) in one.

packages/shared/src/models-hdf5.ts
```diff
@@ -157,6 +161,8 @@ export type Primitive<T extends DType> = T extends NumericType
   ? boolean
   : T extends ComplexType
   ? H5WebComplex
+  : T extends PrintableCompoundType
+  ? Primitive<PrintableType>[]
```
loichuder, line 165: This is where having a subtype `PrintableCompoundType` is useful. We can then infer the `Primitive` of this dtype and therefore the `Value` type when fetching the value in `ValueFetcher`.

The values of compound is an array containing tuple of the field entries. For example, for the compound dataset `compound/dogs` in https://h5web.panosc.eu/h5grove?file=grove.h5 where we have three fields (`name`, `age` and `weight`) the value is
```
[
  ['Rex', 9, 81],
  ['Fido', 3, 27],
]
```

The nice thing about `PrintableType` is that we don't really need to know what is the exact tuple type (here  `[string, number, number]`) as long as the tuple contains printable types which is exactly what `Primitive<PrintableType>[]` describes.

packages/app/src/vis-packs/core/compound/MappedCompoundMatrixVis.tsx
```diff
@@ -0,0 +1,60 @@
...
+  value: Primitive<PrintableType>[][];
+  toolbarContainer: HTMLDivElement | undefined;
+}
+
+function MappedCompoundMatrixVis(props: Props) {
+  const { value, dataset, toolbarContainer } = props;
+  const { fields } = dataset.type;
+
+  const fieldNames = Object.keys(fields).map(capitalize);
+  const cellWidth = Math.max(...Object.values(fields).map(getCellWidth));
```
loichuder, line 30: I modified slightly `getCellWidth` to be able to pass the dtype rather than the dataset to be able to use it for field dtypes.

This way, we can set the width to fit the longest dtype.

packages/app/src/vis-packs/core/matrix/MatrixToolbar.tsx
```diff
@@ -16,8 +16,8 @@ import { useMatrixConfig } from './config';
 const EXPORT_FORMATS: ExportFormat[] = ['npy', 'csv'];
 
 interface Props {
-  dataset: Dataset<ArrayShape, PrintableType>;
-  selection: string | undefined;
+  dataset: Dataset<ArrayShape>;
+  selection?: string | undefined;
```
loichuder, line 20: Allow `selection` to be omitted in `MappedCompoundMatrixVis`. Otherwise, I have to set it to `undefined` explicitely :shrug:

packages/app/src/vis-packs/core/compound/MappedCompoundMatrixVis.tsx
```diff
@@ -0,0 +1,60 @@
...
+  dataset: Dataset<ArrayShape, PrintableCompoundType>;
+  value: Primitive<PrintableType>[][];
+  toolbarContainer: HTMLDivElement | undefined;
+}
+
+function MappedCompoundMatrixVis(props: Props) {
+  const { value, dataset, toolbarContainer } = props;
+  const { fields } = dataset.type;
+
+  const fieldNames = Object.keys(fields).map(capitalize);
```
axelboc, line 29: I would not use `capitalize` here, as some fields may intentionally start with a lower-case letter, e.g. if they correspond to variable or dataset names. How about capitalising the fields in the mock data instead?

packages/shared/src/models-hdf5.ts
axelboc: This is great. 🎉

May I just suggest adding a generic to `CompoundType`?

```ts
export interface CompoundType<T = DType> {
  class: DTypeClass.Compound;
  fields: Record<string, T>;
}

// Then you can use `CompoundType<PrintableType>` directly or make an alias if you prefer:
export type PrintableCompoundType = CompoundType<PrintableType>
```

packages/shared/src/guards.ts
```diff
@@ -388,3 +390,32 @@ export function isTypedNdArray<T extends NumArray>(
...
+
+export function assertCompoundType<S extends Shape>(
+  dataset: Dataset<S>
+): asserts dataset is Dataset<S, CompoundType> {
+  if (!hasCompoundType(dataset)) {
+    throw new Error('Expected dataset to have compound type');
+  }
+}
+
+export function hasPrintableFields<S extends Shape>(
```
axelboc, line 408: Although the current names make sense, it might be best to stick to the more consistent naming: `hasPrintableCompoundType` and `assertPrintableCompoundType`?

packages/shared/src/guards.ts
```diff
@@ -388,3 +390,32 @@ export function isTypedNdArray<T extends NumArray>(
...
+  dataset: Dataset<S, CompoundType>
+): dataset is Dataset<S, PrintableCompoundType> {
+  const { fields } = dataset.type;
+  return Object.values(fields).every((f) => PRINTABLE_DTYPES.has(f.class));
```
axelboc, line 412: Why not use `hasPrintableType` here?

packages/shared/src/guards.ts
```diff
@@ -388,3 +390,32 @@ export function isTypedNdArray<T extends NumArray>(
 ): ndArr is NdArray<Exclude<T, number[]>> {
   return ndArr.dtype !== 'array';
 }
+
+export function hasCompoundType<S extends Shape>(
```
axelboc, line 394: I would maybe move all those new guards/assertions a bit further up, right after the other dtype guards/assertions.

packages/app/src/vis-packs/core/compound/MappedCompoundMatrixVis.tsx
```diff
@@ -0,0 +1,60 @@
...
+      {toolbarContainer &&
+        createPortal(
+          <MatrixToolbar dataset={dataset} cellWidth={cellWidth} />,
+          toolbarContainer
+        )}
+      <MatrixVis
+        dataArray={dataArray}
+        formatter={compoundFormatter}
+        cellWidth={customCellWidth ?? cellWidth}
+        columnHeaders={fieldNames}
```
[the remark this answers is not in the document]
axelboc: Nicely done 👏

packages/app/src/vis-packs/core/compound/MappedCompoundMatrixVis.tsx
axelboc: You could go half-way by calling `getCellWidth` with the compound type and adding a case for that that loops through the fields and recurses:

```ts
getCellWidth(dataset.type)

function getCellWidth(type: PrintableType | PrintableCompoundType): number {
  ...
  if (type.class === DTypeClass.Compound) {
    return Math.max(...Object.values(type.fields).map(getCellWidth));
  }
  ...
}
```

packages/app/src/vis-packs/core/compound/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (Array.isArray(value)) {
+    return formatMatrixComplex(value);
+  }
+
+  if (typeof value === 'number') {
+    return formatMatrixValue(value);
+  }
+
+  return value.toString();
+}
```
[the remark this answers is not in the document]
axelboc: Agreed and agreed 👍

axelboc (approved):
Impressive stuff 🤩

Good call on adding a new container and mapped vis -- I assume re-using the existing ones would have required making some really weird type guards and assertions... 😂

packages/shared/src/guards.ts
loichuder: `hasPrintableType` takes a dataset in input and not a `DType` (which is the type of `fields`)

PR #1162, opened 2022-07-07
Allow to set errors for aux signals in NxSpectrum

### LINKED ISSUE #1029: [NX Spectrum] Errors on auxiliary signals

loichuder:
Fix #1029 

image


packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -109,7 +109,19 @@ export function useMappedArrays(
...
   mapping: DimensionMapping,
   autoScale?: boolean
-): [NdArray<NumArray>[], NdArray<NumArray>[]] {
+): [NdArray<NumArray>[], NdArray<NumArray>[]];
+export function useMappedArrays(
+  values: (NumArray | undefined)[],
+  dims: number[],
+  mapping: DimensionMapping,
+  autoScale?: boolean
+): [(NdArray<NumArray> | undefined)[], (NdArray<NumArray> | undefined)[]];
```
loichuder, line 118: Overload to be able to pass arrays with `undefined` inside while keeping the old typing when there is no `undefined`.

packages/app/src/vis-packs/nexus/hooks.ts
```diff
@@ -63,14 +68,19 @@ export function useAxisMapping(
...
-  const auxValues = useDatasetValues(auxDatasets, selection);
+  const auxValues = useDatasetValues(
+    [...auxDatasets, ...auxErrorsDatasets],
+    selection
+  );
 
   return auxDatasets.map((dataset) => ({
     label: getDatasetLabel(dataset, attrValuesStore),
     value: auxValues[dataset.name],
+    error: auxValues[`${dataset.name}_errors`],
```
loichuder, line 84: `auxErrorsDatasets` may not have the same number of entries than `auxDatasets` (one could imagine having aux datasets with errors and some without) so I can't rely on the index in the array.

I instead reconstruct the key in `auxValues` from the dataset name. Again, not great. Perhaps we could find a way to abstract this reconstruction to be used in `findErrorsDataset` ?

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -85,11 +85,12 @@ export function findSignalDataset(
...
   group: GroupWithChildren,
-  signalName: string
+  signalName?: string
 ): NumArrayDataset | undefined {
-  const dataset =
-    getChildEntity(group, `${signalName}_errors`) ||
-    getChildEntity(group, 'errors');
+  const dataset = getChildEntity(
+    group,
+    signalName ? `${signalName}_errors` : 'errors'
```
loichuder, line 92: Little change to avoid matching the dataset called `errors` to aux datasets

packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -185,11 +186,13 @@ function LineVis(props: Props) {
           color={curveColor}
           curveType={curveType}
         />
-        {auxiliaries.map(({ array }, i) => (
+        {auxiliaries.map(({ array, label, errors }, i) => (
           <DataCurve
-            key={i} // eslint-disable-line react/no-array-index-key
+            key={label}
```
loichuder, line 191: I think we can reasonably expect `label` to be unique

packages/shared/src/mock/metadata-utils.ts
```diff
@@ -303,6 +303,7 @@ export function makeNxDataGroup<
       ...(auxAttr ? [makeNxAuxAttr(auxAttr)] : []),
     ],
     children: [
+      ...(groupOpts.children ?? []),
```
loichuder, line 306: Small modification to be able to add the `secondary_errors` dataset below

axelboc: I think you should add `children` to the destructuring statement instead (with `[]` as default value), so it doesn't get included in `groupOpts` in the first place.

axelboc (changes requested):
Awesome, this did not look trivial... 👏 I've made a couple of minor suggestions, and one more significant about the shape of `auxErrorsDatasets` (which may have unforeseen challenges, I don't know).

packages/shared/src/mock/metadata-utils.ts
loichuder: Sounds good, I will do the same for `attributes` above

packages/app/src/vis-packs/nexus/hooks.ts
axelboc: Ooh smart, taking a look at it now.

packages/app/src/vis-packs/nexus/hooks.ts
```diff
@@ -62,15 +68,22 @@ export function useAxisMapping(
...
+    selection
+  );
 
-  return auxDatasets.map((dataset) => ({
-    label: getDatasetLabel(dataset, attrValuesStore),
-    value: auxValues[dataset.name],
+  return auxDatasets.map(({ signal, errors }) => ({
+    label: getDatasetLabel(signal, attrValuesStore),
+    value: auxValues[signal.name],
+    error: errors ? auxValues[errors.name] : undefined,
```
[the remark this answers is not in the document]
axelboc: I'm also tempted to suggest pluralising the keys in the object: `values` and `errors`, since they each hold an array of data/error values.

Maybe we'll have to do a pass later to decide once and for all when to pluralise those two terms. Because even though `errors` and `values` on their own need to be pluralised when they hold an array, they don't necessarily need to be when used as [attributive nouns](https://english.stackexchange.com/a/397244):

- `errorArray`: the array that holds the error information
- `errorsArray`: the array that holds multiple error values (to convey variety) or the array that "belongs to" the errors (to convey possession)

I'd argue that we don't really need to convey variety or possession, and therefore that the singular form might in fact be more appropriate.

packages/shared/src/mock/metadata-utils.ts
```diff
@@ -290,19 +290,22 @@ export function makeNxDataGroup<
...
     attributes: [
-      ...(groupOpts.attributes ?? []),
+      ...attributes,
       makeStrAttr('signal', signal.name),
       ...(axesAttr ? [makeNxAxesAttr(axesAttr)] : []),
       ...(silxStyle ? [makeSilxStyleAttr(silxStyle)] : []),
       ...(auxAttr ? [makeNxAuxAttr(auxAttr)] : []),
     ],
     children: [
+      ...children,
```
axelboc, line 308: Doesn't really matter, but `...children` and `...attributes` should perhaps be last in the array. There's no risk of overriding anything since we're not destructuring objects, and it feels like custom attributes/children should come last.

packages/app/src/vis-packs/nexus/hooks.ts
axelboc: Much better indeed :clap:

packages/app/src/vis-packs/nexus/hooks.ts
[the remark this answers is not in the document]
loichuder: Make sense

packages/app/src/vis-packs/nexus/models.ts
```diff
@@ -24,13 +24,18 @@ export type NxAttribute =
   | 'SILX_style'
   | 'auxiliary_signals';
 
+export type AuxDatasets = {
+  signal: NumArrayDataset;
+  errors?: NumArrayDataset;
+}[];
```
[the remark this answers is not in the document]
axelboc: Hmm excellent point :+1:

PR #1163, opened 2022-07-09
Add docker and docker-compose support

linupi:
Together with https://github.com/silx-kit/h5grove/pull/70 this is a first attempt to add a Dockerfile to build a production-ready image of h5web and bundle it as h5grove as a stand-alone micro-service in docker-compose to illustrate the use-case.

since there are no official h5web and 5hgrove images in the docker registry yet the docker-compose build context is used to create the corresponding images on the spot.

```
docker-compose build
docker-compose up
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+# 
+# Build stage 1.
+#
+FROM node:16-alpine
+
+COPY . ./h5web/
+WORKDIR /h5web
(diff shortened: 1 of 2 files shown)


loichuder:
After discussion, docker support and examples are now done in a separate repository: https://github.com/silx-kit/h5whale

Further Docker-based developments will happen there.

PR #1166, opened 2022-07-12
Allow overriding `Pan` interaction ID

axelboc:
Follow-up to #1143

The solution of allowing multiple buttons to be passed to the `Pan` component did not allow for the scenario where one and only of the buttons needs to be combined to a modifier key - e.g. middle button without modifier key AND left button with `Ctrl` modifier key:

```tsx
      <Pan id="PanMiddle" button={MouseButton.Middle} />
      <Pan id="PanLeft" button={MouseButton.Left} modifierKey="Control" />
```

### DIFF
--- a/apps/storybook/src/Pan.stories.tsx
+++ b/apps/storybook/src/Pan.stories.tsx
@@ -45,17 +45,43 @@ Disabled.args = {
   disabled: true,
 };
 
+export const TwoComponents: Story<PanProps> = (args) => {
+  return (
+    <VisCanvas
+      title="Pan with middle button, or with <modifierKey> + left button"
+      abscissaConfig={{ visDomain: [-10, 0], showGrid: true }}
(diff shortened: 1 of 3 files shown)


apps/storybook/src/Pan.stories.tsx
```diff
@@ -45,17 +45,40 @@ Disabled.args = {
   disabled: true,
 };
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const TwoComponents: Story<Record<string, never>> = (args) => {
```
axelboc, line 49: I have to set the `args` parameter explicitly to avoid seeing a warning in the UI.

loichuder: You could leave the control for the `modifierKey` of `PanLeft` to make use of `args` if you need to put it anyway :shrug:

loichuder (approved):

PR #1168, opened 2022-07-12
Update storybook with auxiliary errors

loichuder:
Follow-up of #1162

### DIFF
--- a/apps/storybook/src/Utilities.stories.mdx
+++ b/apps/storybook/src/Utilities.stories.mdx
@@ -30,12 +30,13 @@ const myDomain = getDomain(myArray); // [0, 3]
 
 #### getDomains
 
-Find the domains of multiple numeric arrays. Useful when dealing with auxiliary curves.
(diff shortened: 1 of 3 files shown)


apps/storybook/src/Utilities.stories.mdx
```diff
@@ -30,12 +30,13 @@ const myDomain = getDomain(myArray); // [0, 3]
 
 #### getDomains
 
-Find the domains of multiple numeric arrays. Useful when dealing with auxiliary curves.
+Find the domains of multiple numeric arrays. Errors can be accounted for each array by providing an array of error numeric arrays of the same length. Useful when dealing with auxiliary curves.
```
axelboc, line 33:
```suggestion
Find the domains of multiple numeric arrays. Errors can be accounted for each array by providing an array of numeric error arrays of the same length. Useful when dealing with auxiliary curves.
```

axelboc (approved):

PR #1170, opened 2022-07-13
Update three.js to r136

### LINKED ISSUE #1167: [Heatmap] three.js r136 breaks the Heatmap

loichuder:
Fix #1167 

https://github.com/mrdoob/three.js/wiki/Migration-Guide#135--136 tells us that we need to set `needsUpdate` by ourselves now.


packages/lib/src/vis/rgb/utils.ts
```diff
@@ -2,7 +2,7 @@ import type { NumArray } from '@h5web/shared';
 import { getDims, toTypedNdArray } from '@h5web/shared';
 import type { NdArray } from 'ndarray';
 import ndarray from 'ndarray';
-import { DataTexture3D, FloatType, RedFormat, UnsignedByteType } from 'three';
+import { Data3DTexture, FloatType, RedFormat, UnsignedByteType } from 'three';
```
loichuder, line 5: `DataTexture3D` was renamed to `Data3DTexture` in r138: https://github.com/mrdoob/three.js/wiki/Migration-Guide#137--138

I renamed our function to match it

PR #1176, opened 2022-07-26
Use `useCameraState` only when necessary

axelboc:


packages/lib/src/interactions/AxialSelectToZoom.tsx
```diff
@@ -17,18 +16,18 @@ interface Props extends CommonInteractionProps {
...
+  const context = useAxisSystemContext();
   const moveCameraTo = useMoveCameraTo();
 
   const { width, height } = useThree((state) => state.size);
   const camera = useThree((state) => state.camera);
 
-  const { xVisibleDomain, yVisibleDomain } = useCameraState(
-    getVisibleDomains,
-    []
-  );
```
axelboc, line 29: We don't actually need to re-render this component when the visible domains change. We just need to compute the visible domains when we actually need them - i.e. when drawing a selection rectangle (in the render prop of `SelectionTool`) and when completing a selection (in `onSelectionEnd`).

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -17,30 +16,33 @@ interface Props extends CommonInteractionProps {
...
+  function getDataRatio() {
+    const { xVisibleDomain, yVisibleDomain } = getVisibleDomains(
+      camera,
+      context
+    );
+    return Math.abs(
+      (xVisibleDomain[1] - xVisibleDomain[0]) /
+        (yVisibleDomain[1] - yVisibleDomain[0])
+    );
+  }
```
axelboc, line 34: Similarly, we don't need to re-render and recompute the data ratio every time the visible domains change. Instead, we can compute the visible domains and data ratio only when needed - i.e. when drawing a selection rectangle (in the render prop of `SelectionTool`) and when completing a selection (in `onSelectionEnd`).

packages/lib/src/toolbar/floating/ResetZoomButton.tsx
```diff
@@ -1,21 +1,16 @@
...
-
-  useFrame(() => {
-    const isZoomedIn = camera.scale.x < 1 || camera.scale.y < 1;
-    if (isVisible !== isZoomedIn) {
-      setVisible(isZoomedIn);
-    }
-  });
+  const isVisible = useCameraState(({ scale }) => {
+    return scale.x < 1 || scale.y < 1;
+  }, []);
```
axelboc, line 13: Here, `useCameraState` makes sense, since we need to update the DOM, so I'm refactoring the code to actually use it. The behaviour is the same, just shorter code.

packages/lib/src/vis/hooks.ts
```diff
@@ -63,15 +63,11 @@ export function useCameraState<T>(
   const [state, setState] = useState(() => factory(camera, context));
 
   function updateState() {
-    const newState = factory(camera, context);
-
-    if (newState !== state) {
-      setState(newState);
-    }
+    setState(() => factory(camera, context)); // functional update to skip re-renders if state hasn't changed
```
axelboc, line 66: Very small simplification: the documentation of [`useState`](https://reactjs.org/docs/hooks-reference.html#usestate) specifies that when doing a functional state update like this (i.e. by passing a function to `setState`), React does not re-render if the state doesn't change. Basically, it does the equality check for us. 

Also, using the current state from `updateState` like this was pretty bad practice. That's exactly what the functional state update pattern is for. Sorry about that.

packages/lib/src/vis/hooks.ts
```diff
@@ -63,15 +63,11 @@ export function useCameraState<T>(
...
   }
 
   // Update state when context value or dependencies change
-  useEffect(updateState, [context, ...deps]); // eslint-disable-line react-hooks/exhaustive-deps
+  useEffect(updateState, [camera, context, ...deps]); // eslint-disable-line react-hooks/exhaustive-deps
```
axelboc, line 70: Probably doesn't matter, but I prefer to be safe.

packages/lib/src/toolbar/floating/ResetZoomButton.tsx
loichuder: Nice. I would keep `isZoomedIn` as the variable name, though.

packages/lib/src/vis/hooks.ts
loichuder: No worries. Each day is a learning experience :sun_with_face:

loichuder (approved):
Great stuff :100:

Issue #1158, opened 2022-07-06
data sources for h5web/h5wasm


linupi:
I really like the service that is offered at https://h5web.panosc.eu/h5wasm and was wondering what are the main foreseen data sources if one looks at this as a service e.g. to provide links to visualize h5 files e.g. in a documentation. In a way I compare it to the [Jupyter notebook viewer](https://nbviewer.org/). I can clearly see two use-cases, but maybe there are more:

- provide a view for files on github (examples would be files provided by [orso](https://github.com/reflectivity/file_format/tree/master/nexus_formats) or [NeXus](https://github.com/reflectivity/file_format/tree/master/nexus_formats) )
- provide a view for files hosted on zenodo [(e.g. the h5 file in here)](https://zenodo.org/record/6497438 )

The question would be if you'd consider it as useful to have a more easy way of specifying the datafile the user would should look at than just providing a URL (something like the github intergration in the nbviewer) for certain data sources.

### Is your feature request related to a problem?
no 

### Requested solution or feature
a feature

### Alternatives you've considered
obviously on can the url directly but I am not sure if that is the best user experience for the content provider

axelboc:
Hi @linupi, to avoid CORS issues with GitHub, the URL has to start with https://raw.githubusercontent.com/. Check out this repo for an example: https://github.com/oasys-esrf-kit/dabam2d

linupi:
Thanks for the hint regarding https://raw.githubusercontent.com/. For zenodo I figured by now that there is also a way to circumvent the CORS issue by going through the api. E.g. for the [xrr_dataset.h5 in zenodo.6497438](https://zenodo.org/record/6497438)  the correct url would be
```
https://zenodo.org/api/files/b0bfa823-917d-4fe6-9a02-4396743a6037/xrr_dataset.h5
```
where the id in the url can be resolved through the zenodo api (e.g. https://zenodo.org/api/records/6497438 ) which is quite un-intuitive for the user. I guess this is why I'd really like to see the h5wasm service to cope with zenodo and github in a user-friendly way by converting to the correct urls internally.

axelboc:
For a bit of context, https://h5web.panosc.eu/h5wasm is meant to just be a demo: a way for us to test the `H5WasmProvider` and to let people try the H5Web viewer on their own files to gather more valuable feedback. The whole demo app is continuously deployed from the `main` branch, so it can break easily and is definitely not production-ready.

[...]

PR #1179, opened 2022-07-27
Add support of int64 and fix 2d bool support in h5wasm

### LINKED ISSUE #1178: [h5wasm] Arrays of 64-bit integers are not supported
### Describe the bug

When loading a file with _h5wasm_, opening a dataset with the dtype `<i8` raises an error `Expected array or typed array` in the Visualize area.

### To Reproduce

1. Go to https://h5web.panosc.eu/h5wasm?url=https%3A%2F%2Fzenodo.org%2Fapi%2Ffiles%2Fb0bfa823-917d-4fe6-9a02-4396743a6037%2Fxrr_dataset.h5
2. Click on the `/DIP_1/experiment/scan` dataset
3. See error `Expected array or typed array`

The same dataset can be visualized fine when served by h5grove: https://h5web.panosc.eu/h5grove?file=xrr_dataset.h5

### Expected behaviour

Data contained in such datasets should be able to be visualized when loaded with h5wasm.

### More

The failing check is here: https://github.com/silx-kit/h5web/blob/d84a50f776b69da88937b2c99e2cc4e3a2cb3c41/packages/shared/src/guards.ts#L122. The failure comes from the fact that h5wasm encodes the values for such datasets in [BigInt64Array](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/BigInt64Array). 

For some reason, `isTypedArrayLodash` returns `false` for these arrays in the browser (while I had `true` when testing in `node`...). Although, even if the check succeeded, we would end up with an array with `BigInt` objects that must be converted to `Number` before feeding them to visualisations. So in any case, a conversion step is needed in `h5wasm-api`. 

So I see two possiblities:
- Do the conversion ourselves in `getValue` by detecting 64-bit integer datasets and mapping the resulting value.
- Let `to_array` from `h5wasm` handle it for such datasets. [It is supposed to return JSON-compatible values](https://github.com/usnistgov/h5wasm/blob/f1d7547255c32c7723dcd5e30e101e100bc7dc2b/CHANGELOG.md#added) even though it is not the case for now (but we could argue that as `BigInt` is converted to `Number`, `BigInt` arrays should be converted to `Number` arrays).

loichuder:
Fix #1178

@linupi This should fix the issue for the file you gave as example in #1158

### DIFF
--- a/packages/h5wasm/src/guards.ts
+++ b/packages/h5wasm/src/guards.ts
@@ -1,3 +1,5 @@
+import type { Dataset } from '@h5web/shared';
+import { DTypeClass } from '@h5web/shared';
 import {
   Dataset as H5WasmDataset,
   Group as H5WasmGroup,
@@ -92,4 +94,8 @@ export function assertNumericMetadata(
   if (!isNumericMetadata(metadata)) {
     throw new Error('Expected H5Wasm numeric metadata');
   }
+}
+
+export function hasInt64Type(dataset: Dataset) {
+  return dataset.type.class === DTypeClass.Integer && dataset.type.size === 64;
 }
--- a/packages/h5wasm/src/h5wasm-api.ts
+++ b/packages/h5wasm/src/h5wasm-api.ts
@@ -14,6 +14,7 @@ import type {
   GroupWithChildren,
   Shape,
 } from '@h5web/shared';
+import { hasArrayShape } from '@h5web/shared';
 import {
   assertNonNull,
   buildEntityPath,
@@ -24,6 +25,7 @@ import {
 import type { Attribute as H5WasmAttribute } from 'h5wasm';
 import { File as H5WasmFile, ready as h5wasmReady } from 'h5wasm';
 
+import { hasInt64Type } from './guards';
 import {
   assertH5WasmDataset,
   assertH5WasmEntityWithAttrs,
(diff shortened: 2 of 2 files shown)


packages/h5wasm/src/guards.ts
```diff
@@ -93,3 +95,7 @@ export function assertNumericMetadata(
     throw new Error('Expected H5Wasm numeric metadata');
   }
 }
+
+export function hasInt64Type(dataset: Dataset) {
+  return dataset.type.class === DTypeClass.Integer && dataset.type.size === 64;
+}
```
loichuder, line 101: I could put this with other guards in `shared` but it is only useful for h5wasm and we don't need to assert the TS type like the guards in `shared` do.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -54,12 +56,15 @@ export class H5WasmApi extends ProviderApi {
     const h5wDataset = await this.getH5WasmEntity(dataset.path);
     assertH5WasmDataset(h5wDataset);
 
-    if (hasBoolType(dataset)) {
-      const value = h5wDataset.to_array();
+    // h5wasm returns integers for bool and BigInt for int64
+    // So we use to_array instead to have bool and numbers resp.
+    if (hasBoolType(dataset) || hasInt64Type(dataset)) {
```
loichuder, line 61: Two-fold fix:

1. Use `to_array` for `int64` as well to have `BigInt` converted to numbers.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -54,12 +56,15 @@ export class H5WasmApi extends ProviderApi {
...
+      const rawValue = h5wDataset.to_array();
+      const value = hasArrayShape(dataset)
+        ? flattenValue(rawValue, dataset)
+        : rawValue;
```
loichuder, line 65: 2. As `to_array` returns nested arrays for nD datasets and as the visus expect 1D array, flatten it.

The previous code was flattening the result of `to_array` only if there was slicing. That means that 2D arrays were not flattened for 2D+ vis (`Matrix`, `Heatmap`),  which was breaking them.

axelboc (approved):
👏

Issue #1180, opened 2022-08-01
Rethink `DataProviderApi.getEntity()`


axelboc:
## Problem

The current `DataProviderApi` method `getEntity()` has a few issues:

### Manual child cache

In `DataProvider`, when creating `entitiesStore` based on `api.getEntity()`, we have to manage our own cache for child entities. The goal of this cache is to avoid fetching child entities twice, for instance when we first select a group and then one of its child datasets. However, this cache is quite problematic:

- If a child entity gets fetched directly before its parent group (e.g. because of `<App initialPath />`), it is basically fetched twice and can end up in both the `entitiesStore` cache _and_ the internal child cache.
- When we decide to better handle fetching errors for entity metadata (i.e. invalidate/retry), this child cache will be challenging to invalidate. Same if we decide we want an LRU cache for entities.

### API isn't clear, especially for groups

`getEntity` returns a generic `Entity` object. That's because we cannot deduce the type of the entity from its path alone. As a result, someone wanting to implement a new data provider API would have no idea that `getEntity` is supposed to return a `GroupWithChildren` object if the given path points to a group.

## Possible solution

These problems seem to stem from the fact that the data provider API was designed before we decided to save on requests by including children metadata in group responses in H5Grove. `getEntity` was designed to work with a path, and we did not see it as a problem at the time; on the contrary. So we kept the API as is and just made sure that the `entitiesStore` took advantage of the internal optimisation of H5Grove by adding the child cache. We also improved type safety with `GroupWithChildren` and its corresponding assertion.

However, I'm thinking this may no longer be the right way to go and that we may be able to simplify caching and make the API clearer **by replacing `getEntity(path: string): Entity` with `getGroup(path: string): GroupWithChildren`**, and `entitiesStore` with a `groupsStore`. The idea is that the app would always request a group's metadata, never a dataset/datatype's metadata. Here is how this would work for the three instances of `entitiesStore.get()`:

### `EntityList`

`EntityList` receives a `parentPath` as prop, which always refers to a group, and calls `entitiesStore.get(parentPath)`, so replacing this with `groupsStore.get(parentPath)` would be straightforward. We are never in a situation where we need to fetch a child dataset/datatype from a path.

### `MetadataViewer`

`MetadataViewer` receives a `path` as prop, which can refer to any type of entity, since it comes from the `selectedPath` state managed in `App`. The solution here is to store the selected entity instead of path in `App`, and let the `Explorer` inform the `App` when an entity is selected:

```tsx
[...]

PR #1182, opened 2022-08-02
Enforce stronger types for `DataProviderApi.getEntity()`

axelboc:
Stems from #1180 

This PR enforces that `DataProviderApi.getEntity()` cannot return a `Group` without children by introducing a model called `ProvidedEntity`, which is a union of `GroupWithChildren`, `Dataset`, `Datatype` and `UnresolvedEntity`.

I also introduce a model called `ChildEntity`, which is similar but includes `Group` instead of `GroupWithChildren`.

These two models bring quite a bit of clarity throughout the app and allow simplifying a couple of things. For instance, to assert that a `ProvidedEntity` is a `GroupWithChildren`, we now need to check only if it's a group and can just skip the `children` check.

### DIFF
--- a/packages/app/src/providers/DataProvider.tsx
+++ b/packages/app/src/providers/DataProvider.tsx
@@ -1,5 +1,5 @@
-import type { Entity } from '@h5web/shared';
-import { assertGroupWithChildren, isGroup } from '@h5web/shared';
+import type { ChildEntity, Entity, Group } from '@h5web/shared';
+import { isGroup } from '@h5web/shared';
 import type { PropsWithChildren } from 'react';
 import { createContext, useContext, useMemo } from 'react';
 import { createFetchStore } from 'react-suspense-fetch';
@@ -43,25 +43,23 @@ function DataProvider(props: PropsWithChildren<Props>) {
   const { api, children } = props;
 
   const entitiesStore = useMemo(() => {
-    const childCache = new Map<string, Entity>();
+    const childCache = new Map<string, Exclude<ChildEntity, Group>>();
 
     const store = createFetchStore(async (path: string) => {
-      if (childCache.has(path)) {
-        return childCache.get(path) as Entity;
+      const cachedEntity = childCache.get(path);
+      if (cachedEntity) {
+        return cachedEntity;
       }
 
       const entity = await api.getEntity(path);
 
       if (isGroup(entity)) {
-        // Make sure `getEntity` doesn't return groups without `children` proprety
-        assertGroupWithChildren(entity);
-
         // Cache non-group children (datasets, datatypes and links)
-        entity.children
-          .filter((child) => !isGroup(child))
-          .forEach((child) => {
+        entity.children.forEach((child) => {
+          if (!isGroup(child)) {
             childCache.set(child.path, child);
-          });
+          }
+        });
       }
 
       return entity;
--- a/packages/app/src/providers/h5grove/h5grove-api.ts
+++ b/packages/app/src/providers/h5grove/h5grove-api.ts
@@ -2,11 +2,12 @@ import type {
   ArrayShape,
   Attribute,
   AttributeValues,
+  ChildEntity,
   Dataset,
   Entity,
   Group,
-  GroupWithChildren,
   NumericType,
+  ProvidedEntity,
 } from '@h5web/shared';
 import { hasScalarShape, buildEntityPath, EntityKind } from '@h5web/shared';
 import type { AxiosRequestConfig } from 'axios';
@@ -40,7 +41,7 @@ export class H5GroveApi extends DataProviderApi {
     super(filepath, { baseURL: url, ...axiosConfig });
   }
 
-  public async getEntity(path: string): Promise<Entity> {
+  public async getEntity(path: string): Promise<ProvidedEntity> {
     const response = await this.fetchEntity(path);
     return this.processEntityResponse(path, response);
   }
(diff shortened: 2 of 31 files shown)


packages/app/src/providers/DataProvider.tsx
```diff
@@ -43,25 +43,23 @@ function DataProvider(props: PropsWithChildren<Props>) {
...
+      const cachedEntity = childCache.get(path);
+      if (cachedEntity) {
+        return cachedEntity;
       }
 
       const entity = await api.getEntity(path);
 
       if (isGroup(entity)) {
-        // Make sure `getEntity` doesn't return groups without `children` proprety
-        assertGroupWithChildren(entity);
```
axelboc, line 57: `isGroup` is now sufficient to guarantee `GroupWithChildren`.

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -149,37 +150,47 @@ export class H5GroveApi extends DataProviderApi {
...
-    const { name } = response;
+    response: H5GroveEntityResponse,
+    isChild: true
+  ): Promise<ChildEntity>;
+
+  private async processEntityResponse(
+    path: string,
+    response: H5GroveEntityResponse,
+    isChild?: false
+  ): Promise<ProvidedEntity>;
```
axelboc, line 161: I have to add an `isChild` parameter so I can declare signature overrides like in the other provider's APIs. The overrides ensure that when calling `processEntityResponse` without the `isChild` argument in `getEntity`, the return type is `ProvidedEntity` and not just `Entity`.

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -149,37 +150,47 @@ export class H5GroveApi extends DataProviderApi {
...
       };
 
-      if (!children) {
-        /* `/meta` stops at one nesting level
-         * (i.e. children of child groups are not returned) */
+      if (isChild) {
         return baseGroup;
       }
 
-      const groupWithChildren: GroupWithChildren = {
```
axelboc, line 173: Changing the return type to `Promise<ProvidedEntity | ChildEntity>` means we no longer need those intermediary variables (or any type casts).

packages/app/src/providers/hsds/models.ts
```diff
@@ -66,6 +66,11 @@ interface HsdsSymbolicLink {
 /* ------------------- */
 /* ----- ENTITIES----- */
 
+export type BaseHsdsEntity = Pick<
+  HsdsEntity,
+  'id' | 'collection' | 'path' | 'name'
+>;
+
```
axelboc, line 73: This is just to simplify the signatures of all the `process*` methods.

packages/app/src/visualizer/utils.ts
```diff
@@ -73,12 +72,11 @@ function getNxDefaultPath(
       : buildEntityPath(entity.path, defaultPath);
   }
 
-  assertGroupWithChildren(entity);
```
axelboc, line 76: `isGroup` does the job

packages/shared/src/mock/utils.ts
```diff
@@ -31,22 +38,27 @@ export function assertMockAttribute<S extends Shape, T extends DType>(
...
 
-  const pathSegments = path.slice(1).split('/');
-  return pathSegments.reduce<Entity | undefined>(
-    (parentEntity, currSegment) => {
-      return parentEntity && isGroup(parentEntity) && hasChildren(parentEntity)
-        ? getChildEntity(parentEntity, currSegment)
-        : undefined;
-    },
-    mockMetadata
-  );
```
axelboc, line 49: I was struggling to fix the types in this `reduce`, so I refactored it with a recursion. The algorithm is similar to what is done in `HsdsApi.getEntity()`.

packages/shared/src/mock/utils.ts
```diff
@@ -31,22 +38,27 @@ export function assertMockAttribute<S extends Shape, T extends DType>(
...
+  if (!parent) {
+    return undefined;
+  }
+
+  assertGroup(parent);
+  const childName = path.slice(path.lastIndexOf('/') + 1);
+  const child = parent && getChildEntity(parent, childName);
+
+  if (child && isGroup(child)) {
+    assertGroupWithChildren(child);
```
axelboc, line 59: `getChildEntity` has no way to know that it's dealing with mock metadata where all groups have children, so we have to assert. This is the only place left where we need to call `assertGroupWithChildren`, but I've simplified it to accept a group instead of an entity, so it's now only checking for the presence of `children`.

packages/app/src/providers/DataProvider.tsx
```diff
@@ -43,25 +43,23 @@ function DataProvider(props: PropsWithChildren<Props>) {
   const { api, children } = props;
 
   const entitiesStore = useMemo(() => {
-    const childCache = new Map<string, Entity>();
+    const childCache = new Map<string, Exclude<ChildEntity, Group>>();
```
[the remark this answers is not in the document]
loichuder: Wait, why don't we store group metadata ? Does this mean that we need to fetch again when asking for a group path ?

loichuder (commented):
I am a bit torn. On one hand, I see that this brings a lot of simplifications at several places. On the other hand, having multiple entity models like `Entity`, `ProvidedEntity` and `ChildEntity` makes things a bit confusing. Do we really need `Entity` now ? Could we rename `ProvidedEntity` into `Entity` ?

Also, it enforces future providers to follow our line of thought where group metadata contains only one level of metadata making it cumbersome to implement providers where the metadata that contain every level (like in `Mock`). It can be a boon or a pain.

packages/app/src/providers/DataProvider.tsx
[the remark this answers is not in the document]
axelboc: Groups (with children) are stored in the `entitiesStore`'s internal cache. We don't want to store child groups without children in the `childCache`, because if we did, we would end up with groups without children when calling `entitiesStore.get()`... and we definitely don't want that.

axelboc:
> Also, it enforces future providers to follow our line of thought where group metadata contains only one level of metadata making it cumbersome to implement providers where the metadata that contain every level (like in Mock). It can be a boon or a pain.

While I think that the mock provider is an exception and that no other back-end implementations would eagerly include multiple nesting levels, I do agree that this is an unjustified restriction. I think I can try to remove the concept of a `ChildEntity` without losing most of the benefits intended by this PR.

> I am a bit torn. On one hand, I see that this brings a lot of simplifications at several places. On the other hand, having multiple entity models like Entity, ProvidedEntity and ChildEntity makes things a bit confusing. Do we really need Entity now ? Could we rename ProvidedEntity into Entity ?

I hesitated with renaming `Entity` to `BaseEntity` and `ProvidedEntity` to `Entity`, but I didn't like the idea because it would imply that a `Group` without children is not an `Entity`. We use the `Entity` model a lot for utilities and guards that don't care which kinds of entities they're dealing with, and it didn't seem right to use `BaseEntity` for those.

Is the new `ProvidedEntity` still problematic if I remove the `ChildEntity` model?

loichuder (approved):
After discussion, looks good to me

PR #1183, opened 2022-08-08
Add toggle group to change matrix notation

### LINKED ISSUE #1175: [Matrix] Allow other numeric styles than scientific notation

No

Allow the display of the raw data in something other than scientific notation.

loichuder:
Fix #1175 

There are now three display modes, corresponding to [d3-format](https://github.com/d3/d3-format) specifiers :
- **Scientific** (default up to now, corresponds to d3-format `e`)
image

- **Auto**  (`g`)
image

- **Fixed-point** with 3 decimals (2 for complex, `f`)
image

As shown in the screenshots, the change in notation also applies for fields of compound datasets (complex included). For now, there is no way to tune the number of digits, this may come later if needed.


packages/app/src/vis-packs/core/compound/utils.ts
```diff
@@ -1,14 +1,21 @@
...
+  notation: Notation
+) {
   if (Array.isArray(value)) {
-    return formatMatrixComplex(value);
+    return createMatrixComplexFormatter(notation)(value);
   }
 
   if (typeof value === 'number') {
-    return formatMatrixValue(value);
+    return createNumericFormatter(notation)(value);
```
loichuder, line 18: I can't use the updated `getFormatter` as I have access to the `value` here and not to the `dataset`.

packages/lib/src/toolbar/controls/ToggleGroup.tsx
```diff
@@ -66,14 +66,15 @@ function Btn(props: BtnProps) {
...
 }
 
 function ToggleGroup(props: Props) {
-  const { role, ariaLabel, value, disabled, onChange, children } = props;
+  const { role, ariaLabel, value, disabled, onChange, children, id } = props;
 
   return (
     <ToggleGroupContext.Provider value={{ role, value, disabled, onChange }}>
-      <div className={styles.group} role={role} aria-label={ariaLabel}>
+      <div className={styles.group} role={role} aria-label={ariaLabel} id={id}>
```
loichuder, line 77: Small change to allow an `id` prop to be used for `htmlFor` on the `label` of the `ToggleGroup`. This follows the pattern of the cell width control.

packages/app/src/vis-packs/core/compound/utils.ts
axelboc: You're recreating a formatter for every `value` here, though. You should write a `createCompoundFormatter(notation)` function, initialise all the formatters you may need once, and then return a formatter arrow function that uses all these formatters.

That being said, if I remember correctly, the source of the problem is that the `formatter` only receives a `value`, and since we're dealing with a compound dataset, we can't know what the type of that value will be ahead of time.

Could `MatrixVis` maybe pass the `col` index to the `formatter`? Wouldn't this be enough information for `MappedCompoundMatrixVis` to find the correct sub-type for that column? You could then refactor `getFormatter` to accept a `DType` rather than a `Dataset`.

packages/app/src/vis-packs/core/matrix/config.tsx
```diff
@@ -24,10 +28,15 @@ function createStore() {
         setCustomCellWidth: (customCellWidth) => {
           set({ customCellWidth });
         },
+
+        notation: Notation.Auto,
```
axelboc, line 32: Should we keep scientific as the default? This is our first user feedback on this matter, so not sure if other users wouldn't prefer keeping the current behaviour. 🤷

packages/lib/src/toolbar/controls/NotationToggleGroup.tsx
```diff
@@ -0,0 +1,34 @@
...
+const ID = 'h5w-matrix-notation';
+
+function NotationToggleGroup(props: Props) {
+  const { value, onChange } = props;
+
+  return (
+    <div className={styles.root}>
+      <label className={styles.label} htmlFor={ID}>
+        Notation
+      </label>
```
axelboc, line 19: Should the label be a feature of `ToggleGroup` like `Selector`? It can wait until we need a label on another toggle group, though.

axelboc (changes requested):

packages/app/src/vis-packs/core/matrix/config.tsx
loichuder: Good call :+1:

packages/lib/src/toolbar/controls/NotationToggleGroup.tsx
loichuder: Agreed.

loichuder:
Thanks for the great comments. Please have another look

packages/app/src/vis-packs/core/compound/MappedCompoundMatrixVis.tsx
```diff
@@ -44,6 +43,10 @@ function MappedCompoundMatrixVis(props: Props) {
     slicedMapping
   );
 
+  const fieldFormatters = Object.values(fields).map((field) =>
+    getFormatter(field, notation)
+  );
+
```
axelboc, line 49: Nicely done 👏

axelboc (approved):

PR #1184, opened 2022-08-08
Fix stacking context of floating toolbar

loichuder:
| Before | After |
| --- | --- |
| image | image |


axelboc (changes requested):
You also need to create a stacking context, perhaps on `canvasWrapper` so the `z-index` of the floating toolbar does not "leak" to the document's stacking context.

You can replicate this (potential) issue by setting a higher `z-index` value to the floating toolbar, opening the overflow menu of the main toolbar,  and resizing the window so the overflow menu appears under the reset-zoom button.

By setting `z-index: 0` on `.canvasWrapper` (and thus creating a new stacking context, since it already has `position: relative`), the risk of a future stacking problem disappears.

axelboc (approved):

PR #1186, opened 2022-08-09
Fix onSelectionChange not called in SelectionTool

loichuder:
The `onSelectionChange` callback of `SelectionTool` is not called when the mouse moves (see https://h5web-docs.panosc.eu/?path=/story/building-blocks-selection--rectangle where the title should update when dragging a rectangle).

This is a quick fix as a discussion with @axelboc showed that a bigger refactoring could be envisionned to distinguish checks for starting the interaction and checks for continuing the interaction on pointer moves.

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -72,14 +72,14 @@ function SelectionTool(props: Props) {
         return;
       }
 
-      const { worldPt, sourceEvent } = evt;
+      const { worldPt } = evt;
       const dataEndPoint = worldToData(boundPointToFOV(worldPt, camera));
       setEndPoint(dataEndPoint);
 
       const worldEndPoint = dataToWorld(dataEndPoint);
       const worldStartPoint = dataToWorld(startPoint);
 
-      if (onSelectionChange && shouldInteract(sourceEvent)) {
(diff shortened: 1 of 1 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -72,14 +72,14 @@ function SelectionTool(props: Props) {
...
-      const { worldPt, sourceEvent } = evt;
+      const { worldPt } = evt;
       const dataEndPoint = worldToData(boundPointToFOV(worldPt, camera));
       setEndPoint(dataEndPoint);
 
       const worldEndPoint = dataToWorld(dataEndPoint);
       const worldStartPoint = dataToWorld(startPoint);
 
-      if (onSelectionChange && shouldInteract(sourceEvent)) {
+      if (onSelectionChange && isModifierKeyPressed) {
```
loichuder, line 82: `shouldInteract` checks that the `sourceEvent.button` is equal to the interaction button (in this case `MouseButton.Left = 0`). However, for pointer move events, [`sourceEvent.button` is equal to `-1`](https://www.w3.org/TR/pointerevents/#the-button-property) when the button didn't change which makes the check fail. 

The quick fix consists in checking only the modifier key instead of the key and the button (strategy already used in `Pan`).

axelboc (approved):

PR #1187, opened 2022-08-10
Upgrade linting config and make sure JS set-up files are fully linted

axelboc:
Previously some core rules like [`no-undef`](https://eslint.org/docs/latest/rules/no-undef) were not being reported on JS files like `.eslintrc.js`. This is now fixed thanks to a new galex option called `enableJavaScriptSpecificRulesInTypeScriptProject` https://github.com/ljosberinn/eslint-config-galex/issues/662.


apps/demo/package.json
```diff
@@ -33,12 +33,13 @@
     "react-router-dom": "6.3.0"
   },
   "devDependencies": {
+    "@types/node": "16.11.7",
```
axelboc, line 36: Now that the `.eslintrc.js` files are linted, we need to install Node's types package so galex knows to enable ESLint's `node` environment (which includes global variables such as `module`, `require`, etc.)

eslint.shared.js
```diff
@@ -9,91 +9,100 @@ const {
   createTypeScriptOverride,
 } = require('eslint-config-galex/dist/overrides/typescript');
 
-const JS_RULES = {
```
axelboc, line 12: I'm reverting to inlining the rules objects in the `createConfig` call to get better type safety.

eslint.shared.js
```diff
@@ -9,91 +9,100 @@ const {
...
-  'jest/no-conditional-in-test': 'off', // false positives in E2E tests (snapshots), and too strict (disallows using `||` for convenience)
-  'testing-library/no-unnecessary-act': 'off', // `act` is sometimes required when advancing timers manually
-};
-
 module.exports = {
   createConfig: (cwd, dependencies, customOverrides = []) => {
     return createConfig({
       cwd,
-      rules: JS_RULES,
+      env: { es6: true }, // https://github.com/ljosberinn/eslint-config-galex/pull/664
```
axelboc, line 16: Enabling ESLint's `es6` environment so that global variables such as `Promise`, `Set`, etc. can used in JS files. We use `Set` in some `vite.config.js` files. I've opened a PR to fix this in galex: https://github.com/ljosberinn/eslint-config-galex/pull/664

loichuder (approved):
Mr. Galex seems very quick to answer, that's great :smile:

PR #1188, opened 2022-08-10
Lint and type-check all set-up files

axelboc:
`.eslintrc.(c)js`, `jest.config.js`, `vite.config.js`, etc.

### DIFF
--- a/apps/demo/package.json
+++ b/apps/demo/package.json
@@ -18,7 +18,7 @@
     "build": "vite build",
     "serve": "vite preview --port 3000",
     "lint:eslint": "eslint \"**/*.{js,cjs,ts,tsx}\" --max-warnings=0",
-    "lint:tsc": "tsc --noEmit"
+    "lint:tsc": "tsc"
   },
   "dependencies": {
     "@h5web/app": "workspace:*",
--- a/eslint.shared.js
+++ b/eslint.shared.js
@@ -14,7 +14,7 @@ module.exports = {
     return createConfig({
       cwd,
       env: { es6: true }, // https://github.com/ljosberinn/eslint-config-galex/pull/664
-      enableJavaScriptSpecificRulesInTypeScriptProject: true, // to lint `.eslintrc.js` files and the like
+      enableJavaScriptSpecificRulesInTypeScriptProject: true, // to fully lint set-up files like `.eslintrc.js`, `vite.config.js`, etc.
       rules: {
         'sort-keys-fix/sort-keys-fix': 'off', // keys should be sorted based on significance
         'import/no-default-export': 'off', // default exports are common in React
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -18,7 +18,7 @@
     "build": "build-storybook -o build",
     "serve": "pnpx http-server -p 6006 ./build",
     "lint:eslint": "eslint \"**/*.{js,cjs,ts,tsx}\" --max-warnings=0",
-    "lint:tsc": "tsc --noEmit"
+    "lint:tsc": "tsc"
   },
   "dependencies": {
     "@h5web/lib": "workspace:*",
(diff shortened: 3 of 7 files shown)


apps/demo/package.json
```diff
@@ -18,7 +18,7 @@
     "build": "vite build",
     "serve": "vite preview --port 3000",
     "lint:eslint": "eslint \"**/*.{js,cjs,ts,tsx}\" --max-warnings=0",
-    "lint:tsc": "tsc --noEmit"
+    "lint:tsc": "tsc"
```
axelboc, line 21: Already specified in root `tsconfig.json`.

apps/demo/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["src"]
+  "include": ["src", "*.js", ".*.cjs"]
```
axelboc, line 3: Dot files have to be specified explicitly. Advanced glob patterns like `{,.}*.?(c)js` are not supported, unfortunately.

eslint.shared.js
```diff
@@ -10,15 +10,22 @@ const {
...
+   * @param {string} cwd - pass `__dirname` so galex can find the project's `package.json`
+   * @param {import('eslint-config-galex/dist/types').Dependencies} dependencies - dependencies object returned by `getDependencies()`
+   * @param {(import('eslint').Linter.ConfigOverride | null)[]} customOverrides
+   * @returns {import('eslint-config-galex/dist/types').TopLevelESLintConfig}
+   */
   createConfig: (cwd, dependencies, customOverrides = []) => {
     return createConfig({
       cwd,
       env: { es6: true }, // https://github.com/ljosberinn/eslint-config-galex/pull/664
-      enableJavaScriptSpecificRulesInTypeScriptProject: true, // to lint `.eslintrc.js` files and the like
```
axelboc, line 17: Galex provides two options for fully linting JS files in a TypeScript project:

- using the property above, or
- enabling `checkJs` in `tsconfig.json`

I've decided to enable `checkJs` in this repo, which has the added benefit that we can use `tsc` to type check JS/CJS files (once you include them in `tsconfig.json`).

The `enableJavaScriptSpecificRulesInTypeScriptProject` property is more for JS codebases that are migrating to TypeScript (like daiquiri).

package.json
```diff
@@ -26,6 +26,7 @@
     "lint:tsc": "pnpm -r --parallel lint:tsc",
     "lint:cypress:tsc": "tsc --project cypress/tsconfig.json",
     "lint:root:eslint": "eslint \"**/*.{js,cjs,ts,tsx}\" --max-warnings=0",
+    "lint:root:tsc": "tsc --project tsconfig.root.json",
```
axelboc, line 29: Now type-checking JS files at the root of the project.

packages/app/vite.config.js
```diff
@@ -6,7 +6,9 @@ import { defineConfig } from 'vite';
 const dirname = fileURLToPath(new URL('.', import.meta.url));
 
 const [pkg, libPkg, sharedPkg] = ['.', '../lib', '../shared'].map((prefix) =>
-  JSON.parse(fs.readFileSync(path.resolve(dirname, `${prefix}/package.json`)))
+  JSON.parse(
+    fs.readFileSync(path.resolve(dirname, `${prefix}/package.json`), 'utf-8')
```
axelboc, line 10: The only linting / type-checking errors reported were on this line in a few of the `vite.config.js` files. TS wants `JSON.parse` to receive a string, but eslint-plugin-unicorn says it works fine with a buffer... I decided to go with TS' recommendation, since we disable the ESLint rule is a more targetted solution than `@ts-ignore` or `as string`.

axelboc:
Jest did not play well with `checkJs` and I couldn't figure out why. So, I'm giving up on this :cry: 

At least JS files are still fully linted thanks to `enableJavaScriptSpecificRulesInTypeScriptProject`.

PR #1190, opened 2022-08-16
Allow to disable dark mode of App

loichuder:
While working to disable the dark mode in daiquiri, I had the thought that if the switch to dark mode is sometimes undesirable, there should be a way to disable it.

This would avoid the [CSS hack in jupyterlab-h5web](https://github.com/silx-kit/jupyterlab-h5web/blob/8543f67bb93bc6cb19198d5233d1df2c0779758a/style/index.css#L20).

Here is a proposal based on a `disableDarkMode` prop of `App` that gets translated in a `allow-dark-mode` CSS attribute so that CSS dark mode rules can be filtered. I am not convinced this is the best solution so I am drafting it for discussion.

### DIFF
--- a/packages/app/src/App.module.css
+++ b/packages/app/src/App.module.css
@@ -51,6 +51,9 @@
   --h5w-domainSlider-histogram--color: var(--secondary-dark);
   --h5w-domainSlider-histogram-indicator--color: var(--secondary-darker);
   --h5w-interactionHelp-shortcut--bgColor: var(--primary-light);
+
+  --h5w-line--color: darkblue;
+  --h5w-line--colorAux: orangered, forestgreen, red, mediumorchid, olive;
 }
 
 .root *,
@@ -72,7 +75,7 @@
 }
 
 @media (prefers-color-scheme: dark) {
-  .root {
+  .root[data-allow-dark-mode] {
     filter: invert();
 
     /* Set shadows to lighter colors so they remain dark once inverted */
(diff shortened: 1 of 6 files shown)


packages/app/src/App.module.css
```diff
@@ -51,6 +51,9 @@
   --h5w-domainSlider-histogram--color: var(--secondary-dark);
   --h5w-domainSlider-histogram-indicator--color: var(--secondary-darker);
   --h5w-interactionHelp-shortcut--bgColor: var(--primary-light);
+
+  --h5w-line--color: darkblue;
+  --h5w-line--colorAux: orangered, forestgreen, red, mediumorchid, olive;
```
loichuder, line 56: I am reverting part of https://github.com/silx-kit/h5web/pull/1008 by setting back the line color in the app.

In fact, I would argue that the `useCustomColors` hook with dark fallbacks implemented in #1008 should be removed. The dark mode switch is handled at the `app` level while the hook works at the `lib` level.

I have the opinion that the vis (that is `lib`) should be completely oblivious to dark mode as it should be the responsibility of `App` (or of the `App` consumer) to implement it if needed. Further implementations deeper in the hierarchy are not difficult to discover/debug and get in the way of the global handling.

axelboc (commented):
> This would avoid the [CSS hack in jupyterlab-h5web](https://github.com/silx-kit/jupyterlab-h5web/blob/8543f67bb93bc6cb19198d5233d1df2c0779758a/style/index.css#L20).

Awesome!!

packages/app/src/App.module.css
loichuder: Given, we still need to use a hook to get the CSS values in JS, I think we can leave the fallback in JS via `useCustomColors`.

I will remove the `darkFallbackk` though as discussed.

axelboc (approved):
Do you feel like trying to move the fallback to CSS in another PR? Otherwise, I can open an issue.

loichuder:
> Do you feel like trying to move the fallback to CSS in another PR? Otherwise, I can open an issue.

Sure, I can do it while we are at it

PR #1192, opened 2022-08-17
Fix string dataset handling in h5wasm

loichuder:


packages/h5wasm/src/guards.ts
```diff
@@ -51,7 +51,7 @@ export function isNumericMetadata(
 }
 
 export function isStringMetadata(metadata: Metadata) {
-  return metadata.type === 2;
+  return metadata.type === 3;
```
loichuder, line 54: String datasets were given the type `Unknown` as I made a mistake when reading the [type enum](https://github.com/usnistgov/h5wasm/blob/fb407d0c7b2d366485832910d3b43ce70ae6d15e/src/hdf5_util_helpers.d.ts#L10)...

Whoops !

packages/h5wasm/src/utils.ts
```diff
@@ -45,11 +45,13 @@ export function convertMetadataToDType(metadata: Metadata): DType {
...
   if (isStringMetadata(metadata)) {
-    const { size: length, cset } = metadata;
+    const { size, cset, vlen } = metadata;
 
     return {
       class: DTypeClass.String,
-      length,
+      // For variable-length string datatypes, the returned value is the size of the pointer to the actual string and
+      // not the size of actual variable-length string data (https://portal.hdfgroup.org/display/HDF5/H5T_GET_SIZE)
+      length: vlen ? undefined : size,
```
loichuder, line 54: After the fix above, I was wondering why all my string datasets were interpreted as 4-character long. Then, I saw that they were in fact of variable-length, meaning that the `size` field did not reflect the actual string length.

axelboc (approved):

Issue #1177, opened 2022-07-26
Picking API


vallsv:
Now a plot can contain many displayed items, including data arrays.

On my actual use case, the plot contains at least
- Several tiles
- A heatmap

It would be good to provide a helpers for picking data from the data coords.

To retrieve:
- the data as a specific data coord
- the data index at a specific data coord

The plot already have the full knowledge of the transformations. Which can be very complex for items like tiling. So i would prefer not to compute such thing on my side.

The aim for now is to feed a tooltip.

axelboc:
I've done some deeper investigating on this in the past few days.

We are currently limited by the fact that we've opted out of React Three Fiber's events system due to a bug with pointer capture (cf. #994 and #995). The `onPointerUp` event was not always called when clicking on the canvas and releasing the mouse outside of it, leading to the pointer not being released (thus breaking interactions like `Pan` or `SelectToZoom`).

It is my hope that this bug is no longer present in the latest version of R3F. So once we upgrade R3F (as part of the React 18) upgrade, we can try removing `useCanvasEvents` and re-opting-in to R3F's events system instead (with `<mesh onPointerMove />`).

Assuming the pointer capture bug is resolved, opting in to R3F's events system should make implementing a Picking API much more straightforward:

1. The tooltip component registers a `pointermove` handler on a mesh spanning the canvas; thanks to R3F's events system, **all the events received now include an array of intersected Three.js objects**.
2. Components that want to contribute values to the tooltip can somehow (most likely with a zustand store injected via context) register themselves to the tooltip by providing:
   - a reference to their underlying Three.js object;
   - a callback function that accepts data coordinates and returns the value to display in the tooltip.
3. In the `pointermove` handler, the tooltip component computes the data coordinates with `worldToData(evt.unprojectedPoint)` and then loops through the intersected Three.js objects; if it finds one that is registered, it calls the corresponding callback function with the data coordinates; all the values returned by the callbacks are kept in an array.
4. The tooltip component calls a render prop with this array so the consumer can decide how to display them.

Issue #988, opened 2022-03-02
Reduce the number of tiles rendered by TiledHeatmapMesh


t20100:
### Is your feature request related to a problem?

As it is, the `TiledHeatmapMesh` renders the current subsampled image layer according to the resolution, but also all lower layers: in case some tiles are not loaded yet, this "fills" the holes with lower resolution.

### Requested solution or feature

This has two issues:
- It implies to render more tiles than needed
- The `TiledHeatmap` spreads over multiple layer in depth which can be cumbersome when adding more components in the `VisCanvas` (e.g., regions of interest shapes or a second image with transparency on top of it).

It should be possible to achieve the same "progressive" display with only one layer of tiles: The tiles that are not loaded yet can display the corresponding sub-region in the available tiles at lower resolution.
This probably needs to adapt `HeatmapMesh` to pass texture coordinates in order to select the sub-region

PR #1193, opened 2022-08-18
Added tooltip for TiledHeatmapMesh

loichuder:
Related to #1177 

Tried an analytical method as using the raycaster was not convienient at all. Indeed, as the whole layer group is intersected by the ray, we still need to filter manually the tiles. Perhaps I missed something so I could revisit this later.

The tooltip is not working properly for two cases:
- with flipped axes
- with transforms

I am fairly confident I can fix it for flipped axes but we can already discuss the current implementation.


apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -131,6 +131,53 @@ class MandelbrotTilesApi extends TilesApi {
   }
 }
 
+class CheckboardTilesApi extends TilesApi {
```
loichuder, line 134: For early tests, I used tiles showing a checkboard pattern so I kept it as a story.

I seem to remember that you had something similar at one point so don't hesitate to replace/correct it.

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
```diff
@@ -74,25 +76,32 @@ function TiledHeatmapMesh(props: Props) {
...
+            api={api}
+            layer={layer}
+            meshSize={meshSize}
+            visibleBox={visibleBox}
+            {...colorMapProps}
+          />
+        ))}
+      </group>
+      {showTooltip && tooltipLayer !== undefined && (
+        <TiledTooltipMesh layer={tooltipLayer} api={api} meshSize={meshSize} />
```
loichuder, line 102: For now the tooltip component is part of `TiledHeatmapMesh` but is relevant only for one layer that is given in prop.

We could imagine putting `TiledTooltipMesh` in the layer instead but then we should make sure to display only the tooltip of the above layer

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -88,11 +88,7 @@ export function getObject3DVisibleBox(
   return NDC_BOX.clone().applyMatrix4(ndcToObject3DMatrix);
 }
 
-export function scaleToLayer(box: Box3, layerSize: Size, meshSize: Size): Box2 {
-  if (box.isEmpty()) {
-    return new Box2();
-  }
-
+export function getLayerScales(layerSize: Size, meshSize: Size) {
```
loichuder, line 91: As I needed to convert the `world` coordinates in `layer` coordinates, I extracted the scale generation in this utility

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -105,8 +101,34 @@ export function scaleToLayer(box: Box3, layerSize: Size, meshSize: Size): Box2 {
...
+
+  const { xScale, yScale } = getLayerScales(layerSize, meshSize);
+
   return new Box2(
     new Vector2(xScale(box.min.x), yScale(box.min.y)),
     new Vector2(xScale(box.max.x), yScale(box.max.y))
   );
 }
+
+export function scaleFromLayerToTile(layerCoord: Vector2, tileSize: Size) {
```
loichuder, line 120: New util needed for the tooltip: from the position of the mouse in `layer` coordinates (`layerCoord`), we want to know what is the underlying tile offset and the position of the mouse in this tile.

apps/storybook/src/TiledHeatmapMesh.stories.tsx
t20100: Yes, I did the same at first!
At that time, I didn't bother to use a store, I don't think it is needed.

packages/lib/src/vis/tiles/TiledTooltipMesh.tsx
```diff
@@ -0,0 +1,69 @@
...
+      Math.floor(yScale(worldVec.y))
+    );
+  }
+
+  return (
+    <TooltipMesh
+      renderTooltip={(x, y) => {
+        return (
+          <div style={{ display: 'flex', flexDirection: 'column' }}>
+            {`x=${formatTooltipVal(x)}, y=${formatTooltipVal(y)}`}
```
t20100, line 39: I don't get what this is for?

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
```diff
@@ -74,25 +76,32 @@ function TiledHeatmapMesh(props: Props) {
 
   const { colorMap, invertColorMap = false } = colorMapProps;
 
+  const tooltipLayer = layers[layers.length - 1];
```
t20100, line 79: Shouldn't the `tooltipLayer` be the `currentLayerIndex`?

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
t20100: Not sure, as it is, you want a single tooltip at a time.
And to me, it sounds better in the end to have it outside of the `TiledHeatmapMesh` in order to have a single tooltip displayed over multiple meshes, but not for this PR.

t20100 (commented):
As we already discussed, flipped axes support is needed and transforms support would be nice to have but not needed for now.

To me in the end, the tooltip mesh will need to be handled outside of the `TiledHeatmapMesh` because it should handle more than one mesh, but that's better to go step by step and it's fine for me like it is in this PR.

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
loichuder: In essence, this is `currentLayerIndex` but you are right. It would be better to be explicit.

apps/storybook/src/TiledHeatmapMesh.stories.tsx
loichuder: Fine, I removed the store :+1:

loichuder:
> To me in the end, the tooltip mesh will need to be handled outside of the TiledHeatmapMesh because it should handle more than one mesh

Agreed. But it means the variables computed in `TiledHeatmapMesh` (`tooltipLayer` and `meshSize`) should somehow be available outside of it

apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -152,11 +172,7 @@ const Template: Story<TiledHeatmapStoryProps> = (args) => {
       <Zoom />
       <SelectToZoom keepRatio modifierKey="Control" />
       <ResetZoomButton />
-      <group
-        scale={[abscissaConfig.flip ? -1 : 1, ordinateConfig.flip ? -1 : 1, 1]}
-      >
```
loichuder, line 157: The `flip` is now handled at the layer level to avoid flipping the `TiledTooltipMesh`

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
```diff
@@ -46,9 +50,11 @@ function TiledHeatmapMesh(props: Props) {
   );
   const visibleBox = getObject3DVisibleBox(ndcToLocalMatrix);
 
-  const bounds = scaleToLayer(visibleBox, baseLayerSize, meshSize);
+  const { xScale, yScale } = useLayerScales(baseLayerSize, meshSize);
```
loichuder, line 53: Scales are now computed by a hook outside of `scaleToLayer` (renamed `scaleBox`) to be able to access the `AxisContext` and take the flip in account

t20100:
My bad, transforms (scale and position) also need to be supported, and flipped axes handled outside of the `TiledHeatmaComponent`.

axelboc:
> > To me in the end, the tooltip mesh will need to be handled outside of the TiledHeatmapMesh because it should handle more than one mesh
> 
> Agreed. But it means the variables computed in `TiledHeatmapMesh` (`tooltipLayer` and `meshSize`) should somehow be available outside of it

IMO, the most important is that the rendering of the tooltip content be the responsibility of the consumer, so there has to be a `renderTooltip` prop somewhere, and this prop should accept a function that gets called with all the necessary info.

`renderTooltip` can be a prop of `TiledHeatmapMesh` for now. If the prop is not set then the `TiledHeatmapMesh` doesn't render a `TooltipMesh` at all.

You can of course help consumers by providing a component to use inside `renderTooltip` - e.g. `TiledTooltip` (pretty much like `TiledTooltipMesh` but without the `TooltipMesh`): `renderTooltip={tooltipProps => <TiledTooltip {...tooltipProps} />}

In the long term, maybe #988 will bring out some refactoring opportunities to allow having a fully separate `TiledTooltipMesh` component, I don't know... A custom hook for consumers to use together with `TiledHeatmapMesh` and `TiledTooltipMesh` might be an option...

loichuder:
As the support of `transform` is needed, I will need to rework this. I am drafting it in the mean time.

loichuder:
Now handling transforms. I may revert the refactoring of `scaleToLayer` as the new implementation does not need it.

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
```diff
@@ -75,24 +80,28 @@ function TiledHeatmapMesh(props: Props) {
...
+            api={api}
+            layer={layer}
+            meshSize={meshSize}
+            visibleBox={visibleBox}
+            topLayer={i === layers.length - 1}
+            {...colorMapProps}
+          />
+        ))}
+      </group>
+      {showTooltip && <TiledTooltipMesh size={meshSize} />}
```
loichuder, line 103: This is to make the size of `TooltipMesh` match the size of the `TiledHeatmap`

packages/lib/src/vis/tiles/TiledTooltipMesh.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+function TiledTooltipMesh(props: Props) {
+  const { size } = props;
+  const val = useTooltipStore((state) => state.val);
+
+  return (
+    <TooltipMesh
+      size={size}
+      renderTooltip={() => {
```
loichuder, line 18: `renderTooltip` could be exposed

packages/lib/src/vis/tiles/TiledLayer.tsx
```diff
@@ -1,57 +1,93 @@
...
+    props;
 
   const { baseLayerIndex, numLayers, tileSize } = api;
   const layerSize = api.layerSizes[layer];
+  const { xScale, yScale } = useLayerScales(layerSize, meshSize);
+  const { abscissaConfig, ordinateConfig, worldToData } =
+    useAxisSystemContext();
+  const setTooltipValue = useTooltipStore((state) => state.setTooltipValue);
+
+  const onPointerMove = useCallback(
```
axelboc, line 34: If I'm not mistaken, I think you could bring this callback up into `TiledHeatmapMesh` and pass it down as prop only to the top layer.

packages/lib/src/vis/tiles/utils.ts
axelboc: Is this still used?

axelboc (approved):
I reckon that the tooltip store is taking us really close to a shared solution for both the heatmap and the tiled heatmap, where the `TooltipMesh` would remain outside of/decoupled from the `(Tiled)HeatmapMesh` component. This has great potential: it could allow, for instance, consumers of the lib replace the tooltip with a status bar, a bit like on vector graphics software:

image

axelboc:
> Now handling transforms. I may revert the refactoring of scaleToLayer as the new implementation does not need it.

No need, it's clear what's what, and the refactoring is nice 👍

packages/lib/src/vis/tiles/utils.ts
```diff
@@ -88,25 +86,33 @@ export function getObject3DVisibleBox(
...
-  });
+  const pt1 = new Vector2(xScale(box.min.x), yScale(box.min.y));
+  const pt2 = new Vector2(xScale(box.max.x), yScale(box.max.y));
 
-  const yScale = createAxisScale(ScaleType.Linear, {
-    domain: [-meshSize.height / 2, meshSize.height / 2],
-    range: [0, layerSize.height],
-    clamp: true,
-  });
+  return new Box2(pt1.clone().min(pt2), pt1.clone().max(pt2));
```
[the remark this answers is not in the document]
loichuder: Great find! :detective:

PR #1194, opened 2022-08-29
Update peer dep three of h5web/lib

loichuder:
Oversight of https://github.com/silx-kit/h5web/pull/1170

Since we are dependant of the changes introduced in `138` (e.g. renaming from `DataTexture3D` to `Data3DTexture`), the version of `three` in `peerDependencies` should have been updated.


axelboc (approved):
Woops, nice one!

PR #1202, opened 2022-09-06
Upgrade Vite to v3 and remove Windows workarounds

axelboc:
Release notes:

- https://github.com/vitejs/vite/blob/v3.1.0/packages/vite/CHANGELOG.md#main-changes
- https://github.com/vitejs/vite/blob/main/packages/plugin-react/CHANGELOG.md
- https://github.com/gxmari007/vite-plugin-eslint/blob/main/CHANGELOG.md

Biggest win is that Vite 3.1.0 includes an upgrade of `postcss-modules` to v5, which fixes the issue with `composes` on Windows! https://github.com/vitejs/vite/issues/3092

Lots of great stuff in Vite@3 under the hood. Only thing that may be of interest is that it now only supports Firefox v78 and above out of the box. There's a legacy plugin, but I don't think it's worth using for the demo. Unless I'm mistaken, this is irrelevant for library mode since we generate only an ESM bundle and let consumers bundle it in their own app. We'll have to publish a beta to double check.

### DIFF
--- a/apps/demo/package.json
+++ b/apps/demo/package.json
@@ -37,11 +37,11 @@
     "@types/react": "17.0.39",
     "@types/react-dom": "17.0.11",
     "@types/react-router-dom": "5.3.3",
-    "@vitejs/plugin-react": "1.3.2",
+    "@vitejs/plugin-react": "2.1.0",
     "eslint": "8.21.0",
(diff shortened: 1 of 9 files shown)


loichuder (approved):
Seems good to me. Don't forget to test with the beta :wink:

PR #1204, opened 2022-09-12
Document browser support

axelboc:


packages/app/README.md
```diff
@@ -62,6 +62,22 @@ various front-end development stacks:
...
+
+H5Web works out of the box on **Firefox 78 ESR**.
+
+Support for Firefox 68 ESR is possible by polyfilling the `ResizeObserver` API.
+One easy way to do this is with [polyfill.io](https://polyfill.io/v3/):
+
+```html
+<head>
+  <!-- title, meta, link, etc. -->
+  <script src="https://polyfill.io/v3/polyfill.min.js?features=default%2CResizeObserver"></script>
```
axelboc, line 75: I went with `polyfill.io` in the end, as it's the simplest solution. It adds a request, but the response is empty if everything is already available. Polyfilling manually with `@juggle/resize-observer` or the like is more difficult to do right.

.github/workflows/approve-snapshots.yml
```diff
@@ -55,6 +55,7 @@ jobs:
       - name: Run Cypress 🌳
         run: pnpm cypress:run --env updateSnapshots=true
         env:
+          CYPRESS_BASE_URL: http://localhost:3000
```
loichuder, line 58: Note we may instead chose to `serve` the app at port `5173` to have a set-up closer to what we had before

packages/app/README.md
loichuder: I think that's great :+1:

packages/app/README.md
```diff
@@ -62,6 +62,22 @@ various front-end development stacks:
 - [Create React App v4](https://codesandbox.io/s/h5webapp-cra-v4-0y1lj)
 - [Create React App v5](https://codesandbox.io/s/h5webapp-cra-v5-bzmbh1)
 
+### Browser support
+
+H5Web works out of the box on **Firefox 78 ESR**.
+
+Support for Firefox 68 ESR is possible by polyfilling the `ResizeObserver` API.
+One easy way to do this is with [polyfill.io](https://polyfill.io/v3/):
```
loichuder, line 70:
```suggestion
[polyfill.io](https://polyfill.io/v3/) is an easy way to do this:
```

packages/lib/README.md
```diff
@@ -72,3 +72,19 @@ various front-end development stacks:
...
+One easy way to do this is with [polyfill.io](https://polyfill.io/v3/):
+
+```html
+<head>
+  <!-- title, meta, link, etc. -->
+  <script src="https://polyfill.io/v3/polyfill.min.js?features=default%2CResizeObserver"></script>
+</head>
+```
+
+Older versions of Firefox are not supported.
```
loichuder, line 90: Might be good to talk about other browsers ?

Well, we can keep it like this for now

loichuder (approved):
Whoops, forgot to change the review status

packages/lib/README.md
loichuder: Agreed. I had the same line of thought when reading this section

PR #1205, opened 2022-09-12
Set line color feedback in CSS

loichuder:
https://github.com/silx-kit/h5web/pull/1190#pullrequestreview-1103713685

### DIFF
--- a/packages/lib/src/vis/hooks.ts
+++ b/packages/lib/src/vis/hooks.ts
@@ -6,7 +6,6 @@ import { useEffect, useCallback, useMemo, useState } from 'react';
 import type { RefCallback } from 'react';
 import { createMemo } from 'react-use';
 
-import type { CustomColor } from './models';
 import { useAxisSystemContext } from './shared/AxisSystemProvider';
 import type { AxisSystemContextValue } from './shared/AxisSystemProvider';
 import {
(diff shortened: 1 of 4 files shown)


packages/lib/src/vis/hooks.ts
```diff
@@ -75,7 +74,7 @@ export function useCameraState<T>(
 }
 
 export function useCustomColors(
```
loichuder, line 76: The hook could be renamed to something else: `useColorCssVariable` ?

packages/lib/src/vis/line/LineVis.module.css
```diff
@@ -4,6 +4,15 @@
...
   margin: 1rem;
+  --curve-color: var(--h5w-line--color, darkblue);
+  --aux-colors: var(
+    --h5w-line--colorAux,
+    orangered,
+    forestgreen,
+    red,
+    mediumorchid,
+    olive
+  );
```
loichuder, line 15: Following the suggestion in https://github.com/silx-kit/h5web/pull/1190#discussion_r962925905

packages/lib/src/vis/hooks.ts
axelboc: `useCssColors`?

axelboc (approved):
Nice! Glad it works :+1:

PR #1207, opened 2022-09-14
Make auxiliary labels optional in LineVis

loichuder:
Even though labels are always set when calling `LineVis` from `NxSpectrum`, it would be nice to lift this constraint for other uses (such as the example https://codesandbox.io/s/h5weblib-demo-multiple-curves-kwkli)

### DIFF
--- a/packages/lib/src/vis/line/LineVis.tsx
+++ b/packages/lib/src/vis/line/LineVis.tsx
@@ -73,8 +73,8 @@ function LineVis(props: Props) {
 
   assertLength(abscissaValue, dataArray.size, 'abscissa');
   assertLength(errorsArray, dataArray.size, 'error');
-  auxiliaries.forEach(({ label, array }) =>
-    assertLength(array, dataArray.size, `'${label}' auxiliary`)
+  auxiliaries.forEach(({ label, array }, index) =>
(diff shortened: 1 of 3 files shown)


packages/lib/src/vis/line/LineVis.tsx
```diff
@@ -153,12 +153,14 @@ function LineVis(props: Props) {
...
-                  <div className={styles.tooltipAux} key={label}>
+                  <div className={styles.tooltipAux} key={label || index}>
                     <span
                       className={styles.mark}
                       style={{ color: auxColors[index % auxColors.length] }}
                     />
-                    {label} = {formatTooltipVal(array.get(xi))}
+                    {`${label ? `${label} =` : ''}${formatTooltipVal(
+                      array.get(xi)
+                    )}`}
```
axelboc, line 163: You might be able to keep this as two separate JSX expressions (without a space between the two brackets `}{`) 🤷

axelboc (approved):
Good call 💯

packages/lib/src/vis/line/LineVis.tsx
loichuder: True, it looks ugly here

PR #1211, opened 2022-09-27
Allow propagating errors through top-level error boundary

axelboc:
For consumers who want to implement their own error boundary, like in https://gitlab.esrf.fr/ui/myhdf5/-/merge_requests/11

I think this is more flexible than to allow consumers to override the top-level error fallback component, for instance.

Note that I also make `react-error-boundary` an optional peer dependency to reduce bundle size for consumers who do want to provide their own boundary.

### DIFF
--- a/packages/app/README.md
+++ b/packages/app/README.md
@@ -140,6 +140,34 @@ form instead.
 />
 ```
 
+#### `disableDarkMode?: boolean` (optional)
+
+By default, the viewer follows your browser's and/or operating system's dark
+mode setting. This prop disables this beahviour by forcing the viewer into light
+mode.
+
+```tsx
+<App disableDarkMode />
+```
+
+#### `propagateErrors?: boolean` (optional)
+
+The viewer has a top-level `ErrorBoundary` that, by default, handles errors
+thrown outside of the visualization area. These include errors thrown by the
+data provider when fetching metadata for the explorer. If you prefer to
+implement your own error boundary, you may choose to let errors through the
+viewer's top-level boundary:
+
+```tsx
+import { ErrorBoundary } from 'react-error-boundary';
+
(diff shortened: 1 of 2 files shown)


packages/app/README.md
```diff
@@ -140,6 +140,34 @@ form instead.
...
 
+#### `disableDarkMode?: boolean` (optional)
+
+By default, the viewer follows your browser's and/or operating system's dark
+mode setting. This prop disables this beahviour by forcing the viewer into light
+mode.
+
+```tsx
+<App disableDarkMode />
+```
```
loichuder, line 151: OOPS

packages/app/package.json
```diff
@@ -39,7 +39,13 @@
   },
   "peerDependencies": {
     "react": ">=16",
-    "react-dom": ">=16"
+    "react-dom": ">=16",
+    "react-error-boundary": "3.1.4"
+  },
+  "peerDependenciesMeta": {
+    "react-error-boundary": {
+      "optional": true
```
loichuder, line 47: Is `react-error-boundary` optional given the re-throw of errors is done in `onError` of `ErrorBoundary` ?

axelboc: It's optional for the consumer, not for the codebase, of that makes sense.

axelboc: Yeah, no, this doesn't work at all. I have a feeling that the combination [ `peerDependencies` + `{ optional: true }` ] is similar to putting `react-error-boundary` under `optionalDependencies`, in the sense that `@h5web/app` should work regardless of whether `react-error-boundary` is installed or not, which is not what I was after here.

FYI, the difference with `optionalDependencies` is that the package manager tries to install `optionalDependencies` but does not error out if their installation fails. Whereas the package manager does not try to install `peerDependencies` at all.

I'll revert this change.

PR #1216, opened 2022-10-04
Support axis mapping of scale types in `SILX_style`

### LINKED ISSUE #1214: SILX_style axis scales ignored for 2/3D+ arrays
Looking at `getSilxStyle`, I think we actually have a bug there, as we disregard `axisScaleTypes` arrays that contain `undefined` values.

_Originally posted by @axelboc in https://github.com/silx-kit/h5web/pull/1213#discussion_r984435054_

axelboc:
Fix #1214

Most basic fix to let the `axisScaleTypes` array be parsed if it contains `null` values (or `'.'` or whatever -- anything else than a valid scale type is parsed as `undefined`).

This lets the spectrum visualization use the suggested axis scale type for the selected dimension on 2D+ datasets. As you pointed out in #1214, @loichuder, this does not fix the issue of updating the suggested scale type when selecting another dimension. But we agreed this isn't worth fixing until we actually encounter the use case.

### DIFF
--- a/packages/app/src/vis-packs/nexus/models.ts
+++ b/packages/app/src/vis-packs/nexus/models.ts
@@ -1,6 +1,7 @@
 import type {
   ArrayShape,
   ArrayValue,
+  AxisMapping,
   ComplexType,
   Dataset,
   NumArray,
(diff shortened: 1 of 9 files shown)


packages/app/src/vis-packs/nexus/models.ts
```diff
@@ -48,8 +49,6 @@ export interface NxValues<T extends NumericType | ComplexType> {
   title: string;
 }
 
-export type AxisMapping<T> = (T | undefined)[];
```
axelboc, line 51: Moved to the shared package so I can use it in `SilxStyle`.

loichuder (approved):

PR #1220, opened 2022-10-07
Display scalar values as preformatted text

### LINKED ISSUE #1217: [Scalar] Special characters in string are replaced with spaces
### Describe the bug

Original report by @mretegan: multiline strings (containing `\n`) are displayed as a single line. `\n` is indeed replaced by a space.

In fact, it appears that even `\t` is replaced by a single space

### To Reproduce

1. Go to https://h5web.panosc.eu/h5grove?file=grove.h5
2. Click on `scalars/multiline_string`. This dataset contains the string `"This is.\nA multiline.\nString.\nWith a tab:\t."`
3. The multiline string is displayed on a single line and there is a single space between `tab:` and `.`

### Expected behaviour

Special characters should be interpreted correctly (carriage-return, tab, ...)

### Screenshots

For comparison, here is how `silx view` displays the same dataset:

image

loichuder:
Fix #1217

image

### DIFF
--- a/packages/lib/src/vis/scalar/ScalarVis.module.css
+++ b/packages/lib/src/vis/scalar/ScalarVis.module.css
@@ -1,4 +1,5 @@
 .scalar {
+  margin: 0;
   padding: 1rem 1.5rem;
   color: inherit;
   font-family: var(--monospace);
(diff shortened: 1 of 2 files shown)


axelboc (changes requested):
I think there's just a tweak to be done when the content overflows (especially horizontally). There's currently `overflow: auto` on the `mainArea`, or rather on `reflex-container > .reflex-element`. I think we could cancel this react-reflex style with `overflow: visible`, and apply `overflow: auto` specifically to the explorer and to the `visArea`. This would solve overflowing for the raw vis too probably.

Issue #1173, opened 2022-07-19
Add keyboard support to the explorer tree


andygotz:
Is it possible to implement browsing the Nexus tree with the cursor and/or mouse wheel and have the display updated?

I find myself doing this naturally so maybe others do this too.

This is a nice-to-have feature request.

Thanks for considering this.

Andy

loichuder:
@andygotz I reformulated the title of the issue. Please ensure that this was the feature you wanted to request.

For the record, the tree in `silx view` can be navigated with the keyboard but not with the mouse wheel.

loichuder:
I wonder if we should switch to something like https://github.com/lissitz/react-accessible-treeview instead of reimplementing keyboard handlers ourselves.

What do you think @axelboc ?

axelboc:
I had looked at a bunch of libraries at the time, and always had problems (data structure too restrictive, lack of customisation capabilities, missing/unwanted behaviours ...) The fact that an open group doesn't collapse when you first select it is quite a specific feature, for instance. Some libraries also had trouble dealing with dynamic data.

`react-accessible-treeview` has some of these problems and was also left unmaintained for a couple of years until recently. I would be concerned that the recent overhaul of the codebase introduced regressions.

![image](https://user-images.githubusercontent.com/2936402/183437675-d6cda246-3f42-4220-8038-0df53a9c6b1c.png)

That being said, maybe some tree libraries have matured and are worth having another look, so if you feel like doing a POC, please don't hesitate.

Either way, for this feature specifically, I think we should discuss the behaviour in more details before we do anything (e.g. does `ArrowDown` automatically select the next entity or do you have to press `Enter` ; do groups expand/collapse automatically in some cases, etc.)

For the implementation, this will be a good reference: https://www.w3.org/WAI/ARIA/apg/example-index/treeview/treeview-navigation

loichuder:
After a deeper dive, I understand your arguments against external libraries.

I made a quick prototype by myself that we can discuss together.

loichuder:
Here are the envisioned supported features and their status:

- [x] Support for Enter/Space (native)
- [ ] Support for Up/Down https://github.com/silx-kit/h5web/pull/1221
- [ ] Support for Left/Right
- [ ] Support for Home/End

PR #1221, opened 2022-10-07
Add support for ArrowDown/ArrowUp in explorer 

loichuder:
For https://github.com/silx-kit/h5web/issues/1173

Selection with `Enter/Space` is already there as the entries are clickable `button` elements.


packages/app/src/explorer/EntityListContainer.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }, []);
+
+  return (
+    <div
+      ref={ref}
+      onKeyDown={(evt) => {
+        const { key } = evt;
+
+        if (key === 'ArrowDown') {
+          focusOnNext();
```
axelboc, line 50: I'd rename those `focusNext` and `focusPrevious` and return from each `if` block. Not needed but cleaner I think,in case the complexity grows.

packages/app/src/explorer/Explorer.tsx
```diff
@@ -35,12 +35,7 @@ function Explorer(props: Props) {
           />
         }
       >
-        <EntityList
-          level={0}
-          parentPath="/"
-          selectedPath={selectedPath}
-          onSelect={onSelect}
-        />
+        <EntityListContainer selectedPath={selectedPath} onSelect={onSelect} />
```
axelboc, line 38: Composition would be very useful here:

```tsx
<KeyboardController>
  <EntityList {...} />
</KeyboardController>
```

packages/app/src/explorer/utils.ts
```diff
@@ -26,3 +26,22 @@ export function needsNxBadge(
...
 }
+
+export function getActiveButtonIndex(
+  element: Element | null
+): [number, HTMLButtonElement[]] | undefined {
+  if (!element) {
+    return undefined;
+  }
+
+  const activeButton = document.activeElement;
```
axelboc, line 37: What happens when the viewer renders for the first time? Is `activeElement` defined? If not, then perhaps it should default to the first button (i.e. the button of the root entity)?

packages/app/src/explorer/utils.ts
```diff
@@ -26,3 +26,22 @@ export function needsNxBadge(
...
+
+  if (!activeButton || !(activeButton instanceof HTMLButtonElement)) {
```
axelboc, line 39: This probably needs to be a bit stricter, since there are other buttons on the UI. Perhaps `element.contains(activeButton)`, or you could get the list of buttons first and see if it's inside. Same same.

packages/app/src/explorer/utils.ts
```diff
@@ -26,3 +26,22 @@ export function needsNxBadge(
...
+    return undefined;
+  }
+
+  const buttonElementList = [...element.getElementsByTagName('button')]; // eslint-disable-line unicorn/prefer-query-selector
```
axelboc, line 43: If you're not taking advantage of the live `NodeList` (which is probably a good thing), then there's no point in using `getElementsByTagName` over `querySelectorAll`.

packages/app/src/explorer/EntityListContainer.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    const [i, buttonElementList] = res;
+    if (i < 0 || i === buttonElementList.length - 1) {
+      return;
+    }
+    buttonElementList[i + 1].focus();
+  }, []);
+
+  const focusOnPrevious = useCallback(() => {
+    const res = getActiveButtonIndex(ref.current);
```
axelboc, line 31: Returning an array with the index of the button and the list of buttons may not be the simplest abstraction. I feel like this could be simpler if you just had a function `getActiveButton` that would return the active element if it is an explorer button, else `undefined`. The rest can be inlined (listing all the buttons, finding the index of the active one, and focusing the next/previous one) unless a better abstraction is found.

axelboc (commented):
Awesome that it works completely independently! :tada:  First round of feedback.

packages/app/src/explorer/utils.ts
loichuder: `activeElement` is indeed `undefined` at the first render. But I don't think it causes a problem: the callback can only be called once the focus is on a the Explorer, that is to say, one element is focused.

In this case, `activeButton` will always be defined. Unless I missed something ? :thinking:

axelboc: Ah right, you put the `keydown` listener on the explorer. Well, in that case, I would suggest moving it up to the `document` so that people can interact with the keyboard to navigate the explorer when the app first renders, and more generally any time they want.

We just have to be careful not to interfere with the keyboard interactions of other UI elements, like the slider, and basically anything else that can be interacted with (drop-down menus, button groups maybe (to be checked), etc.)

axelboc: I think a condition like `!activeElement || explorerRoot.contains(activeElement)` in the key handlers for the explorer should take care of avoiding those interference in a generic way.

packages/app/src/explorer/hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+    const activeIndex = getActiveIndex(buttonList, ref.current);
+
+    if (
+      activeIndex === undefined ||
+      activeIndex < 0 ||
+      activeIndex === buttonList.length - 1
+    ) {
+      return;
+    }
+    buttonList[activeIndex + 1].focus();
```
axelboc, line 23: Maybe invert the condition to avoid the return (same below)?

packages/app/src/explorer/hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+        evt.preventDefault();
+      }
+    },
+    [focusNext, focusPrevious]
+  );
+
+  useEffect(() => {
+    document.addEventListener('keydown', onKeyDown);
+    return () => document.removeEventListener('keydown', onKeyDown);
+  }, [onKeyDown]);
```
axelboc, line 61: You can use @rect-hookz/web's `useAddEventListener`

packages/app/src/explorer/utils.ts
```diff
@@ -26,3 +26,28 @@ export function needsNxBadge(
...
+  buttonList: HTMLButtonElement[],
+  element: Element
+): number | undefined {
+  const { activeElement } = document;
+
+  if (
+    activeElement instanceof HTMLButtonElement &&
+    element.contains(activeElement)
+  ) {
+    return buttonList.lastIndexOf(activeElement);
```
axelboc, line 40: I think `indexOf` would suffice here, since we don't expect the same button instance to appear multiple times (not technically possible). Same below :point_down:

packages/app/src/explorer/utils.ts
```diff
@@ -26,3 +26,28 @@ export function needsNxBadge(
...
+  }
+
+  if (activeElement === document.body || activeElement === undefined) {
+    const selectedElement = element.querySelector<HTMLButtonElement>(
```
axelboc, line 44: `selectedBtn`?

axelboc (approved):

packages/app/src/explorer/utils.ts
loichuder: Still not short enough to fit on one line :smile:

packages/app/src/explorer/hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+    const buttonList = [...ref.current.querySelectorAll('button')];
+    const activeIndex = buttonList.indexOf(activeElement);
+
+    if (activeIndex > 0) {
+      buttonList[activeIndex - 1].focus();
+    }
+  }, [ref]);
+
+  useKeyboardEvent('ArrowUp', focusPrevious);
+  useKeyboardEvent('ArrowDown', focusNext);
```
axelboc, line 48: Hehe so now, you can actually register the listeners on the explorer with `{ target: ref }`:

```ts
useKeyboardEvent('ArrowUp', focusPrevious, [], { target: ref });
```

axelboc (approved):
Awesome 👍

Issue #1198, opened 2022-09-05
[Heatmap] Don't keep ratio when units don't match


axelboc:
### Is your feature request related to a problem?

See https://github.com/silx-kit/silx/issues/3626

> Working with 2d datasets in NeXus hdf5 with different units on the x- and y-axis I was wondering if there is a way to tell silx view to open a NXdata image view with the option "Do not keep data aspect ration" activated? By default it seems to keep the aspect ratio.

### Requested solution or feature

The `layout` config option currently always defaults to `cover` (i.e. keep ratio). Instead, it could default to `fill` (i.e. don't keep ratio) when the axes have different units.

### Alternatives you've considered

`SILX_style` could be extended with an `aspect` option -- see https://github.com/silx-kit/silx/issues/3626#issuecomment-1228566872.

### Additional context

We'll have to review the way we persist the `layout` configuration option, or whether it still needs to be persisted at all, as a dynamic default is incompatible with persistence across reloads.

t20100:
I made a dummy file with some `NXdata` groups with different cases regarding `@units`: http://www.silx.org/pub/h5web/test_units.h5

PR #1224, opened 2022-10-10
Get all units and labels as part of `useNxData`

axelboc:
In preparation for #1198, labels and units are now retrieved once and for all when calling `useNxData`. We no longer compute labels/retrieves units from `NxValuesFetcher`. This will allow us to access units inside the container in order to pass an initial `layout` value to the heatmap config hook.

### DIFF
--- a/packages/app/src/vis-packs/core/hooks.ts
+++ b/packages/app/src/vis-packs/core/hooks.ts
@@ -54,17 +54,28 @@ export function useDatasetValue<D extends Dataset<ArrayShape | ScalarShape>>(
 export function useDatasetValues<D extends Dataset<ArrayShape | ScalarShape>>(
   datasets: D[],
   selection?: string
-): Record<string, Value<D>> {
+): Value<D>[];
+
+export function useDatasetValues<D extends Dataset<ArrayShape | ScalarShape>>(
+  datasets: (D | undefined)[],
+  selection?: string
+): (Value<D> | undefined)[];
+
+export function useDatasetValues<D extends Dataset<ArrayShape | ScalarShape>>(
+  datasets: (D | undefined)[],
+  selection?: string
+): (Value<D> | undefined)[] {
   const { valuesStore } = useDataContext();
 
-  return Object.fromEntries(
-    datasets.map((dataset) => {
-      const value = valuesStore.get({ dataset, selection });
-      assertDatasetValue(value, dataset);
+  return datasets.map((dataset) => {
+    if (!dataset) {
+      return undefined;
+    }
 
-      return [dataset.name, value];
-    })
-  );
+    const value = valuesStore.get({ dataset, selection });
+    assertDatasetValue(value, dataset);
+    return value;
+  });
 }
 
 export function useBaseArray<T, U extends T[] | TypedArray | undefined>(
--- a/packages/app/src/vis-packs/core/line/MappedLineVis.tsx
+++ b/packages/app/src/vis-packs/core/line/MappedLineVis.tsx
@@ -9,7 +9,6 @@ import type {
 import { createPortal } from 'react-dom';
 
 import type { DimensionMapping } from '../../../dimension-mapper/models';
-import type { Auxiliary } from '../../nexus/models';
 import {
   useMappedArrays,
   useMappedArray,
@@ -26,7 +25,9 @@ interface Props {
   value: NumArray;
   valueLabel?: string;
   errors?: NumArray;
-  auxiliaries?: Auxiliary[];
+  auxLabels?: string[];
+  auxValues?: NumArray[];
+  auxErrors?: (NumArray | undefined)[];
   dims: number[];
   dimMapping: DimensionMapping;
   axisLabels?: AxisMapping<string>;
@@ -43,7 +44,9 @@ function MappedLineVis(props: Props) {
     value,
     valueLabel,
     errors,
-    auxiliaries = [],
+    auxLabels = [],
+    auxValues = [],
+    auxErrors = [],
     dims,
     dimMapping,
     axisLabels,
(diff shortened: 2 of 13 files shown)


packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -54,17 +54,28 @@ export function useDatasetValue<D extends Dataset<ArrayShape | ScalarShape>>(
...
+
+export function useDatasetValues<D extends Dataset<ArrayShape | ScalarShape>>(
+  datasets: (D | undefined)[],
+  selection?: string
+): (Value<D> | undefined)[];
+
+export function useDatasetValues<D extends Dataset<ArrayShape | ScalarShape>>(
+  datasets: (D | undefined)[],
+  selection?: string
+): (Value<D> | undefined)[] {
```
axelboc, line 67: These override allow simplifying the code in `NxValuesFetcher` and getting rid of a couple of custom hooks. We no longer return the datasets' values mapped against their names; instead, we allow `undefined` values arrays in the returned array (in the same place as `undefined` datasets in the input array). The overrides ensure that if we pass an array without `undefined` values, we get a return type without `undefined` values as well.

packages/app/src/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -26,7 +25,9 @@ interface Props {
   value: NumArray;
   valueLabel?: string;
   errors?: NumArray;
-  auxiliaries?: Auxiliary[];
+  auxLabels?: string[];
+  auxValues?: NumArray[];
+  auxErrors?: (NumArray | undefined)[];
```
axelboc, line 30: Passing everything in separate arrays for consistency with axes values/labels. I'm still not hugely satisfied with the mapped components' APIs but by clearly separating values from metadata in the containers, we're a step closer to seeing a good refactoring, I reckon.

packages/app/src/vis-packs/nexus/models.ts
```diff
@@ -23,39 +23,43 @@ export type NxAttribute =
...
-  errorDataset?: NumArrayDataset;
-  axisDatasets: AxisMapping<NumArrayDataset>;
-  axisLabels: AxisMapping<string>;
-  auxDatasets: AuxDatasets;
   titleDataset?: Dataset<ScalarShape, StringType>;
+  signalDef: WithError<DatasetDef<T>>;
+  auxDefs: AuxDef[];
+  axisDefs: AxisMapping<DatasetDef<NumericType>>;
   silxStyle: SilxStyle;
 }
```
axelboc, line 56: The key changes are in this file.

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -230,31 +230,31 @@ export function getSilxStyle(
...
+  const longName =
+    rawLongName && typeof rawLongName === 'string' ? rawLongName : undefined;
+
+  const rawUnits = attrValuesStore.getSingle(dataset, 'units');
+  const units = rawUnits && typeof rawUnits === 'string' ? rawUnits : undefined;
+
+  return {
+    label: longName || (units ? `${dataset.name} (${units})` : dataset.name),
+    unit: units,
+  };
```
axelboc, line 247: Seemed silly/convoluted to retrieve the units first and pass them to `getDatasetLabel` or to get the units twice or make a function for that, so I made one function for retrieving both info.

axelboc:
I broke the aux errors. To be continued.

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
```diff
@@ -15,36 +16,32 @@ interface Props<T extends NumericType | ComplexType> {
...
+  const errors = useDatasetValue(signalDef.errorDataset, selection);
+  const auxValues = useDatasetValues(auxDatasets, selection);
+  const auxErrors = useDatasetValues(auxErrorDatasets, selection);
+  const axisValues = useDatasetValues(axisDatasets);
 
   return (
-    <>
-      {render({ signal, signalLabel, errors, axisValues, auxiliaries, title })}
-    </>
+    <>{render({ title, signal, errors, auxValues, auxErrors, axisValues })}</>
```
loichuder, line 44: Is there a reason for the change in order ? I liked the fact that the variables were sorted by importance

packages/app/src/vis-packs/nexus/models.ts
```diff
@@ -23,39 +23,43 @@ export type NxAttribute =
...
+  T extends NumericType | ComplexType = NumericType | ComplexType
+> extends DatasetInfo {
+  dataset: Dataset<ArrayShape, T>;
+}
+
+type WithError<T extends DatasetDef> = T & {
+  errorDataset?: NumArrayDataset;
+};
+
+export type AuxDef = WithError<DatasetDef<NumericType>>;
```
loichuder, line 41: The chained generic are not very readable in my opinion but I don't see an easy alternative here

loichuder (approved):
Great refactoring :+1:

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
axelboc: I thought this order was conveying importance better 😂 :

1. title of the plot because it tells you what you're looking at ;
2. signal and errors because that's the important plot
3. aux and aux errors because they're still data being plotted, so quite important
4. axis values because I felt they were less important than the rest

Axis stuff between signal and auxiliaries didn't feel right. I don't mind moving the `title` before `axisValues`, since it's a kind of metadata, but as I said, I think the title is actually the most important info of a NeXus plot.

packages/app/src/vis-packs/nexus/models.ts
axelboc: Yeah I hesitated with `DatasetWithErrorDef extends `DatasetDef` but thought the mixin-style type was actually a better pattern in this case. `DatasetWithErrorDef` would have required duplicating the `T extends NumericType | ComplexType = NumericType | ComplexType` generic.

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
loichuder: Yeah but `title` can be `undefined`, no ?

Anyway, do not let this stop you from merging, it is fine.

axelboc: Yeah you're right, I can move it down 👍

axelboc: Actually no, it's always defined since there's a fallback on `signalDef.label`. I'll leave it at the top.

PR #1225, opened 2022-10-11
Extract AxialSelectionTool and export it in lib

loichuder:
I think it is easier to export it rather than reimplementing `getAxialSelection` in daiquiri


axelboc (approved):
Good call! 💯

Issue #1199, opened 2022-09-05
Persistence of `SILX_style` options


axelboc:
### Is your feature request related to a problem?

When we select a dataset with a `SILX_style` attribute (e.g. `/nexus_entry/log_spectrum`) and then another dataset without (e.g. `/nexus_entry/spectrum_with_aux`), the axes scales of the first dataset are persisted.

I've always found this behaviour to be quite annoying. I understand that the point of `SILX_style` is to override any persisted configuration but the fact that the `SILX_style` config itself gets persisted doesn't feel right, since the user doesn't explicitly change the configuration from the toolbar.

I'm reporting this now, since we risk running into a similar issue in #1198 when reviewing the way we persist the `layout` option. If we decide to persist the computed `layout` for the current dataset whenever it differs from the saved `layout`, then we end up in the same situation as above with a configuration option that gets persisted without the user's knowledge/say so.

### Requested solution or feature

The `SILX_style` overrides could simply be applied to the current visualization without being persisted.

axelboc:
This is more challenging that I initially thought in our current architecture. The difficulty comes from the fact that we want the "preferred" scale type (for instance) to apply only until the user decides otherwise. So if the `SILX_style` says the preferred scale type is `log`, the visualization should use `log` by default, but as soon as the user changes the scale type manually, the preferred scale type should be ignored.

### Solution 1: `useState`

```tsx
// In mapped vis (code and variable names simplified for clarity)
const { preferredValue } = props;
const storedValue = useConfig(state => state.value);

const [hasChanged, toggleChanged] = useToggle(false);

// Where we need the value (in the mapped vis and the toolbar)
hasChanged ? storedValue : preferredValue;

// Where we need to change the stored value (in the toolbar)
const setValue= useConfig(state => state.setValue);
setValue(newValue);
toggleChanged();
```

`useToggle` and ternary could be abstracted into custom hook - e.g. `useUntilChanged`, and a monkey-patched `setValue` could be also passed down to the toolbar, but it still leaves a lot of things to wire up, variables to rename, long variable names, etc. And all of this wiring up has to be duplicated in other mapped vis components and for each value (x/y/colormap scale types).

### Solution 2: more props in store

Currently, we update the store via `useEffect` from the toolbar with the values parsed from `SILX_style`. Instead of setting the keys that are persisted (e.g. `scaleType`), we could set other keys that would not persisted (e.g. `preferredXcaleType`). We could then unset them when the persisted key is set.

```tsx
[...]

PR #1227, opened 2022-10-11
Choose best layout based on axis units

### LINKED ISSUE #1198: [Heatmap] Don't keep ratio when units don't match

axelboc:
Fix #1198 

In the mock `NXdata` group below, the axis units are different so the layout is now set to `fill` automatically when the group is selected (i.e. keep ratio disabled):

image

You can play around with the persistence behaviour by switching between a NeXus image and a heatmap visualization and toggling "keep ratio". It works as per `scaleType` and #1199.

There are more test cases in the file provided by @t20100: http://127.0.0.1:5173/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Ftest_units.h5. I did not bring any of them into our mock metadata, since we already have cases with no units as well as different units for the two axes, which is enough to test both conditional branches.

Also, I did not bother dealing with the `NxRGB` visualization since we don't pass axis values/labels to the mapped vis anyway.


loichuder (commented):
In my opinion, if one of the units is not defined, it means we cannot make an educated suggestion on the layout.

 I propose then the following behaviour
- One unit or more undefined: no suggestion (suggest `undefined`)
- Both units defined and **equal**: suggest `cover`
- Both units defined and **different**: suggest `fill`

t20100:
In silx we did it this way:

> Keep aspect ratio by default and disable it when axes have different @units (i.e., keep aspect ratio for axes with the same @units or without @units - if an axis as @units and the other not, it's different @units) https://github.com/silx-kit/silx/issues/3626#issuecomment-1236928823

https://github.com/silx-kit/silx/pull/3660

From what @loichuder propose, I would only make a change for the case of "one unit defined and the other not defined" and consider it as **different** units.
A case for this is a 2D dataset composed of many 1D dataset: rows are indices (unitless) and columns are the coordinates of the data.
And if there is no `@unit` at all, you keep using the current user choice.

axelboc:
Alright, thanks for your inputs. So in the end I'm just changing the case where neither units are defined to no longer enforce the `cover` layout. Now, it just defaults to the user's preference. I've abstracted this logic into a utility function with unit tests.

loichuder (approved):
Great example of refactorings done well: the final PR is quite short :clap:

PR #1232, opened 2022-10-17
Add support for left/right arrows in Explorer

loichuder:
For #1173 

Reminder of the expected behavior for Left/Right: https://www.w3.org/WAI/ARIA/apg/example-index/treeview/treeview-navigation

As this acts on the `expanded` state of the button, it means the event handler should have knowledge (or access) to the state of an entity item and its setter. As our keyboard event handler was up in the `Explorer`, refactoring was needed. 

I had the choice to either:
- Push the state of expanded nodes higher in the `Explorer` so that the hook `useKeyboardSupport` has access to it
- Attach event handlers where the state lives, that is in the `EntityItem` components.

I tried both methods and it seemed that the second one was cleaner and clearer so here we go.


packages/app/src/explorer/EntityItem.tsx
```diff
@@ -55,6 +57,37 @@ function EntityItem(props: Props) {
...
         }}
+        onKeyDown={(e) => {
+          switch (e.key) {
+            case 'ArrowDown':
+              focusNext(e);
+              return;
+
+            case 'ArrowUp':
+              focusPrevious(e);
+              return;
```
loichuder, line 68: `ArrowDown` and `ArrowUp` handlers were moved here as well so that all handlers are attached at the same place.

packages/app/src/explorer/Explorer.tsx
```diff
@@ -18,21 +18,25 @@ function Explorer(props: Props) {
...
 
-  useKeyboardSupport(ref);
-
   useEffect(() => {
     btnRef.current?.focus();
   }, [btnRef]);
 
   return (
-    <div ref={ref} className={styles.explorer} role="tree">
+    <div ref={ref} className={styles.explorer} role="tree" id={EXPLORER_ID}>
```
loichuder, line 26: To get the list of buttons in the Explorer, I need to have a ref to it. As the handlers are deeper, I attach an `id` here to be able to get the reference via `document.querySelector`. We could prop-drill the `ref` or use a context instead.

packages/app/src/explorer/utils.ts
```diff
@@ -26,3 +28,52 @@ export function needsNxBadge(
...
+  );
+  if (parentButton) {
+    parentButton.focus();
+    e.preventDefault();
+  }
+}
+
+export function focusNext(e: { preventDefault: () => void }) {
+  const { activeElement } = document;
+  if (!(activeElement instanceof HTMLButtonElement)) {
```
loichuder, line 55: I choose to no longer check here if `activeElement` is part of `Explorer` to be able to encapuslate the ref to `Explorer` in `getExplorerButtonList`.

Anyway, if it is not part of `Explorer`, `activeIndex` below will be set to `-1` and the focus will not change, giving the same result as before.

packages/app/src/explorer/utils.ts
```diff
@@ -26,3 +28,52 @@ export function needsNxBadge(
...
+  }
+  return [...explorer.querySelectorAll('button')];
+}
+
+export function focusParent(e: { preventDefault: () => void }, path: string) {
+  const buttonList = getExplorerButtonList();
+  const parentPath = path.slice(0, path.lastIndexOf('/')) || '/';
+
+  const parentButton = buttonList.find(
+    (element) => element.getAttribute('aria-label') === parentPath
```
loichuder, line 45: This is the reason why I attached the `path` as `aria-label` in the `EntityItem`. It seemed the easiest way to retrieve the parent button.

packages/app/src/explorer/EntityItem.tsx
```diff
@@ -47,6 +48,7 @@ function EntityItem(props: Props) {
         role="treeitem"
         aria-expanded={isGroup(entity) ? isExpanded : undefined}
         aria-selected={isSelected}
+        aria-label={path}
```
loichuder, line 51: Added to easily retrieve the parent button (for `ArrowLeft`). See below.

packages/app/src/explorer/EntityItem.tsx
```diff
@@ -55,6 +57,37 @@ function EntityItem(props: Props) {
...
+              focusNext(e);
+              return;
+
+            case 'ArrowUp':
+              focusPrevious(e);
+              return;
+
+            case 'ArrowLeft':
+              if (isGroup(entity) && isExpanded) {
+                toggleExpanded(false);
```
loichuder, line 72: I wonder if we should keep a `toggle` here given it is more used as a `set` now.

packages/app/src/explorer/EntityItem.tsx
```diff
@@ -55,6 +57,37 @@ function EntityItem(props: Props) {
...
+                e.preventDefault();
+              } else {
+                focusParent(e, path);
+              }
+              return;
+
+            case 'ArrowRight':
+              if (!isGroup(entity)) {
+                return;
```
loichuder, line 81: `ArrowRight` has no effect for end nodes

packages/app/src/explorer/EntityItem.tsx
```diff
@@ -55,6 +57,37 @@ function EntityItem(props: Props) {
 
           onSelect(path);
         }}
+        onKeyDown={(e) => {
```
axelboc, line 60: might be worth extracting this into a `handleKeyDown` function before the JSX

packages/app/src/explorer/EntityItem.tsx
```diff
@@ -55,6 +57,37 @@ function EntityItem(props: Props) {
...
+
+            case 'ArrowRight':
+              if (!isGroup(entity)) {
+                return;
+              }
+              if (isExpanded) {
+                focusNext(e);
+              } else {
+                toggleExpanded(true);
+                e.preventDefault();
```
axelboc, line 87: Move this before the `if/else` ?

packages/app/src/explorer/utils.ts
```diff
@@ -26,3 +28,52 @@ export function needsNxBadge(
...
+  }
+  return [...explorer.querySelectorAll('button')];
+}
+
+export function focusParent(e: { preventDefault: () => void }, path: string) {
+  const buttonList = getExplorerButtonList();
+  const parentPath = path.slice(0, path.lastIndexOf('/')) || '/';
+
+  const parentButton = buttonList.find(
+    (element) => element.dataset.path === parentPath
```
loichuder, line 45: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v43.0.2/docs/rules/prefer-dom-node-dataset.md

packages/app/src/explorer/EntityItem.tsx
loichuder: Now under `data-path`

packages/app/src/explorer/Explorer.tsx
axelboc: Neh, this is good. The coupling is clear thanks to the shared constant. 👍

packages/app/src/explorer/Explorer.tsx
```diff
@@ -18,21 +18,25 @@ function Explorer(props: Props) {
...
         type="button"
         role="treeitem"
         aria-selected={selectedPath === '/'}
+        aria-label="/"
         onClick={() => onSelect('/')}
+        onKeyDown={(e) => {
+          if (e.key === 'ArrowDown' || e.key === 'ArrowRight') {
+            focusNext(e);
+          }
+        }}
```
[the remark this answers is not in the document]
axelboc: `preventDefault`?

packages/app/src/explorer/utils.ts
```diff
@@ -26,3 +28,52 @@ export function needsNxBadge(
 
   return false;
 }
+
+function getExplorerButtonList(): HTMLButtonElement[] {
+  const explorer = document.querySelector(`#${EXPLORER_ID}`);
+  if (!explorer) {
+    return [];
+  }
+  return [...explorer.querySelectorAll('button')];
```
axelboc, line 37:
```suggestion
  return [...(explorer?.querySelectorAll('button') || [])];
```

packages/app/src/explorer/utils.ts
axelboc: Haha forgot about this API, thanks! Though given that we deal with other kinds of datasets, a small comment would maybe avoid confusion:

```suggestion
    (element) => element.dataset.path === parentPath // `data-path` attribute
```

packages/app/src/explorer/utils.ts
axelboc: Yep, makes sense, and if the button receives the `keydown` event, it means it's focused anyway, so you could just use `evt.currentTarget`.

axelboc (approved):

packages/app/src/explorer/EntityItem.tsx
axelboc: I like the toggle, it makes it clear it's a boolean flag.

packages/app/src/explorer/Explorer.tsx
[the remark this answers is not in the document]
loichuder: `preventDefault` is called by `focusNext`. I don't think we want to call it unless `focusNext` is called ?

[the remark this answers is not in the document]
axelboc: Ah sorry, no it's fine 👍

packages/app/src/explorer/utils.ts
loichuder: In this case, we can disable the rule then and use `getAttribute` instead

packages/app/src/explorer/EntityItem.tsx
loichuder: I wish to let `focusNext` handle `preventDefault` in the `if` branch. 

For example, if we reached the end of the tree and we press `Down`, the default behaviour (scrolling the explorer) should be applied.

axelboc: Good call 💯

PR #1233, opened 2022-10-17
Add support for Home/End key presses

### LINKED ISSUE #1173: Add keyboard support to the explorer tree

loichuder:
Last part: it fixes #1173 :tada: 

That was easier than the support of arrow keys.

### DIFF
--- a/packages/app/src/explorer/Explorer.tsx
+++ b/packages/app/src/explorer/Explorer.tsx
@@ -4,7 +4,7 @@ import { FiFileText, FiRefreshCw } from 'react-icons/fi';
 import { useDataContext } from '../providers/DataProvider';
 import EntityList from './EntityList';
 import styles from './Explorer.module.css';
-import { EXPLORER_ID, focusNext } from './utils';
+import { EXPLORER_ID, focusLast, focusNext } from './utils';
 
 interface Props {
   selectedPath: string;
@@ -33,6 +33,11 @@ function Explorer(props: Props) {
         data-path="/"
         onClick={() => onSelect('/')}
         onKeyDown={(e) => {
+          if (e.key === 'End') {
+            focusLast(e);
+            return;
+          }
+
           if (e.key === 'ArrowDown' || e.key === 'ArrowRight') {
             focusNext(e);
           }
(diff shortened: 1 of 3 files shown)


packages/app/src/explorer/utils.ts
```diff
@@ -74,3 +74,17 @@ export function focusPrevious(e: KeyboardEvent<HTMLButtonElement>) {
     e.preventDefault();
   }
 }
+
+export function focusFirst(e: KeyboardEvent) {
```
loichuder, line 78: For some reason, the types of `@react-hookz/web` do not allow an handler taking an event of type React's `KeyboardEvent`... I had to rename this type to `ReactKeyboardEvent` to avoid shadowing the native `KeyboardEvent` that is accepted by `@react-hookz/web`.

Perhaps worth opening an issue ? :shrug:

packages/app/src/explorer/utils.ts
```diff
@@ -74,3 +74,17 @@ export function focusPrevious(e: KeyboardEvent<HTMLButtonElement>) {
...
+  buttonList[0]?.focus();
+  e.preventDefault();
+}
+
+export function focusLast(e: KeyboardEvent) {
+  const buttonList = getExplorerButtonList();
+
+  buttonList[buttonList.length - 1]?.focus();
+  e.preventDefault();
+}
```
loichuder, line 90: As `focusFirst` and `focusLast` are called in the `Explorer`, I could have used `ref.current` but I preferred to follow the other `focus..` method template.

packages/app/src/explorer/utils.ts
loichuder: Given it is using `addEventListener` under the hood, I think the types are correct in fact.

packages/app/src/explorer/Explorer.tsx
```diff
@@ -22,6 +23,9 @@ function Explorer(props: Props) {
     btnRef.current?.focus();
   }, [btnRef]);
 
+  useKeyboardEvent('Home', focusFirst, [], { target: ref });
+  useKeyboardEvent('End', focusLast, [], { target: ref });
```
axelboc, line 27: Why not just put those in the buttons' event handler? It would keep everything together like before, and the event handler is already registered, so it'd just be adding a couple of `case`s to the `switch`.

axelboc (approved):

PR #1235, opened 2022-10-18
Fix stacking of domain slider tooltip

axelboc:
It is now always above everything, since it can be opened on hover - i.e. when other menus are open. Also, you can't pin it and keep it open while opening another menu, so there's no need for another menu to appear above the domain slider tooltip.

image

image

image


loichuder (approved):
Great ! Thanks !

PR #1236, opened 2022-10-20
Report type-checking errors in development

axelboc:
TS errors were only reported by `pnpm lint:tsc`. This is now corrected thanks to https://vite-plugin-checker.netlify.app/

I'm keeping `vite-plugin-eslint` for now, as the configuration is simpler. We still have the issue that ESLint errors in packages are not reported when developing the demo.

### DIFF
--- a/apps/demo/package.json
+++ b/apps/demo/package.json
@@ -42,6 +42,7 @@
     "eslint-config-galex": "4.2.2",
     "typescript": "4.8.3",
     "vite": "3.1.3",
+    "vite-plugin-checker": "0.5.1",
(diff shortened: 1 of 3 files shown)


loichuder:
Sorry, I don't really understand what this brings.

It adds TS checks when running `vite` and `vite build` ?

axelboc:
Yep, it reports them in the terminal and in a modal in the browser. We had actually lost that when moving to Vite. I thought I'd regoogle it and came across this plugin.

loichuder (approved):
Alrighty then !

Issue #1063, opened 2022-04-07
Allow client-generated exports of datasets


bmaranville:
### Is your feature request related to a problem?

Providers (e.g. h5wasm) that read the HDF5 file directly in the browser are unable to take advantage of server-side-generated exports of datasets or slices (URL-based)

### Requested solution or feature

Allow client-generated exports to be included in the menu in addition to backend-supported URL endpoints e.g. (lazy: Function | URL: string) instead of just URL: string

### Alternatives you've considered

Copying/pasting options discussed in #35 could be relevant, but exports would be nice.

### Additional context

The numpy target format (.NPY) is exceptionally well-documented and would be very easy to implement client-side, as well as CSV/TSV for < 3 dims.  TIFF is probably possible with a library for 2 dims.

axelboc:
@bmaranville, with the next release, you will be able to pass a [`getExportURL` prop](https://github.com/silx-kit/h5web/blob/main/packages/h5wasm/README.md#getexporturl-args--url----promiseurl--blob--undefined-optional) to `H5WasmProvider` to generate your own client-side exports. More info [here](https://github.com/silx-kit/h5web/blob/main/packages/app/README.md#getexporturl-args--url----promiseurl--blob--undefined-optional), along with a client-side export example.

The `H5WasmProvider` could definitely benefit from having a built-in implementation of `getExportURL`, even partial (only 1/2D numeric datasets to CSV, for instance), so please do share with us what you come up with.

bmaranville:
Sounds like you've done all the hard stuff... I'll definitely share any progress I make.

bmaranville:
Something is wrong with the types being given for getExportURL, and the example pseudocode can't even work:
value.forEach and value.map are both forbidden by the compiler, and clearly are needed for writing this function.  Here is the error:
```
[TypeScript] Property 'forEach' does not exist on type 'never'.
/Users/bbm/dev/h5web/apps/demo/src/h5wasm/H5WasmApp.tsx:31:15
    29 |         assert('forEach' in value);
    30 |         input = [];
  > 31 |         value.forEach((val) => input.push(val));
       |               ^^^^^^^
    32 |         // input = (value as Value[]).map((v: unknown) => [v]);
    33 |       }
    34 |       else if (dataset.shape?.length === 0) {
[TypeScript] Parameter 'val' implicitly has an 'any' type.
/Users/bbm/dev/h5web/apps/demo/src/h5wasm/H5WasmApp.tsx:31:24
[...]

PR #1237, opened 2022-10-20
Simplify props API of `ExportMenu`

axelboc:
Preparation for #1063 

The `getFormatURL`  prop was confusing. It made it look like the URLs were generated lazily, which isn't the case -- they're generated during render. Also, we used to end up with two arrays, `formats` and `urls`, so we had to access one by `index` in the `map` loop.

### DIFF
--- a/packages/app/src/vis-packs/core/heatmap/HeatmapToolbar.tsx
+++ b/packages/app/src/vis-packs/core/heatmap/HeatmapToolbar.tsx
@@ -50,6 +50,13 @@ function HeatmapToolbar(props: Props) {
 
   const { getExportURL } = useDataContext();
 
+  const exportEntries =
+    getExportURL &&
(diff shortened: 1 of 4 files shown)


loichuder (approved):
Nice

You could filter `exportEntries` to remove entries with `url = undefined` before passing it to `ExportMenu`. This would remove some checks in `ExportMenu` and avoid passing entries that are not relevant to the `ExportMenu`.

axelboc:
I had this at first, but it meant that the toolbars had to do the filtering, which felt like a bit too much unnecessarily duplication. It may evolve again, so I went for what felt like the simplest solution first.

loichuder:
> I had this at first, but it meant that the toolbars had to do the filtering, which felt like a bit too much unnecessarily duplication. It may evolve again, so I went for what felt like the simplest solution first.

Sure, no qualms about merging this as it is

PR #1238, opened 2022-10-21
Pass dataset value to `getExportURL`

axelboc:
Getting closer to #1063 

I pass the `value` of the dataset, which is available in the mapped vis, to the `getExportURL` method of the data provider. This will allow data providers to generate client-side exports without them having to call `valuesStore.get` (which is not doable outside of the `render` pipeline, as the exports will be generated on demand, when clicking on the `ExportMenu` options).

### DIFF
--- a/packages/app/src/providers/api.ts
+++ b/packages/app/src/providers/api.ts
@@ -4,8 +4,8 @@ import type {
   AttributeValues,
   Dataset,
   Entity,
-  NumericType,
   ProvidedEntity,
+  Value,
 } from '@h5web/shared';
 import type {
   AxiosInstance,
@@ -44,11 +44,12 @@ export abstract class DataProviderApi {
     this.client = axios.create(config);
   }
 
-  public getExportURL?(
-    dataset: Dataset<ArrayShape, NumericType>,
+  public getExportURL?<D extends Dataset<ArrayShape>>(
+    dataset: D,
     selection: string | undefined,
+    value: Value<D>,
     format: ExportFormat
-  ): string | undefined; // `undefined` if format is not supported
+  ): URL | undefined; // `undefined` if format is not supported
 
   public addProgressListener(cb: ProgressCallback): void {
     this.progressListeners.add(cb);
--- a/packages/app/src/providers/h5grove/h5grove-api.ts
+++ b/packages/app/src/providers/h5grove/h5grove-api.ts
@@ -6,10 +6,15 @@ import type {
   Dataset,
   Entity,
   Group,
-  NumericType,
   ProvidedEntity,
+  Value,
 } from '@h5web/shared';
-import { hasScalarShape, buildEntityPath, EntityKind } from '@h5web/shared';
+import {
+  buildEntityPath,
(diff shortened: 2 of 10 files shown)


packages/app/src/providers/api.ts
```diff
@@ -44,11 +44,12 @@ export abstract class DataProviderApi {
...
 
-  public getExportURL?(
-    dataset: Dataset<ArrayShape, NumericType>,
+  public getExportURL?<D extends Dataset<ArrayShape>>(
+    dataset: D,
     selection: string | undefined,
+    value: Value<D>,
     format: ExportFormat
-  ): string | undefined; // `undefined` if format is not supported
+  ): URL | undefined; // `undefined` if format is not supported
```
axelboc, line 52: Just a bit of a stricter return type here to prevent returning invalid URLs.

packages/app/src/vis-packs/core/compound/MappedCompoundMatrixVis.tsx
```diff
@@ -43,14 +44,20 @@ function MappedCompoundMatrixVis(props: Props) {
...
       {toolbarContainer &&
         createPortal(
           <MatrixToolbar
-            dataset={dataset}
-            selection={getSliceSelection(dimMapping)}
             cellWidth={cellWidth}
+            isSlice={selection !== undefined}
+            getExportURL={
+              getExportURL &&
+              ((format) => getExportURL(dataset, selection, value, format))
```
axelboc, line 59: I'm now currying the `getExportURL` function from the mapped vis components to avoid passing `dataset` `selection` and `value` props to all the toolbars...

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -81,7 +89,7 @@ export class H5GroveApi extends DataProviderApi {
       searchParams.set('selection', selection);
     }
 
-    return `${baseURL as string}/data/?${searchParams.toString()}`;
+    return new URL(`${baseURL as string}/data/?${searchParams.toString()}`);
```
loichuder, line 92: You could set the base separately ``new URL(`data/?${searchParams.toString()}`, baseURL as string)`` but I don't mind the current form

loichuder (approved):

packages/app/src/providers/h5grove/h5grove-api.ts
axelboc: Seems that `baseURL` gets truncated down to the origin (protocol + domain) when it is passed as second argument to `new URL()`: the path `/h5grove` is removed. I've reverted back to what I had initially. Good try though.

loichuder: [Indeed](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL):
```js
new URL('/en-US/docs', "https://developer.mozilla.org/fr-FR/toto"); 
// => 'https://developer.mozilla.org/en-US/docs'
```

Well good to know.

PR #1239, opened 2022-10-21
Assert env vars in demo

axelboc:
I had a 404 due to a missing `.env.local` file and it took me a couple of minutes to understand what was going on.


loichuder (approved):
Reinstalling stuff has good sides :grin:

PR #1240, opened 2022-10-24
Support client-side exports and one-time export tokens

### LINKED ISSUE #1063: Allow client-generated exports of datasets

axelboc:
Fix #1063

---

Providers that implement `getExportURL` can now return an async function that gets called when the user actually clicks on an export entry. This function can return either `Blob` or `URL` to support two main use cases:

- `Blob` for generating client-side exports as per #1063 
- `URL` for authenticated back-ends, as per https://gitlab.esrf.fr/ui/daiquiri/-/merge_requests/538#note_195991 (i.e. when a one-time token has to be fetched from an authenticated endpoint)

Note that this is not restrictive. Data provider implementations are free to do whatever they want - e.g.:

- they can perform a client-side export by returning a `data:` URI instead of a `Blob` (with the risk of reaching the URL length limit);
- they can make a `GET` request directly to an authenticated export endpoint (and return a `Blob`) instead of requesting/returning a one-time token.

---

Now to actually resolve #1063 and https://gitlab.esrf.fr/ui/daiquiri/-/merge_requests/538#note_195991, I've added a way for consumers of the providers to pass their own implementations of `getExportURL`.


packages/app/src/providers/api.ts
```diff
@@ -44,12 +45,20 @@ export abstract class DataProviderApi {
...
+   * - `URL`                  Provider has dedicated endpoint for generating server-side exports
+   * - `() => Promise<URL>`   Provider generates single-use export URLs (i.e. signed one-time tokens)
+   * - `() => Promise<Blob>`  Export is to be generated client-side
+   * - `undefined`            Export scenario is not supported
+   */
   public getExportURL?<D extends Dataset<ArrayShape>>(
     dataset: D,
     selection: string | undefined,
     value: Value<D>,
     format: ExportFormat
```
axelboc, line 60: I'm tempted to move the `format` first. What do you think? If you agree, I'll do that in a separate PR.

packages/app/src/providers/mock/mock-api.ts
```diff
@@ -75,6 +81,39 @@ export class MockApi extends DataProviderApi {
...
+    dataset: D,
+    selection: string | undefined,
+    value: Value<D>,
+    format: ExportFormat
+  ): ExportURL {
+    if (this._getExportURL) {
+      this._getExportURL(dataset, selection, value, format);
+    }
+
+    if (
```
axelboc, line 94: For testing purposes, I'm implementing a very basic client-side CSV export for 1D numeric datasets (`/nD_datasets/oneD`, `/nD_datasets/oneD_linear`, etc.)

packages/lib/src/toolbar/controls/ExportEntry.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+function ExportEntry(props: Props) {
+  const { format, url } = props;
+  const label = `Export to ${format.toUpperCase()}`;
+  const filename = `data.${format}`;
+
+  async function handleBtnClick(getURLOrBlob: () => Promise<URL | Blob>) {
+    const anchor = document.createElement('a');
+    anchor.download = filename;
+    document.body.append(anchor);
```
axelboc, line 16: To start a download for which you don't know the URL right away, the idea is to add a temporary link to the DOM (invisibly), set its `href` once the URL is known, and click on it programmatically.

packages/lib/src/toolbar/controls/ExportEntry.tsx
```diff
@@ -0,0 +1,63 @@
...
+        className={styles.linkOption}
+        type="button"
+        onClick={() => void handleBtnClick(url)}
+      >
+        <span className={styles.label}>{label}</span>
+      </button>
+    );
+  }
+
+  return null;
```
axelboc, line 59: I preferred to have `url === undefined` as the fallback without an `if` block instead, as it seemed clearer to make the other conditions explicit.

packages/app/src/providers/models.ts
```diff
@@ -31,5 +31,6 @@ export interface AttrValuesStore extends FetchStore<AttributeValues, Entity> {
 }
 
 export type ExportFormat = 'csv' | 'npy' | 'tiff';
+export type ExportURL = URL | (() => Promise<URL | Blob>) | undefined;
```
axelboc, line 34: What's new is the `() => Promise<URL | Blob>` part

axelboc:
In the future, we may want to improve the APIs of the provider components (especially `H5GroveProvider`), as well as the corresponding back-ends, so that it's easier to support one-time tokens/authenticated exports.

We can also consider bringing some client-side export utilities into `@h5web/h5wasm`.

packages/app/src/providers/mock/mock-api.ts
loichuder: Great initiative :+1:

packages/lib/src/toolbar/controls/ExportEntry.tsx
```diff
@@ -0,0 +1,63 @@
...
+  format: string;
+  url: URL | (() => Promise<URL | Blob>) | undefined;
+}
+
+function ExportEntry(props: Props) {
+  const { format, url } = props;
+  const label = `Export to ${format.toUpperCase()}`;
+  const filename = `data.${format}`;
+
+  async function handleBtnClick(getURLOrBlob: () => Promise<URL | Blob>) {
```
loichuder, line 13: This function could be moved below the check `url instanceof URL` as it will not be used in this case

loichuder (approved):
Promising :100:

loichuder:
> However, in both cases, I would prefer to see consumer implementations first.

Agreed. All the bricks are there but we must first build something with them before deciding on the future abstractions.

packages/lib/src/toolbar/controls/ExportEntry.tsx
axelboc: Functions declarations are hoisted, so this won't do what you think 😉: https://developer.mozilla.org/en-US/docs/Glossary/Hoisting

packages/lib/src/toolbar/controls/ExportEntry.tsx
loichuder: Yep :+1:

PR #1242, opened 2022-10-25
Fallback to built-in `getExportURL` implementations and document prop for every provider

axelboc:


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -73,13 +73,14 @@ export class H5GroveApi extends DataProviderApi {
...
-    value: Value<D>,
-    format: ExportFormat
+    value: Value<D>
   ): ExportURL {
-    if (this._getExportURL) {
-      this._getExportURL(dataset, selection, value, format);
+    const url = this._getExportURL?.(format, dataset, selection, value);
+    if (url) {
+      return url;
     }
```
axelboc, line 84: Refactoring the way the consumer's `getExportURL` function is used so that it enhances the providers' built-in implementations instead of replacing them.

packages/h5wasm/README.md
```diff
@@ -138,3 +138,19 @@ resources.
...
 </H5WasmProvider>
 ```
+
+#### `filename: string` (required)
+
+The name of the file to display in the UI.
+
+#### `buffer: ArrayBuffer` (required)
+
+The entire file's content as a binary buffer.
```
axelboc, line 148: Documenting the other two props.

.vscode/settings.json
```diff
@@ -18,7 +18,7 @@
 
   "eslint.codeAction.disableRuleComment": { "location": "sameLine" },
   "eslint.workingDirectories": [
-    { "pattern": "./apps/*/" },
-    { "pattern": "./packages/*/" }
+    { "pattern": "./apps/*", "!cwd": true },
+    { "pattern": "./packages/*", "!cwd": true }
```
axelboc, line 22: The problem is back and this seems to solve it. If I touch `cwd` in the workspace's ESLint config, ESLint can no longer find the `tsconfig.json` files and rules based on type-checking don't work.

packages/app/README.md
```diff
@@ -197,6 +197,29 @@ parameters to send to the server. If you use one of H5Grove's default API
 implementations, then you'll need to use this prop to pass the `file` query
 parameter as shown above.
 
+#### `getExportURL?: URL | (() => Promise<URL | Blob>) | undefined`
```
loichuder, line 200: This is a function right ?
```suggestion
#### `getExportURL?: (...args) => URL | (() => Promise<URL | Blob>) | undefined` (optional)
```

packages/app/README.md
```diff
@@ -197,6 +197,29 @@ parameters to send to the server. If you use one of H5Grove's default API
...
+
+The `DataProvider#getExportURL` method is used by the toolbars to generate URLs
+for exporting the current dataset/slice in various formats. This prop allows
+providing your own implementation of this method. If you don't or provide a
```
loichuder, line 204: I would put the sentence beginning by `If you don't ...` at the end of the paragraph.

The doc of the other providers point to this paragraph so it feels odd to have a specificity of `H5GroveProvider` described in the middle of it.

loichuder: Also I found the phrasing `If you don't or provide a function that handles only specific export scenarios` confusing.

Suggestion:
```
`H5GroveProvider` falls back to generating URLs based on the `/data` endpoint and `format` query param if you don't provide your own implementation 
```

packages/app/README.md
```diff
@@ -197,6 +197,29 @@ parameters to send to the server. If you use one of H5Grove's default API
 implementations, then you'll need to use this prop to pass the `file` query
 parameter as shown above.
 
+#### `getExportURL?: URL | (() => Promise<URL | Blob>) | undefined`
+
+The `DataProvider#getExportURL` method is used by the toolbars to generate URLs
```
loichuder, line 202:
```suggestion
The `DataProvider#getExportURL` method is used by the toolbars to generate URLs and controls
```

More generally, I think it would be nice to give an example as all this is a bit abstract and hard to figure out.

packages/app/README.md
axelboc: Arf, thanks

packages/app/README.md
axelboc: I've reworded the paragraph you mention, as well as the text in the other providers.

packages/app/README.md
axelboc: I've added three examples.

loichuder (approved):

packages/app/README.md
```diff
@@ -226,6 +293,15 @@ private data.
 
 The path of the file to request.
 
+#### `getExportURL?: (...args) => URL | (() => Promise<URL | Blob>) | undefined` (optional)
+
+See
+[`H5GroveProvider#getExportURL`](https://github.com/silx-kit/h5web/blob/main/packages/app/README.md#getexporturl-args--url----promiseurl--blob--undefined-optional).
```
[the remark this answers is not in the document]
axelboc: Yeah, wasn't sure if NPM generates the same anchor IDs, though.

packages/app/README.md
```diff
@@ -197,6 +197,73 @@ parameters to send to the server. If you use one of H5Grove's default API
...
+getExportURL={(format, dataset, selection) => async () => {
+  const query = new URLSearchParams({ format, path: dataset.path, selection });
+  const response = await fetch(`${AUTH_TOKEN_ENDPOINT}?${query.toString()}`, {
+    headers: { /* authentication header */ }
+  })
+
+  // Response body contains temporary, pre-authenticated export URL
+  return new URL(await response.body());
+}}
+```
```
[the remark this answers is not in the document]
axelboc: Good call

PR #1243, opened 2022-10-25
Allow H5Wasm entities to be null (e.g. datatype)

loichuder:
Before https://github.com/usnistgov/h5wasm/pull/32 is released, entities from h5wasm can be `null` which raises an error in the Provider.

This makes the `null` entities fall back to `Unresolved`.

### DIFF
--- a/packages/h5wasm/src/h5wasm-api.ts
+++ b/packages/h5wasm/src/h5wasm-api.ts
@@ -124,7 +124,7 @@ export class H5WasmApi extends ProviderApi {
     return new H5WasmFile(id, 'r');
   }
 
-  private async getH5WasmEntity(path: string): Promise<H5WasmEntity> {
+  private async getH5WasmEntity(path: string) {
     const file = await this.file;
 
     const h5wEntity = file.get(path);
@@ -173,7 +173,6 @@ export class H5WasmApi extends ProviderApi {
         ...baseGroup,
         children: h5wEntity.keys().map((childName) => {
           const h5wChild = h5wEntity.get(childName);
-          assertNonNull(h5wChild);
 
           const childPath = buildEntityPath(path, childName);
           return this.processH5WasmEntity(childName, childPath, h5wChild, true);
--- a/packages/h5wasm/src/models.ts
+++ b/packages/h5wasm/src/models.ts
@@ -4,7 +4,7 @@ import type {
   CompoundTypeMetadata,
 } from 'h5wasm/src/hdf5_util_helpers';
 
-export type H5WasmEntity = NonNullable<ReturnType<H5WasmGroup['get']>>;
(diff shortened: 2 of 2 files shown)


packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -108,13 +108,7 @@ export class H5WasmApi extends ProviderApi {
   private async getH5WasmEntity(path: string): Promise<H5WasmEntity> {
     const file = await this.file;
 
-    const h5wEntity = file.get(path);
-    assertNonNull(
-      h5wEntity,
-      path === '/' ? `Expected valid HDF5 file` : `No entity found at ${path}`
-    );
```
loichuder, line 115: Actually, I should perhaps keep this assertion for `path = '/'`

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -154,7 +148,6 @@ export class H5WasmApi extends ProviderApi {
         ...baseGroup,
         children: h5wEntity.keys().map((childName) => {
           const h5wChild = h5wEntity.get(childName);
-          assertNonNull(h5wChild);
```
loichuder, line 157: The fallback is already present below so only the removal of assertions are needed

packages/h5wasm/src/h5wasm-api.ts
loichuder: > getH5WasmEntity is never called on anything else than a group by the Explorer

Ah, yes indeed, this is the result of our caching strategy. Well, I would prefer we didn't rely on this but this is only until datatypes are not returned as `null`.

packages/h5wasm/src/models.ts
```diff
@@ -4,7 +4,7 @@ import type {
   CompoundTypeMetadata,
 } from 'h5wasm/src/hdf5_util_helpers';
 
-export type H5WasmEntity = NonNullable<ReturnType<H5WasmGroup['get']>>;
+export type H5WasmEntity = ReturnType<H5WasmGroup['get']>;
```
axelboc, line 7: Still needed?

axelboc (approved):

packages/h5wasm/src/models.ts
loichuder: Well, yes, it is handy and clearer to put `H5WasmEntity` rather `NonNullable<ReturnType<H5WasmGroup['get']>>` (when typing argument from guards for example)

axelboc: Sorry, I meant removing the non-nullable part, but all good 👍

loichuder: Perhaps the diff is not clear but it was indeed removed

PR #1249, opened 2022-10-28
Fix `Value<D>` and export `getExportURL` function type

axelboc:
Partial fix for [#1063 ](https://github.com/silx-kit/h5web/issues/1063#issuecomment-1293771087)

There was a problem with the `Value<D>` type. In advanced cases, like the one identified by @bmaranville, the shape and dtype of the `Dataset` generic was not properly inferred.

`getExportURL` is only called with array-shaped datasets (`Dataset<ArrayShape>`), but `Value<D>` was not being narrowed down at all. Now, `Value<D>` is properly narrowed down to `unknown[] | TypedArray` (if we go down to the barest types).

```tsx
getExportURL={(format, dataset, selection, value) => async () => {
  value.forEach((val) => { /* ... */ }); // `forEach` now works because it is available on both arrays and typed arrays
}}
```

---

I'm also exporting the type of the `getExportURL` function in `@h5web/app` so one can extract the function outside of the JSX:

```tsx
const getExportURL: GetExportURL = (format, dataset, selection, value) => { ... };
```

I'm usually not a fan of function expressions, but in this case, replicating the signature would require exporting a lot of types and would be prone to mistakes. Indeed, with a function declaration, given the permissive nature of TS, consumers may not end up typing the function as strictly as they could and therefore struggle to implement it.

### DIFF
--- a/packages/app/src/index.ts
+++ b/packages/app/src/index.ts
@@ -1,3 +1,5 @@
+import type { DataProviderApi } from './providers/api';
+
 export { default as App } from './App';
 
 export { default as MockProvider } from './providers/mock/MockProvider';
@@ -7,6 +9,7 @@ export { default as H5GroveProvider } from './providers/h5grove/H5GroveProvider'
 export { getFeedbackMailto } from './breadcrumbs/utils';
 export type { FeedbackContext } from './breadcrumbs/models';
 export type { ExportFormat, ExportURL } from './providers/models';
+export type GetExportURL = NonNullable<DataProviderApi['getExportURL']>;
 
 // Context
 export { useDataContext } from './providers/DataProvider';
--- a/packages/app/src/providers/hsds/hsds-api.ts
+++ b/packages/app/src/providers/hsds/hsds-api.ts
@@ -8,7 +8,6 @@ import type {
   ChildEntity,
   ProvidedEntity,
   ArrayShape,
-  DType,
   Value,
 } from '@h5web/shared';
 import { assertGroup } from '@h5web/shared';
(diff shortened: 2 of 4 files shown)


packages/app/src/index.ts
```diff
@@ -7,6 +9,7 @@ export { default as H5GroveProvider } from './providers/h5grove/H5GroveProvider'
 export { getFeedbackMailto } from './breadcrumbs/utils';
 export type { FeedbackContext } from './breadcrumbs/models';
 export type { ExportFormat, ExportURL } from './providers/models';
+export type GetExportURL = NonNullable<DataProviderApi['getExportURL']>;
```
axelboc, line 12: I'm declaring the type alias here directly to avoid cyclic imports.

packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -152,7 +151,7 @@ export class HsdsApi extends DataProviderApi {
     );
   }
 
-  public getExportURL<D extends Dataset<ArrayShape, DType>>(
+  public getExportURL<D extends Dataset<ArrayShape>>(
```
axelboc, line 154: VS Code added `DType` automatically when I implemented the method, but it's the default anyway.

packages/app/src/index.ts
bmaranville: Thanks!  I was realizing I needed to define this function in the application code, and you're right I had to import a lot of types to try to build the function signature.

loichuder (approved):

Issue #903, opened 2021-12-13
Ordinate label can appear above tick labels


axelboc:
![image](https://user-images.githubusercontent.com/2936402/145815263-5a4ef7f2-5b80-41b4-bc82-96262d9996fe.png)

loichuder:
Can be seen also in `/surface_file` plot of https://h5web.panosc.eu/h5wasm?url=https%3A%2F%2Fraw.githubusercontent.com%2Foasys-esrf-kit%2Fdabam2d%2Fmain%2Fdata%2Fdabam2d-001.h5

loichuder:
Issue also raised in visx (that renders the axis): https://github.com/airbnb/visx/issues/1395

axelboc:
[Playing with](https://observablehq.com/@d3/d3-format) some [D3 formats](https://github.com/d3/d3-format#locale_format):

### `.3~g`

```js
d3.format(".3~g")(3234234234234452) // 3.23e+15 (B)
d3.format(".3~g")(123452) // 1.23e+5
d3.format(".3~g")(122) // 122
d3.format(".3~g")(1) // 1 (C)
d3.format(".3~g")(0.12342354345) // 0.123
d3.format(".3~g")(0.0000012342354345) // 0.00000123 (A)
d3.format(".3~g")(0.00000012342354345) // 1.23e-7
d3.format(".3~g")(0.0000000000012342354345) // 1.23e-12
```

- (A) max length of 10 characters. It's weird; I would expect exponent notation to kick in right from `0.0xxx`...
- (B) length of 8, which is what I would have expected the max length to be.
- (C) min length of 1.

### `.3~s`

```js
d3.format(".3~s")(3234234234234452) // 3.23P
d3.format(".3~s")(123452) // 123k
d3.format(".3~s")(122) // 122
d3.format(".3~s")(1) // 1
d3.format(".3~s")(0.12342354345) // 123m
d3.format(".3~s")(0.0000012342354345) // 1.23µ
d3.format(".3~s")(0.00000012342354345) // 123n
d3.format(".3~s")(0.0000000000012342354345) // 1.23p
```

This is a lot more stable with 5 characters max. But do we want SI prefixes even when axes don't have units?

loichuder:
`g` is in fact an alias for `Number.toPrecision`.

`s` is indeed the best candidate to get fixed-width strings. I am not a huge fan of prefixes but it could be an option to activate :shrug:

PR #1250, opened 2022-10-31
Improve tick format to enforce maximum char count

axelboc:
Part 1 of 2 to try to resolve issue #903.

In this PR, I'm improve the formatter for tick values used by the `Axis` component and for the `ColorBar` axis.

The new formatter tries `.3~g` first (automatic mode, equivalent to [`Number.toPrecision(3)`](https://devdocs.io/javascript/global_objects/number/toprecision) as you pointed out @loichuder), and if the resulting string has more than 3 decimal numbers (e.g. `0.0123`), it formats the number again but with `.3~e` instead (exponent notation), which guarantees a maximum of 10 characters (`-0.123e-12`) -- though this extreme case is not very likely to occur, I think.

This seems to be the best compromise that doesn't require using SI prefixes.

Note that I have not adjusted the axis offsets for now to keep the PR focused, so some of the labels may overflow. In my next PR, I'll work on improving all the design aspects: reducing the font-size of tick labels, aligning the color bar with the heatmap, adjusting the axis offsets, etc.

That being said, I did change the way the axis offsets are computed a little, as I couldn't quite understand what was going on any more. It seemed to me that some of the logic was wrong: the `fallback` value was used to make space for labels, for instance. I went with a more explicit solution that doesn't try to be too generic.

None of our mock datasets with screenshots had axis values that were affected by the new formatter, so I decided to change the ordinate values of the scatter visualization to demonstrate the new behaviour.


packages/lib/src/vis/heatmap/ColorBar.tsx
```diff
@@ -83,10 +85,16 @@ function ColorBar(props: Props) {
             numTicks={adaptedNumTicks(gradientLength)}
             tickClassName={styles.tick}
             tickComponent={Tick}
-            tickFormat={axisScale.tickFormat(
-              adaptedNumTicks(gradientLength),
-              '.3~g'
-            )}
+            tickFormat={
+              scaleType === ScaleType.Log
+                ? axisScale.tickFormat(
```
axelboc, line 90: Calling `axisScale.tickFormat` with `.3~g` (i.e. with a precision) is pointless for "linearish" scales (i.e. other than log), since the function's purpose is to compute the precision dynamically.

For log scales, however, it returns empty labels for some ticks to avoid overlap: https://github.com/d3/d3-scale/blob/main/src/log.js#L118-L122. I might try to bring this logic into the app at some point to make this behaviour more explicit, to avoid the `@ts-ignore` below, and to maybe apply the same behaviour for symlog scales, which tend to have overlapping ticks as well.

packages/lib/src/vis/shared/Axis.tsx
```diff
@@ -58,12 +59,19 @@ function Axis(props: Props) {
...
+          width: isAbscissa ? axisLength : offset,
+          height: isAbscissa ? offset : axisLength,
+        }}
+      >
         <AxisComponent
           scale={scale}
           tickFormat={getTickFormatter(domain, axisLength, scaleType)}
           label={label}
-          labelOffset={type === 'abscissa' ? 28 : 42}
+          labelOffset={offset - (isAbscissa ? 32 : 36)}
```
axelboc, line 74: The label offset is now computed from the axis offset, which means that if we adjust axis offsets in the future, the labels won't change position again. Note that the resulting label offsets are now a bit bigger than before (the labels are closer to the edge of the vis area and further from the tick labels).

packages/lib/src/vis/shared/Axis.tsx
```diff
@@ -58,12 +59,19 @@ function Axis(props: Props) {
 
   return (
     <>
-      <svg className={styles.axis} style={svgSize} data-type={type}>
+      <svg
+        className={styles.axis}
+        data-type={type}
+        style={{
+          width: isAbscissa ? axisLength : offset,
```
loichuder, line 66:
```suggestion
          width: isAbscissa ? width : offset,
```

packages/lib/src/vis/shared/Axis.tsx
```diff
@@ -58,12 +59,19 @@ function Axis(props: Props) {
...
+          height: isAbscissa ? offset : axisLength,
+        }}
```
loichuder, line 68:
```suggestion
        style={isAbscissa ? { width, height: offset } : { width: offset, height }}
```

packages/lib/src/vis/utils.ts
```diff
@@ -262,15 +264,14 @@ export function getCombinedDomain(
...
+  const { left, right, top, bottom } = DEFAULT_AXIS_OFFSETS;
   return {
-    left: (hasLabel.left ? fallback : 0) + vertical,
-    bottom: (hasLabel.bottom ? fallback : 0) + horizontal,
-    right: hasLabel.right ? vertical : fallback,
-    top: hasLabel.top ? horizontal : fallback,
+    left: left + (hasLabel.left ? LABEL_OFFSET : 0),
+    right,
+    top: top + (hasLabel.top ? TITLE_OFFSET : 0),
+    bottom: bottom + (hasLabel.bottom ? LABEL_OFFSET : 0),
```
loichuder, line 274: Yeah, that's much better

packages/lib/src/vis/heatmap/ColorBar.tsx
loichuder: I don't understand the TS issue here. Is it a problem in the types of `d3/scale` ?

loichuder (approved):
Good improvement :+1:

packages/lib/src/vis/heatmap/ColorBar.tsx
axelboc: Precisely:

image

... but the code allows for a `function` and calls it: https://github.com/d3/d3-scale/blob/main/src/log.js#L121

axelboc: I would consider improving `@types/d3-format`, but since I'm considering bringing this logic in the codebase to have more control over it (for symlog, but also because the `scale.tickFormat` API feels super confusing).

PR #1253, opened 2022-11-02
Show compression filters info and throw on unsupported filter with h5wasm

axelboc:
image

image


loichuder (approved):
This is great :heart:

PR #1259, opened 2022-11-03
Support stacks of RGB images

axelboc:
As discussed, I don't show an axis mapper for the last dimension (the one that holds the 3 channel values). This was quite straightforward in the end. No need to refactor `DimensionMapper`, `applyMapping` or anything else.

I've updated the `rgb-image` mock entity to hold a stack of 10 RGB images:

image

Of course, in addition to the nD slicing, RGB images can now be transposed:

image


packages/app/src/visualizer/utils.ts
```diff
@@ -102,11 +102,12 @@ function getSupportedNxVis(
 
   const dataset = findSignalDataset(entity, attrValuesStore);
   const isCplx = hasComplexType(dataset);
-  const { interpretation } = attrValuesStore.get(dataset);
+  const { interpretation, CLASS } = attrValuesStore.get(dataset);
 
   if (
-    interpretation === NxInterpretation.RGB &&
-    hasNumDims(dataset, 3) &&
+    (interpretation === NxInterpretation.RGB || CLASS === 'IMAGE') &&
```
axelboc, line 108: Oh forgot to point out here that I'm adding a check for `CLASS === 'IMAGE'` on the signal. I reckon that's why Sandor thought we didn't support RGB images with NeXus: maybe they don't use `interpretation=rgb-image` at FairMAT.

packages/app/src/vis-packs/nexus/containers/NxRgbContainer.tsx
```diff
@@ -16,26 +19,45 @@ function NxRgbContainer(props: VisContainerProps) {
   assertNumericSignal(nxData);
 
   const { signalDef } = nxData;
-  assertNumDims(signalDef.dataset, 3);
+  assertMinDims(signalDef.dataset, 3);
 
   const { shape: dims } = signalDef.dataset;
+  if (dims.at(-1) !== 3) {
```
loichuder, line 25: Nice, I just discovered [at](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Array/at) :joy_cat:

packages/app/src/vis-packs/core/visualizations.ts
```diff
@@ -144,7 +144,8 @@ export const CORE_VIS: Record<Vis, CoreVisDef> = {
       return (
         attrValuesStore.getSingle(dataset, 'CLASS') === 'IMAGE' &&
         hasArrayShape(dataset) &&
-        dataset.shape.length === 3 &&
+        dataset.shape.length >= 3 && // 2 for axes + 1 for RGB channels
+        dataset.shape.at(-1) === 3 && // 3 channels
```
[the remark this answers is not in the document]
loichuder: Doesn't hurt to add it

loichuder (approved):

packages/app/src/vis-packs/nexus/containers/NxRgbContainer.tsx
axelboc: Yeah, thought I'd take it for a spin. We'll have to remember to check if we need to add a polyfill in Braggy/Hibou/....

PR #1261, opened 2022-11-04
Allow histogram indicators to be interactive

loichuder:
As discussed, first version based on `@visx/drag`.

Handy for several reasons:
- API quite close from our `Selection` components
- Possibility to restrict the range of motion
- Works well with `transform={translate...}`

But:
- the `MouseUp` event may be missed when approaching the limits of the histogram **(edit: [fixed!](https://github.com/silx-kit/h5web/pull/1261#discussion_r1015240550))**
- Picking the line itself can be a bit difficult **(edit: fixed by adding an invisible overlaying rectangle to capture pointer events)**
- Does not allow to move the indicator when clicking on a area in the histogram which may be requested by users **(edit: not handled by silx so we can live without it for now ?)**

Also, I did not make the relevant changes in the `DomainSlider` yet to make the included histogram interactive. I thought it could be a separate PR


packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
```diff
@@ -42,15 +50,15 @@ function Histogram(props: Props) {
   const { width, height } = size;
 
   const xScale = H5WEB_SCALES[scaleType].createScale({
-    domain: xDomain,
+    domain: xDomain && extendDomain(xDomain, EXTEND_FACTOR, scaleType),
```
loichuder, line 53: As in `DomainSlider, I now extend the xDomain to have space to move the indicators.

packages/lib/src/toolbar/controls/Histogram/InteractiveIndicators.tsx
```diff
@@ -0,0 +1,61 @@
...
+        x={xMin}
+        onDragEnd={({ dx }) => onChange(xMin + dx, xMax)}
+        restrict={{ xMin: 0, xMax }}
+      >
+        {({ dragStart, dragEnd, dragMove, dx }) => (
+          <g
+            onMouseMove={dragMove}
+            onMouseUp={dragEnd}
+            onMouseDown={dragStart}
+            transform={`translate(${dx}, 0)`}
```
loichuder, line 33: I followed the example of `@visx/drag`: https://airbnb.io/visx/drag-i

packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
```diff
@@ -75,7 +83,22 @@ function Histogram(props: Props) {
...
           />
         )}
-        <HistogramIndicators positions={indicatorPositions} height={height} />
+        <g className={styles.indicators}>
+          {onChange ? (
+            <InteractiveIndicators
+              positions={indicatorPositions}
+              size={size}
+              onChange={(min, max) =>
+                onChange(xScale.invert(min), xScale.invert(max))
```
loichuder, line 92: This allows to only work in SVG coordinates into `InteractiveIndicators`

axelboc: Clever

packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
```diff
@@ -75,7 +83,22 @@ function Histogram(props: Props) {
...
+              }
+            />
+          ) : (
+            <>
+              <Indicator x={indicatorPositions[0]} height={height} />
+              <Indicator x={indicatorPositions[1]} height={height} flipArrow />
```
axelboc, line 98: The height can't be 100%, right? Just checking.

packages/lib/src/toolbar/controls/Histogram/InteractiveIndicators.tsx
axelboc: For the mouse up problem, you might be able to capture/release the pointer with `onMouseDown`/`onMouseUp`. Might be worth using pointer events over mouse events, by the way (but maybe that's not supported by visx).

packages/lib/src/toolbar/controls/Histogram/InteractiveIndicators.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const { positions, size, onChange } = props;
+  const { height, width } = size;
+  const [xMin, xMax] = positions;
+
+  return (
+    <g className={styles.indicators}>
+      <Drag
+        height={height}
+        width={width}
+        x={xMin}
```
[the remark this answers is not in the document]
loichuder: This messes the scales up. I think I will keep the dimensions like this for now

packages/lib/src/toolbar/controls/Histogram/InteractiveIndicators.tsx
loichuder: Capturing the pointer fixed the `onPointerUp` not called when releasing out-of-bounds :tada:

packages/lib/src/toolbar/controls/Histogram/Indicators.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <g className={styles.indicators}>
+      <Drag
+        height={height}
+        width={width}
+        x={xMin}
+        onDragEnd={({ dx }) => onChange?.([xMin + dx, xMax])}
+        restrict={{ xMin: 0, xMax }}
+      >
+        {(dragState) => (
+          <Indicator x={xMin} interactive={!!onChange} {...dragState} />
```
axelboc: Instead of having a separate `interactive` prop, could you maybe just spread `dragState` conditionally?

```tsx
{...onChange && dragState}
```

Then in `Indicator`, you can have:

```tsx
  const handlers = useMemo(
    () => ({
            onPointerMove: dragMove,
            onPointerUp: dragEnd && (e: PointerEvent<Element>) => {
              (e.target as Element).releasePointerCapture(e.pointerId);
              dragEnd(e);
            },
            onPointerDown: dragStart && (e: PointerEvent<Element>) => {
              (e.target as Element).setPointerCapture(e.pointerId);
              dragStart(e);
            }
          }),
    [dragEnd, dragMove, dragStart, interactive]
  );
```

You can add the cursor style conditionally in the JSX based on `dragStart` or whatever.

axelboc (approved):
Nice!! And fairly concise too :clap:

PR #1262, opened 2022-11-07
Support custom axes for NX RGB

axelboc:
image


loichuder (approved):
I don't know if that is worth supporting but all right :+1:

PR #1265, opened 2022-11-07
Make histogram interactive in DomainSlider

loichuder:
That could have been as simple as adding [this line](https://github.com/silx-kit/h5web/compare/interactive-histogram-slider?expand=1#diff-1ddc4e75c70dd0e3279348964b72806822988ea30c61e19a69240832a32d4d65R141).

But due to https://github.com/airbnb/visx/issues/1174, the point positions when dragging are wrongly computed as the domain slider root div is translated via `transform`. This was to center the tooltip with respect to the slider.

To fix this, I reimplemented the centering by adding a flex container around the slider container and the tooltip to center them. This allows to remove `transform` from the histogram hierarchy of parents to work around https://github.com/airbnb/visx/issues/1174.

### DIFF
--- a/packages/lib/src/toolbar/controls/DomainSlider/DomainSlider.tsx
+++ b/packages/lib/src/toolbar/controls/DomainSlider/DomainSlider.tsx
@@ -70,37 +70,38 @@ function DomainSlider(props: Props) {
       onPointerEnter={() => toggleHovered(true)}
       onPointerLeave={() => toggleHovered(false)}
     >
-      <ScaledSlider
-        value={sliderDomain}
-        safeVisDomain={safeDomain}
-        dataDomain={dataDomain}
-        scaleType={scaleType}
-        errors={errors}
-        isAutoMin={isAutoMin}
-        isAutoMax={isAutoMax}
-        onChange={(newValue) => {
-          setSliderDomain(newValue);
-          toggleEditing(false);
-        }}
-        onAfterChange={(hasMinChanged, hasMaxChanged) => {
-          onCustomDomainChange([
-            hasMinChanged ? sliderDomain[0] : customDomain[0],
-            hasMaxChanged ? sliderDomain[1] : customDomain[1],
-          ]);
-        }}
-      />
+      <div className={styles.sliderContainer}>
+        <ScaledSlider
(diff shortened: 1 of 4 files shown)


packages/lib/src/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -70,37 +70,38 @@ function DomainSlider(props: Props) {
...
-        small
-        label="Edit domain"
-        aria-expanded={hovered || isEditing}
-        aria-controls={TOOLTIP_ID}
-        icon={FiEdit3}
-        value={isEditing}
-        onToggle={() => toggleEditing(!isEditing)}
-      />
-
+      <div className={styles.sliderContainer}>
```
loichuder, line 73: The slider and the edit button are now wrapped in their own `div`.

packages/lib/src/toolbar/controls/Histogram/Marker.tsx
```diff
@@ -34,7 +34,7 @@ function Marker(props: Props) {
 
   return (
     <g
-      transform={dragState ? `translate(${dragState.dx}, 0)` : undefined}
+      transform={`translate(${dragState?.dx || 0}, 0)`}
```
loichuder, line 37: Unrelated but made this a bit more concise

packages/lib/src/toolbar/controls/DomainSlider/DomainSlider.tsx
```diff
@@ -137,6 +138,7 @@ function DomainSlider(props: Props) {
             dataDomain={dataDomain}
             value={sliderDomain}
             scaleType={scaleType}
+            onChange={onCustomDomainChange}
```
loichuder, line 141: The meaningful change to add interactivity

loichuder:
There appears to be a little mismatch between the two implementations:

edit_domain diff

axelboc (approved):

axelboc:
The `div` inside which the tooltip used to be has `margin-right: -0.375rem;`. So moving it out of this element shifts the tooltip to the left by this amount, which means that it is now more centered than it was before :100: :smile:

PR #1268, opened 2022-11-08
Do not shrink slider domain below data domain in DomainSlider

loichuder:
To match the behaviour of the now interactive histogram


axelboc (approved):
Nice!

PR #1269, opened 2022-11-08
Implement own createMemo that preserves original function's overrides

axelboc:
Three reasons for this:

1. `createMemo` comes from `react-use`, so bring it in helps us move away from this unmaintained library.
2. The implementation is very short, so it's low-risk, low-maintenance.
3. `react-use`'s version was not playing well with functions with overrides: the memoised function would lose the overrides altogether, which prevented memoising some utility functions. I only found one such example, which I've refactored, but perhaps we had worked around other cases or had moved away from using `createMemo` for other reasons (e.g. `useDomain` cannot use `createMemo(getDomain)` because the two internal calls need to be memoised independently.)

### DIFF
--- a/packages/shared/src/utils.ts
+++ b/packages/shared/src/utils.ts
@@ -2,6 +2,7 @@ import { format } from 'd3-format';
 import ndarray from 'ndarray';
 import type { NdArray, TypedArray } from 'ndarray';
 import { assign } from 'ndarray-ops';
+import { useMemo } from 'react';
 
 import { assertLength, isNdArray, isTypedArray } from './guards';
 import type {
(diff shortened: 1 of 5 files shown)


packages/shared/src/utils.ts
```diff
@@ -182,3 +183,14 @@ export function getDims(dataArray: NdArray): Dims {
   const [rows, cols] = dataArray.shape;
   return { rows, cols };
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function createMemo<P extends any[], R, T extends (...args: P) => R>(
+  fn: T
+): T;
```
axelboc, line 190: I need an override to explicitly declare `T` as return type. If I declare this return type on the function itself, I get an error: `'(...args: P) => R' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint '(...args: P) => R'.`

loichuder (approved):
Thank you so much :heart: 

Please go ahead and merge this so I can use it in #1268

PR #1270, opened 2022-11-09
Leave `flip` handling out of TiledHeatmapMesh

loichuder:
In https://github.com/silx-kit/h5web/pull/1193, I moved the handling of the `flip` of the axes inside the `TiledHeatmapMesh` (more specifically at the `TiledLayer` level) to avoid flipping the new `TooltipMesh` that was part of `TiledHeatmapMesh`.

But:
- `TiledTooltipMesh` is now out of `TiledHeatmapMesh`
- In daiquiri-tomo, the `flip` was already handled by a higher-order component (`LinearVisGroup`, also demonstrated in the `WithTransform` story in H5Web). **The flip of axes was therefore applied twice for tiled layers.**
- In `HeatmapMesh`, the `flip` is handled outside of the component and I think we want the two components to behave similarly.

For all these reasons, I moved the handling of `flip` out of `TiledHeatmapMesh` reverting part of #1193.

### DIFF
--- a/apps/storybook/src/TiledHeatmapMesh.stories.tsx
+++ b/apps/storybook/src/TiledHeatmapMesh.stories.tsx
@@ -188,7 +188,11 @@ const Template: Story<TiledHeatmapStoryProps> = (args) => {
       <Zoom />
       <SelectToZoom keepRatio modifierKey="Control" />
       <ResetZoomButton />
-      <TiledHeatmapMesh {...tiledHeatmapProps} />
+      <group
+        scale={[abscissaConfig.flip ? -1 : 1, ordinateConfig.flip ? -1 : 1, 1]}
+      >
+        <TiledHeatmapMesh {...tiledHeatmapProps} />
+      </group>
       {showTooltip && <TiledTooltipMesh renderTooltip={renderTooltip} />}
     </VisCanvas>
   );
@@ -336,13 +340,13 @@ WithTransforms.args = {
     visDomain: [-2, 1.5],
     isIndexAxis: true,
     showGrid: false,
-    flip: false,
+    flip: true,
   },
   ordinateConfig: {
     visDomain: [0, 2],
     isIndexAxis: true,
     showGrid: false,
-    flip: false,
+    flip: true,
   },
 };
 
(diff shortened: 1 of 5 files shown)


apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -336,13 +340,13 @@ WithTransforms.args = {
...
     showGrid: false,
-    flip: false,
+    flip: true,
   },
   ordinateConfig: {
     visDomain: [0, 2],
     isIndexAxis: true,
     showGrid: false,
-    flip: false,
+    flip: true,
```
loichuder, line 349: I now flip axes in the `WithTransform` story to be closer from the daiquiri-tomo set-up

packages/lib/src/vis/tiles/TiledLayer.tsx
```diff
@@ -40,36 +39,26 @@ function TiledLayer(props: Props) {
   return (
     // Transforms to use level of details layer array coordinates
     <group
-      scale={[abscissaConfig.flip ? -1 : 1, ordinateConfig.flip ? -1 : 1, 1]}
+      position={[-meshSize.width / 2, -meshSize.height / 2, layer / numLayers]}
```
loichuder, line 42: The group that was flipping the layer was removed

packages/lib/src/vis/tiles/hooks.ts
```diff
@@ -9,23 +9,21 @@ import type { TileArray } from './models';
...
-  const xScale = createAxisScale(ScaleType.Linear, {
-    domain: [-meshSize.width / 2, meshSize.width / 2],
-    range: [0, layerSize.width],
-    reverse: abscissaConfig.flip,
-    clamp: true,
-  });
+    const yScale = createAxisScale(ScaleType.Linear, {
+      domain: [-meshSize.height / 2, meshSize.height / 2],
+      range: [0, layerSize.height],
+      clamp: true,
```
loichuder, line 22: I added a `useMemo` but the important change is elsewhere: the `reverse` args were removed: all inner components and utils from `TiledHeatmapMesh` must be oblivious to the `flip`.

axelboc (approved):

packages/lib/src/vis/tiles/hooks.ts
loichuder: Agreed, In that case, I think it makes more sense to call `getLayerScales` into `scaleBox` as both are always called in succession.

t20100 (approved):
LGTM

packages/lib/src/vis/tiles/hooks.ts
t20100: `meshSize` and `layerSize` should be stable, but it's no big deal anyway.

PR #1272, opened 2022-11-09
Reimplement useTimeout to remove react-use

loichuder:
After #1269, we only had two hooks left from `react-use`:
- `useWindowSize`: can now be used from `react-hookz`
- `useTimeout`: reimplemented in this PR

My implementation does not reach feature parity but I think it fits our simple use of the hook.

### DIFF
--- a/packages/app/src/hooks.ts
+++ b/packages/app/src/hooks.ts
@@ -1,0 +1,13 @@
+import { useEffect, useState } from 'react';
+
+export function useTimeout(ms?: number) {
+  const [isReady, setReady] = useState(false);
+
+  useEffect(() => {
+    const id = setTimeout(() => setReady(true), ms);
+
+    return () => clearTimeout(id);
+  }, [ms]);
+
+  return isReady;
+}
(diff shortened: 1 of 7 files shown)


loichuder:
Hm. I broke all the tests relying on fetching. Perhaps `useTimeout` was not doing what I thought it was

axelboc (approved):
Awesome, thanks!! 🎉 I missed that `useWindowSize` had been ported to react-hookz.

packages/app/src/hooks.ts
```diff
@@ -0,0 +1,13 @@
+import { useEffect, useState } from 'react';
+
+export function useTimeout(ms?: number) {
+  const [isReady, setReady] = useState(false);
+
+  useEffect(() => {
+    const id = setTimeout(() => setReady(true), ms);
+
+    return () => clearTimeout(id);
+  });
```
axelboc, line 10: Ah I think it's because you forgot the deps array here.

loichuder: Good catch :fishing_pole_and_fish: ! Fixed everything :tada:

PR #1273, opened 2022-11-09
Export additionnal types in h5web/lib

loichuder:
Here are my thoughts about what should be exported after reviewing https://gitlab.esrf.fr/ui/daiquiri-ui/-/tree/744d64aaac72c79ffafd61b779af93d71f04cc5b/src/components/h5web:
- `bootstrap/DefaultMouseModeOptions`: daiquiri-specific
- `items`: could stay in daiquiri for now
- `DefaultMouseInteraction`: daiquiri-specific
- `LinearVis...`: Used in `TiledHeatmapMesh` story. But apart from that, I don't think it should be exported by us
- `SelectionPoint`: Could be in H5Web but we have no use for it for now. Could be refactored using `useCanvasEvents` though.
- `SvgScene...`: Bound to disappear
- `VisSynchroniser` and `hooks`: daiquiri-specific
- `models`: no longer needed if `SelectionPoint` is refactored
- `utils`: **export `VisScaleType` and `HistogramParams` in H5Web**. `toScaleType` and `stripHistogram` are arguable. The first almost has an equivalent [in HeatmatMesh](https://github.com/silx-kit/h5web/blob/ef80dac22cf959af43eca53f76b3fb7e811c3287/packages/lib/src/vis/heatmap/HeatmapMesh.tsx#L85) and the second could be useful in Braggy.

### DIFF
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -133,6 +133,8 @@ export type {
   Size,
   AxisConfig,
   AxisParams,
+  VisScaleType,
+  HistogramParams,
 } from './vis/models';
 
 export type { D3Interpolator, ColorMap } from './vis/heatmap/models';


loichuder:
@t20100 Don't hesitate to comment on this as well

axelboc (approved):
Agreed!

We should probably think about `toScaleType`, but it's more complex than just exporting a utility, I reckon. It is used for [this prop](https://gitlab.esrf.fr/ui/daiquiri-ui/-/blob/744d64aaac72c79ffafd61b779af93d71f04cc5b/src/components/images/DropdownColormapH5web.tsx#L10), and [this prop](https://gitlab.esrf.fr/ui/daiquiri-ui/-/blob/744d64aaac72c79ffafd61b779af93d71f04cc5b/src/components/tomo/tiling/HistogramDomainSlider.tsx#L42).

In the case of `DomainSlider`, the component accepts `ScaleType` instead of `VisScaleType`, which means that one could pass the gamma scale without an exponent. This should probably be fixed.

For the other case, we could remove the need for `toScaleType` by forcing low-level components to accept a new type as prop: `[ScaleType] | [ScaleType, number]`. We could name it `ScaleTypeDef` or similar. `VisScaleType` would remain as a convenience type for high-level vis components like `Heatmap`, but components like `DomainSlider` would require `ScaleTypeDef`.

If we go this way, then we can make a utility to [convert `VisScaleType` to `ScaleTypeDef`](https://github.com/silx-kit/h5web/blob/ef80dac22cf959af43eca53f76b3fb7e811c3287/packages/lib/src/vis/heatmap/HeatmapMesh.tsx#L85) and remove [`isScaleType`](https://github.com/silx-kit/h5web/blob/main/packages/shared/src/guards.ts#L421) (since the `ScaleType` would always be the first item of the array).

axelboc:
Yep, no worries! `{ type: ScaleType } | { type: ScaleType.Gamma, exponent: number }` and `VisScaleType` can be `ScaleType | ScaleTypeDef`.

PR #1281, opened 2022-11-14
Restrict correctly the marker position when dragging it

loichuder:
Follow-up of #1280 

The `restrict` prop sets the bound of the **mouse** position while I want to restrict the **marker** position. As I extended the interaction area in #1280, there can be a noticeable mismatch between the two. I therefore store this mismatch (in `deltaX`) so that `restrict` bounds are set to the appropriate value.


axelboc (approved):
Well, that's annoying...

PR #1282, opened 2022-11-14
Do not display the compression filters table in h5wasm if no filter

loichuder:
A dataset loaded with h5wasm with no filter shows a disjointed _Compression filters_ header

image

This PR removes the header if no filter is present


axelboc (commented):
Oh well spotted. Shouldn't the logic be on the `MetadataViewer` side instead?

loichuder:
> Oh well spotted. Shouldn't the logic be on the `MetadataViewer` side instead?

You mean the check should be `entity.filters?.length > 0` instead of `entity.filters` ?

axelboc:
Yes, exactly 👍

PR #1284, opened 2022-11-15
Replace `visRatio` and `layout` props with `aspect` prop

### LINKED ISSUE #1275: [VisCanvas] Replace ratio props
The purpose of the `visRatio` and `canvasRatio` props is unclear and doesn't quite model consumer needs for `VisCanvas`.

As an example, below is a component from Daiquiri called [`LinearVisCanvas`](https://gitlab.esrf.fr/ui/daiquiri-ui/-/blob/a57d38b97ef88fc8816decb977efd57f588fa68a/src/components/h5web/LinearVisCanvas.tsx) that computes `visRatio` from the axis domains. The need here is to for the axes to be orthonormal, so the API of `VisCanvas` should make this more straightforward. If `visRatio` was computed internally, the `LinearVisCanvas` abstraction would no longer be required (`VisCanvas` and `LinearVisGroup` could just be used directly).

```tsx
function LinearVisCanvas(props: PropsWithChildren<Props>) {
  const { abscissaConfig, ordinateConfig, title, children } = props;

  const abscissaExtent = Math.abs(
    abscissaConfig.visDomain[1] - abscissaConfig.visDomain[0]
  );
  const ordinateExtent = Math.abs(
    ordinateConfig.visDomain[1] - ordinateConfig.visDomain[0]
  );

  return (
    <VisCanvas
      abscissaConfig={abscissaConfig}
      ordinateConfig={ordinateConfig}
      visRatio={abscissaExtent / ordinateExtent}
      title={title}
    >
      <LinearVisGroup>{children}</LinearVisGroup>
    </VisCanvas>
  );
}
```

In H5Web, the need for orthonormal axes is catered for by the high-level `HeatmapVis` component via the `layout` prop. The H5Web viewer has a `Keep ratio` toggle in the toolbar that affects the value of this prop. With NeXus, the units of the axes also affect the value of the `layout` prop.

I'm wondering whether the `layout` prop should be moved down to `VisCanvas`, to basically replace both `visRatio` and `canvasRatio`. The current layouts are `cover`, `contain` and `fill`, knowing that the first two are both for enforcing an orthonormal basis (`contain` just brings the axes against the visualization). Of course, the downside of this is that we lose the ability to control the exact ratio of the visualization.

This leaves two questions:

- Is the `layout` prop as it stands the right API to enable/disable orthonormal basis?
- Is there a need for more control over the ratio of the visualization than just orthonormal or not?

axelboc:
Fix #1275 

⚠️ **Significant breaking changes**

- `[VisCanvas]` Prop `visRatio` is replaced with prop `aspect`, which [follows matplotlib](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_aspect.html):
  - `'auto'` is equivalent to no `visRatio` (i.e. the heatmap's `'fill'` layout);
  - `'equal'` is equivalent to computing `visRatio` from the axis domains to obtain an orthonormal basis (i.e. the heatmap's `cover` layout) -- so now, this ratio computation is performed internally by `VisCanvas` ;
  - and if `aspect` is a `number`, then it is used as the ratio directly (this is not used by the viewer; it's just for flexibility of the lib and consistency with matplotlib).
- `[HeatmapVis, RgbVis]` Prop `layout` is replaced with prop `aspect`, which is given sane defaults (`'equal'`) and forwarded to `VisCanvas`.

I've also replaced the `layout` property in the config stores and toolbars with a `keepRatio` property, since it is all we care about in the viewer.

### DIFF
--- a/apps/storybook/src/SelectToZoom.stories.tsx
+++ b/apps/storybook/src/SelectToZoom.stories.tsx
@@ -25,7 +25,7 @@ const Template: Story<SelectToZoomProps> = (args) => {
     <VisCanvas
       abscissaConfig={{ visDomain: [-5, 5], showGrid: true }}
       ordinateConfig={{ visDomain: [-0.5, 1.5], showGrid: true }}
-      visRatio={keepRatio ? cols / rows : undefined}
+      aspect={keepRatio ? 'equal' : 'auto'}
     >
       <Pan />
       <Zoom />
--- a/apps/storybook/src/HeatmapVis.stories.tsx
+++ b/apps/storybook/src/HeatmapVis.stories.tsx
@@ -123,7 +123,7 @@ export default {
   args: {
     colorMap: 'Viridis',
     scaleType: ScaleType.Linear,
-    layout: 'cover',
+    aspect: 'equal',
     showGrid: true,
   },
   argTypes: {
@@ -140,9 +140,9 @@ export default {
         ScaleType.Sqrt,
       ],
     },
-    layout: {
+    aspect: {
       control: { type: 'inline-radio' },
-      options: ['cover', 'fill'],
+      options: ['auto', 'equal'],
     },
   },
 } as Meta;
--- a/apps/storybook/src/HeatmapVisDisplay.stories.tsx
+++ b/apps/storybook/src/HeatmapVisDisplay.stories.tsx
@@ -15,11 +15,18 @@ const domain = getDomain(dataArray.data);
 
 const Template: Story<HeatmapVisProps> = (args) => <HeatmapVis {...args} />;
 
-export const Fill = Template.bind({});
-Fill.args = {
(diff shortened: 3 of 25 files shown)


apps/storybook/src/SelectToZoom.stories.tsx
```diff
@@ -25,7 +25,7 @@ const Template: Story<SelectToZoomProps> = (args) => {
     <VisCanvas
       abscissaConfig={{ visDomain: [-5, 5], showGrid: true }}
       ordinateConfig={{ visDomain: [-0.5, 1.5], showGrid: true }}
-      visRatio={keepRatio ? cols / rows : undefined}
+      aspect={keepRatio ? 'equal' : 'auto'}
```
axelboc, line 28: The `keepRatio` prop still makes sense for some components.

PR #1287, opened 2022-11-17
Make interaction components smart with regard to canvas aspect

axelboc:
⚠️ **Breaking change**: `DefaultInteractions` and `SelectToZoom` no longer accept a `keepRatio` prop. Each interaction component adjusts its behaviour automatically when used inside a canvas with an `'equal'` or custom `aspect`.

---

So far, the high-level visualisations had to compute a `keepRatio` flag and pass it to the various interaction components (via `DefaultInteractions`). Since #1284, this flag was computed based on the `aspect` prop passed to `VisCanvas`: `keepRatio={aspect !== 'auto'}`. In practice, this meant that consumers of the low-level components could very well mis-configure the interactions - e.g. by passing `aspect="equal"` to `VisCanvas` and then forgetting to disable axial-zoom interactions.

In this PR, I inject `visRatio` (which is computed internally by `VisCanvas`) via `AxisSystemContext` so that each interaction component can access it and configure itself accordingly when used inside a canvas with an equal-aspect ratio or a custom ratio:

- `XAxisZoom`, `YAxisZoom` and `AxialSelectToZoom` all disable themselves automatically;
- `SelectToZoom` configures itself to keep the canvas' ratio on zoom.

I've added stories to test this behaviour.

### DIFF
--- a/packages/lib/src/interactions/AxialSelectToZoom.tsx
+++ b/packages/lib/src/interactions/AxialSelectToZoom.tsx
@@ -1,5 +1,6 @@
 import { useThree } from '@react-three/fiber';
 
+import { useAxisSystemContext } from '../vis/shared/AxisSystemProvider';
 import AxialSelectionTool from './AxialSelectionTool';
 import SelectionRect from './SelectionRect';
 import { useMoveCameraTo } from './hooks';
@@ -13,6 +14,7 @@ interface Props extends CommonInteractionProps {
 function AxialSelectToZoom(props: Props) {
   const { axis, modifierKey, disabled } = props;
 
+  const { visRatio } = useAxisSystemContext();
   const moveCameraTo = useMoveCameraTo();
 
   const { width, height } = useThree((state) => state.size);
@@ -44,7 +46,7 @@ function AxialSelectToZoom(props: Props) {
       axis={axis}
       id={`${axis.toUpperCase()}SelectToZoom`}
       modifierKey={modifierKey}
-      disabled={disabled}
+      disabled={visRatio !== undefined || disabled}
       onSelectionEnd={onSelectionEnd}
     >
       {({ startPoint, endPoint }) => (
--- a/packages/lib/src/interactions/DefaultInteractions.tsx
+++ b/packages/lib/src/interactions/DefaultInteractions.tsx
@@ -16,17 +16,13 @@ export interface DefaultInteractionsConfig {
   zoom?: ZoomProps | false;
   xAxisZoom?: XAxisZoomProps | false;
   yAxisZoom?: YAxisZoomProps | false;
-  selectToZoom?: Omit<SelectToZoomProps, 'keepRatio'> | false;
+  selectToZoom?: SelectToZoomProps | false;
   xSelectToZoom?: Omit<AxialSelectToZoomProps, 'axis'> | false;
   ySelectToZoom?: Omit<AxialSelectToZoomProps, 'axis'> | false;
 }
 
-interface Props extends DefaultInteractionsConfig {
-  keepRatio?: boolean;
-}
-
-function DefaultInteractions(props: Props) {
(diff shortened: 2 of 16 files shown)


packages/lib/src/interactions/AxialSelectToZoom.tsx
```diff
@@ -44,7 +46,7 @@ function AxialSelectToZoom(props: Props) {
       axis={axis}
       id={`${axis.toUpperCase()}SelectToZoom`}
       modifierKey={modifierKey}
-      disabled={disabled}
+      disabled={visRatio !== undefined || disabled}
```
axelboc, line 49: Okay, so this is the first case of a component deciding, by itself, how to behave based on `visRatio`. Here, it's basically disabling itself inside a canvas with an `'equal'` or custom `aspect`.

Note that, thanks to `|| disabled`, the consumer still has the ability to disable the interaction when used inside a canvas with `aspect="auto"`; but the opposite is not true: the consumer cannot force-enable the interaction inside a canvas with an `'equal'` or custom `aspect`, since this use case is not clearly defined/supported.

packages/lib/src/interactions/DefaultInteractions.tsx
```diff
@@ -16,64 +16,45 @@ export interface DefaultInteractionsConfig {
   zoom?: ZoomProps | false;
   xAxisZoom?: XAxisZoomProps | false;
   yAxisZoom?: YAxisZoomProps | false;
-  selectToZoom?: Omit<SelectToZoomProps, 'keepRatio'> | false;
+  selectToZoom?: SelectToZoomProps | false;
```
axelboc, line 19: `SelectToZoomProps` no longer has this prop.

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -9,14 +9,13 @@ import { useMoveCameraTo } from './hooks';
...
-  keepRatio?: boolean;
-}
+type Props = CommonInteractionProps;
 
 function SelectToZoom(props: Props) {
-  const { keepRatio, ...interactionProps } = props;
-
   const context = useAxisSystemContext();
+  const { visRatio } = context;
+  const keepRatio = visRatio !== undefined;
```
axelboc, line 17: `SelectToZoom` now computes `keepRatio` by itself.

axelboc:
No rush about reviewing this 😉

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -9,14 +9,13 @@ import { useMoveCameraTo } from './hooks';
 import type { CommonInteractionProps, Selection } from './models';
 import { getEnclosedRectangle, getRatioRespectingRectangle } from './utils';
 
-interface Props extends CommonInteractionProps {
-  keepRatio?: boolean;
-}
+type Props = CommonInteractionProps;
```
loichuder, line 12: Is the `Props` type still needed ?

loichuder (approved):
I could imagine a case where the `aspect` given is only valid initially and then, the interaction components should not disable themselves. 

But we don't have a usecase for it so I think the PR goes in the right direction.

packages/lib/src/interactions/SelectToZoom.tsx
axelboc: Just seems like a good thing to keep in case we need to add back more props in the future, and also to have one exported component with props = one matching exported props type.

axelboc:
> I could imagine a case where the aspect given is only valid initially and then, the interaction components should not disable themselves.

The needs/implications of this are not clear to me, but conceptually, I would prefer if this was done by the consumer by passing `aspect` only on first render and then removing/changing it back on second render -- i.e. so that the interactions always remain true to the active `visRatio`.

PR #1288, opened 2022-11-17
Inject `canvasSize` and `canvasRatio` via context and simplify `SelectToZoom` logic

axelboc:
`SelectToZoom` was computing the "ratio to respect" from the visible domains. I came to the realisation that this ratio was always the exact same as the ratio of the canvas, which seems obvious now -- we probably missed it because this wasn't the case when the feature was first developed as we had to iterate quite a bit to get to the current behaviour.

So I decided to inject `canvasRatio` via `AxisSystemContext` and use that directly in `SelectToZoom`. I also decided to inject `canvasSize` for convenience, to avoid using both `useThree` and `useAxisSystemContext` in some places.

### DIFF
--- a/packages/lib/src/interactions/SelectToZoom.tsx
+++ b/packages/lib/src/interactions/SelectToZoom.tsx
@@ -11,24 +11,24 @@ import { getEnclosedRectangle, getRatioRespectingRectangle } from './utils';
 type Props = CommonInteractionProps;
 
 function SelectToZoom(props: Props) {
-  const context = useAxisSystemContext();
-  const { visRatio, canvasRatio } = context;
+  const { canvasSize, canvasRatio, visRatio, dataToWorld } =
+    useAxisSystemContext();
+
+  const { width, height } = canvasSize;
   const keepRatio = visRatio !== undefined;
 
-  const moveCameraTo = useMoveCameraTo();
-
-  const { width, height } = useThree((state) => state.size);
   const camera = useThree((state) => state.camera);
+  const moveCameraTo = useMoveCameraTo();
 
   function onSelectionEnd(selection: Selection) {
     const { startPoint: dataStartPoint, endPoint: dataEndPoint } = selection;
 
     // Work in world coordinates as we need to act on the world camera
-    const [startPoint, endPoint] = (
-      keepRatio
-        ? getRatioRespectingRectangle(dataStartPoint, dataEndPoint, canvasRatio)
-        : [dataStartPoint, dataEndPoint]
-    ).map(context.dataToWorld);
+    const [startPoint, endPoint] = getRatioRespectingRectangle(
+      dataStartPoint,
+      dataEndPoint,
+      keepRatio ? canvasRatio : undefined
+    ).map(dataToWorld);
 
     if (startPoint.x === endPoint.x && startPoint.y === endPoint.y) {
       return;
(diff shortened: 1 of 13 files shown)


packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -13,33 +12,20 @@ type Props = CommonInteractionProps;
 
 function SelectToZoom(props: Props) {
   const context = useAxisSystemContext();
-  const { visRatio } = context;
+  const { canvasSize, canvasRatio, visRatio } = context;
```
loichuder, line 15: `dataToWorld` may also destructured from `context`. 

Also,  keeping the `context` variable is no longer useful since `getVisibleDomains` was removed.

packages/lib/src/vis/shared/AxisSystemProvider.tsx
```diff
@@ -45,8 +47,11 @@ function AxisSystemProvider(props: PropsWithChildren<Props>) {
     children,
   } = props;
 
-  const availableSize = useThree((state) => state.size);
-  const visSize = getSizeToFit(availableSize, visRatio);
+  const canvasSize = useThree((state) => state.size);
```
loichuder, line 50: Does this change anything with regards to rerenders ?

I mean, now that `canvasSize` is in the `context`, it means a change in `size` will change the context and therefore all the consumers, right? But I don't think this is different from the previous behaviour since `visSize` was depending on `canvasSize` and in the context.

packages/lib/src/vis/shared/TooltipMesh.tsx
```diff
@@ -18,10 +17,9 @@ interface Props {
 function TooltipMesh(props: Props) {
   const { guides, renderTooltip, size } = props;
 
-  const { width, height } = useThree((state) => state.size);
-
   // Scales to compute data coordinates from unprojected mesh coordinates
```
loichuder, line 20: The comment may be obsolete and be removed

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -13,33 +12,20 @@ type Props = CommonInteractionProps;
...
 
   function onSelectionEnd(selection: Selection) {
     const { startPoint: dataStartPoint, endPoint: dataEndPoint } = selection;
-    const dataRatio = getDataRatio();
 
     // Work in world coordinates as we need to act on the world camera
     const [startPoint, endPoint] = (
       keepRatio
-        ? getRatioRespectingRectangle(dataStartPoint, dataEndPoint, dataRatio)
+        ? getRatioRespectingRectangle(dataStartPoint, dataEndPoint, canvasRatio)
```
loichuder, line 28: Looking back at these, we might be able to simplify this by allowing `getRatioRespectingRectangle` to accept an `undefined` ratio that triggers an early return.

loichuder (approved):

packages/lib/src/vis/shared/AxisSystemProvider.tsx
axelboc: Precisely, no change to the number of re-renders, since:

1.  `AxisSystemProvider` used to already re-render on canvas `size` changes;
2. It doesn't retrieve anything else from `useThree` and doesn't have any `useState` that could trigger extra renders.

packages/lib/src/interactions/SelectToZoom.tsx
axelboc: Excellent idea. I've also re-used the `.map(scale)` pattern in `RatioSelectionRect`.

PR #1291, opened 2022-11-17
Rename `AxisSystemProvider` to `VisCanvasProvider`

### LINKED ISSUE #1289: Rename `AxisSystemProvider`
We've been talking about it for a while now. Let's start brainstorming some actual names. Here is what needs to be renamed, I think:

- `AxisSystemContext`
- `AxisSystemProvider`
- `useAxisSystemContext` (user-facing)

Here are some suggestions:

- `VisCanvasContext`
- `VisCanvasProvider`
- `useVisCanvas` (preferred), `useVisCanvasContext` (too long), `useCanvasContext` (ok, but can be confused with R3F canvas), `useVisContext` (could expect more that just canvas-related things in there, which is not the idea)

axelboc:
Fix #1289

### DIFF
--- a/packages/lib/src/vis/shared/RatioEnforcer.tsx
+++ b/packages/lib/src/vis/shared/RatioEnforcer.tsx
@@ -2,14 +2,10 @@ import { useThree } from '@react-three/fiber';
 import { useEffect } from 'react';
 
 import { useMoveCameraTo } from '../../interactions/hooks';
+import { useVisCanvasContext } from './VisCanvasProvider';
 
-interface Props {
-  visRatio: number | undefined;
-}
-
-function RatioEnforcer(props: Props) {
-  const { visRatio } = props;
-
+function RatioEnforcer() {
+  const { visRatio } = useVisCanvasContext();
   const camera = useThree((state) => state.camera);
   const moveCameraTo = useMoveCameraTo();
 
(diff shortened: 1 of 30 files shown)


packages/lib/src/vis/shared/RatioEnforcer.tsx
```diff
@@ -2,14 +2,10 @@ import { useThree } from '@react-three/fiber';
...
 
-interface Props {
-  visRatio: number | undefined;
-}
-
-function RatioEnforcer(props: Props) {
-  const { visRatio } = props;
-
+function RatioEnforcer() {
+  const { visRatio } = useVisCanvasContext();
```
axelboc, line 8: Getting `visRatio` from context to remove the prop (we do this in `AxisSystem` too.

packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -68,26 +68,24 @@ function VisCanvas(props: PropsWithChildren<Props>) {
...
-              <ThresholdAdjuster value={raycasterThreshold} />
-            )}
-          </InteractionsProvider>
-        </AxisSystemProvider>
+          <AxisSystem
+            axisOffsets={axisOffsets}
+            title={title}
+            showAxes={showAxes}
+          />
+          <InteractionsProvider>{children}</InteractionsProvider>
```
axelboc, line 82: Moving `InteractionsProvider` around `children`, since the other components don't need it.

loichuder (approved):
Thanks for rolling with `useVisCanvasContext`. I know it is longer but I prefer to be explicit about the fact that it is a context and not any hook.

PR #1294, opened 2022-11-21
Upgrade dependencies

axelboc:
- [TypeScript 4.9](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/)
  - The `satisfies` keyword is going to be super useful! I've added an example in `packages/lib/src/vis/heatmap/interpolators.ts` (commented out for now since Prettier doesn't understand this syntax yet)
- ESLint [8.26](https://github.com/eslint/eslint/releases/tag/v8.26.0) and [8.27](https://github.com/eslint/eslint/releases/tag/v8.27.0)
- [eslint-config-galex](https://github.com/ljosberinn/eslint-config-galex/releases/tag/v4.3.2)
- [react-window](https://github.com/bvaughn/react-window/blob/master/CHANGELOG.md#188)
- [zustand](https://github.com/pmndrs/zustand/releases/tag/v4.1.4)
- [vite 3.2.1, 3.2.2, 3.2.4](https://github.com/vitejs/vite/blob/v3.2.4/packages/vite/CHANGELOG.md)
- [@rest-hookz/web](https://github.com/react-hookz/web/releases?page=3) from 3.2.3 (included) to 19.2.0!
  - Useful to us:
    - [x] The API of [`useLocalStorageValue`](https://react-hookz.github.io/web/?path=/docs/side-effect-uselocalstoragevalue--example) is now more straight-forward. I had to refactor the `useActiveVis` hook in `packages/app/src/visualizer/hooks.ts`.
    - [x] [`useTimeoutEffect`](https://react-hookz.github.io/web/?path=/docs/lifecycle-usetimeouteffect--example) - obviously useful to replace our own `useTimeout`, which I've now done
    - [x] [`useToggle`](https://react-hookz.github.io/web/?path=/docs/state-usetoggle--example)'s setter ignores React events, so we can now do `onClick={toggle}` safely instead of `onClick={() => toggle}`. I've refactored, but only when the `toggle` function was called in the same file as `useToggle` -- if the function was passed as a prop, like `<ToggleBtn onToggle={toggle}>`, I left it alone.
  - Potentially useful (to keep in mind for future refactors and new features):
    - [`useList`](https://react-hookz.github.io/web/?path=/docs/state-uselist--example) - for manipulating an array state; might be useful for `useDimMappingState`?
    - [`useControlledRerenderState`](https://react-hookz.github.io/web/?path=/docs/state-usecontrolledrerenderstate--example) - this is cool: it allows updating a state without re-rendering. Internally, it uses a ref instead of a state; when the setter is called and returns `false`, the ref is updated and nothing else happens; when the setter is called and returns anything else than `false`, the ref is updated and a re-render is artificially triggered. If something else re-renders the component where the hook is used, then the latest value is returned.
    - [`useDeepCompareMemo`](https://react-hookz.github.io/web/?path=/docs/miscellaneous-usedeepcomparememo--example) - like `useMemo` but compares the deps deeply
    - [`useCustomCompareMemo`](https://react-hookz.github.io/web/?path=/docs/miscellaneous-usecustomcomparememo--example) - like `useMemo` but with a custom deps comporator
    - [`useHookableRef`](https://react-hookz.github.io/web/?path=/docs/miscellaneous-usehookableref--example) - might be useful to implement a stateful version of `useRef` at some point (cf. https://github.com/react-hookz/web/discussions/860)
    - [`useRenderCount`](https://react-hookz.github.io/web/?path=/docs/state-userendercount--example) - maybe useful when debugging (but note that `StrictMode` renders everything twice!)
  - Less useful, but worth knowing:
    - [`useIntervalEffect`](https://react-hookz.github.io/web/?path=/docs/lifecycle-useintervaleffect--example) - runs an effect every x ms
    - [`useFunctionalState`](https://react-hookz.github.io/web/?path=/docs/state-usefunctionalstate--page) - like `useState` but returns a function to read the state (to avoid subtle bugs in nested functions, but in most cases, it's fixable with `setState(state => ...)`)
    - [`useMeasure`](https://react-hookz.github.io/web/?path=/docs/sensor-usemeasure--example) now uses a stateful ref internally (with `useHookableRef`) in case the element to measure is rendered conditionally, and also no longer returns a full content rect (`top`, `left`, etc.-- only `width` and `height`


loichuder (approved):
:+1:

Issue #1276, opened 2022-11-10
[VisCanvas] Extract axes/grid rendering logic


axelboc:
I'd like to investigate the feasibility of switching to an API like this:

```tsx
<VisCanvas>
  <Title>My Plot</Title>
  <Axes abscissaLabel="X" />
  <Grid vertical={false} />
</VisCanvas>
```

The idea is to make some of `VisCanvas`'s props such as `showAxes` and `AxisConfig.showGrid` obsolete. The goal is to make the API simpler, but also to remove a lot of internal rendering logic from `VisCanvas`. Toggling the axes, the grid, the plot title or the axis labels becomes a matter of rendering the corresponding components conditionally (or adding ternaries to some of the props), and those components can have their own dedicated documentation sections on the Storybook site.

The main challenge is of course dealing with axis offsets.

axelboc:
Decoupling grids from axes might not be a good idea: https://github.com/silx-kit/h5web/issues/1429#issuecomment-1531368521 – it seems simpler to have one component for each axis (x/y) and keep rendering the corresponding grid inside.

axelboc:
Got a prototype working, but it uncovered an issue: with a declarative API, there's no easy way to "pre-allocate" space next to the canvas for the axes, and since the axes are rendered _after_ the canvas (because of `Html`), the canvas renders first and fills the space then shrinks when the axes are added to the DOM. The effect is pretty awful.

axelboc:
Closed in favour of #1276

loichuder:
> Closed in favour of #1276

I guess you meant https://github.com/silx-kit/h5web/issues/1469 ?

PR #1295, opened 2022-11-22
Allow to hide left axis in histogram

### LINKED ISSUE #1279: [Histogram] Make possible to hide the left axis

In the same vein of #1257, it may be useful for some users to be able to hide the left axis to give more space to the actual histogram. 

Adding a prop to `Histogram` ? Perhaps we could also think about giving even more control on the plot (though `AxisConfigs` ?)

loichuder:
Fix #1279 

In `Histogram`:
image

In `DomainSlider`:
image

The API is a bit clunky but with #1276 in the works, I thought we should not replicate the axis config API. Feedback welcome though.

### DIFF
--- a/apps/storybook/src/DomainSlider.stories.tsx
+++ b/apps/storybook/src/DomainSlider.stories.tsx
@@ -68,6 +68,19 @@ HistogramWithColorMap.args = {
   },
 };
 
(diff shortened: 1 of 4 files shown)


axelboc (approved):
God-enough for now :+1:

Issue #1296, opened 2022-11-24
Minimum zoom threshold


axelboc:
We need to add a threshold to the zoom interaction to avoid zooming on very small areas by mistake - cf. https://gitlab.esrf.fr/ui/daiquiri-tomo/-/issues/220#note_219512

PR #1297, opened 2022-11-24
Avoid duplicate computations in `SelectionTool`

axelboc:
For #1296, I'm gonna need access to the selection start/end points in HTML space. Before I can get there, I need to refactor a few things so it doesn't become a huge mess. I'm gonna try to split things up to avoid regressions and make reviewing easier.

First up, I'm storing the entire `selection` in a state inside `SelectionTool` instead of just the `endPoint`. This avoids recomputing `worldStart/EndPoint` during render.

I'm also fixing a very subtle bug in the persisted selection story, where the active and persisted selections would both be visible for one render.

### DIFF
--- a/apps/storybook/src/Selection.stories.tsx
+++ b/apps/storybook/src/Selection.stories.tsx
@@ -173,7 +173,9 @@ export const Persisted: Story<TemplateProps> = (args) => {
         onSelectionEnd={setPersistedSelection}
         modifierKey={selectionModifierKey}
       >
-        {(selection) => <SelectionComponent {...selection} />}
+        {(selection) =>
+          !persistedSelection && <SelectionComponent {...selection} />
+        }
       </SelectionTool>
       {persistedSelection && (
         <SelectionComponent
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -1,6 +1,6 @@
 import { useKeyboardEvent, useRafState } from '@react-hookz/web';
 import { useThree } from '@react-three/fiber';
-import type { ReactElement } from 'react';
+import type { ReactNode } from 'react';
 import { useCallback, useState } from 'react';
 import type { Vector2 } from 'three';
 
@@ -19,10 +19,9 @@ interface Props extends CommonInteractionProps {
   onSelectionStart?: () => void;
   onSelectionChange?: (points: Selection) => void;
   onSelectionEnd?: (points: Selection) => void;
-  children: (points: Selection) => ReactElement;
+  children: (points: Selection) => ReactNode;
 }
 
-// eslint-disable-next-line sonarjs/cognitive-complexity
 function SelectionTool(props: Props) {
   const {
     id = 'Selection',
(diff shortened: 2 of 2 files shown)


apps/storybook/src/Selection.stories.tsx
```diff
@@ -173,7 +173,9 @@ export const Persisted: Story<TemplateProps> = (args) => {
         onSelectionEnd={setPersistedSelection}
         modifierKey={selectionModifierKey}
       >
-        {(selection) => <SelectionComponent {...selection} />}
+        {(selection) =>
+          !persistedSelection && <SelectionComponent {...selection} />
```
axelboc, line 177: 1. `onSelectionEnd` is called with the selection to persist.
2. The story re-renders with a persisted selection => **both the persisted and active selections are visible at once**
3. The internal active selection state of `SelectionTool` is cleared
4. `SelectionTool` re-renders and returns `null` because there's no longer an active selection.

I couldn't find an easy to way for `SelectionTool` to re-render and return `null` _before_ `onSelectionEnd` gets called, so instead I work around the issue with a simple condition in the render prop.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -19,7 +19,7 @@ interface Props extends CommonInteractionProps {
   onSelectionStart?: () => void;
   onSelectionChange?: (points: Selection) => void;
   onSelectionEnd?: (points: Selection) => void;
-  children: (points: Selection) => ReactElement;
+  children: (points: Selection) => ReactNode;
```
axelboc, line 22: This, along with the extra fragment around `children()` below (for TS), allows the render prop to return `null`, `false`, etc.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -36,18 +36,21 @@ function SelectionTool(props: Props) {
 
   const camera = useThree((state) => state.camera);
   const { dataToWorld, worldToData } = useVisCanvasContext();
+
+  const [startPoint, setStartPoint] = useState<Vector2>();
+  const [selection, setSelection] = useRafState<Selection | undefined>(
```
axelboc, line 41: I now store the entire `selection` instead of just the end point. I know this means the `startPoint` is technically stored twice, but I don't think it's a problem. Especially since, for #1296, I'm gonna need to store more than just the start point in data space.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -142,25 +143,17 @@ function SelectionTool(props: Props) {
...
     { event: 'keydown' }
   );
 
-  if (!startPoint || !endPoint || !isModifierKeyPressed) {
+  if (!selection || !isModifierKeyPressed) {
     return null;
   }
 
-  const worldEndPoint = dataToWorld(endPoint);
-  const worldStartPoint = dataToWorld(startPoint);
```
axelboc, line 156: The main reason for all this (but there is more coming): we no longer compute the selection points in world space twice in a row on every mouse move. Also, we can pass the entire selection object below which saves a few lines.

axelboc:
Ooh, as a nice bonus, the cognitive complexity warning is now gone.

packages/lib/src/interactions/SelectionTool.tsx
loichuder: Would it be possible to work with the `startPoint` in `selection` to remove the `startPoint` state ?

apps/storybook/src/Selection.stories.tsx
loichuder: Not a huge problem but good to be aware of it

loichuder (approved):
Great refacto

axelboc:
> Would it be possible to work with the startPoint in selection to remove the startPoint state ?

Yeah, I'm planning to investigate this next.

PR #1298, opened 2022-11-24
Prevent `onSelectionChange` from affecting selection tool performance

axelboc:
`onSelectionChange` was called on every pointer move event, so it had the potential to affect the performance of the selection tool. I'm moving the call inside a `useEffect` so that it happens asynchronously, after the RAF selection state is updated and after the selection tool and its render prop have had time to re-render.

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -1,7 +1,7 @@
 import { useKeyboardEvent, useRafState } from '@react-hookz/web';
 import { useThree } from '@react-three/fiber';
 import type { ReactNode } from 'react';
-import { useCallback, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import type { Vector2 } from 'three';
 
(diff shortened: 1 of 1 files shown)


loichuder:
Sorry for my ignorance but I don't understand how this improves performance :sweat_smile: 

With this implementation:
- On pointer move, the state of `SelectionTool` changes, triggering a rerender.
- The rerender of `SelectionTool` triggers the execution of the `useEffect` and therefore of `onSelectionChange`.

Therefore, in effect, `onSelectionChange` is still called on every each pointer move event ?

axelboc:
Yes, but it's now called asynchronously:

1. outside of the current animation frame thanks to `useRafState`, which means that if the browser is busy and a selection state update is skipped, `onSelectionChange` is no longer called either;
2. _and_ on the next render thanks to `useEffect`, which means that if the consumer sets a state in `onSelectionState`, the parent component gets re-rendered _after_ the active selection (which is the most time-sensitive thing) instead of before. And since I use `useEffect` and not `useLayoutEffect`, the re-rendered active selection has time to actually be painted by the browser, which maximises the perceived responsiveness of the selection for the user.

PR #1299, opened 2022-11-24
Store full start event in state in `SelectionTool`

axelboc:
The goal is twofold:

1. Keep a reference to the start `htmlPt` for #1296 to avoid having to compute it back from the data point.
2. Keep a reference to the start `worldPt` to avoid having to compute it back from the data point in both `onPointerMove` and `onPointerUp`.

I also fix a couple of naming issues. See comments.

### DIFF
--- a/packages/lib/src/interactions/AxialSelectionTool.tsx
+++ b/packages/lib/src/interactions/AxialSelectionTool.tsx
@@ -31,21 +31,21 @@ function AxialSelectionTool(props: Props) {
       context
     );
 
-    const { startPoint: mouseStartPoint, endPoint: mouseEndPoint } = selection;
-    const startPoint =
+    const { startPoint, endPoint } = selection;
+    const axialStartPoint =
       axis === 'x'
-        ? new Vector2(mouseStartPoint.x, yVisibleDomain[0])
-        : new Vector2(xVisibleDomain[0], mouseStartPoint.y);
-    const endPoint =
+        ? new Vector2(startPoint.x, yVisibleDomain[0])
+        : new Vector2(xVisibleDomain[0], startPoint.y);
+    const axialEndPoint =
       axis === 'x'
-        ? new Vector2(mouseEndPoint.x, yVisibleDomain[1])
-        : new Vector2(xVisibleDomain[1], mouseEndPoint.y);
+        ? new Vector2(endPoint.x, yVisibleDomain[1])
+        : new Vector2(xVisibleDomain[1], endPoint.y);
 
     return {
-      startPoint,
-      endPoint,
-      worldStartPoint: context.dataToWorld(startPoint),
-      worldEndPoint: context.dataToWorld(endPoint),
+      startPoint: axialStartPoint,
+      endPoint: axialEndPoint,
+      worldStartPoint: context.dataToWorld(axialStartPoint),
+      worldEndPoint: context.dataToWorld(axialEndPoint),
     };
   }
 
--- a/packages/lib/src/interactions/RatioSelectionRect.tsx
+++ b/packages/lib/src/interactions/RatioSelectionRect.tsx
@@ -15,22 +15,19 @@ function RatioSelectionRect(props: Props) {
   const { startPoint, endPoint, ratio, ...svgProps } = props;
   const { visSize, dataToWorld, worldToData } = useVisCanvasContext();
 
-  const [dataStartPoint, dataEndPoint] = getRatioRespectingRectangle(
-    startPoint,
-    endPoint,
-    ratio
-  ).map(dataToWorld);
+  const [ratioWorldStartPoint, ratioWorldEndPoint] =
+    getRatioRespectingRectangle(startPoint, endPoint, ratio).map(dataToWorld);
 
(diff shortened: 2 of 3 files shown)


packages/lib/src/interactions/AxialSelectionTool.tsx
```diff
@@ -31,21 +31,21 @@ function AxialSelectionTool(props: Props) {
       context
     );
 
-    const { startPoint: mouseStartPoint, endPoint: mouseEndPoint } = selection;
```
axelboc, line 34: `mouse` made it sound like it was a point in HTML space. I went with renaming the other variables to `axialStartPoint` and `axialEndPoint`. Let me know what you think.

packages/lib/src/interactions/RatioSelectionRect.tsx
```diff
@@ -15,22 +15,19 @@ function RatioSelectionRect(props: Props) {
   const { startPoint, endPoint, ratio, ...svgProps } = props;
   const { visSize, dataToWorld, worldToData } = useVisCanvasContext();
 
-  const [dataStartPoint, dataEndPoint] = getRatioRespectingRectangle(
-    startPoint,
-    endPoint,
-    ratio
-  ).map(dataToWorld);
+  const [ratioWorldStartPoint, ratioWorldEndPoint] =
+    getRatioRespectingRectangle(startPoint, endPoint, ratio).map(dataToWorld);
```
axelboc, line 19: Woops, I mixed things up in https://github.com/silx-kit/h5web/commit/959cfe96bac3fcab0e925134f7062d04afe5d559#diff-64ebff252b1278ccdd5df1c8180f14172dba22e526031abde6ae8d0283108e51. The points go through `dataToWorld`, so they are world points here and data points below... I've renamed again and added prefixes back for clarity.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -16,10 +18,10 @@ import { boundPointToFOV, getModifierKeyArray } from './utils';
 
 interface Props extends CommonInteractionProps {
   id?: string;
-  onSelectionStart?: () => void;
-  onSelectionChange?: (points: Selection) => void;
-  onSelectionEnd?: (points: Selection) => void;
-  children: (points: Selection) => ReactNode;
+  onSelectionStart?: (evt: CanvasEvent<PointerEvent>) => void;
```
axelboc, line 21: I'm now passing the start event through to the callback. No reason not to, really.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -34,9 +36,9 @@ function SelectionTool(props: Props) {
   } = props;
 
   const camera = useThree((state) => state.camera);
-  const { dataToWorld, worldToData } = useVisCanvasContext();
+  const { worldToData } = useVisCanvasContext();
 
-  const [startPoint, setStartPoint] = useState<Vector2>();
+  const [startEvt, setStartEvt] = useState<CanvasEvent<PointerEvent>>();
```
axelboc, line 41: This is the main change: I'm now keeping the entire start event. Relying on a single `selection` state was too confusing. It would have required making the points in the `selection` object optional, which introduced a whole lot of complexity.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -50,84 +52,76 @@ function SelectionTool(props: Props) {
     disabled,
   });
 
+  const computeSelection = useCallback(
```
axelboc, line 55: This code was shared between `onPointerMove` and `onPointerUp`, so I extracted it.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -50,84 +52,76 @@ function SelectionTool(props: Props) {
...
+    (worldEndPt: Vector3) => {
+      assertDefined(startEvt);
+      const { dataPt: startPoint, worldPt: worldStartPoint } = startEvt;
+      const boundWorldEndPoint = boundPointToFOV(worldEndPt, camera);
+
+      return {
+        startPoint,
+        endPoint: worldToData(boundWorldEndPoint),
+        worldStartPoint: new Vector2(worldStartPoint.x, worldStartPoint.y),
```
axelboc, line 64: I can now use `worldPt` almost directly -- just need to convert it to `Vector2`. I have an idea to fix this for another PR.

packages/lib/src/interactions/RatioSelectionRect.tsx
```diff
@@ -15,22 +15,19 @@ function RatioSelectionRect(props: Props) {
...
 
-  const [worldStartPoint, worldEndPoint] = clampRectangleToVis(
-    dataStartPoint,
-    dataEndPoint,
+  const [clampedStartPoint, clampedEndPoint] = clampRectangleToVis(
```
loichuder, line 21:
```suggestion
  const [clampedDataStartPoint, clampedDataEndPoint] = clampRectangleToVis(
```
following your previous namings ?

loichuder (approved):
Good for me :+1:

PR #1300, opened 2022-11-24
Always use `Vector3` for world/camera and `Vector2` for data/html

axelboc:
I thought I'd try to bring the vector conversions under control a bit... Basically we need `Vector3` for camera projections, so I thought we could enforce the use of `Vector3` for points in world/camera space. In data/html space, it seems simpler to use `Vector2` as the `z` index makes less sense (in 2D at least...)

### DIFF
--- a/packages/lib/src/interactions/AxialSelectToZoom.tsx
+++ b/packages/lib/src/interactions/AxialSelectToZoom.tsx
@@ -38,7 +38,7 @@ function AxialSelectToZoom(props: Props) {
     camera.scale.set(zoomRect.width / width, zoomRect.height / height, 1);
     camera.updateMatrixWorld();
 
-    moveCameraTo(zoomRectCenter.x, zoomRectCenter.y);
+    moveCameraTo(zoomRectCenter);
   }
 
   return (
--- a/packages/lib/src/interactions/utils.ts
+++ b/packages/lib/src/interactions/utils.ts
@@ -1,20 +1,19 @@
 import type { Camera } from '@react-three/fiber';
 import { clamp } from 'lodash';
-import type { Vector3 } from 'three';
-import { Vector2 } from 'three';
+import { Vector2, Vector3 } from 'three';
 
 import type { Size } from '../vis/models';
 import { getWorldFOV } from '../vis/utils';
 import type { ModifierKey } from './models';
 
 export function boundPointToFOV(
-  unboundedPoint: Vector2 | Vector3,
+  unboundedPoint: Vector3,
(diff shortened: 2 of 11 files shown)


packages/lib/src/interactions/AxialSelectToZoom.tsx
```diff
@@ -38,7 +38,7 @@ function AxialSelectToZoom(props: Props) {
     camera.scale.set(zoomRect.width / width, zoomRect.height / height, 1);
     camera.updateMatrixWorld();
 
-    moveCameraTo(zoomRectCenter.x, zoomRectCenter.y);
+    moveCameraTo(zoomRectCenter);
```
axelboc, line 41: `moveCameraTo` now accepts a `Vector3`

packages/lib/src/vis/shared/VisCanvasProvider.tsx
```diff
@@ -56,10 +56,17 @@ function VisCanvasProvider(props: PropsWithChildren<Props>) {
   const abscissaScale = getCanvasScale(abscissaConfig, visSize.width);
   const ordinateScale = getCanvasScale(ordinateConfig, visSize.height);
 
-  const worldToData = (vec: Vector2 | Vector3) =>
-    new Vector2(abscissaScale.invert(vec.x), ordinateScale.invert(vec.y));
-  const dataToWorld = (vec: Vector2 | Vector3) =>
-    new Vector2(abscissaScale(vec.x), ordinateScale(vec.y));
+  const dataToWorld = useCallback(
```
axelboc, line 59: Wrapping in `useCallback` here, after all, since everything else in the provider is memoised... It doesn't cost much anyway.

packages/lib/src/interactions/utils.ts
```diff
@@ -66,24 +65,25 @@ export function getEnclosedRectangle(startPoint: Vector2, endPoint: Vector2) {
 }
 
 export function clampPositionToArea(
-  center: Vector2,
+  center: Vector3,
```
loichuder, line 68: Just an idea: it may also be useful to communicate in which space these functions are supposed to act.

loichuder (approved):
Great idea, the calls and signatures of functions are much more consistent :100:

packages/lib/src/interactions/utils.ts
axelboc: Good call :+1: 

I was planning on refactoring these utilities anyway, so I'll keep this for another PR.

PR #1302, opened 2022-11-28
Add camera utilities `worldToHtml`, `htmlToWorld` and `htmlToData`

axelboc:
We already had `dataToHtml`; I'm just adding the missing ones. This helps to extract some of the vector conversions and camera projections from the `useCanvasEvents`.

### DIFF
--- a/apps/storybook/src/Utilities.stories.mdx
+++ b/apps/storybook/src/Utilities.stories.mdx
@@ -194,7 +194,7 @@ Register event listeners on the `canvas` element. Useful to implement custom int
 useCanvasEvents(callbacks: CanvasEventCallbacks): void
 
 const handlePointerDown = useCallback((evt: CanvasEvent<PointerEvent>) => {
-  const { htmlPt, cameraPt, worldPt, dataPt, sourceEvent } = evt;
+  const { htmlPt, worldPt, dataPt, sourceEvent } = evt;
   // ...
 }, []);
 
@@ -234,12 +234,20 @@ To avoid unnecessary re-renders, it performs a deep comparison before updating t
 The utilities documented in this section require access to the R3F `camera` object. They are typically meant to be called inside
 [`useCameraState`](https://h5web-docs.panosc.eu/?path=/story/utilities--page#usecamerastate) callbacks or DOM event listeners.
 
-#### worldToCamera
+#### worldToHtml
 
-Convert a point from world space to camera space
+Convert a point from world space to HTML space
 
 ```ts
-dataToHtml(camera: Camera, worldPt: Vector2 | Vector3): Vector2
+worldToHtml(camera: Camera, context: VisCanvasContextValue, worldPt: Vector3): Vector2
+```
+
+#### htmlToWorld
+
+Convert a point from HTML space to world space
+
+```ts
+htmlToWorld(camera: Camera, context: VisCanvasContextValue, htmlPt: Vector2): Vector3
 ```
 
 #### dataToHtml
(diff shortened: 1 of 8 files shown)


packages/lib/src/interactions/hooks.ts
```diff
@@ -110,28 +110,22 @@ export function useZoomOnWheel(
...
-      const htmlPt = new Vector3(offsetX, offsetY, 0);
-      const cameraPt = htmlPt.clone().applyMatrix4(cameraToHtmlMatrixInverse);
-      const worldPt = cameraPt.clone().unproject(camera);
-      const dataPt = worldToData(worldPt);
+      const htmlPt = new Vector2(offsetX, offsetY);
+      const worldPt = htmlToWorld(camera, context, htmlPt);
+      const dataPt = context.worldToData(worldPt);
 
-      return {
-        htmlPt: new Vector2(htmlPt.x, htmlPt.y),
```
axelboc, line 127: This vector conversion is now abstracted inside `htmltoWorld` (more specifically `htmlToCamera`, which is an internal utilities injected via `VisCanvasProvider`)

packages/lib/src/vis/shared/VisCanvasProvider.tsx
```diff
@@ -15,13 +15,12 @@ export interface VisCanvasContextValue {
...
-  worldToData: (vec: Vector3) => Vector2;
+  dataToWorld: (dataPt: Vector2) => Vector3;
+  worldToData: (worldPt: Vector3) => Vector2;
 
   // For internal use only
-  cameraToHtmlMatrix: Matrix4;
-  cameraToHtmlMatrixInverse: Matrix4;
-  cameraToHtml: (vec: Vector3) => Vector2;
+  cameraToHtml: (cameraPt: Vector3) => Vector2;
+  htmlToCamera: (htmlPt: Vector2) => Vector3;
```
axelboc, line 23: New internal `htmlToCamera` utility here, which allows me to no longer inject the matrices.

apps/storybook/src/Utilities.stories.mdx
```diff
@@ -194,7 +194,7 @@ Register event listeners on the `canvas` element. Useful to implement custom int
 useCanvasEvents(callbacks: CanvasEventCallbacks): void
 
 const handlePointerDown = useCallback((evt: CanvasEvent<PointerEvent>) => {
-  const { htmlPt, cameraPt, worldPt, dataPt, sourceEvent } = evt;
```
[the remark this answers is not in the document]
loichuder: Fine. `cameraToHtml` is used in `braggy` but it can be replaced by the new utilities easily

packages/lib/src/vis/utils.ts
```diff
@@ -350,19 +350,39 @@ export function toArray(arr: NumArray): number[] {
...
+  const cameraPt = worldPt.clone().project(camera);
+  return context.cameraToHtml(cameraPt);
+}
+
+export function htmlToWorld(
+  camera: Camera,
+  context: VisCanvasContextValue,
+  htmlPt: Vector2
+) {
+  return context.htmlToCamera(htmlPt).unproject(camera);
```
loichuder, line 367: I don't know if you did it but it would be nice to check that `worldToHtml(htmlToWorld(vec))` gives the same vector (more or less due to numerical precision)

packages/lib/src/vis/utils.ts
```diff
@@ -350,19 +350,39 @@ export function toArray(arr: NumArray): number[] {
...
+  return worldToHtml(camera, context, worldPt);
+}
+
+export function htmlToData(
+  camera: Camera,
+  context: VisCanvasContextValue,
+  htmlPt: Vector2
+): Vector2 {
+  const worldPt = htmlToWorld(camera, context, htmlPt);
+  return context.worldToData(worldPt);
```
loichuder, line 385: Or even better `dataToHtml(htmlToData(vec))` to test the full chain of transformations

loichuder (approved):
:100:

packages/lib/src/vis/utils.ts
axelboc: I just tried and it does indeed give back roughly the same vector (with some extra tiny decimals as expected).

Issue #1292, opened 2022-11-20
surface plot in 3D


kevinxiaomu:
Does h5web supports ploting 3d surfaces? Some thing like the link below:

https://ts1.cn.mm.bing.net/th/id/R-C.7a82fd9d2f59afa278ffbf1b2d9acd57?rik=9fpfi6I4heIVzw&pid=ImgRaw&r=0

kevinxiaomu:
Do you think this is difficult to implement? I try to implement this, is that possible for you to provide some quick instructions?

在 2022年11月21日，16:10，Loïc Huder ***@***.***> wrote：

﻿

No, the closest vis is the Heatmap that has the color mapping but without 3d rendering.

This is something we might implement in the future if there is enough interest for this 🙂

loichuder:
> Do you think this is difficult to implement?

Depends. In H5Web, honestly yes :sweat_smile: ! We have a lot of components that rely on the postulate that the view is 2D.

But you could try to come up with your own stand-alone implementation with the libraries we use in H5Web: 
- [react-three-fiber](https://github.com/pmndrs/react-three-fiber) for the 3D rendering
- [visx](https://airbnb.io/visx/docs/axis) for the axes and scales

If you do, we could then try to integrate it to H5Web. 

If this appears too overwhelming, do not worry, I may tackle this myself in the future.

loichuder:
@kevinxiaomu We have a first implementation in https://github.com/silx-kit/h5web/pull/1303

Before we continue improving it, could you elaborate on your usecase ? 

For example, would you want to use this surface plot to display HDF5 datasets in the viewer or to use it as a visualization component in other React apps ?

kevinxiaomu:
> @kevinxiaomu We have a first implementation in #1303
> 
> Before we continue improving it, could you elaborate on your usecase ?
> 
> For example, would you want to use this surface plot to display HDF5 datasets in the viewer or to use it as a visualization component in other React apps ?

I would like to use this surface plot to display hdf5 datasets in the web viewer. Something like this:

[...]

PR #1303, opened 2022-11-28
Prototype surface plot component

loichuder:
image

For https://github.com/silx-kit/h5web/issues/1292.

There is a lot of room for improvements but I thought I would ask for a review before it gets too bloated.

### DIFF
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -24,6 +24,7 @@
     "@h5web/lib": "workspace:*",
     "@h5web/shared": "workspace:*",
     "@react-hookz/web": "19.2.0",
+    "@react-three/drei": "8.20.2",
     "@react-three/fiber": "7.0.26",
     "d3-format": "3.1.0",
     "greenlet": "1.1.0",
--- a/apps/storybook/src/SurfaceVis.stories.tsx
+++ b/apps/storybook/src/SurfaceVis.stories.tsx
@@ -1,0 +1,31 @@
+import type { SurfaceVisProps } from '@h5web/lib';
+import { getDomain, SurfaceVis } from '@h5web/lib';
+import { createArrayFromView, getMockDataArray } from '@h5web/shared';
+import { OrbitControls } from '@react-three/drei';
+import type { Meta, Story } from '@storybook/react/types-6-0';
+
+import FillHeight from './decorators/FillHeight';
+
+const dataArray = createArrayFromView(
+  getMockDataArray('/nD_datasets/fourD').pick(0, 0, null, null)
+);
+const domain = getDomain(dataArray.data);
+
+const Template: Story<SurfaceVisProps> = (args) => (
+  <SurfaceVis {...args}>
+    <OrbitControls />
+  </SurfaceVis>
+);
+
+export const Default = Template.bind({});
+Default.args = {
+  dataArray,
+  domain,
+};
+
+export default {
+  title: 'Experimental/SurfaceVis',
+  component: SurfaceVis,
+  parameters: { layout: 'fullscreen', controls: { exclude: 'dataArray' } },
+  decorators: [FillHeight],
+} as Meta;
--- a/packages/lib/src/vis/hooks.ts
+++ b/packages/lib/src/vis/hooks.ts
@@ -7,12 +7,16 @@ import {
 import type { Domain, AnyNumArray } from '@h5web/shared';
 import type { Camera } from '@react-three/fiber';
 import { useFrame, useThree } from '@react-three/fiber';
+import { rgb } from 'd3-color';
 import { useEffect, useCallback, useMemo, useState } from 'react';
 import type { RefCallback } from 'react';
 
+import type { ColorMap } from './heatmap/models';
+import { getInterpolator } from './heatmap/utils';
 import { useVisCanvasContext } from './shared/VisCanvasProvider';
 import type { VisCanvasContextValue } from './shared/VisCanvasProvider';
 import {
+  createAxisScale,
   getAxisDomain,
   getAxisValues,
   getCombinedDomain,
@@ -96,4 +100,23 @@ export function useCssColors(
   const colors = colorProperties.map((p) => styles.getPropertyValue(p).trim());
 
   return [colors, refCallback];
+}
+
+export function useDataToColorScale(
+  scaleType: ScaleType,
+  domain: Domain,
+  colorMap: ColorMap,
+  invertColorMap: boolean
+): (v: number) => [number, number, number] {
+  return useMemo(() => {
+    const numScale = createAxisScale(scaleType, {
+      domain,
+      range: [0, 1],
+    });
+    const interpolator = getInterpolator(colorMap, invertColorMap);
+    return (value: number) => {
+      const color = rgb(interpolator(numScale(value)));
+      return [color.r, color.g, color.b];
+    };
+  }, [colorMap, domain, invertColorMap, scaleType]);
 }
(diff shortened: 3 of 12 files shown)


apps/storybook/package.json
```diff
@@ -24,6 +24,7 @@
     "@h5web/lib": "workspace:*",
     "@h5web/shared": "workspace:*",
     "@react-hookz/web": "14.2.2",
+    "@react-three/drei": "8.20.2",
```
loichuder, line 27: Added `drei` to the storybook for the [OrbitControls](https://drei.pmnd.rs/?path=/story/controls-orbitcontrols--orbit-controls-story).

I think this is the best compromise for now.

apps/storybook/src/SurfaceVis.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+);
+
+export const Default = Template.bind({});
+Default.args = {
+  dataArray,
+  domain,
+};
+
+export default {
+  title: 'Experimental/SurfaceVis',
```
loichuder, line 27: The new `SurfaceVis` can be tested in the Storybook and is deemed as **Experimental**

packages/lib/src/vis/hooks.ts
```diff
@@ -97,3 +101,22 @@ export function useCssColors(
...
+      domain,
+      range: [0, 1],
+    });
+    const interpolator = getInterpolator(colorMap, invertColorMap);
+    return (value: number) => {
+      const color = rgb(interpolator(numScale(value)));
+      return [color.r, color.g, color.b];
+    };
+  }, [colorMap, domain, invertColorMap, scaleType]);
+}
```
loichuder, line 122: Used in both `ScatterVis` and `SurfaceVis`.

packages/lib/src/vis/shared/Canvas.tsx
```diff
@@ -0,0 +1,27 @@
...
+      flat // disable tone mapping
+      frameloop="demand" // disable game loop
+      dpr={[1, 3]} // https://discoverthreejs.com/tips-and-tricks/#performance
+      resize={{ debounce: { scroll: 20, resize: 200 }, scroll: false }} // https://github.com/pmndrs/react-three-fiber/discussions/1906
+    >
+      <ambientLight />
+      {children}
+    </R3FCanvas>
+  );
+}
```
loichuder, line 25: New component `Canvas` to be used in both 2D and 3D vis with the appropriate props set.

packages/lib/src/vis/surface/SurfaceMesh.tsx
```diff
@@ -0,0 +1,58 @@
...
+    domain,
+    colorMap,
+    invertColorMap
+  );
+
+  const {
+    position,
+    color,
+    index: indices,
+  } = useBufferAttributes(dataArray, dataToColorScale);
```
loichuder, line 35: For now, I make use of `BufferAttributes` to set the `position` and the `color`, following the `ScatterPoints` component.

I may change this later to use shaders instead for the color, similarly to `HeatmapMesh`

packages/lib/src/vis/surface/SurfaceShaderMaterial.tsx
```diff
@@ -0,0 +1,25 @@
...
+    }
+  `,
+  fragmentShader: `
+    varying vec3 vertexColor;
+
+    void main() {
+      gl_FragColor = vec4(vertexColor.rgb, 1);
+    }
+  `,
+};
```
loichuder, line 19: Simplified version of the `GlyphMaterial` shader. 

Again, bound to change if I move to a `Heatmap`-like shader

packages/lib/src/vis/surface/SurfaceVis.module.css
```diff
@@ -0,0 +1,11 @@
...
+  flex: 1 1 0%;
+  display: flex;
+  min-width: 0;
+  min-height: 0;
+  margin: 0;
+}
+
+.canvas {
+  margin-right: 1rem;
+}
```
loichuder, line 11: This is to make some space for the `ColorMap` as there is no padding for axes.

packages/lib/src/vis/surface/hooks.ts
```diff
@@ -0,0 +1,60 @@
...
+      for (let col = 0; col < cols; col++) {
+        const value = dataArray.get(row, col);
+        positions.push(col, row, value);
+        colors.push(...dataToColorScale(value));
+      }
+    }
+
+    return {
+      position: new Float32BufferAttribute(positions, 3),
+      color: new Uint8BufferAttribute(colors, 3, true),
```
loichuder, line 27: Quite similar to [`useBufferAttributes`](https://github.com/silx-kit/h5web/blob/d8a5bafec2e3f788592d5c3b417b42b9cdb8dcb3/packages/lib/src/vis/scatter/hooks.ts#L8) of the `ScatterVis`.

Note that here, I choose to create the `BufferAttribute` in the hooks. This made more sense to me and I may change the `ScatterVis` version as well to follow this.

packages/lib/src/vis/surface/hooks.ts
```diff
@@ -0,0 +1,60 @@
...
+          // Second triangle: (x + 1 ,y) -> (x, y + 1) -> (x + 1, y + 1)
+          getIndex(col + 1, row),
+          getIndex(col, row + 1),
+          getIndex(col + 1, row + 1)
+        );
+      }
+    }
+
+    return indices;
+  }, [dataArray]);
```
loichuder, line 57: Difference with the `ScatterVis` hook: for `GL_POINTS`, there is only the need for a single coordinate per data points. But for `GL_TRIANGLES`, I need 6 points (2 triangles) to draw a square per data point (except the last row and column).

Rather than having 5 additional points per datapoints, I make use of [an indexed geometry](https://threejs.org/examples/#webgl_buffergeometry_indexed) to tell in which order the points must be linked to draw the triangles. This allows to reuse points and therefore to have less points in the BufferAttribute for better performance.

Also for performance, I also chose to keep the computation of `index` in a separate loop in its own `useMemo` so that they are not recomputed at each change of `colorMap`.

packages/lib/src/vis/surface/SurfaceMesh.tsx
```diff
@@ -0,0 +1,57 @@
...
+    dataGeometry.setAttribute('position', position);
+    dataGeometry.setAttribute('color', color);
+    dataGeometry.setIndex(indices);
+    invalidate();
+  }, [dataGeometry, invalidate, position, indices, color]);
+
+  return (
+    <group position={[-cols / 2, -rows / 2, 0]}>
+      <mesh geometry={dataGeometry}>
+        <meshBasicMaterial vertexColors side={DoubleSide} />
```
loichuder, line 46: I had a `shaderMaterial` here but the shader was so basic that `meshBasicMaterial` works just as well.

packages/lib/src/vis/surface/hooks.ts
axelboc: I agree. Not sure about the `positions` and `colors` being plain arrays, though. I mean, it probably doesn't make a huge difference, but initialising a buffer and setting values to it is most likely more memory efficient than pushing values to normal arrays.

packages/lib/src/vis/surface/hooks.ts
axelboc: :+1:  - I guess the shader will eventually be useful to avoid this double loop.

axelboc (approved):
Fancy! :sparkles:

packages/lib/src/vis/shared/Canvas.tsx
axelboc: Coming back to this, I think this is very well worth exporting, as experimental or not, whichever you think. If we do so, perhaps we should actually swap the naming - i.e. export it in the lib as `R3FCanvas` and use the default naming of R3F `Canvas` component internally.

packages/lib/src/vis/surface/hooks.ts
loichuder: True, it is done a bit better in the `ScatterVis` so I think I will revisit this in a future PR and try to take the better of both worlds

PR #1305, opened 2022-12-01
Upgrade Testing Library (DOM + user-event)

axelboc:
- Nothing much in https://github.com/testing-library/user-event/releases (from 14.2.0 to 14.4.3) as far as I can tell.
- Some really good stuff in https://github.com/testing-library/dom-testing-library/releases (from 8.13.0 to 8.19.10):
  - improved logging for `ByRole` queries (the `name` is now printed);
  - improved performance for `ByRole` queries (to be confirmed in CI);
  - not new but I've just learnt about the following [debugging feature](https://testing-library.com/docs/queries/about/#screenlogtestingplaygroundurl): `screen.logTestingPlaygroundURL()` -- it logs a URL for debugging the current document in https://testing-playground.com/ which looks really cool.

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -60,10 +60,10 @@
   "devDependencies": {
     "@h5web/shared": "workspace:*",
     "@rollup/plugin-alias": "4.0.2",
-    "@testing-library/dom": "8.13.0",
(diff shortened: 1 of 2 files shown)


loichuder (approved):
Thanks again for taking care of this !

PR #1307, opened 2022-12-01
Upgrade Cypress to v11

axelboc:
From 9.5.1 (inclusive) to 11.2.0: https://docs.cypress.io/guides/references/changelog#9-5-1 (phew!)

Notable changes for us copy-pasted from the CHANGELOG:

- 9.5.4: Updates were made to silence Electron warnings related to being unable to connect to dbus. These errors are normal and expected, and do not result in test failures. Because they are always present when running Electron inside docker containers, it has incorrectly led people to believe it is the root-cause of an error within their test run. By silencing these errors, it will improve the debug experience to allow users to focus on meaningful warning and error messages. 
- 10.0.0: We've also reworked the Cypress app from the ground up to modernize the interface, streamline workflows and integrate better into your overall development experience. https://cypress.io/blog/2022/06/01/cypress-10-release/
- 10.0.0: Using a `cypress.json` configuration file is no longer supported. Replace this configuration file with a `cypress.config.ts`.
- 10.0.0: Cypress has made several configuration option changes including renaming options, removing options, and only allowing some options within a specific testing type. Please run `cypress open` to have our automated migration update these options for you.
  - The guided config migration process was brilliant!
- 10.0.0: The `pluginsFile` option was removed. This option should be replaced with the new [`setupNodeEvents()`](https://docs.cypress.io/guides/references/configuration#setupNodeEvents).
- 10.0.0: New `--e2e` and `--component` CLI flags were added to `cypress run` and `cypress open` to allow a user to run or launch Cypress for a specific testing-type. If a testing-type specific CLI flag is not provided when running `cypress run`, Cypress will run end-to-end tests by default.
  - I added the flag explicitly even for `cypress run`.
- 10.0.0: If a testing type and browser flag are passed to `cypress open` when launching a specific project, the browser will automatically launch.
- 10.0.0: You can now change specs within the same window as running your tests.
- 11.0.0: We have also massively improved our startup performance by shipping a snapshot of our binary instead of the source files.

Other notable changes for the future:

- 10.0.0: Cypress 10 now includes beta support for component testing. 
- 11.0.0: Component Testing is now generally available for projects using React, Next.js, Angular, and Vue!
  - Might be useful to screenshot-test a component in isolation, either to test more logical branches or because it's only available in the lib.
- 10.7.0: Re-introduces the experimental [Cypress Studio feature](https://docs.cypress.io/guides/references/cypress-studio). Cypress Studio provides a visual way to generate tests within Cypress by recording interactions against the application under test.


.eslintrc.js
```diff
@@ -4,11 +4,12 @@ const { createConfig } = require('./eslint.shared');
 
 module.exports = createConfig(__dirname, getDependencies(), [
   {
-    files: ['**/*.spec.ts'],
+    files: ['**/*.cy.ts'],
     rules: {
       'testing-library/await-async-query': 'off', // Cypress has its own way of dealing with asynchronicity
       'testing-library/await-async-utils': 'off', // Cypress has its own way of dealing with asynchronicity
       'testing-library/prefer-screen-queries': 'off', // Cypress provides `cy` object instead of `screen`
+      'sonarjs/no-duplicate-string': 'off', // ioncompatible with Cypress testing syntax
```
axelboc, line 12: Not sure why this starts being a problem now 🤷

cypress.config.ts
```diff
@@ -0,0 +1,16 @@
+import { defineConfig } from 'cypress';
+import { addMatchImageSnapshotPlugin } from 'cypress-image-snapshot/plugin';
+
+export default defineConfig({
+  e2e: {
+    baseUrl: 'http://localhost:5173',
+    supportFile: 'cypress/support.ts',
+    setupNodeEvents(on, config) {
+      addMatchImageSnapshotPlugin(on, config);
```
axelboc, line 9: No more plugin file.

cypress/tsconfig.json
```diff
@@ -8,5 +8,5 @@
       "@types/cypress-image-snapshot"
     ]
   },
-  "include": ["*", "specs"]
+  "include": ["*", "e2e"]
```
axelboc, line 11: I switched to the default folder (it used to be `integration` which I didn't like) to avoid having to specify a regex in the config.

package.json
```diff
@@ -28,16 +28,16 @@
     "lint:root:eslint": "eslint \"**/*.{js,cjs,ts,tsx}\" --max-warnings=0",
     "prettier": "prettier . --cache --check",
     "test": "jest",
-    "cypress": "cypress open",
-    "cypress:run": "cypress run",
+    "cypress": "cypress open --e2e --browser firefox",
```
axelboc, line 31: Starting FF automatically when launching Cypress locally. Doesn't prevent opening the main Cypress window unfortunately.

.eslintrc.js
loichuder: :shrug: 

Slight typo in `incompatible`

loichuder (approved):
Looked intense :smile: 

Good job :+1:

PR #1309, opened 2022-12-01
Use shader in SurfaceVis

loichuder:
`SurfaceMesh` now uses the same shader as the `HeatmapMesh` (now in a `HeatmapMaterial` to be reusable). This avoids recomputing `BufferAttributes` when changing color map settings.

As a reminder, this shader relies on texture coordinates `uv` to look up the data value in the data texture to then compute the appropriate color. 

**This works in `HeatmapMesh` because the use of a plane geometry means that `uv` are essentially normalized coordinates of the geometry**. As `SurfaceMesh` uses a more complex geometry (an indexed `BufferGeometry`), **this postulate is no longer true**. 

To work around this and still be able to reuse the shader as-is, I now set the `uv` "manually" as attributes of the buffer geometry so that they contain the normalized coordinates expected by the shader. 

_Three.js has an example doing exactly this but for [normals](https://github.com/mrdoob/three.js/blob/309b00afb6dcbc5e6c58e72f10eaa8d2e8888c83/examples/webgl_buffergeometry_indexed.html) (see relevant discussion [here](https://discourse.threejs.org/t/indexed-buffer-geometry-normals/3679))._

After discussion with @t20100, we came up with several alternatives:
- Passing the `value` directly as `BufferAttribute` to save a lookup table and avoid computing `uv`
- Leverage in the shader the fact that `value` is in fact simply `position.z`. This may change in the future as we don't want the 3D display to be over-reliant on the data amplitude.

Unfortunately, all of them mean we would have to write more specific shader code. Since coding (and debugging) shaders and abstractions around them can be very challenging, I would make a strong point to keep the shader code minimal and the most reusable possible. Meaning I am relatively happy with the solution presented in this PR.

### DIFF
--- a/packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
@@ -1,0 +1,145 @@
+import type { Domain } from '@h5web/shared';
+import { ScaleType } from '@h5web/shared';
+import { rgb } from 'd3-color';
+import type { NdArray } from 'ndarray';
+import { memo, useMemo } from 'react';
+import type { TextureFilter } from 'three';
+import { DoubleSide } from 'three';
+import { DataTexture, RGBAFormat, UnsignedByteType } from 'three';
+
+import type { VisScaleType } from '../models';
+import type { VisMeshProps } from '../shared/VisMesh';
+import { DEFAULT_DOMAIN, getUniforms, VERTEX_SHADER } from '../utils';
+import { useDataTexture } from './hooks';
+import type { ColorMap, TextureSafeTypedArray } from './models';
+import { getInterpolator, scaleDomain } from './utils';
+
+interface Props extends VisMeshProps {
+  values: NdArray<TextureSafeTypedArray | Uint16Array>; // uint16 values are treated as half floats
+  domain: Domain;
+  scaleType: VisScaleType;
+  colorMap: ColorMap;
+  invertColorMap?: boolean;
+  magFilter?: TextureFilter;
+  alphaValues?: NdArray<TextureSafeTypedArray | Uint16Array>; // uint16 values are treated as half floats
+  alphaDomain?: Domain;
+}
+
+const CMAP_SIZE = 256;
+
+const IMPL_SCALE: Partial<Record<ScaleType, string>> = {
+  [ScaleType.Log]: 'log(value) * oneOverLog10',
+  [ScaleType.SymLog]: 'sign(value) * log(1. + abs(value)) * oneOverLog10',
+  [ScaleType.Sqrt]: 'sqrt(value)',
+};
+
+const IMPL_IS_SUPPORTED: Partial<Record<ScaleType, string>> = {
+  [ScaleType.Log]: 'value > 0.',
+  [ScaleType.Sqrt]: 'value >= 0.',
+};
+
+function HeatmapMaterial(props: Props) {
+  const {
+    values,
+    domain,
+    scaleType: visScaleType,
+    colorMap,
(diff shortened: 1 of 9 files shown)


packages/lib/src/vis/surface/hooks.ts
```diff
@@ -2,31 +2,28 @@ import type { NumArray } from '@h5web/shared';
...
     const positions: number[] = [];
-    const colors: number[] = [];
+    const uvs: number[] = [];
 
     for (let row = 0; row < rows; row++) {
       for (let col = 0; col < cols; col++) {
         const value = dataArray.get(row, col);
         positions.push(col, row, value);
-        colors.push(...dataToColorScale(value));
+        uvs.push((col + 0.5) / cols, (row + 0.5) / rows);
```
loichuder, line 18: Not really understood why but there was a shift of `[0.5, 0.5]` between the applied color and the point it referred to :shrug:

packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+          if (withAlpha == 1) {
+            gl_FragColor.a = oneOverAlphaRange * (texture2D(alpha, coords).r - alphaMin);
+          }
+        }
+      }
+    `,
+  };
+
+  return <shaderMaterial args={[shader]} side={DoubleSide} />;
```
loichuder, line 150: Note the addition of `side` that changes nothing for `HeatmapMesh` but makes the material visible from both sides in 3D

packages/lib/src/vis/surface/hooks.ts
loichuder: After investigation, I think this is because in `HeatmapVis`, the "pixel" color is applied to the center that is between the data points while we want the color to be applied to the data point for the `SurfaceVis`.

If you checkout the branch, you can check this yourself by setting `showPoints` to `true` and removing the `filter` of `HeatmapMaterial` in `SurfaceMesh`.

packages/lib/src/vis/surface/SurfaceMesh.tsx
```diff
@@ -20,30 +21,28 @@ function SurfaceMesh(props: Props) {
...
-    invertColorMap
-  );
+  const safeDataArray = useTextureSafeNdArray(dataArray);
 
-  const {
-    position,
-    color,
-    index: indices,
-  } = useBufferAttributes(dataArray, dataToColorScale);
+  const { position, index: indices, uv } = useBufferAttributes(dataArray);
```
axelboc, line 26: Now that the returned values are of type `BufferAttribute`, it might make more sense to keep the name `index` as is instead of renaming it.

loichuder: You are right. I don't know why there was still this back-and-forth

packages/lib/src/vis/surface/SurfaceMesh.tsx
```diff
@@ -20,30 +21,28 @@ function SurfaceMesh(props: Props) {
...
-    dataGeometry.setAttribute('color', color);
+    dataGeometry.setAttribute('uv', uv);
     dataGeometry.setIndex(indices);
     invalidate();
-  }, [dataGeometry, invalidate, position, indices, color]);
+  }, [dataGeometry, invalidate, position, indices, uv]);
 
   return (
     <group position={[-cols / 2, -rows / 2, 0]}>
       <mesh geometry={dataGeometry}>
```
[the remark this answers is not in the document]
loichuder: Agreed, I got the same feeling :+1:

PR #1311, opened 2022-12-05
Fix build to work on Windows

PeterC-DLS:
Commands for cleaning up files in build:css steps are not cross-platform so replace with node code

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -29,7 +29,7 @@
   },
   "scripts": {
     "build": "vite build && run-p build:*",
-    "build:css": "vite build --config vite.styles.config.js && concat dist/temp/style.css dist/style.css -o dist/styles.css && rm -Rf dist/style.css dist/temp",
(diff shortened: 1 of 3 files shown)


axelboc (changes requested):
Hey @PeterC-DLS, good catch! For conciseness, we could use a package like [`rimraf`](https://github.com/isaacs/rimraf#cli). You can run it with `pnpm dlx rimraf <path> <other-path>` to avoid adding it as a dependency.

PeterC-DLS:
True, I tried with [del-cli](https://www.npmjs.com/package/del-cli) but that did not work.

PeterC-DLS:
@axelboc , I decided to add `rimraf` as a dev dependency so the build can run offline (if all packages are installed).

axelboc:
Alright, makes sense. In that case, you don't need `npx`.

axelboc (approved):
Thanks!

PR #1312, opened 2022-12-05
Upgrade `eslint-config-galex`, now with `eslint-plugin-etc`

axelboc:
- [Galex release notes](https://github.com/ljosberinn/eslint-config-galex/releases/tag/v4.4.0)
- README of the new plugin, [`eslint-config-etc`](https://github.com/cartant/eslint-plugin-etc)
- [Galex' config for the new plugin](https://github.com/ljosberinn/eslint-config-galex/compare/v4.3.2...v4.4.0#diff-a86b1b1fb0b30afb2920ad7cc59e77b7fce52c6af6343c1788ab66ee06e4a421)
- The rule I've looking forward to for a while, [no-misused-generics](https://github.com/cartant/eslint-plugin-etc/blob/main/docs/rules/no-misused-generics.md), which allowed me to fix a few issues in this PR (see comments).


packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -123,9 +123,10 @@ export function findAssociatedDatasets(
 ): (NumArrayDataset | undefined)[] {
   const dsetList = attrValuesStore.getSingle(group, type) || [];
   const dsetNames = typeof dsetList === 'string' ? [dsetList] : dsetList;
-  assertArray<string>(dsetNames);
+  assertArray(dsetNames);
 
   return dsetNames.map((name) => {
+    assertStr(name);
```
axelboc, line 129: We were type-casting via a generic, here, so if the array contained anything else than `string` we may have landed on a bug that would have been hard to track down.

packages/lib/src/interactions/InteractionsProvider.tsx
```diff
@@ -79,10 +79,11 @@ function InteractionsProvider(props: { children: ReactNode }) {
       }
 
       // If conflicting interactions, the one with the most modifier keys take precedence
-      const [maxKeyInteraction] = matchingInteractions
-        .sort((a, b) => a.modifierKeys.length - b.modifierKeys.length)
-        .reverse();
+      matchingInteractions.sort(
+        (a, b) => b.modifierKeys.length - a.modifierKeys.length
+      );
```
axelboc, line 84: This is https://github.com/cartant/eslint-plugin-etc/blob/main/docs/rules/no-assign-mutated-array.md. Here, it's now clear that the array is sorted in place. Note that I also remove the need for `reverse()` by tweaking the comparator function.

packages/shared/src/mock/utils.ts
```diff
@@ -61,14 +61,15 @@ export function findMockEntity(path: string): ProvidedEntity | undefined {
   return child;
 }
 
+// eslint-disable-next-line etc/no-misused-generics
```
axelboc, line 64: Here, I'm explicitly allowing the caller to type-cast via the generic, as this function is used in Storybook for convenience and with known data. Note that it still works without `<T>` below because of `as T[]` on the last line.

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -78,10 +78,10 @@ export function useDatasetValues<D extends Dataset<ArrayShape | ScalarShape>>(
   });
 }
 
-export function useBaseArray<T, U extends T[] | TypedArray | undefined>(
-  value: U,
+export function useBaseArray<T extends unknown[] | TypedArray | undefined>(
+  value: T,
```
axelboc, line 82: This is the main change throughout. Before, you could do `useBaseArray<string, string[]>(...)` to effectively type-cast the return value to `NdArray<string>` -- i.e. like doing `useBaseArray(...) as NdArray<string>` but less explicitly, which was bad.

.github/workflows/lint-test.yml
```diff
@@ -27,6 +27,12 @@ jobs:
           node-version-file: 'package.json'
           cache: 'pnpm'
 
+      - name: Cache Cypress binary 📌
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/Cypress
+          key: cypress-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+
```
axelboc, line 35: When the `pnpm-lock.yaml` changes and we push code to the repo, if the `lint` job succeeds but not the `e2e`, we end up with a cache for the pnpm store but not the Cypress binaries. When that happens, on subsequent pushes, the `e2e` job always fails because it can never find the binaries, and the binaries never get cached. Only way to resolve is to change `pnpm-lock.yaml`, push and hope that the `e2e` passes.

To resolve this, I now cache the Cypress binaries in every job. This is not ideal. I've opened an issue in `actions/setup-node` to solve this more elegantly: https://github.com/actions/setup-node/issues/638

loichuder: Fine for now, thanks for investigating other ways :+1:

packages/app/src/vis-packs/core/hooks.ts
loichuder: One less generic is a win for me :+1:

packages/shared/src/utils.ts
```diff
@@ -184,13 +184,11 @@ export function getDims(dataArray: NdArray): Dims {
...
-) {
-  return (...args: P) => useMemo<R>(() => fn(...args), args); // eslint-disable-line react-hooks/exhaustive-deps
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export function createMemo<T extends (...args: any[]) => any>(fn: T): T;
+
+export function createMemo<T extends (...args: any[]) => any>(fn: T) {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument,react-hooks/exhaustive-deps
+  return (...args: any[]) => useMemo<any>(() => fn(...args), args);
 }
+/* eslint-enable @typescript-eslint/no-explicit-any */
```
loichuder, line 194: Might be worth putting `createMemo` in its own file to allow to use the disable of rules by file

loichuder (approved):
Nice :slightly_smiling_face:

packages/lib/src/vis/utils.ts
```diff
@@ -189,11 +189,16 @@ export function getCanvasScale(
   canvasSize: number
 ): AxisScale {
   const { scaleType, visDomain, flip } = config;
-  const canvasRange = [-canvasSize / 2, canvasSize / 2];
+
+  const flipFactor = flip ? -1 : 1;
+  const canvasRange = [
+    (-canvasSize / 2) * flipFactor,
+    (canvasSize / 2) * flipFactor,
+  ];
```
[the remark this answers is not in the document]
axelboc: Ha... How did you find this? `visx`'s documentation doesn't mention it... I had to look in the code... https://github.com/airbnb/visx/blob/master/packages/visx-scale/src/operators/reverse.ts#L14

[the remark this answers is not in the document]
loichuder: I think I stumbled upon it in another project :joy_cat:

PR #1316, opened 2022-12-07
Use tuples to store selection rectangles

axelboc:
Not a huge win (at least for now) in terms of amount of code, but it does feel clearer and it definitely simplifies naming start/end points variables when destructuring, and opens the door to a few simplifications (e.g. one prop/parameter instead of two for `SelectionRect`, `getEnclosedRectangle`, etc.) I haven't made those simplifications yet to keep the PR concise.

### DIFF
--- a/apps/storybook/src/Selection.stories.tsx
+++ b/apps/storybook/src/Selection.stories.tsx
@@ -1,4 +1,4 @@
-import type { Selection, ModifierKey, Axis } from '@h5web/lib';
+import type { ModifierKey, Axis, Selection, Rect2 } from '@h5web/lib';
 import { AxialSelectionTool } from '@h5web/lib';
 import {
   Pan,
@@ -12,9 +12,26 @@ import {
 import type { Meta, Story } from '@storybook/react';
 import { format } from 'd3-format';
 import { useState } from 'react';
-import type { Vector2 } from 'three';
 
 import FillHeight from './decorators/FillHeight';
+
+const formatCoord = format('.2f');
+
+function getTitle(selection: Rect2 | undefined) {
+  if (!selection) {
+    return 'No selection';
+  }
+
+  const [start, end] = selection;
+  return `Selection from (${formatCoord(start.x)}, ${formatCoord(
+    start.y
+  )}) to (${formatCoord(end.x)}, ${formatCoord(end.y)})`;
+}
+
+const SELECTION_COMPONENTS = {
+  line: SelectionLine,
+  rectangle: SelectionRect,
+};
 
 interface TemplateProps {
   selectionType: 'line' | 'rectangle';
@@ -24,18 +41,12 @@ interface TemplateProps {
   stroke?: string;
 }
 
-function vectorToStr(vec: Vector2) {
-  return `(${format('.2f')(vec.x)}, ${format('.2f')(vec.y)})`;
-}
-
 const Template: Story<TemplateProps> = (args) => {
   const { selectionType, selectionModifierKey, panModifierKey, ...svgProps } =
     args;
 
   const [activeSelection, setActiveSelection] = useState<Selection>();
-
-  const SelectionComponent =
-    selectionType === 'line' ? SelectionLine : SelectionRect;
+  const SelectionComponent = SELECTION_COMPONENTS[selectionType];
 
   if (selectionModifierKey === panModifierKey) {
     return (
@@ -48,13 +59,7 @@ const Template: Story<TemplateProps> = (args) => {
 
   return (
     <VisCanvas
-      title={
-        activeSelection
-          ? `Selection from ${vectorToStr(
-              activeSelection.startPoint
-            )} to ${vectorToStr(activeSelection.endPoint)}`
-          : 'No selection'
-      }
+      title={getTitle(activeSelection?.data)}
       abscissaConfig={{ visDomain: [-10, 0], showGrid: true }}
       ordinateConfig={{ visDomain: [50, 100], showGrid: true }}
     >
@@ -67,7 +72,13 @@ const Template: Story<TemplateProps> = (args) => {
         onSelectionEnd={() => setActiveSelection(undefined)}
         modifierKey={selectionModifierKey}
       >
-        {(selection) => <SelectionComponent {...selection} {...svgProps} />}
+        {({ data: [dataStart, dataEnd] }) => (
+          <SelectionComponent
+            startPoint={dataStart}
+            endPoint={dataEnd}
+            {...svgProps}
+          />
+        )}
       </SelectionTool>
     </VisCanvas>
   );
@@ -137,9 +148,7 @@ export const Persisted: Story<TemplateProps> = (args) => {
   const { selectionType, selectionModifierKey, panModifierKey } = args;
 
   const [persistedSelection, setPersistedSelection] = useState<Selection>();
-
-  const SelectionComponent =
-    selectionType === 'line' ? SelectionLine : SelectionRect;
+  const SelectionComponent = SELECTION_COMPONENTS[selectionType];
 
   if (selectionModifierKey === panModifierKey) {
     return (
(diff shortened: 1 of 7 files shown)


apps/storybook/src/Selection.stories.tsx
```diff
@@ -67,7 +66,13 @@ const Template: Story<TemplateProps> = (args) => {
         onSelectionEnd={() => setActiveSelection(undefined)}
         modifierKey={selectionModifierKey}
       >
-        {(selection) => <SelectionComponent {...selection} {...svgProps} />}
+        {({ data: [dataStart, dataEnd] }) => (
+          <SelectionComponent
+            startPoint={dataStart}
+            endPoint={dataEnd}
```
axelboc, line 72: I'm hoping I can refactor this in my next PR to avoid the array destructuring. Also, we have access to the world selection, so we could avoid the `dataToWorld` conversion in `Selection<Rect|Line>`.

packages/lib/src/interactions/AxialSelectToZoom.tsx
```diff
@@ -22,17 +22,12 @@ function AxialSelectToZoom(props: Props) {
...
 
   function onSelectionEnd(selection: Selection) {
-    // Work in world coordinates as we need to act on the world camera
-    const { worldStartPoint, worldEndPoint } = selection;
-
-    if (
-      worldStartPoint.x === worldEndPoint.x ||
-      worldStartPoint.y === worldEndPoint.y
-    ) {
+    const [worldStart, worldEnd] = selection.world;
```
axelboc, line 25: I've removed most of the `<...>Point` suffixes to keep things more readable.

packages/lib/src/interactions/AxialSelectionTool.tsx
```diff
@@ -26,30 +25,26 @@ function AxialSelectionTool(props: Props) {
   const { worldToData } = useVisCanvasContext();
   const camera = useThree((state) => state.camera);
 
-  const toAxialSelection = useCallback(
```
axelboc, line 29: `useCallback` is not neeeded. You can hide whitespace for a clearer diff.

packages/lib/src/interactions/AxialSelectionTool.tsx
```diff
@@ -26,30 +25,26 @@ function AxialSelectionTool(props: Props) {
...
-        worldEndPoint: axialWorldEndPoint,
-      };
-    },
-    [axis, camera, worldToData]
-  );
+  function toAxialSelection(selection: Selection): Selection {
+    const [worldStart, worldEnd] = selection.world;
+    const { bottomLeft, topRight } = getWorldFOV(camera);
+
+    const axialWorldSelection: Rect3 =
```
axelboc, line 32: One ternary that returns a `Rect3` allows me to map through it below to convert it to data space.

packages/lib/src/interactions/AxialSelectionTool.tsx
```diff
@@ -26,30 +25,26 @@ function AxialSelectionTool(props: Props) {
...
+            new Vector3(worldEnd.x, topRight.y, 0),
+          ]
+        : [
+            new Vector3(bottomLeft.x, worldStart.y, 0),
+            new Vector3(topRight.x, worldEnd.y, 0),
+          ];
+
+    return {
+      world: axialWorldSelection,
+      data: axialWorldSelection.map(worldToData) as Rect2,
```
axelboc, line 45: `as Rect2` because of the usual tuple mapping issue

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -53,18 +51,16 @@ function SelectionTool(props: Props) {
...
+      const { worldPt: worldEnd } = evt;
+      const boundWorldEnd = boundWorldPointToFOV(worldEnd, camera);
 
       return {
-        startPoint,
-        endPoint: worldToData(boundWorldEndPoint),
-        worldStartPoint,
-        worldEndPoint: boundWorldEndPoint,
+        world: [worldStart, boundWorldEnd],
+        data: [dataStart, worldToData(boundWorldEnd)],
```
axelboc, line 63: I'm still hoping to be able to store the selection internally only in world space, and to call `onSelectionChange` and the `children` render prop with this world selection only. I feel like that's all you need during the interaction and that `onSelectionEnd` should be the only callback to receive the selection in data space (since that's when the selection becomes relevant to the app and may need to be persisted).

I'll keep working towards this.

packages/lib/src/interactions/models.ts
```diff
@@ -8,12 +8,14 @@ export enum MouseButton {
...
-  endPoint: Vector2;
-  worldStartPoint: Vector3;
-  worldEndPoint: Vector3;
+  world: Rect3;
+  data: Rect2;
 }
 
+export type Rect<T extends Vector2 | Vector3> = [T, T];
+export type Rect2 = Rect<Vector2>;
+export type Rect3 = Rect<Vector3>;
```
axelboc, line 17: The new `Rect` models I mentioned yesterday.

loichuder (approved):
As you said, very few improvements but I get this impression this moves in the right direction

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -21,20 +21,16 @@ function SelectToZoom(props: Props) {
...
-    ).map(dataToWorld);
+    ).map(dataToWorld); // work in world coordinates as we need to act on the world camera
 
-    if (startPoint.x === endPoint.x || startPoint.y === endPoint.y) {
+    if (worldStart.x === worldEnd.x || worldStart.y === worldEnd.y) {
       return;
     }
 
-    const zoomRect = getEnclosedRectangle(startPoint, endPoint);
+    const zoomRect = getEnclosedRectangle(worldStart, worldEnd);
```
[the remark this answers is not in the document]
axelboc: Yeah, yep, this is gonna change. 👍

packages/lib/src/interactions/models.ts
axelboc: It might be convenient for typing utility functions that can work with both `Rect2` and `Rect3`. None yet, though. FYI, this came useful when I tried to add a generic to the `SelectionTool` component -- the idea was that if you passed `transformSelection: (worldSelection: Rect3) => Rect2`, then you'd end up with `Rect2` in the callbacks -- it worked but wasn't quite right for consumers for various reasons.

loichuder: > It might be convenient for typing utility functions that can work with both Rect2 and Rect3.

Yeah, that was my first guess. No complains keeping it like this for now.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -49,10 +50,8 @@ const Template: Story<TemplateProps> = (args) => {
   return (
     <VisCanvas
       title={
-        activeSelection
-          ? `Selection from ${vectorToStr(
-              activeSelection.startPoint
-            )} to ${vectorToStr(activeSelection.endPoint)}`
+        strSelection
+          ? `Selection from ${strSelection[0]} to ${strSelection[1]}`
```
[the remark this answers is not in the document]
axelboc: So much nicer, thanks. I've also added a `SELECTION_COMPONENTS` record to avoid the ternaries when picking the right `SelectionComponent`.

apps/storybook/src/Selection.stories.tsx
```diff
@@ -12,10 +12,27 @@ import {
 import type { Meta, Story } from '@storybook/react';
 import { format } from 'd3-format';
 import { useState } from 'react';
-import type { Vector2 } from 'three';
 
 import FillHeight from './decorators/FillHeight';
 
+const formatVec = format('.2f');
```
loichuder, line 18: Should rather be `formatCoord` but great improvement for `getTitle` :+1:

PR #1317, opened 2022-12-08
Fix infinite render loop in axial selection tool stories

axelboc:
**EDIT** The problem described below is still accurate, but the solutions I ended up with have changed -- cf. https://github.com/silx-kit/h5web/pull/1317#issuecomment-1344085794

---

Alright, so the axial selection tool stories had not been working for a few days, ever since I moved the `onSelectionChange` call in `SelectionTool` in a `useEffect` (89f5dac3eb1052ca279ef00bc1c00228afadc7bf). Then, the introduction of `transformSelection` made it worse (b3ff1b38f099233eb311f6b6da54ce62f0911b94)...

This is all because some of the selection stories have a state that is updated via `onSelectionChange` (to display the selection in the canvas' title), which means that when the selection changes, the entire `VisCanvas` gets re-rendered and this triggers an infinite render loop.

Obviously, in a real app, storing such a fast changing state at this level would be a very bad idea, but it's still not great that it leads to an infinite render loop. So let's start with identifying the cause of the render loop:

- Before I moved `onSelectionChange` to `useEffect` in `SelectionTool`, the `onSelectionChange` callback in `AxialSelectionTool` was not properly memoised: when the `onSelectionChange` prop was set (which is the case in the axial selection story), the callback would get a new reference on every render, and therefore on every mouse move. This was slow but not critical; `onSelectionChange` used to be called inside the mouse move handler in `SelectionTool`, so the handler was recreated and re-registered on every mouse move, but since a re-render doesn't trigger a mouse move, there was no risk of an infinite render loop.
- By moving the `onSelectionChange` call into `useEffect`, the effect started being called on every re-render and we ended up with an infinite render loop.

So the first solution is to **memoise `onSelectionChange` in `AxialSelectionTool`**.

However, this doesn't prevent an infinite render loop if the user forgets to memoise their own `onSelectionChange` callback! For this reason, I decided to **revert the move to `useEffect`** in `SelectionTool`. Sure, `onSelectionChange` can now again slow down the interaction if it's computationally-heavy (e.g. if it triggers a big re-render), but it's not as bad a risk as an infinite render loop. In other words, it's back to being an optimisation problem for the consumer rather than a debugging problem. To show one way of doing this optimisation properly, I changed `useState` in the stories with `useThrottledState`-- the improvement in fluidity of the selection interactions is significant.

As mentioned, `transformSelection` was also causing an infinite render loop, and for the same reason. Even though removing the `useEffect` resolved the issue, it still not great, because the event handlers still end up being recreated/re-registered on every render in the stories.

So I started by **reverting the "unmemoisation" of `toAxialSelection`** in `AxialSelectionTool` (from https://github.com/silx-kit/h5web/commit/59fbb27858f6bfaa82b55011245a39ea8900c826#diff-4541ea5f19d2f849664c8659e651c97bab5f81635e77388e24ac9fcd2b67fdfaL29). However this turned out not to be enough: `toAxialSelection` depends on `worldToData`, and `worldToData` is actually recreated on every render! That's because `worldToData` depends on the axis scales, which are themselves recreated on every render.

So basically, if we don't memoise the scales, we (or our users) risk running into an infinite loop issue again in the future! So I decided to go ahead and **memoise the scales**. I added a boolean to `createMemo` to perform a deep comparison of the dependencies (with react-hookz' `useDeepCompareMemo`), and I used it to create a memoised version of `getCanvasScale`. The reason for the deep comparison is that users are unlikely to want to memoise the axis config objects (`abscissaConfig` and `ordinateConfig`)...

As a bonus, I discovered a subtle bug when clicking quickly after finishing a selection. There was a race condition between the moment the `startEvt` state was reset in `onPointerUp` and the moment the next call to `onPointerMove` occurred. I fixed it by moving `startEvt` into a ref.

### DIFF
--- a/apps/storybook/src/AxialSelectToZoom.stories.tsx
+++ b/apps/storybook/src/AxialSelectToZoom.stories.tsx
@@ -20,6 +20,7 @@ import { useMockData } from './hooks';
 const { oneD, twoD } = mockValues;
 
 const Template: Story<AxialSelectToZoomProps> = (args) => {
+  const { modifierKey } = args;
   const domain = useDomain(oneD);
   assertDefined(domain);
 
@@ -28,7 +29,7 @@ const Template: Story<AxialSelectToZoomProps> = (args) => {
       abscissaConfig={{ visDomain: [0, oneD.length], showGrid: true }}
       ordinateConfig={{ visDomain: domain, showGrid: true }}
     >
-      <Pan />
+      <Pan modifierKey={modifierKey?.length === 0 ? 'Control' : undefined} />
       <Zoom />
       <AxialSelectToZoom {...args} />
       <ResetZoomButton />
@@ -78,7 +79,7 @@ export const DisabledInsideEqualAspectCanvas: Story<AxialSelectToZoomProps> = (
       ordinateConfig={{ visDomain: [0, 20], showGrid: true }}
       aspect="equal"
     >
-      <Pan />
+      <Pan modifierKey="Control" />
       <Zoom />
       <AxialSelectToZoom {...args} />
       <ResetZoomButton />
@@ -91,6 +92,11 @@ export const DisabledInsideEqualAspectCanvas: Story<AxialSelectToZoomProps> = (
       />
     </VisCanvas>
   );
+};
+
+DisabledInsideEqualAspectCanvas.argTypes = {
+  modifierKey: { control: false },
+  disabled: { control: false },
 };
 
 export default {
--- a/apps/storybook/src/DefaultInteractions.stories.tsx
+++ b/apps/storybook/src/DefaultInteractions.stories.tsx
@@ -1,20 +1,39 @@
-import { DefaultInteractions, ResetZoomButton, VisCanvas } from '@h5web/lib';
+import {
+  DefaultInteractions,
+  HeatmapMesh,
+  mockValues,
+  ResetZoomButton,
+  ScaleType,
+  VisCanvas,
+} from '@h5web/lib';
 import type { DefaultInteractionsConfig } from '@h5web/lib/src/interactions/DefaultInteractions';
 import type { Meta, Story } from '@storybook/react';
 
 import FillHeight from './decorators/FillHeight';
+import { useMockData } from './hooks';
+
+const { twoD } = mockValues;
 
 export const InsideAutoAspectCanvas: Story<DefaultInteractionsConfig> = (
   args
 ) => {
+  const { values, domain } = useMockData(twoD, [20, 41]);
+
   return (
     <VisCanvas
-      abscissaConfig={{ visDomain: [0, 10], showGrid: true }}
-      ordinateConfig={{ visDomain: [0, 10], showGrid: true }}
+      abscissaConfig={{ visDomain: [0, values.shape[1]], showGrid: true }}
+      ordinateConfig={{ visDomain: [0, values.shape[0]], showGrid: true }}
       aspect="auto"
     >
       <DefaultInteractions {...args} />
       <ResetZoomButton />
+
+      <HeatmapMesh
+        values={values}
+        domain={domain}
+        colorMap="Viridis"
+        scaleType={ScaleType.Linear}
+      />
     </VisCanvas>
   );
 };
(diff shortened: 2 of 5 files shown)


apps/storybook/src/AxialSelectToZoom.stories.tsx
```diff
@@ -28,7 +29,7 @@ const Template: Story<AxialSelectToZoomProps> = (args) => {
       abscissaConfig={{ visDomain: [0, oneD.length], showGrid: true }}
       ordinateConfig={{ visDomain: domain, showGrid: true }}
     >
-      <Pan />
+      <Pan modifierKey={modifierKey?.length === 0 ? 'Control' : undefined} />
```
axelboc, line 32: In a few stories, the pan interaction was conflicting with the selection interaction. So I know set the pan's `modifierKey` dynamically so it never conflicts.

apps/storybook/src/DefaultInteractions.stories.tsx
```diff
@@ -1,35 +1,63 @@
...
     >
       <DefaultInteractions {...args} />
       <ResetZoomButton />
+
+      <HeatmapMesh
+        values={values}
+        domain={domain}
+        colorMap="Viridis"
+        scaleType={ScaleType.Linear}
+      />
```
axelboc, line 36: The select-to-zoom interaction is constrained by the axis domains, which makes it look broken on an empty canvas.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -83,43 +84,52 @@ function SelectionTool(props: Props) {
...
     (evt: CanvasEvent<PointerEvent>) => {
-      if (startEvt) {
-        setSelection(computeSelection(evt));
+      if (!startEvtRef.current) {
+        return;
+      }
+      const newSelection = transformSelection(computeSelection(evt));
+
+      if (isModifierKeyPressed) {
+        onSelectionChange?.(newSelection);
```
axelboc, line 101: Bringing back the `onSelectionChange` call inside `onPointerMove`.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -133,24 +143,18 @@ function SelectionTool(props: Props) {
...
-      onSelectionChange(transformSelection(selection));
-    }
-  }, [selection, isModifierKeyPressed, onSelectionChange, transformSelection]);
-
   if (!selection || !isModifierKeyPressed) {
     return null;
   }
 
-  return <>{children(transformSelection(selection))}</>;
+  return <>{children(selection)}</>;
```
axelboc, line 157: I now store the transformed selection to avoid computing it twice.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -83,43 +84,52 @@ function SelectionTool(props: Props) {
...
   const onPointerMove = useCallback(
     (evt: CanvasEvent<PointerEvent>) => {
-      if (startEvt) {
-        setSelection(computeSelection(evt));
+      if (!startEvtRef.current) {
+        return;
+      }
+      const newSelection = transformSelection(computeSelection(evt));
+
+      if (isModifierKeyPressed) {
```
loichuder, line 100: I think we already discussed this but why the condition is `isModifierKeyPressed` rather than `shouldInteract` ?

loichuder (approved):

packages/lib/src/interactions/SelectionTool.tsx
axelboc: Instinctively, I'd say that we don't need to because the interaction was allowed to start. Maybe also the buttons in the event aren't right in `pointermove` or something like that? I wonder if the question should rather be if the condition in `onPointerUp` should be `isModifierKeyPressed` instead of `shouldInteract`...  😂

loichuder: Yeah, that rings a bell. Perhaps we should rename `shouldInteract` into `shouldInteractionStart` in the future.

Anyway, fine for me.

axelboc:
As discussed, change of plan: I've decided to use my new friend [`useSyncedRef`](https://react-hookz.github.io/web/?path=/docs/miscellaneous-usesyncedref--example) to wrap the callbacks in up-to-date but stable refs. By using those wrapped callback refs in `useEffect` and in the pointer event handlers, we guarantee that the effect and handlers aren't called/recreated every time the callbacks change. This completely removes the burden of memoising the callbacks from consumers!

Here is a list of what changed from my initial solution:

- The call to `onSelectionChange` in `SelectionTool` is back inside a `useEffect`.
- The `toAxialSelection` function in `AxialSelectionTool` is no longer memoised.
- The axis scales are no longer memoised and I've removed the `deep` flag from `createMemo`. (In a real app, `VisCanvas` should really not be re-rendered that often...)
- All the callbacks in `SelectionTool` are now wrapped in `useSyncedRef`. (Re-rendering `SelectionTool` often may be more difficult for consumers to avoid, which is why optimising it with `useSyncedRef` is important.)
- The `transformSelection` callback is now called inside `computeSelection` for simplicity.

loichuder (approved):

PR #1318, opened 2022-12-09
Invoke selection start/end callbacks asynchronously

axelboc:
Now that the callbacks are stable, we can really embrace `useEffect`. This PR moves the calls to `onSelectionStart` and `onSelectionEnd` into the `useEffect`.

In addition to reducing the potential impact on performance of consumers doing something really heavy in the callbacks, it also allows me to easily fix two bugs:

1. `onSelectionStart` is now called only when the user actually starts dragging. So a simple click no longer creates a selection and no longer invokes any of the callbacks.
2. `onSelectionEnd` is now called _after_ the render prop, which means that if the consumer persists the selection in `onSelectionEnd` and displays it, we no longer see two selections stacked on top of each other for one render. This allows us to remove the workaround for this subtle bug from the `Persisted` selection story.

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -1,5 +1,9 @@
-import { assertDefined } from '@h5web/shared';
-import { useKeyboardEvent, useRafState, useSyncedRef } from '@react-hookz/web';
+import {
+  useKeyboardEvent,
+  usePrevious,
+  useRafState,
+  useSyncedRef,
+} from '@react-hookz/web';
 import { useThree } from '@react-three/fiber';
 import type { ReactNode } from 'react';
 import { useRef } from 'react';
@@ -18,8 +22,8 @@ import { boundWorldPointToFOV, getModifierKeyArray } from './utils';
 interface Props extends CommonInteractionProps {
   id?: string;
   transformSelection?: (selection: Selection) => Selection;
-  onSelectionStart?: (evt: CanvasEvent<PointerEvent>) => void;
-  onSelectionChange?: (selection: Selection) => void;
+  onSelectionStart?: (selection: Selection) => void;
+  onSelectionChange?: (selection: Selection | undefined) => void;
   onSelectionEnd?: (selection: Selection) => void;
   children: (selection: Selection) => ReactNode;
 }
@@ -46,7 +50,10 @@ function SelectionTool(props: Props) {
   const { worldToData } = useVisCanvasContext();
 
   const startEvtRef = useRef<CanvasEvent<PointerEvent>>();
+  const hasSuccessfullyEndedRef = useRef<boolean>(false);
+
   const [selection, setSelection] = useRafState<Selection>();
+  const prevSelection = usePrevious(selection);
 
   const modifierKeys = getModifierKeyArray(modifierKey);
   const isModifierKeyPressed = useModifierKeyPressed(modifierKeys);
@@ -57,22 +64,6 @@ function SelectionTool(props: Props) {
     disabled,
   });
 
-  const computeSelection = useCallback(
-    (evt: CanvasEvent<PointerEvent>): Selection => {
-      assertDefined(startEvtRef.current);
-      const { dataPt: dataStart, worldPt: worldStart } = startEvtRef.current;
-
-      const { worldPt: worldEnd } = evt;
-      const boundWorldEnd = boundWorldPointToFOV(worldEnd, camera);
-
-      return transformSelectionRef.current({
-        world: [worldStart, boundWorldEnd],
-        data: [dataStart, worldToData(boundWorldEnd)],
-      });
-    },
-    [camera, transformSelectionRef, worldToData]
-  );
-
   const onPointerDown = useCallback(
     (evt: CanvasEvent<PointerEvent>) => {
       const { sourceEvent } = evt;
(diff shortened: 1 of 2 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -46,7 +50,10 @@ function SelectionTool(props: Props) {
   const { worldToData } = useVisCanvasContext();
 
   const startEvtRef = useRef<CanvasEvent<PointerEvent>>();
+  const hasSuccessfullyEndedRef = useRef<boolean>(false);
```
axelboc, line 53: This ref allows me to track whether the modifier keys where indeed correctly pressed when the `pointerup` event occurred, and also to indicate whether the selection ended "successfully" - i.e. that it didn't get cancelled with <kbd>Escape</kbd>.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -46,7 +50,10 @@ function SelectionTool(props: Props) {
...
+
   const [selection, setSelection] = useRafState<Selection>();
+  const prevSelection = usePrevious(selection);
```
axelboc, line 56: This allows me to tell whether the selection is starting/changing/ending.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -57,22 +64,6 @@ function SelectionTool(props: Props) {
     disabled,
   });
 
-  const computeSelection = useCallback(
```
axelboc, line 60: I've moved this into `onPointerMove` since it's only called there now.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -18,8 +22,8 @@ import { boundWorldPointToFOV, getModifierKeyArray } from './utils';
 interface Props extends CommonInteractionProps {
   id?: string;
   transformSelection?: (selection: Selection) => Selection;
-  onSelectionStart?: (evt: CanvasEvent<PointerEvent>) => void;
-  onSelectionChange?: (selection: Selection) => void;
+  onSelectionStart?: (selection: Selection) => void;
```
loichuder, line 25: Do you remember why we had the `evt` as parameter here instead of `selection` ?

Also, it is fortunate that we derive several interfaces from these `Props` (I was wondering why you didn't need to change anything in `AxialSelectionTool` :smile: )

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -131,10 +129,30 @@ function SelectionTool(props: Props) {
...
+
+    if (prevSelection && !selection && hasSuccessfullyEndedRef.current) {
+      hasSuccessfullyEndedRef.current = false;
+      onSelectionEndRef.current?.(prevSelection);
+      return;
+    }
+
+    if (prevSelection && selection) {
+      onSelectionChangeRef.current?.(
+        isModifierKeyPressed ? selection : undefined
```
loichuder, line 145: Could you elaborate a bit on the reason of this change ?

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -131,10 +129,30 @@ function SelectionTool(props: Props) {
...
 
   useEffect(() => {
-    if (selection && isModifierKeyPressed) {
-      onSelectionChangeRef.current?.(selection);
+    if (!prevSelection && selection) {
+      onSelectionStartRef.current?.(selection);
+      return;
+    }
+
+    if (prevSelection && !selection && hasSuccessfullyEndedRef.current) {
```
loichuder, line 137: Not sure if this is the effect of a December Friday afternoon but these conditions feel hard to process for me :joy_cat:  .

I would suggest to add some comments but perhaps having the explanation later will make things crystal clear :smile_cat:

packages/lib/src/interactions/SelectionTool.tsx
axelboc: Hehe, yeah it's good. I had added `evt` in a previous PR in case we needed it in the future or in case a consumer needed it, but we don't actually do need it yet. Now, by the time this is called, a `selection` object is already computed, so it makes sense to expose it I think.

packages/lib/src/interactions/SelectionTool.tsx
axelboc: Ah yeah, good point. Basically, this replicates the behaviour of the `children` render prop: we didn't use to inform `onSelectionChange` subscribers when the selection was no longer "valid"/"active"/"relevant" (whatever you want to call it), when the modifier key was released. The callback stopped being called with no other cues whatsoever. Now, they get `undefined` once when the modifier key is released, and then the latest `selection` object when the modifier key is pressed again.

packages/lib/src/interactions/SelectionTool.tsx
axelboc: `prevSelection` holds the selection of the previous render.

- When the selection starts, it's `undefined` and the current `selection` is defined, so we know to call `onSelectionStart`.
- When the selection continues, it's defined, and the current `selection` is also defined, so we know to call `onSelectionChange`.
- When the selection ends, it's defined but the current `selection` is `undefined`, so we know to call `onSelectionEnd`.

`hasSuccessfullyEnded` is just a flag so we don't call `onSelectionEnd`:

- when the user actually cancels the selection with `Escape`; or
- when the call to `shouldInteract` returns `false` in `onPointerUp` (i.e. when the user has released a modifier key before releasing the pointer).

packages/lib/src/interactions/SelectionTool.tsx
loichuder: I see. Clever!

That would need a bit of documentation I think though.

axelboc: User-facing? It is demonstrated in the Storybook, technically. Or do you mean dev-facing documentation? I'll add a comment.

packages/lib/src/interactions/SelectionTool.tsx
loichuder: Yeah I wonder if making  `hasSuccessfullyEnded` smarter could simplify all this. 

It could be containing the current status of the selection process (e.g. `started`, `inProgress`, `ended`, `invalid`) so that we don't rely on `selection === undefined` that is an ambiguous condition that needs further tests to know the exact state of the selecting process.

axelboc: The ref solution works because it tracks a very specific thing. If I start complexifying it, I run the risk of introducing race conditions (e.g. what if `onPointerMove` is called before the `useEffect` has had a chance to run?) I could store a more complex `selection` state with a `status` key for instance, but in that case, we'd be better off using a full-on state machine, otherwise we're sure to re-introduce tricky state-related bugs. I felt the ref was a simple, bug-free solution for the problem at hand.

loichuder: I understand. But the chain of conditions feel really convoluted to me so I feel it should at least be commented.

axelboc: I've added a bunch of comments, let me know if they make sense.

loichuder (approved):

packages/lib/src/interactions/SelectionTool.tsx
loichuder: They do. Thanks :+1:

PR #1319, opened 2022-12-12
Use single cache for both pnpm store and Cypress binaries in CI

axelboc:
`actions/setup-node` does not and will not provide a way to pass additional paths to its internal `actions/cache` instance, so I move caching into a separate step that takes care of caching both the pnpm store and Cypress binaries.

### DIFF
--- a/.github/workflows/approve-snapshots.yml
+++ b/.github/workflows/approve-snapshots.yml
@@ -22,22 +22,25 @@ jobs:
         with:
           ref: ${{ steps.getBranchName.outputs.branch }}
 
+      - name: Use Node 16 🕹️
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: 'package.json'
+
       - name: Install pnpm ⚙️
         uses: pnpm/action-setup@v2
         with:
           version: 7.x
 
-      - name: Use Node 16 🕹️
-        uses: actions/setup-node@v3
-        with:
-          node-version-file: 'package.json'
-          cache: 'pnpm'
-
-      - name: Cache Cypress binary 📌
+      - name: Restore cache 📌
         uses: actions/cache@v3
         with:
-          path: ~/.cache/Cypress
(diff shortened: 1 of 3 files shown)


.github/workflows/approve-snapshots.yml
```diff
@@ -22,22 +22,25 @@ jobs:
...
+        with:
+          version: 7.x
+
+      - name: Restore cache 📌
         uses: actions/cache@v3
         with:
-          path: ~/.cache/Cypress
-          key: cypress-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+          path: |
+            ~/setup-pnpm/node_modules/.bin/store
```
axelboc, line 39: The path is hardcoded for now, as I don't want to add an [extra step](https://github.com/pnpm/action-setup#use-cache-to-reduce-installation-time) just to run `pnpm store path`. I've suggested storing the path as an output of the `pnpm/action-setup` step in https://github.com/pnpm/action-setup/issues/68#issuecomment-1346181326. To be continued.

.github/workflows/approve-snapshots.yml
```diff
@@ -22,22 +22,25 @@ jobs:
...
         uses: actions/setup-node@v3
         with:
           node-version-file: 'package.json'
-          cache: 'pnpm'
 
-      - name: Cache Cypress binary 📌
+      - name: Install pnpm ⚙️
+        uses: pnpm/action-setup@v2
+        with:
+          version: 7.x
```
axelboc, line 33: Moving this step after the Node installation step, now that we no longer use the built-in caching feature of `actions/setup-node`.

.github/workflows/approve-snapshots.yml
loichuder: Yeah, it is fine like this, no worries.

loichuder (approved):

PR #1321, opened 2022-12-12
Add SurfaceVis container and toolbar

loichuder:
For #1292 and first part of #1303


axelboc (changes requested):
Would you mind maybe adding an env var to allow enabling the surface vis in the viewer? I'm not a huge fan of putting code that's not demonstrable somehow. It will be more convenient for you to have such a feature flag while the feature is still under development, and it will allow us to show progress in meetings.

loichuder:
> Would you mind maybe adding an env var to allow enabling the surface vis in the viewer? I'm not a huge fan of putting code that's not demonstrable somehow. It will be more convenient for you to have such a feature flag while the feature is still under development, and it will allow us to show progress in meetings.

Great idea :+1:

loichuder:
I added a check on a env variable but it is quick and dirty.

Also, it is done in the `app` so that it makes tests fail. Please have a look and tell me if you have a better way of dealing with this.

loichuder:
Used a query param `experimental` in the end.

PR #1322, opened 2022-12-12
Add SVG overlay and `SvgElement` helper

axelboc:
Basically the replacement for SVG scene.

SVG elements wrapped in `SvgElement` get appended to this new global `svg` overlay container via a portal, similarly to floating toolbar controls. When it comes to naming, I hestiated with `SvgShape`, `Svg`, etc. but had issues with each. We can discuss this in person if you like @loichuder .

### DIFF
--- a/packages/lib/src/vis/shared/VisCanvas.tsx
+++ b/packages/lib/src/vis/shared/VisCanvas.tsx
@@ -45,6 +45,7 @@ function VisCanvas(props: PropsWithChildren<Props>) {
       })
     : NO_OFFSETS;
 
+  const [svgOverlay, setSvgOverlay] = useState<SVGSVGElement>();
   const [floatingToolbar, setFloatingToolbar] = useState<HTMLDivElement>();
 
   return (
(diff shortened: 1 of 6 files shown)


packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -76,6 +78,13 @@ function VisCanvas(props: PropsWithChildren<Props>) {
             <ThresholdAdjuster value={raycasterThreshold} />
           )}
         </VisCanvasProvider>
+
+        <Html>
+          <svg
+            ref={(elem) => setSvgOverlay(elem || undefined)}
+            className={styles.svgOverlay}
+          />
+        </Html>
```
loichuder, line 87: Don't you need to set the size to the `canvasSize` so that svg elements are properly positioned ?

loichuder (approved):
Looks good :slightly_smiling_face:

packages/lib/src/vis/shared/VisCanvas.tsx
axelboc: The size is set in CSS to 100%, so I think it's fine.

PR #1323, opened 2022-12-14
Refactor zoom and selection computations with `Box3`

axelboc:
This was a difficult but rewarding one. I considered splitting in multiple commits/PR, but the overall reasoning would have been lost in the process. I did revert a few changes as I wrote this, though. So here goes nothing...

---

At its core, this PR deals with the interaction utilities in `lib/src/interactions/utils.ts`: `getRatioRespectingRectangle`, `getEnclosedRectangle` `clampPositionToArea` and `clampRectangleToVis`. Let's first go over them to see what they do (or rather used to do) and why they need to be refactored:

### `getRatioRespectingRectangle`

Takes a rectangle as two vectors and a ratio, and expands the rectangle either vertically or horizontally (and equally on both sides) to match the ratio. It is used in `SelectToZoom`, when a selection ends, as the first step to computing the area to zoom on. It is also used in `RatioSelectionRect` (itself used by `SelectToZoom`), as the first step to displaying the effective zoom selection.

#### Why refactor it?

- The name is a bit awkward.
- The internal logic takes particular care to keep the "orientation" of the input rectangle, but we don't actually care about this: we just need a box to zoom on and draw on screen.
- We compute the [size](https://threejs.org/docs/?q=box#api/en/math/Box3.getSize) and [center](https://threejs.org/docs/?q=box#api/en/math/Box3.getCenter) of the rectangle, which `Box3` is very god at doing.
- `Box3` provides a method for [expanding a box by a vector](https://threejs.org/docs/?q=box#api/en/math/Box3.expandByVector) (`x` for horizontal expansion, `y` for vertical expansion).

### `getEnclosedRectangle`

Takes a rectangle as two vectors and returns the `width`, `height` and `center` vector of the rectangle. This is a convenience utility that is used right after both occurrences of `getRatioRespectingRectangle` (see above) and also in `clampRectangleToVis` (see below). It's actually surprising we weren't also using it directly in `getRatioRespectingRectangle` since we also compute the center/size of a rectangle in there...

#### Why refactor it?

- This is basically the same as `box3.getCenter()` and `box3.getSize()`.
- The code cares about the "orientation" of the input rectangle, a complexity that `Box3` can abstract away for us.
- Computing both the center and size every time might be inefficient if you only need one or the other.

### `clampPositionToArea`

Even though it's quite concise, this interaction utility is actually the most obscure. It basically takes a box (as a center and a size) and an "area" box (only as a size, assuming that the center of the area is at `Vector(0, 0, 0)`), and shifts the box so it remains within the area box. What it returns is the center of the shifted box.

This is used in:

- `useMoveCameraTo` to prevent moving the FOV outside of the bounds of the visualization;
- `clampRectangleToVis` (itself used in `RatioSelectionRect`, see below) to keep the effective zoom rectangle within the bounds of the visualization (note: not within the FOV, see below).

#### Why refactor it?

- The name is confusing: the input box is not actually clamped but shifted. It even supports passing a box bigger than the area box, in which case the box ends up centered around the area box (e.g. which can happen with the select-to-zoom interaction, when `keepRatio` is enabled and the canvas is larger/taller than the visualization => the zoom box spills over the visualization bounds).
- In `useMoveCameraTo`, we effectively pass the target FOV we want the camera to have as well as the visualization area we want to keep the FOV in. This works great, but the API does not make this clear at all.
- In `clampRectangleToVis`, we pass the box we ideally want to zoom on as well as the visualization area. Again the API isn't clear and it requires calling `getEnclosedRectangle` first. But there's a bigger problem: the fact that we pass the visualization area means that the zoom box can actually overflow the FOV of the camera. This is a bit tricky to replicate: you need to zoom in the middle of the visualization and then make a selection with such a ratio that the zoom box overflows the FOV; then, you can reduce your selection in one direction to see the zoom box move in and out of the FOV.
- `Box3` has a method to [translate a box](https://threejs.org/docs/?q=box#api/en/math/Box3.translate).

### `clampRectangleToVis`

This is basically just a wrapper around `clampPositionToArea` to make this function's API more palatable to `RatioSelectionRect` specifically - i.e. it takes a rectangle as two vectors and calls `getEnclosedRectangle` to get the center and size of the rectangle that `clampPositionToArea` requires; it also returns the shifted rectangle instead of just the shifted center so we can easily pass `startPoint` and `endPoint` to `SelectionRect`.

#### Why refactor it?

- The name is confusing because the rectangle is not clamped but shifted -- "kept within" the visualization area.
- As discussed above, the fact that the zoom box sometimes stays within the FOV (thanks to `SelectionTool`) and sometimes moves out of it is confusing from the user's viewpoint.

---

The refactoring consists mainly in **embracing Three's `Box3` class**. 

At first, I tried using `Box3` directly in the utilities above but their APIs (based on parameters like `startPoint`, `endPoint`, `center`, etc.) quickly got in the way and the code quickly became unreadable. Then, I tried making new utilities with APIs that worked better with `Box3`, but this led to a lot of duplication and inefficiencies (initialising and cloning lots of vectors/boxes).

In the end, and after much consideration, I decided to make my own `Box` class that extends `Box3`. I added:

- a couple of getters for `center` and `size` to abstract `Box3`'s more convoluted `getCenter` and `getSize` methods;
- a couple of convenience static methods for creating boxes:
  - `Box.empty()`, because by default, Three [sets the `min` and `max` vectors](https://github.com/mrdoob/three.js/blob/master/src/math/Box3.js#L5) to `+Infinity` and `-Infinity` respectively, which makes expanding the box by a set amount challenging... They really like to complicate things...
  - `Box.fromPoints()`, which is a shortcut for [`setFromPoints`](https://threejs.org/docs/?q=box#api/en/math/Box3.setFromPoints) that accepts individual vectors instead of an array (`...points: Vector3[]` instead of `points: Vector3[]`);
- a convenience method, `expandBySize` for expanding a box by a given `width` and `height` from the center (i.e. the expansion is spread equally both vertically and horizontally).
- two methods that effectively replace and improve upon `getRatioRespectingRectangle` and `clampRectangleToVis`:
  - `expandToRatio`, to expand a box to a given ratio (either horizontally or vertically but always equally on both sides so the center of the box doesn't change);
  - `keepWithin`, to keep a box within another "area" box (by shifting it if needed with `translate`) -- as per `clampPositionToArea`'s original behaviour, if the area box is smaller than the source box, the source box becomes centered around the area box.

As per other `Box3` methods and in line with Three's coding style, the new `Box` methods mutate the instance and are chainable. As you'll see, this is quite convenient and makes the code very readable.

---

A second important aspect of the refactoring concerns `moveCameraTo`. As I was refactoring things with `Box3`, I realised two things:

1. `moveCameraTo` doesn't care about the position of the FOV; it only cares about its size;
2. there's a much simpler way to get the size of the FOV that doesn't require unprojecting the `bottomLeft` and `topRight` corners with `getWorldFOV`: we can simply multiply the camera's `scale` by `canvasSize`!

These realisations allowed me to simplify/clarify the code a lot by means of the new `Box` class. Since the new solution doesn't make use of the camera's projection matrix, it also means we no longer have to call `camera.updateMatrixWorld()` when we update the `scale` before calling `moveCameraTo`, which removes a hidden coupling and is a big win in my book!

---

Hopefully this gives you enough context to understand the diff and the comments within, where I explain the implementation details. 😅


packages/lib/src/interactions/AxialSelectToZoom.tsx
```diff
@@ -15,26 +14,16 @@ interface Props extends CommonInteractionProps {
 function AxialSelectToZoom(props: Props) {
   const { axis, modifierKey, disabled } = props;
 
-  const { canvasSize, visRatio } = useVisCanvasContext();
-  const moveCameraTo = useMoveCameraTo();
+  const { visRatio } = useVisCanvasContext();
+  const zoomOnBox = useZoomOnBox();
```
axelboc, line 18: The function returned by this new hook updates the camera's scale and calls `moveCameraTo`. It removes duplication between `SelectToZoom` and `AxialSelectToZoom`.

packages/lib/src/interactions/AxialSelectToZoom.tsx
```diff
@@ -15,26 +14,16 @@ interface Props extends CommonInteractionProps {
...
 
-    const zoomRect = getEnclosedRectangle(worldStart, worldEnd);
-    const { center: zoomRectCenter } = zoomRect;
-
-    // Change scale first so that moveCameraTo computes the updated camera bounds
-    camera.scale.set(zoomRect.width / width, zoomRect.height / height, 1);
-    camera.updateMatrixWorld();
-
-    moveCameraTo(zoomRectCenter);
+    zoomOnBox(Box.fromPoints(...worldSelection));
```
axelboc, line 26: See `useZoomOnBox` for implementation. The scale update logic doesn't change; I just remove `camera.updateMatrixWorld();` and access `box.size` and `box.center` instead of calling `getEnclosedRectangle`.

packages/lib/src/interactions/RatioSelectionRect.tsx
```diff
@@ -1,36 +0,0 @@
...
-  const { visSize, dataToWorld, worldToData } = useVisCanvasContext();
-
-  const [ratioWorldStartPoint, ratioWorldEndPoint] =
-    getRatioRespectingRectangle(startPoint, endPoint, ratio).map(dataToWorld);
-
-  const [clampedDataStartPoint, clampedDataEndPoint] = clampRectangleToVis(
-    ratioWorldStartPoint,
-    ratioWorldEndPoint,
-    visSize
-  ).map(worldToData);
```
axelboc, line 25: I'm removing this component. See `computeZoomBox` in `SelectToZoom` to see what this code becomes.

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -1,69 +1,76 @@
...
   const camera = useThree((state) => state.camera);
-  const moveCameraTo = useMoveCameraTo();
+  const zoomOnBox = useZoomOnBox();
 
-  function onSelectionEnd(selection: Selection) {
-    const [worldStart, worldEnd] = getRatioRespectingRectangle(
-      ...selection.data,
-      keepRatio ? canvasRatio : undefined
-    ).map(dataToWorld); // work in world coordinates as we need to act on the world camera
+  function computeZoomBox(worldSelection: Rect3): Box {
```
axelboc, line 22: This function is called from both the render prop, to compute the zoom box to display via `SelectionRect`, and from `onSelectionEnd`, to compute the box to pass to `zoomOnBox`.

First, it creates a box for the given `worldSelection`, then, if `keepRatio` is enabled, it expands the box to match the ratio and then "keeps" this box within the camera's FOV. To compute the FOV box, I use the same trick as discussed for `moveCameraTo`: I multiply the scale by the canvas size.

As explained, this "fixes" the behaviour of `SelectToZoom` slightly, as the zoom box used to be kept within the visualization area, not within the FOV.

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -1,69 +1,76 @@
...
+  }
 
-    // Change scale first so that moveCameraTo computes the updated camera bounds
-    camera.scale.set(zoomRect.width / width, zoomRect.height / height, 1);
-    camera.updateMatrixWorld();
+  function onSelectionEnd(selection: Selection) {
+    const [worldStart, worldEnd] = selection.world;
+    if (worldStart.x === worldEnd.x || worldStart.y === worldEnd.y) {
+      return;
+    }
```
axelboc, line 42: I've left all the empty box validation alone for now, as I want to deal with them all at once when add the zoom selection threshold.

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -1,69 +1,76 @@
...
 
-    moveCameraTo(zoomRectCenter);
+    zoomOnBox(computeZoomBox(selection.world));
```
axelboc, line 44: Note that we didn't use to bother calling `clampRectangleToVis` before `moveCameraTo`, since `moveCameraTo` already takes care of keeping the FOV within the visualization area.

Now, `computeZoomBox` always calls `keepWithin`. I think the explicitness and consistency between `onSelectionEnd` and the render prop is good. It might even allow us to remove the call to `moveCameraTo` from `zoomOnBox` and to just set the camera position directly to the box' center.

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -1,69 +1,76 @@
...
+              fillOpacity={keepRatio ? 0 : 0.25}
+              strokeDasharray={keepRatio ? '4' : undefined}
             />
-          )}
-        </>
-      )}
+            {keepRatio && (
+              <SelectionRect
+                startPoint={worldToData(zoomBox.min)}
+                endPoint={worldToData(zoomBox.max)}
```
axelboc, line 65: If we end up adding low-level components that accept `worldCoords: Rect3`, we might not need the extra `zoomBox` constant and `worldToData` conversions.

packages/lib/src/interactions/box.ts
```diff
@@ -0,0 +1,61 @@
+import { Box3, Vector3 } from 'three';
+
+import type { Size } from '../vis/models';
+
+const ZERO_VECTOR = new Vector3(0, 0, 0);
+
+class Box extends Box3 {
+  public get size(): Size {
+    const { x: width, y: height } = this.getSize(new Vector3());
+    return { width, height };
```
axelboc, line 10: `getSize` returns a vector. I'm still a little unsure as to whether this is a good thing or not. Destructuring a size vector with `{ x: width, y: height }` or creating one with `new Vector3(width, height, 0)` definitely doesn't feel right, but a vector is useful for working with methods like `expandByVector` or for making operations on it. For now, I've decided to stick with our `Size` data structure and I've also added an `expandBySize` method to abstract away `expandByVector`.

packages/lib/src/interactions/box.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  public get center(): Vector3 {
+    return this.getCenter(new Vector3());
```
axelboc, line 14: Note that we should avoid accessing `box.center` and `box.size` multiple times in the same method. It is better to use destructuring: `const { center, size } = box;`.

packages/lib/src/interactions/box.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const centerClampingBox = Box.empty(area.center).expandBySize(
+      Math.max(areaWidth - size.width, 0),
+      Math.max(areaHeight - size.height, 0)
+    );
+
+    const shift = centerClampingBox
+      .clampPoint(center, new Vector3())
+      .sub(center)
+      .setZ(0); // cancel `z` shift
```
axelboc, line 55: The downside of working with `Box3` and `Vector3`: we end up having to correct the `z` component in a few places so the camera keeps `z=5` and doesn't move behind the scene... Perhaps we can revisit this in the future if we think of something clever.

packages/lib/src/interactions/hooks.ts
```diff
@@ -5,46 +5,67 @@ import { useCallback, useEffect } from 'react';
...
+      const fovHeight = canvasSize.height * scale.y;
 
-      const clampedPosition = clampPositionToArea(
-        worldPt,
-        { width, height },
-        visSize
+      // Create box to clamp target camera position, to disallow panning outside of visualization bounds
+      const clampingBox = Box.empty(new Vector3(0, 0, position.z)).expandBySize(
+        Math.max(visSize.width - fovWidth, 0),
+        Math.max(visSize.height - fovHeight, 0)
```
axelboc, line 38: This is a similar computation as in `keepWithin`, but the center of the clamping box is different. Instead of trying to abstract things like we had done with `clampPositionToArea`, I kept the duplication. We can revisit this later as well if needed.

packages/lib/src/interactions/hooks.ts
```diff
@@ -5,46 +5,67 @@ import { useCallback, useEffect } from 'react';
...
       );
 
-      position.set(clampedPosition.x, clampedPosition.y, position.z);
-
+      const target = clampingBox.clampPoint(worldPt, new Vector3());
+      position.copy(target);
```
axelboc, line 42: Could also write this `clampingBox.clampPoint(worldPt, position)`, but I think it's better to make the mutation of the `position` vector explicit.

packages/lib/src/interactions/hooks.ts
```diff
@@ -5,46 +5,67 @@ import { useCallback, useEffect } from 'react';
...
+  return useCallback(
+    (zoomBox: Box) => {
+      const { width, height } = canvasSize;
+
+      // Update camera scale first (since `moveCameraTo` relies on camera scale)
+      const { width: zoomWidth, height: zoomHeight } = zoomBox.size;
+      camera.scale.set(zoomWidth / width, zoomHeight / height, 1);
+
+      // Then move camera position
+      moveCameraTo(zoomBox.center);
```
axelboc, line 66: As mentioned, I might remove `moveCameraTo` and just set the camera position directly in a future PR.

packages/lib/src/interactions/hooks.ts
```diff
@@ -88,17 +109,16 @@ export function useZoomOnWheel(
...
+      // Scale change in position according to zoom
+      const delta = camera.position.clone().sub(worldPt);
+      if (sourceEvent.deltaY < 0) {
+        delta.multiply(zoomVector);
+      } else {
+        delta.divide(zoomVector);
+      }
 
-      moveCameraTo(oldPosition.add(scaledDelta));
+      moveCameraTo(worldPt.clone().add(delta));
```
axelboc, line 121: Small refactoring to make the mutations of `delta` and the use of `worldPt` more explicit.

packages/lib/src/vis/shared/RatioEnforcer.tsx
```diff
@@ -17,7 +17,6 @@ function RatioEnforcer() {
     const targetScale = Math.max(camera.scale.x, camera.scale.y);
     camera.scale.x = targetScale;
     camera.scale.y = targetScale;
-    camera.updateMatrixWorld();
```
axelboc, line 20: As explained, this is no longer needed now that `moveCameraTo` no longer relies on the camera's projection matrix.

packages/lib/src/interactions/SelectToZoom.tsx
loichuder: Yes, that makes sense

packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -1,69 +1,76 @@
...
+    if (!keepRatio) {
+      return zoomBox;
     }
 
-    const zoomRect = getEnclosedRectangle(worldStart, worldEnd);
-    const { center: zoomRectCenter } = zoomRect;
+    const fovBox = Box.empty(position).expandBySize(
+      width * scale.x,
+      height * scale.y
+    );
```
loichuder, line 33: We could store this box in the context if needed

loichuder (approved):
Some nitpicks here and there but overall, I think this improves greatly the code.

Well done :1st_place_medal:

packages/lib/src/interactions/box.ts
```diff
@@ -0,0 +1,61 @@
...
+  public static empty(center = ZERO_VECTOR) {
+    return new Box(center.clone(), center.clone());
+  }
+
+  public static fromPoints(...points: Vector3[]) {
+    return new Box().setFromPoints(points);
+  }
+
+  public expandBySize(width: number, height: number): this {
+    return this.expandByVector(new Vector3(width, height, 0).divideScalar(2));
```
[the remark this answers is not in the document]
axelboc: Hmm nope :smile:

packages/lib/src/interactions/box.ts
```diff
@@ -0,0 +1,61 @@
...
+    const { center, size } = this;
+    const { width: areaWidth, height: areaHeight } = area.size;
+
+    const centerClampingBox = Box.empty(area.center).expandBySize(
+      Math.max(areaWidth - size.width, 0),
+      Math.max(areaHeight - size.height, 0)
+    );
+
+    const shift = centerClampingBox
+      .clampPoint(center, new Vector3())
```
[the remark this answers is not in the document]
axelboc: The problem is that with your suggestion, `clampedCenter` gets mutated, so after the second statement, it no longer holds what it says it holds. And if we add `clone()`, it's less efficient... not by much, obviously, but still.

packages/lib/src/interactions/hooks.ts
axelboc: Yeah, but with `keepWithin`, we'd be creating a box, clamping the center of that box to another box, subtracting the original center to the clamped point to get a `shift` vector, and then translating the box by that vector, only to get the center of the translated box...  Whereas in `moveCameraTo`, all we only need is to clamp a point to a box, that's it.

Basically, using the `Box` abstraction in `moveCameraTo` feels overkill and inefficient.

I could create a helper such as `getFovSize(camera, context)` to abstract away the `canvasSize * scale` computation, at least... And maybe let `expandBySize` accept a `Size` object instead of two values... Not sure.

packages/lib/src/interactions/SelectToZoom.tsx
axelboc: Well, it depends on the camera scale, so it would re-render consumers of the context on every wheel/selection zoom. I'm considering more adding a camera-based utility similar to `getWorldFOV`, like `getFovSize(camera, context)`.

packages/lib/src/interactions/box.ts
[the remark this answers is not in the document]
loichuder: All right, no worries keeping it like this then. 

I just had to read this several times to understand what it does and my reflex in this case is to suggest a change.

packages/lib/src/interactions/hooks.ts
loichuder: > Basically, using the Box abstraction in moveCameraTo feels overkill and inefficient.

In terms of implementation perhaps but conceptually, I think it makes a lot of sense. You are already creating a `clampingBox` so there is only the need for a a second one:

```ts
      // Similar to `clampingBox`
      const visBox = new Box().setFromCenterAndSize(
        new Vector3(0, 0, position.z),
        new Vector3(visSize.width, visSize.height, 0)
      );
      // New box to be added
      const fovBox = new Box().setFromCenterAndSize(
        worldPt,
        new Vector3(fovWidth, fovHeight, 0)
      );
     
      // Actual operation
      fovBox.keepWithin(visBox);

      position.copy(zoomBox.center);
```

Having a `setFromCenterAndSize` that accepts a `Size` object rather than a `Vector3` could also help simplifying the code.

axelboc: Yeah, alright, though I prefer the API `Box.empty(position).expandBySize(width, height)`. I could add a static method like `Box.fromCenterAndSize()`, but it seems like it should accept two arguments, not three, as I like passing the size as two values.

axelboc: Done, much nicer indeed. I've added a static method called `Box.fromSize(size: Size)` to create a box with a given size and centered on `(0, 0, 0)`. If the need is for a box with a different center, then `Box.empty(center).expandBySize(width, height)` is still the way to go.

loichuder: Fine for me. You merge whenever you want :octopus:

PR #1324, opened 2022-12-14
Add nice property to AxisConfig

PeterC-DLS:
This passes through to D3 and allows domains to be extended to nice values

### DIFF
--- a/packages/lib/src/vis/shared/Axis.tsx
+++ b/packages/lib/src/vis/shared/Axis.tsx
@@ -50,13 +50,13 @@ function Axis(props: Props) {
     isIndexAxis,
     showGrid,
     label,
-    nice: _nice = false,
+    nice = false,
   } = config;
   // Restrain ticks scales to visible domains
   const scale = createAxisScale(scaleType, {
     domain,
     range: flipAxis ? [axisLength, 0] : [0, axisLength],
-    nice: _nice,
+    nice,
   });
 
   const [AxisComponent, GridComponent] = COMPONENTS[type];
--- a/packages/lib/src/vis/shared/Axis.tsx
+++ b/packages/lib/src/vis/shared/Axis.tsx
@@ -45,11 +45,12 @@ function Axis(props: Props) {
   const isAbscissa = type === 'abscissa';
   const axisLength = isAbscissa ? width : height;
 
-  const { scaleType = ScaleType.Linear, isIndexAxis, showGrid, label } = config;
+  const { scaleType = ScaleType.Linear, isIndexAxis, showGrid, label, nice : _nice = false } = config;
   // Restrain ticks scales to visible domains
   const scale = createAxisScale(scaleType, {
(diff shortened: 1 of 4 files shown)


axelboc:
Hi @PeterC-DLS, thanks for the suggestion. However, knowing that D3's `nice()` function effectively just [rounds the domain](https://github.com/d3/d3-scale/blob/main/src/nice.js#L15-L16), I'm not sure I see the value in complexifying the API of `VisCanvas` when this rounding can be done so easily by the consumer. Also, by using the domains "as is" inside `VisCanvas`, we provide a behaviour that is clear, explicit and easy to understand, which is very important in my opinion. What's your view on this @loichuder?

PeterC-DLS:
D3 adds this [nice](https://github.com/d3/d3-scale/blob/83555bd759c7314420bd4240642beda5e258db9e/src/linear.js#L19) function which extends the domain to nearest tick values. This makes auto-scaling axes a little nicer than rounding the min/max values.

axelboc:
Ah right, thanks for clarifying 👍

packages/lib/src/vis/shared/Axis.tsx
```diff
@@ -45,11 +45,18 @@ function Axis(props: Props) {
...
+    isIndexAxis,
+    showGrid,
+    label,
+    nice: _nice = false,
+  } = config;
   // Restrain ticks scales to visible domains
   const scale = createAxisScale(scaleType, {
     domain,
     range: flipAxis ? [axisLength, 0] : [0, axisLength],
+    nice: _nice,
```
axelboc, line 59: Any particular reason for renaming the variable here and below?

axelboc (commented):
Thanks for adding a story! I'll let Loïc take a look next week and we'll discuss it.

packages/lib/src/vis/shared/Axis.tsx
PeterC-DLS: Not really. I will tidy up and use literal property value shorthands too.

axelboc (approved):

loichuder:
No qualms about the implementation but not too keen on the idea.

In my view, `VisCanvas` is a low-level component where it is up to the caller to tune the domains to their needs. A `nice` parameter may have its place in higher level abstractions such as `...Vis` components.

Also, having `nice` means that the following code may no longer be valid:

```ts
// In a children of `VisCanvas`:
const { abscissaConfig, abscissaScale } = useVisCanvasContext();

// `nice` changed the domain of the scale so that domains are now differents
abscissaConfig.visDomain[0] === abscissaScale.domain[0]; // returns false if `nice` is used
abscissaConfig.visDomain[1] === abscissaScale.domain[1];  // returns false if `nice` is used
```

On the other hand, it may be relevant for `AxisConfig` to reflect the scale config. So if you feel a strong need for it, I can live with the option being there.

PR #1325, opened 2022-12-15
Fix link in contributing

PeterC-DLS:


axelboc:
🤔 seems to be working for me on both Firefox and Chrome: https://github.com/silx-kit/h5web/blob/main/CONTRIBUTING.md#quick-start-

The emoji is removed form the heading's `id`, but the space that precedes it is preserved as a dash by GitHub.

PeterC-DLS:
My IDE (the Mylyn Wikitext plugin) flags it up and also the link didn't work in its MD preview render without this change. I guess it does not support GH's flavour of Markdown.

axelboc:
I reckon we might as well remove the emoji completely, it will be neater. Would you mind @PeterC-DLS?

PeterC-DLS:
No, I can do that as well as another minor spacing fix.

PR #1328, opened 2022-12-16
Allow rendering app with preferred vis in tests and improve testing of vis tabs

axelboc:
Part three of bringing in test improvements from https://github.com/silx-kit/h5web/pull/1119

### DIFF
--- a/packages/app/src/__tests__/CorePack.test.tsx
+++ b/packages/app/src/__tests__/CorePack.test.tsx
@@ -1,8 +1,9 @@
 import { mockValues } from '@h5web/shared';
-import { screen } from '@testing-library/react';
+import { screen, within } from '@testing-library/react';
 
 import {
-  findVisSelectorTabs,
+  findSelectedVisTab,
+  findVisTabs,
   mockConsoleMethod,
   renderApp,
 } from '../test-utils';
@@ -11,11 +12,8 @@ import { Vis } from '../vis-packs/core/visualizations';
 test('visualise raw dataset', async () => {
   await renderApp('/entities/raw');
 
-  const tabs = await findVisSelectorTabs();
-  expect(tabs).toHaveLength(1);
-  expect(tabs[0]).toHaveTextContent(Vis.Raw);
-  expect(tabs[0]).toHaveAttribute('aria-selected', 'true');
-
+  await expect(findVisTabs()).resolves.toEqual([Vis.Raw]);
+  await expect(findSelectedVisTab()).resolves.toBe(Vis.Raw);
   await expect(screen.findByText(/"int": 42/)).resolves.toBeVisible();
 });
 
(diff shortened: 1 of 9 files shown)


packages/app/src/__tests__/CorePack.test.tsx
```diff
@@ -11,11 +12,8 @@ import { Vis } from '../vis-packs/core/visualizations';
 test('visualise raw dataset', async () => {
   await renderApp('/entities/raw');
 
-  const tabs = await findVisSelectorTabs();
-  expect(tabs).toHaveLength(1);
-  expect(tabs[0]).toHaveTextContent(Vis.Raw);
-  expect(tabs[0]).toHaveAttribute('aria-selected', 'true');
-
+  await expect(findVisTabs()).resolves.toEqual([Vis.Raw]);
```
axelboc, line 15: Async for now, until I bring back the testing code that deals with loaders.

packages/app/src/__tests__/CorePack.test.tsx
```diff
@@ -31,101 +29,105 @@ test('visualise scalar dataset', async () => {
...
   jest.runAllTimers();
   await expect(screen.findByRole('figure')).resolves.toBeVisible();
 
-  // Move to third slice to check that entire dataset is fetched
+  // Move to third slice
   await user.type(d0Slider, '{ArrowUp}');
 
-  await expect(screen.findByRole('figure')).resolves.toBeVisible();
+  // Wait for new slicing to apply to Line visualization to confirm that no more slow fetching is performed
+  await expect(screen.findByTestId('2,0,x', undefined)).resolves.toBeVisible();
```
loichuder, line 130: Could you elaborate on the reason of this change and the need for a `testid` ?

packages/app/src/__tests__/DimensionMapper.test.tsx
```diff
@@ -20,25 +21,31 @@ test('display mapping for X axis when visualizing 2D dataset as Line', async ()
...
 });
 
-test('display mappings for X and Y axes when visualizing 2D dataset as Heatmap', async () => {
-  const { user, selectVisTab } = await renderApp('/nD_datasets/twoD');
-
-  await selectVisTab(Vis.Heatmap);
+test('control mapping for X and Y axes when visualizing 2D dataset as Heatmap', async () => {
+  const { user } = await renderApp({
+    initialPath: '/nD_datasets/twoD',
+    preferredVis: Vis.Heatmap,
```
loichuder, line 45: Not sure if we want to set the preferredVis here. 

We want to rely on the fact that the `Heatmap` vis is the default vis for 2D datasets, no ? And let `preferredVis` be the focus of a dedicated test ?

packages/app/src/__tests__/MetadataViewer.test.tsx
```diff
@@ -1,14 +1,12 @@
...
-import { queryVisSelector, renderApp } from '../test-utils';
+import { renderApp } from '../test-utils';
 
 test('switch between "display" and "inspect" modes', async () => {
   const { user } = await renderApp();
 
   // Switch to "inspect" mode
   await user.click(screen.getByRole('tab', { name: 'Inspect' }));
-
-  expect(queryVisSelector()).not.toBeInTheDocument();
```
loichuder, line 11: Doesn't hurt to keep it ?

packages/app/src/test-utils.tsx
```diff
@@ -15,17 +16,25 @@ interface RenderAppResult extends RenderResult {
...
     initialPath: '/',
     ...optsObj,
   };
 
+  if (preferredVis) {
+    window.localStorage.setItem(
+      'h5web:preferredVis',
+      JSON.stringify(preferredVis)
+    );
+  }
```
loichuder, line 36: I know we clear the `localStorage` after each test but should we need to clear it aso **before** if we start relying on this setting ?

packages/app/src/__tests__/CorePack.test.tsx
```diff
@@ -31,101 +29,105 @@ test('visualise scalar dataset', async () => {
   // Integer scalar
   const { selectExplorerNode } = await renderApp('/entities/scalar_int');
 
+  await expect(findVisTabs()).resolves.toEqual([Vis.Scalar]);
```
loichuder, line 32: Looks better :sparkles: !

loichuder (approved):

packages/app/src/__tests__/CorePack.test.tsx
axelboc: Ah yeah, so basically we don't have a way in the DOM to know when the visualization component has actually "applied" the new mapping/slicing. The title doesn't change, the color bar bounds may or may not change, etc. Since there is a debounce, we also can't count on the fact that the visualization is updated right away, so the `testid` gives us a way to wait for the visualization to update, which guarantees that the test has completed and that React won't perform any more work after that.

packages/app/src/__tests__/DimensionMapper.test.tsx
axelboc: We're testing the `DimensionMapper` here, not the detection/selection of the supported visualizations (which is tested in `CorePack.test.ts`), and not the preferred visualization feature (which is tested in `VisSelector.test.ts`).

packages/app/src/__tests__/MetadataViewer.test.tsx
axelboc: Fair enough

packages/app/src/test-utils.tsx
axelboc: Doing both is not needed, since when Jest starts, it creates a fresh jsdom environment, so there's never anything in `localStorage`. We could clear it _before_ instead of after each test, but conceptually, it feels like a clean up step rather than a preparation step.

loichuder: All right :+1:

packages/app/src/__tests__/CorePack.test.tsx
axelboc: In principle, I'd agree with you, but in this case, the information is really not easy to convey to a user by text (unless you had something in mind), and it really is the mapping/slicing information that we need to check, because it's the thing that changes. Adding this information to a caption/title would also duplicate information that is already present on the UI, thus actually _reducing_ accessibility (by slowing down screen reader users or adding clutter to the UI for sighted users).

I think in this case, `testid` (and the comment) is what keeps the test clear/concise/readable and also reliable. It makes the underlying testing complexity explicit instead of trying to hide it behind a workaround.

loichuder: Fair point. Let's keep the `testid` then :+1:

PR #1330, opened 2022-12-16
Abstract away enabling fake timers in tests

axelboc:
Part five of bringing in test improvements from https://github.com/silx-kit/h5web/pull/1119


loichuder (approved):
Nice :slightly_smiling_face:

PR #1331, opened 2022-12-16
Add doc on feature tests and make minor test improvements

axelboc:
Part six, and final, of bringing in test improvements from https://github.com/silx-kit/h5web/pull/1119

I diffed all the test files and brought a few more things over, including some of the documentation from: https://github.com/silx-kit/h5web/pull/1119/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055

Only thing that's left and that can't be ported is all the async `Suspense` stuff: waiting for all loaders to disappear; using `getBy` instead of `findBy` in tests with real timers; and using `findBy` with a timeout in tests with fake timers. This stuff requires upgrading  React Testing Library to v13, which requires upgrading React to v18.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,6 +12,9 @@
   - [Fixing and formatting](#fixing-and-formatting)
   - [Editor integration](#editor-integration)
 - [Testing](#testing)
+  - [Feature tests](#feature-tests)
+    - [Fake timers](#fake-timers)
+    - [Debugging](#debugging)
   - [Visual regression](#visual-regression)
 - [Deployment](#deployment)
 - [Release process](#release-process)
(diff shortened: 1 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -206,6 +209,62 @@ install the recommended extensions.
...
+component that suspends inside a test **finishes loading before the end of that
+test**. This is where Testing Library's
+[asynchronous methods](https://testing-library.com/docs/dom-testing-library/api-async)
+come in.
+
+#### Fake timers
+
+To allow developing and testing loading interfaces, as well as features like
+cancel/retry, `MockProvider` adds an artificial delay of 3s (`SLOW_TIMEOUT`) to
+some requests, notably to value requests for datasets prefixed with `slow_`.
```
loichuder, line 242:
```suggestion
value requests for datasets prefixed with `slow_`.
```

CONTRIBUTING.md
```diff
@@ -206,6 +209,62 @@ install the recommended extensions.
...
+
+```ts
+console.debug(prettyDOM()); // if you use `console.log` without mocking it, the test will fail
+console.debug(prettyDOM(screen.getByText('foo'))); // you can also print out a specific element
+```
+
+To ensure that the entire DOM is printed out in the terminal, you may have to
+set environment variable `DEBUG_PRINT_LIMIT`
+[to a large value](https://testing-library.com/docs/dom-testing-library/api-debugging#automatic-logging)
+when calling `pnpm test`.
```
loichuder, line 266: Should we add the variable in the `.env` file ?

loichuder (approved):

CONTRIBUTING.md
axelboc: There's also a delay when fetching the metadata for entity `slow_metadata`... :joy:

CONTRIBUTING.md
axelboc: Done. I've added a separate section to explain that devs should override this in a file called `.env.test.local`

axelboc: I remembered why this wasn't in `.env`.

`.env` is in the `demo` project. When we run the tests in the repo, `.env` is not read... I'll revert on `main`.

loichuder: How do you proceed then to set this limit ? Thourgh the shell ?

axelboc: yep

axelboc: as mentioned on the RTL website

PR #1332, opened 2022-12-20
Replace `SelectionRect/Line` with `SvgRect/Line` in world space

axelboc:
POC of having low-level SVG components that work in world space instead of data space. I also introduce a component called `DataToWorld` that consumers can use directly instead of having to create a component just to retrieve `dataToWorld` from context. The `PersistDataSelection` story shows an example of usage.

The idea behind this is:

- to avoid unnecessary `world -> data -> world` back-and-forth conversions => components such as `SelectionTool` can work in world space all the way, so the extra conversions are wasteful and complicate the code;
- to avoid unnecessary `data -> world` conversions when panning/zooming => `dataToWorld` is unaffected by camera scale/position, so components that need to update on every frame (like a persisted selection rectangle) shouldn't have to do this extra conversion;
- to keep a clear layer of separation between data space and world space => everything that depends on the camera should work in world space, and everything else in data space;
- to allow consumers to compose their own SVG-based, world-space components without having to do any space conversions themselves => for instance for a ROI component with a circumscribed circle (in pseudo-code):

```tsx
function MyROI(props: { coords: Rect3 }) {
  const { center, size } = Box.fromPoints(...coords);

  return (
    <>
      <SvgRect coords={coords} />
      <SvgCircle center={center} radius={Math.max(size.width, size.height) / 2} />
      <SvgDragHandle coords={coords[0]} /> { /* naïve implementation, obviously */}
      <SvgDragHandle coords={coords[1]} />
    </>
  );
}

// // data -> world conversion => clear, efficient layer of separation
<DataToWorld coords={dataCoords}>
  (...worldCoords) => <MyROI coords={worldCoords} />
</DataToWorld>
```

---

If this is to your liking, I will document `DataToWorld`, `SvgRect` and `SvgLine` in Storybook (and ask for your review again) before merging.

**EDIT:** done!


apps/storybook/src/SelectionTool.stories.tsx
```diff
@@ -95,11 +88,9 @@ CustomStyles.argTypes = {
   stroke: { control: { type: 'color' } },
 };
 
-export const Persisted: Story<
-  Pick<TemplateProps, 'selectionModifierKey' | 'panModifierKey'>
-> = (args) => {
+export const PersistedDataSelection: Story<TemplateProps> = (args) => {
   const { selectionModifierKey, panModifierKey } = args;
-  const [persistedSelection, setPersistedSelection] = useState<Selection>();
+  const [persistedDataSelection, setPersistedDataSelection] = useState<Rect2>();
```
axelboc, line 93: Consumers will typically want to persist a selection in data space.

apps/storybook/src/SelectionTool.stories.tsx
```diff
@@ -122,24 +113,33 @@ export const Persisted: Story<
...
-        <SelectionRect
-          startPoint={persistedSelection.data[0]}
-          endPoint={persistedSelection.data[1]}
-        />
+      {persistedDataSelection && (
+        <DataToWorld coords={persistedDataSelection}>
+          {(...worldCoords) => (
+            <SvgRect coords={worldCoords} fill="teal" fillOpacity="0.6" />
+          )}
+        </DataToWorld>
```
axelboc, line 131: This shows how to use `DataToWorld` along with the new `SvgRect` component. Without it, since the story component renders `VisCanvas`, I wouldn't be able to retrieve `dataToWorld` from the context without creating a new component.

packages/shared/src/models-vis.ts
```diff
@@ -36,3 +36,7 @@ export interface Dims {
   rows: number;
   cols: number;
 }
+
+// MappedTupple<number[], string> => string[]
+// MappedTupple<[number, number], string> => [string, string]
+export type MappedTuple<T, U> = { [index in keyof T]: U };
```
axelboc, line 42: This is the only slightly tricky bit, really. Hopefully the commented examples are self-explanatory.

loichuder (approved):
Slick :100:

loichuder (approved):

PR #1335, opened 2023-01-04
Fix experimental vis not appearing

loichuder:
I think our process for enabling experimenal vis was not working properly: the SurfaceVis did appear only when hot-reloading a dev version of the project.

I think this is due to the fact that `window.H5WEB_EXPERIMENTAL` is read before it is set by the `DemoApp` component. Since `window.H5WEB_EXPERIMENTAL` is `undefined`, the SurfaceVis is not enabled (expect if the reading of the variable is triggerred by a hot reloading).

To fix this, I now read `window.H5WEB_EXPERIMENTAL` in the `supportsDataset` function of the SurfaceVis so that its value is always up to date.

### DIFF
--- a/packages/app/src/visualizer/utils.ts
+++ b/packages/app/src/visualizer/utils.ts
@@ -83,8 +83,8 @@ function getSupportedCoreVis(
     (vis) => isDataset(entity) && vis.supportsDataset(entity, attrValueStore)
   );
 
-  return supportedVis.length > 1 && supportedVis[0].name === Vis.Raw
-    ? supportedVis.slice(1)
+  return supportedVis.length > 1
+    ? supportedVis.filter((vis) => vis.name !== Vis.Raw)
     : supportedVis;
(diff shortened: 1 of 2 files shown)


packages/app/src/visualizer/utils.ts
```diff
@@ -83,8 +83,8 @@ function getSupportedCoreVis(
     (vis) => isDataset(entity) && vis.supportsDataset(entity, attrValueStore)
   );
 
-  return supportedVis.length > 1 && supportedVis[0].name === Vis.Raw
-    ? supportedVis.slice(1)
+  return supportedVis.length > 1
+    ? supportedVis.filter((vis) => vis.name !== Vis.Raw)
```
loichuder, line 87: Rewritten this to not rely on the ordering of the vis

axelboc (approved):
Ah of course, good call 👍

PR #1339, opened 2023-01-05
Fix selection change callback not called on first pointer move

axelboc:
Very minor, subtle bug: consumers of `SelectionTool` registering an `onSelectionChange` callback were not being notified of the first pointer move event. `SelectionTool` was calling `onSelectionStart` but not `onSelectionChange` after the first pointer move event. This is now corrected.

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -139,26 +139,24 @@ function SelectionTool(props: Props) {
   useEffect(() => {
     const { selection, isCancelled } = selectionState;
 
-    // Selection state is now defined => selection has started
-    if (!prevSelectionState?.selection && selection) {
-      onSelectionStartRef.current?.(selection);
-      return;
-    }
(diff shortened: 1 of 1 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -139,26 +139,24 @@ function SelectionTool(props: Props) {
...
-     * - the selection was not cancelled with Escape,
-     * - and the modifier key was not released before the pointer. */
-    if (prevSelectionState?.selection && !selection && !isCancelled) {
-      onSelectionEndRef.current?.(prevSelectionState.selection);
-      return;
-    }
+    if (selection) {
+      // Previous selection was undefined and current selection is now defined => selection has started
+      if (!prevSelectionState?.selection) {
+        onSelectionStartRef.current?.(selection);
```
axelboc, line 145: No `return` here, so `onSelectionChange` gets called.

loichuder (approved):

PR #1340, opened 2023-01-06
Invoke `SelectionTool` render prop and callbacks with transformed selection

axelboc:
I realised that the transformed selection as well as the future `isValid` boolean can be derived from the raw selection state; they don't need to be part of the local state. As a result, the justification for having an `isCancelled` flag in the state made in #1338 no longer holds.

So in this PR, I'm going back a little by first reverting #1338 (sorry for the back and forth...) Then, I compute the transformed selection (which I refer to as the "effective" selection in comments) as a derived state in a `useMemo`, and I pass this effective `selection` object to the render prop and to the callbacks.

Finally, now that the render prop of `SelectionTool` has access to both the raw and effective selections, I can refactor `SelectToZoom` in order to compute the effective zoom selection in a `transformSelection` callback instead of twice, in both the render prop and the `onSelectionEnd` callback.

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -22,7 +22,7 @@ import { boundWorldPointToFOV, getModifierKeyArray } from './utils';
 interface Props extends CommonInteractionProps {
   id?: string;
   transformSelection?: (selection: Selection) => Selection;
-  onSelectionStart?: (selection: Selection) => void;
+  onSelectionStart?: () => void;
   onSelectionChange?: (selection: Selection | undefined) => void;
   onSelectionEnd?: (selection: Selection) => void;
   children: (selection: Selection) => ReactNode;
@@ -132,7 +132,7 @@ function SelectionTool(props: Props) {
     if (selection) {
       // Previous selection was undefined and current selection is now defined => selection has started
       if (!prevSelection) {
-        onSelectionStartRef.current?.(selection);
+        onSelectionStartRef.current?.();
       }
 
       // Either way, current selection is defined, so invoke change callback
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -19,11 +19,6 @@ import type { CanvasEvent, CommonInteractionProps, Selection } from './models';
 import { MouseButton } from './models';
 import { boundWorldPointToFOV, getModifierKeyArray } from './utils';
 
-interface SelectionState {
-  selection: Selection | undefined;
-  isCancelled?: boolean;
-}
-
 interface Props extends CommonInteractionProps {
   id?: string;
   transformSelection?: (selection: Selection) => Selection;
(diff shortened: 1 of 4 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -19,18 +21,16 @@ import type { CanvasEvent, CommonInteractionProps, Selection } from './models';
...
-  selection: Selection | undefined;
-  isCancelled?: boolean;
-}
-
 interface Props extends CommonInteractionProps {
   id?: string;
   transformSelection?: (selection: Selection) => Selection;
-  onSelectionStart?: (selection: Selection) => void;
-  onSelectionChange?: (selection: Selection | undefined) => void;
+  onSelectionStart?: () => void;
```
axelboc, line 27: With #1339, `onSelectionChange` is always called after `onSelectionStart`, so it makes less sense to pass the `selection` object to `onSelectionStart`.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -19,18 +21,16 @@ import type { CanvasEvent, CommonInteractionProps, Selection } from './models';
...
+  onSelectionChange?: (
+    selection: Selection | undefined,
+    rawSelection: Selection
```
axelboc, line 30: `onSelectionChange` (as well as the render prop below) now receives the effective selection object as first parameter (i.e. the transformed selection), and the raw selection object as second parameter. The effective selection is the most important information so it makes sense for it to come first.

Note that the default `transformSelection` callback is just a pass-through, so the effective selection is guaranteed to be defined when the raw selection is defined.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -95,14 +92,12 @@ function SelectionTool(props: Props) {
...
       const boundWorldEnd = boundWorldPointToFOV(worldEnd, camera);
 
-      setSelectionState({
-        selection: transformSelectionRef.current({
-          world: [worldStart, boundWorldEnd],
-          data: [dataStart, worldToData(boundWorldEnd)],
-        }),
+      setRawSelection({
+        world: [worldStart, boundWorldEnd],
+        data: [dataStart, worldToData(boundWorldEnd)],
```
axelboc, line 97: I'm no longer storing the transformed/effective selection in the local state.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -130,49 +123,63 @@ function SelectionTool(props: Props) {
...
     { event: 'keydown' }
   );
 
-  useEffect(() => {
-    const { selection, isCancelled } = selectionState;
+  // Compute effective selection
+  const selection = useMemo(
+    () => rawSelection && transformSelectionRef.current(rawSelection),
+    [rawSelection, transformSelectionRef]
+  );
```
axelboc, line 136: The transformed/effective selection is now computed as a derived state, during render. It's memoised, of course, so as to not trigger the effect unnecessarily.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -130,49 +123,63 @@ function SelectionTool(props: Props) {
...
+
+  // Keep track of previous effective selection state
+  const prevSelection = usePrevious(selection);
```
axelboc, line 139: We now keep track of the previous "effective" selection. This makes more sense, since the effective selection is what the consumer cares most about. This is why I named the variables `selection` and `prevSelection`, and the local state `rawSelection`, instead of the opposite (i.e. `effectiveSelection` and `selection` for the local state).

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -130,49 +123,63 @@ function SelectionTool(props: Props) {
...
+      if (!prevSelection) {
+        onSelectionStartRef.current?.();
       }
 
-      // Either way, current selection is defined, so invoke change callback
+      // Either way, current selection is defined, so invoke change callback with effective selection object
       onSelectionChangeRef.current?.(
-        isModifierKeyPressed ? selection : undefined // don't pass selection object if user is not pressing modifier key
+        isModifierKeyPressed ? selection : undefined, // don't pass selection object if user is not pressing modifier key
+        rawSelection
```
axelboc, line 153: Note that contrary to the effective selection, the raw selection is always passed to the callback.

PR #1341, opened 2023-01-09
Use `Vector3` in all coordinate spaces

axelboc:
Using `Vector2` in HTML and data spaces is actually really inconvenient -- it was making it difficult to explore solutions for switching `SelectionTool` to HTML space. By using `Vector3` in all three coordinate spaces, we can for instance use the `Box` abstraction in all three spaces without pain.


apps/storybook/src/SelectionTool.stories.tsx
```diff
@@ -1,4 +1,4 @@
-import type { ModifierKey, Rect2, Selection } from '@h5web/lib';
+import type { ModifierKey, Rect, Selection } from '@h5web/lib';
```
axelboc, line 1: `Rect2` is removed, and `Rect3` becomes `Rect`.

loichuder (approved):

PR #1342, opened 2023-01-09
Move camera-based space conversion utilities to context

axelboc:
`dataToHtml` and the like were just too annoying to use directly, without `useCameraState`. You had to pass the entire `context` object to them, which was cumbersome. By moving them into the context, they can access what they need directly, which also means we no longer need internal utilities `cameraToHtml` and `htmlToCamera`.

Note, however, that `dataToHtml` and the others still accept the `camera` as first parameter to avoid falling back into the problem we had initially, which was that it wasn't clear that those utilities relied on the camera.

My brain also likes the consistency of having all the space-conversion utilities in one place.


apps/storybook/src/Context.stories.mdx
```diff
@@ -15,17 +15,21 @@ useVisCanvasContext(): VisCanvasContextValue
...
+| `dataToWorld`    | Converts a vector from data space to world space (calls `abscissaScale()` and `ordinateScale`)                                                                    | <code>(vec: Vector3) => Vector3</code>                 |
+| `dataToHtml`     | Converts a vector from data space to HTML space                                                                                                                   | <code>(camera: Camera, vec: Vector3) => Vector3</code> |
+| `worldToHtml`    | Converts a vector from world space to HTML space                                                                                                                  | <code>(camera: Camera, vec: Vector3) => Vector3</code> |
+| `worldToData`    | Converts a vector from world space to data space (calls `abscissaScale.invert()` and `ordinateScale.invert`)                                                      | <code>(vec: Vector3) => Vector3</code>                 |
```
loichuder, line 31:
```suggestion
| `worldToData`    | Converts a vector from world space to data space (calls `abscissaScale.invert` and `ordinateScale.invert`)                                                      | <code>(vec: Vector3) => Vector3</code>                 |
```

loichuder (approved):
Good call :+1:

PR #1343, opened 2023-01-09
Compute selection in HTML space in `SelectionTool`

axelboc:
Almost there!

Since checking the validity of a selection requires HTML coordinates, I'm introducing those to `SelectionTool`. You'll notice that it actually simplifies a bunch of stuff that can be done more easily in HTML space than in world space. This is due to the simple fact that, in HTML space, the field of view is always the canvas, regardless of the camera's position/scale.


packages/lib/src/interactions/SelectToZoom.tsx
```diff
@@ -1,57 +1,57 @@
...
 
-    return {
-      world: [zoomBox.min, zoomBox.max],
-      data: [worldToData(zoomBox.min), worldToData(zoomBox.max)],
-    };
+    const html = zoomBox.toRect();
+    const world = html.map((pt) => htmlToWorld(camera, pt)) as Rect;
+    const data = world.map(worldToData) as Rect;
+
+    return { html, world, data };
```
axelboc, line 47: Converting to all three spaces here as well.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -88,16 +93,14 @@ function SelectionTool(props: Props) {
         return;
       }
 
-      const { dataPt: dataStart, worldPt: worldStart } = startEvtRef.current;
-      const { worldPt: worldEnd } = evt;
-      const boundWorldEnd = boundWorldPointToFOV(worldEnd, camera);
+      const { htmlPt: htmlStart } = startEvtRef.current;
+      const html: Rect = [htmlStart, canvasBox.clampPoint(evt.htmlPt)];
```
axelboc, line 97: Now we just clamp the HTML end point to the canvas.

loichuder (approved):

PR #1344, opened 2023-01-10
Revert adding `SvgRect` and `SvgLine` and render SVG elements directly instead

axelboc:
The render prop of `SelectionTool` and `AxialSelectionTool` now have access to the current selection in HTML space, so `SvgRect` and `SvgLine` are no longer relevant.

Using raw SVG elements directly means that we can render a bunch of them together inside a single `SvgElement` (which is used to move out of the R3F rendered and back into the ReactDOM renderer). In fact, rendering a single shape is rarely the intention (e.g. ROI drag handles).

The other benefit of this whole refactoring process is that users no longer have to bother dealing with world space, which reduces cognitive load. Indeed, `DataToWorld` now becomes `DataToHtml` -- the goal is similar: to convert a bunch of points from data space to HTML space without having to retrieve the camera and `dataToHtml`, but HTML space is much easier to reason about since it's what the user sees on the screen.


axelboc (commented):
I'm planning on adding a few more stories in future PRs, before the release, to document `SvgElement` as well as the `transformSelection` prop.

packages/lib/src/interactions/utils.ts
```diff
@@ -1,7 +1,31 @@
...
+
+import type { ModifierKey, Rect } from './models';
 
 export function getModifierKeyArray(
   keys: ModifierKey | ModifierKey[] | undefined = []
 ): ModifierKey[] {
   return Array.isArray(keys) ? keys : [keys];
 }
+
+export function getSvgRectAttrs(rect: Rect): SVGProps<SVGRectElement> {
```
loichuder, line 11: I get why you used `SVGProps<SVGRectElement>` but it is a bit too broad to my taste.

I think dedicated types `{x: number, y: number, width: number, height: number}` would be much clearer for consumers for the function.

Also, you could rename the function to `getSvgRectCoords` or similar since it only returns coordinates.

packages/lib/src/interactions/utils.ts
```diff
@@ -1,7 +1,31 @@
...
+
+  return {
+    x: Math.min(start.x, end.x),
+    y: Math.min(start.y, end.y),
+    width: Math.abs(end.x - start.x),
+    height: Math.abs(end.y - start.y),
+  };
+}
+
+export function getSvgLineAttrs(rect: Rect): SVGProps<SVGLineElement> {
```
loichuder, line 22: Same comment here

loichuder (approved):

packages/lib/src/interactions/utils.ts
axelboc: I went with the renaming but I chose to remove the return type altogether and let TS infer it. I had put a return type as a way to check that I wasn't returning invalid `rect` attributes from inside the function, but this is checked just as well when the function is used on a `rect` element. I could have used `Pick<SVGProps<...>, '...'>` but letting TypeScript do it's job seems much cleaner here.

PR #1345, opened 2023-01-11
Support validating selections

axelboc:
At last, the `SelectionTool` can now report whether the selection is valid or not!

By default, it checks for a default width and height of 20px, which seems to be a sensible default behaviour for rectangle selections. Since the validity check is performed on the transformed selection, this behaviour works perfectly for both `SelectToZoom` and `AxialSelectToZoom`:

valid-selection

valid-selection-rect

However, the default validation behaviour does not work for line selection (e.g. for drawing line profiles), as in this case, what matters is the length of the line, not the size of the box. To cover this use case (and most likely others), `SelecionTool` simply accepts a custom validation function. This function has access to the current selection in all three coordinate spaces, so it's really flexible:

valid-selection-line

### DIFF
--- a/packages/lib/src/interactions/AxialSelectToZoom.tsx
+++ b/packages/lib/src/interactions/AxialSelectToZoom.tsx
@@ -5,8 +5,10 @@ import AxialSelectionTool from './AxialSelectionTool';
 import SvgElement from './SvgElement';
 import Box from './box';
 import { useZoomOnSelection } from './hooks';
-import type { Selection, CommonInteractionProps } from './models';
+import type { CommonInteractionProps } from './models';
 import { getSvgRectCoords } from './utils';
+
+const MIN_SIZE = 20;
 
 interface Props extends CommonInteractionProps {
   axis: Axis;
@@ -18,28 +20,24 @@ function AxialSelectToZoom(props: Props) {
   const { visRatio } = useVisCanvasContext();
   const zoomOnSelection = useZoomOnSelection();
 
-  function onSelectionEnd(selection: Selection) {
-    const { size } = Box.fromPoints(...selection.html);
-    if (size.width > 0 && size.height > 0) {
-      zoomOnSelection(selection);
-    }
-  }
-
   return (
     <AxialSelectionTool
       axis={axis}
       id={`${axis.toUpperCase()}SelectToZoom`}
       modifierKey={modifierKey}
       disabled={visRatio !== undefined || disabled}
-      onSelectionEnd={onSelectionEnd}
+      validate={({ html }) => Box.fromPoints(...html).hasMinSize(MIN_SIZE)}
+      onValidSelection={zoomOnSelection}
     >
-      {({ html: htmlSelection }) => (
+      {({ html: htmlSelection }, _, isValid) => (
         <SvgElement>
           <rect
             {...getSvgRectCoords(htmlSelection)}
             fill="white"
+            fillOpacity={isValid ? 0.25 : 0}
             stroke="black"
-            fillOpacity={0.25}
+            strokeDasharray={isValid ? undefined : 4}
+            style={{ transition: 'fill-opacity 0.2s' }}
           />
         </SvgElement>
       )}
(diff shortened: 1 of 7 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -134,12 +146,19 @@ function SelectionTool(props: Props) {
...
-  // Keep track of previous effective selection state
+  // Determine if effective selection respects the minimum size threshold
+  const isValid = useMemo(
+    () => !!selection && validateRef.current(selection),
+    [selection, validateRef]
+  );
+
+  // Keep track of previous effective selection and validity
   const prevSelection = usePrevious(selection);
+  const prevIsValid = usePrevious(isValid);
```
axelboc, line 161: Very slightly annoying to have to keep track of the previous validation state as well (for `onSelectionEnd` and `onValidSelection`), but the alternatives are awful:

- like adding `isValid` to the `Selection` mode, which doesn't make sense since `transform()` receives a `Selection` object but the validity of the selection is checked afterwards;
- or attempting yet again to store the transformed selection and validity flag in local state...

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -164,25 +184,33 @@ function SelectionTool(props: Props) {
      * - selection was not cancelled with Escape, and
      * - modifier key was released after pointer (if applicable). */
     if (prevSelection && hasSuccessfullyEndedRef.current) {
+      assertDefined(prevIsValid);
```
axelboc, line 187: `prevIsValid` is `undefined` on initial render, but since we know that `prevSelection` is defined, we know we're no longer in the initial render.

packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -28,33 +29,44 @@ import { getModifierKeyArray } from './utils';
...
 }
 
 function SelectionTool(props: Props) {
   const {
     id = 'Selection',
     modifierKey,
     disabled,
-    transformSelection = (selection) => selection,
+    transform = (selection) => selection,
+    validate = ({ html }) => Box.fromPoints(...html).hasMinSize(20),
```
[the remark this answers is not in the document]
loichuder: Yeah. Given the general character of `SelectionTool`, I am not sure it makes sense to have a default validation. It may be weird for first-time consumers of the component.

I would put this validation behaviour in the components where it is needed.

packages/lib/src/interactions/SelectionTool.tsx
loichuder: :shrug:

packages/lib/src/interactions/SelectionTool.tsx
loichuder: Related to https://github.com/silx-kit/h5web/pull/1345/files#r1066859012: if you add `isValid to the Selection state, this `assert` may not be needed. This could give an argument in favour of this solution :shrug:

loichuder (approved):

packages/lib/src/interactions/SelectionTool.tsx
axelboc: I'm gonna keep it as is. Having a simple state makes all the event handling logic, as well the logical structure of the component, much simpler to understand. If a refactoring is needed in the future, I think it will be around those two `usePrevious` calls.

packages/lib/src/interactions/SelectionTool.tsx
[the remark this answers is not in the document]
axelboc: Interesting thought, you're probably right. I'll move it out :+1:

packages/lib/src/interactions/AxialSelectToZoom.tsx
```diff
@@ -18,28 +17,23 @@ function AxialSelectToZoom(props: Props) {
...
+      {({ html: htmlSelection }, _, isValid) => (
         <SvgElement>
           <rect
             {...getSvgRectCoords(htmlSelection)}
             fill="white"
+            fillOpacity={isValid ? 0.25 : 0}
             stroke="black"
-            fillOpacity={0.25}
+            strokeDasharray={isValid ? undefined : 4}
+            style={{ transition: 'fill-opacity 0.2s' }}
```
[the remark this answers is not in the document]
axelboc: I'll do this after merging the two PRs.

PR #1347, opened 2023-01-12
Add dashed and thick line support to DataCurve

PeterC-DLS:
This PR enhances `DataCurve` by adding new attributes to its props which:
1. Allow dashed lines by using `LineDashedMaterial`.
2. Support line widths other than 1. WebGL  does not support setting its line primitive to other widths so enable them by using an different Line implementation from `three-stdlib`.

### DIFF
--- a/packages/lib/src/vis/line/DataCurve.tsx
+++ b/packages/lib/src/vis/line/DataCurve.tsx
@@ -5,10 +5,14 @@ import type {
   Object3DNode,
   ThreeEvent,
 } from '@react-three/fiber';
-import { useCallback, useLayoutEffect, useState } from 'react';
-import { BufferGeometry, Color, Line, Vector2, Vector3 } from 'three';
-import { Line2, LineGeometry, LineMaterial } from 'three-stdlib';
-import type { LineMaterialParameters } from 'three-stdlib';
+import { useCallback, useLayoutEffect, useMemo, useState } from 'react';
+import { BufferGeometry, Color, Line, Vector2, type Vector3 } from 'three';
+import {
+  Line2,
+  LineGeometry,
+  LineMaterial,
+  type LineMaterialParameters,
+} from 'three-stdlib';
 
 import ErrorBars from './ErrorBars';
 import GlyphMaterial from './GlyphMaterial';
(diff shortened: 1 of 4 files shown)


packages/lib/src/vis/line/DataCurve.tsx
```diff
@@ -52,15 +78,34 @@ function DataCurve(props: Props) {
...
+  const lineGeometry = useState(() => new LineGeometry())[0];
 
+  function getFlatData(): number[] {
+    return lineWidth === 1 ? [] : flattenCoordinates(points.data);
+  }
+  const flatData = useMemo(getFlatData, [lineWidth, points.data]);
+  if (lineWidth === 1 || showGlyphs) {
+    bufferGeometry.setFromPoints(points.data);
+  } else {
+    lineGeometry.setPositions(flatData);
```
axelboc, line 97: There's a bit of a performance issue here: [one loop](https://github.com/silx-kit/h5web/pull/1347/files#diff-87080b3b5aeb767584fe4e9df8460c5f30453d44d12ea990e960c03c93326a6bR54) for to go from `[vec1, vec2, ...]` to `[x1, y1, z1, x2, y2, z2, ...]`, then [another loop](https://github.com/pmndrs/three-stdlib/blob/main/src/lines/LineGeometry.js#L18) to go from that to line segments.

Any chance we could initialise a buffer geometry manually, in order to convert the input vectors array to a buffer of line segments in a single pass?

axelboc (commented):
Hey, thanks for that! It looks very promising.

Took me a bit of time to understand the need for `three-stdlib`, until I remembered that `lineWidth` gets ignored by WebGL. Maybe explain things a bit in the PR description next time? We'd really appreciate it! :wink:

Also, having a little heads up about what you're planning to contribute would be super nice! :grin: Issues are really good for that, as they allow discussing needs, scope and implementation strategy, which can often prevent wasting development time.

Thanks again for the contribution, though. It's really great to see `@h5web/lib` getting some traction :blush:

packages/lib/src/vis/line/DataCurve.tsx
```diff
@@ -95,17 +140,43 @@ function DataCurve(props: Props) {
...
+          {gapSize === 0 ? (
+            <lineBasicMaterial color={color} />
+          ) : (
+            <lineDashedMaterial
+              color={color}
+              dashSize={dashSize}
+              gapSize={gapSize}
+            />
+          )}
+        </line_>
```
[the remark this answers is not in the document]
PeterC-DLS: It's both performance and appearance.

packages/lib/src/vis/line/DataCurve.tsx
PeterC-DLS: I think this logic is a little buggy: we need both geometries set when `showGlyphs` and `lineWith != 1`. I could use  `LineSegments2` and conditionally compute two arrays with one loop.

PeterC-DLS:
Hi @axelboc, I emailed the above address Monday last week but haven't had any reply. Has a spam filter or two ate our messages?

axelboc:
Most likely, sorry about that! 😅 I've managed to find your email, so hopefully it works the other way around.

PR #1349, opened 2023-01-16
Document transforming selections and rendering SVG elements

axelboc:
Peek 2023-01-16 15-12

Peek 2023-01-16 15-37

### DIFF
--- a/apps/storybook/src/SvgElement.stories.module.css
+++ b/apps/storybook/src/SvgElement.stories.module.css
@@ -1,0 +1,12 @@
+.rect {
+  fill: teal;
+  fill-opacity: 0.2;
+  stroke: teal;
+  stroke-width: 2;
+  pointer-events: auto;
+}
+
+.rect:hover {
+  fill: darkgoldenrod;
+  stroke: darkorange;
+}
--- a/apps/storybook/src/SvgElement.stories.tsx
+++ b/apps/storybook/src/SvgElement.stories.tsx
@@ -1,0 +1,69 @@
+import {
+  DataToHtml,
+  Pan,
+  ResetZoomButton,
+  SvgElement,
+  SvgLine,
+  SvgRect,
+  VisCanvas,
+  Zoom,
+} from '@h5web/lib';
+import type { Meta, Story } from '@storybook/react';
+import { Vector3 } from 'three';
+
+import styles from './SvgElement.stories.module.css';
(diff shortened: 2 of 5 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -28,7 +30,11 @@ import { getModifierKeyArray } from './utils';
 
 interface Props extends CommonInteractionProps {
   id?: string;
-  transform?: (rawSelection: Selection) => Selection;
+  transform?: (
+    rawSelection: Selection,
+    camera: Camera,
+    context: VisCanvasContextValue
```
axelboc, line 36: I'm now passing the camera and context to the `transform` callback so the consumer can use all the space-conversion functions without having to retrieve them manually. This is necessary for the story, since we're outside of `VisCanvas`.

apps/storybook/src/SvgElement.stories.module.css
```diff
@@ -0,0 +1,12 @@
+.rect {
+  fill: teal;
+  fill-opacity: 0.2;
+  stroke: teal;
+  stroke-width: 2;
+  pointer-events: auto;
```
loichuder, line 6: Note this blocks panning/zooming when hovering the rectangle

loichuder (approved):

apps/storybook/src/SvgElement.stories.module.css
axelboc: Yep, we'll have to find a solution for this at some point...

apps/storybook/src/SvgElement.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+          new Vector3(2, 8),
+          new Vector3(4, 6),
+          new Vector3(3, 2),
+          new Vector3(6, 4),
+          new Vector3(6, 6),
+          new Vector3(7, 7),
+        ]}
+      >
+        {(pt1, pt2, pt3, pt4, pt5, pt6) => (
+          <SvgElement>
```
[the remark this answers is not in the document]
axelboc: Nope, `DataToHtml` cannot be inside `SvgElement` since it requires access to the R3F context. `SvgElement` changes the renderer so we lose access to that context. You can have multiple `SvgElement` and wrap each of them with a `DataToHtml`, though. It's the same problem as with `Html`.

PR #1351, opened 2023-01-18
Remove `type` fields in prepublish scripts instead of CI

axelboc:
This helps with publishing H5Web's packages manually - e.g. to a local registry like Verdaccio: https://github.com/silx-kit/h5web/discussions/1334#discussioncomment-4707965.

I will publish a beta version once merged to confirm that this work as expected.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -334,6 +334,19 @@ new tag to the remote repository. This, in turn, triggers the _Release_ workflow
 on the CI, which builds and publishes the packages to NPM (with `pnpm publish`)
 and deploys the Storybook site.
 
+> A few things happen when `pnpm publish` is run inside each package's
+> directory:
+>
+> 1. First, a `prepublish` script is triggered that removes the `type` field
+>    from the package's `package.json`. The reason for this workaround is
+>    explained in [#1219](https://github.com/silx-kit/h5web/issues/1219).
+> 2. Then, pnpm modifies `package.json` further by merging in the content of the
+>    [`publishConfig` field](https://pnpm.io/package_json#publishconfig).
+> 3. Finally, the package gets published to NPM. Note that it's possible to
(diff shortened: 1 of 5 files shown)


packages/app/package.json
```diff
@@ -34,7 +34,8 @@
     "lint:eslint": "eslint \"**/*.{js,cjs,ts,tsx}\" --max-warnings=0",
     "lint:tsc": "tsc",
     "test": "jest",
-    "analyze": "pnpx source-map-explorer dist/index.js --no-border-checks"
+    "analyze": "pnpx source-map-explorer dist/index.js --no-border-checks",
+    "prepublish": "pnpx dot-json package.json -d type"
```
axelboc, line 38: Found another package than `jq` with a much simpler CLI, and one that works on Windows.

CONTRIBUTING.md
```diff
@@ -334,6 +334,11 @@ new tag to the remote repository. This, in turn, triggers the _Release_ workflow
 on the CI, which builds and publishes the packages to NPM (with `pnpm publish`)
 and deploys the Storybook site.
 
+> `pnpm publish` triggers a `prepublish` script that removes the `type` field
+> from `package.json` before publishing each package. The reason for this
+> workaround is explained in
+> [#1219](https://github.com/silx-kit/h5web/issues/1219).
```
loichuder, line 340: Since we are at it, we may as well remind that this process changes the `package.json` via `publishConfig` since this is not obvious from the get-go.

loichuder (approved):

CONTRIBUTING.md
axelboc: https://github.com/silx-kit/h5web/pull/1351/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R337-R350

PR #1352, opened 2023-01-18
Allow passing longer abscissa array that data array to `LineVis`

### LINKED ISSUE #1350: Problem with NeXus data
### Describe the Bug

NeXus data supports storing axes as either bin centers or bin boundaries, which means that an axis can have one more bin boundary than bins. When viewing this type of file, h5web will display the error `Expected abscissa array to have length N instead of N+1`

### To Reproduce

1.  Unzip the attached file [chopper.nxs.zip](https://github.com/silx-kit/h5web/files/10436650/chopper.nxs.zip)
2. Upload the file using `H5Wasm`
3. Click on `NX Line`
4. The error message should be visible

### Expected behaviour

NeXpy (https://github.com/nexpy) is able to display these files correctly.

### Screenshots

Screen Shot 2023-01-17 at 10 42 47 AM

axelboc:
Fix #1350 

I'm loosening the assertion on the length of the `abscissa` array by allowing it to be longer than the data array. If it is, it just gets sliced to the data array's length. I don't think we need to be stricter than that.

Another option was to remove the extra value from the abscissa array inside the NX containers, but it would not have solved the issue for consumers of `@h5web/lib`.

### DIFF
--- a/packages/shared/src/mock/metadata.ts
+++ b/packages/shared/src/mock/metadata.ts
@@ -151,7 +151,7 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
           signal: makeDataset('twoD_complex', complexType, [2, 2], {
             valueId: 'twoD_cplx',
           }),
-          axes: { position: makeDataset('position', intType, [2]) },
+          axes: { position: makeDataset('position', intType, [3]) },
           axesAttr: ['.', 'position'],
         }),
         makeNxDataGroup('complex_spectrum', {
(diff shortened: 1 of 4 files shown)


packages/shared/src/mock/metadata.ts
```diff
@@ -151,7 +151,7 @@ export const mockMetadata = makeNxGroup(mockFilepath, 'NXroot', {
           signal: makeDataset('twoD_complex', complexType, [2, 2], {
             valueId: 'twoD_cplx',
           }),
-          axes: { position: makeDataset('position', intType, [2]) },
+          axes: { position: makeDataset('position', intType, [3]) },
```
axelboc, line 154: I'm adding a value to the mock abscissa array of `/nexus_entry/complex` to test this use case in both `NX Heatmap` and `NX Line`. As expected, there is no visual change (except to the `/nexus_entry/complex/position` dataset itself, of course).

loichuder (approved):
Great call :+1:

PR #1357, opened 2023-01-24
Broaden arguments and props to use NumArray for #1356

### LINKED ISSUE #1356: Make `DataCurve.abscissas` an `NumArray`
I'm passing `ndarray`s about for x/y values before calling `DataCurve` and restricting `x` values to be `number[]` seems odd as `DataCurve.ordinates` is a `NumArray`.

Change `DataCurve.abscissas` to a `NumArray`.

Use `useToArray` to convert but that's implemented as a spread in `vis/utils` and does not scale well.

### Additional context
Scanning through the code, I see `lodash.toArray` is also used in `ScatterVis.tsx`. Also, `range` is used from `lodash` and `d3-array` - is this worth rationalising?

PeterC-DLS:
Fix #1356 

Avoid converting to number[] if possible, remove useToArray hook

### DIFF
--- a/packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
+++ b/packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
@@ -4,7 +4,7 @@ import { AxisBottom, AxisLeft } from '@visx/axis';
 import { scaleLinear } from '@visx/scale';
 
 import { useSafeDomain } from '../../../vis/heatmap/hooks';
-import { useToArray, useCombinedDomain, useDomain } from '../../../vis/hooks';
+import { useCombinedDomain, useDomain } from '../../../vis/hooks';
 import type { HistogramParams } from '../../../vis/models';
 import { H5WEB_SCALES } from '../../../vis/scales';
 import Tick from '../../../vis/shared/Tick';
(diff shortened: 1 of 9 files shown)


packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
```diff
@@ -67,19 +65,24 @@ function Histogram(props: Props) {
 
   const markerPositions = safeValue.map(xScale) as Domain;
 
+  const rects = [] as JSX.Element[];
```
axelboc, line 68:
```suggestion
  const rects: JSX.Element[] = [];
```

axelboc (commented):
Makes sense!

packages/lib/src/vis/utils.ts
```diff
@@ -348,7 +345,8 @@ export function getUniforms(
 }
 
 export function toArray(arr: NumArray): number[] {
-  return isTypedArray(arr) ? [...arr] : arr;
+  // eslint-disable-next-line unicorn/prefer-spread
+  return isTypedArray(arr) ? Array.from(arr) : arr;
```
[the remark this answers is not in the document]
PeterC-DLS: I thought it may be limited to a maximum number of arguments like when used in a function call (e.g. `Math.max(...arr)`) but it seems to be parsed and interpreted differently. I will revert.

packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
PeterC-DLS: Done

axelboc (approved):

PR #1359, opened 2023-01-27
Document and test `Box` utility class

axelboc:


apps/storybook/src/Utilities.stories.mdx
```diff
@@ -266,3 +266,155 @@ import { findMockEntity, getMockDataArray } from '@h5web/lib';
...
+box.hasMinSize(5, 12); // false
+box.hasMinSize(10, 12); // true
+```
+
+#### keepWithin
+
+Shift the box so it remains within the bounds of a given box.
+If the given box is too small to fit the source box, then the source box will end up overflowing the given box equally on both sides (horizontally, vertically or both).
+
+```ts
```
loichuder, line 405:
```suggestion
```ts
keepWithin(area: Box): this
```

apps/storybook/src/Utilities.stories.mdx
```diff
@@ -266,3 +266,155 @@ import { findMockEntity, getMockDataArray } from '@h5web/lib';
...
+const boundingBox = Box.empty().expandBySize(20, 20); // min: (-10, -10, 0); max: (10, 10, 0)
+
+const box = Box.empty(new Vector3(10, 10)).expandBySize(10, 10); // min: (5, 5, 0); max: (15, 15, 0)
+const shiftedBox = box.clone().keepWithin(boundingBox); // min: (0, 0, 0); max: (10, 10, 0)
+```
+
+#### toRect
+
+Convert the box to an array of two points - i.e. `[min, max]`. Note that if the box was initialised from an array of two points,
+`toRect` is unlikely to return those same two points.
```
loichuder, line 415:
```suggestion
`toRect` may not return those same two points.
```

apps/storybook/src/Utilities.stories.mdx
```diff
@@ -266,3 +266,155 @@ import { findMockEntity, getMockDataArray } from '@h5web/lib';
...
+
+```ts
```
loichuder, line 417:
```suggestion
```ts
toRect(): [Vector3, Vector3]
```

packages/lib/src/interactions/box.test.ts
```diff
@@ -0,0 +1,203 @@
...
+
+describe('size', () => {
+  it('should compute box size', () => {
+    const box1 = Box.empty();
+    expect(box1.size).toEqual({ width: 0, height: 0 });
+
+    const box2 = Box.empty(new Vector3(10, 10));
+    expect(box2.size).toEqual({ width: 0, height: 0 });
+
+    const box3 = Box.empty().expandByPoint(new Vector3(10, -5));
```
loichuder, line 15: Could we use `expandBySize` here instead of `expandByPoint` to make the test easier to read ?

packages/lib/src/interactions/box.test.ts
```diff
@@ -0,0 +1,203 @@
...
+});
+
+describe('empty', () => {
+  it('should create empty box at (0, 0, 0)', () => {
+    const box = Box.empty();
+
+    expect(box.size).toEqual({ width: 0, height: 0 });
+    expect(box.center).toMatchObject(ZERO_COORDS);
+
+    // `new Box()` initializes min and max to (+∞, +∞, +∞)` and (-∞, -∞, -∞) respectively
```
loichuder, line 40: The comment is a bit misleading here
```suggestion
    // Check that min and max are not (+∞, +∞, +∞)` and (-∞, -∞, -∞) as it would if using `new Box()` 
```

packages/lib/src/interactions/box.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    const box2 = Box.empty(new Vector3(10, 10));
+    expect(box2.center).toMatchObject({ x: 10, y: 10, z: 0 });
+
+    const box3 = Box.empty().expandByPoint(new Vector3(10, -5));
+    expect(box3.center).toMatchObject({ x: 5, y: -2.5, z: 0 });
+  });
+});
+
+describe('empty', () => {
+  it('should create empty box at (0, 0, 0)', () => {
```
loichuder, line 34:
```suggestion
  it('should create an empty box at (0, 0, 0)', () => {
```

packages/lib/src/interactions/box.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    expect(box.size).toEqual({ width: 0, height: 0 });
+    expect(box.center).toMatchObject(ZERO_COORDS);
+
+    // `new Box()` initializes min and max to (+∞, +∞, +∞)` and (-∞, -∞, -∞) respectively
+    expect(box.min).toMatchObject(ZERO_COORDS);
+    expect(box.max).toMatchObject(ZERO_COORDS);
+    expect(box.min).not.toBe(box.max); // different `Vector3` instances
+  });
+
+  it('should create empty box at given coordinates', () => {
```
loichuder, line 46:
```suggestion
  it('should create an empty box at given coordinates', () => {
```

packages/lib/src/interactions/box.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    expect(box.min).toMatchObject(expectedCoords);
+    expect(box.max).toMatchObject(expectedCoords);
+  });
+});
+
+describe('fromPoints', () => {
+  it('should create box that includes all given points', () => {
+    const box = Box.fromPoints(new Vector3(10, 10), new Vector3(-5, 12));
+    expect(box.min).toMatchObject({ x: -5, y: 10 });
+    expect(box.max).toMatchObject({ x: 10, y: 12 });
```
loichuder, line 61: You could use [`containsPoint`](https://threejs.org/docs/?q=box#api/en/math/Box3.containsPoint) to check if the points are inside the box

packages/lib/src/interactions/box.test.ts
```diff
@@ -0,0 +1,203 @@
...
+  });
+
+  it('should create empty box when called with single point', () => {
+    const vec = new Vector3(-5, 12);
```
loichuder, line 65: I think it is best continuing your pattern of declaring the `box` rather than the `vec`
```suggestion
    const box = Box.fromPoints(new Vector3(-5, 12));
```

packages/lib/src/interactions/box.ts
```diff
@@ -32,18 +32,24 @@ class Box extends Box3 {
   }
 
   public expandBySize(width: number, height: number): this {
-    return this.expandByVector(new Vector3(width / 2, height / 2, 0));
+    const { size } = this;
+
+    // Prevent shrinking to "negative" size - i.e. make sure `this.isEmpty()` is never `true`
```
loichuder, line 37: Why `isEmpty` is relevant here ?

packages/lib/src/interactions/box.ts
```diff
@@ -32,18 +32,24 @@ class Box extends Box3 {
...
+    const w = width < 0 ? Math.max(width, -size.width) : width;
```
loichuder, line 38: I wonder if we could make this condition more explicit
```suggestion
    const w = width < 0 && width < -size.width ? -size.width : width;
```

Up to you

loichuder (approved):

packages/lib/src/interactions/box.test.ts
axelboc: I try to avoid `a`, `an`, `the`, etc. in test labels to keep them short. This is the writing style that is used in most of our tests at the moment.

packages/lib/src/interactions/box.test.ts
loichuder: Sure but since you are only checking `size` here, I would find it easier to have something like:

```ts
const newSize = { width: 10, height: 20 }
const box3 = Box.empty().expandBySize(newSize);
expect(box3.size).toEqual(newSize);
```

packages/lib/src/interactions/box.ts
axelboc: Good call

axelboc: Hmm after reading it again, the duplication of `-size.width` feels a bit less clear to me for some reason.

loichuder: Ideally, the condition should make clear that the final `width` is `0` but I couldn't find an easy way to do this since `width` and `height` are treated at the same time

packages/lib/src/interactions/box.ts
axelboc: I've improved the comment, you're right :100: -- a better approach would perhaps be to fix `expandByVector`, but overkill for now.

axelboc:
Thanks a lot!

PR #1360, opened 2023-01-27
Invoke `onSelectionEnd` callback when cancelling selection

axelboc:
Fixes a minor bug in the `SelectionTool` and `AxialSelectionTool` stories. Now, when a selection is cancelled with Escape or by releasing the modifier key, the title of the canvas is reset to "No selection" as expected. This is achieved by calling `onSelectionEnd` with `undefined` when the selection is cancelled.

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -42,7 +42,7 @@ interface Props extends CommonInteractionProps {
     rawSelection: Selection,
     isValid: boolean
   ) => void;
-  onSelectionEnd?: (selection: Selection, isValid: boolean) => void;
+  onSelectionEnd?: (selection: Selection | undefined, isValid: boolean) => void;
   onValidSelection?: (selection: Selection) => void;
   children: (
     selection: Selection,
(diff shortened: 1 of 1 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -185,18 +185,19 @@ function SelectionTool(props: Props) {
...
-     * - selection was not cancelled with Escape, and
-     * - modifier key was released after pointer (if applicable). */
-    if (prevSelection && hasSuccessfullyEndedRef.current) {
+    // Previous selection was defined and current selection is now undefined => selection has ended.
+    if (prevSelection) {
       assertDefined(prevIsValid);
-      hasSuccessfullyEndedRef.current = false;
-      onSelectionEndRef.current?.(prevSelection, prevIsValid);
+      onSelectionEndRef.current?.(
+        hasSuccessfullyEndedRef.current ? prevSelection : undefined, // don't pass selection object if Escape pressed or modifier key released
```
loichuder, line 192:
```suggestion
        hasSuccessfullyEndedRef.current ? prevSelection : undefined, // pass `undefined` if Escape pressed or modifier key released
```

loichuder (approved):
Does this mean we need to add specific treatement for `undefined` in callbacks of components using `SelectionTool` ?

axelboc:
The local states in the stories were already receiving `undefined` as their default values (before any selections occurred), so no change needed, but yes, the idea is now that `onSelectionEnd` callbacks must handle `undefined` selections.

PR #1361, opened 2023-02-01
Refactor camera-based utilities and simplify `useCameraState` usage

axelboc:
Looking at `getVisibleDomains` this morning reminded me of a couple of refactorings I had been planning. Since they introduce breaking changes, I thought I'd squeeze them into the v7 release. I've made separate commits for everything, but here is a summary and some explanations:

- `getWorldFOV` becomes `getFovBox` and has to now be retrieved from the vis context. Instead of unprojecting two camera points, I now create a `Box` from the camera's position and scale like in `useMoveCameraTo` (which means I now use `getFovBox` in `useMoveCameraTo`). Since the new implementation relies on the canvas' size, it made sense to move it to the vis context and to remove the need of the second `context` parameter.
- Since I document `getFovBox` in the Storybook, I might as well also document `canvasBox` (which is like the FOV box but in HTML space ^^).
- Similarly to `getFovBox`, I move `getVisibleDomains` (which relies on `worldToData`) to the vis context and remove the second `context` parameter`.
- With those two utilities moved to the context, there's less justification for the callback of `useCameraState` to receive the `context` as second parameter, so I remove it.
- Then, I refactor `useCameraState` with `useSyncedRef`. The dependencies array is now gone, so there's even less point in passing the whole `context` object to the callback.
- Finally, I add an `equalityFn` parameter to `useCameraState` so consumers can skip renderings if they need to. I can see this being useful to optimise the `TiledHeatmapVis`, for instance.


apps/storybook/src/Context.stories.mdx
```diff
@@ -4,32 +4,35 @@ import { Meta } from '@storybook/addon-docs';
...
+| `canvasSize`        | Canvas size (equivalent to `useThree((state) => state.size)`)                                                                                                                                                                                        | <code>Size</code>                                      |
+| `canvasRatio`       | Canvas ratio (i.e. `width / height`)                                                                                                                                                                                                                 | <code>number</code>                                    |
+| `canvasBox`         | [`Box`](https://h5web-docs.panosc.eu/?path=/story/utilities--page#box) spanning the canvas in HTML space                                                                                                                                             | <code>Box</code>                                       |
```
axelboc, line 22: This is new ...

apps/storybook/src/Context.stories.mdx
```diff
@@ -4,32 +4,35 @@ import { Meta } from '@storybook/addon-docs';
...
+| `htmlToData`        | Converts a vector from HTML space to data space                                                                                                                                                                                                      | <code>(camera: Camera, vec: Vector3) => Vector3</code> |
+| `getFovBox`         | Get a [`Box`](https://h5web-docs.panosc.eu/?path=/story/utilities--page#box) spanning the field of view of the camera in world space. The second parameter, `center`, which defaults to `camera.position`, can be overridden for advanced use cases. | <code>(camera: Camera, center?: Vector3) => Box</code> |
+| `getVisibleDomains` | Get the field of view of the camera as axis domains - i.e. in data space.                                                                                                                                                                            | <code>(camera: Camera) => VisibleDomains</code>        |
```
axelboc, line 36: ... and this.

apps/storybook/src/Utilities.stories.mdx
```diff
@@ -206,53 +206,30 @@ useCanvasEvents({ onPointerDown: handlePointerDown }); // also supported: `onPoi
...
-
-const htmlPt = useCameraState(
-  (camera, context) => dataToHtml(camera, new Vector3(x, y)),
-  [x, y, dataToHtml]
-);
-```
+useCameraState<T>(
+  factory: (camera: Camera) => T,
+  equalityFn: (prev: T, next: T) => boolean
+): T
```
axelboc, line 216: No more dependencies array since I'm using `useSyncedRef` to wrap the callback (instead of `useCallback`). This simplifies things a lot for consumers since they no longer have to configure ESLint.

packages/lib/src/index.ts
```diff
@@ -132,7 +130,7 @@ export type {
 } from './interactions/models';
 export { MouseButton } from './interactions/models';
 
-export type { Domain, Dims, Axis } from '@h5web/shared';
+export type { Domain, VisibleDomains, Dims, Axis } from '@h5web/shared';
```
axelboc, line 133: I extract the return type of `getVisibleDomains` into `@h5web/shared`, next to `Domain`.

packages/lib/src/interactions/hooks.ts
```diff
@@ -20,26 +20,22 @@ const ONE_VECTOR = new Vector3(1, 1, 1);
...
   const invalidate = useThree((state) => state.invalidate);
 
   return useCallback(
     (worldPt: Vector3) => {
-      const { position, scale } = camera;
-
+      const { position } = camera;
       const visBox = Box.fromSize(visSize);
-      const fovBox = Box.empty(worldPt)
-        .expandBySize(canvasSize.width * scale.x, canvasSize.height * scale.y)
```
axelboc, line 34: This is basically the new implementation of `getFovBox`. The only difference with `getVisibleDomains` is that I need the to create a box for the "wanted" camera FOV, so I have to initialise the FOV box with a custom `center` point. That's the reason for the second, optional `center` parameter, which defaults to `camera.position`.

packages/lib/src/interactions/svg/SvgRect.tsx
```diff
@@ -12,14 +12,13 @@ function SvgRect(props: Props) {
...
 
   // Shrink/grow rectangle to simulate positioning stroke inside/outside
   // https://stackoverflow.com/questions/7241393/can-you-control-how-an-svgs-stroke-width-is-drawn
-  const offset =
-    stroke && strokePosition
-      ? (strokeWidth / 2) * (strokePosition === 'inside' ? 1 : -1)
-      : 0;
+  const offset = strokePosition
+    ? (strokeWidth / 2) * (strokePosition === 'inside' ? 1 : -1)
+    : 0;
```
axelboc, line 21: There was a regression here: since I've replaced the `stroke` prop with the `stroke` CSS property in `SelectToZoom`, the `offset` logic above was no longer working (i.e. `offset` was always `0`). As a fix, I remove `stroke` from the condition, assuming that if the consumer passes prop `strokePosition`, it's because the `SvgRect` has a stroke... This works, but a similar problem occurs if the stroke width is set via CSS... Not sure how to fix this... Reading the styles dynamically is requires rendering the component once first, which complexifies things... Instead of `strokePosition`, I could just ask for the `offset` directly, but then the user may forget to update the prop when adjusting the stroke width... Any thoughts?

packages/lib/src/vis/hooks.ts
```diff
@@ -54,23 +54,20 @@ export function useDomains(
...
-
-  // Update state when context value or dependencies change
-  useEffect(updateState, [camera, context, ...deps]); // eslint-disable-line react-hooks/exhaustive-deps
-
-  // Update state when canvas is redrawn (i.e. on zoom/pan)
-  useFrame(updateState);
+  useFrame(() => {
+    const next = factoryRef.current(camera);
+    if (!equalityFn || !equalityFn(state, next)) {
+      setState(next);
```
axelboc, line 68: In case this isn't clear in the docs, when `equalityFn` is not set, I rely on React to perform a strict equality check. The comment I had  before was wrong: `setState` always performs an equality check regardless of how you call it.

packages/lib/src/vis/hooks.ts
```diff
@@ -98,14 +95,9 @@ export function useCssColors(
...
 export function useHtmlCoords<T extends Vector3[]>(
   ...worldCoords: T
-): MappedTuple<T, Vector3> {
+): MappedTuple<T> {
   const { worldToHtml } = useVisCanvasContext();
-  return useCameraState(
-    (camera) =>
-      worldCoords.map((pt) => worldToHtml(camera, pt)) as MappedTuple<
-        T,
-        Vector3
```
axelboc, line 107: The second generic of `MappedTuple` is now optional: it assumes you want a tuple of the same type. This avoids this sort of ugly wrapping.

packages/lib/src/vis/shared/VisCanvasProvider.tsx
```diff
@@ -126,6 +129,26 @@ function VisCanvasProvider(props: PropsWithChildren<Props>) {
...
+    (camera: Camera, center: Vector3 = camera.position): Box => {
+      const { scale } = camera;
+      return Box.empty(center).expandBySize(width * scale.x, height * scale.y);
+    },
+    [width, height]
+  );
+
+  const getVisibleDomains = useCallback(
+    (camera: Camera): VisibleDomains => {
+      const [dataMin, dataMax] = getFovBox(camera).toRect().map(worldToData);
```
axelboc, line 142: I like how concise this is 😁

packages/shared/src/models-vis.ts
```diff
@@ -37,6 +42,10 @@ export interface Dims {
   cols: number;
 }
 
+// MappedTupple<number[]> => number[]
```
loichuder, line 45: Worth adding 
`// MappedTupple<string[]> => string[]`
?

packages/lib/src/interactions/svg/SvgRect.tsx
axelboc: There's indeed a CSS-only solution to the inside-stroke problem: it consists in doubling up the stroke and using `clip-path` to crop it and keep only the part inside the rectangle. However, this introduces an implicit coupling: users have to know and remember to double up the stroke width. For outside strokes, there's a solution with `mask`, but it hides the inside of the rectangle, so it's not compatible with `fill`.

It's a complex problem to solve, so I'm not fond of the idea of consumers having to deal with it themselves. Since I can't come up with a perfect solution, I say we just go with a convention: don't use a CSS property if there's an equivalent SVG attribute.

- I've now re-added `stroke &&` to the condition like it was before.
- I've put back all the props on `SvgRect` that I had moved to CSS, in `SelectToZoom` and `AxialSelectToZoom`.
- I've kept only the `transition` property in the CSS, with a single `selection` class.

PR #1362, opened 2023-02-03
Make `aspect` prop of `VisCanvas` really consistent with matplotlib

axelboc:
In #1284 I say that the new `aspect` prop of `VisCanvas` works like matplotlib's [`Axes.set_aspect`](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_aspect.html), but this was not actually true. When passing a number, this number was used as the `visRatio` directly. The beahviour is now in line with matplotlib - i.e. for `aspect={2}`, we get:

image


axelboc:
FYI, I've just realised I forgot to check the stories.

loichuder (approved):

Issue #1348, opened 2023-01-12
Search for string in object names


bmaranville:
### Is your feature request related to a problem?

HDF5 is *hierarchical*, and sometimes users are looking for a particular group or dataset by name, and don't know where to start.  Given that many NeXus files are in practice deeply nested and contain hundreds of groups and datasets, this can be quite a challenge and an annoyance for using NeXus files.
 

### Requested solution or feature

Apparently at the SNS there is a search capability in their web frontend for NeXus files, that allows finding a dataset/group by substring within a given file.  Users have requested that this functionality be implemented in H5Web as well.  In the SNS implementation, they seem to present a filtered tree which only contains paths for which the substring is matched.

#### Description of feature:
User types a string into a search box, and matched paths within the file are presented.  It could be a clickable list of matching paths, or a filtered tree, or ...

### Alternatives you've considered

At the moment, search can be clumsily achieved by using the browser's search and exploding all the paths in the HDF5 tree.  An option to explode all paths is not available.

### Additional context

Clearly implementing this without modifying the existing API of access tools like h5grove or h5wasm will be slow and inefficient, requiring recursive lookups from the frontend.  

An ancillary requirement of this feature then also be to add either a search function to the access APIs, or a function to simply return a list of all paths in the file.  Either of those can be efficiently implemented in h5py or h5wasm or ... using the underlying visititems functionality that maps directly onto the HDF5 C api.

loichuder:
For insight, this was already discussed for silx at some point: https://github.com/silx-kit/silx/issues/3575

loichuder:
Hello @bmaranville 

I am working on a prototype at https://github.com/silx-kit/h5web/pull/1358. 

Here is the search in action:
[Screencast from 08-02-2023 13:43:00.webm](https://user-images.githubusercontent.com/42204205/217533240-991f931a-b90f-4b22-b5e7-36f24fdc6875.webm)

Among the rough edges, you can see that we settled for now on **a UI with a Previous/Next button to reach matches of the search** rather than filtering paths in the explorer.

The search for matches is not by the back-end/provider but by the React component. **The only job of the provider is to return a list of paths on which the search is done**. In this case, I added an endpoint to this purpose in `h5grove` that is called by `H5GroveProvider`.

Are these decisions sensible for you ? Will they answer your need ? 

This is still in early stages so I will be happy to have your opinion to further shape the feature.

loichuder:
[...]

PR #1367, opened 2023-02-20
Add API for searchable paths

loichuder:
Pre-requisite for #1348 

Adds a `getSearchablePaths` mathods to API to return a list of paths in the group. This method can be implemented by the APIs that support it (for now, `h5grove` and `h5wasm` that was updated to 0.4.10)

### DIFF
--- a/packages/app/src/providers/DataProvider.tsx
+++ b/packages/app/src/providers/DataProvider.tsx
@@ -27,6 +27,7 @@ export interface DataContextValue {
   getExportURL?: DataProviderApi['getExportURL'];
   addProgressListener: (cb: ProgressCallback) => void;
   removeProgressListener: (cb: ProgressCallback) => void;
+  getSearchablePaths?: DataProviderApi['getSearchablePaths'];
 }
 
 const DataContext = createContext({} as DataContextValue);
@@ -113,6 +114,7 @@ function DataProvider(props: PropsWithChildren<Props>) {
         getExportURL: api.getExportURL?.bind(api),
         addProgressListener: api.addProgressListener.bind(api),
         removeProgressListener: api.removeProgressListener.bind(api),
+        getSearchablePaths: api.getSearchablePaths?.bind(api),
       }}
     >
       {children}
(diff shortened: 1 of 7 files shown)


packages/app/src/providers/DataProvider.tsx
```diff
@@ -102,6 +104,13 @@ function DataProvider(props: PropsWithChildren<Props>) {
     });
   }, [api]);
 
+  const { getSearchablePaths } = api;
+  const searchablePaths = getSearchablePaths
+    ? createFetchStore(async (path: string) =>
+        getSearchablePaths.bind(api)(path)
+      )
+    : undefined;
```
loichuder, line 112: Main discussion point here: the exposed `getSearchablePaths` could be simply `api.getSearchablePaths?.bind(api)` (as `getExportURL`).

**However** I would like the result to be cached so I used a `fetchStore` but this seems a bit overengineered to store a list of paths. Moreover, I don't expect to call `getSearchablePaths` with another `path` in arg than `/`.

axelboc: Yeah, I wouldn't bother with the fetch store, especially since we've been talking about refactoring all this fetching business.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -111,6 +111,19 @@ export class H5WasmApi extends ProviderApi {
...
   }
 
+  public async getSearchablePaths(root: string): Promise<string[]> {
+    const file = await this.file;
+
+    const h5wEntity = file.get(root);
+
+    if (isH5WasmGroup(h5wEntity)) {
+      // Build absolute paths since .paths() are relative
+      return h5wEntity.paths().map((path) => `${root}${path}`);
```
axelboc, line 121: Are we sure that we want absolute paths? Or did we discuss this already?

loichuder: > Or did we discuss this already?

Not really. But we have to somehow get the absolute path at some point if we want to reach for the entity afterwards. So it made sense for me to use absolute paths from the get-go.

axelboc: ah yeah, of course, fair point :+1:

PR #1368, opened 2023-02-21
Fix regression in `useCameraState` when dependencies change

axelboc:
I went a bit too far in #1361 by removing the `useEffect` and `deps` array parameter from `useCameraState`. This change broke two library components: `DataToHtml` and `Annotation`, as well as any use of `useCameraState` where the factory function relies on external dependencies. Without `useEffect`, `useCameraState` was no longer recomputing the state when the dependencies changed.

In this PR, I fix this regression by restoring the dependencies array parameter and recommending once again to configure linting rule `exhaustive-deps`. However, instead of restoring `useEffect`, which was triggering an extra render on every dependency change, I replace `useState` with `useRef` and update the ref via `useMemo`. This means that the state is now updated synchronously during the current render. It becomes a "computed state" as opposed to a "derived state".

To avoid future regressions, I also add a story to test that changing the props of the `Annotation` component dynamically works as expected. The screen recording below, while janky due to a low recording framerate, shows that the annotation follows the pointer pretty closely. Skipping the extra render should make a significant difference in more complex components.

Peek 2023-02-21 13-44

### DIFF
--- a/apps/storybook/src/Annotation.stories.tsx
+++ b/apps/storybook/src/Annotation.stories.tsx
@@ -1,14 +1,16 @@
-import { Annotation, DefaultInteractions, VisCanvas } from '@h5web/lib';
+import type { CanvasEvent } from '@h5web/lib';
+import { Annotation, useCanvasEvents } from '@h5web/lib';
+import { useRafState } from '@react-hookz/web';
 import type { Meta, Story } from '@storybook/react';
+import { useCallback } from 'react';
+import type { Vector3 } from 'three';
 
+import DefaultCanvas from './decorators/DefaultCanvas';
 import FillHeight from './decorators/FillHeight';
+import { formatCoord } from './utils';
 
 export const Default: Story = () => (
-  <VisCanvas
-    abscissaConfig={{ visDomain: [0, 41], showGrid: true }}
-    ordinateConfig={{ visDomain: [0, 20], showGrid: true }}
-  >
-    <DefaultInteractions />
+  <>
     <Annotation x={10} y={16}>
       HTML annotation positioned at (10, 16)
     </Annotation>
@@ -67,15 +69,11 @@ export const Default: Story = () => (
         </svg>
       </>
     </Annotation>
-  </VisCanvas>
+  </>
 );
 
 export const WithZoom: Story = () => (
-  <VisCanvas
-    abscissaConfig={{ visDomain: [0, 41], showGrid: true }}
-    ordinateConfig={{ visDomain: [0, 20], showGrid: true }}
-  >
-    <DefaultInteractions />
+  <>
     <Annotation x={10} y={16} scaleOnZoom style={{ width: 230 }}>
       HTML annotation at (10, 16) that scales with zoom.
     </Annotation>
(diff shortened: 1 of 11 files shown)


apps/storybook/src/Annotation.stories.tsx
```diff
@@ -1,14 +1,16 @@
...
+import DefaultCanvas from './decorators/DefaultCanvas';
 import FillHeight from './decorators/FillHeight';
+import { formatCoord } from './utils';
 
 export const Default: Story = () => (
-  <VisCanvas
-    abscissaConfig={{ visDomain: [0, 41], showGrid: true }}
-    ordinateConfig={{ visDomain: [0, 20], showGrid: true }}
-  >
-    <DefaultInteractions />
```
axelboc, line 11: Moving this to a decorator so that it doesn't get re-rendered when the `FollowPointer` story re-renders. This is required to test `useCameraState` properly, because when the whole canvas re-renders, it also invokes all the `useFrame`s, so it becomes impossible to isolate updates made via `useMemo`.

packages/lib/src/vis/hooks.ts
```diff
@@ -53,21 +53,34 @@ export function useDomains(
...
 
-  const factoryRef = useSyncedRef(factory);
-  const [state, setState] = useState(() => factoryRef.current(camera));
+  const stateRef = useRef<T>(); // ref instead of state to avoid re-render when deps change
+  const factoryRef = useSyncedRef(factory); // ensure `useMemo` always sees latest `factory` reference
+
+  useMemo(() => {
+    // Compute state synchronously when deps change
+    stateRef.current = factoryRef.current(camera);
+  }, deps); // eslint-disable-line react-hooks/exhaustive-deps
```
axelboc, line 68: No re-render here — the ref gets updated and returned synchronously.

Note that `useMemo` is important to avoid calling `factory` unnecessarily. Without it, we'd get: 

- re-render unrelated to camera or dependency change >> `factory` (unnecessary!)
- zoom >> `useFrame` >> `factory` >> re-render >> `factory` (again!)

packages/lib/src/vis/hooks.ts
```diff
@@ -53,21 +53,34 @@ export function useDomains(
...
-    if (!equalityFn || !equalityFn(state, next)) {
-      setState(next);
+
+    // ... unless state hasn't change
+    if (equalityFn(stateRef.current as T, next)) {
+      return;
     }
+
+    stateRef.current = next;
+    rerender();
```
axelboc, line 80: `useRerender` allows us to split the state update and the re-rendering, which is impossible to do with `useState`. This way we can update the state in `useMemo` synchronously, without re-rendering.

packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
```diff
@@ -37,11 +37,11 @@ function TiledHeatmapMesh(props: Props) {
   const { canvasSize, visSize } = useVisCanvasContext();
   const meshSize = size ?? visSize;
 
-  const groupRef = useRef<Object3D>(null);
+  const groupRef = useRef<Group>(null);
```
axelboc, line 40: Unrelated, just something I noticed along the way.

packages/lib/src/vis/hooks.ts
```diff
@@ -53,21 +53,34 @@ export function useDomains(
...
+    stateRef.current = factoryRef.current(camera);
+  }, deps); // eslint-disable-line react-hooks/exhaustive-deps
 
   useFrame(() => {
+    // Recompute state and re-render on every frame
     const next = factoryRef.current(camera);
-    if (!equalityFn || !equalityFn(state, next)) {
-      setState(next);
+
+    // ... unless state hasn't change
```
loichuder, line 74:
```suggestion
    // ... unless state hasn't changed
```

loichuder (approved):
Works for me :slightly_smiling_face:

PR #1374, opened 2023-03-02
HeatmapMaterial NaN color can be configured and now keep colors for Infinity

### LINKED ISSUE #1372: [HeatmapMesh] NaN, +inf, -inf are all displayed the same way
### Describe the bug

+inf, -inf, nan are displayed the same way from `HeatmapMesh`.

Tested with `float16`, but it's probably the same with `float32`.

### Expected behaviour

We would expect as default behaviour:

- NaN -> transparent for 
- +inf -> highest color of the lut
- -inf -> lowest color of the lut

loichuder:
Fix #1372

### DIFF
--- a/packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
@@ -1,9 +1,11 @@
 import type { Domain } from '@h5web/shared';
 import { ScaleType } from '@h5web/shared';
+import type { RGBColor } from 'd3-color';
 import { rgb } from 'd3-color';
 import type { NdArray } from 'ndarray';
 import { memo, useMemo } from 'react';
 import type { TextureFilter } from 'three';
+import { Vector4 } from 'three';
 import { DoubleSide } from 'three';
 import { DataTexture, RGBAFormat, UnsignedByteType } from 'three';
 
@@ -23,7 +25,10 @@ interface Props extends VisMeshProps {
   magFilter?: TextureFilter;
   alphaValues?: NdArray<TextureSafeTypedArray | Uint16Array>; // uint16 values are treated as half floats
   alphaDomain?: Domain;
+  badColor?: RGBColor | string;
 }
+
+const DEFAULT_BAD_COLOR = rgb(255, 255, 255, 0);
 
 const CMAP_SIZE = 256;
 
@@ -48,6 +53,7 @@ function HeatmapMaterial(props: Props) {
     magFilter,
     alphaValues,
     alphaDomain = DEFAULT_DOMAIN,
+    badColor = DEFAULT_BAD_COLOR,
   } = props;
 
   const dataTexture = useDataTexture(values, magFilter);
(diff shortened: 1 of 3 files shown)


loichuder:
Following comments on #1372, I will draft this and work on a configurable solution instead

loichuder:
Made `bad/NaN` color configurable in `HeatmapMesh`/`HeatmapMaterial`.

Infinity color is not configurable but will now use the color of the upper (resp. lower) limit of the color bar for `+Inf` (resp. `-Inf`)

@t20100 @axelboc Please have a look and see if this is all right for you

t20100:
Sounds good to me!

packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
```diff
@@ -23,8 +25,11 @@ interface Props extends VisMeshProps {
   magFilter?: TextureFilter;
   alphaValues?: NdArray<TextureSafeTypedArray | Uint16Array>; // uint16 values are treated as half floats
   alphaDomain?: Domain;
+  badColor?: RGBColor | string;
 }
 
+const DEFAULT_NAN_COLOR = rgb(255, 255, 255, 0);
```
axelboc, line 31: `DEFAULT_BAD_COLOR`?

packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
```diff
@@ -106,8 +120,8 @@ function HeatmapMaterial(props: Props) {
       uniform float alphaMin;
       uniform float oneOverAlphaRange;
       uniform int withAlpha;
+      uniform vec4 nanColor;
```
axelboc, line 123: Rename to `badColor` as well?

packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
```diff
@@ -91,6 +99,12 @@ function HeatmapMaterial(props: Props) {
       withAlpha: alphaValues ? 1 : 0,
       alphaMin: alphaDomain[0],
       oneOverAlphaRange: 1 / (alphaDomain[1] - alphaDomain[0]),
+      nanColor: new Vector4(
+        badColorAsRgb.r / 255,
+        badColorAsRgb.g / 255,
+        badColorAsRgb.b / 255,
+        badColorAsRgb.opacity ?? 1
```
axelboc, line 106: not sure it's worth falling back to `1` since `rgb()` seems to take care of initialising the opacity to 1 by default -e.g. for named colours: https://github.com/d3/d3-color/blob/958249d3a17aaff499d2a9fc9a0f7b8b8e8a47c8/src/color.js#L221

and if the user provides an `Rgb` instance directly, the opacity is initialised with `+opacity` in the constructor, which means it always ends up being a number or `NaN` (and if it's `NaN`, it's a programmer error so we shouldn't fallback silently).

axelboc (approved):

packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
loichuder: Excellent remark :+1:

PR #1375, opened 2023-03-03
Add support for unsigned 64-bit integers in h5wasm

### LINKED ISSUE #1371: Add support to view Integer (unsigned), 64-bit, little-endian

loichuder:
Fix #1371 

Oversight in the `BigInt` conversion: we only checked for regular integers and not unsigned.


axelboc (approved):
Good find :+1:

PR #1376, opened 2023-03-03
Remove background mesh from `TiledheatmapMesh`

t20100:
I need to avoid displaying the "background" color while the `TiledHeatmapMesh` is loading, i.e., not displaying this `mesh`:

https://github.com/silx-kit/h5web/blob/83c5c513ee0e6d3c6e05b29ede5353ae77f254e6/packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx#L79-L84

~~This PR proposes to add a `background` prop that allows to configure this.~~
~~I hesitated to completely remove this `mesh` and leave it to the users to implement it, but since it's there...~~

~~By default it uses the color of the smallest value of the colormap, it accepts a color as well as 'none' to disable it.~~

This PR removes the background mesh. If needed, it can be added below the `TiledHeatmapMesh`.

### DIFF
--- a/packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
+++ b/packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
@@ -2,7 +2,6 @@ import { clamp, range } from 'lodash';
 import { useRef } from 'react';
 import type { Group } from 'three';
 
-import { getInterpolator } from '../heatmap/utils';
 import { useCameraState } from '../hooks';
 import type { Size } from '../models';
 import { useVisCanvasContext } from '../shared/VisCanvasProvider';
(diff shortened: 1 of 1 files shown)


packages/lib/src/vis/tiles/TiledHeatmapMesh.tsx
```diff
@@ -72,16 +74,19 @@ function TiledHeatmapMesh(props: Props) {
...
 
   return (
     <group ref={groupRef}>
-      <mesh position={[0, 0, -0.1]}>
-        <planeGeometry args={[meshSize.width, meshSize.height]} />
-        <meshBasicMaterial
-          color={getInterpolator(colorMap, invertColorMap)(0)}
-        />
-      </mesh>
+      {backgroundColor !== 'none' && (
```
t20100, line 84: Alternative could be `null` instead of `'none'` (as in CSS) :shrug:

axelboc:
`TiledHeatmapMesh` is still experimental, so you can break things if you like. Is there a need for keeping a fallback background color when loading?

t20100:
> Is there a need for keeping a fallback background color when loading?

The main reason is that it makes a poor default: Without background, if loading tiles is fast (which is what you want ;)), it blinks a lot with `displayLowerResolutions=false` when you zoom-in and the level of details changes. Using a background color from the colormap mitigates this.

axelboc:
Up to you, but I'd be in favour of simplifying the component by removing the fallback bg, as this is a low-level component and providing loading fallbacks is typically the concern of parent components.

t20100:
OK, let's remove the background for now.

axelboc (approved):

Issue #1377, opened 2023-03-06
Customizing toolbar


PeterC-DLS:
### Is your feature request related to a problem?
We have customized some of the toolbar controls to group them together into modal dialogue boxes and move others to the overflow menu.

![colormap-modal](https://user-images.githubusercontent.com/1381719/223065038-54958546-27e1-4e30-8eb1-9374c82cd3db.png)

### Requested solution or feature
To do so, we require some minor changes to existing toolbar components
1. Allow children to be initially added to the overflow menu
2. Export some components
3. ~Allow children to included in a toggle button~

PeterC-DLS:
Thanks, Axel. This is what we have at the moment:
![colormap-tools](https://user-images.githubusercontent.com/1381719/223088531-431f0c3a-9471-4383-a8cb-4e6f6b66b2c7.png)
1. I think a static menu is fine but then in more crowded situations, a 2nd menu may be confusing.
2. Great.
3. Actually, this is not needed any longer as I moved away from re-using toggle button to the vanilla buttion for the `ColorMapOption` button in the screenshot.

axelboc:
1. Makes sense. I could see a pattern where static menus would merge into the overflow menu, as demonstrated here: https://www.telerik.com/kendo-angular-ui/components/toolbar/responsive-toolbar/ — maybe too complicated, I'll have to think about it more. Re-using the overflow menu is definitely easier for the time being.

axelboc:
Closing this as resolved now that `DomainControls` and `DomainSlider` are exported in `main` (as a reminder, the old `DomainSlider` export is now called `DomainWidget`). I won't go further on the idea of having static menus that merge into the overflow menu, at least for the time being. We can open a new issue when the time comes if the idea becomes more relevant.

I probably won't publish the next major release until September, but in the meantime I've published a beta: `v9.0.0-beta.0`. Feel free to try it out and to let me know if the new exports meet your needs. 😉

PeterC-DLS:
Thanks for the refactor. We are maintaining a fork so will have a look at updating that to reduce our divergence.

PR #1379, opened 2023-03-06
Export more toolbar control components and domain error enum

PeterC-DLS:
This is to address part of https://github.com/silx-kit/h5web/issues/1377 and allows us to reuse `BoundEditor`, `ScaledSlider` and `ColorMapOption` in our custom modal dialogue.

### DIFF
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -15,7 +15,11 @@ export { default as Btn } from './toolbar/controls/Btn';
 export { default as LinkBtn } from './toolbar/controls/LinkBtn';
 export { default as ToggleBtn } from './toolbar/controls/ToggleBtn';
 export { default as ToggleGroup } from './toolbar/controls/ToggleGroup';
+export { default as BoundEditor } from './toolbar/controls/DomainSlider/BoundEditor';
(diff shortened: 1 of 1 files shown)


axelboc:
I'm not sure I like the approach of exposing all those internal components.

It seems to me that what you need is more a component that encapsulates everything that is currently rendered inside `DomainTooltip`'s [`.tooltipControls` container](https://github.com/silx-kit/h5web/blob/86df8d5bbd443adc0b4eda4a8515324b4be76d37/packages/lib/src/toolbar/controls/DomainSlider/DomainTooltip.tsx#L74), doesn't it?

Exposing `ScaledSlider` is also problematic, as you probably have to replicate a whole bunch of logic from `DomainSlider` to make it work the same way as in H5Web.

Instead, I see no problem in refactoring the code to properly expose a `DomainSlider` and a `DomainControls` component that are both fully decoupled from the tooltip UI.

axelboc:
Closing in favour of https://github.com/silx-kit/h5web/issues/1382

PeterC-DLS:
No problem, as it's what we replicated with https://github.com/DiamondLightSource/davidia/blob/main/client/component/DomainControls.tsx

PR #1380, opened 2023-03-06
Allow other elements to a toggle button

PeterC-DLS:
This is to address part of https://github.com/silx-kit/h5web/issues/1377 and allows us to use something other than text or icon in a toggle button


PeterC-DLS:
Not needed

Issue #1181, opened 2022-08-01
Possible "ignore" netCDF-4 fill value/missing values?


mathomp4:
### Is your feature request related to a problem?

So, I've been having fun with the the fun h5web vscode extension with my netCDF files, but today I encountered something that hadn't yet: missing values (aka fill values).

I made a quick little netCDF-Python script to make up a variable where the fill value is 1e15:
```bash
❯ ncdump -hsc test_miss.nc4
netcdf test_miss {
dimensions:
	nlocs = 1000 ;
variables:
	float nlocs(nlocs) ;
		nlocs:_Storage = "contiguous" ;
		nlocs:_Endianness = "little" ;
	float value(nlocs) ;
		value:_FillValue = 1.e+15f ;
		value:_Storage = "contiguous" ;
		value:_Endianness = "little" ;

// global attributes:
		:_NCProperties = "version=2,netcdf=4.8.1,hdf5=1.12.1" ;
		:_SuperblockVersion = 2 ;
		:_IsNetcdf4 = 1 ;
		:_Format = "netCDF-4" ;
}
```
but in this script, I didn't fill every point in the `value` variable, just some of them (randomly). When you do this, the output from `ncdump` looks like:

```bash
❯ ncdump -v value test_miss.nc4 | head -12
netcdf test_miss {
dimensions:
	nlocs = 1000 ;
variables:
	float nlocs(nlocs) ;
	float value(nlocs) ;
		value:_FillValue = 1.e+15f ;
data:

 value = 8.252128, _, 3.278462, 1.717357, 3.64509, 5.895305, 6.32355,
    9.360811, _, _, _, 1.007396, 4.095286, _, _, _, _, _, _, _, _, 9.35124,
    _, _, _, _, _, _, _, _, _, _, 6.430166, _, 5.538579, 4.860384, _,
```

`ncdump` sees we have a `_FillValue` set, and shows them as "missing".

But since the h5web extension doesn't know that (since I'm guessing netCDF-4 doesn't define the fill value as an HDF5 fill value...if that exists?), then you get this when you view it:

<img width="864" alt="Screen Shot 2022-08-01 at 1 19 56 PM" src="https://user-images.githubusercontent.com/2982494/182206349-c765ec83-58c2-4a2c-9199-61393eba682d.png">

Now, if you turn on points and use a log scale for y:

[...]

PR #1381, opened 2023-03-09
Ignore `_FillValue` in `LineVis`

loichuder:
For #1181 

I think it is best to look at each commit separately.


packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -160,3 +160,19 @@ export function useSlicedDimsAndMapping(
...
+    return undefined;
+  }
+
+  const fillValue = fillValueArray[0];
+  if (typeof fillValue !== 'number') {
+    return undefined;
+  }
+
+  return fillValue;
+}
```
loichuder, line 178: This hook may have to be changed. In the file supplied in #1381, `_FillValue` is a array of one value but [the specs](https://docs.unidata.ucar.edu/netcdf-c/current/attribute_conventions.html) tell that:

> If _FillValue is defined then it should be scalar and of the same type as the variable.

So this array parsing may not be needed

packages/app/src/vis-packs/core/line/LineVisContainer.tsx
```diff
@@ -53,6 +56,7 @@ function LineVisContainer(props: VisContainerProps) {
               title={entity.name}
               toolbarContainer={toolbarContainer}
               config={config}
+              invalidValue={fillValue}
```
loichuder, line 59: I have then to do a bit a prop drilling to use this `fillValue` in the domain computation of `MappedLineVis` and in the generation of points in the `DataCurve`.

packages/shared/src/utils.ts
```diff
@@ -182,3 +184,11 @@ export function getDims(dataArray: NdArray): Dims {
   const [rows, cols] = dataArray.shape;
   return { rows, cols };
 }
+
+export function isValid(value: number, invalidValue?: number) {
+  if (invalidValue === undefined) {
+    return true;
+  }
+
+  return Math.abs((value - invalidValue) / invalidValue) > 1e-2;
```
loichuder, line 193: Usual problem when comparing floats, the equality check with `===` may not be satisfied due to a lack of precision.

So instead, I consider the value invalid if it lies in `+/- 1%` of `invalidValue`. This criteria is arbitrary and may be revisited in the future.

packages/app/src/vis-packs/core/hooks.ts
t20100: Then it's probably worth also supporting `_FillValue` being a `number`.

packages/shared/src/utils.ts
t20100: Why is this an issue?
`number` is supposed to be float64, and only `_FillValue` should be considered invalid, right?

t20100: My bad, netCDF specs defines the comparison to do:

 > If the _FillValue is positive then it defines a valid maximum, otherwise it defines a valid minimum. For integer types, there should be a difference of 1 between the _FillValue and this valid minimum or maximum. For floating point types, the difference should be twice the minimum possible (1 in the least significant bit) to allow for rounding error.

t20100:
Having a quick look at [the specs](https://docs.unidata.ucar.edu/netcdf-c/current/attribute_conventions.html) you mentioned, having `_FillValue` defining an invalid value is a special case (maybe a common one though) since netCDF has other attributes to control the `valid_range`:

```
‍If neither valid_min, valid_max nor valid_range is defined then generic applications should define a valid range as follows. If the data type is byte and _FillValue is not explicitly defined, then the valid range should include all possible values. Otherwise, the valid range should exclude the _FillValue (whether defined explicitly or by default) as follows. If the _FillValue is positive then it defines a valid maximum, otherwise it defines a valid minimum. For integer types, there should be a difference of 1 between the _FillValue and this valid minimum or maximum. For floating point types, the difference should be twice the minimum possible (1 in the least significant bit) to allow for rounding error. 
``` 

I don't know how much of this is really needed though.

packages/shared/src/utils.ts
axelboc: I'm surprised that this is needed as well. With binary transfer, I don't get how we can lose precision. Perhaps there is a problem in the example file, or does this only happen with mock data?

There are lots of [edge cases](https://floating-point-gui.de/errors/comparison/) to comparing floats, unfortunately... [This package](https://github.com/micksatana/nearly-equal/blob/master/src/nearly-equal.ts) implements the `nearlyEqual` function from the article, but it's no longer maintained (though we could duplicate the function).

[This package](https://www.npmjs.com/package/float-equal) seems promising as well, as it is 1) making use of `Number.EPSILON`; 2) avoiding divisions; 3) adjusting the epsilon as a last resort: https://github.com/sindresorhus/float-equal/blob/main/index.js

If we go down the float comparison route, we definitely need to invoke the comparison code only when necessary to avoid computing something like `Number.EPSILON * Math.min(Math.abs(a), Math.abs(b))` on integers, for instance... 😅 

**EDIT**: Disregard the above since NetCDF specifies a comparison algorithm. It is all the more important that the comparison function be provided by the code that knows about NetCDF4, i.e. `@h5web/app`.

packages/app/src/vis-packs/core/hooks.ts
axelboc: Yeah, might as well support both... Maybe it was allowed by an old version of the spec.

axelboc (changes requested):

packages/shared/src/utils.ts
loichuder: After checking, it seems the loss of precision happens **in the binary transfer**.

By using JSON to get data, I get `1000000000000000` for the values in the array as expected.

But with the binary transfer, I get `999999986991104` instead :shrug:

loichuder: It is due to the fact that the data is stored in `float32` that is not able to store `1e15` in an accurate way. 

By using the JSON transfer, we use JS `Number` that is `float64` and therefore able to store `1e15` accurately.

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -160,3 +161,34 @@ export function useSlicedDimsAndMapping(
...
+
+  if (!Array.isArray(rawFillValue)) {
+    return undefined;
+  }
+  const fillValue = rawFillValue[0];
+
+  return typeof fillValue === 'number' ? fillValue : undefined;
+}
+
+export function useIgnoreValue(
```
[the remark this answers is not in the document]
loichuder: > this is like `getModifierKeyArray` so it might be worth abstracting this into a generic `wrap` utillity of sorts :shrug: 

Or use [castArray](https://lodash.com/docs/4.17.15#castArray) ?

>  const fillValue = typeof wrappedFillValue[0] === 'number' ? new Float32Array(wrappedFillValue)[0] : undefined;

Not sure if you meant that `Float32Array` should be used everywhere but I think it is needed to derive the `TypedArray` from the dataset `type`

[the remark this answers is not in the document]
axelboc: ah yeah, perfect!

packages/app/src/providers/h5grove/utils.ts
```diff
@@ -45,54 +40,6 @@ export function isExternalLinkResponse(
...
-        return undefined;
-      case 32:
-        return Float32Array;
-      case 64:
-        return Float64Array;
-    }
-  }
-
-  return undefined;
-}
```
loichuder, line 94: Moved a level above

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -160,3 +162,26 @@ export function useSlicedDimsAndMapping(
...
+
+  const rawFillValue = attrValuesStore.getSingle(dataset, '_FillValue');
+  const wrappedFillValue = castArray(rawFillValue);
+
+  const DTypedArray = typedArrayFromDType(dataset.type);
+
+  // Cast fillValue in the type of the dataset values to be able to use `===` for the comparison
+  const fillValue =
+    DTypedArray && typeof wrappedFillValue[0] === 'number'
+      ? new DTypedArray(wrappedFillValue as number[])[0]
```
axelboc, line 179: Arf hadn't thought that wrapping as array instead of unwrapping would lead to requiring this type cast. Oh well, it can stay, too bad.

axelboc (approved):

PR #1383, opened 2023-03-10
Put `App` root into fullscreen instead of document

### LINKED ISSUE #1365: Full screen mode not working as expected inside JupyterLab
### Describe the bug

The full screen button puts the entire JupyterLab UI into full screen mode rather than just H5Web. This is probably the case in other applications that embed `@h5web/app`.

### To Reproduce

1. Run JupyterLab with [`jupyterlab-h5web`](https://github.com/silx-kit/jupyterlab-h5web) extension installed
2. Open any HDF5 file
3. Click on full screen button in top bar

### Expected behaviour

The H5Web cell should take up the whole screen.

### Context

- OS: Ubuntu
- Browser: Firefox
- H5Web context: jupyterlab_h5web

axelboc:
Fix #1365 

This fixes fullscreen mode in JupyterLab and other integrations, where H5Web is part of a broader application.


packages/app/src/App.tsx
```diff
@@ -55,6 +55,7 @@ function App(props: Props) {
     >
       <ReflexContainer
         className={styles.root}
+        data-fullscreen-root
```
axelboc, line 58: Easier than a ref, since I can query the element when the user clicks on the fullscreen button.

packages/app/src/breadcrumbs/BreadcrumbsBar.tsx
```diff
@@ -34,13 +34,10 @@ function BreadcrumbsBar(props: Props) {
...
+  const isFullscreen = !!document.fullscreenElement;
 
-  const [isFullscreen, toggleFullScreen] = useToggle(
-    !!document.fullscreenElement
-  );
-  useEventListener(document.documentElement, 'fullscreenchange', () => {
-    toggleFullScreen();
-  });
+  const rerender = useRerender();
+  useEventListener(document, 'fullscreenchange', rerender);
```
axelboc, line 40: Simplifying the logic here so that I don't have to add a listener to the element that enters fullscreen (which would require a ref). Note that the `fullscreenchange` event bubbles up to the document.

packages/app/src/breadcrumbs/BreadcrumbsBar.tsx
```diff
@@ -79,8 +76,10 @@ function BreadcrumbsBar(props: Props) {
           label="Go full screen"
           onClick={() => {
             if (!document.fullscreenElement) {
-              void document.documentElement.requestFullscreen();
-            } else if (document.exitFullscreen) {
```
axelboc, line 83: Not sure why I was checking for the existence of `exitFullscreen` here. It is well supported.

axelboc:
I've put the `data-fullscreen-root` attribute on the `ReflexContainer`, which is inside the `App`'s top-level error boundary. It is worth noting that if the error boundary is triggered while the app is in fullscreen, the browser automatically leaves fullscreen mode (because the `ReflexContainer` is removed from the DOM). I think this is better behaviour than to add a wrapper around `ReflexContainer` and to see uncaught errors in fullscreen.

packages/app/src/breadcrumbs/BreadcrumbsBar.tsx
loichuder: Actually, perhaps a ref would reduce the implicit coupling introduced by the `data-fullscreen-root` field.

You could use the ref to both attach the listener to the relevant fullscreen element and to replace the `querySelector` below.

loichuder (approved):
I had a suggestion to use the ref instead but I don't mind the actual implementation.

PR #1384, opened 2023-03-10
Allow controlling min select-to-zoom sizes

axelboc:
I'm adding a prop called `minZoom` to both `SelectToZoom` and `AxialSelectToZoom` to allow changing the default min zoom size of 20px.

### DIFF
--- a/apps/storybook/src/AxialSelectToZoom.stories.tsx
+++ b/apps/storybook/src/AxialSelectToZoom.stories.tsx
@@ -62,6 +62,9 @@ MultipleModifierKeysY.args = {
   modifierKey: ['Control', 'Shift'],
 };
 
+export const MinZoom = Template.bind({});
+MinZoom.args = { minZoom: 200 };
+
 export const Disabled = Template.bind({});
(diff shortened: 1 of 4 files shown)


loichuder (commented):
I understand the rationale behind the default minimal value when it was not configurable. 

But now, I think it would make sense to **not** set a default value for `minZoom` and have no check when it is not supplied. It would be up to consumers such as `@h5web/app` to set this minimal value themselves.

What do you think ?

axelboc:
While I agree with your sentiment, having "no check when it is not supplied" isn't great as the bug of dragging while clicking is subtle and might not be noticed by developers. `minZoom` has to be provided somehow, so it can either be mandatory or have a default value of at least a few pixels.

I would personally prefer to keep a sensible default on the low-level components, perhaps less opinionated than 20px, like 5 or 10? 

If you prefer, I can make `minZoom` required on the low-level components, but then I'd still need a default for `DefaultInteractions`.

loichuder (approved):

PR #1387, opened 2023-03-14
Bring stacking under control

### LINKED ISSUE #1373: Not able to display a Tooltip over an Annotation
### Describe the bug

Screenshot from 2023-03-01 17-13-59

The red cross is an `Annotation` and the lines are something basic i guess.

I am not able to display a `TooltipMesh` over that.

I dont understand the behaviour, there is no orders, it sounds like a layer is selected by kind.

### Expected behaviour

I expect last elements to be displayed over the previous elements.

axelboc:
Fix #1373 

I'm introducing a bunch of CSS variables to control the `z-index` stacking of the various elements that appear on top of the WebGL canvas. The idea is that, by default, anything rendered via `Html` appears underneath the tooltip and floating toolbar, and in front the `svg` overlay and axis grid.

I'm also documenting how the stacking works, how to avoid the stacking from changing when stuff unmount/remount, and how to go about tweaking the stacking if the need arises.


loichuder (approved):
Few suggestions to improve the readability but overall great write-up :+1:

apps/storybook/src/Stacking.stories.mdx
```diff
@@ -0,0 +1,123 @@
...
+    Foo
+  </Annotation>
+
+  <SvgElement>
+    <SvgRect fill={bool ? 'skyblue' : 'transparent'} />
+  </SvgElement>
+</>
+```
+
+### What if DOM order cannot be changed or guaranteed?
```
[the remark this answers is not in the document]
axelboc: I've reworded the first sentence of this section to:

> If you can't restructure your React tree to stack HTML and SVG elements in the right order,
or if you can't avoid unmounting/remounting some occurrences of `Html` and `SvgElement`,
the last resort is to control the stacking order explicitly with `z-index`.

apps/storybook/src/Stacking.stories.mdx
```diff
@@ -0,0 +1,123 @@
...
+that this entails:
+
+```tsx
+<Html>
+  <svg style={{ zIndex: 1 }}>{/* ... */}</svg>
+</Html>
+```
+
+### Crossing stacking boundaries
+
```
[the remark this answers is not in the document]
axelboc: I've reworded the section.

apps/storybook/src/Stacking.stories.mdx
```diff
@@ -0,0 +1,125 @@
...
+  <SvgElement>
+    <SvgRect fill={bool ? 'skyblue' : 'transparent'} />
+  </SvgElement>
+</>
+```
+
+### What if DOM order of HTML elements cannot be changed or guaranteed?
+
+If you can't restructure your React tree to render `Html` elements in the order you want them to be stacked,
```
loichuder, line 84:
```suggestion
If you can't restructure your React tree to render `Html` elements in an order following the desired stacking,
```

apps/storybook/src/Stacking.stories.mdx
```diff
@@ -0,0 +1,125 @@
...
+or if you can't avoid unmounting/remounting them, the last resort is to control the stacking order explicitly with `z-index`.
+
+Knowing that, by default, `Html`'s internal `div` element has no `z-index`,
+which is equivalent to `z-index: 0`, you can move an `Html` element in front of,
+or behind, other `Html` elements simply by giving it a positive or negative
+`z-index` (respectively).
```
loichuder, line 90: Lots of info in this: I suggest to break it down in two sentences.

```suggestion
By default, `Html`'s internal `div` element has no `z-index`,
this is equivalent to `z-index: 0`. 
You can therefore move an `Html` element in front of,
or behind, other `Html` elements simply by giving it a positive or negative
`z-index` (respectively).
```

apps/storybook/src/Stacking.stories.mdx
```diff
@@ -0,0 +1,125 @@
...
+</>
+```
+
+> To avoid messing with the stacking order of other components (grid, tooltip, etc.), please remain within the range [-999, 999].
+
+### What if DOM order of SVG elements cannot be changed or guaranteed?
+
+For `SvgElement`, the problem is more complicated as SVG 1.1 does not provide an equivalent to `z-index`.
+A workaround is to not use `SvgElement` at all and instead render an `svg` element yourself via `Html`, with the caveats
+that this entails (i.e. you have to position and size the SVG element yourself):
```
loichuder, line 108: :+1:

apps/storybook/src/Stacking.stories.mdx
axelboc: Good call 👍

apps/storybook/src/Stacking.stories.mdx
```diff
@@ -0,0 +1,123 @@
...
+The closest container to all these elements, `canvasWrapper`, creates a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context).
+This allows its children (both direct and nested) to be stacked in relation with one another within this stacking context (unless some of them define their own stacking context).
+
+By default, the stacking order is as follows, from furthest (behind) to closest (in front):
+
+1. axis system (including the axis grid);
+2. SVG elements rendered via [`SvgElement`](http://localhost:6006/?path=/story/building-blocks-svgelement--default) into the `svg` overlay, in DOM order;
+3. HTML elements rendered via [`Html`](http://localhost:6006/?path=/story/building-blocks-html--default), in DOM order (including [`Annotation`](http://localhost:6006/?path=/story/building-blocks-annotation--default) and [`Overlay`](http://localhost:6006/?path=/story/building-blocks-overlay--default));
+4. tooltip
+5. floating toolbar (typically with reset zoom button)
```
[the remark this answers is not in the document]
axelboc: I've added a schema and emphasised/reworded a few things (in a separate commit to ease your review).

apps/storybook/src/Stacking.stories.mdx
```diff
@@ -4,33 +4,54 @@ import { Meta } from '@storybook/addon-docs';
...
+│   │ 2. SVG elements                  │
+└── ┤   ┌──────────────────────────────┴───┐
+    │   │ 3. HTML elements                 │
+    └── ┤   ┌──────────────────────────────┴───┐
+        │   │ 4. Tooltip                       │
+        └── ┤   ┌──────────────────────────────┴───┐
+            │   │ 5. Floating toolbar              │
+            └── ┤                                  │
+                │                                  │
+                └──────────────────────────────────┘
```
loichuder, line 36: That's so nice :joy:

loichuder (approved):

apps/storybook/src/Stacking.stories.mdx
axelboc: Took me age 😂

PR #1390, opened 2023-03-17
Refactor and optimise creation of `ScatterPoints` geometry

axelboc:
As part of the investigative work around thick/dashed lines, I'm trying to come up with an efficient pattern for instantiating and updating Three.js geometries.

---

### Why is this needed?

This comes from the observation that Three's `BufferGeometry`, notably with its `setFromPoints` method, promotes highly inefficient patterns for instantiating and updating geometries.

[`setFromPoints`](https://github.com/mrdoob/three.js/blob/master/src/core/BufferGeometry.js#L273) creates a `position` buffer attribute from an array of vectors: `[Vector3(1, 2, 3), Vector3(4, 5, 6)]` becomes `Float32BufferAttribute([1, 2, 3, 4, 5, 6])`.

I'll pass quickly on the fact that, internally, `setFromPoints` pushes values to a plain JS array, which is not as efficient as initialising a `Float32Array` with the right size. I think it's because it accepts both `Vector3` and `Vector2` in the input array, so it can't tell ahead of time what the size of the final buffer will be.

The bigger problem is that `setFromPoints` requires preparing an input array that contains exactly the coordinates we want to end up with in the `position` attribute. In other words, we typically end up with two loops: one that loops through the data and creates one or more position vectors for each data point, and another one inside `setFromPoints` to flatten those vectors.

This gets even worse when initialising multiple geometries from the same data array - e.g. in `DataCurve`, we need a geometry for the lines, and potentially two more for the error bars, so we can end up calling `setFromPoints` three times!

While digging into the code of `BufferGeometry`, I found more inefficiencies:

- [`BufferGeometry#setIndex`](https://github.com/mrdoob/three.js/blob/master/src/core/BufferGeometry.js#L64), which we use in `SurfaceMesh`, calls [`arrayNeedsUint32`](https://github.com/mrdoob/three.js/blob/master/src/utils.js#L33-L45), which loops through the entire indices array for the sole purpose of deciding whether to use a `Uint32Array` or a `Uint16Array`, which utterly defeats the purpose of trying to optimise the type of `TypedArray` used for the index...
- In `ScatterPoints`, before this PR, we used to initialise a "normalized" `uint8` buffer attribute. The purpose of `normalized` is to automatically convert the integer values that we set in the buffer (in `[0 255]`) to floats (in `[0 1]`). In other words, at some point, the `Uint8Array` turns into a `Float32Array` (couldn't find exactly where), which is wasteful in terms of memory.

In addition to these inefficiencies, Three's `BufferGeometry` does not help with code readability. It can hide some of the complexity of creating the `position` buffer attribute via `setFromPoints` for basic objects (e.g. line, points), but as soon as custom shaders are involved (e.g. scatter points, thick/dashed lines, etc.), that's no longer enough. We end up with code that uses `setFromPoints` mingled with code that creates custom buffer attributes.

### My overall vision

I'd like to find a pattern that allows us to loop through the data once and only once, regardless of how many geometries we need to initialise/update and regardless of how many attributes and other features (e.g. index) those geometries have.

---

Before I tackle `DataCurve`, I thought I'd start with a simpler component with a single geometry. `ScatterPoints` was a good candidate because it doesn't make use of `setFromPoints`, so it was already pretty inefficient and only looping through the data once.

All I do in this PR is just optimise a couple of things and structure the code in a way that can be more easily replicated in other components.


packages/lib/src/vis/scatter/ScatterPoints.tsx
```diff
@@ -71,28 +72,46 @@ function ScatterPoints(props: Props) {
...
-  const { position, color } = useBufferAttributes(
-    abscissas,
-    ordinates,
-    data,
-    dataToColorScale
-  );
+  const dataGeometry = useMemo(() => {
+    const geometry = new BufferGeometry();
+    geometry.setAttribute('position', createBufferAttr(length, 3));
+    geometry.setAttribute('color', createBufferAttr(length, 3));
```
axelboc, line 96: We now initialise the buffer attributes and their internal typed arrays only once, and re-initialise them only when `data.length` changes. As long as the number of values doesn't change, we can reuse the same instances.

Note that we were already initialising the typed arrays to the exact lengths we need, so that hasn't changed. But that's not what we do in some other places, so `createBufferAttr` will prove useful I think.

Note also that I'm now using a `Float32BufferAttribute` even for `color` instead of a "normalized" `Uint8BufferAttribute`, for the reason explained in the PR description.

packages/lib/src/vis/scatter/ScatterPoints.tsx
```diff
@@ -71,28 +72,46 @@ function ScatterPoints(props: Props) {
...
+    return geometry;
+  }, [length]);
 
   useLayoutEffect(() => {
```
axelboc, line 100: I'm now looping through the `data` array inside the `useLayoutEffect` instead of in a `useMemo` beforehand. This allows me to mutate the buffers instead of recreating instances.

packages/lib/src/vis/scatter/ScatterPoints.tsx
```diff
@@ -71,28 +72,46 @@ function ScatterPoints(props: Props) {
...
+    data.forEach((val, index) => {
+      const { x, y, z } = indexToPosition(index);
+      position.setXYZ(index, x, y, z);
+
+      const { r, g, b } = valueToColor(val);
+      color.setXYZ(index, r / 255, g / 255, b / 255); // normalize RGB channels
+    });
+
+    color.needsUpdate = true;
+    position.needsUpdate = true;
```
axelboc, line 112: Since the instances are re-used we need to mark them as updated.

packages/lib/src/vis/scatter/ScatterPoints.tsx
```diff
@@ -71,28 +72,46 @@ function ScatterPoints(props: Props) {
...
-    dataGeometry.setAttribute('position', new BufferAttribute(position, 3));
-    dataGeometry.setAttribute('color', new BufferAttribute(color, 3, true));
+    const { color, position } = dataGeometry.attributes;
+
+    data.forEach((val, index) => {
+      const { x, y, z } = indexToPosition(index);
+      position.setXYZ(index, x, y, z);
+
+      const { r, g, b } = valueToColor(val);
+      color.setXYZ(index, r / 255, g / 255, b / 255); // normalize RGB channels
```
axelboc, line 108: 1. Here, I'm doing the normalisation myself to avoid the `Uint8Array` to `Float32Array` conversion, and also to be more explicit (the "normalized" feature is enabled by passing `true` as third argument to `new BufferAttribute()`, which I find very obscure).
2. `valueToColor` now returns an `RgbColor` directly instead of an array of three numbers. This saves us a spread or a loop through those three values (when calling `BufferAttribute#set`).

packages/lib/src/vis/scatter/hooks.ts
```diff
@@ -1,66 +1,6 @@
...
-
-      const x = abscissaScale(abscissas[index]);
-      const y = ordinateScale(ordinates[index]);
-
-      const hasFiniteCoords = Number.isFinite(x) && Number.isFinite(y);
-
-      /* Render points with NaN/Infinity coordinates (i.e. values <= 0 in log)
-       * at origin to avoid Three warning, and outside of camera's field of view
-       * to hide them and any segments connecting them. */
-      position.set([x, y, hasFiniteCoords ? 0 : CAMERA_FAR], 3 * index);
```
axelboc, line 35: There was a bug here: we weren't doing the "at origin" part like in `DataCurve`.

packages/lib/src/vis/scatter/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+): (index: number) => { x: number; y: number; z: number } {
+  return (index: number) => {
+    const x = abscissaScale(abscissas[index]);
+    const y = ordinateScale(ordinates[index]);
+
+    /* Render points with NaN/Infinity coordinates (i.e. values <= 0 in log)
+     * at origin to avoid Three warning, and outside of camera's field of view
+     * to hide them and any segments connecting them. */
+    const hasFiniteCoords = Number.isFinite(x) && Number.isFinite(y);
+    return hasFiniteCoords ? { x, y, z: 0 } : { x: 0, y: 0, z: CAMERA_FAR };
```
axelboc, line 26: The fallback case now renders the point at `(0, 0, ...)` as expected.

packages/lib/src/vis/scatter/hooks.ts
axelboc: Ah, because the comment seems to indicate that it's the moving of the point to the origin that removes the warning, (and that moving it behind the camera is just for hiding the resulting line segments). I think this is right, since the coordinates in the `position` attribute are passed directly to the vertex shader, and it seems logical to me that the shader wouldn't like dealing with infinite coordinates.

loichuder: Ha! Yes, you are right ! :+1:

PR #1391, opened 2023-03-21
Add search feature in the explorer

### LINKED ISSUE #1348: Search for string in object names

loichuder:
Fix #1348 

[Screencast from 21-03-2023 17:00:24.webm](https://user-images.githubusercontent.com/42204205/226667755-a74d3820-f15a-4e3a-9694-3536218cdf89.webm)

### DIFF
--- a/apps/demo/src/styles.css
+++ b/apps/demo/src/styles.css
@@ -17,6 +17,7 @@ body {
 
 body {
   --primary: #c0da74;
+  --primary-lighter: #eaf0cd;
   --primary-light: #d4e09b;
   --primary-dark: #9aae5d;
   --primary-bg: #f5fbef;
--- a/packages/app/src/App.tsx
+++ b/packages/app/src/App.tsx
@@ -8,10 +8,10 @@ import { ReflexContainer, ReflexElement, ReflexSplitter } from 'react-reflex';
 import styles from './App.module.css';
 import EntityLoader from './EntityLoader';
 import ErrorFallback from './ErrorFallback';
+import Sidebar from './Sidebar';
 import VisConfigProvider from './VisConfigProvider';
 import BreadcrumbsBar from './breadcrumbs/BreadcrumbsBar';
 import type { FeedbackContext } from './breadcrumbs/models';
-import Explorer from './explorer/Explorer';
 import MetadataViewer from './metadata-viewer/MetadataViewer';
 import { useDataContext } from './providers/DataProvider';
 import Visualizer from './visualizer/Visualizer';
@@ -60,12 +60,12 @@ function App(props: Props) {
         orientation="vertical"
       >
         <ReflexElement
-          className={styles.explorer}
+          className={styles.sidebar}
           style={{ display: isExplorerOpen ? undefined : 'none' }}
           flex={25}
           minSize={150}
         >
-          <Explorer selectedPath={selectedPath} onSelect={onSelectPath} />
+          <Sidebar selectedPath={selectedPath} onSelect={onSelectPath} />
         </ReflexElement>
 
         <ReflexSplitter
--- a/packages/app/src/explorer/EntityItem.tsx
+++ b/packages/app/src/explorer/EntityItem.tsx
@@ -2,6 +2,7 @@ import { isGroup } from '@h5web/shared';
 import type { ChildEntity } from '@h5web/shared';
 import { useToggle } from '@react-hookz/web';
 import type { CSSProperties, KeyboardEvent } from 'react';
+import { useRef } from 'react';
 import { useCallback } from 'react';
 import { Suspense, useEffect } from 'react';
 import { ErrorBoundary } from 'react-error-boundary';
@@ -31,6 +32,8 @@ function EntityItem(props: Props) {
   const { path, entity, level, selectedPath, onSelect } = props;
   const isSelected = path === selectedPath;
 
+  const btnRef = useRef<HTMLButtonElement>(null);
+
   const [isExpanded, toggleExpanded] = useToggle(false);
 
   useEffect(() => {
@@ -42,6 +45,13 @@ function EntityItem(props: Props) {
       toggleExpanded(true);
     }
   }, [entity, path, selectedPath, toggleExpanded]);
+
+  // When tabbing in, restore focus on the selected element
+  useEffect(() => {
+    if (isSelected) {
+      btnRef.current?.focus();
+    }
+  }, [btnRef, isSelected]);
 
   const handleKeyDown = useCallback(
     (e: KeyboardEvent<HTMLButtonElement>) => {
@@ -93,6 +103,7 @@ function EntityItem(props: Props) {
       role="none"
     >
       <button
+        ref={btnRef}
         className={styles.btn}
         type="button"
         role="treeitem"
--- a/packages/app/src/explorer/Explorer.tsx
+++ b/packages/app/src/explorer/Explorer.tsx
@@ -18,9 +18,14 @@ function Explorer(props: Props) {
   const ref = useRef<HTMLDivElement>(null);
   const btnRef = useRef<HTMLButtonElement>(null);
 
+  const isRootSelected = selectedPath === '/';
+
+  // When tabbing in, restore focus on the root button if it was selected
   useEffect(() => {
-    btnRef.current?.focus();
-  }, [btnRef]);
+    if (isRootSelected) {
+      btnRef.current?.focus();
+    }
+  }, [btnRef, isRootSelected]);
 
   return (
     <div ref={ref} className={styles.explorer} role="tree" id={EXPLORER_ID}>
@@ -29,7 +34,7 @@ function Explorer(props: Props) {
         className={styles.fileBtn}
         type="button"
         role="treeitem"
-        aria-selected={selectedPath === '/'}
+        aria-selected={isRootSelected}
         data-path="/"
         onClick={() => onSelect('/')}
         onKeyDown={(e) => {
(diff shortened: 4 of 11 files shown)


apps/demo/src/styles.css
```diff
@@ -17,6 +17,7 @@ body {
 
 body {
   --primary: #c0da74;
+  --primary-lighter: #eaf0cd;
```
loichuder, line 20: New shade of `primary` for the new bar with the explorer/search icons

packages/app/src/ExplorerTabs.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const { selectedPath, onSelect } = props;
+
+  const [tab, setTab] = useState<Tab>(Tab.Explore);
+
+  const [searchValue, setSearchValue] = useState<string>('');
+
+  const { getSearchablePaths } = useDataContext();
+
+  if (!getSearchablePaths) {
+    return <Explorer selectedPath={selectedPath} onSelect={onSelect} />;
```
loichuder, line 29: In case `getSearchablePaths` is not defined, the bar with the explorer/search icons is not displayed.

packages/app/src/explorer/EntityItem.tsx
```diff
@@ -43,6 +46,12 @@ function EntityItem(props: Props) {
     }
   }, [entity, path, selectedPath, toggleExpanded]);
 
+  useEffect(() => {
+    if (isSelected) {
+      btnRef?.current?.focus();
+    }
+  }, [btnRef, isSelected]);
```
loichuder, line 53: This is to restore the focus on the selected element when tabbing from search tab to the explorer tab.

packages/app/src/explorer/Explorer.tsx
```diff
@@ -18,9 +18,13 @@ function Explorer(props: Props) {
...
 
+  const isRootSelected = selectedPath === '/';
+
   useEffect(() => {
-    btnRef.current?.focus();
-  }, [btnRef]);
+    if (isRootSelected) {
+      btnRef?.current?.focus();
+    }
+  }, [btnRef, isRootSelected]);
```
loichuder, line 27: Same as before but for the `/` path since the file button does not use `EntityItem`

packages/app/src/search/SearchTab.tsx
```diff
@@ -0,0 +1,50 @@
...
+  onChange: (newValue: string) => void;
+  selectedPath: string;
+  onSelect: (path: string) => void;
+}
+
+function SearchTab(props: Props) {
+  const { paths, value, onChange, selectedPath, onSelect } = props;
+
+  const matches = value
+    ? paths.filter((path) => getNameFromPath(path).includes(value))
```
loichuder, line 17: The search algorithm is very basic for now: get the name of the entity and see if it includes the pattern

packages/app/src/search/SearchTabContainer.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const [paths, setPaths] = useState<string[]>();
+
+  useEffect(() => {
+    const fetchPaths = async () => {
+      setPaths(await getSearchablePaths(BASE));
+    };
+    if (!paths) {
+      void fetchPaths();
+    }
+  }, [paths, getSearchablePaths]);
```
loichuder, line 33: Since the fetch store was dropped in #1367, I use a simple `useEffect` to handle the fetching. This could be moved up in the hierarchy to cache the `paths`.

packages/app/src/ExplorerTabs.tsx
```diff
@@ -0,0 +1,74 @@
...
+  Explore = 'explore',
+  Search = 'search',
+}
+
+function ExplorerTabs(props: Props) {
+  const { selectedPath, onSelect } = props;
+
+  const [tab, setTab] = useState<Tab>(Tab.Explore);
+
+  const [searchValue, setSearchValue] = useState<string>('');
```
loichuder, line 24: The state storing the `value` of the search `input` is handled here so that it can be restored when tabbing from the search tab to the explorer tab and back to the search tab.

packages/app/src/App.tsx
```diff
@@ -65,7 +65,7 @@ function App(props: Props) {
           flex={25}
           minSize={150}
         >
-          <Explorer selectedPath={selectedPath} onSelect={onSelectPath} />
+          <ExplorerTabs selectedPath={selectedPath} onSelect={onSelectPath} />
```
axelboc, line 68: What do you think of using a ui-less tab library, like https://headlessui.com/react/tabs? It would take care of the tab switching logic and of all the accessibility features for us.

The tabs code can still live in a separate component (to handle the `getSearchablePaths` logic and `searchText` state), but I would rename this component `Sidebar` (and the class of the parent reflex container `sidebarArea` maybe?), as when `getSearchablePaths` is not supported, we don't render any tabs.

packages/app/src/explorer/EntityItem.tsx
```diff
@@ -43,6 +46,12 @@ function EntityItem(props: Props) {
     }
   }, [entity, path, selectedPath, toggleExpanded]);
 
+  useEffect(() => {
+    if (isSelected) {
+      btnRef?.current?.focus();
```
axelboc, line 51: The first question mark should not be needed (same in next component).

packages/app/src/ExplorerTabs.tsx
```diff
@@ -0,0 +1,74 @@
+import { useState } from 'react';
+import { FiList, FiSearch } from 'react-icons/fi';
+
+import styles from './ExplorerTabs.module.css';
+import Explorer from './explorer/Explorer';
+import { useDataContext } from './providers/DataProvider';
+import SearchBoxContainer from './search/SearchTabContainer';
```
axelboc, line 7: Inconsistent file name here, though I'd probably rename this component `SearchContainer`, and its child `Search`. The term `SearchTab` is not quite right, since the "tab" is the element you click on -- it's technically a "panel" -- but I don't think it's worth mentioning in the component name, since it doesn't contain any code that makes it a tab panel specifically.

packages/app/src/search/SearchTabContainer.tsx
axelboc: Good for now. Eventually, it'd be good to show the search UI, let the user enter a search term, and show a spinner underneath the search field until the paths are fetched.

axelboc (commented):
Other general comments for a separate PR, perhaps:

- Implementing `getSearchablePaths` in the mock API
- Adding feature tests

packages/app/src/search/SearchTab.module.css
```diff
@@ -0,0 +1,21 @@
...
+.inputRow {
+  display: flex;
+}
+
+.inputRow input {
+  border-radius: 0.5rem;
+  margin: 0.25rem;
+  box-shadow: none;
+  height: 1.875rem;
+}
```
axelboc, line 15: Current aspect:

image

Nitpicking here, but it'd be good, I think, if the search box:
- had a bit more space all around it
- had a bit more padding inside it
- filled the sidebar in width

... and if the total height of `.inputRow` matched the height of the explorer's root group, so that the first search result and first child entity line up perfectly.

axelboc: The results could also use a bit of horizontal padding -- it should probably match the horizontal padding of the explorer's root group.

packages/app/src/search/SearchEntry.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <li key={path} className={styles.entity} role="none">
+      <button
+        className={styles.btn}
+        type="button"
+        role="treeitem"
+        data-path={path}
+        onClick={onClick}
+        aria-selected={isSelected}
+      >
```
axelboc, line 23: For now, since we show the full path, it'd be good to add a `title` attribute so users can see the full, uncropped path on hover.

packages/app/src/App.tsx
loichuder: I tried to replicate the tab logic we had for the Vis but I wouldn't mind using something else.

packages/app/src/search/SearchTab.module.css
loichuder: I updated the aspect according to your comments

packages/app/src/App.tsx
axelboc: Meh, it can wait. I'm fine if you just do the renaming.

packages/app/src/search/SearchTab.module.css
axelboc: Looks great! :star_struck:

axelboc (approved):

PR #1392, opened 2023-03-22
Ignore FillValue in HeatmapVis

### LINKED ISSUE #1181: Possible "ignore" netCDF-4 fill value/missing values?

loichuder:
Fix #1181 

I had to refactor a bit to revert part of #1381 where we replace the `fillValue` prop by the `ignoreValue` function.

In `HeatmapMaterial`. the check has to be done in shader so having the `ignoreValue` function is not useful.  Instead, I passed the  `fillValue` to be able to pass it as uniform to do the check in the shader.

For this, I broke down (again!) the parsing of `fillValue` and the creation of the `ignoreValue` callback in two hooks to use them in the different contexts:
- In `LineVisContainer` and its children, we only need the `ignoreValue` so I call the two hooks to generate `ignoreValue` and leave the rest of #1381 unchanged.
- In `HeatmapVisContainer`, I only get `fillValue` and pass it to the children (`MappedHeatmapVis`)
- In `MappedHeatmapVis`, we need `ignoreValue` so I use the hook to derive it from the `fillValue` prop but I still pass `fillValue` to the children
- In `HeatmapMaterial`, as explained, I need `fillValue` to implement the check in the shader so I get it from the props.

### DIFF
--- a/packages/app/src/vis-packs/core/heatmap/HeatmapVisContainer.tsx
+++ b/packages/app/src/vis-packs/core/heatmap/HeatmapVisContainer.tsx
@@ -10,6 +10,7 @@ import { useDimMappingState } from '../../../dimension-mapper/hooks';
 import VisBoundary from '../../VisBoundary';
 import type { VisContainerProps } from '../../models';
 import ValueFetcher from '../ValueFetcher';
+import { useFillValue } from '../hooks';
 import { getSliceSelection } from '../utils';
 import MappedHeatmapVis from './MappedHeatmapVis';
 import { useHeatmapConfig } from './config';
@@ -25,6 +26,8 @@ function HeatmapVisContainer(props: VisContainerProps) {
   const [dimMapping, setDimMapping] = useDimMappingState(dims, 2);
 
   const config = useHeatmapConfig();
+
+  const fillValue = useFillValue(entity);
 
   return (
     <>
@@ -45,6 +48,7 @@ function HeatmapVisContainer(props: VisContainerProps) {
               title={entity.name}
               toolbarContainer={toolbarContainer}
               config={config}
+              fillValue={fillValue}
             />
           )}
         />
--- a/packages/app/src/vis-packs/core/heatmap/MappedHeatmapVis.tsx
+++ b/packages/app/src/vis-packs/core/heatmap/MappedHeatmapVis.tsx
@@ -5,7 +5,11 @@ import { createPortal } from 'react-dom';
 
 import type { DimensionMapping } from '../../../dimension-mapper/models';
 import { useDataContext } from '../../../providers/DataProvider';
-import { useMappedArray, useSlicedDimsAndMapping } from '../hooks';
+import {
+  useIgnoreFillValue,
+  useMappedArray,
+  useSlicedDimsAndMapping,
+} from '../hooks';
 import { DEFAULT_DOMAIN, getSliceSelection } from '../utils';
 import HeatmapToolbar from './HeatmapToolbar';
 import type { HeatmapConfig } from './config';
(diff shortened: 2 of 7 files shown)


loichuder:
You can use the `value_2D` dataset in `test_miss.nc4` in `h5grove` if you want to test yourself.

axelboc (commented):
Arf, this is frustrating... but I've done some thinking. **Isn't this pretty much like masking?** Could we initialise a separate texture that "marks" the coordinates to be ignored?

The code that initialises the mask's typed array would then call the same `ignoreValue` function that is called when computing the domain.

In an ideal world, we'd even fill in this typed array _while_ computing the domain, to have a single loop, but we can't have everything we want... Anyway, the second loop can easily be skipped when the `ignoreValue` function is undefined, so it's not a big deal.

loichuder:
Excellent thinking ! I made a second attempt with the mask and it feels much nicer :wink: 

I am closing this PR in favour of the mask solution in #1395

PR #1393, opened 2023-03-24
Refactor and optimise creation of `DataCurve` geometries

axelboc:
Follows #1390

I'm refactoring `DataCurve` like I did with `ScatterPoints`. Here's what changes concretely in this PR:

- Buffer attributes are now instantiated once instead of being re-created each time.
- Buffer attributes are now instantiated with a pre-defined size instead of via dynamically-sized, plain JS arrays. This requires preparing error bars for every data point, regardless of whether an error bar is needed/valid at each point (with the `(0, 0, CAMERA_FAR)` trick).
- When errors are available, we now loop through the data only once to create all three geometries (data points, error caps, error bars), instead of three times.
- I've decided to initialise error geometries and render `ErrorBars` regardless of whether `showErrors` is enabled or disabled. The `showErrors` condition is now part of the `visible` condition of the Three objects involved in rendering the error bars and caps. I think it's better to have a bit more work in the main loop even when errors are toggled off, than to run a full second loop when the errors are toggled on. Note however that like before, if a dataset doesn't have `errors`, the error geometries are never created.
- Error bars are now rendered for error values equal to 0 (of course, when this happens, it looks like a single cap on top of the data point).
- Error bars are no longer rendered for error values lower than 0 (until now, the top/bottom bars and caps were just swapped, so it was impossible to tell the error value was invalid).


packages/lib/src/vis/line/DataCurve.tsx
```diff
@@ -1,14 +1,21 @@
...
 import ErrorBars from './ErrorBars';
 import GlyphMaterial from './GlyphMaterial';
-import { useCanvasPoints } from './hooks';
+import { useValueToErrorPositions, useValueToPosition } from './hooks';
 import { CurveType, GlyphType } from './models';
 
+/* Render points with NaN/Infinity coordinates (i.e. values <= 0 in log)
+ * at origin to avoid Three warning, and outside of camera's field of view
+ * to hide them and any segments connecting them. */
+const CAMERA_FAR = 1000; // R3F's default
```
axelboc, line 17: This constant and comment will eventually move into a dedicated subclass that extends `BufferGeometry`. This class will provide an abstraction for initialising geometries and updating buffer attributes, so it should hugely clean up the `useMemo` and `useLayoutEffect` code below.

packages/lib/src/vis/line/DataCurve.tsx
```diff
@@ -54,15 +61,106 @@ function DataCurve(props: Props) {
...
+      const { topCap, bottomCap, bar } = valueToErrorPositions(
+        value,
+        index,
+        pos
+      );
+
+      if (topCap) {
+        errorPositions.caps.setXYZ(index * 2 + 1, topCap[0], topCap[1], 0);
+      } else {
+        errorPositions.caps.setXYZ(index * 2 + 1, 0, 0, CAMERA_FAR);
```
axelboc, line 130: Again, all the `else` blocks will be hidden away in a method of the future subclass of `BufferGeometry`.

packages/lib/src/vis/line/hooks.ts
```diff
@@ -1,60 +1,6 @@
...
-        ? new Vector3(x, y, 0)
-        : /* Render points with NaN/Infinity coordinates (i.e. values <= 0 in log)
-           * at origin to avoid Three warning, and outside of camera's field of view
-           * to hide them and any segments connecting them. */
-          new Vector3(0, 0, CAMERA_FAR);
-
-      dataPoints.push(dataVector);
-
-      const error = errors?.[index];
-      if (!error || !hasFiniteCoords) {
```
axelboc, line 38: This condition is why error bars where not being rendered before for error values equal to `0`.

packages/lib/src/vis/line/hooks.ts
```diff
@@ -1,60 +1,6 @@
...
-        return;
-      }
-
-      const yErrBottom = ordinateScale(val - error);
-      const yErrTop = ordinateScale(val + error);
-
-      if (Number.isFinite(yErrBottom)) {
-        const errBottomVector = new Vector3(x, yErrBottom, 0);
-        errorBarSegments.push(errBottomVector, dataVector);
-        errorCapPoints.push(errBottomVector);
```
axelboc, line 48: We were pushing to a plain JS array conditionally, which is quite bad for memory management when dealing with large amounts of data.

packages/lib/src/vis/line/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    const showBottomCap = Number.isFinite(yCapBottom);
+
+    const yBarTop = showTopCap ? yCapTop : y;
+    const yBarBottom = showBottomCap ? yCapBottom : y;
+    const showBar = showTopCap || showBottomCap;
+
+    return {
+      topCap: showTopCap ? [x, yCapTop] : undefined,
+      bottomCap: showBottomCap ? [x, yCapBottom] : undefined,
+      bar: showBar ? [x, yBarTop, x, yBarBottom] : undefined,
```
axelboc, line 69: One downside of this implementation is that I'm initialising three arrays for each data point to be able to pass the coordinates back to `DataCurve`. Another reason why I'm not 100% convinced by that hooks, but it's a compromise between performance and readability.

packages/lib/src/vis/line/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    const yCapTop = ordinateScale(value + error);
+    const yCapBottom = ordinateScale(value - error);
+
+    const showTopCap = Number.isFinite(yCapTop);
+    const showBottomCap = Number.isFinite(yCapBottom);
+
+    const yBarTop = showTopCap ? yCapTop : y;
+    const yBarBottom = showBottomCap ? yCapBottom : y;
+    const showBar = showTopCap || showBottomCap;
```
loichuder, line 64: I can't seem to remember what happens if the top cap can be displayed but not the bottom cap.

Do we want to show the full bar in this case ?

loichuder (approved):
Yeah, not convinced either by the hooks. 

But let's move forward with this see how it goes with future refactorings.

packages/lib/src/vis/line/utils.ts
axelboc: This flag helps me decide whether or not to render _a_ bar at all, but the length of the bar is controlled by the `yBarTop` and `yBarBottom` logic, which can by either `yCapTop /Bottom` or the data point's `y` coordinate.

PR #1394, opened 2023-03-27
Test search in Cypress

loichuder:
Fix https://github.com/silx-kit/h5web/pull/1391#pullrequestreview-1352175079

> - Implementing getSearchablePaths in the mock API
> - Adding feature tests


cypress/e2e/app.cy.ts
```diff
@@ -231,6 +231,41 @@ describe('/mock', () => {
...
+    cy.findByRole('tab', { name: 'Open explorer panel' }).click();
+    cy.findExplorerNode('scatter').should('have.attr', 'aria-selected', 'true');
+
+    // Check that the search value is kept when tabbing back from Explorer
+    cy.findByRole('tab', { name: 'Open search panel' }).click();
+    cy.findByRole('textbox', { name: 'Path to search' }).should(
+      'have.value',
+      'scatter'
+    );
+  });
```
loichuder, line 267: These two tests could be merged to avoid duplicating the early steps. 

But I thought that the resulting test would be too long.

cypress/index.d.ts
```diff
@@ -1,6 +1,7 @@
 declare namespace Cypress {
   interface Chainable {
     waitForStableDOM(iteration?: number): void;
+    findExplorerNode(name: string): ReturnType<typeof cy.findByRole>;
```
loichuder, line 4: New command to find an explorer node without selecting it. I used it to check if a node was indeed selected in the explorer after selecting it in the search panel.

packages/app/src/Sidebar.tsx
```diff
@@ -38,6 +38,7 @@ function Sidebar(props: Props) {
           role="tab"
           aria-selected={tab === Tab.Explore}
           onClick={() => setTab(Tab.Explore)}
+          aria-label="Open explorer panel"
```
loichuder, line 41: Added some aria-labels to be able to query the element

packages/app/src/providers/mock/mock-api.ts
```diff
@@ -115,6 +116,26 @@ export class MockApi extends DataProviderApi {
...
+
+    if (!isGroup(entity)) {
+      return [entity.path];
+    }
+
+    return entity.children.reduce<string[]>(
+      (acc, child) => [...acc, ...this.getEntityPaths(child.path)],
+      [entity.path]
+    );
+  }
```
loichuder, line 137: Implementation of `getSearchablePaths` for the mock API as asked. This was also used for the Cypress tests.

cypress/e2e/app.cy.ts
axelboc: Since this is E2E, I wouldn't see a problem with merging the two. I wouldn't bother expecting for the search and `NX Scatter` tabs to be selected, though.

You could maybe add an expectation that the `scatter` node receives focus when tabbing back to the explorer.

packages/app/src/Sidebar.tsx
axelboc: Very minor again, but taking example on VS Code, I would shorten those to `Explorer` and `Search`: 

image

Screen reader users would already know they're on a tab and understand that each tab opens the corresponding panel.

packages/app/src/providers/mock/mock-api.ts
axelboc: Very nice :clap:

axelboc (approved):

PR #1396, opened 2023-03-28
Fix Cypress snapshots directory path

loichuder:
Cypress snapshots were no longer checked for since there was a mismatch between the spec file name `app.cy.ts` and the path to the snapshots directory `snapshots/app.spec.ts`.

I renamed the snapshots directory to `snapshots/app.cy.ts` to fix this.


axelboc (approved):

loichuder:
For the record, here is a compare since the renaming: https://github.com/silx-kit/h5web/compare/be8b2fd45a21a1db92640b74b4311733e9b089d1...5146c1c45caa05d52a6c1134277903df535d3bcb

Issue #1388, opened 2023-03-16
Event propagation to canvas


axelboc:
### Is your feature request related to a problem?

Currently, `useCanvasEvents` registers its event handlers on the WebGL `canvas`. As a result, any element positioned on top of the canvas (via `Html`, `SvgElement`, `FloatingControl`, etc.) **intercepts every event before it can reach the `canvas` element**, thus preventing interaction (zoom, pan, etc.). To work around this and let events reach the canvas, we use `pointer-events: none` on every overlayaing element.

The problem arises with interactive overlaying elements - e.g. a resize handle on an SVG ROI; the ROI itself, to be able to move it around, etc. Currently, we use `pointer-events: auto` to re-enable intercepting events on specific interactive elements, but the downside is that it then re-intercepts _every_ event! There's no way to fine-grain which events to propagate and which to stop -- **it's all or nothing**.

An example of why this is problematic: if we want a ROI to be draggable, then there's no way to let through wheel events so users can still zoom on the visualisation underneath.

### Requested solution or feature

**`useCanvasEvents` could register its event handlers on `.r3fRoot` instead of `canvas` in order to take advantage of event bubbling**.

The upside of this solution is its simplicity — we don’t try to “catch” and re-fire events or anything like that, and we don’t need to set `pointer-events: none` on every overlaying element. The downside is that it's an opt-out-only solution: _every_ event bubbles by default, so now we need to stop events that we don't want to let through to the canvas.

For instance, in this configuration, pressing down on the reset zoom button in the floating toolbar and then dragging the cursor away would pan the visualization. In this case, I'd see a few ways to "opt out" to prevent the pan:

1. Put `onPointerDown={evt => evt.stopPropagation()}` on the reset zoom button.
    - Upsides: 1) fine-grained control; 2) very explicit, no hidden magic.
    - Downsides: 1) we have to do this for every event we want to stop (most interactions rely on `pointerdown` to start, but this may not be the case in the future or with custom interactions); 2)  if users implement their own floating controls, they have to remember to stop events themselves; 3) users have to do this for every element they want to make interactive.
1. Add a `div` wrapper inside `FloatingControl` and stop events there.
    - Upside: 1) every control stops the same events and users don’t have to do it themselves.
    - Downside: 1) less control over which events to stop; 2) hidden behaviour; 3) solution specific to floating controls.
1. In `useCanvasEvents`'s event handlers, detect if the target of the event is a `button` and stop processing the event if that’s the case. 
[...]

PR #1397, opened 2023-03-28
Fix stacking order of canvas

axelboc:
Regression introduced in #1387 ... so much for bringing stacking under control :sweat: 

The canvas was still stacked implicitly at `z-index: 0`, so in front of the grid and SVG elements... Sorry about that. So now it moves behind everything.

Unfortunately, in doing so, `r3fRoot`—which is still stacked implicitly at `z-index: 0` (and cannot be stacked explicitly without creating a new stacking context and breaking the stacking order, as explained in a comment in the CSS)—captures events before the reach the canvas... I had to work around this by moving `pointer-events: none` to `r3fRoot` instead of setting it on every overlay, and then re-enabling it to `auto` on the canvas and floating toolbar.

This feels better in the end, since it removes a bunch of `pointer-events: none` and saves us and consumers from forgetting to add them.

--- 

Note that this may very well change when dealing with #1388. To be continued...

### DIFF
--- a/packages/lib/src/vis/shared/VisCanvas.module.css
+++ b/packages/lib/src/vis/shared/VisCanvas.module.css
@@ -6,20 +6,34 @@
 
   /*
    * Stacking order, from furthest to closest:
-   * 1. axis system (i.e. most notably the grid)
-   * 2. SVG overlay
-   * 3. anything rendered via `Html` (`Overlay`, `Annotation`, etc.) -- by default with no z-index, which is equivalent to `z-index: 0`
-   * 4. tooltip
-   * 5. floating toolbar
+   * 1. WebGL canvas
+   * 2. axis system (i.e. most notably the grid)
+   * 3. SVG overlay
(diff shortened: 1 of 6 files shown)


packages/lib/src/vis/shared/VisCanvas.module.css
```diff
@@ -6,29 +6,42 @@
...
   /*
    * Stacking order, from furthest to closest:
+   * 1. WebGL canvas
    * 1. axis system (i.e. most notably the grid)
    * 2. SVG overlay
-   * 3. anything rendered via `Html` (`Overlay`, `Annotation`, etc.) -- by default with no z-index, which is equivalent to `z-index: 0`
-   * 4. tooltip
+   * 3. anything rendered via `Html` (`Overlay`, `Annotation`, etc.) — by default with no z-index, which is equivalent to `z-index: 0`
+   * 4. tooltip + guides
    * 5. floating toolbar
```
loichuder, line 14: Missed the increment in numbers here.

loichuder (approved):
Thanks!

PR #1400, opened 2023-03-28
Ensure no files are created when E2E test pass in CI

axelboc:
I'm using https://github.com/marketplace/actions/clean-git-status to make sure there are no untracked files after the Cypress tests pass (when run without `--updateSnapshots=true`).

This can occur in two cases, as we've experienced just now and in the past:

1. When adding a new snapshot test, which takes a brand new snapshot.
2. When renaming the spec file, which puts new snapshots in a different folder.

In both cases, the CI wasn't failing, so we weren't notified and the reference snapshots would not get updated.


.github/workflows/lint-test.yml
```diff
@@ -125,6 +125,10 @@ jobs:
         env:
           CYPRESS_TAKE_SNAPSHOTS: true
 
+      - name: Require approval when new snapshots are created 🔎
+        uses: numtide/clean-git-action@v1
+        if: success()
```
axelboc, line 130: This is not technically required, as it is the default - i.e. if a step fails, the following steps don't run. In this case, I want to make it clearer that this is sort of an `if/else` with the step below.

axelboc (commented):
I've renamed the spec file for testing purposes—I'll rename it back after your review:

image

loichuder (approved):

PR #1402, opened 2023-03-30
Align sidebar elements with breadcrumbs bar and toolbar

axelboc:
I'm formalising the heights of various elements so things line up properly between the sidebar and the main area.

The core change is that I'm giving the same height to all the elements that have the potential to be lined up with each other: the sidebar's tab bar, the breadcrumbs bar, the explorer's root entity button, the search input row, the metadata viewer's header rows.

Giving the same height to the breadcrumbs bar and to the toolbar simplifies things a lot because when the search is enabled, the search input row must line up with the toolbar, and when the search is disabled, the explorer's root entity button must line up with the breadcrumbs bar.

I introduce a CSS variable in the viewer to control this height: `--toolbar-height`. Seems logical that the toolbar be the element that other elements line up to, since it's kind of the most important one.

I also add a CSS variable to control the height of the toolbar buttons: `--toolbarBtn-height`, which I use for explorer buttons too, and I map those two variables to two more variables in the lib: `--h5w-toolbar--height` and `--h5w-btn--height`. I give those default values such that the toolbar and buttons in Braggy (notably) keep the same height as before.

This is what it looks like now:

image

image


packages/app/src/search/Search.module.css
```diff
@@ -5,14 +5,24 @@
...
+  flex: 1 1 0%;
+  height: var(--toolbarBtn-height);
+  padding: 0 0.5rem;
+  border: none;
   border-radius: 0.5rem;
-  margin: 0.625rem;
-  padding: 0.25rem;
-  box-shadow: none;
+  box-shadow: 0 0 2px dimgray;
+  outline: none;
```
axelboc, line 20: Replacing the outline with a box-shadow for cross-browser consistency and for consistency with toolbar inputs (i.e. cell width in Matrix vis and bound editors in domain tooltip). That being said, I use a "flat" shadow instead of a "raised" shadow, since the sidebar's styles are "flatter" that the toolbar's.

packages/app/src/visualizer/VisSelector.module.css
```diff
@@ -7,7 +7,7 @@
   composes: btnClean from global;
   display: flex;
   align-items: center;
-  padding: 0.375rem 1rem;
+  padding: 0.125rem 1rem 0;
```
axelboc, line 10: Removing the now-obsolete padding top/bottom, but keeping a bit to align tab labels better.

packages/lib/src/toolbar/Toolbar.module.css
```diff
@@ -56,15 +56,15 @@ a.btn {
...
+  height: var(--h5w-toolbar-btn--height, 1.875rem);
   padding: 0 0.5rem;
   border-radius: 0.5rem;
   transition: background-color 0.05s ease-in-out, box-shadow 0.05s ease-in-out;
   white-space: nowrap;
 }
 
 .btn[data-small] > .btnLike {
-  height: 1.75rem;
+  height: calc(var(--h5w-toolbar-btn--height, 1.875rem) - 0.125rem);
```
axelboc, line 67: The height of small toolbar buttons is now computed.

packages/lib/src/toolbar/Toolbar.module.css
```diff
@@ -4,7 +4,7 @@
   z-index: 1; /* for toolbar menus to appear above visualizations (overflow, selectors) */
   display: flex;
   min-width: 0;
-  height: 2.5rem;
+  height: var(--h5w-toolbar--height, 2.5rem);
```
axelboc, line 7: As mentioned, the default values are smaller to avoid visual regressions in existing applications.

loichuder (approved):
Aligning the search elements indeed felt a bit hacky so it is a nice improvement

PR #1403, opened 2023-03-30
Update Cypress reference snapshots

axelboc:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


cypress/snapshots/app.cy.ts/nximage.snap.png
axelboc, line 1: This screenshot shows the effect of the new transparent color for bad values, I think.

axelboc: Obviously the checkerboard doesn't appear in the viewer.

PR #1404, opened 2023-03-30
Deprecate prop `explorerOpen` in favour of `sidebarOpen`

axelboc:
As discussed, I'm deprecating `explorerOpen` instead of removing it to avoid a breaking change in the short term. We'll remove it in v8.


loichuder (approved):

loichuder:
Note that you could also rename the test `start with the explorer closed` om `app.cy.ts`. Up to you.

PR #1410, opened 2023-04-04
Upgrade dependencies, including zustand

axelboc:
- https://github.com/d3/d3-array/releases 3.2.2 and 3.2.3, notably https://github.com/d3/d3-array/pull/264
- https://github.com/prettier/prettier/blob/main/CHANGELOG.md#285 TS 5 support
- https://github.com/axios/axios/releases 1.2.3 and above
- https://github.com/react-hookz/web/releases/tag/v23.0.0 – React says getting rid of the "state updated after unmount" warning, which is what `useSafeState` is for, is not worth it. They've actually silenced the warning in React 18. We weren't using `useSafeState` anyway.
- https://github.com/pmndrs/zustand/releases 4.1.6 and above – biggest change for this PR: the `createContext` API is deprecated; they now recommend using React's `createContext` directly: https://github.com/pmndrs/zustand#react-context


packages/app/src/vis-packs/core/matrix/MatrixToolbar.tsx
```diff
@@ -9,27 +9,27 @@ import {
...
 import type { ExportFormat, ExportURL } from '../../../providers/models';
-import { useMatrixConfig } from './config';
+import type { MatrixVisConfig } from './config';
 
 const EXPORT_FORMATS: ExportFormat[] = ['npy', 'csv'];
 
 interface Props {
   cellWidth: number;
   isSlice: boolean;
+  config: MatrixVisConfig;
```
axelboc, line 19: Passing the config as prop to be consistent with other toolbars.

eslint.shared.js
```diff
@@ -99,6 +99,9 @@ module.exports = {
             // Disallow shadowing variables for an outer scope, as this can cause bugs
             // when the inner-scope variable is removed, for instance
             '@typescript-eslint/no-shadow': 'error',
+
+            // Warn on deprecated APIs (TypeScript strikes them out but doesn't report them)
+            'etc/no-deprecated': 'warn',
```
axelboc, line 104: Deprecated APIs were not reported. This is on the radar of the guy developing Galex.

packages/lib/src/vis/tiles/store.ts
```diff
@@ -1,4 +1,4 @@
-import create from 'zustand';
+import { create } from 'zustand';
```
axelboc, line 1: Note the difference between `create` and `createStore`. The former returns a custom hook, while the latter returns a vanilla store object. It's the latter that can be injected via a context and retrieved with `useStore(myVanillaStore)` (or `useStore(useContext(MyVanillaStoreContext)`). Here we use a simple global store, so we can use the hook API.

packages/app/src/App.tsx
```diff
@@ -28,8 +28,8 @@ interface Props {
 
 function App(props: Props) {
   const {
-    explorerOpen = true,
-    sidebarOpen: initialSidebarOpen = explorerOpen,
+    explorerOpen = true, // eslint-disable-line etc/no-deprecated
+    sidebarOpen: initialSidebarOpen = explorerOpen, // eslint-disable-line etc/no-deprecated
```
axelboc, line 32: Shame that the linting rule reports our own deprecations, but I guess it makes sense... :sweat_smile:

packages/lib/src/vis/shared/Tick.tsx
```diff
@@ -4,7 +4,8 @@ function Tick(props: TickRendererProps) {
   const { formattedValue, ...tickProps } = props;
 
   return (
-    /* Increase `dy` slightly to fix alignment and spacing with tick lines */
+    // Increase `dy` slightly to fix alignment and spacing with tick lines
+    // @ts-expect-error (probably caused by visx being on React 18)
```
loichuder, line 8: Might be fixed by updating `@types/react`

loichuder (approved):

PR #1412, opened 2023-04-05
Upgrade dependencies

axelboc:
Mostly dev deps this time except for `react-error-boundary`. Nothing exciting:

- https://github.com/bvaughn/react-error-boundary/releases/tag/4.0.0
- https://github.com/Swatinem/rollup-plugin-dts/blob/master/CHANGELOG.md
- https://github.com/rollup/rollup/releases?page=2
- https://github.com/isaacs/rimraf/blob/main/CHANGELOG.md
- https://github.com/facebook/jest/releases
- https://docs.cypress.io/guides/references/changelog


packages/app/src/vis-packs/core/raw/RawVisContainer.tsx
```diff
@@ -11,7 +11,7 @@ function RawVisContainer(props: VisContainerProps) {
   assertNonNullShape(entity);
 
   return (
-    <VisBoundary resetKey={entity}>
+    <VisBoundary>
```
axelboc, line 14: Just a quick thing I noticed while looking at our usage of `ErrorBoundary`. There's no need for this `resetKey` prop here, since when `entity` changes, the path changes and therefore the whole subtree is unmounted/remounted. In other components, we pass either `dimMapping` or nothing like here when there's no dimension mapper.

loichuder (approved):
Sorry for the late review

PR #1414, opened 2023-04-12
Rename prop `maskValues` to `mask`

axelboc:
This PR suggests renaming the `maskValues` prop in `HeatmapMesh` and `HeatmapMaterial` to just `mask`.

I know `maskValues` takes inspiration from prop `alphaValues`, but I feel like this is a bit different, since a "mask" in this context implies that it's an array, so the `Values` suffix feels redundant. More importantly, `maskValues` can be interpreted as a callback, like `ignoreValue`, since `mask` can be a verb. 

In addition to this, I document the `ignoreValue` and the now-called `mask` props in Storybook:

image

image

image

image

### DIFF
--- a/packages/lib/src/vis/heatmap/HeatmapVis.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapVis.tsx
@@ -6,9 +6,7 @@ import {
   ScaleType,
 } from '@h5web/shared';
 import type { NdArray } from 'ndarray';
-import ndarray from 'ndarray';
 import type { ReactElement, ReactNode } from 'react';
-import { useMemo } from 'react';
 
 import type { DefaultInteractionsConfig } from '../../interactions/DefaultInteractions';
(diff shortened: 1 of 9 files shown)


packages/lib/src/vis/heatmap/HeatmapVis.tsx
```diff
@@ -81,18 +79,7 @@ function HeatmapVis(props: Props) {
...
-        ? ndarray(
-            Uint8Array.from(
-              dataArray.data.map((v) => (ignoreValue(v) ? 255 : 0))
-            ),
-            dataArray.shape
-          )
-        : undefined,
-    [dataArray, ignoreValue]
-  );
+  const maskArray = useMask(dataArray, ignoreValue);
```
axelboc, line 82: I'm also extracting this into a utility+memoised hook

loichuder (approved):
Great clean-up :+1:

PR #1416, opened 2023-04-13
Upgrade Storybook to v7

axelboc:
First part of the upgrade.

- [Release notes](https://github.com/storybookjs/storybook/releases/tag/v7.0.4) and [Migration guide (overview)](https://storybook.js.org/docs/react/migration-guide/)
- [Migration guide (in depth)](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-65x-to-700) (I'll give this another look at the end of the upgrade process to make sure I haven't missed anything)
- [Design changes](https://storybook.js.org/blog/7-0-design-alpha/)
- [Framework API](https://storybook.js.org/blog/framework-api/) and [First-class Vite support](https://storybook.js.org/blog/first-class-vite-support-in-storybook/) – They now provide convenient packages to configure Storybook for both a specific front-end library (e.g. React, Vue, etc.) and a specific builder (e.g. webpack, Vite, etc.). In our case, we used to configure Storybook with React (obviously) and the webpack builder (we had even migrated to the latest webpack 5 builder), so I could have just used `@storybook/react-webpack`. However, since Storybook 7 now has first-class support for Vite, I went straight to `@storybook/react-vite` and it works great! There's a bunch of warnings in the console due to Rollup, but apparently they can be safely ignored.
- [Storybook Ecosystem CI](https://storybook.js.org/blog/storybook-ecosystem-ci/) – They have a new CI that runs everyday against the latest releases of React, Vue, Vite, webpack etc. to catch problems. There's even a status page: https://storybook.js.org/status?ref=storybookblog.ghost.io
- [Storybook 7 Docs](https://storybook.js.org/blog/storybook-7-docs/) – Previously, the "documentation" page of a story was accessible via a "Docs" tab in the toolbar of each story ... but that page was the same for every story in a folder, so it made little sense. Now it's much better: the "Docs" page for each folder appears in the navigation sidebar. For the rest, we had already migrated to MDX v2, so easy-peasy.

The second part of the upgrade will consist in upgrading the stories from v2 to v3 of the Component Story Format (CSF). Storybook v7 is backwards-compatible with CSF v2, so I think it's best to do the CSF upgrade separately.

- [CSF v3](https://storybook.js.org/blog/storybook-csf3-is-here/) – seems fairly straightforward, but the diff will be horrible.
- [Improved type safety](https://storybook.js.org/blog/improved-type-safety-in-storybook-7/) – uses TS' new `satisfies` keyword.


axelboc:
Due to typing issues, I'll branch off here to work on the CSF upgrade and merge back in here before merging into `main`. You can still review this PR and ignore the failing CI check.

apps/storybook/src/decorators/FillHeight.tsx
```diff
@@ -1,9 +1,14 @@
...
 
-function FillHeight(MyStory: Story) {
+function FillHeight(Story: StoryFn, context: StoryContext) {
   return (
-    <div style={{ display: 'grid', height: '100vh' }}>
-      <MyStory />
+    <div
+      style={{
+        display: 'grid',
+        height: context.viewMode === 'story' ? '100vh' : '25rem', // fixed height in docs mode
```
axelboc, line 8: This improves documentation pages by reducing the height of the iframes of the documented stories.

loichuder (approved):
:clap:

PR #1417, opened 2023-04-14
Require pnpm v8

axelboc:
There's a bunch of default configuration changes and a new lockfile version that is apparently more resilient to merge conflicts: https://github.com/pnpm/pnpm/releases/tag/v8.0.0

Config changes that are worth mentioning:

- [`auto-install-peers`](https://pnpm.io/npmrc#auto-install-peers): now enabled by default – as the name implies, pnpm now installs missing peer deps automatically, so we can remove a bunch of overrides from `package.json`. npm decided to reinstate the automatic installation of peer dependencies in v7+, which brought out a whole bunch of problems ... but pnpm can actually do it safely thanks to its non-flat `node_modules` structure.
- [`resolution-mode`](https://pnpm.io/npmrc#resolution-mode): changed from `highest` (`^1.0.0` => `1.2.3`) to `lowest-direct` (`^1.0.0` => `1.0.0`) – This is really good, since it means that the versions specified in `package.json` are the versions that get installed, regardless of `^` or `~` (when there's no lockfile of course). The `^` and `~` characters become useful only for commands such as `pnpm outdated`, or `pnpm update` (without `--latest`). This config change is not relevant to us, since we have a lockfile and we use exact dependency versions anyway (except for specific `@types` dependencies, so we can run `pnpm -r update` to update them easily).
- [`resolve-peers-from-workspace-root`](https://pnpm.io/npmrc#resolve-peers-from-workspace-root): seems convenient but I like the explicitness of installing peer deps in packages that need them, so even though it's now enabled by default, I haven't made any changes to our `package.json` files.

### DIFF
--- a/.github/workflows/approve-snapshots.yml
+++ b/.github/workflows/approve-snapshots.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Install pnpm ⚙️
         uses: pnpm/action-setup@v2
         with:
-          version: 7.x
+          version: 8.x
 
       - name: Restore cache 📌
         uses: actions/cache@v3
@@ -52,7 +52,7 @@ jobs:
         run: pnpm build
 
       - name: Serve ⚡️
-        run: pnpm serve & pnpx wait-on http://localhost:5173
+        run: pnpm serve & pnpm dlx wait-on http://localhost:5173
 
       - name: Run Cypress 🌳
         run: pnpm cypress:run --env updateSnapshots=true
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,8 @@
 - [Development](#development)
   - [`pnpm` v7 cheat sheet](#pnpm-v7-cheat-sheet)
   - [Dependency management](#dependency-management)
-  - [Workspace dependencies](#workspace-dependencies)
+    - [DefinitelyTyped packages](#definitelytyped-packages)
+    - [Workspace dependencies](#workspace-dependencies)
   - [Icon set](#icon-set)
 - [Build](#build)
   - [Package builds](#package-builds)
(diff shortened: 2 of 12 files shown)


.github/workflows/approve-snapshots.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         run: pnpm build
 
       - name: Serve ⚡️
-        run: pnpm serve & pnpx wait-on http://localhost:5173
+        run: pnpm serve & pnpm dlx wait-on http://localhost:5173
```
axelboc, line 55: `pnpx` is not actually documented, so I wonder if they're planning to drop it. To be safe, I'm switching to the more verbose command.

CONTRIBUTING.md
```diff
@@ -65,23 +66,21 @@ pnpm start
...
-
-`pnpm.overrides` is currently used to force the version of ESLint to the one
-required by `eslint-config-galex`. This is needed because, in the `demo`
-project, `vite-plugin-eslint` depends on an older version of ESLint.
-
-### Workspace dependencies
+
+> If you run into peer dependency warnings and other package resolution issues,
+> note that `pnpm` offers numerous solutions for dealing with them, like
+> [`pnpm.peerDependencyRules.allowedVersions`](https://pnpm.io/package_json#pnpmpeerdependencyrulesallowedversions).
```
axelboc, line 72: Simplifying this paragraph to only mention the option we actually use.

CONTRIBUTING.md
```diff
@@ -65,23 +66,21 @@ pnpm start
...
+
+#### [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) packages
+
+The major versions of `@types/*` packages must be aligned with the major
+versions of the packages they provide types for—i.e. `foo@x.y.z` requires
+`@types/foo@^x`.
+
+For convenience, some `@types` packages can be quickly upgraded to their latest
+minor/patch version by running `pnpm up -r`.
```
axelboc, line 81: Thought I'd rewrite this a bit as well, notably to shorten the command to just `pnpm up -r`. We only specify `^` on those exact packages, so the result is the same and it's easier to remember.

loichuder (approved):

PR #1418, opened 2023-04-14
Upgrade TypeScript and Galex + switch to new type-only import/export syntax

axelboc:
- [simple-import-sort](https://github.com/lydell/eslint-plugin-simple-import-sort) replaces `import/order` to bring case-insensitive and named imports sorting.
- TS 5.0 introduces a new setting called [`verbatimModuleSyntax`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#verbatimmodulesyntax) to help with identifying and removing type-only imports at compile time. This setting, which replaces two other settings now deprecated, enforces the use of the following syntax for type-only imports: `import { type Foo } from 'foo'` (instead of `import type { Foo } from 'foo'`). The linting rule  [consistent-type-specifier-style](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/consistent-type-specifier-style.md) (with option `prefer-inline`) has an automatic fixer for this which helped switch to the new syntax.
- With the above change, regular and type-only named imports can now be merged into the same statements — I've enabled a rule to check for duplicated imports and fixed them all with the help of VS Code's _Organise imports_ feature (since the ESLint rule doesn't come with an automatic fixer).
- Rule `consistent-type-specifier-style` is not yet able to enforce the `export { type Foo }` syntax, so I've made the change manually for now.
- To keep our packages' entry files readable, I've disabled sorting of `export` statements, and I've kept type-only exports separate from regular exports.

To help with the review, the most relevant files are `eslint.shared.js`, `tsconfig.json` and the various `package.json`. Everything else is just changes to imports/exports. Note also that I've been very careful not to re-order CSS files.


axelboc:
Closing as Jest doesn't like `verbatimModuleSyntax` at all... It wants me to switch to running test with ES modules support, which is still [highly experimental](https://jestjs.io/docs/ecmascript-modules). It got me nowhere :sob:

PR #1419, opened 2023-04-14
Upgrade TypeScript and Galex

axelboc:
Supersedes #1418 -- I'm only upgrading TS and Galex; I'm not switching to the new setting [`verbatimModuleSyntax`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#verbatimmodulesyntax) (cf. #1418 for what that setting was about).

- [simple-import-sort](https://github.com/lydell/eslint-plugin-simple-import-sort) replaces `import/order` and brings case-insensitive sorting, as well as sorting of named imports.
- I enable rule `import/no-duplicates` to check for duplicated imports and fix them all with the rule's automatic fixer.
- To keep our packages' entry files readable, I disable sorting of `export` statements.

[Galex' release notes](https://github.com/ljosberinn/eslint-config-galex/releases) and [TypeScript 5.0's release blog post](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#verbatimmodulesyntax) for reference. The new `"moduleResolution": "builder"` option looks very promising to allow multiple entry points in packages. I'll test it later on.


cypress/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../tsconfig.json",
   "compilerOptions": {
     "isolatedModules": false,
+    "sourceMap": false,
```
axelboc, line 5: Fix for Cypress after TS upgrade — `sourceMap: true` in our root `tsconfig.json` is incompatible with another option set by Cypress :shrug:

packages/app/src/vis-packs/core/utils.ts
```diff
@@ -71,7 +71,7 @@ export function applyMapping<T>(
 export function getSliceSelection(
   dimMapping?: DimensionMapping
 ): string | undefined {
-  if (!dimMapping || !dimMapping.some(isNumber)) {
+  if (!dimMapping?.some(isNumber)) {
```
axelboc, line 74: The only linting fix unrelated to imports.

tsconfig.json
```diff
@@ -22,6 +22,7 @@
     "noFallthroughCasesInSwitch": true,
     "forceConsistentCasingInFileNames": true,
     "importsNotUsedAsValues": "error",
+    "ignoreDeprecations": "5.0",
```
axelboc, line 25: Silencing the deprecation of `importsNotUsedAsValues`

packages/app/src/vis-packs/core/utils.ts
loichuder: Not a fan of the combination of `!` with `?.`. I find it hard to process mentally.

tsconfig.json
loichuder: I hope you will remember why this line (and the `sourceMap: false` from `cypress` ) is here cause I will not :joy:

loichuder (approved):

packages/app/src/vis-packs/core/utils.ts
axelboc: Yeah agreed, I'll revert and silence the error

tsconfig.json
axelboc: Hehe :joy:  I guess this line is easy: you remove it and you get a warning in the IDE. The other warning shows up when running Cypress, so it's less obvious.

PR #1425, opened 2023-04-20
Convert stories to CSF3

axelboc:
This converts all the stories to CSF3. I've reviewed things along the way, which led me to add/remove/simplify/refactor some stories. I'll try to point out every such case.

I really like the how much simpler it is to compose stories and share template/render function, notably across files. I also like having the `meta` object at the top, before the stories. 

It would be awesome if we could move `VisCanvas`, `DefaultInteractions` and the like into decorators, as this would save us from declaring custom `render` functions in a lot of stories (directly or via spreading - e.g. `...Default`). Unfortunately, this is still not possible for us because of the R3F renderer; we get errors saying that ReactDOM doesn't understand Three elements and cannot render them. Storybook seems to render stories separately from their decorators, with the ReactDOM renderer; it must take the resulting DOM elements and pass them to the decorators after the fact. :shrug:

- Relevant migration articles: https://storybook.js.org/blog/storybook-csf3-is-here/, https://storybook.js.org/blog/improved-type-safety-in-storybook-7/
- Relevant documentation articles: https://storybook.js.org/docs/react/writing-stories/introduction, https://storybook.js.org/docs/react/writing-stories/stories-for-multiple-components (note that you have to select "TypeScript 4.9 in the drop-down at the top to see the `satisfies` syntax)
  - Regarding the docs, I find that they are still a bit thin when it comes to creating stories with more args than accepted by the documented component. I don't believe [this example](https://storybook.js.org/docs/react/writing-stories/stories-for-multiple-components#creating-a-template-component) passes type-checking, since the `List` component accepts a `children` prop, not an `items` prop. Long story short, I found that I had to use `satisfies Meta<MyCompPropsWithExtraArgs>` (and `type Story = StoryObj<MyCompPropsWithExtraArgs>`) instead of `Meta<typeof MyComp>` (and `type Story = StoryObj<typeof meta>`).


apps/storybook/src/LineVis.stories.tsx
```diff
@@ -31,114 +29,17 @@ const errorsArray = ndarray(
...
-  excludeStories: ['LineVisTemplate'],
-  parameters: { layout: 'fullscreen', controls: { sort: 'requiredFirst' } },
   decorators: [FillHeight],
+  parameters: {
+    layout: 'fullscreen',
+    controls: { sort: 'requiredFirst' },
+  },
   args: {
+    dataArray,
+    domain: undefined, // compute dynamically in each story based on scale type, errors and auxiliaries
```
axelboc, line 42: This is an important change. I'm letting the domain be computed dynamically in the stories by default to avoid errors when the user changes the scale type with the controls. Users can still set a `domain` of their own choosing with the controls if they want.

apps/storybook/src/LineVis.stories.tsx
```diff
@@ -157,4 +58,103 @@ export default {
...
+      ]);
+
+    return <LineVis {...args} domain={effectiveDomain} />;
+  },
+} satisfies Story;
+
+export const Domain = {
+  ...Default,
+  args: {
+    domain: [1, 20], // compatible with all scale types
```
axelboc, line 86: Changing the `domain` here with one that works will all the scales.

apps/storybook/src/LineVis.stories.tsx
```diff
@@ -31,114 +29,17 @@ const errorsArray = ndarray(
...
-};
-
-export const AuxiliaryArrays = Template.bind({});
-AuxiliaryArrays.args = {
-  dataArray: primaryArray,
-  auxiliaries: [{ array: secondaryArray }, { array: tertiaryArray }],
-};
-
-export const AuxiliaryLabels = Template.bind({});
-AuxiliaryLabels.args = {
```
axelboc, line 99: This felt redundant. I've moved the labels into the `AuxiliaryArrays` story. I guess we no longer demonstrate what the tooltip looks like when there are no labels, but I don't think that's very interesting.

apps/storybook/src/MatrixVis.stories.tsx
```diff
@@ -12,41 +11,61 @@ const complexDataArray = getMockDataArray<H5WebComplex>(
...
+type Story = StoryObj<typeof meta>;
+
+export const Default = {
+  args: {
+    dataArray,
+    formatter: (val) => formatMatrixValue(val as number),
+  },
+} satisfies Story;
+
+export const CellWidth = {
```
axelboc, line 37: New story to document `cellWidth` on its own (previously via the `Complex` story).

apps/storybook/src/MatrixVis.stories.tsx
```diff
@@ -12,41 +11,61 @@ const complexDataArray = getMockDataArray<H5WebComplex>(
...
+  args: {
+    ...Default.args,
+    cellWidth: 250,
+  },
+} satisfies Story;
+
+export const StaticHeaderCells = {
```
axelboc, line 44: Renamed from `StaticIndexCells` to match the wording used in the code.

apps/storybook/src/MatrixVis.stories.tsx
```diff
@@ -12,41 +11,61 @@ const complexDataArray = getMockDataArray<H5WebComplex>(
...
+} satisfies Story;
+
+export const TypedArray = {
+  args: {
+    dataArray: toTypedNdArray(dataArray, Float32Array),
+    formatter: (val) => formatMatrixValue(val as number),
+  },
+} satisfies Story;
+
+export const ColumnHeaders = {
```
axelboc, line 66: New story to document `columnHeaders`.

apps/storybook/src/Overlay.stories.tsx
```diff
@@ -1,35 +1,45 @@
...
-    </VisCanvas>
-  );
-};
-
-export default {
+const meta = {
   title: 'Building Blocks/Overlay',
-  component: Default,
-  parameters: { layout: 'fullscreen' },
+  component: Overlay,
```
axelboc, line 8: Now documenting `Overlay`'s props in the controls panel (but with disabled controls).

apps/storybook/src/SelectionTool.stories.tsx
```diff
@@ -17,230 +12,215 @@ import {
...
-              fillOpacity="0.3"
-            />
-          </SvgElement>
-        )}
-      </SelectionTool>
-
-      {persistedDataSelection && (
-        <DataToHtml points={persistedDataSelection}>
-          {(...htmlSelection) => (
+        <Pan modifierKey={modifierKey?.length === 0 ? 'Control' : undefined} />
```
axelboc, line 54: The pan modifier key is now computing dynamically like in other stories. This allows the args/controls to match the real props of the `Pan` component.

apps/storybook/src/TiledHeatmapMesh.stories.tsx
```diff
@@ -1,390 +0,0 @@
-import type {
```
axelboc, line 1: This file moves into a new `TiledHeatmapMesh` folder. The API classes, utilities and the `TileParams` model all move to separate files.

apps/storybook/src/TiledHeatmapMesh/TiledHeatmapMesh.stories.tsx
```diff
@@ -0,0 +1,207 @@
...
+        <group
+          scale={[
+            abscissaConfig.flip ? -1 : 1,
+            ordinateConfig.flip ? -1 : 1,
+            1,
+          ]}
+        >
+          <TiledHeatmapMesh {...tiledHeatmapProps} />
+        </group>
+        <TiledTooltipMesh renderTooltip={renderTooltip} />
```
axelboc, line 94: I've removed the extra `showTooltip` and `showAxes` args. I'm now always showing the tooltips and I've removed the `NoAxes` story (since this is already documented by a dedicated `VisCanvas` story).

apps/storybook/src/TiledHeatmapMesh/checkerboard-api.ts
```diff
@@ -0,0 +1,53 @@
...
+import { clamp } from 'lodash';
+import type { NdArray } from 'ndarray';
+import ndarray from 'ndarray';
+import { createFetchStore } from 'react-suspense-fetch';
+import type { Vector2 } from 'three';
+
+import type { TileParams } from './models';
+import { areTilesEqual } from './utils';
+
+const getCheckerboardArray = greenlet(
```
axelboc, line 13: The checkerboard tiles arrays are now computed in a worker like with the mandlebrot API. This gets rid of a lot of jank when playing with the `Checkerboard` story. (Note also the renaming of the word "checkboard" to "checkerboard" everywhere.)

apps/storybook/src/TiledHeatmapMesh/checkerboard-api.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+);
+
+export class CheckerboardTilesApi extends TilesApi {
+  private readonly store;
+
+  public constructor(size: Size, tileSize: Size) {
+    super(tileSize, getLayerSizes(size, tileSize));
+
+    this.store = createFetchStore(
```
axelboc, line 25: Using a suspense store like in the mandlebrot API in order for the `get` method to remain synchronous.

apps/storybook/src/Toolbar.stories.tsx
```diff
@@ -84,60 +90,42 @@ const Template: Story<TemplateProps> = (args) => {
...
-    docs: {
-      source: { type: 'code' },
-      page: () => (
-        <>
-          <Title />
-          <p>
-            The source code of the <em>Default</em> story demonstrates a basic
-            use of the <code>Toolbar</code> component and associated toolbar
-            controls. To reveal the code, click on <em>Show code</em> below the
-            preview.
```
axelboc, line 129: I ddin't keep these paragraphs. I think the new documentation page is clear enough.

apps/storybook/src/Toolbar.stories.tsx
```diff
@@ -84,60 +90,42 @@ const Template: Story<TemplateProps> = (args) => {
...
-DocumentInteractions.args = {
-  narrow: true,
-  interactions: [
-    { shortcut: 'Wheel', description: 'Turn' },
-    { shortcut: 'Space', description: 'Accelerate' },
-    { shortcut: 'Ctrl+Alt', description: 'Do a backflip' },
+export const Responsive = {
+  ...Default,
+  decorators: [
+    (Story: StoryFn) => (
```
axelboc, line 105: Using a decorator here to remove the custom `narrow` arg/control.

apps/storybook/src/TooltipMesh.stories.tsx
```diff
@@ -1,76 +1,83 @@
...
+  args: {
+    guides: 'horizontal',
+    renderTooltip: (x, y) => (
+      <TooltipContent text={'guides="horizontal"'} x={x} y={y} />
+    ),
+  },
+} satisfies Story;
 
-export default {
-  ...VisCanvasStoriesConfig,
```
axelboc, line 62: We used to include the controls/args from `VisCanvas` but them remove them via `parameters.controls.include` below...

apps/storybook/src/TooltipMesh.stories.tsx
```diff
@@ -1,76 +1,80 @@
...
-
-export const Guides = Template.bind({});
-Guides.args = {
-  tooltipValue: 'guides="both"',
-  guides: 'both',
-  panZoom: false,
-};
+  ),
+  args: {
+    renderTooltip: (x, y) => <TooltipContent x={x} y={y} />,
```
axelboc, line 48: Now providing `renderTooltip` as an arg (facilitated by abstracting the rendering logic into `TooltipContent`)

apps/storybook/src/TooltipMesh.stories.tsx
```diff
@@ -1,76 +1,80 @@
...
+  },
+} satisfies Story;
 
-export const WithPanZoom = Template.bind({});
-WithPanZoom.args = {
-  tooltipValue: '<Pan /> and <Zoom /> must come first',
-  guides: 'vertical',
-  panZoom: true,
-};
+export const HorizontalGuide = {
```
axelboc, line 52: Adding separate stories for `guides="horizontal"` and `guides="vertical"`.

axelboc (commented):
New record? :rofl: 

image

As mentioned, I recommend side-by-side review. To visualise changes in the individual stories's definitions, tick "Hide whitespace"; to visualise changes to the `meta` objects, untick "Hide whitespace".

Good luck... :sweat_smile:

apps/storybook/src/Pan.stories.tsx
```diff
@@ -93,4 +28,85 @@ export default {
...
+  args: {
+    disabled: true,
+  },
+} satisfies Story;
+
+export const TwoComponents = {
+  render: (args) => {
+    return (
+      <VisCanvas
+        title="Pan with middle button, or with <modifierKey> + left button"
```
loichuder, line 90: Idea: `<modifierKey>` could be dynamically set to `${args.modifierKey}` ?

apps/storybook/src/Zoom.stories.tsx
```diff
@@ -49,4 +18,43 @@ export default {
...
+  ...Default,
+  args: {
+    modifierKey: ['Control'],
+  },
+} satisfies Story;
+
+export const MultipleModifierKeys = {
+  ...Default,
+  args: {
+    modifierKey: ['Control', 'Shift', 'Alt'],
```
loichuder, line 51: I don't think it is worth having all three keys: it is a bit janky to test. Let's remove one.

loichuder (approved):
Thanks for the hard work :clap:

apps/storybook/src/Pan.stories.tsx
loichuder: Good call, I forgot that `modifierKey` could be an array.

PR #1427, opened 2023-04-20
Enforce slightly stricter TS config

axelboc:
Inspired by https://www.npmjs.com/package/@tsconfig/strictest. I tried enabling a few more options but in the end, I've made only one change to `tsconfig.json`. The other options specified in `@tsconfig/strictest` that we don't already specify are either:

- too crazy and not worth the pain (`exactOptionalPropertyTypes`, `noUncheckedIndexedAccess` and `noPropertyAccessFromIndexSignature`); or
- overly restrictive in development and already reported by ESLint anyway (`allowUnreachableCode`, `noUnusedLocals`, `noUnusedParameters`).

They did allow me to detect and fix a few minor issues, though, which is nice.

### DIFF
--- a/packages/app/src/providers/h5grove/h5grove-api.ts
+++ b/packages/app/src/providers/h5grove/h5grove-api.ts
@@ -112,7 +112,7 @@ export class H5GroveApi extends DataProviderApi {
     const { data } = await handleAxiosError(
       () =>
         this.client.get<H5GroveEntityResponse>(`/meta/`, { params: { path } }),
-      (status, errorData) => {
+      (_, errorData) => {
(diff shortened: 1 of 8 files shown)


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -276,7 +276,6 @@ export class H5GroveApi extends DataProviderApi {
   }
 
   private async processAttrsMetadata(
-    path: string,
```
loichuder, line 279: What an oversight :joy:

loichuder (approved):

PR #1431, opened 2023-05-02
Enforce stricter scale types

axelboc:
Multiple components accepting scale types as props (either directly or via an object prop), as well as utility functions and hooks, were wrongly allowed to receive:

- `ScaleType.Gamma` without a corresponding exponent
- `ScaleType.Sqrt` and `ScaleType.Gamma` even when those scale types were not meant to be supported

The first issue was due to type `VisScaleType` incorrectly using `Omit` instead of `Exclude` to deal with the special case of `ScaleType.Gamma`. The second issue was due to props and parameters being typed with `ScaleType` without consideration for unsupported scales.

I've done a few more refactorings along the way in separate commits, like renaming the confusing `AxisScale` and `createAxisScale` to just `Scale` and `createScale`, and splitting the `H5WEB_SCALES` record, which was dealing with totally unrelated responsibilities.


packages/lib/src/vis/utils.ts
```diff
@@ -46,16 +66,27 @@ const adaptedLogTicksThreshold = scaleLinear({
...
 
-export function createAxisScale(
+export function createScale(
   scaleType: VisScaleType,
-  config: VisxScaleConfig | ScaleGammaConfig
-): AxisScale {
-  if (isScaleType(scaleType)) {
-    return H5WEB_SCALES[scaleType].createScale(config);
+  config: ScaleConfig
+): Scale {
```
axelboc, line 72: As mentioned, there's definitely an opportunity here to use a generic to infer a stricter return type. This would allow `abscissaScale` and `ordinateScale` to have `type AxisScale =  PickD3Scale<AxisScaleType, number>` instead of just `Scale` (which includes gamma/sqrt). However, I preferred to keep this for a future PR.

packages/lib/src/vis/utils.ts
```diff
@@ -46,16 +66,27 @@ const adaptedLogTicksThreshold = scaleLinear({
...
+  if (Array.isArray(scaleType)) {
+    const [, exponent] = scaleType;
+    return scaleGamma({ ...(config as ScaleGammaConfig), exponent });
   }
 
-  const [, exponent] = scaleType;
-  return H5WEB_SCALES[ScaleType.Gamma].createScale({ ...config, exponent });
```
axelboc, line 58: Typing is tricky to get right with this pattern (we used to type the `config` parameter with `any`...) – especially if we decide to introduce generics. The `switch` statement gets rid of this complexity by removing a layer of abstraction.

packages/app/src/vis-packs/core/complex/ComplexLineToolbar.tsx
```diff
@@ -43,11 +44,13 @@ function ComplexLineToolbar(props: Props) {
         label="X"
         value={xScaleType}
         onScaleChange={setXScaleType}
+        options={AXIS_SCALE_TYPES}
```
axelboc, line 47: Note that we had forgotten to pass `options` arrays, which meant that all scales were shown in the drop-down. Since `xScaleType` is now typed to `AxisScaleType`, we get proper errors when we don't pass `options`.

packages/shared/src/models-vis.ts
```diff
@@ -30,6 +30,9 @@ export enum ScaleType {
   Gamma = 'gamma',
 }
 
+export type ColorScaleType = Exclude<ScaleType, 'gamma'>;
+export type AxisScaleType = Exclude<ColorScaleType, 'sqrt'>;
```
loichuder, line 34:
```suggestion
export type AxisScaleType = Exclude<ScaleType, 'sqrt' | 'gamma'>;
```

Might be clearer to remove the interdependency of both types

packages/lib/src/toolbar/controls/ScaleSelector/ScaleSelector.tsx
```diff
@@ -1,22 +1,17 @@
...
 
-interface Props {
+interface Props<T extends ScaleType> {
   label?: string;
-  value: ScaleType;
-  onScaleChange: (scale: ScaleType) => void;
-  options?: ScaleType[];
+  value: T;
+  onScaleChange: (scale: T) => void;
+  options: T[];
```
loichuder, line 10: Nice :+1:

packages/lib/src/vis/scales.ts
```diff
@@ -1,57 +0,0 @@
...
-import SqrtIcon from '../toolbar/controls/ScaleSelector/SqrtIcon';
-import type { AxisScale } from './models';
-import { scaleGamma } from './scaleGamma';
-
-interface H5WebScale {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  createScale: (config: any) => AxisScale;
-  Icon: IconType;
-  label: string;
-  validMin: number;
```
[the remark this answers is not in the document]
loichuder: Good call :+1:

loichuder (approved):

PR #1424, opened 2023-04-20
Update h5wasm to 0.4.11

loichuder:
Fix https://github.com/silx-kit/h5web/issues/1421#issuecomment-1513061952 (related to https://github.com/usnistgov/h5wasm/issues/49)

PR #1433, opened 2023-05-22
Parse and display chunks when present in h5wasm metadata

loichuder:
https://github.com/silx-kit/h5web/pull/1424#issuecomment-1515863874

Unfortunately, this does not work for some reason. See the path `/gzip` in https://h5web.panosc.eu/h5grove?file=compressed.h5 that has `Chunk size` for h5grove and not for h5wasm.

Perhaps an issue to forward to Brian.


axelboc (approved):

loichuder:
Reported at https://github.com/usnistgov/h5wasm/issues/53

loichuder:
> Unfortunately, this does not work for some reason.

In fact, this is working fine. I may have mistaken since the demo https://h5web.panosc.eu was not updated accordingly.

PR #1353, opened 2023-01-18
Upgrade dependencies

axelboc:
- [axios 1.2.0 to 1.2.2](https://github.com/axios/axios/releases)
  - There's a breaking change with type `AxiosRequestConfig` in 1.2.3 that affects the `axios-hooks` (amongst others https://github.com/axios/axios/issues/5476) -- this might be corrected in the next release of `axios`.
- [d3-array 3.2.1](https://github.com/d3/d3-array/releases/tag/v3.2.1)
- [@testing-library/dom 8.19.0 to 8.20.0](https://github.com/testing-library/dom-testing-library/releases)
- [@react-hookz/web 19.2.0 to 22.0.0](https://github.com/react-hookz/web/releases)
- [vite 3.2.4 to 4.0.4](https://github.com/vitejs/vite/blob/v4.0.4/packages/vite/CHANGELOG.md)
    - v4 brings `@vitejs/plugin-react-swc` as an alternative to `@vitejs/plugin-react` to use SWC instead of Babel in developement. I'll try it at some point, but not in this PR.
    - There are also some browser compatibility changes, so we'll have to be careful upgrading Vite to v4 in other projects, like Braggy or Daiquiri.
- [Cypress 11.2.0 to 12.3.0](https://docs.cypress.io/guides/references/changelog#12-0-0)
    - Cypress 12 automatically [clears the browser context between tests](https://docs.cypress.io/guides/references/migration-guide#test-isolation), including `localStorage`, so I removed our own bit of code that did that.

PR #1434, opened 2023-05-23
Upgrade Vite to v4

axelboc:
I had tried the upgrade before (cf. https://github.com/silx-kit/h5web/pull/1353) but had run into issues when running tests in Braggy. Turns out Vite 4 upgrades its internal version of Rollup to v3, which changes a bunch of default configuration options: https://rollupjs.org/migration/#changed-defaults

Jest loads CommonJS scripts when available (otherwise it loads ESM modules and transforms them to CommonJS before running the tests). The problem was that some of the `require` statements in the CommonJS bundle of `@h5web/lib` were returning objects with a `default` property (instead of the content of that `default` property directly). In other words, some ESM imports in the lib were transformed into `require('...')` instead of `require('...').default` by Vite/Rollup.

The culprit was [`output.interop`](https://rollupjs.org/configuration-options/#output-interop), which controls how Rollup converts ESM `import` statements into CommonJS `require` statements. The default value changed from `'compat'` to `'default'`, so I'm now explicitly setting it back to `'compat'`. I'll let the docs explain what these options mean (if you have the courage).

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "eslint-config-galex": "4.5.2",
     "identity-obj-proxy": "3.0.0",
     "jest": "29.5.0",
-    "jest-transform-stub": "2.0.0",
     "npm-run-all": "4.1.5",
     "prettier": "2.8.7",
     "ts-jest": "29.1.0",
@@ -54,7 +53,9 @@
     "peerDependencyRules": {
       "allowedVersions": {
         "cypress-image-snapshot>cypress": "*",
-        "jest-image-snapshot>jest": "*"
+        "jest-image-snapshot>jest": "*",
+        "@phenomnomnominal/tsquery>typescript": "5.x",
+        "eslint-plugin-etc>typescript": "5.x"
       }
     },
     "requiredScripts": [
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -115,6 +115,8 @@ export { scaleGamma } from './vis/scaleGamma';
 export { useCanvasEvents } from './interactions/hooks';
 export { default as Box } from './interactions/box';
 
+export { COLOR_SCALE_TYPES, AXIS_SCALE_TYPES } from '@h5web/shared';
+
 // Models
 export { INTERPOLATORS } from './vis/heatmap/interpolators';
 export { ScaleType } from '@h5web/shared';
@@ -132,7 +134,14 @@ export type {
 } from './interactions/models';
 export { MouseButton } from './interactions/models';
 
-export type { Domain, VisibleDomains, Dims, Axis } from '@h5web/shared';
+export type {
+  Domain,
+  VisibleDomains,
+  Dims,
+  Axis,
+  ColorScaleType,
+  AxisScaleType,
+} from '@h5web/shared';
 
 export type {
   Aspect,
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -54,7 +53,9 @@
     "peerDependencyRules": {
       "allowedVersions": {
         "cypress-image-snapshot>cypress": "*",
-        "jest-image-snapshot>jest": "*"
+        "jest-image-snapshot>jest": "*",
+        "@phenomnomnominal/tsquery>typescript": "5.x",
+        "eslint-plugin-etc>typescript": "5.x"
```
axelboc, line 58: Fixing some peer deps warnings I hadn't fixed when upgrading `eslint-config-galex`. After I asked for it, pnpm added a CLI flag to check for peer deps issues: `pnpm i --resolution-only`, which is quite convenient.

packages/lib/src/index.ts
```diff
@@ -115,6 +115,8 @@ export { scaleGamma } from './vis/scaleGamma';
 export { useCanvasEvents } from './interactions/hooks';
 export { default as Box } from './interactions/box';
 
+export { COLOR_SCALE_TYPES, AXIS_SCALE_TYPES } from '@h5web/shared';
```
axelboc, line 118: Sorry for dumping this in here. I'm exporting these arrays and types (`ColorScaleType` and `AxisScaleType`) so I can use them in Braggy.

packages/lib/src/toolbar/OverflowMenu.tsx
```diff
@@ -1,17 +1,17 @@
...
 import styles from './OverflowMenu.module.css';
 import Separator from './Separator';
 
-interface Props {
-  children: ReactElement[];
-}
+interface Props {}
 
-function OverflowMenu(props: Props) {
+function OverflowMenu(props: PropsWithChildren<Props>) {
```
axelboc, line 12: When debugging the error in Braggy, I thought the issue came from here at first. I refactored the type of the `children` prop to be more lenient. In summary, I allow any children to be passed to `OverflowMenu`, which simplifies the JSX code in `Toolbar`, but then I make sure to take into account only valid `ReactElement` children below.

packages/lib/src/toolbar/Toolbar.tsx
```diff
@@ -20,7 +20,7 @@ function Toolbar(props: PropsWithChildren<Props>) {
 
   /* Convert `children` to flat array by traversing nested arrays and fragments.
    * (Note that `flattenChildren` guarantees stable string keys regardless of JSX logic.) */
-  const allChildren = flattenChildren(children).filter(isReactElement);
+  const allChildren = flattenChildren(children).filter(isValidElement);
```
axelboc, line 23: Turns out React already has an equivalent utility function/type guard.

packages/lib/src/toolbar/Toolbar.tsx
```diff
@@ -65,10 +63,8 @@ function Toolbar(props: PropsWithChildren<Props>) {
       </div>
 
       <OverflowMenu>
-        {[
-          ...initialOverflows,
-          ...outOfView.filter((child) => child.type !== Separator),
-        ]}
+        {overflowChildren}
+        {outOfView.filter((child) => child.type !== Separator)}
```
axelboc, line 67: Much cleaner: taking advantage of the fact that arrays can be rendered as is in the JSX.

Issue #1313, opened 2022-12-06
HDF5 Dimension Scales


axelboc:
It would be nice to support HDF5's dimension scales. We've received multiple feature requests relating to those, including in a couple of recent emails. Dimension scales are apparently used quite extensively in NetCDF4 files, to describe how to plot datasets (axes, units, etc.).

In his email, Jean-Christophe describes, for instance, how to create a time-based dimension scale (which could then be "attached" to a dataset - e.g. `values.dims[0].attach_scale(time)`).

```python
df_modified = pd.to_datetime(df.index.values) - time # Difference of time from a given date
df_modified_str = df_modified.total_seconds().to_numpy()
time_dset = group.create_dataset ('time', data=df_modified_str)
time_dset.attrs["long_name"] = "UTC Time"
time_dset.attrs["description"] = ModelHDFLevel4Lumina.TIME.__doc__    
time_dset.attrs["calendar"] = "standard"
time_dset.attrs["units"] = f"seconds since {time.strftime('%Y-%m-%d %H:%M:%S')}"
time_dset.make_scale('time')
```

More reading:

- [Primer on HDF5 Dimension Scales on StackOverflow](https://stackoverflow.com/questions/41704019/how-to-use-hdf5-dimension-scales-in-h5py)
- [h5py documentation](https://docs.h5py.org/en/latest/high/dims.html)
- A [blog post](https://www.unidata.ucar.edu/blogs/developer/en/entry/netcdf4_shared_dimensions) with a dump of a sample NetCDF4 file that shows HDF5 Dimension Scales.
- The [NetCDF Climate and Forecast Conventions](https://cfconventions.org/cf-conventions/cf-conventions.html) (cf. section [_4.4. Time Coordinate_](https://cfconventions.org/cf-conventions/cf-conventions.html#time-coordinate), for the naming convention used in the `units` attribute above to indicate that a dimension scale dataset contains relative timestamps.)

axelboc:
[h5wasm@0.6.9](https://github.com/usnistgov/h5wasm/releases/tag/v0.6.9) brings support for reading dimension scales.

axelboc:
@zhqrbitee any chance you could share a sample file with us?

NAThompson:
[hdf5-dimension-scales.patch](https://github.com/user-attachments/files/30951345/hdf5-dimension-scales.patch)

@axelboc : Let me know if you'd like this to be turned into a full merge request.

PR #1435, opened 2023-05-23
Show NX axis labels as hints in axis mappers

### LINKED ISSUE #1426: Legend for LineVis, HeatmapVis, MatrixVis

For multi-dimensional data sets, it is not obvious what the parameters labelled D0, D1, D2,... are (e.g. which physical quantities are behind D0, D1, ...), that can be selected for the x- and y-axes in MatrixVis, LineVis and HeatmapVis plots.

It would help to have a small legend which defines what D0, D1, D2, ... are (depending on whatever names are provided by the user, for example, D0: energy, D1: intensity etc.) if there is more than one Di.

axelboc:
Fix #1426 (it's only for NeXus for now, but we have a dedicated epic for supporting HDF5 dimension scales: #1313)

This is a very basic implementation — probably not the best UX, but I haven't found a better idea for now. @cmmngr, is this what you had in mind?

Peek 2023-05-23 14-33


cmmngr:
Looks great. This is exactly what we had in mind.

packages/lib/src/toolbar/controls/ToggleGroup.tsx
```diff
@@ -24,16 +29,17 @@ function useToggleGroupProps(): ToggleGroupProps {
   return context;
 }
 
-interface BtnProps {
+interface BtnProps extends HTMLAttributes<HTMLButtonElement> {
```
loichuder, line 32: The `extends` is not needed since we do not propagate `props` to `button`, right ?

loichuder (approved):

PR #1437, opened 2023-05-25
Upgrade h5wasm and switch to TS' `"moduleResolution": "bundler"`

axelboc:
[H5wasm v0.5.0](https://github.com/usnistgov/h5wasm/releases/tag/v0.5.0) brings the ability to [dynamically link](https://emscripten.org/docs/compiling/Dynamic-Linking.html) WASM modules at runtime, which will hopefully allow us to support HDF5 compression plugins in the `@h5web/h5wasm` package (notably for [myHDF5](https://myhdf5.hdfgroup.org/) and the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=h5web.vscode-h5web)).

In addition, [h5wasm v0.5.2](https://github.com/usnistgov/h5wasm/releases/tag/v0.5.2) (v0.5.1 was skipped) fixes a package distribution issue that prevented it from working with TS's new [`moduleResolution: "bundler"` setting](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler). I've now switched to this setting, but note that Cypress doesn't like it in the root `tsconfig.json` file, so I had to create a dedicated `tsconfig.bundler.json` file, which the projects' tsconfigs now extend.

`moduleResolution: "bundler"` means that TS can now read [`exports` declarations](https://nodejs.org/api/packages.html#package-entry-points) in `package.json` files, which should hopefully allow us to split the `@h5web/shared` package's entrypoint so Vite/Rollup can tree shake the output bundles better.

### DIFF
--- a/packages/h5wasm/src/models.ts
+++ b/packages/h5wasm/src/models.ts
@@ -1,8 +1,9 @@
-import type { Group as H5WasmGroup, Metadata } from 'h5wasm';
 import type {
   CompoundTypeMetadata,
   EnumTypeMetadata,
-} from 'h5wasm/src/hdf5_util_helpers';
+  Group as H5WasmGroup,
+  Metadata,
+} from 'h5wasm';
 
 export type H5WasmEntity = ReturnType<H5WasmGroup['get']>;
 
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,7 +8,6 @@
     "allowJs": true,
     "skipLibCheck": true,
     "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
     "isolatedModules": true,
     "resolveJsonModule": true,
     "noEmit": true,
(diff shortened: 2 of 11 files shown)


packages/h5wasm/src/models.ts
```diff
@@ -1,8 +1,9 @@
-import type { Group as H5WasmGroup, Metadata } from 'h5wasm';
 import type {
   CompoundTypeMetadata,
   EnumTypeMetadata,
-} from 'h5wasm/src/hdf5_util_helpers';
+  Group as H5WasmGroup,
+  Metadata,
+} from 'h5wasm';
```
axelboc, line 6: `h5wasm@0.5.2` also adds those missing types.

tsconfig.json
```diff
@@ -8,7 +8,6 @@
     "allowJs": true,
     "skipLibCheck": true,
     "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
```
axelboc, line 11: This is automatically enabled when `esModuleInterop` is enabled.

loichuder (approved):

Issue #1422, opened 2023-04-19
[DomainSlider] Moving one histogram marker toggles off both min and max autoscale


axelboc:
### To Reproduce

1. Go to https://h5web-docs.panosc.eu/?path=/story/toolbar-domainslider--histogram (autoscale on for both min and max)
2. Open domain slider tooltip and move min or max marker on histogram
3. Observe how autoscale is now off for both min and max instead of just min or just max (depending on marker that was moved).

### Expected behaviour

Autoscale should be switched off only for the bound corresponding to the marker that is moved. This is what happens when interacting with the main slider.

PR #1439, opened 2023-05-31
Moving one histogram marker now toggles off only the linked autoscale

loichuder:
Fix https://github.com/silx-kit/h5web/issues/1422


axelboc (approved):
Nice change! So... breaking change, technically :sweat_smile:

PR #1442, opened 2023-06-05
Split Storybook deployment workflow from package publishing

axelboc:
... and try [new GitHub Pages deployment](https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta/) ([more info](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow)). Instead of deploying via a `github-pages` branch, we can now deploy to a special [`github-pages` "environment"](https://github.com/silx-kit/h5web/settings/environments/1423492/edit).

This allows us to remove the `.nojekyll` workaround, since Jekyll is not involved in the new deployment process. Obviously, I'll have to merge this PR to try the new workflow, but I've added a manual trigger, so I won't have to publish a tag.

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -1,0 +1,64 @@
+name: Deploy Storybook
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - v*
+      - '!v*-beta.*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🏷️
+        uses: actions/checkout@v3
+
+      - name: Set up Node 🕹️
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: 'package.json'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install pnpm ⚙️
+        uses: pnpm/action-setup@v2
+        with:
+          version: 8.x
+
+      - name: Restore cache 📌
+        uses: actions/cache@v3
+        with:
+          path: |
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,64 @@
+name: Deploy Storybook
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - v*
+      - '!v*-beta.*'
```
axelboc, line 8: Note that I'm now excluding beta tags, so that the Storybook gets deployed only for official releases. We can still trigger the workflow manually if need be.

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ubuntu-latest
+
+    permissions:
+      pages: write
+      id-token: write
+
+    steps:
+      - name: Deploy Storybook 🚀
+        id: deployment
+        uses: actions/deploy-pages@v2
```
axelboc, line 60: This action looks for the `github-pages.tar` artifact and publishes it to an environment called `github-pages`.

.github/workflows/publish-packages.yml
```diff
@@ -1,4 +1,4 @@
-name: Release
```
axelboc, line 1: I've renamed this workflow.

loichuder (approved):
Good stuff :+1:

PR #1443, opened 2023-06-07
Type `abscissaScale` and `ordinateScale` more strictly

axelboc:
Follow-up to #1431 to enforce stricter types for the internal `createScale` and `getCanvasScale` utility functions, and therefore for `abscissaScale` and `ordinateScale` (exposed through `VisCanvasContext`).

To do so, I've added a generic to `Scale`, in order for `abscissaScale` and `ordinateScale` to be typed as `Scale<AxisScaleType>`, and I've reintroduced `AxisScale` (removed in #1431, as before it incorrectly meant "any of the supported scales") as an alias to this type.

The [Storybook](https://h5web-docs.panosc.eu/?path=/docs/context--docs) was still saying that the type of `abscissaScale` and `ordinateScale` was `AxisScale` even though it had basically been renamed to `Scale` in #1431. But with the stronger typing and the reintroduction of `AxisScale`, it is now correct.

### DIFF
--- a/packages/lib/src/vis/models.ts
+++ b/packages/lib/src/vis/models.ts
@@ -48,6 +48,9 @@ export interface AxisConfig {
 }
 
 export type VisScaleType = ColorScaleType | [ScaleType.Gamma, number];
+export type ExtractScaleType<V extends VisScaleType> = V extends ColorScaleType
+  ? V
+  : ScaleType.Gamma;
 
 export interface ScaleGammaConfig {
   domain?: Domain;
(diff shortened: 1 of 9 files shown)


packages/lib/src/vis/models.ts
```diff
@@ -48,6 +48,9 @@ export interface AxisConfig {
 }
 
 export type VisScaleType = ColorScaleType | [ScaleType.Gamma, number];
+export type ExtractScaleType<V extends VisScaleType> = V extends ColorScaleType
+  ? V
+  : ScaleType.Gamma;
```
axelboc, line 53: This is the TS equivalent of extracting the `ScaleType` enum value from a `VisScaleType` parameter (if array: `visScaleType[0]`; else: `visScaleType`).

packages/lib/src/vis/shared/VisCanvasProvider.tsx
```diff
@@ -68,8 +68,8 @@ function VisCanvasProvider(props: PropsWithChildren<Props>) {
     [width, height]
   );
 
-  const abscissaScale = getCanvasScale(abscissaConfig, visSize.width);
-  const ordinateScale = getCanvasScale(ordinateConfig, visSize.height);
+  const abscissaScale = getCanvasAxisScale(abscissaConfig, visSize.width);
+  const ordinateScale = getCanvasAxisScale(ordinateConfig, visSize.height);
```
axelboc, line 72: Also renaming this utility for more clarity.

packages/lib/src/vis/utils.ts
```diff
@@ -66,6 +68,11 @@ const adaptedLogTicksThreshold = scaleLinear({
   range: [0.8, 1.4],
 });
 
+export function createScale<
+  V extends VisScaleType,
+  S extends ExtractScaleType<V>
+>(scaleType: V, config: ScaleConfig<S>): Scale<S>;
```
axelboc, line 74: Couldn't do it without a function override. It's simpler to keep the main function signature "generic" to avoid resorting to workarounds inside the function body, and to enforce a stricter signature for the caller with an override.

loichuder (approved):

Issue #1382, opened 2023-03-10
[Toolbar] `DomainSlider` and `DomainControls` components decoupled from tooltip UI


axelboc:
Part of #1377 -- see also https://github.com/silx-kit/h5web/pull/1379#issuecomment-1463589489

The idea is to allow using the domain slider and controls (bound editors, min/max autoscale, histogram, etc.) without being constrained by the tooltip UI (in #1377 specifically, the UI requires the domain controls to be in a modal dialog), and without having to re-implement all the logic from the current `DomainSlider` and `DomainTooltip` components.

PR #1445, opened 2023-06-07
Rename `DomainSlider` to `DomainWidget`

axelboc:
First step in addressing #1382 and #1377. This is a **breaking change**, obviously. The idea is to then rename `ScaledSlider` to `DomainSlider` and export it.

The end goal is to have the following components exported:

- `DomainWidget`: the whole toolbar control that allows choosing a custom domain (i.e. slider + tooltip);
- `DomainSlider`: the actual slider;
- `DomainControls`: the content of the tooltip (bounds inputs + auto-scale buttons).

The tooltip behaviour will remain the responsibility of `DomainWidget`.

I haven't renamed the CSS variables for now -- I'll do it after I rename/create the other components.


packages/lib/src/toolbar/controls/DomainWidget/DomainWidget.module.css
```diff
@@ -19,15 +19,3 @@
 .root[data-disabled] > .sliderContainer {
   opacity: 0.5; /* edit button is itself disabled, so don't fade it more */
 }
-
-.slider {
```
axelboc, line 23: Moved to `ScaledSlider.module.css`

axelboc:
I've created a branch called `domain-widget` into which I'll merge this and the next PRs. I won't merge `domain-widget` into `main` until we're ready for another major release.

apps/storybook/src/Customization.mdx
```diff
@@ -146,7 +146,7 @@ as you see fit. For instance, if you'd like to change the color of the curve of
 | `--h5w-selector-option-focus--outlineColor` | `gray`                                       | Outline color of focused option      |
 | `--h5w-selector-groupLabel--color`          | `gray`                                       | Text color of group labels in menu   |
 
-##### Domain slider
+##### Domain widget
 
 | Name                                                 | Default                    | Description                                    |
```
loichuder, line 151: Do we need to change the CSS variable names ?

loichuder (approved):

loichuder:
We already have https://github.com/silx-kit/h5web/pull/1439 which is technically a breaking change so you can merge to main if you wish.

apps/storybook/src/Customization.mdx
axelboc: I'll do it after I rename/create the other components, because most will not likely change (if `ScaledSlider` becomes `DomainSlider` and is exported, then the variables can still be called `--h5w-domainSlider-<...>`.

loichuder: Sure. Just wanted to be sure that you had it in mind but it was an unneeded concern apparently :wink:

PR #1448, opened 2023-06-21
Rename, export and document domain slider

axelboc:
Second step (after #1445) in addressing #1382 and #1377.

`ScaledSlider` becomes `DomainSlider` and is now exported and documented in the Storybook:

domain-slider


apps/storybook/src/DomainWidget.stories.tsx
```diff
@@ -4,12 +4,12 @@ import { COLOR_SCALE_TYPES } from '@h5web/shared';
...
 import { useState } from 'react';
 
-import Center from './decorators/Center';
+import DomainWidgetDecorator from './decorators/DomainWidgetDecorator';
 
 const meta = {
   title: 'Toolbar/DomainWidget',
   component: DomainWidget,
-  decorators: [Center],
+  decorators: [DomainWidgetDecorator],
```
axelboc, line 12: I noticed the domain widget stories required a minimum height for the domain tooltip to be usable. Instead of adding `min-height` to `Center`, which would have made it less generic, I decided to turn `Center` into a decorator dedicated to `DomainWidget` and stopped using it for the `Histogram` stories, which work fine without it.

axelboc:
The good news is that everything needed to properly use `DomainSlider` was already exported. I only had to put the pieces together like in `DomainWidget`.

It's quite nice that we now have an example of using the `DomainSlider` in complete isolation.

loichuder (approved):

PR #1455, opened 2023-07-19
Rename `DomainTooltip` to `DomainControls` and move tooltip logic to `DomainWidget`

axelboc:
Third step (after #1445 and #1448) in addressing #1382 and #1377.

In the first commit, `DomainTooltip` becomes `DomainControls` and the tooltip logic moves up into `DomainWidget`. In the second commit, I export `DomainControls` and document it in Storybook:

image

The final step will be to double check if any of the CSS customisation properties need to be renamed.

### DIFF
--- a/apps/storybook/src/DomainControls.stories.tsx
+++ b/apps/storybook/src/DomainControls.stories.tsx
@@ -1,0 +1,73 @@
+import { DomainControls } from '@h5web/lib';
+import type { Domain } from '@h5web/shared';
+import { useToggle } from '@react-hookz/web';
+import type { Meta, StoryObj } from '@storybook/react';
+import { useEffect, useState } from 'react';
+
+const meta = {
+  title: 'Toolbar/DomainControls',
+  component: DomainControls,
+  argTypes: {
+    sliderDomain: { control: false },
+    errors: { control: false },
+    isAutoMin: { control: false },
+    isAutoMax: { control: false },
+    isEditingMin: { control: false },
+    isEditingMax: { control: false },
+  },
+} satisfies Meta<typeof DomainControls>;
+
+export default meta;
+type Story = StoryObj<typeof DomainControls>;
+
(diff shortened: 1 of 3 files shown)


apps/storybook/src/DomainControls.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    useEffect(() => {
+      setSliderDomain(dataDomain);
+    }, [dataDomain]);
+
+    return (
+      <div style={{ maxWidth: '18rem' }}>
+        <DomainControls
+          sliderDomain={sliderDomain}
+          dataDomain={dataDomain}
+          errors={{}}
```
axelboc, line 42: I'm not bothering with the errors, since this would involve managing a `customDomain` and computing `visDomain` and `safeVisDomain` (which requires knowing the `scaleType`).

I decided to keep things simple and manage only a simple `sliderDomain` state.

packages/lib/src/toolbar/controls/DomainWidget/DomainWidget.tsx
```diff
@@ -106,48 +106,62 @@ function DomainWidget(props: Props) {
...
+        className={styles.tooltip}
+        role="dialog"
+        aria-label="Edit domain"
+        hidden={!hovered && !isEditing}
       >
-        {histogram && (
-          <Histogram
+        <div className={styles.tooltipInner}>
+          {histogram && (
+            <Histogram
```
axelboc, line 119: The diff is crap, but the move of the tooltip logic is fairly straightforward. The only thing worth mentioning is that the `Histogram` is no longer passed as `children` prop of `DomainTooltip` (now `DomainControls`).

It's `DomainWidget` that takes care of rendering `Histogram` directly in the body of the tooltip (`.tooltipInner`) and of laying it and `DomainControls` out appropriately with CSS.

PR #1456, opened 2023-07-19
Rename CSS vars following `DomainWidget` refactor

axelboc:
Fourth and final step in addressing https://github.com/silx-kit/h5web/issues/1377. Closes https://github.com/silx-kit/h5web/issues/1382.

A lot of the CSS variables (currently prefixed with `--h5-domainSlider-`) relate to the slider itself (track, thumbs, etc.), which is now named `DomainSlider` and exported, so those don't change.

Here are the variables that change:

- `--h5w-domainSlider-tooltip--bgColor` => `--h5w-domainWidget-popup--bgColor`
  - Note that I'm replacing the term "tooltip" with "popup" to avoid confusion with the tooltip we show on top of visualisations. It's also kind of the wrong term in the context of the domain widget, since `DomainControls` is quite complex and the popup can be pinned open. It's more of a popover, really, but the term popup is a bit more generic and can apply to all toolbar popups (like the overflow menu). It's also consistent with the ARIA attribute `aria-haspopup`.
  - Also, it turns out that `--h5w-domainSlider-tooltip--bgColor` was actually applied to all toolbar "tooltips" in `Toolbar.module.css`. I introduce a new variable to fix this: `--h5w-toolbar-popup--bgColor` and use it as default for `--h5w-domainWidget-popup--bgColor`.
- `--h5w-domainSlider-tooltip--colorAlt` => `--h5w-domainControls--colorAlt`
- `--h5w-domainSlider-boundInput--shadowColor` => `--h5w-domainControls-boundInput--shadowColor`
- `--h5w-domainSlider-boundInput-focus--shadowColor` => `--h5w-domainControls-boundInput-focus--shadowColor`
- `--h5w-domainSlider-boundInput-editing--bgColor` => `--h5w-domainControls-boundInput-editing--bgColor`
- `--h5w-domainSlider-boundInput-editing--borderColor` => `--h5w-domainControls-boundInput-editing--borderColor`
- `--h5w-domainSlider-histogram--bgColor` => `--h5w-histogram--bgColor`
- `--h5w-domainSlider-histogram--color` => `--h5w-histogram--color`
- `--h5w-domainSlider-histogram--color` => `--h5w-histogram-marker--color`


loichuder (approved):
> Note that I'm replacing the term "tooltip" with "popup" to avoid confusion with the tooltip we show on top of visualisations.

Good call :+1:

PR #1458, opened 2023-07-25
Skip linting and type checking when building for production to save time

axelboc:
Linting and type checking in the demo project can be run separately from the production build (and potentially in parallel) with `pnpm lint:eslint` and `pnpm lint:tsc`. This is already the case in the CI (we build the demo before running the E2E tests), so we were basically doubling up on the work.


axelboc:
This seems to shave off roughly 20s off the production build time.

PR #1462, opened 2023-07-26
Improve `RawVis` fallback message when dataset is too big to display

### LINKED ISSUE #917: Large mixed type dataset logged to the console can be missleading
### Describe the bug

Too large to display mixed type dataset leads to the message: `The dataset is too big to be displayed and was logged to the console instead.`
This seems missleading to me, since not everyone will look into the dev tools to open the console.

### To Reproduce

1. Go to https://h5web.panosc.eu/?file=epics.h5
2. Click on `arrays` -> `Vdata with mixed types`
4. Message is: `The dataset is too big to be displayed and was logged to the console instead.`

### Expected behaviour

To avoid missleading message:
- the data could not be displayed at all: `The dataset is too big to be displayed.`, or
- the data could be displayed on-demand: `This dataset is big,` [`display it`](), if possible with a foldable json viewer.

### Context

- H5Web context: demo

axelboc:
This is a proposal to fix #917. I'm basically just improving the fallback message to clarify which console it's talking about and how users can access it:

image

### DIFF
--- a/packages/app/src/__tests__/CorePack.test.tsx
+++ b/packages/app/src/__tests__/CorePack.test.tsx
@@ -21,7 +21,7 @@ test('log raw dataset to console if too large', async () => {
   const logSpy = mockConsoleMethod('log');
   await renderApp('/entities/raw_large');
 
-  await expect(screen.findByText(/dataset is too big/)).resolves.toBeVisible();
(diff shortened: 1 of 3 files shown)


loichuder (approved):
Very nice !

t20100 (approved):
Nice!

PR #1463, opened 2023-07-26
Allow canceling select-to-zoom interaction with right click

### LINKED ISSUE #1370: Cancel ZoomMode interaction with right mouse button click

User interaction convenience

During the zoom interaction i can click on other buttons.

For now if i do the interaction with the left button, the right button will open a menu, which is not part of the user task at this stage.

The opposite button is usually used to cancel the user task (or any other buttons).

Could it be possible to implement that behaviour?

Thx.

axelboc:
Fix #1370 -- this is quite a neat interaction, actually.

Couple of notes:

- Pressing the context menu key on the keyboard also cancels the interaction.
- Firefox does not fire the [`contextmenu` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event) at all when <kbd>Shift</kbd> is pressed. So with the proposed solution, a select-to-zoom interaction configured with the `Shift` modifier key cannot be cancelled with right click.

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -1,5 +1,6 @@
 import { assertDefined } from '@h5web/shared';
 import {
+  useEventListener,
   useKeyboardEvent,
   usePrevious,
   useRafState,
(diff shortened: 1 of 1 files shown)


loichuder (approved):
Neat, indeed :+1:

axelboc:
Bug fix: https://github.com/silx-kit/h5web/commit/5da1e6a480b28412947eb12d113323975ba2c04f

PR #1464, opened 2023-07-27
Allow overriding default validation logic of `SelectToZoom`

### LINKED ISSUE #1407: There is no way to zoom on a small slice
Sounds like the contraint to avoid zoom on click is taking the constraint on x and y independently.

As result there is no way to zoom on a small slice.

image

The 2 dimensions should be used together (distance or distance^2 or manhattan distance).

Also it would be much better to apply this constrain only at the start of the interaction.
If i create create a big rect and reduce it progressively during the interaction, the constraint should not be applied.

axelboc:
Fix #1407


loichuder:
If it is only to allow arbitrary zoom box sizes, is the `minZoom` prop not enough ? Putting it to `0` would essentially disable the size check.

However, this is the right way if this is to allow more involved validation by consumers.

PR #1465, opened 2023-08-09
Refactor `Html` with portal

axelboc:
### Context

`Html` allows rendering HTML elements from inside React Three Fiber's renderer. It works by appending a `div` to the DOM and rendering its children into it with `ReactDOM.render()`. The `div` is appended to `r3fRoot` by default, to `canvasWrapper` when `overflowCanvas` is set, or to a custom container when `container` is set.

The problem currently is that the children of `Html` always end up wrapped in an extra `div`. This is a bit annoying when debugging the DOM and it's quite limiting as the extra `div` can make styling cumbersome and can lead to invalid HTML (e.g. `div > svg`). We worked around this extra `div` before in `FloatingControl` and `SvgElement` by rendering the children into another container through a portal. 

The `container` prop has other issues, like the fact that it has no effect when `overflowCanvas` is set or that it's buggy with dynamically rendered containers (i.e. async refs) – the children get rendered into `r3fRoot` first until the container becomes available.

### Proposed changes

In this PR, I refactor `Html` to render its children through a portal, thus removing the extraneous `div`. The children now end up directly under `r3fRoot` or, if `overflowCanvas` is set, under `canvasWrapper`.

To avoid the pitfall of dynamically rendered containers and keep `Html` as light as possible, I remove the `container` prop altogether. Equivalent behaviour can be achieved by the consumer of the `Html` component with a portal (like in `FloatingControl` and `SvgElement`). Yes, in this case, it's basically portal-ception, since a portal is rendered inside another portal, but my quick tests and research don't seem to point to any negative performance implications.

So not much changes in practice:

- if an element needs to go into `r3fRoot` or `canvasWrapper`, we still use `<Html>` or `<Html overflowCanvas>`;
- if multiple elements need to go into the same custom container, we still use `<Html>{createPortal(..., container)}</Html>`.

I took the opportunity to improve the documentation of `Html` in the Storybook:

image

image

image

### DIFF
--- a/packages/lib/src/vis/shared/Html.tsx
+++ b/packages/lib/src/vis/shared/Html.tsx
@@ -1,48 +1,44 @@
+import { assertNonNull } from '@h5web/shared';
 import { useThree } from '@react-three/fiber';
-import type { ReactNode } from 'react';
+import type { PropsWithChildren } from 'react';
 import { useLayoutEffect, useState } from 'react';
-import ReactDOM from 'react-dom';
+import ReactDOM, { createPortal } from 'react-dom';
 
 interface Props {
-  overflowCanvas?: boolean; // allow children to overflow above axes
-  container?: HTMLElement;
-  children?: ReactNode;
+  overflowCanvas?: boolean;
 }
 
-function Html(props: Props) {
-  const {
-    overflowCanvas = false,
-    container: customContainer,
-    children,
-  } = props;
+function Html(props: PropsWithChildren<Props>) {
+  const { overflowCanvas = false, children } = props;
 
   const r3fRoot = useThree((state) => state.gl.domElement.parentElement);
-  const canvasWrapper = r3fRoot?.parentElement;
+  assertNonNull(r3fRoot);
 
-  // Choose DOM container to which to append `renderTarget`
-  // (with `canvasWrapper`, `Html` children are allowed to overflow above the axes)
-  const container =
-    customContainer || (overflowCanvas ? canvasWrapper : r3fRoot);
+  const canvasWrapper = r3fRoot.parentElement;
+  assertNonNull(canvasWrapper);
 
-  const [renderTarget] = useState(() => document.createElement('div'));
+  const portalContainer = overflowCanvas ? canvasWrapper : r3fRoot;
+
+  const [renderContainer] = useState(() => {
+    const div = document.createElement('div');
+    div.setAttribute('hidden', '');
+    return div;
+  });
 
   useLayoutEffect(() => {
-    ReactDOM.render(<>{children}</>, renderTarget); // eslint-disable-line react/jsx-no-useless-fragment
-  }, [children, renderTarget]);
+    ReactDOM.render(createPortal(children, portalContainer), renderContainer);
+  }, [children, portalContainer, renderContainer]);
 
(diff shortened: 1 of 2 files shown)


packages/lib/src/vis/shared/Html.tsx
```diff
@@ -1,48 +1,42 @@
...
 
-  const [renderTarget] = useState(() => document.createElement('div'));
+  const portalContainer = overflowCanvas ? canvasWrapper : r3fRoot;
 
-  useLayoutEffect(() => {
-    ReactDOM.render(<>{children}</>, renderTarget); // eslint-disable-line react/jsx-no-useless-fragment
-  }, [children, renderTarget]);
+  const [renderContainer] = useState(() => {
+    const div = document.createElement('div');
+    div.setAttribute('hidden', '');
```
axelboc, line 24: While I'm at it, I hide the `div` that acts as the `ReactDOM.render` container, since it is now empty.

packages/lib/src/vis/shared/Html.tsx
```diff
@@ -1,48 +1,42 @@
...
+    return div;
+  });
 
   useLayoutEffect(() => {
-    return () => {
-      ReactDOM.unmountComponentAtNode(renderTarget);
```
axelboc, line 35: Turns out that having this in a separate effect with only `renderTarget` as dep was a bug. `renderTarget` never changes—it's `container` that changes—so the first clean-up function was not always called when it should have been.

The bug is gone now, since we always append `renderTarget` (now `renderContainer`) into `r3fRoot`, which never changes (so both clean-up functions would get called when `Html` unmounts), but it's still cleaner to have them in a single clean-up function.

packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -49,8 +49,10 @@ function VisCanvas(props: PropsWithChildren<Props>) {
       })
     : NO_OFFSETS;
 
-  const [svgOverlay, setSvgOverlay] = useState<SVGSVGElement>();
-  const [floatingToolbar, setFloatingToolbar] = useState<HTMLDivElement>();
+  const [svgOverlay, setSvgOverlay] = useState<SVGSVGElement | null>(null);
+  const [floatingToolbar, setFloatingToolbar] = useState<HTMLDivElement | null>(
+    null,
+  );
```
axelboc, line 55: It's simpler for `svgOverlay` and `floatingToolbar` to never be `undefined` (see below). Also, I find that `null` better conveys the fact that the containers are there but we just can't get their refs synchronously.

axelboc:
One alternative I considered to solve the problem of conditionally-rendered containers is to keep the `container` prop but to change its type to `HTMLElement | null | undefined`. This would allow passing `null` to indicate that a `container` is expected but not yet available, thus allowing `Html` to not render its children until the container becomes available.

The downside of this solution is that it doesn't actually prevent consumers from hitting the initial problem if they mistakenly store their container in a state with type `HTMLElement | undefined`... By forcing them to render their own portal, they are required to check that their container exists before calling `createPortal`, so it doesn't matter if the initial value is `undefined` or `null`.

loichuder (approved):
Good thinking :+1:

PR #1466, opened 2023-08-09
Clean up and improve `Annotation` stories

axelboc:
While working on refactoring `Html` in #1465 , I noticed that the `Annotation` stories were in need of a good clean-up.

I took full advantage of Storybook by splitting each annotation example into its own story, which means that the controls now work as expected and the props are properly listed in the _Docs_ page.

I also:

- removed the SVG story, since `Annotation` is no longer the recommended way to render SVG elements;
- added a story to document `overflowCanvas`.

This actually uncovered a bug: annotations that overflow the canvas are not positioned properly, since they are positioned relatively to the `canvasWrapper` element, which includes the axes. I have an idea to fix this, but I'll continue working on declarative axes first.

### DIFF
--- a/apps/storybook/src/Annotation.stories.tsx
+++ b/apps/storybook/src/Annotation.stories.tsx
@@ -15,7 +15,11 @@ import { formatCoord } from './utils';
 
 const meta = {
   title: 'Building Blocks/Annotation',
-  parameters: { layout: 'fullscreen' },
+  component: Annotation,
+  parameters: {
+    layout: 'fullscreen',
+    controls: { sort: 'requiredFirst' },
+  },
   decorators: [
     (Story) => (
       <VisCanvas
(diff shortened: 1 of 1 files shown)


apps/storybook/src/Annotation.stories.tsx
```diff
@@ -28,94 +32,101 @@ const meta = {
...
+    <PointerTracker>
+      {(x, y) => (
+        <Annotation
+          {...args}
+          x={x + 0.5} // slight offset from pointer
+          y={y - 0.5}
+          style={{ whiteSpace: 'nowrap' }}
+        >{`x=${formatCoord(x)}, y=${formatCoord(y)}`}</Annotation>
+      )}
+    </PointerTracker>
```
axelboc, line 120: Just added `{...args}`

loichuder:
image

Just a thought: it would be nice to highlight the `(10, 16)` point above with a red cross for example.

loichuder (approved):

axelboc:
Good call!

image
image

PR #1467, opened 2023-08-10
Fix position of overflowing annotations by refactoring `VisCanvas`

axelboc:
As mentioned in #1466, annotations with `overflowCanvas` do not work as expected: they are positioned in relation to `canvasWrapper`, the root element rendered by `VisCanvas`, which defines padding for the axes.

In this PR, I remedy this problem by refactoring `VisCanvas` once again in order to have a `canvasWrapper` element with the same size as the `canvas` that, unlike `r3fRoot`, does not hide overflow.

The root element rendered by `VisCanvas` is now called `visCanvas` (this may change in later refactors) and no longer has any padding. Instead, it takes on the role of `axisSystem` by setting up the CSS grid that lays everything out:

- `canvasWrapper` is now a child of `visCanvas` and positioned with `grid-area`.
- The axes and the grids are rendered as direct children of `visCanvas` and also positioned with `grid-area`.

So instead of:

```
.canvasWrapper (padding)
  .r3fRoot (hidden overflow)
    canvas
    .svgOverlay
    .floatingToolbar
  .axisSystem
    .axis
    .grid
    .axis
    .grid
```

... we now have:

```
.visCanvas (no padding)
  .canvasWrapper (visible overflow)
    .r3fRoot (hidden overflow)
      canvas
      .svgOverlay
      .floatingToolbar
  .axis
  .grid
  .axis
  .grid
``` 

I had tried similar refactors before but always ran into overflow issues. The key that unlocked it this time is the use of `minmax(0, 1fr)` instead of `1fr` for sizing the grid area reserved for the canvas. This has a similar effect to `min-width: 0` and `min-height: 0` when using Flexbox. See https://css-tricks.com/preventing-a-grid-blowout/

Another important reason this layout works is because CSS Grid allows positioning multiple grid items into the same grid area. The items then behave as if they were positioned absolutely on top of one another. In our case, the axis grids are positioned in the same CSS Grid area as `canvasWrapper`. I feel like this will enable further refactors.


packages/lib/src/vis/shared/AxisSystem.module.css
```diff
@@ -1,21 +1,8 @@
-.axisSystem {
```
axelboc, line 1: This container no longer exists. The CSS Grid properties move to `.visCanvas` in `VisCanvas.module.css`. `pointer-events` and `z-index` move directly on `.axis` and `.grid`.

packages/lib/src/vis/shared/AxisSystem.module.css
```diff
@@ -52,20 +40,8 @@
   text-anchor: middle;
 }
 
-.title {
```
axelboc, line 55: The title is now rendered directly under `.visCanvas`, so I've moved the style to `VisCanvas.module.css`.

packages/lib/src/vis/shared/AxisSystem.tsx
```diff
@@ -1,52 +1,55 @@
...
-          config={ordinateConfig}
-          domain={visibleDomains.yVisibleDomain}
-          canvasSize={canvasSize}
-          offset={axisOffsets.left}
-          showAxis={showAxes}
-          flipAxis
-        />
-      </div>
+    <Html>
+      {createPortal(
```
axelboc, line 31: Using the same technique as in `SvgElement` and `FloatingControl`.

packages/lib/src/vis/shared/AxisSystem.tsx
```diff
@@ -1,52 +1,55 @@
...
   const visibleDomains = useCameraState(getVisibleDomains, [getVisibleDomains]);
 
   return (
-    // Append to `canvasWrapper` instead of `r3fRoot`
-    <Html overflowCanvas>
-      <div
-        className={styles.axisSystem}
-        style={{
-          gridTemplateColumns: `${axisOffsets.left}px 1fr ${axisOffsets.right}px`,
-          gridTemplateRows: `${axisOffsets.top}px 1fr ${axisOffsets.bottom}px`,
```
axelboc, line 28: Moved to `VisCanvas`.

packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -54,34 +54,31 @@ function VisCanvas(props: PropsWithChildren<Props>) {
...
-          <InteractionsProvider>{children}</InteractionsProvider>
-          <ViewportCenterer />
-          <RatioEnforcer />
-          {raycasterThreshold !== undefined && (
-            <ThresholdAdjuster value={raycasterThreshold} />
-          )}
-        </VisCanvasProvider>
+      {showAxes && title && <p className={styles.title}>{title}</p>}
+
+      <div className={styles.canvasWrapper}>
```
axelboc, line 65: This is the new wrapper. Couldn't think of a better name for now. With the CSS Grid on `visCanvas`, things might change a bit more and this wrapper might in fact be removed.

cypress/snapshots/app.cy.ts/edit_domain.snap.png
axelboc, line 1: For some reason, a couple of pixels change in the shadows of the toolbar controls (e.g. slider thumbs). 🤷

packages/lib/src/vis/shared/AxisSystem.tsx
```diff
@@ -1,52 +1,55 @@
...
 
 function AxisSystem(props: Props) {
-  const { axisOffsets, title, showAxes } = props;
+  const { axisOffsets, showAxes } = props;
   const { canvasSize, abscissaConfig, ordinateConfig, getVisibleDomains } =
     useVisCanvasContext();
 
+  const visCanvas = useThree(
+    (state) => state.gl.domElement.parentElement?.parentElement?.parentElement,
+  );
```
loichuder, line 23: That seems a bit brittle. 

Could we imaging exposing a ref to `visCanvas` through the `VisCanvas` ?

packages/lib/src/vis/shared/VisCanvas.tsx
loichuder: `r3fRootContainer` ?

loichuder (approved):
:+1:

packages/lib/src/vis/shared/AxisSystem.tsx
axelboc: Good call. I'll do it in one of my later PRs to avoid rebasing conflicts.

packages/lib/src/vis/shared/VisCanvas.tsx
loichuder: Sure. It was a wild suggestion

axelboc: #1479

PR #1471, opened 2023-08-11
Move SVG overlay and floating toolbar out of `r3fRoot`

axelboc:
#1467 allows further simplification:

- `svgOverlay` can be rendered directly into `canvasWrapper` instead of `r3fRoot`. This opens up an opportunity to let SVG elements overflow in the future; however, for now, the overflow remains hidden. It also removes the need to render the overlay with `Html`, which most likely provides a very small performance gain on initial render.
- `floatingToolbar` can be rendered into `visCanvas` and positioned with CSS Grid properties. This paves the way to https://github.com/silx-kit/h5web/issues/1388#issuecomment-1673269269 and also removes the need for `Html`.


packages/lib/src/vis/shared/VisCanvas.module.css
```diff
@@ -68,10 +69,9 @@
...
 .floatingToolbar {
-  position: absolute;
-  right: 0;
-  bottom: 0;
-  display: flex;
+  grid-area: canvas;
+  justify-self: end;
+  align-self: end;
   z-index: var(--h5w-zi-floatingToolbar);
-  pointer-events: auto;
```
axelboc, line 76: No longer needed, since `floatinToolbar` is no longer inside `r3fRoot`/`canvasWrapper`, so `pointer-events` are never disabled.

packages/lib/src/vis/shared/VisCanvas.module.css
```diff
@@ -68,10 +69,9 @@
 }
 
 .floatingToolbar {
-  position: absolute;
-  right: 0;
-  bottom: 0;
-  display: flex;
+  grid-area: canvas;
+  justify-self: end;
+  align-self: end;
```
axelboc, line 74: Taking full advantage of CSS Grid 🎉

packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -49,8 +48,10 @@ function VisCanvas(props: PropsWithChildren<Props>) {
       })
     : NO_OFFSETS;
 
-  const [svgOverlay, setSvgOverlay] = useState<SVGSVGElement>();
-  const [floatingToolbar, setFloatingToolbar] = useState<HTMLDivElement>();
+  const [svgOverlay, setSvgOverlay] = useState<SVGSVGElement | null>(null);
+  const [floatingToolbar, setFloatingToolbar] = useState<HTMLDivElement | null>(
+    null,
+  );
```
axelboc, line 54: Brought from #1465:



This simplifies setting the `ref`s of `svgOverlay` and `floatingToolbar` below. Also, I find that `null` better conveys the fact that the containers are there but we just can't get their refs synchronously.

packages/lib/src/vis/shared/VisCanvas.tsx
```diff
@@ -79,21 +80,16 @@ function VisCanvas(props: PropsWithChildren<Props>) {
...
-              ref={(elem) => setFloatingToolbar(elem || undefined)}
-              className={styles.floatingToolbar}
-            />
-          </Html>
         </R3FCanvas>
+
+        <svg
+          ref={setSvgOverlay}
+          className={styles.svgOverlay}
+          overflow="hidden"
```
axelboc, line 88: User agents set `overflow: hidden` on `svg` elements that are not document roots ([source](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/overflow); this is why we have to set `overflow: visible` on the axes and the grid), but I prefer it to be explicit here.

cypress/snapshots/app.cy.ts/auxspectrum.snap.png
axelboc, line 1: Just a random change to the painting order of the error bars and caps materials.

loichuder (approved):
Great !

PR #1474, opened 2023-08-16
Revert workaround of bundling in `react-keyed-flatten-children`

axelboc:
`react-keyed-flatten-children` [now has an ESM bundle](https://github.com/grrowl/react-keyed-flatten-children/blob/master/CHANGELOG.md), which means we should be able to revert the workaround that consisted in inlining it in the lib's output bundle (cf. #914). It is now marked as an `external` in the Vite config, just like every other prod dependency.


axelboc:
`react-keyed-flatten-children` now requires React 18. I'm not seeing any problem in development, but probably safer to wait until #1119 is merged.

PR #1476, opened 2023-08-18
Fix MouseButton export

PeterC-DLS:


axelboc:
On the need to always open a PR :sweat: – cheers for that. FYI, I'll push a refactor of the internals before midday, to avoid the need to call `getModifierKeyArray`.

axelboc (approved):

PeterC-DLS:
You're welcome!

PR #1475, opened 2023-08-17
Export bits from interactions

PeterC-DLS:
This allows for an external implementation of a multiclick selection tool

PR #1477, opened 2023-08-18
Simplify and document usage of `useInteraction` and `useModifierKeyPressed`

axelboc:
Following #1475, I'm documenting the newly exported `useInteraction` and `useModifierKeyPressed` hooks.

I'm also taking the opportunity to do a bit of a refactoring now, to avoid a breaking change down the line:

- `InteractionEntry` (the second parameter of `useInteraction`) becomes `InteractionConfig`
- We no longer need to cast the `modifierKey` prop to an array before calling `useInteraction` and `useMdifierKeyPressed` (thus removing the need for `getModifierKeyArray`)
- I'm introducing an `Interaction` class internally to encapsulate the casting of the `modiferKeys` and `buttons` arrays and the logic of whether an interaction "matches" a mouse event.

### DIFF
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -124,30 +124,20 @@ export {
   useInteraction,
   useModifierKeyPressed,
 } from './interactions/hooks';
-export { MouseButton } from './interactions/models';
-export { getModifierKeyArray } from './interactions/utils';
 export { default as Box } from './interactions/box';
 
+// Constants
 export { COLOR_SCALE_TYPES, AXIS_SCALE_TYPES } from '@h5web/shared';
-
-// Models
 export { INTERPOLATORS } from './vis/heatmap/interpolators';
+
+// Enums
 export { ScaleType } from '@h5web/shared';
 export { CurveType, GlyphType } from './vis/line/models';
 export { ImageType } from './vis/rgb/models';
 export { Notation } from './vis/matrix/models';
+export { MouseButton } from './interactions/models';
 
-export type {
-  ModifierKey,
-  Rect,
-  Selection,
-  CanvasEvent,
-  CanvasEventCallbacks,
-  InteractionInfo,
-  InteractionEntry,
-  CommonInteractionProps,
-} from './interactions/models';
-
+// Models
 export type {
   Domain,
   VisibleDomains,
@@ -175,6 +165,17 @@ export type {
 
 export type { D3Interpolator, ColorMap } from './vis/heatmap/models';
 export type { ScatterAxisParams } from './vis/scatter/models';
+
+export type {
+  ModifierKey,
+  Rect,
+  Selection,
+  CanvasEvent,
+  CanvasEventCallbacks,
+  InteractionInfo,
+  InteractionConfig,
+  CommonInteractionProps,
+} from './interactions/models';
 
 // Mock data and utilities
 export {
--- a/packages/lib/src/interactions/InteractionsProvider.tsx
+++ b/packages/lib/src/interactions/InteractionsProvider.tsx
@@ -1,16 +1,13 @@
 import type { ReactNode } from 'react';
 import { createContext, useCallback, useContext, useState } from 'react';
 
-import type { InteractionEntry, ModifierKey, MouseButton } from './models';
+import { Interaction } from './interaction';
+import type { InteractionConfig } from './models';
 
 export interface InteractionsContextValue {
-  registerInteraction: (id: string, value: InteractionEntry) => void;
+  registerInteraction: (id: string, config: InteractionConfig) => void;
   unregisterInteraction: (id: string) => void;
   shouldInteract: (id: string, event: MouseEvent) => boolean;
-}
-
-interface MapEntry extends InteractionEntry {
-  id: string;
 }
 
 const InteractionsContext = createContext({} as InteractionsContextValue);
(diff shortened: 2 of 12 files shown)


packages/lib/src/index.ts
```diff
@@ -124,30 +124,20 @@ export {
...
-export { getModifierKeyArray } from './interactions/utils';
 export { default as Box } from './interactions/box';
 
+// Constants
 export { COLOR_SCALE_TYPES, AXIS_SCALE_TYPES } from '@h5web/shared';
-
-// Models
 export { INTERPOLATORS } from './vis/heatmap/interpolators';
+
+// Enums
```
axelboc, line 133: Separating out enums to avoid falling into the trap of exporting them with `export type`.

packages/lib/src/interactions/InteractionsProvider.tsx
```diff
@@ -1,18 +1,15 @@
 import type { ReactNode } from 'react';
 import { createContext, useCallback, useContext, useState } from 'react';
 
-import type { InteractionEntry, ModifierKey, MouseButton } from './models';
+import { Interaction } from './interaction';
+import type { InteractionConfig } from './models';
 
 export interface InteractionsContextValue {
-  registerInteraction: (id: string, value: InteractionEntry) => void;
```
axelboc, line 7: The second parameter is no longer stored "as is" in the interactions map below. It is now a config object used to construct an instance of the `Interaction` class.

packages/lib/src/interactions/InteractionsProvider.tsx
```diff
@@ -45,28 +42,12 @@ function InteractionsProvider(props: { children: ReactNode }) {
...
-      }
-
       if (!interactionMap.has(interactionId)) {
         throw new Error(`Interaction ${interactionId} is not registered`);
       }
 
       const matchingInteractions = registeredInteractions.filter(
-        ({ modifierKeys: keys, button, disabled }) =>
-          !disabled && isButtonPressed(button) && areKeysPressed(keys),
+        (interaction) => interaction.matches(event),
```
axelboc, line 50: The logic of `isButtonPressed` and `areKeysPressed` is now encapsulated in the `Interaction` class.

packages/lib/src/interactions/InteractionsProvider.tsx
```diff
@@ -77,13 +58,12 @@ function InteractionsProvider(props: { children: ReactNode }) {
         return matchingInteractions[0].id === interactionId;
       }
 
-      // If conflicting interactions, the one with the most modifier keys take precedence
-      matchingInteractions.sort(
-        (a, b) => b.modifierKeys.length - a.modifierKeys.length,
+      // If conflicting interactions, the one with the most modifier keys takes precedence
+      const maxKeysInteraction = matchingInteractions.reduce((acc, next) =>
+        next.modifierKeys.length > acc.modifierKeys.length ? next : acc,
```
axelboc, line 63: Reducing the array is a bit more efficient than sorting it.

packages/lib/src/interactions/Pan.tsx
```diff
@@ -25,18 +24,13 @@ function Pan(props: Props) {
...
-    disabled,
-  });
+  const shouldInteract = useInteraction(id, { button, modifierKey, disabled });
 
   const camera = useThree((state) => state.camera);
   const moveCameraTo = useMoveCameraTo();
 
   const startOffsetPosition = useRef<Vector3>(); // `useRef` to avoid re-renders
-  const isModifierKeyPressed = useModifierKeyPressed(modifierKeys);
+  const isModifierKeyPressed = useModifierKeyPressed(modifierKey);
```
axelboc, line 33: The array casting is now done internally, to avoid having to call and export `getModifierKeyArray`.

packages/lib/src/interactions/interaction.ts
```diff
@@ -0,0 +1,44 @@
+import { castArray } from 'lodash';
+import type { ModifierKey } from 'react';
+
+import type { InteractionConfig } from './models';
+import { MouseButton } from './models';
+
+export class Interaction {
+  public readonly buttons: MouseButton[];
+  public readonly modifierKeys: ModifierKey[];
+  public readonly isWheel: boolean;
```
axelboc, line 10: Note that I'm no longer storing `'Wheel'` in `buttons` like it was done in `InteractionsProvider`. It makes the constructor a bit more verbose, but it clarifies the matching logic a little.

The `InteractionConfig` still requires providing `'Wheel'` through the `button` prop for convenience but by design, it still disallows `[MouseButton.Left, 'Wheel']`, for instance.

loichuder (approved):
Nice !

packages/lib/src/interactions/interaction.ts
```diff
@@ -0,0 +1,44 @@
...
+      this.buttons = castArray(button);
+      this.isWheel = false;
+    }
+
+    this.modifierKeys = castArray(modifierKey);
+    this.isEnabled = !disabled;
+  }
+
+  public matches(event: MouseEvent): boolean {
+    return (
```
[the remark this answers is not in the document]
axelboc: Yeah on the other hand, early returns require inverting the conditions ... except the last one, which, in this case, I'm not hugely fond of:

```ts
!this.isEnabled { return false; } 
!this.modifierKeys.every((key) => event.getModifierState(key)) { return false; }
return this.modifierKeys.every((key) => event.getModifierState(key));
```

PR #1479, opened 2023-08-21
Inject `visCanvas` through context and rename `canvasWrapper` to `canvasArea`

axelboc:
Follow-up to #1467

### DIFF
--- a/apps/storybook/src/Html.stories.tsx
+++ b/apps/storybook/src/Html.stories.tsx
@@ -171,6 +171,8 @@ export const OverflowCanvas = {
 
 export const Portal = {
   render: () => {
+    const { visCanvas } = useVisCanvasContext();
+
     const [containerMounted, toggleContainer] = useToggle(true);
     const [customContainer, setCustomContainer] =
       useState<HTMLDivElement | null>(null);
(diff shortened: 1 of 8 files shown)


apps/storybook/src/Html.stories.tsx
```diff
@@ -201,10 +203,44 @@ export const Portal = {
...
+                  , the root <code>div</code> rendered by <code>VisCanvas</code>
+                  , which includes the axes. You can then take advantage of the
+                  CSS Grid defined on <code>visCanvas</code> to position your
+                  elements.
+                </p>
+                <p>
+                  However, note that your elements will interfere with canvas
+                  interactions (zoom, pan, etc.), since pointer events will no
+                  longer be able to bubble up to <code>canvasArea</code>. You
+                  can remedy this with <code>pointer-events: none</code>.
```
axelboc, line 234: While I'm at it, adding a bit more documentation here hinting at how interactions work under the hood.

packages/lib/src/interactions/hooks.ts
```diff
@@ -62,15 +62,15 @@ export function useZoomOnSelection() {
 }
 
 function useWheelCapture() {
-  const canvasWrapper = useCanvasWrapper();
+  const canvasArea = useCanvasArea();
```
loichuder, line 65: `useCanvasArea` still relies on the DOM structure. Would it be possible to leverage the work of this PR to avoid this ?

For example, putting the ref on `canvasArea` instead of `visCanvas` and exposing `visCanvas` as `canvasArea.parentElement` ? Or the other way round ?

axelboc: I remember now why I was retrieving `visCanvas` via `useThree` instead of injecting it through context. For every state we add to `VisCanvas` to store a DOM element in the context, we basically add a new re-render on initial mount.

R3F probably re-renders once when it gets a reference to the `canvas` element. By then, we know that the canvas' parent elements are also in the DOM, so by accessing them with `parentElement`, we save a lot of processing.

I do agree that this is not very elegant, though. I'll have a bit more of a think.

axelboc: What's not super elegant is to have custom hooks for accessing specific parents of the `canvas`. Maybe I can still call `gl.domElement.parentElement[.parentElement]` to retrieve the elements, but do so in `VisCanvasProvider` instead of in custom hooks. No `useState`, no extra renders, no custom hooks.

loichuder: > Maybe I can still call gl.domElement.parentElement[.parentElement] to retrieve the elements, but do so in VisCanvasProvider instead of in custom hooks. No useState, no extra renders, no custom hooks.

Yeah, sounds good to me :+1:

axelboc: > Hm. Is it mandatory to use a state rather than a ref ?

Yeah, because the `ref` is `null` until React actually renders the DOM element. It'd be fine if we used the reference to the DOM element inside a `useEffect`, but we use it during render (with `createPortal(<></>, targetElem)`)

loichuder (approved):
Looks good to me. At least, the dependency on the DOM architecture is on a single place.

PR #1480, opened 2023-08-21
Minor monorepo tweaks

axelboc:
Following the set-up of a monorepo based on H5Web's, I've identified a couple of opportunities to tweak things a bit.


loichuder:
Great :+1:

PR #1481, opened 2023-08-25
Improve hook to register events on canvas

axelboc:
I'm rethinking `useCanvasEvents` a bit. The hook is now called `useCanvasEvent` (singular) and allows registering only one event listener at a time. This brings the API closer to that of `useEventListener` from react-hookz and avoids registering event listeners that are not actually used.

Additional benefits of the new implementation:

- It allows more events to be registered, not just `pointerdown`, `pointerup`, `pointermove` and `wheel` — so basically any event that has `offsetX` and `offsetY` (i.e. any event that extends `MouseEvent`, including `WheelEvent`, `PointerEvent`, `DragEvent`).
- It uses `useSyncedRef` so consumers don't have to wrap their listeners in `useCallback` — this simplifies a bunch of code.
- It allows overriding the options passed to `addEventListener`, notably `passive`, which might be useful at some point notably for `wheel` events (currently, the only place where we do need to configure `passive` is in `useWheelCapture` but it doesn't rely on `useCanvasEvent`).

### DIFF
--- a/packages/lib/src/interactions/YAxisZoom.tsx
+++ b/packages/lib/src/interactions/YAxisZoom.tsx
@@ -1,5 +1,10 @@
 import { useVisCanvasContext } from '../vis/shared/VisCanvasProvider';
-import { useCanvasEvents, useInteraction, useZoomOnWheel } from './hooks';
+import {
+  useCanvasEvent,
+  useInteraction,
+  useWheelCapture,
+  useZoomOnWheel,
+} from './hooks';
 import type { CommonInteractionProps } from './models';
 
 type Props = CommonInteractionProps;
@@ -19,7 +24,8 @@ function YAxisZoom(props: Props) {
     y: shouldInteract(sourceEvent),
   });
 
-  useCanvasEvents({ onWheel: useZoomOnWheel(isZoomAllowed) });
+  useWheelCapture();
+  useCanvasEvent('wheel', useZoomOnWheel(isZoomAllowed));
 
   return null;
 }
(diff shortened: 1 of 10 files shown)


packages/lib/src/interactions/YAxisZoom.tsx
```diff
@@ -19,7 +24,8 @@ function YAxisZoom(props: Props) {
     y: shouldInteract(sourceEvent),
   });
 
-  useCanvasEvents({ onWheel: useZoomOnWheel(isZoomAllowed) });
+  useWheelCapture();
```
axelboc, line 27: Moving this call out of `useZoomOnWheel` to be more explicit.

packages/lib/src/interactions/models.ts
```diff
@@ -52,3 +45,9 @@ export interface UseDragState {
   isDragging: boolean;
   startDrag: (evt: PointerEvent) => void;
 }
+
+export type MouseEventName = {
+  [Key in keyof GlobalEventHandlersEventMap]: GlobalEventHandlersEventMap[Key] extends MouseEvent
+    ? Key
+    : never;
+}[keyof GlobalEventHandlersEventMap];
```
axelboc, line 53: I did not come up with this... https://medium.com/dailyjs/typescript-create-a-condition-based-subset-types-9d902cea5b8c

Basically you go from `{ pointerdown: PointerEvent, resize: ResizeEvent, ... }`, to `{ pointerdown: PointerEvent, resize: never, ... }`, to `'pointerdown' | ...`.

PR #1485, opened 2023-09-01
Add JSON export to Raw visualization

axelboc:
The `Raw` visualization now supports exporting datasets to JSON when using the h5grove, h5wasm and mock providers.

This provides a much better solution to issue #917 than PR #1462 . As such, large raw datasets that can't be rendered on the screen are no longer logged to the console.

image

image

Concerning large datasets, I took the opportunity to do some more testing and decided to increase the threshold for when a dataset is considered to be large by a few orders of magnitude. This is purely empirical...

### DIFF
--- a/packages/app/src/providers/mock/mock-api.ts
+++ b/packages/app/src/providers/mock/mock-api.ts
@@ -61,6 +61,10 @@ export class MockApi extends DataProviderApi {
   public async getValue(params: ValuesStoreParams): Promise<unknown> {
     const { dataset, selection } = params;
 
+    if (dataset.name === 'raw_large') {
+      return { str: '.'.repeat(1_000_000) };
+    }
+
     if (dataset.name === 'error_value') {
       throw new Error('error');
     }
@@ -91,6 +95,13 @@ export class MockApi extends DataProviderApi {
     const url = this._getExportURL?.(format, dataset, selection, value);
     if (url) {
       return url;
+    }
+
+    if (format === 'json') {
+      return async () => {
+        const json = JSON.stringify(value, null, 2);
+        return new Blob([json]);
+      };
     }
 
     if (
--- a/packages/lib/src/toolbar/OverflowMenu.module.css
+++ b/packages/lib/src/toolbar/OverflowMenu.module.css
@@ -23,7 +23,9 @@
 .menuList {
   composes: popupInner from './Toolbar.module.css';
   display: grid;
+  grid-template-columns: 1fr;
   grid-gap: 0.25rem;
+  justify-items: flex-start;
   margin: 0;
   padding: 0.375rem 0.25rem;
   list-style-type: none;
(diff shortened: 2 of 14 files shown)


packages/app/src/providers/mock/mock-api.ts
```diff
@@ -61,6 +61,10 @@ export class MockApi extends DataProviderApi {
   public async getValue(params: ValuesStoreParams): Promise<unknown> {
     const { dataset, selection } = params;
 
+    if (dataset.name === 'raw_large') {
+      return { str: '.'.repeat(1_000_000) };
+    }
```
axelboc, line 66: Now generating this particular dataset dynamically to not clutter the memory.

packages/lib/src/toolbar/OverflowMenu.module.css
```diff
@@ -23,7 +23,9 @@
 .menuList {
   composes: popupInner from './Toolbar.module.css';
   display: grid;
+  grid-template-columns: 1fr;
   grid-gap: 0.25rem;
+  justify-items: flex-start;
```
axelboc, line 28: Fix positioning of menus that are nested inside the overflow menu so they are properly aligned with their buttons.

packages/lib/src/toolbar/controls/ExportMenu.tsx
```diff
@@ -8,11 +8,12 @@ import styles from './Selector/Selector.module.css';
 
 interface Props {
   entries: ExportEntryProps[];
-  isSlice: boolean;
+  isSlice?: boolean;
+  align?: 'center' | 'left' | 'right';
```
axelboc, line 12: Allow aligning the export menu to the right of the "Export" button to fix overflow issues in the Raw and Matrix toolbars.

t20100 (approved):
Nice!
Neater than dumping to the console.

packages/lib/src/vis/raw/RawVis.tsx
```diff
@@ -8,23 +10,13 @@ function RawVis(props: Props) {
   const { value } = props;
   const valueAsStr = JSON.stringify(value, null, 2);
```
loichuder, line 11: This step could be time consuming if `value` is too big. I wonder if we could infer the size in another lighter way.

axelboc: Yeah, I did wonder about that. On the other hand, the `value` first has to be downloaded, which takes a _lot_ more time than `JSON.stringify`. Even in VS Code we have a 2GB limit on the size of the HDF5 file, so I'm not hugely concerned. Given that the raw vis is for datasets that have complicated dtypes, I'm not sure there's an easier way to approximate the size... So I'd prefer to wait for someone to hit this potential bottleneck before trying to fix it.

loichuder: Sounds good to me :+1:

PR #1486, opened 2023-09-01
Parse nested array types returned by h5wasm

axelboc:
I was playing with compound datasets as part of #1485 when I noticed this "Unknown" type. Definitely a bit of a niche fix...

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/packages/h5wasm/src/guards.ts
+++ b/packages/h5wasm/src/guards.ts
@@ -61,6 +61,10 @@ export function isStringMetadata(metadata: Metadata) {
   return metadata.type === 3;
 }
 
+export function isArrayMetadata(metadata: Metadata) {
+  return metadata.type === 10;
+}
+
 export function isCompoundMetadata(
   metadata: Metadata,
 ): metadata is CompoundMetadata {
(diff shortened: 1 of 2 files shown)


packages/h5wasm/src/guards.ts
```diff
@@ -61,6 +61,10 @@ export function isStringMetadata(metadata: Metadata) {
   return metadata.type === 3;
 }
 
+export function isArrayMetadata(metadata: Metadata) {
+  return metadata.type === 10;
+}
```
loichuder, line 66: Do you think we could use the `H5T_class_t` directly from `h5wasm` ?

axelboc: Not in its current state, since `H5T_class_t` is an `interface`, not an `enum`. I'll write it down in case I find time to open a PR on the h5wasm repo. :+1:

loichuder: Ah shoot, indeed. 

> I'll write it down in case I find time to open a PR on the h5wasm repo.

Thanks

PR #1490, opened 2023-09-12
Upgrade dependencies

axelboc:
A good round of upgrades! I'll deal with TS, Vite, R3F and Three in separate PRs.

---

- [axios](https://github.com/axios/axios/releases) from 1.3.4 to 1.5.0
- [@visx/*](https://github.com/airbnb/visx/blob/master/CHANGELOG.md) from 3.1.2 to 3.3.0
- [d3-array 3.2.4](https://github.com/d3/d3-array/releases/tag/v3.2.4)
- [h5wasm](https://github.com/usnistgov/h5wasm/releases) from 0.5.2 to 0.6.2
  - Not relevant to us but it is now possible to create compressed datasets.
  - Also not relevant but could be in the future for [#616](https://github.com/silx-kit/h5web/issues/616) or myHDF5: it is now possible to slice datasets with a step size.
- [zustand](https://github.com/pmndrs/zustand/releases) from 4.3.7 to 4.4.1
  - It doesn't affect us, but I think v4.4 basically removes `useStore(..., shallow)`. You can still create a store with `shallow` but then all calls to `useStore` for that store will use `shallow`. The reasoning is explained [here](https://github.com/pmndrs/zustand/discussions/1937) and [here](https://github.com/pmndrs/zustand/discussions/1937#discussioncomment-6962978).

---

- [Storybook](https://github.com/storybookjs/storybook/releases) from 7.1.1 to 7.4.1
- [Jest + jest-environment-jsdom 29.7.0](https://github.com/jestjs/jest/releases/tag/v29.7.0)
- [dom-testing-library](https://github.com/testing-library/dom-testing-library/releases) from 8.20.0 to 9.3.1
- [@testing-library/jest-dom](https://github.com/testing-library/jest-dom/releases) from 5.17.0 to 6.1.3
- [@testing-library/react](https://github.com/testing-library/react-testing-library/releases) from 13.2.0 to 14.0.0
  - Supposedly gets rid of some `act` warnings, but not the one in `DomainWidget.test.tsx`...
- [Cypress](https://docs.cypress.io/guides/references/changelog) from 12.9.0 to 13.1.0
[...]

PR #1491, opened 2023-09-13
Upgrade Vite and fix generated declaration files

axelboc:
Vite 4.4 adds a couple of convenient keyboard shortcuts to the CLI, including one for opening the browser. As a result, I decided to disable opening the browser by default, as this can be annoying sometimes.

Unrelated: after the `rollup` upgrade in #1490, we were getting a lot of warnings when building packages. It turned out that the `index.d.ts` generated for each package was including all the React types! So I'm fixing this.

Also, there was a type import from `@h5web/app/src` in `@h5web/h5wasm` that shouldn't have been there.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,6 +28,9 @@ pnpm install
 pnpm start
 ```
 
+Once the development server has started, press `o` to open the development URL
+in your browser, or `h` to show all the available keyboard shortcuts.
+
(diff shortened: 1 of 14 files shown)


loichuder (approved):
> As a result, I decided to disable opening the browser by default, as this can be annoying sometimes.

Not sure how I feel about that. I'll have to make my mind when developing.

PR #1492, opened 2023-09-13
Shorten `flex` shorthand properties

axelboc:
If unspecified, the defaults for a flex item are:

- `flex-grow: 0`
- `flex-shrink: 1`
- `flex-basis: auto`

Very often, what we want is `flex: 1 1 0` - i.e. for the item to grow/shrink to fill the available space regardless of its intrinsic size (so that all flex items inside a container get sized equally).

The [`flex`](https://devdocs.io/css/flex) shorthand knows this is the most common case and that people don't generally want `flex-basis: auto`, so when one or two unit-less values are specified (for `flex-grow` and `flex-shrink`) it sets `flex-basis` to `0` instead of keeping the default value of `auto`:

- `flex: 1` <=> `flex: 1 1 0`
- `flex: 1 0` <=> `flex:  1 0 0`

I used to think that this behaviour was kind of obscure and always preferred to write `flex: 1 1 0%`. After seeing that modern CSS tooling like [Lightning CSS](https://lightningcss.dev/) minifies `flex: 1 1 0%` to `flex: 1`, I thought I'd reconsider my stance. After all, the `flex` shorthand is trying to help, so perhaps I was silly to not take advantage of it.

When it comes to `0%` vs `0` for `flex-basis`, this used to be a [workaround for ... IE11](https://github.com/philipwalton/flexbugs#flexbug-4)!! I think it's time for me to move on... 😅

### DIFF
--- a/packages/app/src/dimension-mapper/SlicingSlider.module.css
+++ b/packages/app/src/dimension-mapper/SlicingSlider.module.css
@@ -28,7 +28,7 @@
 }
 
 .track {
-  flex: 1 0 auto;
+  flex: 1;
   margin: calc(var(--thumb-height) / 2) 0;
   width: 2px;
   background-color: var(--secondary-dark);
--- a/packages/lib/src/toolbar/controls/DomainWidget/BoundEditor.module.css
+++ b/packages/lib/src/toolbar/controls/DomainWidget/BoundEditor.module.css
@@ -14,7 +14,7 @@
 }
 
 .value {
-  flex: 1 0 auto;
+  flex: 1;
   width: 8.5em; /* -9.99999e+999 without overflowing */
   margin-right: 0.375rem;
   padding: 0.25rem 0.375rem;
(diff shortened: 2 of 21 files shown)


packages/lib/src/toolbar/controls/DomainWidget/BoundEditor.module.css
```diff
@@ -14,7 +14,7 @@
 }
 
 .value {
-  flex: 1 0 auto;
```
axelboc, line 17: We specify a `width` below, so the item will not shrink below that width regardless of `flex-shrink` and `flex-basis`.

packages/app/src/dimension-mapper/SlicingSlider.module.css
```diff
@@ -28,7 +28,7 @@
 }
 
 .track {
-  flex: 1 0 auto;
```
axelboc, line 31: This used to be a workaround to `1 1 0%` for flex items inside column flex containers in legacy browsers. Just like `0%`, it is no longer relevant.

loichuder (approved):
> I used to think that this behaviour was kind of obscure

Yeah, me too. But now that you spell the reasoning behind it, it makes sense in the end.

Issue #1498, opened 2023-09-21
Need H5Web to support plot True/False and Enum signals


Yayanzjg:
### Requested solution or feature

I am dealing with many signals with True/False value or kActive/kNotActive. It would be fantastic if H5Web can support to plot this kind of two-value signal as 0/1.

axelboc:
Hi @Yayanzjg, I've transferred the issue to the `h5web` repository. Will reply there :blush:

axelboc:
Hi @Yayanzjg, could you please share an example HDF5 file and maybe a screenshot of the visualization you're after made with another tool like matplotlib?

xbaelus:
Hi! @axelboc I'd love to have this in **vscode-h5web**, but since the issue here is still open, not sure if I should open another issue there?

axelboc:
@xbaelus you're in the right place! Thanks for letting us know.

axelboc:
@xbaelus please keep an eye out for the next releases of H5Web and vscode-h5web. Might take us a few weeks. 😉

xbaelus:
Thanks a lot for taking care of this @axelboc ! :smile:

PR #1499, opened 2023-09-25
Support visualizing nD boolean datasets as Line and Heatmap

axelboc:
Proposal implementation for #1498 with some unresolved questions.

I had a couple of options:

1. Convert the boolean array to a numeric arrays before passing it to the mapped visualization components (current proposal).
2. Let the boolean array go through and let `useDomain` handle boolean arrays (to return `[0, 1]` obviously) – apart from that, it turns out that `DataCurve` handles `boolean` array out of the box (though taking advantage of this may not be such a great idea... :shrug: :smile:)

Option 2 seemed to complexify a bunch lot of code in terms of typing. Option 1 obviously requires mapping through the `values` array, but seems like a good compromise in terms of complexity (even though it's not trivial either in terms of typing – see below.)

A third option could be to let the data provider convert the boolean values to numbers when visualizing as line/heatmap; or maybe even to request a numeric array straight from the back-end.

Either way, there's a bit of unresolved complexity when it comes to the `Value` type (i.e. what should `Value<Dataset<ArrayShape, NumericType | BooleanType>>` resolve to?), since for the matrix vis we want actual booleans  (`boolean[]`), while for the line/heatmap vis, we want either `TypedArray | number[] | boolean[]` (and not `(number | boolean)[]`!) for option 1 or `TypedArray | number[]` for option 2/3/4.

image

image

Other unresolved question: the issue mentions enums, which makes sense since (if I recall) booleans are stored as enums in HDF5. However, we do treat them separately in H5Web. Perhaps that wasn't such a good idea? :shrug:


loichuder (commented):
Honestly, this PR looks quite all right to me. Sure, it may not seem the prettiest since you need to propagate typing changes down the line, but there is no way around it. The alternatives you listed would have been even more verbose in terms of typings.

Perhaps you can introduce a `NumericLikeType` to go along with the new guards/assertions ?

loichuder:
Also, once there is a real usecase with `enums`, perhaps we will find a better way to harmonize this. But in the mean time, this is a quite sensible implementation.

axelboc:
I might just investigate option 3/4 a bit more, because with the current solution, when using `H5WasmProvider`, the data is retrieved as an array of numbers, converted to an array of booleans, and then back to an array of numbers. Seems like a bit of a waste.

.gitattributes
axelboc, line 1: I think this file was actually preventing git conflicts from being resolved automatically in `pnpm-lock.yml` when rebasing.

packages/app/src/vis-packs/core/matrix/utils.ts
```diff
@@ -51,7 +51,7 @@ export function getFormatter(
     return createNumericFormatter(notation);
   }
 
-  return (val) => (val as string).toString();
+  return (val) => (val as string | boolean).toString();
```
axelboc, line 54: unrelated -- just a tiny bit more accurate

packages/app/src/vis-packs/core/utils.ts
```diff
@@ -83,3 +89,11 @@ export function getSliceSelection(
 export function getImageInteractions(keepRatio: boolean) {
   return keepRatio ? BASE_INTERACTIONS : INTERACTIONS_WITH_AXIAL_ZOOM;
 }
+
+export function toNumArray(arr: ArrayValue<NumericLikeType>): NumArray {
+  if (typeof arr[0] === 'boolean') {
+    return arr.map((val) => (val ? 1 : 0));
```
axelboc, line 95: This apparently performs better than `Number()` on some browsers. `Number()` performs type coercion, which is more complicated than a simple ternary.

packages/app/src/vis-packs/core/utils.ts
```diff
@@ -83,3 +89,11 @@ export function getSliceSelection(
...
+  }
+
+  return arr as NumArray;
```
axelboc, line 98: Scratched my head trying to remove this `as`, but two problems make this difficult:

- The value array is separated from the dataset object that describes the type, so narrowing the type of the dataset object does not help narrow the type of the value array...
- `ArrayValue<NumericType | BooleanType>` resolves to `(number | boolean)[] | TypedArray`. I actually [found a way](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types) to get `number[] | boolean[] | TypedArray` but it broke a bunch of code related to compound datasets ... and it makes sense for compound value arrays to have type `(number | string | ...)[] | TypedArray)`.

packages/lib/src/vis/heatmap/HeatmapVis.tsx
```diff
@@ -30,7 +30,7 @@ interface Props {
   aspect?: Aspect;
   showGrid?: boolean;
   title?: string;
-  dtype?: NumericType;
+  dtype?: NumericLikeType;
```
axelboc, line 33: Not a fan of `@h5web/lib` knowing about `NumericLikeType` (or any other `DType`). I'm thinking I might change this to `string` and move the call to `formatNumLikeType` into the mapped components at some point.

packages/app/src/vis-packs/core/utils.ts
loichuder: Yeah. I have the feeling there is always a way to remove `as` but sometimes it is not worth the complexity.

loichuder (approved):
Well done :partying_face:

PR #1512, opened 2023-10-16
Pass `dtype` as string to visualization components

axelboc:
I wasn't fond of `LineVis` and `HeatmapVis` being aware of some of the `DType` models—those should concern only the H5Web viewer.

### DIFF
--- a/packages/app/src/vis-packs/core/utils.ts
+++ b/packages/app/src/vis-packs/core/utils.ts
@@ -5,7 +5,7 @@ import type {
   NumArray,
   NumericLikeType,
 } from '@h5web/shared';
-import { createArrayFromView } from '@h5web/shared';
+import { createArrayFromView, DTypeClass, isNumericType } from '@h5web/shared';
 import { isNumber } from 'lodash';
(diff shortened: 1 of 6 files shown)


packages/app/src/vis-packs/core/utils.ts
```diff
@@ -97,3 +97,14 @@ export function toNumArray(arr: ArrayValue<NumericLikeType>): NumArray {
...
+const TYPE_STRINGS: Record<NumericLikeType['class'], string> = {
+  [DTypeClass.Bool]: 'bool',
+  [DTypeClass.Integer]: 'int',
+  [DTypeClass.Unsigned]: 'uint',
+  [DTypeClass.Float]: 'float',
+};
+
+export function formatNumLikeType(type: NumericLikeType): string {
+  return `${TYPE_STRINGS[type.class]}${isNumericType(type) ? type.size : ''}`;
+}
```
axelboc, line 110: Moved from `@h5web/lib` to `@h5web/app`.

loichuder (approved):
Makes sense since the dtype is only passed to be converted as string.

Good thinking :+1:

PR #1513, opened 2023-10-23
Migrate CodeSandboxes

axelboc:
@loichuder, as discussed, I've forked your CodeSandboxes in my own account:

- [Multiple curves](https://codesandbox.io/p/sandbox/h5weblib-multiple-curves-46kppn)
- [Heatmap with color bar](https://codesandbox.io/p/sandbox/h5weblib-heatmap-with-tooltip-4nc9hp)
- [Gamma vs Power scale](https://codesandbox.io/p/sandbox/h5weblib-gamma-vs-power-kksyd7)
- [Value-to-index scale](https://codesandbox.io/p/sandbox/h5weblib-value-to-index-scale-9ldn9k)

They're now all based on the `h5weblib-vite` CodeSandbox, which uses Vite, pnpm, React 18, TS 5, etc.


loichuder (approved):
Looks good !

I wonder if it is the same for you,  but the first time I land on the sandbox, the displayed file is the `.gitignore` which is not super helpful.

Perhaps we can provide the URL pointing to `App.tsx` instead (e.g. https://codesandbox.io/p/sandbox/h5weblib-multiple-curves-46kppn?file=%2Fsrc%2FApp.tsx%3A1%2C1) ?

axelboc:
Great idea!

PR #1514, opened 2023-10-24
Fix overly permissive regex

axelboc:
I've activated CodeQL analysis on the repo to try it out: https://github.com/silx-kit/h5web/settings/security_analysis

It detected an issue in a regex: https://github.com/silx-kit/h5web/security/code-scanning/1

### DIFF
--- a/packages/app/src/providers/h5grove/utils.ts
+++ b/packages/app/src/providers/h5grove/utils.ts
@@ -54,11 +54,11 @@ export function convertH5GroveDtype(dtype: H5GroveDtype): DType {
 }
 
 function convertDtypeString(dtype: string): DType {
-  const regexp = /([<>=|])?([A-z])(\d*)/u;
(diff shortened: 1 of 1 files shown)


loichuder (approved):
Good to know.

I wonder if we could be even more restrictive. This part of the regex looks for the letter representing the `dataType` and this letter can only be a few (as evidenced by the `switch` below).

axelboc:
Hmm but if the regex doesn't match, we throw an error; and the error is more for cases where the format of the dtype string is invalid rather than unknown. When the format looks right but the dtype "letter" is not recognised, then we return an unknown type. So I'd rather be lenient on the letter that can be matched.

However, maybe I should reword the error message to "Invalid dtype" to avoid confusion.

axelboc:
I should add that this relates to the testing of the provider that I'd like to put in place: I'm hoping it will help make this kind of dtype parsing a bit more exhaustive and therefore stricter.

PR #1518, opened 2023-10-31
Fix flaky visual diffing tests

axelboc:
I'm replacing `text-decoration: underline dotted`, which is a [relatively new shorthand syntax](https://caniuse.com/text-decoration) and seems to be the cause of our flaky visual diffing tests, with an old-school `border-bottom` to see if it improves things.


axelboc:
All tests passed on first attempt :partying_face:

loichuder (approved):

PR #1522, opened 2023-11-03
Lazy-load compression filters in `H5WasmProvider`

axelboc:
Thanks to the new [h5wasm-plugins](https://github.com/h5wasm/h5wasm-plugins) package (kudos @bmaranville!!), I'm able to load compression plugins on demand in `H5WasmProvider`. :tada: 

I was hoping that everything could be taken care of by `H5WasmProvider` under the hood, but since `@h5web/h5wasm` is a library, it's not that simple. The plugin files have to be loaded into Emscripten's file system as array buffers, which means they first have to be fetched lazily somehow ... and only the consumer application knows how to do that:

- perhaps the application is bundled with Vite or webpack, which means the plugins can be imported as static assets (i.e. to get their file system URLs) ` and then fetched as array buffers;
- perhaps the application doesn't use a bundler and just serves the plugin files at known URLs;
- perhaps the application runs in an environment that prevents the use of `fetch` altogether (e.g. VS Code webview). 

The solution I came up with is to expose a `getPlugin` prop. If the prop is implemented by the consumer app, then it will be called with the name of a known plugin when the time comes to load it into Emscripten. It is then up to the consumer application to return an array buffer (or `undefined` if the plugin is not supported).

`H5WasmProvider` holds a mapping of known plugins against their corresponding HDF5 filter IDs. For now, this list will need to be kept up to date as more plugins are added to `h5wasm-plugins`. In the future, we could add a prop to let consumers declare additional plugins without having to wait for H5Web updates.

This PR brings support for the following **7 compression filters** in the h5wasm demo: Blosc, bzip2, LZ4, LZF, SZ, ZFP and Zstandard. You can try it with this file, which comes from the `h5wasm-plugins` repo: https://www.silx.org/pub/h5web/filters.h5

### DIFF
--- a/packages/h5wasm/README.md
+++ b/packages/h5wasm/README.md
@@ -120,7 +120,7 @@ means that:
 
 - your build tool must understand the BigInt notation (e.g.
   [babel-plugin-syntax-bigint](https://babeljs.io/docs/en/babel-plugin-syntax-bigint))
-- your application will only run in browsers that
+- your application will run only in browsers that
   [support BigInt](https://caniuse.com/bigint).
 
 ### External links are not resolved
@@ -165,3 +165,42 @@ See
 `H5WasmProvider` does not provide a fallback implementation of `getExportURL` at
 this time, so if you don't provide your own, the export menu will remain
 disabled in the toolbar.
+
+#### `getPlugin?: (name: string) => Promise<ArrayBuffer | undefined>`
+
+If provided, this aysnchronous function is invoked when loading a compressed
+dataset. It receives the name of a compression plugin as parameter and should
+return:
+
+- the compression plugin's source file as `ArrayBuffer`,
+- or `undefined` if the plugin is not available.
+
+`@h5web/h5wasm` is capable of identifying and requesting the plugins supported
+by the
+[`h5wasm-plugins@0.0.1`](https://github.com/h5wasm/h5wasm-plugins/tree/v0.0.1)
+package: `blosc`, `bz2`, `lz4`, `lzf`, `szf`, `zfp`, `zstd`.
+
+A typical implementation of `getPlugin` in a bundled front-end application might
+look like this:
+
+```ts
+/*
+ * Import the plugins' source files as static assets (i.e. as URLs).
+ * The exact syntax may vary depending on your bundler (Vite, webpack ...)
+ * and may require extra configuration/typing.
+ */
+import blosc from 'h5wasm-plugins/plugins/libH5Zblosc.so';
+import bz2 from 'h5wasm-plugins/plugins/libH5Zbz2.so';
+// ...
+
+const PLUGINS = { blosc, bz2 /* ... */ };
+
+async function getPlugin(name: string): Promise<ArrayBuffer | undefined> {
+  if (!PLUGINS[name]) {
+    return undefined;
+  }
+
+  const response = await fetch(PLUGINS[name]);
+  return response.arrayBuffer();
+}
+```
--- a/packages/h5wasm/src/h5wasm-api.ts
+++ b/packages/h5wasm/src/h5wasm-api.ts
@@ -22,8 +22,8 @@ import {
   EntityKind,
   hasArrayShape,
 } from '@h5web/shared';
-import type { Attribute as H5WasmAttribute } from 'h5wasm';
-import { File as H5WasmFile, Module, ready as h5wasmReady } from 'h5wasm';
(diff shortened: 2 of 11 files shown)


packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -135,21 +146,59 @@ export class H5WasmApi extends DataProviderApi {
...
+    // Replace default plugins path
+    module.remove_plugin_search_path(0);
+    module.insert_plugin_search_path(PLUGINS_PATH, 0);
 
-    // `FS` is guaranteed to be defined once H5Wasm is ready
-    // https://github.com/silx-kit/h5web/pull/1082#discussion_r858613242
-    assertNonNull(FS);
+    // Create plugins folder on Emscripten virtual file system
+    // @ts-expect-error
+    module.FS.mkdirTree(PLUGINS_PATH);
```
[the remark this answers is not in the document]
axelboc: https://github.com/usnistgov/h5wasm/pull/62

packages/h5wasm/README.md
```diff
@@ -165,3 +165,42 @@ See
...
+/*
+ * Import the plugins' source files as static assets (i.e. as URLs).
+ * The exact syntax may vary depending on your bundler (Vite, webpack ...)
+ * and may require extra configuration/typing.
+ */
+import blosc from 'h5wasm-plugins/plugins/libH5Zblosc.so';
+import bz2 from 'h5wasm-plugins/plugins/libH5Zbz2.so';
+// ...
+
+const PLUGNS = { blosc, bz2 /* ... */ };
```
loichuder, line 196:
```suggestion
const PLUGINS = { blosc, bz2 /* ... */ };
```

packages/h5wasm/src/h5wasm-api.ts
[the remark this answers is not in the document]
loichuder: I see this is already merged so at next h5wasm release, we should get a linter warning saying that the `@ts-expect-error` is no longer needed, right ?

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -135,21 +146,59 @@ export class H5WasmApi extends DataProviderApi {
...
+      }
+
+      const pluginPath = `${PLUGINS_PATH}/libH5Z${plugin}.so`;
+
+      if (h5Module.FS.analyzePath(pluginPath).exists) {
+        continue; // plugin already loaded
+      }
+
+      const buffer = await this.getPlugin?.(plugin);
+      if (buffer) {
```
loichuder, line 196: Shouldn't you return an error if `buffer` is undefined ? 

Even if the plugin exists in `PLUGINS_BY_FILTER_ID`, it may be that the `getPlugin` method does not implement it and you may want to notify the user about it.

packages/h5wasm/README.md
```diff
@@ -165,3 +165,42 @@ See
...
+
+async function getPlugin(name: string): Promise<ArrayBuffer | undefined> {
+  if (!PLUGINS[name]) {
+    return undefined;
+  }
+
+  const response = await fetch(PLUGINS[name]);
+  return response.arrayBuffer();
+}
```
loichuder, line 205: Yeah, it is bit weird since in the end, every consumer application will implement the same function as above...

loichuder (approved):
Amazing :star_struck:

packages/h5wasm/src/h5wasm-api.ts
axelboc: True! I didn't want to throw when `getPlugin` is not implemented, and this threw me off :drum:

packages/h5wasm/README.md
axelboc: Well, you could imagine calling `fetch` a little differently, for instance if the plugins are served from a different domain. And in the VS Code extension, I'll have to use `postMessage`.

loichuder: Yeah, it will be similar but different. Which is the worst in programming.

packages/h5wasm/src/h5wasm-api.ts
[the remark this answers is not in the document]
bmaranville: I was working on a different feature in h5wasm, so this already made it to release in v0.6.9

packages/h5wasm/src/h5wasm-api.ts
axelboc: Done. The error message is slightly different from the other one to make sure we can distinguish them.

PR #1524, opened 2023-11-10
Support Blosc2 and Bitshuffle plugins

axelboc:
With [hwasm@0.6.10](https://github.com/usnistgov/h5wasm/releases), the Blosc2 plugin now works as expected! :tada: 

<details>
<summary>Initial PR description</summary>

Opening as draft because Blosc2 is not working as expected. The `blosc2` dataset of the test file `filters.h5` is not correctly decompressed. You can try it locally: http://localhost:5173/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Ffilters.h5

image

---

What's more, there's definitely something weird going on — it seems that plugins have an impact of the decompression of datasets compressed with other plugins. For instance:

1. Open `filters.h5`
2. Select `lz4` dataset => looks correctly decompressed
3. Select `blosc2` then `lz4` again => still works
4. Reload
5. Select `blosc2` first, then `lz4` => no longer works

image

Other example: if you select `zstd` first and then `blosc2`, it looks different than before:

image

</details>

### DIFF
--- a/apps/demo/src/h5wasm/plugin-utils.ts
+++ b/apps/demo/src/h5wasm/plugin-utils.ts
@@ -1,6 +1,9 @@
 // Import compression plugins as static assets (i.e. as URLs)
 // cf. `vite.config.ts` and `src/vite-env.d.ts
+import { Plugin } from '@h5web/h5wasm';
 import blosc from 'h5wasm-plugins/plugins/libH5Zblosc.so';
+import blosc2 from 'h5wasm-plugins/plugins/libH5Zblosc2.so';
+import bshuf from 'h5wasm-plugins/plugins/libH5Zbshuf.so';
 import bz2 from 'h5wasm-plugins/plugins/libH5Zbz2.so';
 import lz4 from 'h5wasm-plugins/plugins/libH5Zlz4.so';
 import lzf from 'h5wasm-plugins/plugins/libH5Zlzf.so';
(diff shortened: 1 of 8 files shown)


packages/h5wasm/src/h5wasm-api.test.ts
```diff
@@ -25,7 +25,7 @@ async function loadTestFile(): Promise<ArrayBuffer> {
   return resp.arrayBuffer();
 }
 
-test('test file matches snapshot', async () => {
+test.skip('test file matches snapshot', async () => {
```
axelboc, line 28: Since yesterday, the `H5WasmApi` test snapshot seems to be different on my machine and on the CI. Some of the `UInt8Array`s that contain memory addresses have slightly different bytes... This is unrelated to this PR (`main` is failing as well), so I'm skipping the test until I can figure out a solution.

apps/demo/src/h5wasm/plugin-utils.ts
```diff
@@ -1,25 +1,30 @@
 // Import compression plugins as static assets (i.e. as URLs)
 // cf. `vite.config.ts` and `src/vite-env.d.ts
+import { Plugin } from '@h5web/h5wasm';
```
axelboc, line 3: Introducing an enum so it's easier for consumer applications to be exhaustive and not forget to load/support a plugin.

loichuder: Should it be documented ?

loichuder (approved):

apps/demo/src/h5wasm/plugin-utils.ts
axelboc: Good catch! I've updated the README of `@h5web/h5wasm`.

Issue #1525, opened 2023-11-10
import problem on @h5web/h5wasm components when using NextJS


borin98:
### Describe the bug

Hello people. The description is that i tried to create a simple NextJs using @h5web/h5wasm. But for some reason, this package is having a import the import error " Module not found: Can't resolve './' ". Can you guys help me out with this task ?

### To Reproduce

1. install the project with NextJs basic stac : npx create-next-app@latest
2. Use the following options
```
✔ What is your project named? … test-h5web
✔ Would you like to use TypeScript? … No / Yes
✔ Would you like to use ESLint? … No / Yes
✔ Would you like to use Tailwind CSS? … No / Yes
✔ Would you like to use `src/` directory? … No / Yes
✔ Would you like to use App Router? (recommended) … No / Yes
✔ Would you like to customize the default import alias (@/*)? … No / Yes
Creating a new Next.js app in ~/test-h5web

Using npm.

Initializing project with template: app-tw 


Installing dependencies:
- react
- react-dom
- next

Installing devDependencies:
- typescript
- @types/node
- @types/react
- @types/react-dom
- autoprefixer
- postcss
- tailwindcss


added 117 packages, and audited 118 packages in 9s

found 0 vulnerabilities
Initialized a git repository.

Success! Created bla at ~/test-h5web
```
3. use 
```
yarn add @h5web/app @h5web/h5wasm
```
4. Then, replace ```src/app/page.tsx``` with the code
```
"use client"
import '@h5web/app/dist/styles.css';

import React, { useState } from 'react';
import type { ChangeEvent } from 'react';
import { App } from '@h5web/app';
import { H5WasmProvider } from '@h5web/h5wasm';

interface H5File {
  filename: string;
  buffer: ArrayBuffer;
}

function MyApp() {
  const [h5File, setH5File] = useState<H5File>();

  function handleFileInputChange(evt: ChangeEvent<HTMLInputElement>) {
    const file = evt.target.files?.[0];
    if (!file) {
      return;
    }

    /* `H5WasmProvider` expects an HDF5 file in the form of an `ArrayBuffer`.
     * The way you get this buffer is up to you. Here, we show a simple file picker
     * and use the FileReader API to process the selected file. */
    const reader = new FileReader();
    reader.addEventListener('load', () => {
      setH5File({
        filename: file.name,
        buffer: reader.result as ArrayBuffer,
      });
    });

    reader.readAsArrayBuffer(file);
  }

  if (!h5File) {
    return (
[...]

PR #1528, opened 2023-11-24
Use mjs extension for ESM builds

axelboc:
As recommended by https://publint.dev/@h5web/h5wasm@10.0.0.

I will release a beta to test if this helps with #1525.


axelboc:
It does not help, unfortunately.

PR #1529, opened 2023-11-27
Remove barrel file from @h5web/shared

axelboc:
Before v5, TypeScript could not understand the `exports` property in `package.json` that allows defining multiple entry points in a package. As a result we had to re-export everything from a single file, `src/index.ts`, and import everything from `@h5web/shared`. Unfortunately, this type of "barrel" file is really bad for performance in development, since any change to any module in the shared package would trigger an HMR update of every module that depend on the shared package. This used to really ruin the development experience...

Now that we've fully switched to `moduleResolution: 'bundler'`, which does understand `exports`, I was able to add export entries for each file in `@h5web/shared`.

The only challenge was to make sure Rollup knew where to find the shared compiled declaration files in `packages/shared/dist-ts` when generating `packages/<app,lib,h5wasm>/dist/index.d.ts`.

HMR performance is now much, much better. It's still not perfect, and we could probably also provide multiple entry points in the lib package, but it's a start!

### DIFF
--- a/packages/app/rollup.config.js
+++ b/packages/app/rollup.config.js
@@ -2,6 +2,7 @@ import alias from '@rollup/plugin-alias';
 import { defineConfig } from 'rollup';
 import dts from 'rollup-plugin-dts';
 
+import { aliasEntries } from '../shared/rollup.utils.js';
 import { externals } from './vite.config.js';
 
 export default defineConfig({
@@ -9,15 +10,7 @@ export default defineConfig({
   output: [{ file: 'dist/index.d.ts', format: 'es' }],
   external: [...externals, /\.css$/u],
   plugins: [
-    alias({
-      entries: [
-        {
-          // Make sure rollup-plugin-dts can find shared types
-          find: '@h5web/shared',
-          replacement: '../shared/dist-ts/index.d.ts',
-        },
-      ],
-    }),
+    alias({ entries: aliasEntries }), // make sure rollup-plugin-dts can find type declatations in @h5web/shared
     dts(),
   ],
 });
--- a/packages/lib/src/styles.css
+++ b/packages/lib/src/styles.css
@@ -1,2 +1,2 @@
-/* Global lib styles for demo/Storybook */
-@import '@h5web/shared/src/styles.css'; /* global shared styles */
+/* Global lib styles for demo/Storybook (imported via `packages/app/src/styles.css`) */
+@import '@h5web/shared/styles.css'; /* global shared styles */
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -6,6 +6,21 @@
   "type": "module",
   "sideEffects": false,
   "main": "src/index.ts",
+  "exports": {
+    "./createMemo": "./src/createMemo.ts",
+    "./guards": "./src/guards.ts",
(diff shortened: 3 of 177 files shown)


packages/app/rollup.config.js
```diff
@@ -2,22 +2,15 @@ import alias from '@rollup/plugin-alias';
...
-    alias({
-      entries: [
-        {
-          // Make sure rollup-plugin-dts can find shared types
-          find: '@h5web/shared',
-          replacement: '../shared/dist-ts/index.d.ts',
-        },
-      ],
-    }),
+    alias({ entries: aliasEntries }), // make sure rollup-plugin-dts can find type declatations in @h5web/shared
```
axelboc, line 13: All the rollup config files need the same alias entries, so I've put the code in a shared file, `packages/shared/rollup.utils.js`.

packages/shared/package.json
```diff
@@ -6,6 +6,21 @@
...
+    "./models-vis": "./src/models-vis.ts",
+    "./utils": "./src/utils.ts",
+    "./mock/metadata-utils": "./src/mock/metadata-utils.ts",
+    "./mock/metadata": "./src/mock/metadata.ts",
+    "./mock/models": "./src/mock/models.ts",
+    "./mock/utils": "./src/mock/utils.ts",
+    "./mock/values": "./src/mock/values.ts",
+    "./styles.css": "./src/styles.css",
+    ".": "./src/index.ts"
+  },
```
axelboc, line 23: The new entrypoints of `h5web/shared`. Note that `.` is mandatory, so I've made it point to an empty module.

packages/shared/rollup.utils.js
```diff
@@ -0,0 +1,15 @@
...
+
+const dirname = fileURLToPath(new URL('.', import.meta.url));
+const pkg = JSON.parse(fs.readFileSync(path.resolve(dirname, 'package.json')));
+
+export const aliasEntries = Object.fromEntries(
+  Object.entries(pkg.exports)
+    .filter(([, value]) => value.endsWith('.ts'))
+    .map(([key, value]) => [
+      key.replace(/^\./u, '@h5web/shared'),
+      value.replace(/^\.\/src\/(.+)\.ts$/u, '../shared/dist-ts/$1.d.ts'),
```
axelboc, line 13: This means, for instance, that import `@h5web/shared/utils` is aliased to `../shared/dist-ts/utils.d.ts`.

packages/lib/src/styles.css
```diff
@@ -1,2 +1,2 @@
 /* Global lib styles for demo/Storybook */
```
loichuder, line 1: Actually, this file is only imported by `packages/app/src/styles.css` so I guess the comment should be updated ?

packages/shared/rollup.utils.js
loichuder: Worth putting as comment IMO

loichuder (approved):
I hope this will not come back to bite us when using the packages in other contexts...

I guess it should not since this is only internal stuff but I am never sure :joy:

packages/shared/rollup.utils.js
loichuder: The beauty of reg exps: what you don't write in code, you need to write it in comments :joy_cat:

PR #1537, opened 2024-01-05
Generate mock data lazily and improve API

axelboc:
I realised that the `mockValues` and `mockMetadata` objects were not properly tree-shaken in consumer applications of `@h5web/lib` or `@h5web/app` (at least with ones using Vite as their bundler). As a result, those two giant data structures were being created **on every page load of myHDF5**, just to name one.

**Tree-shaking works better (only?) with functions**, so the idea is to turn `mockValues` and `mockMetadata` into functions. This has benefits even in projects that do need these two structures (e.g. in the demo project), since it means that we can generate them **lazily**, when they are actually needed (e.g. when navigating to the mock demo).

I had tried to generate the mock metadata and values lazily before, but never really got to a point that I found satisfactory. The API to access or generate entities and arrays that we use in our Jest test suites and in the Storybook was always quite convoluted.

This time, I had the idea of generating ndarrays instead of plain, nested JS arrays in `mockValues`, which unlocked a lot of things, thanks to the simple fact that ndarrays store the shape of their underlying arrays. For instance, this means that we:

- no longer need to hard code the dimensions of our mock arrays everywhere (like `[20, 41]`...) when we generate the mock metadata;
- no longer need to find a mock dataset's metadata (just so we can know its shape) in order to construct an ndarray from its value.

Additionally, since the internal array inside an ndarray is already flattened, we no longer need to deal with this in `MockApi`.

---

The PR is huge particularly because I took the opportunity to clean the metadata a bit and to shorten the metadata utility functions to make the whole thing a bit more readable.


packages/app/src/providers/mock/mock-api.ts
```diff
@@ -75,16 +75,14 @@ export class MockApi extends DataProviderApi {
       await this.cancellableDelay(params);
     }
 
-    assertMockDataset(dataset);
-    const { value: rawValue } = dataset;
-    const value = this.processRawValue(dataset, rawValue);
+    const { value } = dataset;
+    assertDefined(value, 'Expected mock dataset to have value');
```
axelboc, line 79: We don't need to process the value anymore, since we already store flat arrays in the mock metadata (cf. the `array()` metadata utility function).

packages/app/src/vis-packs/core/visualizations.test.ts
```diff
@@ -28,188 +26,211 @@ const mockStore = {
...
-    expect(supportsDataset(strImageDataset, mockStore)).toBe(false);
+    expect(supportsDataset(imageStr, mockStore)).toBe(false);
+  });
+
+  it('should not support dataset with non-array shape', () => {
+    expect(supportsDataset(imageScalar, mockStore)).toBe(false);
+  });
+});
+
+describe('Compound Matrix', () => {
```
axelboc, line 220: We were missing some tests for the `CompoundMatrix` visualization.

packages/shared/src/guards.ts
```diff
@@ -458,7 +458,9 @@ export function isNdArray<T extends Data>(
   return 'data' in arr;
 }
 
-export function isTypedArray<T, U extends TypedArray>(arr: U | T[]): arr is U {
+export function isTypedArray<T extends TypedArray>(
+  arr: T | unknown[],
```
axelboc, line 462: One of the generics was unnecessary.

axelboc (commented):
First round of comments, but I still need to make some changes before the review.

packages/shared/src/mock/utils.ts
```diff
@@ -60,21 +56,3 @@ export function findMockEntity(path: string): ProvidedEntity | undefined {
...
-  assertDefined(entity);
-  assertDataset(entity);
-  assertMockDataset(entity);
-
-  const { value } = entity;
-  assertArray(value);
-  assertArrayShape(entity);
-
-  return ndarray(value.flat(entity.shape.length - 1) as T[], entity.shape);
-}
```
axelboc, line 80: As explained in the PR description, since the factory functions in `mockValues` generate ndarrays, we no longer need this utility, which used to look for a dataset's dimensions in the mock metadata in order to be able to generate an ndarray with the right shape.

packages/shared/src/mock/values.ts
```diff
@@ -1,189 +1,241 @@
...
-  twoD.map((arrOneD) => arrOneD.map((val) => val * multiplier)),
-);
-const fourD = arr4.map((divider) =>
-  threeD.map((arrTwoD) =>
-    arrTwoD.map((arrOneD) => arrOneD.map((val) => Math.sin(val / divider))),
-  ),
-);
+export function cplx(real: number, imag: number): H5WebComplex {
+  return [real, imag];
+}
```
axelboc, line 9: Small utility to generate properly typed `H5WebComplex` arrays - e.g. `cplx(1, 2)` instead of `[1, 2] as H5WebComplex`

packages/shared/src/mock/values.ts
```diff
@@ -1,189 +1,241 @@
...
 
-const rgbFlat = rgb.flat(Infinity) as number[];
+const oneD = () => ndarray(range1().map((val) => val ** 2));
 
-const oneD_cplx = range(1, 11).map((val) => [
-  val * Math.cos((val * 3.14) / 10),
-  val * Math.sin((val * 3.14) / 10),
-]);
+const oneD_bool = () =>
+  ndarray([true, false, false, true, true, true, false, true, false, false]);
```
axelboc, line 24: No need to provide a shape when creating 1D arrays.

packages/shared/src/mock/values.ts
```diff
@@ -1,189 +1,241 @@
...
+          [cplx(2, 0), cplx(1.41, 1.41), cplx(1, 1.73), cplx(0, 2)],
+          [cplx(4, 0), cplx(2.82, 2.82), cplx(2, 3.46), cplx(0, 4)],
+          [cplx(1, 0), cplx(0.71, 0.71), cplx(0.5, 0.87), cplx(0, 1)],
+        ],
+        [
+          [cplx(-1, 0), cplx(0.87, -0.5), cplx(0.92, -0.38), cplx(1, 0)],
+          [cplx(-4, 0), cplx(3.46, -2), cplx(3.7, -1.53), cplx(4, 0)],
+          [cplx(-8, 0), cplx(6.93, -4), cplx(7.39, -3.06), cplx(8, 0)],
+        ],
+      ].flat(2),
```
axelboc, line 156: Note that I'm always specifying the number of levels to flatten to ensure we get the correct `NdArray<...[]>` return type. It doesn't work as well with `Infinity`.

packages/shared/src/mock/values.ts
```diff
@@ -1,189 +1,241 @@
...
+    ndarray(
+      twoD().data.map((v) => v / 2),
+      [20, 41],
+    ),
+  position: () => ndarray([-1, 1, 3]), // pixel boundaries (N + 1)
+  scatter_data: () =>
+    ndarray(range1().map((val) => Math.cos((val * 3.14) / 40))),
+  Y_scatter: () =>
+    ndarray(range1().map((_, i) => ((i % 10) + (i % 5)) / 123_456)),
+} satisfies Record<string, () => NdArray<ArrayValue<DType>>>;
```
axelboc, line 240: Very important: this helps make sure `mockValues` contains only functions and that they all return ndarrays. If `mockValues` were to contain anything else, then we'd get a complicated type error in the `array()` function.

packages/shared/src/models-hdf5.ts
```diff
@@ -74,7 +74,7 @@ export type NumArrayDataset = Dataset<ArrayShape, NumericType>;
 
 export type Shape = ArrayShape | ScalarShape | null;
 export type ArrayShape = number[];
-export type ScalarShape = never[];
+export type ScalarShape = [];
```
axelboc, line 77: I think the empty tuple type makes more sense here (instead of the empty array type) and is perhaps a little stricter. Doesn't make a difference in practice though, so sorry for the noise.

Ideally, `ArrayShape` should be `[number, ...number]` to enforce at least one number and to make sure `assertArrayShape` actually excludes the empty tuple (which it doesn't at the moment, I realised), but this led to some problems... Might investigate another time.

packages/shared/src/mock/metadata-utils.ts
```diff
@@ -184,20 +188,37 @@ export function makeDataset<S extends Shape, T extends DType>(
...
+export function array(
+  name: string,
+  opts: EntityOpts & { type?: DType; valueId?: MockValueId } = {},
+): MockDataset<ArrayShape> {
+  const { type, valueId = name, ...entityOpts } = opts;
+  const arr = mockValues[valueId as MockValueId]();
+
+  return dataset(
+    name,
+    type || guessType(arr.data[0]),
```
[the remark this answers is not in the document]
loichuder: Would be nice to convey that `guessType` does not work for arrays but only for single values. Don't know how though.

packages/shared/src/mock/metadata-utils.ts
```diff
@@ -258,31 +273,29 @@ export function makeSilxStyleAttr(
...
->(
+export function nxData<T extends Record<string, MockDataset<ArrayShape>>>(
   name: string,
   opts: {
-    signal: MockDataset<ArrayShape, NumericType | ComplexType>;
-    errors?: MockDataset<ArrayShape, NumericType>;
-    title?: MockDataset<ScalarShape, StringType>;
+    signal: MockDataset<ArrayShape>;
+    errors?: MockDataset<ArrayShape>;
+    title?: MockDataset<ScalarShape>;
```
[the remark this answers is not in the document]
loichuder: It is fine. It is not like the typing here is strictly (got it?) needed.

loichuder (approved):
Seems all fine to me ! Well done !

packages/app/src/__tests__/MetadataViewer.test.tsx
```diff
@@ -59,7 +59,7 @@ test('inspect empty dataset', async () => {
   const shapeRow = screen.getByRole('row', { name: /^Shape/ });
   const typeRow = screen.getByRole('row', { name: /^Type/ });
   expect(shapeRow).toHaveTextContent(/None/);
-  expect(typeRow).toHaveTextContent(/Integer, 32-bit, little-endian/);
+  expect(typeRow).toHaveTextContent(/Unknown/);
```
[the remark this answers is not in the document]
axelboc: True, true. Well, still a good opportunity to demonstrate handling of unknown types in the metadata viewer, I guess.

packages/shared/src/mock/metadata-utils.ts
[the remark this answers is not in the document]
axelboc: True, but not sure either. Not too worried, though, since it's an internal function.

apps/storybook/src/LineVisScales.stories.tsx
```diff
@@ -1,10 +1,10 @@
 import type { LineVis } from '@h5web/lib';
-import { getMockDataArray, ScaleType } from '@h5web/lib';
+import { mockValues, ScaleType } from '@h5web/lib';
 import type { Meta, StoryObj } from '@storybook/react';
 
 import LineVisStoriesMeta, { Default } from './LineVis.stories';
 
-const dataArrayForXLog = getMockDataArray('/nexus_entry/log_spectrum/X_log');
+const dataArrayForXLog = mockValues.X_log(); // eslint-disable-line new-cap
```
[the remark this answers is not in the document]
axelboc: yep :grin:

PR #1539, opened 2024-01-09
Add mock scalar attributes and other tweaks

axelboc:
Just a few tweaks. See individual commits. I'll move things to `@h5web/app` in my next PR.


packages/app/src/__tests__/MetadataViewer.test.tsx
```diff
@@ -36,12 +36,14 @@ test('inspect scalar dataset', async () => {
...
   expect(column).toBeVisible();
-  expect(nameRow).toHaveTextContent(/scalar_num/);
-  expect(pathRow).toHaveTextContent(/\/entities\/scalar_num/);
-  expect(shapeRow).toHaveTextContent(/Scalar/);
-  expect(typeRow).toHaveTextContent(/Float, 64-bit, little-endian/);
+  expect(nameRow).toHaveTextContent(/scalar_num$/);
+  expect(pathRow).toHaveTextContent(/\/entities\/scalar_num$/);
+  expect(typeRow).toHaveTextContent(/Float, 64-bit, little-endian$/);
+  expect(shapeRow).toHaveTextContent(/Scalar$/);
+  expect(attrRow).toHaveTextContent(/0$/);
```
axelboc, line 46: Adding an expectation for the newly added scalar attribute.

packages/shared/src/mock/metadata.ts
```diff
@@ -29,14 +30,26 @@ export function makeMockFile(): MockFile {
...
+          attributes: [scalarAttr('attr', cplx(1, 5))],
+        }),
+        scalar('scalar_compound', ['foo', 2], {
+          type: compoundType({ str: strType(), int: intType(8) }),
+          attributes: [
+            scalarAttr('attr', ['foo', 2], {
+              type: compoundType({ str: strType(), int: intType(8) }),
+            }),
+          ],
+        }),
```
axelboc, line 52: Adding a scalar compound dataset. It's currently rendered with the _Raw_ visualization as `["foo",2]`, so the idea will be to improve on that at some point.

packages/shared/src/mock/metadata.ts
```diff
@@ -29,14 +30,26 @@ export function makeMockFile(): MockFile {
...
+          attributes: [scalarAttr('attr', cplx(1, 5))],
+        }),
+        scalar('scalar_compound', ['foo', 2], {
+          type: compoundType({ str: strType('ASCII', 3), int: intType(8) }),
+          attributes: [
+            scalarAttr('attr', ['foo', 2], {
+              type: compoundType({ str: strType('UTF-8'), int: intType(8) }),
+            }),
+          ],
+        }),
```
axelboc, line 52: Adding a scalar compound dataset. It's currently rendered with the Raw visualization as ["foo",2], so the idea will be to improve on that at some point.

packages/app/src/__tests__/MetadataViewer.test.tsx
loichuder: `0` is kind of a special value, we should perhaps use another number ? :shrug:

loichuder (approved):

PR #1540, opened 2024-01-09
Rename shared modules

axelboc:
First, I rename the models modules so that the namespace that matters most becomes the prefix:

- `models-hdf5.ts` => `hdf5-models.ts`
- `models-nexus.ts` => `nexus-models.ts`
- `models-vis.ts` => `vis-models.ts`

This is more in line with how we've named other common files in the repo, like `metadata-utils`, `test-utils`, `plugin-utils`, etc.

Then I rename `utils.ts` (at the root of the shared package, still) to `vis-utils.ts` and move two utilities (`getChildEntity` and `buildEntityPath`) into a new file called `hdf5-utils.ts`.

So now we have a clearer structure: `hdf5-models`+`hdf5-utils` and `vis-models`+`vis-utils`.

That's a big diff, obviously, so I'll stop here for this PR. Next up, I'm planning on moving some of the generic metadata utilities (like `dataset()`, `attr()`, `intType()`, etc.) to the new `hdf5-utils` so I can use them in the providers (including in `@h5web/h5wasm`, which is why they must remain in `@h5web/shared`). The utilities that are dedicated to generating mock metadata (`scalar()`, `array()`, etc.) will then move to the `@h5web/app` package, along with `makeMockFile()`.


loichuder (approved):
Sounds sensible to me. You can go forward with your plan.

PR #1541, opened 2024-01-11
Move shared mock files and utilities

axelboc:
I've made separate commits, but in summary:

- `mock/<models,utils,values>.ts` move to root of shared package: `mock-<models,utils,values>.ts`
- `makeMockFile` and `findMockEntity` move to the app package
- All the mock metadata utilities remain in the shared package after all (in `mock-utils.ts`), but the dtype utilities move to `hdf5-utils.ts`. The dtype utilities are very useful and entirely generic; I will use them in the providers in my next PR. The remaining utilities are really specific to generating mock entities because they do stuff with the paths (to avoid specifying the full path of every entity in `makeMockFile`); also, they are used here and there in a few tests, like for testing `getChildEntity` in the shared package, which is why I didn't move them to the app package.
- The little `cplx()` utility also moves to `hdf5-utils.ts`.
- I introduce a new `arrayAttr()` utility to create array attributes; it accepts an array value and infers the dtype from it.


packages/app/src/providers/mock/mock-api.ts
```diff
@@ -10,25 +9,27 @@ import type {
...
 import type { ExportFormat, ExportURL, ValuesStoreParams } from '../models';
 import { sliceValue } from '../utils';
+import { makeMockFile } from './mock-file';
+import { findMockEntity } from './utils';
 
 export const SLOW_TIMEOUT = 3000;
 
 export class MockApi extends DataProviderApi {
-  private readonly mockFile: MockFile;
+  private readonly mockFile: GroupWithChildren;
```
axelboc, line 32: I removed this alias after all, since it's only used here and in `makeMockFile`, and since the latter is now in this folder as well, I didn't fee like creating a `models.ts` file just for that... :sweat_smile:

packages/shared/src/mock-utils.ts
```diff
@@ -1,118 +1,52 @@
...
   }
-
-  return unknownType();
 }
 
 /* ---------------------- */
 /* ----- ATTRIBUTES ----- */
 
-export function attr<S extends Shape, T extends DType>(
+export function attribute<S extends Shape, T extends DType>(
```
axelboc, line 49: Using the full word for the base utility, in line with `dataset()`, `datatype()` and `group()`, which allows me to use the abbreviation for a variable above.

packages/app/src/providers/mock/utils.ts
```diff
@@ -3,33 +3,12 @@ import {
...
-  }
-}
-
-export function assertMockAttribute<S extends Shape, T extends DType>(
-  attribute: Attribute<S, T>,
-): asserts attribute is MockAttribute<S, T> {
-  if (!('value' in attribute)) {
-    throw new Error('Expected mock attribute');
-  }
-}
```
axelboc, line 32: These move to `mock-utils.ts`.

loichuder (approved):

PR #1542, opened 2024-01-11
Use dtype utils in providers + add utilities to create `ArrayType` and `EnumType`

axelboc:


packages/app/src/providers/hsds/utils.ts
```diff
@@ -90,19 +102,25 @@ function convertHsdsCompoundType(
...
-  };
+  );
+}
+
+function convertHsdsEnumType(hsdsType: HsdsEnumType): EnumType | BooleanType {
+  const { base, mapping } = hsdsType;
+  assertHsdsNumericType(base);
+
+  const type = enumType(convertHsdsNumericType(base), mapping);
+  return isBoolEnumType(type) ? boolType() : type; // booleans stored as enums by h5py
```
axelboc, line 123: I've also added a utility to abstract away from the providers the check of whether an enum dtype corresponds to the boolean dtype.

packages/app/src/providers/hsds/utils.ts
```diff
@@ -115,37 +133,25 @@ export function convertHsdsType(hsdsType: HsdsType): DType {
...
+      );
 
     case 'H5T_ARRAY':
     case 'H5T_VLEN':
       return {
         class:
           hsdsType.class === 'H5T_ARRAY' ? DTypeClass.Array : DTypeClass.VLen,
         base: convertHsdsType(hsdsType.base),
-        dims: hsdsType.dims,
+        ...(hsdsType.dims && { dims: hsdsType.dims }),
```
axelboc, line 147: We don't want `dims: undefined`. This was caught by switching from `toEqual` to `toStrictEqual` in the tests.

packages/app/src/providers/mock/mock-file.ts
```diff
@@ -52,6 +54,26 @@ export function makeMockFile(): GroupWithChildren {
...
+          ],
+        }),
+        scalar('scalar_enum', 2, {
+          type: enumType(intType(8, true), { FOO: 0, BAR: 1, BAZ: 2 }),
+          attributes: [
+            scalarAttr('attr', 2, {
+              type: enumType(intType(8, true), { FOO: 0, BAR: 1, BAZ: 2 }),
+            }),
+          ],
+        }),
```
axelboc, line 76: Adding some mock scalar datasets to test `ArrayType` (with fixed or variable length), and `EnumType`.

packages/shared/src/hdf5-utils.ts
```diff
@@ -33,7 +35,7 @@ export function buildEntityPath(
 /* ----- TYPES ----- */
 
 export function intType(
-  size: 8 | 16 | 32 | 64 = 32,
+  size = 32,
```
axelboc, line 38: Had to loosen this up to allow creating numeric types from computed sizes - i.e. `length * 8`.

packages/app/src/providers/h5grove/utils.test.ts
```diff
@@ -1,100 +1,69 @@
...
-      size: 32,
-      endianness: Endianness.LE,
-    });
-    expect(convertH5GroveDtype('>u8')).toEqual({
-      class: DTypeClass.Unsigned,
-      size: 64,
-      endianness: Endianness.BE,
-    });
+    expect(convertH5GroveDtype('<i4')).toStrictEqual(
+      intType(32, false, Endianness.LE),
```
loichuder, line 17: Not a huge fan of the fact we have to inspect `intType` to know what this `false/true` is for.

Add a helper `uintType` instead and keep `intType` for signed integers ?

packages/app/src/providers/hsds/utils.ts
loichuder: :fishing_pole_and_fish:

loichuder (approved):

packages/app/src/providers/h5grove/utils.test.ts
axelboc: I hesitated but you're right, let's do that.

axelboc:
- `intType` vs `uintType`: https://github.com/silx-kit/h5web/pull/1542/commits/4b7273115ce6257866941ada8d966c5ce2cc12f7
- I've also pushed a commit to strengthen `CompoundType` https://github.com/silx-kit/h5web/pull/1542/commits/efb0d541946da8df471244fc5377f259ec81c0c8

packages/shared/src/hdf5-utils.ts
```diff
@@ -73,14 +67,41 @@ export function cplxType(
...
   return { class: DTypeClass.Compound, fields };
 }
 
-export function printableCompoundType(
-  fields: Record<string, PrintableType>,
-): PrintableCompoundType {
-  return { class: DTypeClass.Compound, fields };
+export const printableCompoundType = compoundType<
+  Record<string, PrintableType>
+>;
```
axelboc, line 78: Taking advantage of a quite recent TypeScript syntax here: https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#instantiation-expressions

loichuder (approved):

PR #1546, opened 2024-01-15
Sync package versions when bumping monorepo version to simplify CI release workflow

axelboc:
The main purpose of this PR (which stems from #1543, which I'll rebase on this one) is to be able to pack a package into a tarball without having to worry about its version (i.e. without ending up with a packed package that has version `0.0.1`).

To do so, I add a `version` script (which runs as part of `pnpm version`) that copies the bumped version number of the monorepo into each package's `package.json`. This allows those `package.json` files to be committed with their real version numbers instead of the placeholder `0.0.1`.

As a side benefit, the _Publish packages_ CI workflow is now a lot shorter and less magical: it just builds the packages and publishes them!

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -48,38 +48,8 @@ jobs:
       - name: Build packages 📦
         run: pnpm packages
 
-      - name: Set @h5web/lib version 📌
-        uses: reedyuk/npm-version@1.1.1
-        with:
-          version: ${{ steps.packageVersion.outputs.PACKAGE_VERSION }}
-          package: 'packages/lib'
-
-      - name: Set @h5web/app version 📌
-        uses: reedyuk/npm-version@1.1.1
-        with:
-          version: ${{ steps.packageVersion.outputs.PACKAGE_VERSION }}
-          package: 'packages/app'
-
-      - name: Set @h5web/h5wasm version 📌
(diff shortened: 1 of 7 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -48,38 +48,8 @@ jobs:
...
-      - name: Publish @h5web/app 🥳
-        run: cd packages/app && pnpm publish --access public --provenance --no-git-checks --tag $NPM_TAG
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-          NPM_TAG: "${{ contains(steps.packageVersion.outputs.PACKAGE_VERSION, 'beta') && 'next' || 'latest' }}"
-
-      - name: Publish @h5web/h5wasm 🥳
-        run: cd packages/h5wasm && pnpm publish --access public --provenance --no-git-checks --tag $NPM_TAG
+      - name: Publish packages 🥳
+        run: pnpm -r publish --access public --provenance --no-git-checks --tag $NPM_TAG
```
axelboc, line 52: Just discovered that `pnpm publish` supports the `--recursive` flag. I've checked: it runs only on public packages (without `private: true` in their `package.json`), so it doesn't try to publish the shared package, the Storybook and the demo. :+1:

loichuder (approved):

PR #1547, opened 2024-01-15
Link to artifact URL from PR when visual diffing fails

axelboc:
The [upload-artifact](https://github.com/actions/upload-artifact) action finally exposes an `artifact-url` output! This was probably [their most requested features](https://github.com/actions/upload-artifact/issues/50).

This means we can provide a link to download the artifacts from the PR when visual diffing fails. I do this using the [create-or-update-comment](https://github.com/peter-evans/create-or-update-comment) action.


axelboc:
Arf, I can't get this to work because of the fact that we skip `pull_request` events on local PRs. So I have access to the `artifact-url` only in jobs triggered by the `push` event, which don't give easy access to the PR number (since it may not exist yet).

PR #1548, opened 2024-01-15
Check out Prettier's experimental ternary formatting

axelboc:
Prettier is asking for feedback on its ternary formatting experimentation. All the context here: https://prettier.io/blog/2023/11/13/curious-ternaries

I don't intend to merge this. Just thought you might be interested in giving your own input @loichuder.

### DIFF
--- a/apps/demo/src/h5wasm/UrlForm.tsx
+++ b/apps/demo/src/h5wasm/UrlForm.tsx
@@ -63,19 +63,17 @@ function UrlForm(props: Props) {
             required
           />
           <button id="h5wasm-url" className={styles.urlBtn} type="submit">
-            {isLoading ? (
+            {isLoading ?
               <FiLoader className={styles.urlLoader} aria-label="Loading..." />
-            ) : (
-              'Go'
-            )}
+            : 'Go'}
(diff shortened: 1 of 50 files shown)


loichuder (approved):
I always advised against the use of shared ternaries in my reviews but if one has to do it, the new syntax is just great :heart_eyes: 

Look how clean this one looks: https://github.com/silx-kit/h5web/pull/1548/files#diff-0cbc21f1d5715550ea2a6d0f46e110e651f2b5492bf51629f298388e4f962b65R172

axelboc:
I quite like the fact that the question mark is closer to the condition; it makes a lot of sense. I also like the indentation of the then/else clauses.

Only downside for me is that the condition tends to move to (or stay on) a new line more often than it needs to. So it's not really a problem with the new ternary syntax, but more with how ternaries integrate with the code that surrounds them. Like here and the one just after: https://github.com/silx-kit/h5web/pull/1548/files?diff=split&w=0#diff-0217bbf7bde59b0a50e0c55dca473df1dff3275ed66a4480fb69e34f596b665eR70-R73

PR #1549, opened 2024-01-16
Switch monorepo root to ESM

axelboc:
Can't remember what was holding us back on this. Maybe nothing, maybe Jest... :shrug: I've checked things thoroughly and can't see any problems.

Hopefully ESLint will release its [v9](https://eslint.org/blog/2023/11/whats-coming-in-eslint-9.0.0/) soon, as I think it will support writing the config as ESM instead of CommonJS. For now, it's still not supported: https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file-formats


loichuder (approved):
This should not change anything for the h5web packages, right ?

axelboc:
> This should not change anything for the h5web packages, right ?

Nope!

PR #1550, opened 2024-01-19
Test `H5WasmApi`

axelboc:
The goal of this PR is to test the low-level `H5WasmApi` class — more specifically, **the metadata and values it returns for HDF5 datasets with various types and shapes**.

To do so, I've written a Python script (inspired from [silx' script](https://github.com/silx-kit/silx/blob/d8a666dcfab4f071f38615ac495937cf85e97f95/tools/create_h5_sample.py)) to generate a sample HDF5 file with h5py. In this file, I try to generate HDF5 datasets of as many kinds as I can (`H5T_INTEGER`, `H5T_COMPOUND`, etc.) and with as many interesting dtype combinations as I can think of (`int8`, `complex64`, etc.).

The idea is to be able to test:
- that h5wasm returns the values as accurately as possible (without loss of precision, type coercion, etc.);
- that h5wasm returns the metadata as accurately as possible (with all the information we need to parse it correctly);
- that `H5WasmApi` doesn't interfere with the values in unwanted ways;
- that `H5WasmApi` creates our internal `DType` objects from the raw metadata accurately.

More specifically, this test will help identify:
- values that are clamped/rounded/unsupported by h5wasm and/or H5Web;
- dtypes that are currently unknown/unsupported by h5wasm and/or H5Web;
- inconsistencies between data providers in the values/metadata they provide;
- inaccuracies in the way we generate mock data for testing purposes.

> I purposely did not to add any attributes to the sample file, as they didn't seem to add any testing value (the dtype parsing is the same, and h5wasm reads/processes datasets and attributes values the same way ... I hope). I also did not add groups, links, datatypes, and NeXus/RGB/scatter data as this falls out of the scope of what I'm wanting to test (at least for the moment).

You can explore the sample file here: https://h5web.panosc.eu/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Fsample.h5

The test itself consists in a single, very basic **snapshot test**:

1. I start by loading the sample file, initialising the API and calling `api.getEntity('/')` to get the metadata of the root group and its children;
2. Then, for each child dataset, I get the metadata I'm interested in and request the dataset's value with `api.getValue({ dataset })`. (I've kept the sample file flat to keep this test code as concise as possible.)
3. Finally, I take a snapshot of this array of dataset metadata+values. If the snapshot matches the existing, committed snapshot, then the test passes; otherwise it fails.


packages/h5wasm/create_h5_sample.py
```diff
@@ -0,0 +1,215 @@
...
+        "enum_int32",
+        256,
+        h5py.enum_dtype({"A": 256, "B": 257}, np.int32),
+    )
+
+    # === H5T_VLEN ===
+
+    # https://docs.h5py.org/en/stable/special.html#arbitrary-vlen-data
+    scalar_vlen = add_scalar(h5, "vlen_int8", dtype=h5py.vlen_dtype(np.int8))
+    scalar_vlen[()] = [0, 1]
```
axelboc, line 211: I'm trying to create a scalar dataset with an `H5T_VLEN` type (itself with an `H5T_INTEGER` base type). Not 100% this is what I'm doing above...

packages/h5wasm/create_h5_sample.py
```diff
@@ -0,0 +1,215 @@
...
+        "ascii_fixed",
+        np.string_("Some text"),
+    )
+    add_scalar(h5, "utf8_vlen", "Some text")
+    add_array(h5, "utf8_vlen", np.array(["foo", "bar", "baz"], h5py.string_dtype()))
+    add_scalar(h5, "utf8_fixed", "Some text", h5py.string_dtype("utf-8", 9))
+
+    # === H5T_BITFIELD ===
+
+    # h5py syntax unknown
```
axelboc, line 108: I've you know, let me know :wink:

packages/h5wasm/create_h5_sample.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # === H5T_VLEN ===
+
+    # https://docs.h5py.org/en/stable/special.html#arbitrary-vlen-data
+    scalar_vlen = add_scalar(h5, "vlen_int8", dtype=h5py.vlen_dtype(np.int8))
+    scalar_vlen[()] = [0, 1]
+
+    # === H5T_ARRAY ===
+
+    # h5py syntax unknown
```
axelboc, line 215: I've you know, let me know :wink:

packages/h5wasm/src/h5wasm-api.test.ts
```diff
@@ -0,0 +1,64 @@
...
+const REMOTE_TEST_FILE = 'http://www.silx.org/pub/h5web/sample.h5'; // `https` would complicate things...
+
+async function loadTestFile(): Promise<ArrayBuffer> {
+  if (existsSync(LOCAL_TEST_FILE)) {
+    return readFile(LOCAL_TEST_FILE);
+  }
+
+  const resp = await fetch(REMOTE_TEST_FILE);
+  return resp.arrayBuffer();
+}
```
axelboc, line 29: The test first looks for the sample file at `packages/h5wasm/dist-h5/sample.h5`, as this is where the Python script creates it. If the sample file hasn't been generated locally, then the test will fetch it from silx.org.

packages/h5wasm/src/h5wasm-api.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+async function getValueOrError(
+  api: H5WasmApi,
+  dataset: Dataset<ArrayShape | ScalarShape>,
+): Promise<unknown> {
+  try {
+    return await api.getValue({ dataset });
+  } catch (error) {
+    return error;
```
axelboc, line 38: I want the error to appear in the snapshot file.

packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -256,16 +256,18 @@ export class H5WasmApi extends DataProviderApi {
...
         kind: EntityKind.Dataset,
-        attributes: this.processH5WasmAttrs(h5wEntity.attrs),
-        shape,
+        shape: metadata.shape,
         type: convertMetadataToDType(metadata),
-        rawType: dtype,
+        chunks: metadata.chunks ?? undefined,
         filters,
+        attributes: this.processH5WasmAttrs(h5wEntity.attrs),
+        rawType,
```
axelboc, line 270: I'm now storing the full raw type metadata instead of just h5wasm's `dtype` string. This ensures we have access to all the raw information in the snapshot file and in the metadata viewer (to check that we haven't made a mistake or lost something in our `DType` objects).

packages/h5wasm/package.json
```diff
@@ -60,6 +60,7 @@
     "eslint": "8.56.0",
     "eslint-config-galex": "4.5.2",
     "react": "18.2.0",
+    "react-dom": "18.2.0",
```
axelboc, line 63: I had an error when running the test. `react-dom` is a peer dep of `@h5web/app`, which is a peer dep of `@h5web/h5wasm`, so `react-dom` has to be installed as a dev dep in `@h5web/h5wasm`.

packages/h5wasm/src/__snapshots__/h5wasm-api.test.ts.snap
```diff
@@ -0,0 +1,1918 @@
...
+      "littleEndian": false,
+      "signed": false,
+      "size": 8,
+      "total_size": 1,
+      "type": 7,
+      "vlen": false,
+    },
+    "shape": [],
+    "type": {
+      "class": "Unknown",
```
axelboc, line 1672: We can get some easy wins in the parsing of the dtype.

packages/h5wasm/src/__snapshots__/h5wasm-api.test.ts.snap
```diff
@@ -0,0 +1,1918 @@
...
+      "type": 0,
+      "vlen": false,
+    },
+    "shape": [],
+    "type": {
+      "class": "Integer",
+      "endianness": "little-endian",
+      "size": 64,
+    },
+    "value": -9223372036854776000,
```
axelboc, line 185: The exact value in the file is -9,223,372,036,854,775,808, so there's a bit of rounding happening somewhere. There are other cases like this of course.

packages/h5wasm/src/__snapshots__/h5wasm-api.test.ts.snap
```diff
@@ -0,0 +1,1918 @@
...
+      "type": 1,
+      "vlen": false,
+    },
+    "shape": [],
+    "type": {
+      "class": "Float",
+      "endianness": "little-endian",
+      "size": 16,
+    },
+    "value": [Error: Float16 not supported],
```
axelboc, line 428: Another opportunity for improvement.

loichuder (approved):

packages/h5wasm/create_h5_sample.py
```diff
@@ -0,0 +1,248 @@
...
+                    np.array([0, 1], np.uint64),
+                ),
+                (
+                    np.array([4, 5], np.float32),
+                    np.array([0, 1, 2], np.uint64),
+                ),
+            ],
+            [("arr", np.float32, (2,)), ("vlen", h5py.vlen_dtype(np.uint64))],
+        ),
+    )
```
axelboc, line 198: I've managed to unlock the correct syntax for creating a compound with an `H5T_ARRAY` and an `H5T_VLEN` field. :tada:

packages/h5wasm/create_h5_sample.py
```diff
@@ -0,0 +1,248 @@
...
+    scalar_vlen = add_scalar(h5, "vlen_int8", dtype=h5py.vlen_dtype(np.int8))
+    scalar_vlen[()] = [0, 1]
+
+    # https://docs.h5py.org/en/stable/special.html#arbitrary-vlen-data
+    scalar_vlen = add_array(
+        h5, "vlen_int64", shape=(3,), dtype=h5py.vlen_dtype(np.int64)
+    )
+    scalar_vlen[0] = [0]
+    scalar_vlen[1] = [0, 1]
+    scalar_vlen[2] = [0, 1, 2]
```
axelboc, line 244: I've also added a 1D `H5T_VLEN` dataset.

PR #1552, opened 2024-01-23
Test `H5GroveApi`

axelboc:
Diffing the two files (after removing the `rawType` keys) is quite enlightening: https://www.diffchecker.com/0lEq6wKG/ (`H5GroveApi` on the left, `H5WasmApi` on the right).

I can already see some obvious bugs, some potential quick wins, and some more complicated inconsistencies.

### DIFF
--- a/packages/h5wasm/src/h5wasm-api.test.ts
+++ b/packages/h5wasm/src/h5wasm-api.test.ts
@@ -1,3 +1,7 @@
+import { existsSync } from 'node:fs';
+import { readFile } from 'node:fs/promises';
+import path from 'node:path';
+
 import {
   assertDataset,
   assertGroup,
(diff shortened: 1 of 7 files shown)


packages/h5wasm/src/h5wasm-api.test.ts
```diff
@@ -1,22 +1,19 @@
...
-} from '@h5web/shared/hdf5-models';
-import { existsSync } from 'fs';
-import { readFile } from 'fs/promises';
-import path from 'path';
 import { expect, test } from 'vitest';
 
 import { H5WasmApi } from './h5wasm-api';
 
-const LOCAL_TEST_FILE = path.resolve(__dirname, '../dist-h5/sample.h5');
+const LOCAL_TEST_FILE = path.resolve(process.cwd(), 'dist/sample.h5');
```
axelboc, line 16: I've moved the Python script to the root of the monorepo. The sample file is now generated into a `dist` folder also at the root. I've considered moving it to the shared package, but it didn't feel right, because the sample file is only imported from `@h5web/h5wasm` (and only when it exists, otherwise it falls back to fetching it from silx.org).

I preferred to make it clear that the script and sample file are part of the set up of the repository as a whole and not of a specific project.

loichuder (approved):

PR #1555, opened 2024-01-26
Refactor providers to notably remove unneeded type guards

axelboc:
A bit of clean-up before I start improving the parsing of the dtypes in both `H5WasmApi` and `H5GroveApi`. Mostly, I remove a bunch of type guards that are not necessary and I move some functional utilities out of the classes. I recommend reviewing commit by commit.


loichuder (approved):
Nice !

PR #1556, opened 2024-01-29
Improve dtype parsing with h5wasm

axelboc:
See test snapshot for detailed improvements. I'm also refactoring the `parseDType` function to remove the need for type guards.

### DIFF
--- a/packages/h5wasm/src/models.ts
+++ b/packages/h5wasm/src/models.ts
@@ -1,22 +1,5 @@
-import type {
-  CompoundTypeMetadata,
-  EnumTypeMetadata,
-  Group as H5WasmGroup,
-  Metadata,
-} from 'h5wasm';
+import type { Group as H5WasmGroup } from 'h5wasm';
 
 export type H5WasmEntity = ReturnType<H5WasmGroup['get']>;
 
 export type H5WasmAttributes = H5WasmGroup['attrs'];
-
-export interface CompoundMetadata extends Metadata {
-  compound_type: CompoundTypeMetadata;
-}
-
-export interface NumericMetadata extends Metadata {
-  type: 0 | 1;
-}
-
-export interface EnumMetadata extends Metadata {
-  enum_type: EnumTypeMetadata;
-}
--- a/packages/h5wasm/src/utils.ts
+++ b/packages/h5wasm/src/utils.ts
@@ -1,26 +1,26 @@
-import { assertDefined } from '@h5web/shared/guards';
+import { assertDefined, isNumericType } from '@h5web/shared/guards';
 import type {
   Attribute,
   ChildEntity,
   DType,
   Group,
-  NumericType,
   ProvidedEntity,
   Shape,
 } from '@h5web/shared/hdf5-models';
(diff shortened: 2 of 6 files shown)


packages/h5wasm/src/models.ts
```diff
@@ -1,22 +1,5 @@
...
-  Metadata,
-} from 'h5wasm';
+import type { Group as H5WasmGroup } from 'h5wasm';
 
 export type H5WasmEntity = ReturnType<H5WasmGroup['get']>;
 
 export type H5WasmAttributes = H5WasmGroup['attrs'];
-
-export interface CompoundMetadata extends Metadata {
-  compound_type: CompoundTypeMetadata;
```
axelboc, line 13: I now simply use `assertDefined` to make sure that `compound_type` exists when the HDF5 type `class` is `H5T_COMPOUND`.

packages/h5wasm/src/utils.ts
```diff
@@ -162,83 +152,86 @@ function parseAttributes(h5wAttrs: H5WasmAttributes): Attribute[] {
...
-): NumericType {
-  const { signed, size: length, littleEndian } = metadata;
-  const size = length * 8;
-  const endianness = littleEndian ? Endianness.LE : Endianness.BE;
+export function parseDType(metadata: Metadata): DType {
+  const { type: h5tClass, size } = metadata;
 
-  if (isIntegerMetadata(metadata)) {
-    return signed ? intType(size, endianness) : uintType(size, endianness);
+  if (h5tClass === H5TClass.Integer) {
```
axelboc, line 158: Introducing this `H5TClass` enum, which I'll reuse for h5grove.

packages/h5wasm/src/utils.ts
```diff
@@ -162,83 +152,86 @@ function parseAttributes(h5wAttrs: H5WasmAttributes): Attribute[] {
...
+    const { signed, littleEndian } = metadata;
+    return intOrUintType(signed, size * 8, toEndianness(littleEndian));
```
axelboc, line 160: I'm also adding a few utilities like this to keep this `parseDType` function concise (and share the logic with h5grove later). The other two are `compoundOrCplxType()` and `enumOrBoolType()`

packages/h5wasm/src/models.ts
loichuder: I guess we could we keep it and add the field `type: H5TClass.Compound` to avoid resorting to a type guard.

loichuder (approved):

packages/h5wasm/src/models.ts
axelboc: H5Wasm's `Metadata` interface has all the fields and some of them (like `compound_type`) are optional. To avoid `assertDefined`, I would have to turn this `Metadata` type into a discriminated union (`CompoundMetadata | EnumMetadata | ...`), which is what I'm planning to do for h5grove. In h5wasm, however, there's typically one field that is different for each type (`compound_type`, `enum_type`, etc.), so `assertDefined` is much simpler I think.

Issue #1554, opened 2024-01-26
Visualize pictures stored as binary (PNG, JPEG, etc.)


axelboc:
### Discussed in https://github.com/silx-kit/h5web/discussions/1551

<div type='discussions-op-text'>

<sup>Originally posted by **Flydroid** January 23, 2024</sup>
Hi, 

I have been evaluating different options to store images in an HDF5 file. 
Because build-in methods come with a large file size, we are storing png files as binary data (u8 array) in HDF5. 
What would be really cool is, if h5web would support the decoding of the PNG data to then display the image. 

Unfortunately i have don't have typescript experience to really contribute directly.
</div>
<br>

- The goal would be to support the most common picture formats supported by modern browsers (at least PNG and JPEG), as the decoding would be left to the browser via the `img` tag.
- The pictures binary data would have to be stored as `H5T_OPAQUE` datasets in HDF5 in order to trigger the appropriate visualization in H5Web. I'm still debating whether this should be a new visualization called _Opaque_ or  the existing _Raw_ visualization but I'm tipping towards the latter.
- If the visualization detects the [magic numbers](https://en.wikipedia.org/wiki/List_of_file_signatures) of a known picture format, it generates a blob URL from the binary data and pass it to an `img` element's `src` attribute so the browser can display the image.
- Ideally, we should also support displaying stacks of pictures stored as multi-dimensional opaque datasets (i.e. one image at a time, with sliders on the left).

PR #1560, opened 2024-02-01
Add byte string array and float128 max value to sample HDF5 file

axelboc:


packages/h5wasm/src/__snapshots__/h5wasm-api.test.ts.snap
```diff
@@ -1745,8 +1769,8 @@ exports[`test file matches snapshot 1`] = `
           0,
           0,
           0,
-          80,
-          191,
+          72,
+          203,
```
axelboc, line 1773: I guess this is a side-effect of adding a new dataset to the file.

create_h5_sample.py
```diff
@@ -113,6 +117,11 @@ def print_h5t_class(dataset):
     # === H5T_OPAQUE ===
 
     add_scalar(h5, "byte_string", np.void(b"\x00\x11\x22"))
+    add_array(
+        h5,
+        "byte_string",
+        np.array([np.void(b"\x00"), np.void(b"\x11"), np.void(b"\x22")]),
+    )
```
axelboc, line 124: Adding this as a sanity check in preparation for #1554

packages/h5wasm/src/__snapshots__/h5wasm-api.test.ts.snap
loichuder: Wait, what ? How adding new datasets changes the value of existing ones ?

axelboc: I think those are basically buffer addresses in the file.

loichuder: Oh got it :+1:

loichuder (approved):

PR #1561, opened 2024-02-01
Fetch float16 and float128 as binary using `dtype=safe`

axelboc:


packages/app/src/providers/h5grove/__snapshots__/h5grove-api.test.ts.snap
```diff
@@ -670,13 +670,13 @@ exports[`test file matches snapshot 1`] = `
...
     },
-    "value": [
+    "value": Float64Array [
       0,
       1,
       2,
       3,
       4,
-      null,
+      Infinity,
```
axelboc, line 679: The maximum finite float128 value gets converted to float64 `Infinity`. I would have expected Numpy to clamp it to the maximum finite float64 value instead. :thinking: 

Now that I think of it, a similar thing happens with the smallest float128 value greater than 0 (the value of the `float128_scalar` dataset in the sample file) => h5grove sends 0 in float64 — not the smallest float64 value greater than 0, as I would have expected.

vallsv: > The maximum finite float128 value gets converted to float64 Infinity

Sounds like a safe behaviour. It would be weird to convert a finite value to another while you know it's not the same value.

axelboc: > It would be weird to convert a finite value to another while you know it's not the same value.

That's exactly what happens when you cast an int64 to int32, though. Obviously `inf` cannot be represented in int32, but still, in float, it's a special value... The rounding to 0 makes more sense in that regard.

I'm just asking in case the code in h5grove is wrong somehow — maybe the float128 are cast by Python before numpy can convert them properly, or something of the sort. :shrug:

axelboc:
> Sorry, took a while to get back to this and I would need some context to refresh my mind.
> 
> Why using `dtype=safe` makes it possible to fetch float16 and float128 as binary ?

No worries!

There's no `Float16Array` and `Float128Array` in JS, so if we fetched float16/128 datasets with `format=bin` without `dtype=safe`, we'd get binary we don't understand (at least without using custom libraries like https://github.com/petamoriken/float16 or https://github.com/munrocket/double.js).

When fetching float16/128 datasets with `format=bin` and `dtype=safe`, h5grove [converts them to float32/64](https://github.com/silx-kit/h5grove/blob/main/h5grove/utils.py#L207-L213) respectively. With this knowledge, we can adjust our provider logic to say that when fetching float16, the response buffer should be passed into a `Float32Array` and when fetching float128, into a `Float64Array`. (Before, we were saying that float16/128 datasets had no corresponding typed array, which led to fetching them as JSON).

packages/app/src/providers/h5grove/utils.ts
```diff
@@ -198,3 +204,24 @@ export function parseDType(type: H5GroveType): DType {
...
+
+  if (dtypeClass === DTypeClass.Float && dtype.size === 16) {
+    return Float32Array; // also saves a number[]->Float32Array conversion in texture-based visualizations
+  }
+
+  if (dtypeClass === DTypeClass.Float && dtype.size === 128) {
+    return Float64Array;
+  }
+
+  return undefined;
```
loichuder, line 226: It might be cleaner to `return typedArrayFromDType(type)` here to avoid the `||` in `h5grove-api.ts`.

loichuder (approved):

packages/app/src/providers/h5grove/utils.ts
axelboc: Ooh clever

PR #1571, opened 2024-02-09
Upgrade deps and remove lodash

axelboc:
Big batch of updates, in no particular order:

- h5wasm [v0.7.1](https://github.com/usnistgov/h5wasm/releases/tag/v0.7.1) with support for object and region references.
- Wouter [stable v3](https://github.com/molefrog/wouter/releases/tag/v3.0.0) (no notable changes since RC)
- Storybook [from 7.6.4 to 7.6.13](https://github.com/storybookjs/storybook/releases) with a fix for a pre-transform error with Vite 5
- React Three Fiber [from 8.15.12 to 8.15.16](https://github.com/pmndrs/react-three-fiber/releases)
- @react-three/drei [from 9.92.3 to 9.97.3](https://github.com/pmndrs/drei/releases)
- DOM Testing Library [9.3.4](https://github.com/testing-library/dom-testing-library/releases/tag/v9.3.4)
- @testing-library/jest-dom [from 6.1.5 to 6.4.2](https://github.com/testing-library/jest-dom/releases) with a new `toHaveRole` matcher that could be useful
-  @testing-library/react [14.2.0 and 14.2.1](https://github.com/testing-library/react-testing-library/releases) with an option to enable strict mode render (but I don't think I'm going to enable it, since the goal is to test as closely to production as possible)
- User Event [14.5.2](https://github.com/testing-library/user-event/releases/tag/v14.5.2)
- Axios [from 1.6.2 to 1.6.7](https://github.com/axios/axios/releases) with better async error stack traces
- Cypress [from 13.6.1 to 13.6.4](https://docs.cypress.io/guides/references/changelog), which allows removing the `module: "bundler"` workaround in `tsconfig.json`.
- nanoid [5.0.5](https://github.com/ai/nanoid/blob/main/CHANGELOG.md#505)
- react-error-boundary [4.0.12](https://github.com/bvaughn/react-error-boundary/releases/tag/4.0.12) with support for `null` fallback prop (useful for the error boundary of the `NX` badges in the explorer)
- rollup [from 4.9.0 to 4.9.6](https://github.com/rollup/rollup/releases)
- @visx/axis [from 3.5.0 to 3.8.0](https://github.com/airbnb/visx/blob/master/CHANGELOG.md) ‑ had to change the way we reset the axis label styles from visx
- Prettier [from 3.1.1 to 3.2.5](https://github.com/prettier/prettier/blob/main/CHANGELOG.md)
- @react-hookz/web [24.0.<0,1,2,3,4>](https://github.com/react-hookz/web/releases), now ESM only
- zustand [4.5.0](https://github.com/oslabs-beta/Zusty) with a new convenient `getInitialState` function on store objects.
- three [0.160.0 and 0.161.0](https://github.com/mrdoob/three.js/releases), now ESM only
- JSDOM [24.0.0](https://github.com/jsdom/jsdom/releases/tag/24.0.0) reverting the switch to a new selector engine which had worse performance
- Vite [5.0.10, 5.0.11 and 5.1.0](https://github.com/vitejs/vite/blob/v5.1.0/packages/vite/CHANGELOG.md)
- vite-plugin-checker [0.6.3 and 0.6.4](https://github.com/fi3ework/vite-plugin-checker/releases)
- Vitest [from 1.1.3 to 1.2.1](https://github.com/vitest-dev/vitest/releases)

I noticed `drei` removed `lodash` from its codebase, and since this seems to be a trend, I'm doing the same. Turns out we've got alternatives for most of it, either in vanilla JS of course (e.g. `typeof val === 'number'` isn't that long to write after all...), or in other libs: `MathUtils.clamp()` in `three`, `range()` in `d3-array`, and `useThrottledCallback` in `@react-hookz/web`. When vanilla JS was a bit too verbose, I abstracted into my own utility functions (`castArray` and `areSameDims`).

This shaves off another 34 kB from the demo's bundle (in addition to the 50 kB saved in https://github.com/silx-kit/h5web/pull/1564). We should see an equivalent impact in consumer apps.


packages/app/src/metadata-viewer/MetadataViewer.module.css
```diff
@@ -11,6 +11,7 @@
 
 .table th {
   font-weight: 500;
+  text-transform: capitalize;
```
axelboc, line 14: Replaces lodash's `capitalize()`.

packages/app/src/vis-packs/core/line/config.tsx
```diff
@@ -55,7 +54,13 @@ function createLineConfigStore() {
...
       {
         name: 'h5web:line',
-        partialize: (state) => omit(state, ['autoScale']),
+        partialize: (state) => ({
+          curveType: state.curveType,
+          showGrid: state.showGrid,
+          xScaleType: state.xScaleType,
+          yScaleType: state.yScaleType,
+          showErrors: state.showErrors,
+        }),
```
axelboc, line 63: Might as well be explicit about what we do keep.

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -273,3 +273,10 @@ export function assertComplexSignal(
 ): asserts nxData is NxData<ComplexType> {
   assertComplexType(nxData.signalDef.dataset);
 }
+
+export function areSameDims(dims1: ArrayShape, dims2: ArrayShape): boolean {
+  return (
+    dims1.length === dims2.length &&
+    dims1.every((dim, index) => dim === dims2[index])
+  );
+}
```
axelboc, line 282: This can be generalised if we ever need to compare other arrays, but it's much simpler than lodash's `isEqual` for sure, since the latter supports _everything_ (objects, deeply nested arrays, etc.)

packages/shared/src/guards.ts
```diff
@@ -126,10 +125,24 @@ export function assertArray(val: unknown): asserts val is unknown[] {
   }
 }
 
+function isTypedArray(val: unknown): boolean {
```
axelboc, line 128: Much simpler implementation than lodash's as well, since I think lodash parses the output of `toString`... Also, it will be simpler to check for bigint typed arrays when the type comes.

packages/shared/src/guards.ts
```diff
@@ -458,12 +471,6 @@ export function isNdArray<T extends Data>(
   return 'data' in arr;
 }
 
-export function isTypedArray<T extends TypedArray>(
-  arr: T | unknown[],
-): arr is T {
-  return !Array.isArray(arr);
-}
-
```
axelboc, line 466: This function was confusingly named. Also, it's so simple that it really doesn't warrant its own type guard.

packages/shared/src/hdf5-models.ts
```diff
@@ -207,13 +207,14 @@ export type ArrayValue<T extends DType> =
...
-      : never
-  : never;
+export type Value<D extends Dataset> =
+  D extends Dataset<infer S, infer T>
+    ? S extends ScalarShape
+      ? Primitive<T>
+      : S extends ArrayShape
+        ? ArrayValue<T>
+        : never
+    : never;
```
axelboc, line 217: Prettier formatting

loichuder (approved):
Nice :+1:

PR #1574, opened 2024-02-21
Add control to heatmap toolbar to flip `X` axis

### LINKED ISSUE #1573: [Heatmap] Add option to flip x-axis in the toolbar

I would like to be able to flip the x axis in the heatmap vizualization, just like I can currently flip the y axis.

Remote sensing data sometimes is produced that way. Swapping the axes and then rotating the axis formerly known as x gives me a rotated view, which is better but not ideal.

Thank you for an awesome tool.

axelboc:
Fix #1573 

- :sparkles: `< Heatmap >` Add toolbar control to flip `X` axis
- :sparkles: `[HeatmapVis]` New prop `flipXAxis` to flip heatmap horizontally
- :warning: Deprecate components `FlipYAxis` and `GridToggler` (to be removed in H5Web 12) — use `ToggleBtn` directly instead.


apps/storybook/src/HeatmapVisDisplay.stories.tsx
```diff
@@ -1,15 +1,15 @@
...
+import ndarray from 'ndarray';
 
 import HeatmapVisStoriesMeta from './HeatmapVis.stories';
 
-const dataArray = mockValues.twoD();
+const { dataArray } = HeatmapVisStoriesMeta.args;
+const dataArrayNoSym = ndarray(
+  dataArray.data.map((val, index) => (index % 41 >= 20 ? val * 5 : val)),
+  dataArray.shape,
+);
```
axelboc, line 12: Modifying `twoD` array to break the horizontal symmetry, otherwise the `FlipXAxis` story doesn't make much sense.

apps/storybook/src/Toolbar.stories.tsx
```diff
@@ -76,7 +76,13 @@ export const Default = {
           />
 
           <Separator />
-          <GridToggler value={showGrid} onToggle={toggleGrid} />
```
axelboc, line 79: A bunch of other toggles don't have their own components, so it didn't seem to make sense to still have `GridToggler` and `FlipYAxisToggler`. I decided to deprecate them rather than to create more such components (like `FlipXAxisToggler`) or to turn `FlipYAxisToggler` into a more generic `FlipAxisToggler`.

cypress/e2e/app.cy.ts
```diff
@@ -205,7 +205,28 @@ describe('/mock', () => {
     });
 
     if (Cypress.env('TAKE_SNAPSHOTS')) {
-      cy.matchImageSnapshot('edit_domain');
+      cy.matchImageSnapshot('heatmap_domain');
+    }
+  });
+
+  it.only('flip heatmap', () => {
```
axelboc, line 212: Adding a screenshot test.

packages/app/src/vis-packs/core/heatmap/HeatmapToolbar.tsx
```diff
@@ -73,7 +78,20 @@ function HeatmapToolbar(props: Props) {
 
       <Separator />
 
-      <FlipYAxisToggler value={flipYAxis} onToggle={toggleYAxisFlip} />
+      <ToggleBtn
+        label="X"
```
axelboc, line 82: Shortening the visible labels to just "X" and "Y", as the icon is quite explicit.

apps/storybook/src/HeatmapVisDisplay.stories.tsx
axelboc: I can definitely add it to the mock values you're right. I'm not sure about using it in the Cypress test, though, as this would require adding a new dataset to the mock demo or replacing the value of one of the exiting ones. I think it's fine to use the uint8 dataset instead, as it is already asymetrical.

loichuder: > I'm not sure about using it in the Cypress test, though, as this would require adding a new dataset to the mock demo or replacing the value of one of the exiting ones

Ah yes, adding a dataset changes all the screenshots since the explorer is open. Oh well, let's go with `uint8` then.

PR #1577, opened 2024-02-22
Add support for auxiliary signals in NX Image visualization

### LINKED ISSUE #1500: Support auxiliary signals for heatmaps
Example: https://h5web.panosc.eu/h5grove?file=wdnexamples/dectris.h5

The line plot supports auxiliary signals (plots them all together with the @signal) while the heatmap does not.

It is enough to select only one signal to plot, the @signal being the default while @auxiliary_signals are the alternatives. For example a dropdown would work.

axelboc:
Fix #1500 

I considered many implementation and UI options but eventually settled with the simplest I could think of:

- In terms of implementation:
  - I worked mostly inside `NxImageContainer` -- I didn't change anything in `DimensionMapper`, `MappedHeatmapVis`, `HeatmapToolbar`, `HeatmapVis`, etc.
  - The UI is rendered in the same container as the container (i.e. I didn't use a portal).
  - I added a single local state to the container, which is not persisted.
- In terms of UI/UX, I went with a very basic group of radio buttons just above the visualization. Can't do any simpler than that!

nx-image-aux


packages/app/src/dimension-mapper/DimensionMapper.module.css
```diff
@@ -1,5 +1,5 @@
 .mapper {
-  flex: none;
+  grid-area: dimMapper;
```
axelboc, line 2: Changing from a flexbox to a grid layout for `.visArea` so I can position the new UI element above the visualization. It saves me from adding wrappers in `NxImageContainer`.

packages/app/src/vis-packs/core/visualizations.test.ts
```diff
@@ -72,135 +72,135 @@ describe('Raw', () => {
   const { supportsDataset } = CORE_VIS.Raw;
 
   it('should support any dataset', () => {
-    expect(supportsDataset(scalarInt, mockStore)).toBe(true);
-    expect(supportsDataset(twoDStr, mockStore)).toBe(true);
+    expect(supportsDataset(scalarInt)).toBe(true);
+    expect(supportsDataset(twoDStr)).toBe(true);
```
axelboc, line 76: Quick win from using `satisfies` when defining the core and NeXus visualizations.

packages/app/src/vis-packs/nexus/NxSignalPicker.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+function NxSignalPicker<T extends NumericType | ComplexType>(props: Props<T>) {
+  const { definitions, onChange } = props;
+
+  return (
+    <p className={styles.bar}>
+      <span className={styles.text}>Signal:</span>
+      {definitions.map((def, index) => (
+        <label key={def.dataset.name} className={styles.label}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
axelboc, line 19: Pardon my French, but this rule is so dumb... I might disable it one day...

packages/app/src/vis-packs/nexus/NxSignalPicker.tsx
```diff
@@ -0,0 +1,35 @@
...
+          <input
+            type="radio"
+            name="signal"
+            value={def.dataset.name}
+            defaultChecked={index === 0}
```
axelboc, line 24: I left the radio buttons uncontrolled.

packages/app/src/vis-packs/nexus/containers/NxComplexImageContainer.tsx
```diff
@@ -8,16 +8,17 @@ import { useHeatmapConfig } from '../../core/heatmap/config';
...
-import { assertComplexSignal, guessKeepRatio } from '../utils';
+import { guessKeepRatio } from '../utils';
 
 function NxComplexImageContainer(props: VisContainerProps) {
   const { entity, toolbarContainer } = props;
   assertGroup(entity);
 
   const nxData = useNxData(entity);
-  assertComplexSignal(nxData);
+  assertComplexNxData(nxData);
```
axelboc, line 21: Removing a couple of duplicate type guards.

packages/app/src/vis-packs/nexus/NxSignalPicker.tsx
loichuder: Yeah, it also got in my way a few times

packages/lib/src/vis/heatmap/HeatmapVis.module.css
```diff
@@ -1,4 +1,5 @@
 .root {
+  grid-area: vis;
```
loichuder, line 2: Could this get in the way of uses of `HeatmapVis`/`LineVis`... by themselves in consumer apps ?

packages/app/src/vis-packs/nexus/containers/NxImageContainer.tsx
```diff
@@ -33,8 +37,24 @@ function NxImageContainer(props: VisContainerProps) {
...
 
+  const nxDataToFetch =
+    selectedDef === signalDef
+      ? { ...nxData, auxDefs: [] } // don't fetch auxiliaries
+      : {
+          ...nxData,
+          signalDef: selectedDef, // fetch only selected auxiliary
+          auxDefs: [],
+          titleDataset: undefined,
+        };
```
loichuder, line 48: If I am not mistaken, this can be rewritten more simply:

```suggestion
  const nxDataToFetch =
       {
          ...nxData,
          signalDef: selectedDef,
          auxDefs: [],  // don't fetch auxiliaries
          ...(selectedDef !== signalDef ?  { titleDataset: undefined } : {})
        };
```

But why is it needed to conditionally set `titleDataset` ?

loichuder (approved):
Looks good !

I now wonder if we should not use this radio button UI for Phase/Amplitude selection for complex datasets instead of the dropdown ?

image

packages/app/src/vis-packs/nexus/containers/NxImageContainer.tsx
axelboc: Yep totally!

Regarding the title dataset, the idea is to show the auxiliary signal's label (and potentially `long_name`/`unit` attributes) as the title of the plot. If a `titleDataset` is always provided, then its value is used regardless of which signal is selected — I didn't think this was what we wanted.

axelboc: (If you agree with this behaviour, I'll add a comment to clarify.)

loichuder: Yeah all right.

axelboc:
> I now wonder if we should not use this radio button UI for Phase/Amplitude selection for complex datasets instead of the dropdown ?

Interesting thought. The current Amp/Phase drop-down is indeed an exception in the toolbar: it controls the _what_ instead of the _how_ like other controls. I'll take a note to revisit this at a later time.

loichuder:
> it controls the _what_ instead of the _how_ like other controls.

Exactly :+1:

PR #1587, opened 2024-03-07
Fetch opaque datasets with h5grove as binary

axelboc:
This change is in preparation for #1554.

Opaque datasets were so far fetched with h5grove as JSON, which led to back-end errors for datasets that cannot be encoded to strings (like `datetime64_scalar`).

I now fetch opaque datasets as binary and return `UInt8Array` values from `H5GroveProvider` for consistency with `H5WasmProvider`.

With h5grove@2.0.0, this change leads to 422 errors for all opaque datasets. This will be fixed with https://github.com/silx-kit/h5grove/pull/89 and a new release of h5grove. This PR therefore constitutes a breaking change.

While I'm at it, I also improve the display of `UInt8Array` values slightly in `RawVis` by simply calling `toString()` instead of `JSON.stringify()`:

image image

### DIFF
--- a/packages/lib/src/vis/raw/RawVis.tsx
+++ b/packages/lib/src/vis/raw/RawVis.tsx
@@ -8,7 +8,11 @@ interface Props {
 
 function RawVis(props: Props) {
   const { value } = props;
-  const valueAsStr = JSON.stringify(value, null, 2);
+
+  const valueAsStr =
+    value instanceof Uint8Array
+      ? `Uint8Array [ ${value.toString()} ]`
+      : JSON.stringify(value, null, 2);
 
   return (
     <div className={styles.root}>
--- a/packages/app/src/App.module.css
+++ b/packages/app/src/App.module.css
@@ -148,6 +148,7 @@
 
 .error {
   composes: error from global;
+  grid-area: vis;
 }
 
 .error > span {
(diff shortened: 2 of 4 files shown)


axelboc:
I've updated the h5grove demo back-end following the merge of https://github.com/silx-kit/h5grove/pull/89. As planned, we no longer get 422 errors when fetching opaque datasets.

packages/lib/src/vis/raw/RawVis.tsx
```diff
@@ -8,7 +8,11 @@ interface Props {
 
 function RawVis(props: Props) {
   const { value } = props;
-  const valueAsStr = JSON.stringify(value, null, 2);
+
+  const valueAsStr =
+    value instanceof Uint8Array
```
loichuder, line 13: This could probably be made more generic for other types of `TypedArray`.

axelboc: Yeah but I didn't think the raw vis would ever receive any other kinds of typed arrays... When it's opaque, it's `Uint8Array`; h5wasm returns `Uint8Array` in a few of other cases, but otherwise it's typically a plain JS array/object. Other cases are handled by other visualizations. I didn't think it was worth making this more generic before encountering a real use case.

loichuder: Sensible :+1:

PR #1589, opened 2024-03-14
Fix style ordering issues caused by composing from imported CSS modules

axelboc:
Vite has a [new plugin](https://github.com/privatenumber/vite-css-modules) that fixes composition of CSS Modules by treating them as JS modules. It is planned to be integrated into Vite core but it may take a while.

There's one case that is not "fixed" by the plugin, which is when composing from a module that is itself imported from a React component. It's all explained here: https://github.com/privatenumber/vite-css-modules/issues/7. To address this, I decided to refactor the code to never compose from a module that is also imported. I've explained and documented this in the `CONTRIBUTING` guide.

The plugin, in addition to the refactoring above, allow us to remove all of our `!important` workarounds and bring much needed explanations and reassurance that there is indeed a way to use CSS Modules safely.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,7 @@
     - [DefinitelyTyped packages](#definitelytyped-packages)
     - [Workspace dependencies](#workspace-dependencies)
   - [Icon set](#icon-set)
+  - [CSS Modules](#css-modules)
 - [Build](#build)
   - [Package builds](#package-builds)
 - [Code quality](#code-quality)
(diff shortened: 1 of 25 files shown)


CONTRIBUTING.md
```diff
@@ -114,6 +115,89 @@ this problem in the form of the
...
+modules that are composed from other modules (`utils.module.css` in the example
+above) should be used solely for composition and never be imported. In other
+words, you should never compose from a class located in the CSS Modules file of
+another React component:
+
+```css
+/*
+  >>> BAD <<<
+  If `OtherComponent.module.css` gets imported after `ThisComponent.module.css,
+  the `.baseBtn` rule will end up after `.thisBtn` and `color: blue` will win.
```
loichuder, line 165:
```suggestion
  the `.baseBtn` rule will end up after `.thisBtn` and `color: blue` will be applied.
```

loichuder (approved):

PR #1590, opened 2024-03-14
Support JPEG/PNG images stored as opaque datasets in Raw vis

### LINKED ISSUE #1554: Visualize pictures stored as binary (PNG, JPEG, etc.)

axelboc:
Fix #1554 

I've added a simple PNG red dot to the mock file and a JPEG image to `grove.h5` (both with CC0 licence from Wikimedia Commons).

image


packages/lib/src/vis/raw/RawVis.tsx
```diff
@@ -1,13 +1,23 @@
...
-  const { value } = props;
+  const { value, title } = props;
+
+  if (value instanceof Uint8Array && isImage(value)) {
+    return (
+      <div className={styles.root}>
+        <img src={URL.createObjectURL(new Blob([value]))} alt={title} />
+      </div>
+    );
+  }
```
loichuder, line 20: From an architecture point of view, it would perhaps make more sense to add a specific `Vis` with `supportsDataset` equal to `value instanceof Uint8Array && isImage(value)` ?

loichuder (approved):
Spooky fish :skull: :fish:

packages/lib/src/vis/raw/RawVis.tsx
axelboc: Sorry for not reporting on this in the issue, but after investigating, I realised that values are fetched by the visualization containers, so `supportsDataset` can't have access to it. And if I add an `Opaque` visualization for opaque datasets, then it would also have to handle datasets that don't contain JPEG/PNG, which would duplicate the `Raw` visualization. I could make an `OpaqueContainer` that renders a new `RawImageVis` if an image is detected, or `RawVis` if not, but it felt unnecessary.

loichuder: Ah yes, didn't realize the condition was on the value. Fine then.

PR #1593, opened 2024-03-15
Fix `grid-area` regression

axelboc:
Specifying `grid-area: vis` in the visualizations did introduce a regression in the end. Turns out that if a [grid area identifier](https://devdocs.io/css/grid-area#custom-ident) is not defined in the parent grid, the cell ends up on a new row/column for reasons that I [don't quite clearly understand](https://stackoverflow.com/a/57750033/758806)... Since we wrap all our stories with `display: grid` in Storybook, the high-level visualization stories were all messed up.

So I went ahead with the plan I had to pass the `grid-area: vis` style from `@h5web/app` down to the visualizations in `@h5web/lib` via a custom `className` prop. I've also added a `style` prop to all the visualizations.

### DIFF
--- a/packages/lib/src/vis/heatmap/HeatmapVis.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapVis.tsx
@@ -9,7 +9,12 @@ import type { DefaultInteractionsConfig } from '../../interactions/DefaultIntera
 import DefaultInteractions from '../../interactions/DefaultInteractions';
 import ResetZoomButton from '../../toolbar/floating/ResetZoomButton';
 import { useAxisDomain, useValueToIndexScale } from '../hooks';
-import type { Aspect, AxisParams, VisScaleType } from '../models';
+import type {
+  Aspect,
+  AxisParams,
+  ClassStyleAttrs,
+  VisScaleType,
+} from '../models';
 import TooltipMesh from '../shared/TooltipMesh';
 import VisCanvas from '../shared/VisCanvas';
 import { DEFAULT_DOMAIN } from '../utils';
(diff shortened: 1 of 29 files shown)


packages/lib/src/vis/surface/SurfaceVis.tsx
```diff
@@ -22,7 +23,7 @@ interface Props {
   showPoints?: boolean;
 }
 
-function SurfaceVis(props: PropsWithChildren<Props>) {
+function SurfaceVis(props: PropsWithChildren<Props> & ClassStyleAttrs) {
```
axelboc, line 26: This visualization's `Props` interface is used in `SurfaceMesh`, so I could not extend it. I went with this syntax instead.

packages/lib/src/vis/heatmap/HeatmapVis.tsx
```diff
@@ -81,7 +88,12 @@ function HeatmapVis(props: Props) {
   const maskArray = useMask(dataArray, ignoreValue);
 
   return (
-    <figure className={styles.root} aria-label={title} data-keep-canvas-colors>
+    <figure
+      className={`${styles.root} ${className}`}
```
axelboc, line 92: Just concatenating the two classes by hand; I don't think it's worth an extra library.

packages/lib/src/vis/surface/SurfaceVis.tsx
loichuder: Perhaps `SurfaceVisProps` should be in a `models` file then to be extended/imported in both components :shrug:

loichuder (approved):

packages/lib/src/vis/surface/SurfaceVis.tsx
axelboc: In the end, I created a dedicated `Props` interface in `SurfaceMesh`.

PR #1594, opened 2024-03-18
Upgrade Storybook

axelboc:
Upgrade to Storybook 8:

- [Release announcement](https://storybook.js.org/blog/storybook-8/)
- [Migration guide](https://storybook.js.org/docs/migration-guide?ref=storybookblog.ghost.io)
- [Full changelog](https://github.com/storybookjs/storybook/blob/next/CHANGELOG.md#800)

<details><summary><em>Migration of <code>argTypesRegex</code> postponed due to <code>@storybook/test</code> breaking <code>expect</code> types in our Vitest test suites</em></summary>

The main change is that `on<Event>` args now have to be **[mocked explicitly](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#implicit-actions-can-not-be-used-during-rendering-for-example-in-the-play-function)**.

Previously, we had `actions: { argTypesRegex: '^on[A-Z].*' }` in `preview.ts`, which meant that Storybook would automatically pass a mock function for every `on<Event>` argument. We would then spread all the `args` on the component in the JSX, and every time one of the event handlers was called, an "action" would be logged in the Actions panel:

image

Now the mocking has to be explicit via the `fn` function of the `@storybook/test` package (which is basically Vitest's [`fn` function](https://vitest.dev/api/vi.html#vi-fn)).

While working on this, I realised that a lot of the time, we pass our own `on<Event>` handlers in the stories to manage local state. In this situation, the implicit mock event handler is overridden and therefore never called. I've fixed this by calling the mock event handlers explicitly. This makes some of the stories a bit more verbose, but it's worth it I think.

> Note that I don't call some mock event handlers like `onSelectionChange`, as this can generate some lag.

</details>


apps/storybook/src/Stacking.mdx
```diff
@@ -51,36 +51,38 @@ The stacking order is organised into six layers, from furthest (behind) to close
 The example below demonstrates the default stacking order. You can hover to test the stacking of the tooltip and crosshairs,
 and zoom in to test the stacking of the reset-zoom button in the floating toolbar.
 
-<Canvas>
-  <div style={{ display: 'flex', height: '25em' }}>
-    <VisCanvas
-      abscissaConfig={{ visDomain: [0, 41], showGrid: true }}
-      ordinateConfig={{ visDomain: [0, 20], showGrid: true }}
+{/* https://github.com/storybookjs/storybook/issues/19964 */}
+{/* <Canvas> */}
```
axelboc, line 55: Commenting this out for now since it causes an error. the JSX below still works; we just lose the ability to inspect the source code. I'll keep an eye on the closed issue linked above and open a new one if need be.

apps/storybook/src/DomainWidget.stories.tsx
```diff
@@ -12,6 +12,7 @@ const meta = {
   decorators: [DomainWidgetDecorator],
   args: {
     customDomain: undefined,
+    onCustomDomainChange: () => {},
```
axelboc, line 15: This is the only required event handler prop in the entire lib, so I get an error if it's not specified in `args`.

I'm passing a no-op function for now, but at some point it will receive an explicit mock function: `onCustomDomainChange: fn()` (cf. updated PR description).

apps/storybook/src/Stacking.mdx
loichuder: Weird.

loichuder (approved):
Thanks :)

PR #1599, opened 2024-03-25
Install `vite-css-modules` plugin in lib/app packages

### LINKED ISSUE #1598: Styling regression in dimension mapper
### Describe the bug

Looks like #1589 caused a styling regression:

image

It only happens when using `@h5web/app` as a library - i.e. when importing the viewer's styles from `@h5web/app/styles.css` (which points to `packages/app/dist/styles.css`), so there's a problem with how that file is generated. The problem doesn't occur in the demo, where styles are imported from their source - i.e. from `packages/app/src/styles.css`.

### To Reproduce

https://codesandbox.io/p/devbox/h5webapp-vite-5c204

axelboc:
Fix #1598 

So... fun story, I forgot to add the `vite-css-modules` plugin to the Vite configs of `@h5web/app` and `@h5web/lib`. :sweat_smile:

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -90,6 +90,7 @@
     "rollup-plugin-dts": "6.1.0",
     "typescript": "5.0.4",
     "vite": "5.1.6",
+    "vite-css-modules": "1.4.2",
     "vitest": "1.3.1"
   }
 }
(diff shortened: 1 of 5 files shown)


axelboc:
The regression didn't catch my eyes when I tested the beta `@h5web/app` package in the sandbox.

Issue #1582, opened 2024-02-27
Webworker h5wasm provider (for random access to local files)


bmaranville:
### Is your feature request related to a problem?

Currently for the h5wasm provider, the entire file must be loaded into memory before use (it is written to the MEMFS virtual file system provided by Emscripten)

This puts an upper limit of 2GB (?) on the size of files that can be used with the h5wasm provider, and can cause memory issues for users (entire file in memory).

The only advantage of this system is that file access (once loaded) is very very fast.

### Requested solution or feature

Use the WORKERFS Emscripten file system and a webworker-based h5wasm provider, which allows random-access to files on the users' computer without loading the entire thing into memory.

### Alternatives you've considered

The new File System Access API could also solve this problem, where users could mount a local folder for working with and have random access to the files in that folder.  This API is only fully implemented on Chrome-based browsers, however.

### Additional context

Here is an example worker:
```javascript
// h5wasm_worker.js
importScripts("https://unpkg.com/comlink/dist/umd/comlink.js");
importScripts("https://cdn.jsdelivr.net/npm/h5wasm@0.7.1/dist/iife/h5wasm.js")

const WORKING_DIRECTORY = '/working_directory';

async function load_file(file) {
  // file is of type File
  const { FS } = await h5wasm.ready;
  const { filesystems: { WORKERFS } } = FS;
  const { name: filename, size } = file;
  if (!FS.analyzePath(WORKING_DIRECTORY).exists) {
    FS.mkdir(WORKING_DIRECTORY);
  }
  if (api.file !== null) {
    api.file.close();
    // only use a single file at a time;
    // unmount the previous filesystem
    FS.unmount(WORKING_DIRECTORY);
  }

  FS.mount(WORKERFS, { files: [ file ] }, WORKING_DIRECTORY);
  const h5wasmFile = new h5wasm.File(`${WORKING_DIRECTORY}/${filename}`);
  api.file = h5wasmFile;
  return;
}

async function get_entity(path = '/') {
  if (api.file === null) {
    return null;
  }
  return api.file.get(path);
}

async function get_type(path) {
  const entity = await get_entity(path);
  return entity.type;
}

async function get_attrs(path) {
  const entity = await get_entity(path);
  const attrs = entity.attrs;
  return attrs;
}

// Group functions:
async function get_keys(group_path) {
  const entity = await get_entity(group_path);
  if (entity === null) {
    return null;
  }
  // assert entity instanceof h5wasm.Group;
[...]

PR #1602, opened 2024-04-04
Upgrade h5wasm to support empty datasets

axelboc:
Support for empty datasets and attributes was [merged](https://github.com/usnistgov/h5wasm/pull/67) and released in v0.7.3. I'm upgrading and also making a few unrelated changes ahead of opening a PR for #1582


packages/app/src/index.ts
```diff
@@ -17,10 +17,10 @@ export type { DataContextValue } from './providers/DataProvider';
 
 // Undocumented
 export { assertEnvVar } from '@h5web/shared/guards';
+export { assertNonNull } from '@h5web/shared/guards';
 
 // Undocumented (for @h5web/h5wasm)
 export { default as DataProvider } from './providers/DataProvider';
 export { DataProviderApi } from './providers/api';
 export type { ValuesStoreParams } from './providers/models';
-export { getNameFromPath, getValueOrError } from './providers/utils';
```
axelboc, line 25: I move `getNameFromPath` to the shared package so I can later import it in a web worker. For some reason, importing it from here led to pending promises as explained in https://github.com/silx-kit/h5web/issues/1582#issuecomment-2034729074

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -36,12 +36,12 @@ export class H5GroveApi extends DataProviderApi {
     super(filepath, { baseURL: url, ...axiosConfig });
   }
 
-  public async getEntity(path: string): Promise<ProvidedEntity> {
+  public override async getEntity(path: string): Promise<ProvidedEntity> {
```
axelboc, line 39: Abstract methods don't require this `override` keyword (since technically they are being _implemented_ not _overridden_). However, I do find it makes the class more readable as explained in this [TS issue](https://github.com/microsoft/TypeScript/issues/47250).

packages/app/src/providers/mock/mock-api.ts
```diff
@@ -50,16 +50,7 @@ export class MockApi extends DataProviderApi {
     return entity;
   }
 
-  public async getAttrValues(entity: Entity): Promise<AttributeValues> {
```
axelboc, line 53: Just moving this down below `getValue` for consistency between all provider API classes.

packages/h5wasm/src/__snapshots__/h5wasm-api.test.ts.snap
```diff
@@ -460,13 +460,13 @@ exports[`test file matches snapshot 1`] = `
...
     },
-    "shape": [],
+    "shape": null,
     "type": {
       "class": "Float",
       "endianness": "little-endian",
       "size": 32,
     },
-    "value": undefined,
+    "value": null,
```
axelboc, line 469: The main change: empty datasets/attributes now have `null` shapes and values, just like in h5grove. They no longer appear as "Scalar" in the metadata viewer.

packages/h5wasm/src/utils.ts
```diff
@@ -241,6 +245,7 @@ export function parseDType(metadata: Metadata): DType {
   if (h5tClass === H5TClass.Array) {
     const { array_type } = metadata;
     assertDefined(array_type);
+    assertNonNull(array_type.shape);
```
axelboc, line 248: Expectedly, we now need to assert that the shape is non-null in a couple of places.

loichuder (approved):
All good :slightly_smiling_face:

PR #1606, opened 2024-04-10
Add _Nx Note_ visualization

### LINKED ISSUE #1603: Add rich display for NXnotes

When saving processed data in workflows, we usually save the parameters in a [NXnote group](https://manual.nexusformat.org/classes/base_classes/NXnote.html#index-0). The NXnote group contain a `data` dataset that contain the parameters as JSON and a `type` dataset that contain the MIME type (here `application/json`).

When I want to inspect the parameters in H5Web, I look at the `data` dataset but since it uses the scalar visualization, I get the JSON serialized as a single line which is not very readable.

I would like H5Web to be able to take advantage of the MIME type to be able to present me with a rich display of the `data` dataset. The same way NxHeatmap is used when clicking on the NXdata group, we could have a NxNote visualisation when clicking on the NXnote.

None.

### Additional context

See `/results/integrate/configuration` at https://h5web.panosc.eu/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Fnxnote.h5 for an example

axelboc:
Fix #1603 

image

Nothing crazy, but there is a subtlety: the MIME type is stored in a dataset and not an attribute (:shrug:), which means that with our current architecture, we don't have access to it when looking for supported visualizations. We can only detect that the group has `NX_class=NXnote`, render `NxNoteContainer`, and only then fetch the values of the `data` and `type` datasets. So if the MIME type is not JSON, we get:

image

When ideally, it would be nicer to show the default fallback (without the _NX Note_ tab):

image

I could refactor the visualization detection code so it allows fetching dataset values. The downside is that it will trigger the suspense fallback that wraps the whole `Visualizer`. I guess it's okay as long as we fetch only datasets that contain metadata and not actual data... Any thoughts @loichuder?


packages/app/src/ErrorFallback.tsx
```diff
@@ -3,21 +3,25 @@ import type { FallbackProps } from 'react-error-boundary';
 import styles from './App.module.css';
 import { CANCELLED_ERROR_MSG } from './providers/utils';
 
-function ErrorFallback(props: FallbackProps) {
-  const { error, resetErrorBoundary } = props;
+interface Props extends FallbackProps {
+  className?: string;
```
axelboc, line 7: So I can pass `grid-area: vis` from `VisBoundary`.

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -267,3 +274,11 @@ export function areSameDims(dims1: ArrayShape, dims2: ArrayShape): boolean {
     dims1.every((dim, index) => dim === dims2[index])
   );
 }
+
+export function parseJson(value: string): unknown {
```
axelboc, line 278: Small function to rethrow a nicer error.

loichuder (approved):
Thank you so much :heart_eyes:

PR #1608, opened 2024-04-12
Allow flipping RGB visualizations on X and Y

axelboc:
We received a request by email to allow flipping RGB visualizations on X and Y.

So far we were always applying a flip on Y. It is still the default, but users can now toggle it off (in addition to flipping on X if they like).

image


apps/storybook/src/RgbVis.stories.tsx
axelboc, line 1: Documenting `RgbVis` while I'm at it.

loichuder (approved):
I would have kept `flipY` on by default but ok !

axelboc:
> I would have kept `flipY` on by default but ok !

Yeah, why not.

PR #1615, opened 2024-04-18
Fix trying to close file twice in strict mode

axelboc:
Following the stale state issue in myHDF5 that was fixed in #1568, I started seeing HDF5 warnings in development: the h5wasm provider was trying to close the same file twice.

That's because when React in strict mode re-renders a second time to detect side effects, it doesn't update the state (i.e. `setPrevApi(api)` has no effect) so `prevApi !== api` evaluates to `true` again and `prevApi.cleanUp()` is called again.

Now that HDF5 errors are thrown instead of logged as warnings, they really stand out in the console. So I thought I'd try to get rid of them.

None of the React-y solutions (which involve either a `key` or a `useEffect` with a clean-up function) are satisfactory, as explained in #1568. In the end, all I want is to call the clean-up function when the `useMemo` callback is re-invoked, which can be done by just storing the previous API instance in a ref instead of a state.

### DIFF
--- a/packages/h5wasm/src/local/worker.ts
+++ b/packages/h5wasm/src/local/worker.ts
@@ -38,7 +38,6 @@ async function openFile(file: File): Promise<bigint> {
 
 async function closeFile(fileId: bigint): Promise<number> {
   const h5wasm = await h5wasmReady;
-  h5wasm.flush(fileId);
   return h5wasm.ccall('H5Fclose', 'number', ['bigint'], [fileId]);
 }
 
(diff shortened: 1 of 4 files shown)


packages/h5wasm/src/local/worker.ts
```diff
@@ -38,7 +38,6 @@ async function openFile(file: File): Promise<bigint> {
 
 async function closeFile(fileId: bigint): Promise<number> {
   const h5wasm = await h5wasmReady;
-  h5wasm.flush(fileId);
```
axelboc, line 41: While I'm at it, I remove those `flush` calls from both providers, as they are totally unnecessary.

PR #1632, opened 2024-04-29
Tweak camera frustrum to provide wider `z` range

### LINKED ISSUE #1626: z-range is only -3..3

I use a lot the z-index.

But i saw that the range is clamped to something quite small like 0..3 or -3...3

Outside, a component is hidden.

I have to use a lot of layers for my use cases.

Sure i still can divide 0..3 by an infinite amount of values. But i feel like more easy to deal with integer.

Use a bigger range for the z-buffer, maybe something like 0..1000 or -1000..1000, i don't know.

Rescale the z-depth on my side.

### Additional context

I use @h5web/lib": "7.0.2-beta.0

axelboc:
Fix #1626

I went with `{ near: 0, far: 1000, position: [0, 0, 1000] }` to give a visible `z` range of `[0 1000]` as explained in https://github.com/silx-kit/h5web/issues/1626#issuecomment-2082921944.

To make the relation between the camera frustrum/position settings and the visible `z` range, I've defined three constants and exported/documented them in `@h5web/lib`.

### DIFF
--- a/apps/storybook/src/Utilities.mdx
+++ b/apps/storybook/src/Utilities.mdx
@@ -405,6 +405,16 @@ writing your own geometries:
 - `ScatterPointsGeometry`, used in [`ScatterPoints`](https://h5web-docs.panosc.eu/?path=/docs/visualizations-scattervis--docs) (undocumented)
 - `SurfaceMeshGeometry`, used in [`SurfaceMesh`](https://h5web-docs.panosc.eu/?path=/docs/experimental-surfacevis--docs) (experimental)
 
+### Constants
+
+#### `Z_MIN = 0`, `Z_MAX = 1000`
+
+R3F/Three objects inside `VisCanvas` with a `z` position in the range `[Z_MIN Z_MAX]`
+are in view of the scene's orthographic camera. Objects positioned further than `Z_MIN`
+(`z < 0`) or closer than `Z_MAX` (`z > 1000`, i.e. behind the camera) are not in view of
+the camera, and therefore not rendered at all. By default, objects are positioned in view
+of the camera, at `z = 0`.
+
 ### Mock values
 
 The `mockValues` object can be used to generate mock ndarrays for testing purposes:
--- a/packages/lib/src/vis/utils.ts
+++ b/packages/lib/src/vis/utils.ts
@@ -39,13 +39,20 @@ import type {
 } from './models';
 import { scaleGamma } from './scaleGamma';
 
-export const DEFAULT_DOMAIN: Domain = [0.1, 1];
+// Orthographic camera configuration (cf. `R3FCanvas.tsx`)
+export const CAMERA_NEAR = 0;
+export const CAMERA_FAR = 1000;
+export const CAMERA_Z = CAMERA_FAR;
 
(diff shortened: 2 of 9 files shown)


apps/storybook/src/Utilities.mdx
```diff
@@ -405,6 +405,15 @@ writing your own geometries:
...
 
+### Constants
+
+#### `CAMERA_Z`
+
+The `z` position of the canvas' orthographic camera: **1000**.
+
+The camera's frustrum is configured to `{ near: 0.1, far: CAMERA_Z *2 }`, so the visible `z` range is `[-CAMERA_Z CAMERA_Z[`.
+This means that any child of `VisCanvas` positioned outside of the range `-1000 <= z < 1000` is not rendered.
+
```
axelboc, line 416: Still planning on improving/splitting up this file but thought I'd at least document the constant for now.

apps/storybook/src/Utilities.mdx
```diff
@@ -405,6 +405,16 @@ writing your own geometries:
 - `ScatterPointsGeometry`, used in [`ScatterPoints`](https://h5web-docs.panosc.eu/?path=/docs/visualizations-scattervis--docs) (undocumented)
 - `SurfaceMeshGeometry`, used in [`SurfaceMesh`](https://h5web-docs.panosc.eu/?path=/docs/experimental-surfacevis--docs) (experimental)
 
+### Constants
```
axelboc, line 408: Still planning on improving/splitting up this file but thought I'd at least document the constants here for now.

loichuder: Do you think it is worth documenting the constants ? We don't expect consumers from the `lib` to use them, no ?

packages/lib/src/vis/utils.ts
```diff
@@ -41,11 +41,18 @@ import { scaleGamma } from './scaleGamma';
...
+export const CAMERA_NEAR = 0;
+export const CAMERA_FAR = 1000;
+
+/* The `z` position of the orthographic camera.
+ * Visible `z` range for objects in the scene: [0 1000]
+ * i.e. (CAMERA_Z - CAMERA_FAR) <= z <= (CAMERA_Z - CAMERA_NEAR) */
+export const CAMERA_Z = CAMERA_FAR;
+
+/* `z` coordinate located outside of the camera's frustrum (too far).
+ * Useful to render points with NaN/infinite coordinates without Three warnings. */
```
loichuder, line 54:
```suggestion
 * Used to hide points with NaN/infinite coordinates without Three warnings. */
```

loichuder (approved):

apps/storybook/src/Utilities.mdx
axelboc: Fair enough, I'll remove them.

loichuder: This being said, giving the consumers the limits of the camera FOV would perhaps be helpful ?

axelboc: I've added two constants: `Z_MIN` and `Z_MAX`, and documented them. It's clearer indeed, including in the code/comments.

loichuder (approved):

PR #1633, opened 2024-05-02
Add toggle to fit raw image (on by default)

### LINKED ISSUE #1630: Allow fitting Raw Image to window size

Right now when storing image files with high resolutions as opaque datasets and visualizing them they do not fit into the display area and get scrollbars associated with them. While this makes sense when wanting to view details in the images it is a bit problematic if the image in question is just an overview picture with high resolution that is best viewed as a whole.

Add a toggle to fit/not-fit an image within the display area.

Scaling down the image myself or adding a "thumbnail" low-res version but that just duplicates data.

axelboc:
Fix #1630

I add a toggle button to the _Raw_ visualization toolbar and a (persisted) config store to hold the state. The button is toggled on by default. When the dataset isn't a binary image, the button is disabled.

I chose to create a separate visualization component called `RawImageVis` to deal specifically with JPEG/PNG images; it's the `RawVisContainer` that now decides which visualization component to use.

fit-raw-image

### DIFF
--- a/packages/app/src/vis-packs/core/configs.ts
+++ b/packages/app/src/vis-packs/core/configs.ts
@@ -1,3 +1,4 @@
+export { RawConfigProvider } from './raw/config';
 export { MatrixConfigProvider } from './matrix/config';
 export { LineConfigProvider } from './line/config';
 export { HeatmapConfigProvider } from './heatmap/config';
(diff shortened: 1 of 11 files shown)


loichuder (approved):
I like it ! Both in the code and in the UI !

Issue #1578, opened 2024-02-22
Don't use debounce on slice slider (update as fast as possible)


bmaranville:
### Is your feature request related to a problem?

When using the slicing slider (in DimensionMapper) for heatmaps, the plot is not updated smoothly when the slider is moved, even when using h5wasm provider (which should allow very fast updates).

The debounce that is applied is preventing updates to the plot.

### Requested solution or feature

Instead of debouncing the slider signal, I think a different approach is merited here and would provide huge gains in responsiveness:  

1. add a separate independent requestAnimationFrame loop that checks if the heatmap needs to be redrawn
   - if "needs redraw" flag is true and not "drawing" flag, set "drawing" = true: 
     - set "needs redraw" = false
     - start redraw with current slider value (redraw operation should do the data slicing, which could be slow if coming from a server)
     - when fetch + redraw is complete, set "drawing" = false and continue animation loop
   - if "needs redraw" is false, do nothing and continue animation loop.
2. when the slider is moved, set the "needs redraw" flag to true.

The advantage of this system over debounce is that it updates as fast as the (data slice + plot) operation can be carried out, rather than arbitrarily delaying execution of the slice by the debounce timers.  There is natural rate-limiting to slice requests from e.g. H5Grove because no more than one request will be issued at a time.  Providers with very fast slice turnarounds (e.g. h5wasm) will be able to update nearly instantly (smooth visual scrolling through datasets!)

### Additional context

I have implemented this type of buffering on many occasions for smooth updates of visualizations from user inputs.  I'm happy to help with this.

I couldn't find a nice requestAnimationFrame equivalent to the `react-hookz/web` debounce tools being used, so it might have to be coded by hand rather than using a library.

bmaranville:
Here is a demo with debouncing turned off:

https://github.com/silx-kit/h5web/assets/686570/4d791ea8-c83f-49c7-a1e5-a16caccbb209

loichuder:
Great call @bmaranville, I wholly agree.

The debounce indeed only makes sense when using providers that make requests for each slice such as H5Grove or HSDS. It only gets in the way of responsiveness when using the H5Wasm provider.

axelboc:
I agree that the debouncing is not ideal. Our main goal at the time was to avoid spamming h5grove with concurrent requests while keeping the slider responsive. I think this was before h5wasm came along.

[...]

PR #1634, opened 2024-05-03
Remove debounce on dimension mapping state to reduce lag

axelboc:
Baby step to address #1578

1. We first added debouncing on the dimension mapping state in #888 to address #879.
2. In #920, we then realised that this also debounced the axis mappers (the D0/D1/.../Dn button groups), so we opened #975 with the goal of moving the debounce to the sliders.
3. Turns out that in that PR, we added debouncing to the sliders without actually removing it on the dimension mapping state... So we ended up with two debounces, for a total of 350 ms.

In this PR, I therefore remove the debounce on the dimension mapping state, thus saving 100ms both when slicing and changing axis dimensions.


loichuder (approved):
Sorry for the delay. Well spotted :+1:

PR #1567, opened 2024-02-07
Refactor export menu with new headless UI library

axelboc:
`react-aria-menubutton` is no longer maintained. It is still on React 17 and CommonJS, which is causing issues in our Cypress tests and for consumer applications.

In this PR, I'm starting the migration to another headless UI library called **[Ariakit](https://ariakit.org/)**, which seems like a well-maintained, featureful, Typescript-first library. I've considered and discarded other options:

- [React Aria](https://react-spectrum.adobe.com/react-aria/index.html) from Adobe — I'm not a fan of the fact that it [renders class names](https://react-spectrum.adobe.com/react-aria/styling.html#class-names) by default;
- [Headless UI](https://headlessui.com/) from Tailwind — the number of components is quite limited, which might be a hindrance in the future;
- [Reach UI](https://reach.tech/) — no longer maintained

I'll continue the migration in separate PRs.

PR #1642, opened 2024-05-16
Refactor Export menu with Floating UI

axelboc:
Supersedes #1567

Since Ariakit and React Aria were unsatisfactory options with regard to bundle size, I decided to try a more specialised solution: [Floating UI](https://floating-ui.com/docs/getting-started) (a full rebuild of `popper.js` and `react-popper`). It's really well designed; all based on hooks (along with helper components that I didn't end up using).

The two main hooks I use are `useFloating`, which provides the refs and the popover's positioning logic, and `useListNavigation`, which provides the arrow up/down navigation (including opening the menu on arrow down, looping, focus management, etc.) This weighs roughly 30 kB, which is way more reasonable than the other two headless UI libraries I tried...

At first I was also using `useDismiss`, to close the menu on <kbd>Escape</kbd> and click outside, and `<FloatingFocusManager />` to close the menu when the focus leaves the menu. But both of those added an extra 16 kB. Since we already use `useClickOutside` and `useKeyboardListener` from react-hookz, and since `<FloatingFocusManager />` has a lot of features we don't need (e.g. focus trap for modals), I decided to reimplement the same behaviours myself. As I said, Floating UI is very well designed, and this was [super easy to do](https://floating-ui.com/docs/custom-hooks).

All in all, I'm very happy with the result at last, so I will continue the migration from `react-aria-menubutton` in separate PRs by refactoring other components like `Selector`.


axelboc:
Now with all the required ARIA props ... sorry for the spam. Again, to save bundle size, I didn't use [`useRole`](https://floating-ui.com/docs/useRole) but only replicated it's logic for `role="menu"`. I also just find it useful to make the ARIA attributes explicit in the source code, since they are relied upon for styling.

packages/lib/src/toolbar/controls/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  anchor.href =
+    urlOrBlob instanceof Blob ? URL.createObjectURL(urlOrBlob) : urlOrBlob.href;
+  anchor.click();
+
+  URL.revokeObjectURL(anchor.href);
+  anchor.remove();
+}
+
+export const floatingMinWidth = {
```
loichuder, line 21: Upper case ?

packages/lib/src/toolbar/controls/hooks.ts
```diff
@@ -0,0 +1,44 @@
...
+import { useClickOutside, useKeyboardEvent } from '@react-hookz/web';
+import type { FocusEvent } from 'react';
+import { useCallback, useMemo } from 'react';
+
+/* Custom dismiss interaction hook for Floating UI widgets.
+ * Reduces bundle size (~16 kB) by replacing:
+ * - `useDismiss` (Escape key, click outside)
+ * - `FloatingFocusManager` (focus out) */
+export function useFloatingDismiss(context: FloatingContext): ElementProps {
+  const { refs, onOpenChange } = context;
```
loichuder, line 11: This is the kind of situation where I find that the `on...` convention for handlers does not hold well.

Oh well.

loichuder (approved):

packages/lib/src/toolbar/controls/utils.ts
axelboc: Hmm I see your point, but camel case matches the other Floating UI middlewares.

packages/lib/src/toolbar/controls/hooks.ts
axelboc: Agreed, I could pass `toggle` directly, but it felt unnecessarily cumbersome.

PR #1644, opened 2024-05-17
Refactor Selector with Floating UI

axelboc:
This was a bit more complicated because of the option groups in the color map selector, since it forced me to use [FloatingList](https://floating-ui.com/docs/FloatingList) and `useListItem`.

But ... I can now remove `react-aria-menubutton` at last :tada: 

Next steps:

- Refactor `OverflowMenu` and `InteractionHelp` with Floating UI to gain a few features (auto-placement, better accessibility and focus management, etc.)
- Refactor option groups and a few other bits and pieces in `Selector`
- Deprecate `align` prop of `ExportMenu`
- Abstract duplicated logic into custom hooks


packages/lib/src/toolbar/controls/Selector/Option.tsx
```diff
@@ -0,0 +1,35 @@
...
+  activeIndex: number | null;
+  selectedIndex: number;
+  getItemProps: UseInteractionsReturn['getItemProps'];
+  onSelect: (index: number) => void;
+}
+
+function Option(props: PropsWithChildren<Props>) {
+  const { activeIndex, selectedIndex, children, getItemProps, onSelect } =
+    props;
+  const { ref, index } = useListItem();
```
axelboc, line 17: This hook works together with `FloatingList`. It retrieves the index of the option based on rendering order (regardless of whether they are part of an option group) and registers it in the `listRef` array (cf. `Selector`) so `useListNavigation` can navigate between options.

packages/lib/src/toolbar/controls/Selector/Selector.tsx
```diff
@@ -28,54 +37,136 @@ function Selector<T extends string>(props: Props<T>) {
...
+      offset(6),
+      size({
+        padding: 12,
+        apply({ availableHeight, elements, rects }) {
+          Object.assign(elements.floating.style, {
+            maxHeight: `${Math.min(availableHeight, MENU_MAX_HEIGHT)}px`,
+            minWidth: `${rects.reference.width}px`,
+          });
+        },
+      }),
```
axelboc, line 66: This is an improved version of the `floatingMinWidth` middleware that also sets the max height of the popup so it never overflows the viewport. I will use it in `ExportEntry` as well eventually.

packages/lib/src/toolbar/controls/Selector/Selector.tsx
```diff
@@ -28,54 +37,136 @@ function Selector<T extends string>(props: Props<T>) {
...
-                      value={value}
-                    />
-                  </ul>
-                </li>
-              ))}
-            </ul>
-          )}
-        </Menu>
-      </Wrapper>
+          <FloatingList elementsRef={listRef}>
```
axelboc, line 132: This is the `FloatingList` I was talking about. It's basically just a context under the hood.

packages/lib/src/toolbar/controls/Selector/Selector.tsx
```diff
@@ -28,54 +37,136 @@ function Selector<T extends string>(props: Props<T>) {
...
-    optionComponent: Option,
+    optionComponent: OptionComp,
   } = props;
 
-  const { height: winHeight } = useWindowSize();
-  const menuMaxHeight = winHeight - MENU_TOP - MENU_BOTTOM;
+  const [isOpen, toggle] = useToggle();
+  const [activeIndex, setActiveIndex] = useState<number | null>(null);
+  const [selectedIndex, setSelectedIndex] = useState(() => {
+    return getAllOptions(options).indexOf(value);
```
axelboc, line 46: I have to find for the index of the selected option to properly initialise the `selectedIndex` state. This requires flatting the record when option groups are involved like in the color map drop-down.

packages/lib/src/toolbar/controls/Selector/Selector.tsx
```diff
@@ -28,54 +37,136 @@ function Selector<T extends string>(props: Props<T>) {
...
-          </div>
-        </Button>
+      <button
+        ref={refs.setReference}
+        id={referenceId}
+        className={styles.btn}
+        type="button"
+        disabled={disabled}
+        role="combobox"
+        aria-labelledby={`${label ? labelId : ''} ${selectedOptionId}`}
```
axelboc, line 108: Note that I'm combining the label of the selector with the selected option. Floating UI's `useRole` doesn't do this, so I'm glad I didn't use it; I saw this pattern in React Aria.

loichuder (approved):

PR #1647, opened 2024-05-21
Refactor `OverflowMenu` and `InteractionHelp` with Floating UI

axelboc:
The selector menus, the export menu, and the overflow and interaction help popovers now all behave the same way in terms of focus management, dismissal, positioning, etc.

In this PR, I've also streamlined the CSS across all these components, simplified the DOM in a few places (unnecessary wrappers, etc.) and fixed a couple of flexbox alignment regressions that I've only just noticed.

I've left the `DomainWidget` alone for now because the toggling logic is more complex.


apps/storybook/src/Customization.mdx
```diff
@@ -135,13 +135,13 @@ as you see fit. For instance, if you'd like to change the color of the curve of
 | `--h5w-btnRaised--shadowColor`        | `gray`                                             | Box shadow color of raised buttons                        |
 | `--h5w-btnRaised-hover--shadowColor`  | `dimgray`                                          | Box shadow color of raised buttons on hover               |
 
-##### Selectors
+##### Selectors & menus
```
axelboc, line 138: `ExportMenu` now also uses the `--h5w-selector` variables. Didn't think it was worth renaming everything to `--h5w-menu`, especially since we use the term "menu" for the overflow menu even though it is technically more of a popover than a menu... Also, the `--h5w-selector` prefix is meant to describe the entire control, whereas ARIA considers that the "menu" is the container that pops open and contains the options. Anyways...

This is all a bit confusing, but the important thing is that `Selector` and `ExportMenu` share very similar styles (their "menus" both contain a number of clickable options), so they make use of the same variables (`--h5w-selector`) and classes (`.menu`, `.arrowIcon`, `.btnOption`, etc.) -- cf. `Toolbar.module.css`.

They no longer compose from the `popup` class (cf. `toolbar/utils.modules.css`), which is now used only by `DomainWidget`, `OverflowMenu` and `InteractionHelp` and has its own CSS variable (`--h5w-toolbar-popup--bgColor`, which defaults to `white` with 90% opacity, whereas `--h5w-selector-menu--bgColor` defaults to `white` with 100% opacity).

cypress/snapshots/app.cy.ts/heatmap_2D_inverted_cmap.snap.png
axelboc, line 1: Regression in the alignment of the controls in the overflow menu (right-aligned brings them closer to the trigger). Note also that I've increased the clickable area of the buttons (orange outline).

packages/lib/src/toolbar/controls/DomainWidget/DomainWidget.module.css
```diff
@@ -20,18 +20,21 @@
...
 
-.popup {
-  composes: popup from '../../utils.module.css';
-  z-index: 2; /* above overflow and selector menus */
+.popupOuter {
+  position: absolute;
+  z-index: 2; /* above other floating elements (overflow menu, selectors, etc.) */
+  bottom: 1px; /* guarantees overlap with toolbar so the popup doesn't close when the pointer moves into it */
+  padding-top: 6px; /* matches other floating elements */
+  transform: translateY(100%);
```
axelboc, line 28: We used to have two wrappers: `popup` and `popupInner`, but Floating UI no longer requires the extra level nesting; `InteractionHelp` and `OverflowMenu` only need the `popupInner` styles (the background color and box shadow).

So I rename `popupInner` to `popup`, and since the old `popup` class is now only used in `DomainWidget`, I move it here and rename it to `popupOuter`.

packages/lib/src/toolbar/controls/ExportMenu.tsx
```diff
@@ -13,8 +13,8 @@ import { useId, useRef, useState } from 'react';
 import { FiDownload } from 'react-icons/fi';
 import { MdArrowDropDown } from 'react-icons/md';
 
+import toolbarStyles from '../Toolbar.module.css';
 import { useFloatingDismiss } from './hooks';
-import styles from './Selector/Selector.module.css';
```
axelboc, line 17: The styles shared between `Selector` and `ExportMenu` now live in `Toolbar.module.css`.

packages/lib/src/toolbar/controls/InteractionHelp.tsx
```diff
@@ -14,35 +22,61 @@ interface Props {
...
-    <div className={overflowMenuStyles.root} ref={rootRef}>
-      <ToggleBtn
-        icon={FiHelpCircle}
-        iconOnly
-        onToggle={toggleHelpMenu}
-        label="Show help"
-        value={isHelpMenuOpen}
-      />
-      <div className={overflowMenuStyles.menu} hidden={!isHelpMenuOpen}>
-        <div className={overflowMenuStyles.menuList}>
```
axelboc, line 36: This is now a `ul`.

packages/lib/src/toolbar/controls/Selector/Selector.tsx
```diff
@@ -143,31 +142,29 @@ function Selector<T extends string>(props: Props<T>) {
                 </Option>
               ))
             ) : (
-              <ul className={styles.list}>
+              <ul className={styles.groups}>
                 {Object.entries(options).map(([groupLabel, groupOptions]) => (
                   <li key={groupLabel}>
                     <span className={styles.groupLabel}>{groupLabel}</span>
-                    <ul className={styles.list}>
```
axelboc, line 150: Removing the list markup inside each group. Too much noise for screenreader users.

packages/lib/src/toolbar/controls/Selector/models.ts
```diff
@@ -1,6 +1,3 @@
 import type { ReactElement } from 'react';
 
-export type OptionComponent<T> = (props: {
-  option: T;
-  isSelected?: boolean;
```
axelboc, line 5: This prop doesn't seem to be needed anymore (to differentiate an option rendered inside the floating menu or on the trigger once selected).

loichuder (approved):

PR #1652, opened 2024-05-22
Refactor `Selector` API and implement `ComplexVisTypeSelector`

axelboc:
**BREAKING** 

- I replace the `optionComponent` prop of `Selector` with a cleaner `renderOption` prop. This simplifies a few things, including removing the need for dedicated option components (`ScaleOption` and `ColorMapOption`).
- I replace the now-removed `ColorMapOption`, which was exported in the lib to address #1377, with a new component called `ColorMapGradient` that takes care only of rendering a little gradient preview for a given color map (which should in fact cover more customisation use cases).

**NEW FEATURE**

- I create a component called `ComplexVisTypeSelector` (in line with `ColorMapSelector` and `ScaleSelector`) to reduce duplication, and I export it in the lib;
- I move enum `ComplexVisType` to the shared package and I export it in the lib; I also replace its values and stop relying on them for display purposes (axis label, plot title, selector options), as this seemed very brittle and prevented displaying different labels in different places.


packages/app/src/vis-packs/core/complex/config.tsx
```diff
@@ -20,7 +20,7 @@ function createComplexConfigStore() {
       }),
       {
         name: 'h5web:complex',
-        version: 2,
+        version: 3,
```
axelboc, line 23: The values of the `ComplexVisType` enum have changed, so I bump the versions of the complex configs to avoid issues.

packages/lib/src/toolbar/controls/Selector/Selector.tsx
```diff
@@ -87,7 +80,7 @@ function Selector<T extends string>(props: Props<T>) {
 
   function handleSelect(index: number, option: T) {
     setSelectedIndex(index);
-    toggle(false);
+    context.onOpenChange(false);
```
axelboc, line 83: In case I decide to hide the `isOpen` state and `toggle` setter (along with `useFloating`) inside a reusable custom hook.

loichuder (approved):

PR #1653, opened 2024-05-22
Clean up toolbar deprecations

axelboc:
Since #1652  has a breaking change, I go ahead and remove some deprecated components, as well as the obsolete `align` prop of `ExportMenu`. I also re-use the generic `Btn` component in `ExportMenu`, `InteractionHelp` and `OverflowMenu` and fix a small ARIA bug.

**BREAKING**

- Remove deprecated components `GridToggler` and `FlipYAxisToggler` — use `ToggleBtn` directly instead.
- `[ExportMenu]` Remove prop `align` — the popup now remains within the bounds of the viewport automatically, thanks to Floating UI.
- `[Btn]` Prop `onClick` now receives the `MouseEvent` as first parameter, so make sure you don't pass a state setter as is, like `onClick={toggleOpen}`.

**NEW FEATURE**

- `[Btn]` Add prop `withArrow` to display a drop-down arrow.
- `[Btn]` Forward `ref`.
- `[Btn]` Allow forwarding any valid attributes to the `button` element (`HTMLAttributes<HTMLButtonElement>`)


packages/lib/src/toolbar/controls/Btn.tsx
```diff
@@ -1,48 +1,51 @@
...
+    withArrow,
     disabled,
-    onClick,
-    ...ariaAttrs
+    ...btnProps
   } = props;
 
   return (
     <button
+      ref={ref}
```
axelboc, line 31: Forwarding the `ref` for Floating UI.

packages/lib/src/toolbar/controls/ExportMenu.tsx
```diff
@@ -68,25 +60,18 @@ function ExportMenu(props: Props) {
...
         id={referenceId}
-        className={toolbarStyles.btn}
-        type="button"
+        label={`Export${isSlice ? ' slice' : ''}`}
+        icon={FiDownload}
+        withArrow
         disabled={availableEntries.length === 0}
         aria-haspopup="menu"
-        aria-expanded={isOpen || undefined}
+        aria-expanded={isOpen}
```
axelboc, line 71: This is the small ARIA bug: when the popup is closed, the `button` should have attribute `aria-expanded="false"`.

loichuder (approved):

PR #1654, opened 2024-05-22
Refactor `Selector` and `ExportMenu` with custom hook

axelboc:
`Selector` and `ExportMenu` now call the same custom hook `useFloatingMenu`, which ensures they use the exact same Floating UI configuration (including the `size` middleware that sets the menu's max width and height).


loichuder (approved):
Looks good !

PR #1655, opened 2024-05-29
Improve h5grove/h5wasm API testing set-up

axelboc:
- I was following an issue on the Vitest repo and finally saw a way to move test configuration to environment variables.
- While I'm at it, I add two env vars to be able to skip running the h5grove/h5wasm API tests (they always fail on my laptop because of the lack of `float128` support)
- And I add a check to make sure that the h5grove support server is running before the test starts (to get a clearer error message).

### DIFF
--- a/packages/app/.env.test
+++ b/packages/app/.env.test
@@ -3,5 +3,11 @@
 # Copy and configure the variables below in your own `.env.test.local` file
 #####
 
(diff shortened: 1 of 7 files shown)


loichuder (approved):
Nice :sunglasses:

Issue #1621, opened 2024-04-22
String padding


axelboc:
### Discussed in https://github.com/silx-kit/h5web/discussions/1612

<div type='discussions-op-text'>

<sup>Originally posted by **benajamin** April 15, 2024</sup>
One of the difficulties new adopters of HDF5 come across is that there are many ways to write strings (fixed vs variable length, null terminated or not). Many tools don't clearly tell what they are reading or writing, and some programs will crash or malfunction if they don't get an expected string type. The hdfview utility is the one that I find most useful for debugging string issues and it would be great if the h5web viewer could also provide this detailed string info in the "inspect" mode and therefore help debug HDF5 string bugs.</div>

<div type='discussions-op-text'>

<sup>Originally posted by **benajamin** April 20, 2024</sup>
The string padding info would be great to have but can be hidden behind the Inspect accordion since it is probably only interesting for debugging. I don't think that there is a definite right or wrong type of padding to use, I have just seen incompatibilities between programs reading and writing HDF5 that caused bugs that were difficult to diagnose because the critical padding information was not easy to expose.</div>

axelboc:
- [PR open on h5grove](https://github.com/silx-kit/h5grove/pull/93) 
- [PR open on h5wasm](https://github.com/usnistgov/h5wasm/pull/72) 
- HSDS already exposes the padding:
    ![image](https://github.com/silx-kit/h5web/assets/2936402/17629228-a14a-41d1-a66c-18932685c919)

PR #1658, opened 2024-06-04
Normalise H5T enums and use them to initialise types

axelboc:
[h5wasm@0.7.5](https://github.com/usnistgov/h5wasm/releases) has just been released with the new `strpad` metadata property to address #1621.

Since this is another case where we need to map an [H5T enum](https://docs.hdfgroup.org/hdf5/v1_14/_h5_tpublic_8h.html#title3) number to a more readable string, and since both h5grove and h5wasm expose a few of those raw enum values now (dtype class, charset, sign, order), I thought I'd first normalise things a bit...

1. I create a `h5t.ts` file in the shared package to hold the H5T enums and their mappings to readable strings. 
The enums and their keys are named as closely to the HDF5 spec as possible (though I don't specify keys for errors (`-1`) and reserved values — I'll explain why in a comment.)
2. I refactor the dtype factories (`stringType()`, `intType()`, etc.) to accept H5T enum values; it is now there responsibility to convert those values to readable strings (`'little-endian'`, etc.) using the mappings in `h5t.ts` instead of the providers.

Hopefully this makes sense. I'll add some more comments.

### DIFF
--- a/packages/h5wasm/src/utils.ts
+++ b/packages/h5wasm/src/utils.ts
@@ -3,6 +3,7 @@ import {
   assertNonNull,
   isNumericType,
 } from '@h5web/shared/guards';
+import { H5T_CLASS, H5T_ORDER, H5T_SIGN } from '@h5web/shared/h5t';
 import type {
   Attribute,
   ChildEntity,
@@ -11,12 +12,7 @@ import type {
   ProvidedEntity,
   Shape,
 } from '@h5web/shared/hdf5-models';
-import {
-  DTypeClass,
-  Endianness,
-  EntityKind,
-  H5TClass,
-} from '@h5web/shared/hdf5-models';
+import { DTypeClass, EntityKind } from '@h5web/shared/hdf5-models';
 import {
   arrayType,
   bitfieldType,
@@ -29,7 +25,6 @@ import {
   referenceType,
   strType,
   timeType,
-  toCharSet,
   unknownType,
 } from '@h5web/shared/hdf5-utils';
 import type { Metadata } from 'h5wasm';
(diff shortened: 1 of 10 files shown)


packages/h5wasm/src/utils.ts
```diff
@@ -163,36 +158,41 @@ function parseAttributes(h5wAttrs: H5WasmAttributes): Attribute[] {
...
   const { type: h5tClass, size } = metadata;
 
-  if (h5tClass === H5TClass.Integer) {
+  if (h5tClass === H5T_CLASS.INTEGER) {
     const { signed, littleEndian } = metadata;
-    return intOrUintType(signed, size * 8, toEndianness(littleEndian));
+    return intOrUintType(
+      signed ? H5T_SIGN.SIGN_2 : H5T_SIGN.NONE,
+      size * 8,
+      littleEndian ? H5T_ORDER.LE : H5T_ORDER.BE,
```
axelboc, line 170: h5wasm exposes two booleans: `signed` and `littleEndian`. It might be better to expose the raw H5T enum values instead/as well, like in h5grove, since the enums have more than two values: https://docs.hdfgroup.org/hdf5/v1_14/_h5_tpublic_8h.html#title3 (`H5T_NSGN`, `H5T_ORDER_MIXED`, etc.)

packages/shared/src/h5t.ts
```diff
@@ -0,0 +1,53 @@
...
+  BE = 1,
+  VAX = 2,
+  MIXED = 3,
+  NONE = 4,
+}
+
+export enum H5T_SIGN {
+  NONE = 0, // unsigned
+  SIGN_2 = 1, // signed
+  NSGN = 2,
```
axelboc, line 31: Not sure what this enum member represents, TBH.

packages/shared/src/h5t.ts
```diff
@@ -0,0 +1,53 @@
...
+  [H5T_ORDER.BE]: 'big-endian',
+  [H5T_ORDER.VAX]: 'VAX',
+  [H5T_ORDER.MIXED]: 'mixed',
+  [H5T_ORDER.NONE]: 'none',
+} as const;
+
+export const H5T_TO_CHAR_SET = {
+  [H5T_CSET.ASCII]: 'ASCII',
+  [H5T_CSET.UTF8]: 'UTF-8',
+} as const;
```
axelboc, line 53: These are the mappings from H5T enums to readable strings. Could have used arrays, but this felt more explicit. The `as const` are important to generate the union types `CharSet` and `Endianness` in `hdf5-models.ts`.

As mentioned, I did not add error/reserved members to the enums, since the idea is that if such a value is received from a provider, the mapping will return `undefined` - e.g. `H5T_TO_ENDIANNESS[-1 as H5T_ORDER] => undefined`.

To make this safe, I added `| undefined` in the relevant places in `hdf5-models.ts`.

packages/h5wasm/src/utils.ts
loichuder: Is it planned than h5wasm eventually exposes the enum values ?

packages/shared/src/h5t.ts
loichuder: Even the HDF people don't know apparently: https://docs.hdfgroup.org/hdf5/v1_14/_h5_tpublic_8h.html#af7bfee2db210a12b9290eba85d730a71

Perhaps @t20100 has an idea ?

loichuder (approved):
Nicely done :+1:

packages/h5wasm/src/utils.ts
axelboc: Not yet, but I'll make a note of it.

PR #1659, opened 2024-06-04
Parse HDF5 string padding metadata

### LINKED ISSUE #1621: String padding

axelboc:
Fix #1621 

As discussed, it remains visible only under the _Inspect_ accordion:

image


package.json
```diff
@@ -54,6 +54,9 @@
     "wait-on": "7.2.0"
   },
   "pnpm": {
+    "overrides": {
+      "eslint-config-galex>typescript": "$typescript"
+    },
```
axelboc, line 59: ESLint complains about the new `module: 'preserve'` setting in `tsconfig`. Why now that I'm updating `h5wasm` and not when I actually switched to that setting and upgraded typescript, I have no idea :shrug:

tsconfig.json
```diff
@@ -1,12 +1,15 @@
 {
   "compilerOptions": {
-    "module": "preserve",
+    "module": "esnext",
+    "moduleResolution": "bundler",
```
axelboc, line 4: For some reason, ESLint started complaining about `preserve` after I upgraded `h5wasm`... 🤷 Reverting for now.

loichuder (approved):
Nice !

PR #1660, opened 2024-06-05
Bring in `react-suspense-fetch` into codebase

axelboc:
Since [`react-suspense-fetch`](https://github.com/dai-shi/react-suspense-fetch) is no longer maintained but provides a good abstraction for our data providers (probably even after the next [React 19](https://react.dev/reference/react/use) upgrade), and since we would really benefit from being able to introspect the internal cache of the `valueStore` to address #1578 properly once and for all, I bring in the library's source code into our codebase.

Of course, I also tweak it to my liking a bit:

- I swap the `Input` and `Result` generics (`<Result, Input>` just felt unnecessarily confusing).
- I remove the restriction to always have to call `prefetch` before `get`.
- I take advantage of the new `preset` method, which was added in a recent version we hadn't upgraded to because of the previous point, to remove our own cache for group children in `DataProvider`.

In my next PR, I'll try to use the stores' built-in features to cancel requests and evict aborted responses from the cache, instead of our own complicated code in `DataProviderApi`.

### DIFF
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -25,7 +25,6 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-icons": "5.2.1",
-    "react-suspense-fetch": "0.4.1",
     "three": "0.164.1"
   },
   "devDependencies": {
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -54,6 +54,9 @@
     "wait-on": "7.2.0"
   },
   "pnpm": {
+    "overrides": {
+      "eslint-config-galex>typescript": "$typescript"
+    },
```
axelboc, line 59: Please ignore this, it will be removed once I merge #1659 and restore `moduleResolution: "bundler"`.

loichuder: Okay :smile:

loichuder (approved):
Man, the `preset` addition is so nice :+1:

PR #1661, opened 2024-06-05
Abort requests with signal via fetch store

axelboc:
We used to set up our own axios `CancelToken` to cancel value requests but it's [deprecated](https://github.com/axios/axios?tab=readme-ov-file#canceltoken-deprecated) in favour of the web platform's [AbortController](https://github.com/axios/axios?tab=readme-ov-file#abortcontroller).

Moreover, in newer versions, react-suspense-fetch creates its own `AbortController` for each request and passes it on to the store's fetch function. This allows aborting requests (and then evicting them) through the stores instead of having to keep track of ongoing and cancelled requests ourselves in `DataProviderApi`.

The challenge was that react-suspense-fetch was not providing a way to tell which requests have been aborted in order to evict them from the cache when after the error boundary appears. So I had to tweak things a bit.

### DIFF
--- a/packages/app/src/vis-packs/VisBoundary.tsx
+++ b/packages/app/src/vis-packs/VisBoundary.tsx
@@ -23,7 +23,7 @@ function VisBoundary(props: Props) {
       fallbackRender={(args) => (
         <ErrorFallback className={visualizerStyles.vis} {...args} />
       )}
-      onError={() => valuesStore.evictCancelled()}
+      onError={() => valuesStore.evictErrors()}
     >
       <Suspense fallback={<ValueLoader message={loadingMessage} />}>
         {children}
--- a/packages/shared/src/react-suspense-fetch.ts
+++ b/packages/shared/src/react-suspense-fetch.ts
@@ -1,13 +1,14 @@
 type FetchFunc<Input, Result> = (
   input: Input,
-  options: { signal: AbortSignal },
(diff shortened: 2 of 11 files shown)


packages/app/src/vis-packs/VisBoundary.tsx
```diff
@@ -23,7 +23,7 @@ function VisBoundary(props: Props) {
       fallbackRender={(args) => (
         <ErrorFallback className={visualizerStyles.vis} {...args} />
       )}
-      onError={() => valuesStore.evictCancelled()}
+      onError={() => valuesStore.evictErrors()}
```
axelboc, line 26: Note that I now evict _all_ errored requests, not just the ones that were aborted, which means that requests that errored for other reasons will be retried the next time `valueStore.prefetch()` or `valueStore.get()` is called with the same inputs. This is a great unexpected benefit from this refactoring -- until now, users had to reload the app for errors to be cleared from the fetch cache. 🎉

packages/shared/src/react-suspense-fetch.ts
```diff
@@ -1,21 +1,25 @@
 type FetchFunc<Input, Result> = (
   input: Input,
-  options: { signal: AbortSignal },
+  signal: AbortSignal,
```
axelboc, line 3: Simplifying the `fetchFunc` signature.

PR #1662, opened 2024-06-05
Load compression plugins for virtual datasets with h5wasm

axelboc:
To address https://github.com/silx-kit/vscode-h5web/issues/43

Compression plugins required to read virtual datasets, which have compressed datasets as sources, will now be loaded automatically in both `H5WasmProvider` and `H5WasmLocalFileProvider`.

h5wasm [now exposes](https://github.com/usnistgov/h5wasm/releases/tag/v0.7.5) a `virtual_sources` array in the `Metadata` object, which I parse and store in our `Dataset` model. I then get the filters of every virtual source dataset located within the same file (`file_name: '.'`) and load the corresponding compression plugins.

By storing the virtual sources in the `Dataset` model, we open the door to showing this information in the `MetadataViewer` in the future if the provider supports it.

image


packages/h5wasm/src/h5wasm-api.ts
```diff
@@ -163,23 +163,26 @@ export class H5WasmApi extends DataProviderApi {
...
-        console.warn(
-          `Compression filter ${filter.id} not supported (${filter.name})`,
-        );
-        continue;
-      }
+    const pluginsToLoad = allFilters
+      .map(({ id }) => PLUGINS_BY_FILTER_ID[id])
+      .filter(isDefined);
 
+    for await (const plugin of pluginsToLoad) {
```
axelboc, line 185: This just bring the code in line with `H5WasmLocalFileProvider`.

loichuder (approved):

PR #1663, opened 2024-06-05
Remove useless fragments

axelboc:
I found an old stash from a past attempt at upgrading TypeScript from 5.0 to 5.2. Turns out I missed this refactoring opportunity when I looked at it again in #1636.

I had also taken some notes of interesting changes in 5.1 and 5.2 in case this is of interest:

### [Typescript 5.1](https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/)

- [Easier Implicit Returns for undefined-Returning Functions](https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#easier-implicit-returns-for-undefined-returning-functions): a `return` statement is still required for all code paths, and it feels weird having `return;` at the end of a function; I much prefer the explicitness of `return undefined;`. Might help simplify other cases that I'm not seeing. 🤷‍♀️
- [Decoupled Type-Checking Between JSX Elements and JSX Tag Types](https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#decoupled-type-checking-between-jsx-elements-and-jsx-tag-types) - at last, no more wrapping stuff in unnecessary fragments, like `return <>{children}</>`! 🎉
- [Linked Cursors for JSX Tags](https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#linked-cursors-for-jsx-tags) 🎉
- [Avoiding Unnecessary Type Instantiation](https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#avoiding-unnecessary-type-instantiation) - FYI, type checking performance improvements, notably for MUI

### [TypeScript 5.2](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/)

- [Easier Method Usage for Unions of Arrays](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#easier-method-usage-for-unions-of-arrays) - we can finally loop through arrays of type ~`number[] | TypedArray`~ `number[] | string[]` with  `filter`, `map`, etc. but now that I'm looking through the code, I can no longer recall a specific case where we had to work around this limitation... EDIT: we encountered this in `getBounds` with `NumArray`, a union of plain and typed arrays, which isn't covered here.
- [Copying Array Methods](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#copying-array-methods) - immutable `arr.toReversed()`, `arr.toSorted()`, etc. are now supported by TS ... but they still need to be polyfilled for older browsers, so not worth using yet.
- [Comma Completions for Object Members](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#comma-completions-for-object-members) - nice!
- [Inline Variable Refactoring](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#comma-completions-for-object-members) - cool also

### DIFF
--- a/packages/lib/src/interactions/SelectionTool.tsx
+++ b/packages/lib/src/interactions/SelectionTool.tsx
@@ -45,6 +45,7 @@ interface Props extends CommonInteractionProps {
   ) => ReactNode;
 }
 
+// eslint-disable-next-line sonarjs/cognitive-complexity
 function SelectionTool(props: Props) {
   const {
     id = 'Selection',
@@ -213,7 +214,7 @@ function SelectionTool(props: Props) {
   }
 
   assertDefined(rawSelection);
-  return <>{children(selection, rawSelection, isValid)}</>;
+  return children(selection, rawSelection, isValid);
 }
 
 export type { Props as SelectionToolProps };
--- a/packages/lib/src/vis/rgb/utils.ts
+++ b/packages/lib/src/vis/rgb/utils.ts
@@ -26,7 +26,7 @@ export function toRgbSafeNdArray(
 
   if (ndArr.dtype === 'int8') {
     return ndarray(
-      Uint8Array.from([...ndArr.data].map((val) => val + 128)),
+      Uint8Array.from(ndArr.data.map((val) => val + 128)),
       ndArr.shape,
     );
   }
--- a/apps/storybook/src/Annotation.stories.tsx
+++ b/apps/storybook/src/Annotation.stories.tsx
@@ -162,6 +162,5 @@ function PointerTracker(props: {
     setCoords([x, y]);
   });
 
-  // eslint-disable-next-line react/jsx-no-useless-fragment
-  return <>{coords ? children(...coords) : null}</>;
+  return coords ? children(...coords) : null;
 }
--- a/apps/storybook/src/SelectionTool.stories.tsx
+++ b/apps/storybook/src/SelectionTool.stories.tsx
@@ -89,7 +89,7 @@ export const ModifierKey = {
 export const Validation = {
   ...Rectangle,
   args: {
-    validate: ({ html }) => Box.fromPoints(...html).hasMinSize(100),
+    validate: ({ html }) => Box.fromPoints(...(html as Rect)).hasMinSize(100),
   },
 } satisfies Story;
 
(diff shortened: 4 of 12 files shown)


packages/lib/src/interactions/SelectionTool.tsx
```diff
@@ -45,6 +45,7 @@ interface Props extends CommonInteractionProps {
   ) => ReactNode;
 }
 
+// eslint-disable-next-line sonarjs/cognitive-complexity
```
axelboc, line 48: 🤷 😂

packages/lib/src/vis/rgb/utils.ts
```diff
@@ -26,7 +26,7 @@ export function toRgbSafeNdArray(
 
   if (ndArr.dtype === 'int8') {
     return ndarray(
-      Uint8Array.from([...ndArr.data].map((val) => val + 128)),
+      Uint8Array.from(ndArr.data.map((val) => val + 128)),
```
axelboc, line 29: Unrelated: no need to copy the array since the `map` doesn't mutate the source array.

loichuder: :fishing_pole_and_fish:

loichuder (approved):

loichuder:
> ### [TypeScript 5.2](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/)
> 
>     * [Easier Method Usage for Unions of Arrays](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#easier-method-usage-for-unions-of-arrays) - we can finally loop through arrays of type `number[] | TypedArray` with
>       `filter`, `map`, etc. but now that I'm looking through the code, I can no longer recall a specific case where we had to work around this limitation... 🤷‍♀️
> 

IIRC, it was for the bounds computation: https://github.com/silx-kit/h5web/blob/86848a7fe2fd9e11563276c3ddc41eae764711e6/packages/shared/src/vis-utils.ts#L128

axelboc:
> IIRC, it was for the bounds computation: 

Ooh nice! :fishing_pole_and_fish:

axelboc:
Arf no luck, seems to work only with unions of plain arrays, like `number[] | string[]`, not with unions of typed and plain arrays like in the case of the `NumArray` type in `getBounds`. :cry:

PR #1664, opened 2024-06-07
Refactor h5wasm providers to use same worker-based API

axelboc:
:warning: **Breaking change:** I've renamed `H5WasmProvider` to **`H5WasmBufferProvider`** for clarity.

The two h5wasm providers now both use the same worker-based API. In fact, they even use the same `H5WasmApi` class. Where it differs is in the two providers themselves: `H5WasmBufferProvider` calls a new method in the worker API called `remote.openFileBuffer(buffer)`, and `H5WasmLocalFileProvider` calls `remote.openLocalFile(file)` (previously named `openFile`). Both of these methods return the file's identifier provided by the HDF5 library, which is then passed to the `H5WasmApi` constructor. That's it!

The rest of the diff is basically about replacing `h5wasm-api.ts` with the content of `h5wasm-local-file-api.ts`, removing the `local` folder, and removing a whole bunch of obsolete utilities.

One significant benefit of this refactoring (in addition to removing a lot of code) is that consumers who use the two providers no longer end up with h5wasm twice in their bundle. For instance, the demo's bundle was basically cut in half, from 10 MB back down to 5 MB.

Additionally, I was so far testing only `H5WasmApi` (the one with the buffer) and not `H5WasmLocalFileApi` (because of the web worker). Since we now have a single `H5WasmApi`, I had no choice but to get the tests to work with it. This was challenging because Vitest runs in a Node environment, which doesn't support Web Workers. There's a package called `@vitest/worker` that simulates workers in Node, but it is not compatible with Comlink... In the end, the very simple solution I found is to import and use the worker API object directly (no Web Worker, no Comlink).


packages/h5wasm/src/H5WasmBufferProvider.tsx
```diff
@@ -13,13 +14,22 @@ interface Props {
...
 
   const prevApiRef = useRef<H5WasmApi>();
 
   const api = useMemo(() => {
-    const newApi = new H5WasmApi(filename, buffer, getExportURL, getPlugin);
+    const remote = getH5WasmRemote();
+    const fileId = remote.openFileBuffer(buffer);
+
+    const newApi = new H5WasmApi(
+      remote,
```
axelboc, line 27: Technically, since `remote` is a singleton, I could call `getH5WasmRemote` again inside `H5WasmApi`. However, this signature allows me to pass the worker API object directly when testing (and not call `getH5WasmRemote` at all).

loichuder (approved):
:100:

Issue #1572, opened 2024-02-12
Keep zoom level and axis positions fixed between files


jpmvferreira:
Given that this tool is made to quickly have a glance at data (1D, 2D or 3D), I think it would be convenient to add an option that allows the user to keep the same zoom level and axis positions fixed as the user is browsing through different files.

To illustrate what I mean, let us say that I have a HDF5 file where each dataset corresponds to a 3D array at a given time.
If I want to see the differences at z=0 with time, I just set a fixed data range for the color map and browse the different files.

https://github.com/silx-kit/h5web/assets/41581666/cc0e6741-f674-49be-8877-80c2f981f8a5

However, if I happen to want to compare the snapshots at different values of z, then every time I change between datatsets I have to manually adjust the z coordinate correspondingly (apologies for the ghost cursor in the next video).

https://github.com/silx-kit/h5web/assets/41581666/3803232e-8b1a-442f-96e4-5217e7fb3c4a

Also, if I only want to see differences in a particular region, the zoom level also changes as I go through the datasets.

https://github.com/silx-kit/h5web/assets/41581666/d1a27497-f93f-4768-a411-77265cfc1120

The two previous issues are also true when it comes to line plots, with an additional issue: whereas in the heatmap you can choose an arbitrary data range, here you can either set "auto-scale" on or off, but off does not allow you to pick any values whatsoever.

Apologies for not following the feature request template that was provided, but I feel like I could explain myself better this way.

axelboc:
@jpmvferreira thanks for the feedback! It's really three requests in one, but we're on board with all of them. 😊 Here is what I'm thinking of implementing in more details:

- [ ] Maintain dimension mapping+slicing state when switching between two datasets with the same exact dimensions, and forget any memorised state as soon as a dataset with different dimensions is selected.
- [ ] Show _Pin zoom_ button next to _Reset zoom_ in floating toolbar once the user starts zooming; when the user toggles it on, maintain zoom level across datasets for the corresponding low-level visualization (heatmap, line, RGB, etc.) until the user toggles it off, clicks "Reset zoom", or zooms out completely.
- [ ] Replace _Auto-scale_ button in _Line_ vis toolbar with custom domain slider like in _Heatmap_ vis toolbar.

Would that answer your needs?

jpmvferreira:
Yes those looks like perfect solutions!

joaonizer:
[...]

PR #1665, opened 2024-06-10
Replace auto-scale toggle with domain widget on line vis

axelboc:
As suggested in https://github.com/silx-kit/h5web/issues/1572, I replace the _Autoscale_ toggle with a custom domain widget on the line (and complex line) visualization.

By removing the _Autoscale_ toggle, we loose the ability to find the domain of the entire dataset and to set it automatically as the domain of the visualization.

However, the toggle also came with the side effect of triggering a fetch of the entire dataset, which could have very bad consequences with very large datasets (e.g. unresponsive server or browser), so I think removing it is a good thing. I also like the fact that the behaviour (and code) is now consistent between the line and heatmap visualizations.

Peek 2024-06-10 15-42


packages/app/src/vis-packs/core/complex/MappedComplexLineVis.tsx
```diff
@@ -29,43 +32,68 @@ function MappedComplexLineVis(props: Props) {
...
           <ComplexLineToolbar
-            disableAutoScale={dims.length <= 1} // with 1D datasets, `baseArray` and `dataArray` are the same so auto-scaling is implied
+            dataDomain={dataDomain}
             config={config}
             lineConfig={lineConfig}
           />,
           toolbarContainer,
         )}
 
-      <MappedLineVis
```
axelboc, line 58: Since I now need to pass a domain to the complex line toolbar here, I can't just let `MappedLineVis` compute it any more — I have to do it in this component. So I decided to remove the coupling between this component and `MappedLineVis`. It was the only "mapped" component doing something like this, so it's consistent with the others now.

packages/app/src/vis-packs/core/line/config.tsx
```diff
@@ -44,24 +48,12 @@ function createLineConfigStore() {
...
-        toggleAutoScale: () => {
-          set((state) => ({ autoScale: !state.autoScale }));
-        },
-
         showErrors: true,
         toggleErrors: () => set((state) => ({ showErrors: !state.showErrors })),
       }),
       {
         name: 'h5web:line',
-        partialize: (state) => ({
```
axelboc, line 57: We no longer need to remove `autoScale` before persisting.

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -99,58 +99,45 @@ export function useMappedArray<T extends unknown[] | TypedArray | undefined>(
   value: T,
   dims: number[],
   mapping: DimensionMapping,
-  autoScale?: boolean,
```
axelboc, line 102: It's as if `autoScale` was always true, now, so the `value` array always corresponds to the slice, never to the whole dataset. This opens the door to a bunch of simplifications, since it means we no longer need to 1) slice datasets in `applyMapping`, and 2) return a separate ndarray from `useMappedArray(s)` (which we needed to compute the domain of the entire dataset).

For now, I just remove the `autoScale` parameter. I will do more refactoring separately.

loichuder (approved):
Amazing how this led to all these simplifications :astonished:

PR #1673, opened 2024-06-18
Add way to reset fetch cache of relevant providers

axelboc:
Addresses https://github.com/silx-kit/h5web/discussions/1553#discussioncomment-9787236

I was wrong about `key` not resetting the state of children components, so I'm adding a `resetKeys` array prop to three providers: `H5GroveProvider`, `HsdsProvider` and `H5WasmLocalFileProvider` to allow resetting the fetch cache manually. This prop is inspired from react-error-boundary.

I've tested and it works great: the suspense fallbacks in the explorer and in the visualizer are triggered again as expected. This even works with local files, which will benefit the VS Code extension.

### DIFF
--- a/packages/app/README.md
+++ b/packages/app/README.md
@@ -286,16 +286,14 @@ specific export scenarios. In this case, or if you don't provide a function at
 all, `H5GroveProvider` falls back to generating URLs based on the `/data`
 endpoint and `format` query param.
 
-#### `key?: Key` (optional)
+#### `resetKeys?: unknown[]` (optional)
 
-If the content of the current file changes and you want to ensure that the
-viewer refetches the latest metadata and dataset values, you can take advantage
-of
-[React's `key` attribute](https://react.dev/reference/react/useState#resetting-state-with-a-key).
-Changing the value of the `key` will force a remount of `H5GroveProvider` and
-clear its internal fetching cache.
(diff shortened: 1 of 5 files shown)


packages/app/README.md
```diff
@@ -286,16 +286,13 @@ specific export scenarios. In this case, or if you don't provide a function at
...
 
-If the content of the current file changes and you want to ensure that the
-viewer refetches the latest metadata and dataset values, you can take advantage
-of
-[React's `key` attribute](https://react.dev/reference/react/useState#resetting-state-with-a-key).
-Changing the value of the `key` will force a remount of `H5GroveProvider` and
-clear its internal fetching cache.
+Use this prop to reset the provider's internal fetch cache, for instance when
+the content of the current file changes and you want the viewer to refetch the
+latest metadata and dataset values.
```
loichuder, line 293: I think we should keep the link to the React doc that explains what a key is.

Or explain a bit more how to use this prop:

> In `resetKeys`, you can put variables that when changed, will reset the provider's internal fetch cache.

loichuder (approved):

packages/app/README.md
axelboc: Good call and good formulation :+1: 

I'm not adding a link to the React doc, as I feel it could create confusion.

PR #1674, opened 2024-06-19
Visualize scalar compounds

### LINKED ISSUE #1649: Show field names in Scalar view for Compound Datatypes

Right now when viewing an array of compound datatypes there are column headers that contain the field names. When viewing a scalar compound value that isn't the case and it becomes quite useless that way.

Implement same view for scalar values as for arrays.

axelboc:
Fix #1649

image

- I've made the _Compound_ visualization its own thing (it still uses `MatrixVis` internally, but the tab shows "Compound").
- I'm reusing `CompoundVisContainer` and `MappedCompoundVis` (previously `CompoundMatrixVisContainer` and `MappedCompoundMatrixVis`) so they now handle both scalar and array datasets.


packages/lib/src/vis/matrix/Grid.tsx
```diff
@@ -31,7 +32,8 @@ function Grid(props: ClassStyleAttrs) {
           rowCount={rowCount}
           columnWidth={cellSize.width}
           columnCount={columnCount}
-          overscanCount={8}
+          overscanColumnCount={8}
+          overscanRowCount={16}
```
axelboc, line 36: Deprecation

packages/app/src/setupTests.ts
```diff
@@ -53,3 +55,27 @@ afterEach(() => {
...
+
+// Mock the size of elements measured with `useMeasure`
+// Use attribute `data-test-size="<width>,<height>"` to specify the element's size
+vi.mock('@react-hookz/web', async (importOriginal) => {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+  const mod = await importOriginal<typeof import('@react-hookz/web')>();
+
+  return {
+    ...mod,
+    useMeasure: () => {
```
axelboc, line 67: `useMeasure` relies on `ResizeObserver` which is not supported by JSDOM. As a result, the `MatrixVis` visualization used to render an empty container. I now fake the width and height of the container so that `react-window` renders something.

packages/app/src/test-utils.tsx
```diff
@@ -81,6 +81,10 @@ export async function renderApp(
 
     selectVisTab: async (name) => {
       await user.click(await screen.findByRole('tab', { name }));
+
+      if (!withFakeTimers) {
+        await waitForAllLoaders();
+      }
```
loichuder, line 87: What is this for ?

loichuder (approved):

packages/app/src/test-utils.tsx
axelboc: Ah, so far we were using `selectVisTab` only in `VisSelector.test.ts`, to assert that the selected tab had changed. Now we're using it to test that the selected visualization renders correctly, so we need to wait for the data to load, like we do in `selectExplorerNode`.

PR #1676, opened 2024-06-20
Support complex auxiliaries

### LINKED ISSUE #1670: Complex auxiliary signals
### Discussed in https://github.com/silx-kit/h5web/discussions/1669

<sup>Originally posted by **joaonizer** June 12, 2024</sup>
Hi,

I have a group `SParameters` that is a `NXentry` and subgroups `S11`,..,`SNN` that are `NXdata`. Each one of them has their own `data`. Within `SParameters` I've created `SoftLink()`'s to the `data` in the other subgroups so I can plot them all using `auxiliary_signals` in a new `NXdata` called `ALL`. You can see them all here:

image

Unfortunately, the  display says "Expected dataset to have numeric type"

I am following the example under `spectrum_with_aux` in [h5web:mock](https://h5web.panosc.eu/mock):

image

image

`axes` is not an array as the data is 1D. I even tried adding `"."` and making it an array but it did not work.

Here is the .hdf5 file:
[data_file.zip](https://github.com/user-attachments/files/15809334/wg_straight_2024-06-12.18-11-20-493584_e5a788a0-cb5e-4141-81ab-e418940d8feb.zip)

Again, thanks for the help! 🙇‍♂️ </div>

axelboc:
Fix #1670 

image


packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -147,7 +148,6 @@ export function findAssociatedDatasets(
     assertDefined(dataset, `Expected child entity "${name}" to exist`);
     assertDataset(dataset, `Expected child "${name}" to be a dataset`);
     assertArrayShape(dataset);
-    assertNumericType(dataset);
```
axelboc, line 150: I struggled with type-checking a bit because of this type assertion and because `axes` datasets are still expected to have a numeric type, not a complex type. I considered asserting for numeric or complex type here conditionally based on the `type` parameter, but this felt like a really bad idea.

In the end, it was much easier to remove the assertion from here and to move it inside `findAxesDatasets` instead. I then created a new `findAuxiliaryDatasets` function that asserts for numeric or complex type.

At the end we still get `NxData<NumericType | ComplexType>` (but now with both the signal and auxiliaries of type `Dataset<ArrayShape, NumericType | ComplexType>`), and we narrow that to just `NumericType` or just `ComplexType` with `assertNumericNxData` or `assertComplexNxData`.

.github/workflows/lint-test.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.cache/pypoetry/virtualenvs
-          key: poetry-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
+          key: poetry-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}-0 # increment to reset cache
```
axelboc, line 98: I had an issue with a corrupted Poetry cache in the CI.

loichuder (approved):
Cool stuff! 

Do you intend to also add supports of aux for `NxComplexImage` ? You already made more than half of the way there.

axelboc:
> Do you intend to also add supports of aux for NxComplexImage ? You already made more than half of the way there.

Yeah, might as well

PR #1680, opened 2024-06-24
Support complex image auxiliaries

axelboc:
image


cypress/e2e/app.cy.ts
```diff
@@ -399,29 +399,29 @@ describe('/mock', () => {
       }
     });
 
-    it('visualize 2D complex signal as NxImage', () => {
```
axelboc, line 402: I've swapped the two mock complex `NXdata` groups to have the spectrum first and the image second, for consistency with the other mock `NXdata` groups, and I've renamed the `complex` `NXdata` group to `complex_image` for clarity.

This is why all the snapshots have changed below and why the diff of `app.cy.ts` looks weird here (since I've swapped the specs as well).

cypress/snapshots/app.cy.ts/nximage_complex_2d.snap.png
axelboc, line 1: Note the signal picker in this screenshot.

axelboc:
@loichuder merging so I can work on the release, but feel free to double check when you get the time.

loichuder:
All good!

PR #1688, opened 2024-08-01
Support visualizing enum datasets and signals

### LINKED ISSUE #1498: Need H5Web to support plot True/False and Enum signals

axelboc:
A bunch of commits to make enum datasets first-class citizens and fix #1498.

I've made separate commits, but the gist is that I now treat enum datasets as "numeric-like", just like I did witth booleans. All providers already return numeric values anyway, so it was relatively easy.

In addition to supporting enum datasets in Line/Heatmap, and enum signals/auxiliaries in NX Line/Heatmap, I also add enum support to the Matrix and Scalar vis. In these two visualizations, I even display the enum key rather than the numeric value.

image

image

image

image


loichuder (approved):
Looks good. 

But I wonder: you assume that enums always have numeric underlying values. Is that always the case ? Can HDF5 enum have string values ?

axelboc:
> Apparently it is always integers: https://docs.hdfgroup.org/hdf5/v1_12/group___e_n_u_m.html ?

Yeah I remember seeing that it's typically uint8 but allowed to be signed or unsigned int. No floats, strings, etc.

PR #1692, opened 2024-08-07
Upgrade h5wasm and improve boolean and vlen support

axelboc:
Two changes in [h5wasm v0.7.6](https://github.com/usnistgov/h5wasm/releases/tag/v0.7.6):

### Full support for the `H5T_VLEN` datatype (metadata + value)

This brings h5wasm in line with h5grove. The only difference is that, for numeric vlen datasets, h5wasm is able to return an array of vlen typed arrays (`TypedArray[]`), whereas h5grove is forced to return an array of plain arrays (`number[][]`) -- this doesn't make any difference, in the viewer though.

image
image

> See https://github.com/usnistgov/h5wasm/issues/78 and https://github.com/usnistgov/h5wasm/pull/77

### Booleans are now returned as numbers (like enums)

This removes the unnecessary `number->boolean->number` conversion for the line and heatmap vis.

Since h5grove and HSDS still return booleans for now, I decided to allow for both throughout (`Primitive<BooleanType` => `number | boolean`). As a result:

- If `MappedLineVis` and `MappedHeatmapVis` receive booleans, they convert them to numbers as before (via `useToNumArray`).
- If the _Scalar_ and _Matrix_ vis receive numbers, they put them through a new formatter called `formatBool` to convert them to strings (just like they do with enums). This is quite beneficial for the _Matrix_ vis, since it means only the values rendered on the screen are converted.

> See https://github.com/usnistgov/h5wasm/pull/76

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,13 @@
     "wait-on": "7.2.0"
   },
   "pnpm": {
+    "packageExtensions": {
+      "@typescript-eslint/eslint-plugin": {
+        "dependencies": {
+          "typescript": "5.0.3"
+        }
+      }
+    },
     "peerDependencyRules": {
       "allowedVersions": {
         "@phenomnomnominal/tsquery>typescript": "5.x",
--- a/packages/app/src/providers/h5grove/__snapshots__/h5grove-api.test.ts.snap
+++ b/packages/app/src/providers/h5grove/__snapshots__/h5grove-api.test.ts.snap
@@ -1895,6 +1895,37 @@ exports[`test file matches snapshot 1`] = `
     ],
   },
   {
+    "name": "vlen_float32_1D",
+    "rawType": {
+      "base": {
+        "class": 1,
+        "dtype": "<f4",
+        "order": 0,
+        "size": 4,
+      },
+      "class": 9,
+      "dtype": "|O",
+      "size": 16,
+    },
+    "shape": [
+      2,
+    ],
+    "type": {
+      "base": {
+        "class": "Float",
+        "endianness": "little-endian",
+        "size": 32,
+      },
+      "class": "Array (variable length)",
+    },
+    "value": [
+      [],
+      [
+        0,
+      ],
+    ],
+  },
+  {
     "name": "vlen_int64_1D",
     "rawType": {
       "base": {
--- a/packages/app/src/vis-packs/core/matrix/utils.ts
+++ b/packages/app/src/vis-packs/core/matrix/utils.ts
@@ -1,6 +1,12 @@
 import { Notation } from '@h5web/lib';
-import { isComplexType, isEnumType, isNumericType } from '@h5web/shared/guards';
+import {
+  isBoolType,
+  isComplexType,
+  isEnumType,
+  isNumericType,
+} from '@h5web/shared/guards';
 import type {
+  BooleanType,
   ComplexType,
   NumericType,
   PrintableCompoundType,
@@ -11,6 +17,7 @@ import type { ValueFormatter } from '@h5web/shared/vis-models';
 import {
   createComplexFormatter,
   createEnumFormatter,
+  formatBool,
 } from '@h5web/shared/vis-utils';
 import { format } from 'd3-format';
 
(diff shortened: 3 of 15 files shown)


package.json
```diff
@@ -53,6 +53,13 @@
     "wait-on": "7.2.0"
   },
   "pnpm": {
+    "packageExtensions": {
+      "@typescript-eslint/eslint-plugin": {
+        "dependencies": {
+          "typescript": "5.0.3"
```
axelboc, line 59: Temporary workaround until galex is updated, to prevent an unwanted change in the lockfile and ensure that `@typescript-eslint` runs with a version of typescript it supports.

packages/app/src/providers/h5grove/__snapshots__/h5grove-api.test.ts.snap
```diff
@@ -1894,6 +1894,37 @@ exports[`test file matches snapshot 1`] = `
       1,
     ],
   },
+  {
+    "name": "vlen_float32_1D",
```
axelboc, line 1898: Additional vlen dataset to test two things:

- what happens when an item in the vlen dataset is not initialized => providers should return an empty array or typed array;
- that h5wasm returns typed arrays (can't test this with the `vlen_int64_1D` dataset below since bigint typed arrays are converted to plain number arrays).

See also Python code in `create_h5_sample.py`.

packages/app/src/vis-packs/core/matrix/utils.ts
```diff
@@ -50,11 +57,15 @@ export function getFormatter(
     return createNumericFormatter(notation);
   }
 
+  if (isBoolType(type)) {
+    return formatBool as ValueFormatter<BooleanType>;
```
axelboc, line 61: Still hoping to improve the typing of those formatters at some point...

packages/h5wasm/src/__snapshots__/h5wasm-api.test.ts.snap
```diff
@@ -1722,21 +1731,30 @@ exports[`test file matches snapshot 1`] = `
           0,
           1,
         ],
-        "Uint8Array (unstable)",
+        [
+          0,
+        ],
```
axelboc, line 1736: not a typed array because the vlen's base dtype is `uint64`

packages/h5wasm/src/__snapshots__/h5wasm-api.test.ts.snap
```diff
@@ -1871,15 +1889,15 @@ exports[`test file matches snapshot 1`] = `
...
-    "value": [
-      true,
-      false,
-      true,
-      true,
-      false,
-      false,
-      true,
-      false,
+    "value": Int8Array [
```
axelboc, line 1892: Note that booleans are stored as `int8` by HDF5, not as `uint8`

packages/h5wasm/src/__snapshots__/h5wasm-api.test.ts.snap
```diff
@@ -2064,15 +2082,62 @@ exports[`test file matches snapshot 1`] = `
...
+    "type": {
+      "base": {
+        "class": "Float",
+        "endianness": "little-endian",
+        "size": 32,
+      },
+      "class": "Array (variable length)",
+    },
+    "value": [
+      Float32Array [],
```
axelboc, line 2136: - typed array 🎉 
- empty because no vlen array was explicitly assigned at this index 🎉

packages/lib/src/vis/raw/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { assertDefined, assertNonNull } from '@h5web/shared/guards';
+import type { TypedArray } from 'ndarray';
+
+export function getTypedArrayClass(arr: TypedArray): string {
+  const res = /^\[object (.+)\]$/u.exec(Object.prototype.toString.call(arr));
+  assertNonNull(res);
+
+  const [, arrClass] = res;
+  assertDefined(arrClass);
+  return arrClass;
```
loichuder, line 10: Since it is just for formatting perhaps we should return a default value (e.g. `TypedArray`) instead of asserting ?

loichuder (approved):

packages/lib/src/vis/raw/utils.ts
axelboc: For some reason, I had found a much simpler way to do this but hadn't pushed it: `value.constructor.name` :sweat_smile:

PR #1693, opened 2024-08-08
Use zustand store inside suspense fetch store to track progress

axelboc:
This replaces the pub-sub pattern I had implemented in `DataProviderApi`, and considerably simplifies subscribing to the progress state in `ValueLoader`.

Basically, every fetch store now initialises a zustand store to track progress. Each "fetch input" (e.g. dataset+selection for `valuesStore`) is associated with a progress percentage in the zustand store, and the store can be subscribed to with zustand's `useStore` hook.

### DIFF
--- a/packages/app/src/providers/api.ts
+++ b/packages/app/src/providers/api.ts
@@ -1,4 +1,3 @@
-/* eslint-disable promise/prefer-await-to-callbacks */
 import type {
   ArrayShape,
   AttributeValues,
@@ -7,6 +6,7 @@ import type {
   ProvidedEntity,
   Value,
 } from '@h5web/shared/hdf5-models';
+import type { OnProgress } from '@h5web/shared/react-suspense-fetch';
 import type {
   AxiosInstance,
   AxiosProgressEvent,
@@ -16,18 +16,10 @@ import type {
 } from 'axios';
 import axios from 'axios';
 
-import type {
-  ExportFormat,
-  ExportURL,
-  ProgressCallback,
-  ValuesStoreParams,
-} from './models';
+import type { ExportFormat, ExportURL, ValuesStoreParams } from './models';
 
 export abstract class DataProviderApi {
   protected readonly client: AxiosInstance;
-  protected progress = new Map<ValuesStoreParams, number>();
-
-  private readonly progressListeners = new Set<ProgressCallback>();
 
   public constructor(
     public readonly filepath: string,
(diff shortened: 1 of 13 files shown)


packages/app/src/providers/api.ts
```diff
@@ -93,21 +74,16 @@ export abstract class DataProviderApi {
...
-  private notifyProgressChange() {
-    this.progressListeners.forEach((cb) => cb([...this.progress.values()]));
-  }
-
   public abstract getEntity(path: string): Promise<ProvidedEntity>;
+
   public abstract getValue(
     params: ValuesStoreParams,
     signal?: AbortSignal,
+    onProgress?: OnProgress,
```
axelboc, line 85: Both `signal` and `onProgress` are optional so I don't have to pass them when testing the provider APIs.

packages/app/src/providers/mock/mock-api.ts
```diff
@@ -132,43 +136,6 @@ export class MockApi extends DataProviderApi {
   }
 
   public override async getSearchablePaths(path: string): Promise<string[]> {
-    return this.getEntityPaths(path);
-  }
-
-  private getEntityPaths(entityPath: string): string[] {
```
axelboc, line 138: Unrelated: I moved a couple of methods out of the `MockApi` class and turned them into utilities instead.

packages/app/src/vis-packs/ValueLoader.tsx
```diff
@@ -42,13 +37,17 @@ function ValueLoader(props: Props) {
...
-            <div className={styles.progressBars}>
-              {progress.slice(0, MAX_PROGRESS_BARS).map((val, index) => (
-                <progress className={styles.progress} key={index} value={val} /> // eslint-disable-line react/no-array-index-key
+          <div className={styles.progressBars}>
+            {[...ongoing.entries()]
+              .slice(0, MAX_PROGRESS_BARS)
+              .map(([key, val]) => (
+                <progress
+                  className={styles.progress}
+                  key={`${key.dataset.path}_${key.selection || ''}`}
```
axelboc, line 46: I now have access to the `Input` of the ongoing fetch — which, for the values store, is an object with a dataset and selection.

packages/shared/package.json
```diff
@@ -35,6 +35,9 @@
     "ndarray-ops": "1.2.2",
     "react": ">=18"
   },
+  "optionalDependencies": {
+    "zustand": "4.5.4"
```
axelboc, line 39: Optional, so I don't have to install it in the storybook package.

loichuder (approved):

PR #1694, opened 2024-08-08
Remove axios from `DataProviderApi` base class and refactor axios error handling

axelboc:
In this PR, I'm finally removing the `cancellableFetchValue` method from the `DataProviderApi` base class. This method made little sense ever since the arrival of h5wasm, which doesn't require making HTTP requests at all.

This should get rid of `axios` from consumer bundles that use `H5WasmBufferProvider` or `H5WasmLocalFileProvider`. It also means that if we expose `DataProviderApi` in the future, consumers may be able to develop their own network-based provider without having to use `axios` as the fetching library.

So it's now up to `H5GroveProvider` and `HsdsProvider` to initialise and use their own axios client. I had to refactor axios error handling a bit to make this a bit more palatable, but  overall I find the result less convoluted than before. I especially like that the `DataProviderApi` base class no longer has any method implementations.

### DIFF
--- a/packages/app/src/ErrorFallback.tsx
+++ b/packages/app/src/ErrorFallback.tsx
@@ -10,7 +10,27 @@ interface Props extends FallbackProps {
 function ErrorFallback(props: Props) {
   const { className = '', error, resetErrorBoundary } = props;
 
-  if (error.cause || error.cause instanceof Error) {
+  if (error.message === CANCELLED_ERROR_MSG) {
+    return (
+      <p className={`${styles.error} ${className}`}>
+        {CANCELLED_ERROR_MSG}
+        <span>–</span>
+        <button
+          className={styles.retryBtn}
+          type="button"
+          onClick={() => resetErrorBoundary()}
+        >
+          Retry?
+        </button>
+      </p>
+    );
+  }
+
+  if (
+    error.cause &&
+    error.cause instanceof Error &&
+    error.message !== error.cause.message
+  ) {
     const { message } = error.cause;
     return (
       <details className={`${styles.detailedError} ${className}`}>
@@ -20,23 +40,7 @@ function ErrorFallback(props: Props) {
     );
   }
 
-  return (
-    <p className={`${styles.error} ${className}`}>
-      {error.message}
-      {error.message === CANCELLED_ERROR_MSG && (
-        <>
-          <span>–</span>
-          <button
-            className={styles.retryBtn}
-            type="button"
-            onClick={() => resetErrorBoundary()}
-          >
-            Retry?
-          </button>
-        </>
-      )}
-    </p>
-  );
+  return <p className={`${styles.error} ${className}`}>{error.message}</p>;
 }
 
 export default ErrorFallback;
--- a/packages/app/src/providers/api.ts
+++ b/packages/app/src/providers/api.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/member-ordering */
 import type {
   ArrayShape,
   AttributeValues,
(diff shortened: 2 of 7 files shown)


packages/app/src/ErrorFallback.tsx
```diff
@@ -10,7 +10,27 @@ interface Props extends FallbackProps {
 function ErrorFallback(props: Props) {
   const { className = '', error, resetErrorBoundary } = props;
 
-  if (error.cause || error.cause instanceof Error) {
+  if (error.message === CANCELLED_ERROR_MSG) {
```
axelboc, line 13: Moving this up because I now use a new utility function called `processAxioError` to throw "Request cancelled" errors, and this utility adds the original error as `cause` when rethrowing. (And I don't need to render an accordion for "Request cancelled" errors, since the cause error message is simply "cancelled".)

packages/app/src/ErrorFallback.tsx
```diff
@@ -10,7 +10,27 @@ interface Props extends FallbackProps {
...
+          Retry?
+        </button>
+      </p>
+    );
+  }
+
+  if (
+    error.cause &&
+    error.cause instanceof Error &&
+    error.message !== error.cause.message
```
axelboc, line 32: No longer rendering an accordion if the cause error message is the same as the error message.

packages/app/src/providers/api.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/member-ordering */
```
axelboc, line 1: This rule forces me to have non-abstract methods before abstract methods. `getExportURL` and `getSearchablePaths` can't be abstract because they're optional (i.e. if I make them abstract, I have to implement them in all providers) but I still prefer them to appear after the main abstract methods like `getValue`).

packages/app/src/providers/utils.ts
```diff
@@ -6,32 +6,14 @@ import type {
...
-      const errorToThrow = getErrorToThrow(status, data);
-
-      if (errorToThrow) {
-        throw new Error(errorToThrow);
-      }
-    }
-
-    throw error;
-  }
-}
```
axelboc, line 33: This utility function was not very elegant to use. I replaced it with `processAxiosError` below, which basically contain what was inside the try/catch block of `handleAxiosError`.

Another change, however, is that `getMessageToThrow` receives the whole `AxiosError` instead of just the response. This allows using `processAxiosError` to detect and process "cancelled" errors (which don't have a response since they're cancelled).

packages/app/src/providers/utils.ts
```diff
@@ -94,3 +76,28 @@ export async function getValueOrError(
...
+    const messageToThrow = getMessageToThrow(error);
+    if (messageToThrow) {
+      return new Error(messageToThrow, { cause: error });
+    }
+  }
+
+  return error;
+}
+
+export function createAxiosProgressHandler(onProgress: OnProgress | undefined) {
```
axelboc, line 94: Just a small helper to avoid duplication.

packages/app/src/providers/utils.ts
```diff
@@ -94,3 +76,28 @@ export async function getValueOrError(
     return error;
   }
 }
+
+export function processAxiosError(
+  error: unknown,
+  getMessageToThrow: (error: AxiosError) => string | false | undefined,
```
loichuder, line 82: It feels weird to have to pass a callback when you already have access to the argument `error` in the caller. It makes the calling code way less readable IMO.

Perhaps pass `messageToThrow` directly and have it `undefined` if `error` is not an `AxiosError` ?

loichuder (approved):

packages/app/src/providers/utils.ts
axelboc: The goal was to abstract away the condition `error instanceof AxiosError`, but I agree that the callback is not super readable. In this case, I prefer to remove the abstraction altogether and inline the error handling in the provider APIs.

This is where wretch's [chainable API](https://github.com/elbywan/wretch?tab=readme-ov-file#catchers-) would really shine. Perhaps I will migrate to it at some point.

axelboc: I've pushed a separate commit: https://github.com/silx-kit/h5web/pull/1694/commits/47a4e9cc159726f42d66985af123e29a76e84adc

loichuder: Excellent, that's way clearer

PR #1698, opened 2024-08-21
Maintain dimension mapping when possible

axelboc:
This addresses the first part of issue #1572 (cf. first bullet point in https://github.com/silx-kit/h5web/issues/1572#issuecomment-1945556759).

"when possible" means when:

- toggling inspect mode;
- switching to a dataset with no dimension mapper (scalar dataset, group, etc.) and back;
- switching to a visualization with the same number of axes (e.g. between _Heatmap_ and _Matrix _);
- switching to a dataset with the exact same dimensions (both number and size).

maintain-dim-mapping

If the visualization changes when switching dataset, then the last two conditions must **both** be met for the dimension mapping state to be maintained.

In other words, the state is **maintained** when, for instance:

- switching between two `NXdata` groups with signals with exact same dimensions and interpretations;
- switching between an `NXData` group with a 1D signal (_NX Line_), and the signal itself (_Line_);
- switching between an `NXData` group with a 2D+ signal with implicit or explicit `image` interpretation (_NX Heatmap_), and the signal itself (_Heatmap_);
- switching between an `NXData` group with a 2D+ signal with `spectrum` interpretation (_NX Line_), and the signal itself if the `preferredVis` is _Line_.

maintain-dim-mapping-nx

However, the state is **reset** when, for instance:

- switching between two `NXData` groups with signals with exact same dimensions but different interpretations (e.g. `image` vs `spectrum`);
- switching between an `NXData` group with a 2D+ signal with `spectum` interpretation (_NX Line_), and the signal itself (_Heatmap_) when no `preferredVis` is set.


packages/app/src/App.tsx
```diff
@@ -84,21 +85,25 @@ function App(props: Props) {
...
-                  <MetadataViewer
-                    path={selectedPath}
-                    onSelectPath={onSelectPath}
-                  />
-                ) : (
-                  <Visualizer path={selectedPath} />
-                )}
-              </Suspense>
-            </ErrorBoundary>
+            <DimMappingProvider>
```
axelboc, line 88: Putting the new zustand store's provider in `App` to support maintaining dim mapping state when toggling inspect mode.

packages/app/src/__tests__/DimensionMapper.test.tsx
```diff
@@ -100,14 +100,92 @@ test('slice through 2D dataset', async () => {
   const { user } = await renderApp({
     initialPath: '/nD_datasets/twoD',
     preferredVis: Vis.Line,
-    withFakeTimers: true, // required since React 18 upgrade (along with `waitForAllLoaders` below)
```
axelboc, line 103: This workaround doesn't seem needed anymore.

packages/app/src/__tests__/DimensionMapper.test.tsx
```diff
@@ -100,14 +100,92 @@ test('slice through 2D dataset', async () => {
...
+
+  // Swap axes for D0 and D1
+  await user.click(getDimMappingBtn('x', 0));
+
+  // Switch to dataset with different dimensions
+  await selectExplorerNode('twoD_cplx');
+
+  expect(getDimMappingBtn('x', 0)).not.toBeChecked();
+  expect(getDimMappingBtn('x', 1)).toBeChecked();
+});
```
axelboc, line 191: Note that I purposely did not add tests for more complicated cases, like with `NXdata` groups and signals, as I consider them to be simple consequences of the more well-defined cases above. If we find a bug or a precise scenario where the dim mapping state is maintained when we would prefer it to be reset—or the other way around—then and only then I'll add a specific test for it.

packages/app/src/dimension-mapper/store.tsx
```diff
@@ -0,0 +1,72 @@
...
+        ...Array.from({ length: dims.length - axesCount }, () => 0),
+        ...(dims.length > 0
+          ? ['y' as const, 'x' as const].slice(-axesCount)
+          : []),
+      ]
+    : state.mapping;
+
+  useEffect(() => {
+    state.reset(dims, axesCount, mapping);
+  }, [isStale]); // eslint-disable-line react-hooks/exhaustive-deps
```
axelboc, line 69: Obviously, we don't want to reset the state whenever the `dims` and `mapping` array references change. Hopefully when React 19 comes out and we upgrade to it, we can wrap the `reset` call in an effect event so ESLint doesn't complain about the missing `useEffect` deps.

axelboc (commented):
Pretty straightforward in the end :tada:

packages/app/src/__tests__/DimensionMapper.test.tsx
loichuder: Yep, sounds good :+1:

loichuder (approved):
Looks good :clap:

PR #1703, opened 2024-09-03
Enforce stricter types for formatters in `MappedMatrixVis`

axelboc:
Follows #1702 

As planned, the problem became much simpler: all the typing complexity related to formatters in `MappedMatrixVis` can now be moved into the `getFormatter` utility function, now called `getCellFormatter`.

I add a type guard called `assertNdArrayValue` that allows narrowing down an `NdArray` given an already narrowed-down `DType`. After this type guard, `dataArray.get(row, col)` is properly typed and we can pass the value to a strictly typed formatter without any casting.

I'll do something similar in `ScalarVis`.

---

The broader issue is that every time we narrow down a dataset object (e.g. `Dataset<ArrayShape, PrintableType>` => `Dataset<ArrayShape, EnumType>`), we have to narrow down the value right after with `assertArrayValue`/`assertNdArrayValue`/`assertPrimitiveValue` (e.g. `ArrayValue<PrintableType>` => `ArrayValue<EnumType>`). This is because we don't store dataset values within dataset objects. Maybe I'll try to tackle this one day.

### DIFF
--- a/packages/app/src/vis-packs/core/compound/MappedCompoundVis.tsx
+++ b/packages/app/src/vis-packs/core/compound/MappedCompoundVis.tsx
@@ -14,7 +14,7 @@ import visualizerStyles from '../../../visualizer/Visualizer.module.css';
 import { useMappedArray, useSlicedDimsAndMapping } from '../hooks';
 import type { MatrixVisConfig } from '../matrix/config';
 import MatrixToolbar from '../matrix/MatrixToolbar';
-import { getCellWidth, getFormatter } from '../matrix/utils';
+import { getCellFormatter, getCellWidth } from '../matrix/utils';
 import { getSliceSelection } from '../utils';
 
 interface Props {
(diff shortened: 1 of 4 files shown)


axelboc:
Was a bit too quick on this one: it doesn't work at all with printable compound datasets...

Ideally, compound ndarrays would be typed with a tuple (`NdArray<[Primitive<Field0Type>, Primitive<Field1Type>, ...]>`) so that the return type of `compoundNdArray.get(0)` would be automatically inferred as `Primitive<Field0Type>` (which we could pass to a strictly typed formatter: `ValueFormatter<Field0Type>`).

Unfortunately, this doesn't work: the return type is always inferred as `Primitive<Field0Type> | Primitive<Field1Type> | ...`. It's a limitation of the `ndarray` library and I'm not sure it's fixable.

Seems that we're stuck with returning a weakly typed formatter from `getFormatter` (`ValueFormatter<PrintableType>`). Anyways, at least the typing weakness no longer leaks into `MatrixVis`, so it wasn't all worthless.

Issue #1679, opened 2024-06-20
64-bit integers not shown correctly in h5web display


gapost:
### Describe the bug

64-bit signed and unsigned integers are not always displayed correctly in the h5web display window.

### To Reproduce

1. Go to https://h5web.panosc.eu/h5grove?file=sample.h5
2. Select the `int64_scalar` dataset
3. Select `Display` view on the right panel
4. Value shown is: `-9223372036854776000`

### Expected behaviour

The value returned by `h5dump -d int64_scalar sample.h5` is: `-9223372036854775808`

Similar happens with uint64:
uint64_scalar datatype as shown on h5web window: `18446744073709552000`
while h5dump reports: `18446744073709551615`

### Context

- OS: Linux
- Browser: Firefox
- Version: 127.0
- H5Web context: h5web

axelboc:
Indeed, big integers (int/uint64) numbers are not currently supported. They get [converted](https://github.com/silx-kit/h5web/pull/1581) to primitive JS numbers (float64) [to avoid errors](https://github.com/silx-kit/h5web/issues/1536), which is why you're seeing inaccurate values.

We would definitely like to improve support for big integers, though, so we'll use this issue to report any progress. Thanks!

axelboc:
### Scope

The scope of this issue is to make bigints (as in primitive JS `bigint` values) **first-class citizens** in `@h5web/app` so that they can be displayed without any loss of precision in the _Scalar_, _Matrix_ and _Compound_ visualizations.

The scope does **not** include maintaining bigint precision in any of the WebGL-based visualizations like _Heatmap_, _Line_, _RGB_, and the NeXus visualizations, or showing exact bigint values in the tooltips of these visualizations.

### Challenges

#### Typing

`bigint` and `number` do not play well together in JS: most operators support only `number` or only `bigint` operands. For instance, `1 + 1n` throws a type error; instead, one has to cast one operand or the other: `1 + Number(1n)` (potential loss of precision) or `BigInt(1) + 1n`.

This means that if, in some parts of the codebase, we end up with the union type `number | bigint`, it can get ugly very fast. We have to avoid this, which means we also have to avoid `(number | bigint)[]` and `Int8Array | Uint8Array | ... | BigInt64Array | BigUint64Array`.

[...]

PR #1706, opened 2024-09-06
Expose utility function to enable bigint serialization in consumer apps

axelboc:
In preparation to adding support for bigints (cf. #1679), I add a few mock datasets with `bigint` literals and I implement an opt-in way to enable big int serialization for the _Raw_ visualization and metadata viewer in consumer apps.

For `JSON.stringify()` to support bigint, one has to implement `BigInt.prototype.toJSON`. I really don't want to do this in `@h5web/app` or `@h5web/lib` because extending built-in objects is [dangerous](https://eslint.org/docs/latest/rules/no-extend-native). If H5Web is integrated into an app that already implements its own JSON serialization, H5Web could unknowingly break this parent app. That's why I think it's safer to give this responsibility to the consumer app.

Since the goal is for providers to return bigint values/typed arrays if possible, and since there will always be cases where bigints need to be serialized for display (attributes, nested compounds, etc.), this should at least provide a much more robust and future-proof solution for #1536.

image

image

image


packages/shared/src/mock-values.ts
```diff
@@ -17,6 +17,11 @@ const range9 = () => range(1, 11);
 
 const oneD = () => ndarray(range1().map((val) => val ** 2));
 
+const oneD_bigint = () =>
+  ndarray(
+    range8().map((_, i) => BigInt(Number.MAX_SAFE_INTEGER) + BigInt(i - 5)),
```
axelboc, line 22: I make the values cross the max safe integer threshold. This will hopefully be interesting when visualizing as line/heatmap after casting to float64/float32.

packages/app/README.md
```diff
@@ -389,6 +389,40 @@ import { getFeedbackMailto } from '@h5web/app';
...
+[big integers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#bigint_type):
+
+```jsx
+enableBigIntSerialization();
+createRoot(document.querySelector('#root')).render(<MyApp />);
+```
+
+This is recommended if you're working with a provider that supports 64-bit
+integers — i.e. one that may provide dataset and attribute values that include
+primitive `bigint` numbers.
```
loichuder, line 405: Give an example of such a provider ?

packages/app/README.md
```diff
@@ -389,6 +389,40 @@ import { getFeedbackMailto } from '@h5web/app';
...
+
+The _Raw_ visualization and metadata inspector rely on `JSON.stringify()` to
+render dataset and attribute values. By default, `JSON.stringify()` does not
+know how to serialize `bigint` numbers and throws an error if it encounters one.
+`enableBigIntSerialization()` teaches `JSON.stringify()` to convert big integers
+to strings:
```
loichuder, line 411: I think what you wrote in the PR description is worth mentioning: 

> if H5Web is integrated into an app that already implements its own JSON serialization, H5Web could unknowingly break this parent app 

Corollary: if your app already implements `BigInt` serialization, `enableBigIntSerialization` is not needed

loichuder (approved):

PR #1709, opened 2024-09-09
Fetch boolean data from h5grove as binary

axelboc:
Follows #1692 

Requires [h5web@2.3.0](https://github.com/silx-kit/h5grove/releases/tag/v2.3.0), which I've just released. I've also updated the demo server on Bosquet.

I'll add client-side code to support datatypes in my next PR.


packages/app/src/providers/h5grove/__snapshots__/h5grove-api.test.ts.snap
```diff
@@ -1431,6 +1431,11 @@ exports[`test file matches snapshot 1`] = `
               "size": 64,
             },
             "bool": {
+              "base": {
+                "class": "Integer",
+                "endianness": "little-endian",
+                "size": 8,
+              },
```
axelboc, line 1438: Enums have a base type. I need to know what that is in order to use the right typed array constructor. It was already returned by h5grove, just not parsed.

Technically, it's probably always going to be `int8` with h5grove/h5py, but I prefer to be strict about it.

packages/app/src/providers/utils.ts
```diff
@@ -72,7 +73,7 @@ export async function getValueOrError(
   try {
     return await api.getValue({ dataset });
   } catch (error) {
-    return error;
+    return isAxiosError(error) ? error.message : error;
```
axelboc, line 76: Just a nicer output in the snapshot when a request errors during local testing.

packages/lib/src/vis/models.ts
```diff
@@ -87,12 +81,6 @@ export interface AxisParams {
 
 export type Coords = [x: number, y: number];
 
-export type PrintableType =
```
axelboc, line 90: Duplicate type that was used by `MatrixVis` (and actually missing `EnumType`) before I removed it in #1702 (I had done that multiple times, but it got lost in my many recent stashes).

packages/app/src/providers/utils.ts
```diff
@@ -8,13 +8,14 @@ import type {
...
 import type { AxiosProgressEvent } from 'axios';
+import { isAxiosError } from 'axios';
 
 import type { DataProviderApi } from './api';
 
 export const CANCELLED_ERROR_MSG = 'Request cancelled';
 
 export function typedArrayFromDType(dtype: DType) {
-  if (isEnumType(dtype)) {
+  if (isEnumType(dtype) || isBoolType(dtype)) {
```
loichuder, line 18: Sorry for the dumb question: isn't bool a particular type of `enum` ? So that `isEnumType` should encompass bool ?

loichuder (approved):

packages/app/src/providers/utils.ts
axelboc: We split the two in our internal types and the purpose of these guards is to narrow our internal types. We need two separate types so `Primitive<BooleanType>` can resolve to `number | boolean`. But we may be able to reconsider this choice with https://github.com/silx-kit/h5web/issues/1679#issuecomment-2333937430

PR #1712, opened 2024-09-16
Fix jitter when resizing explorer sidebar on Matrix vis

axelboc:
This fixes two jittering issues when resizing the explorer sidebar:

- Some control labels would sometimes wrap onto two lines. I noticed this with the _Cell width_ control on the _Matrix_ vis, but the same could most likely happen with `Selector` labels. I now prevent wrapping in both cases.
- The _Matrix_ visualization would overflow and cause sidebars to appear and disappear quickly.

| BEFORE | AFTER |
| --- | --- |
| matrix-resize-jitter | matrix-resize-jitter-fixed |


loichuder (approved):
Nice !

loichuder:
I wish this would have been merged before the demo last week :joy:

PR #1715, opened 2024-10-07
Upgrade TypeScript

axelboc:
TypeScript 5.5 can:

- [infer type prediactes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates), notably when filtering arrays - e.g. `myArr.filter(foo => !!foo).map(foo => foo.toString())`
- better [narrow down `obj[key]` expressions](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#control-flow-narrowing-for-constant-indexed-accesses)
- check the [syntax of regular expressions](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#regular-expression-syntax-checking)
- identify [missing type annotations](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#isolated-declarations) that may prevent future tools from type-checking in parallel with emitting declaration files (a bit early days to enable it, though)

TypeScript 5.6 can:

- catch condition that are [always true or always false](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#disallowed-nullish-and-truthy-checks) by mistake
- forbid [side effect imports](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#the---nouncheckedsideeffectimports-option) (too early to use due to ESLint)
- emit declarations without type-checking (too early to use, though)
- provide better [code completions](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#granular-commit-characters) based on commit characters
- be configured to [no longer suggest some auto-imports](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#exclude-patterns-for-auto-imports)
  - I'm taking advantage of this to force VS Code to stop trying to import `log` from Node's `console` or from Three.js (so the `console.log()` snippet becomes the first suggestion)


tsconfig.json
```diff
@@ -5,24 +5,19 @@
     "target": "esnext",
     "lib": ["dom", "dom.iterable", "esnext"],
     "jsx": "react-jsx",
-    "allowJs": true,
```
axelboc, line 8: Not actually needed.

tsconfig.json
```diff
@@ -5,24 +5,19 @@
...
     "skipLibCheck": true,
     "esModuleInterop": true,
     "isolatedModules": true,
     "resolveJsonModule": true,
     "noEmit": true,
     "incremental": true,
     "sourceMap": true,
-    "declaration": true,
-    "declarationMap": true,
```
axelboc, line 17: Moved to `tsconfig` files in `packages` subfolders, where it is actually needed. This was causing issues in the `storybook` project.

tsconfig.json
```diff
@@ -5,24 +5,19 @@
...
     "removeComments": true,
     "strict": true,
     "noImplicitOverride": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
     "forceConsistentCasingInFileNames": true,
-    "importsNotUsedAsValues": "error",
-    "ignoreDeprecations": "5.0",
```
axelboc, line 25: I linked to an article that discussed this, I think: https://johnnyreilly.com/typescript-5-importsnotusedasvalues-error-eslint-consistent-type-imports -- it's ESLint's job

axelboc:
Superseded by #1730

PR #1716, opened 2024-10-30
Make h5web compatible with newer node versions.

NAThompson:


axelboc:
Hi @NAThompson, this is expected. The development environment is pinned to Node 20.x for stability. I recommend using a Node version manager like [Volta](https://docs.volta.sh/guide/getting-started) to switch version easily.

That being said, we might move to Node 22 soon since it is now the LTS version. More info on [Node versioning here](https://nodejs.org/en/about/previous-releases).

PR #1726, opened 2025-01-10
Replace Galex ESLint config

axelboc:
[Galex](https://github.com/ljosberinn/eslint-config-galex) is no longer maintained, unfortunately, notably because of the transition to ESLint 9's [flat config](https://eslint.org/docs/latest/use/configure/migration-guide) format. Being dependent on it was starting to cause issues, notably with TypeScript.

Having not found a suitable replacement, I've decided to write my own config, in the hope that we can later use it in other projects at the ESRF.

- The new config relies on **ESLint v9** and the new **flat config** format — basically one big array of config objects that looks like this:
  ```js
  [
    {
      name: 'h5web/defaults/rules/<plugin>',
      plugins: { 'my-plugin': myPlugin },
      rules: {
        'my-plugin/my-rule': 'error',
      }
    },
    // ...
  }
  ```
- The set-up remains otherwise the same: one **base config file** at the root of the monorepo imported by an ESLint config entrypoint (now called `eslint.config.js` instead of `.eslintrc.cjs`) **inside each project** + at the root.
- Like Galex, the new config **detects** the dependencies actually installed inside each project (React, TypeScript, etc.) and enables only the relevant plugins and rules.
- Here is what's happening in terms of [**plugins**](https://github.com/ljosberinn/eslint-config-galex/blob/master/package.json#L92):
  - `eslint-plugin-tailwindcss` and `@next/eslint-plugin-next` are dropped, obviously
  - `eslint-plugin-etc` is dropped as it doesn't support ESLint 9 and is not actively maintained
  - `eslint-plugin-sonarjs` is dropped because most of the rules are either very opinionated/stylistic or duplicate core ESLint rules and rules from other plugins (like the new regexp plugin below)
  - `eslint-config-prettier` is dropped because I prefer to be careful not to enable stylistic rules that may conflict with Prettier myself (this is made a lot easier by the fact that stylistic rules are no longer part of the core ESLint package)
  - `eslint-plugin-jest` and related plugins are replaced with `eslint-plugin-vitest`
  - [`eslint-plugin-regexp`](https://ota-meshi.github.io/eslint-plugin-regexp/rules/) is added to properly lint regular expressions
  - [`@stylistic/eslint-plugin-js`](https://eslint.style/packages/js) is added for one specific stylistic rule not enforced by Prettier: [`spaced-comment`](https://eslint.style/rules/js/spaced-comment)
- The new config supports linting **JS projects** (i.e. without TypeScript) but it does make the assumption that all `.js` files use **ESM** (i.e. that `package.json` has `"type": "module"`).
- I decided to not mix and match rules from different plugins within each flat config object. Instead, for each plugin, I always have one config object that enables the plugin on as many files as possible and with as many rules as possible, and then additional config objects that disable some of the plugin's rules on specific files (e.g. on TS files, on test files, etc.)

As you'll see with the many changes in the source code, the new config is a little stricter. I'll try to explain what's going on a bit more with some comments.


packages/lib/src/toolbar/MeasuredControl.tsx
```diff
@@ -20,7 +20,9 @@ function MeasuredControl(props: Props) {
         }
       }}
     >
-      {({ measureRef }) => (
+      {(
+        { measureRef }, // eslint-disable-line @typescript-eslint/unbound-method
```
axelboc, line 24: This is because the library declares `measureRef` as a method:

```ts
measureRef(/* params */): void;

// instead of:
measureRef: (/* params */) => void;
```

packages/lib/src/toolbar/controls/Selector/Selector.tsx
```diff
@@ -77,7 +77,7 @@ function Selector<T extends string>(props: Props<T>) {
         aria-labelledby={`${label ? labelId : ''} ${currentOptionId}`}
         aria-haspopup="listbox"
         aria-expanded={isOpen}
-        aria-controls={(isOpen && context.floatingId) || undefined}
```
axelboc, line 80: This comes from `unicorn/consistent-destructuring`, which detects property access on a variable that has already been partially destructured. Unfortunately, I had to later disable it because it would give a lot of false positives with type guards like `isDataset(entity) && entity.chunks`

packages/lib/src/toolbar/controls/ToggleGroup.tsx
axelboc, line 1: Moved to `ToggleGroup` folder and split into multiple files.

packages/lib/src/vis/utils.ts
```diff
@@ -272,13 +272,13 @@ export function getTickFormatter(
...
   availableSize: number,
   scaleType: ScaleType,
-): (val: number | { valueOf(): number }) => string {
+): (val: number | { valueOf: () => number }) => string {
   if (scaleType !== ScaleType.Log) {
     return formatTick;
   }
 
   // If available size allows for all log ticks to be rendered without overlap, use default formatter
-  const [min, max] = domain[0] > 0 ? domain : [-domain[1], -[domain[0]]];
```
axelboc, line 281: Pretty sure this was a bug, but I need to double check.

tsconfig.json
```diff
@@ -21,10 +21,8 @@
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
     "forceConsistentCasingInFileNames": true,
-    "importsNotUsedAsValues": "error",
-    "ignoreDeprecations": "5.0",
```
axelboc, line 25: I'll enforce `import { type Foo } from ...` in my next PR.

tsconfig.json
```diff
@@ -21,10 +21,8 @@
...
     "useDefineForClassFields": true,
     "useUnknownInCatchVariables": true
   },
-  "include": ["*", "cypress"]
+  "include": ["*"]
```
axelboc, line 27: I'll add the `lint:root:tsc` script I mentioned in a future PR.

eslint.config.base.js
```diff
@@ -0,0 +1,1266 @@
...
+    react: withReact,
+    storybook: withStorybook,
+    vitest: withVitest,
+    cypress: withCypress,
+    reactTestingLibrary: withReactTestingLibrary,
+    cypressTestingLibrary: withCypressTestingLibrary,
+  } = { ...DEFAULT_OPTS, ...opts };
+
+  return tseslint.config(
+    _.compact([
```
axelboc, line 65: lodash's `compact` utility allows passing `withTypeScript && { ... }` into the array instead of `...(withTypeScript ? [{ ... }] : [])`.

eslint.config.base.js
```diff
@@ -0,0 +1,1266 @@
...
+         * ESLint core rules: https://eslint.org/docs/latest/rules/
+         */
+        name: 'h5web/defaults/rules/core',
+        rules: {
+          'accessor-pairs': 'error',
+          'array-callback-return': [
+            'error',
+            { checkForEach: true }, // enforce no return in `forEach`
+          ],
+          // 'arrow-body-style': 'off', // use of curly braces in arrow functions should first be dictated by readabillity
```
axelboc, line 129: I declare every single rule explicitly so there are no surprises. I keep those that are disabled commented to not pollute the display when inspecting the config with `pnpm --filter @h5web/shared lint:eslint --inspect-config` or the like.

packages/lib/src/vis/utils.ts
loichuder: Looks like it. Weird it wasn't picked up by TS

loichuder (approved):

packages/app/src/providers/hsds/utils.ts
```diff
@@ -73,15 +73,15 @@ export function convertHsdsShape(shape: HsdsShape): Shape {
...
+  const regex = /H5T_(?:IEEE|STD)_([A-Z])(\d+)(BE|LE)/u;
   const matches = regex.exec(base);
 
   if (!matches) {
     throw new Error(`Unrecognized base ${base}`);
   }
 
   const [, sign, sizeStr, h5tOrderStr] = matches;
-  const size = Number.parseInt(sizeStr, 10);
+  const size = Number.parseInt(sizeStr);
```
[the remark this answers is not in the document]
axelboc: I don't think it was https://github.com/ljosberinn/eslint-config-galex/blob/1792b61043d9f8e5c08be924f6a999318cfc2cf1/src/plugins/eslint-core.ts#L1176

I've been enforcing this myself thinking it was still relevant. Well, there's still the case that, without a radix specified, the `0x` prefix parses the string as hexadecimal. For a user-provided input (e.g. domain bound in the domain widget), this might not be what we're after... but then maybe it is (or at least doesn't matter).

packages/lib/src/vis/utils.ts
axelboc: I dug into this a bit.

1. First observation: when an array that contains a single number is coerced to a number, the result is the number itself (i.e. the number is unwrapped). In other words, `Number([42])` evaluates to `42`, and `-[42]` evaluates to `-42`. This means that `-[domain[0]]` gives the exact same value as `-domain[0]`. Go figure... :shrug: 
2. This means that the fix of removing the extra array is valid in the sense that it maintains the same logic.
3. However, this piece of code is flawed: when `VisCanvas` is used directly from the lib, one can pass a fully negative domain in whichever order, so we have no guarantee that `min < max`. When it happens that `min > max`, then of course `max / min < 1`, which results in showing all tick labels. For instance with `[-10, -15000]`:

    image

axelboc: The existing logic can be simplified to `const [min, max] = domain;`, since the domain is either fully positive of fully negative in log scale (otherwise the D3 scale gives us `[NaN, NaN]`).

But the problem of the inverted domain remains. I feel like we're missing assertions in `VisCanvas` to check that the ordinate and abscissa domains are indeed `[min, max]`. For descending values, the axis is supposed to be flipped with `flip: true`, not by flipping the domain... Am I missing something @loichuder?

axelboc: Merging as is, since the current fix doesn't change the logic. I'll open a PR with another fix, at least to make the logic more robust and fix the edge case.

loichuder: This has been a while since I looked at this...

AFAIR, we implemented the `flip` after the domain logic. This perhaps means that this bit of code should have been removed altogether. It needs some more testing.

axelboc: For reference, the `-[domain[0]]` issue came straight from the initial implementation of the log ticks display improvement in #229

loichuder: Good to know, thanks for digging this

PR #1732, opened 2025-01-15
Switch to inline type imports

axelboc:
The idea is to avoid duplicate imports (one for types and one for values). Having the `type` keyword right before each import is also clearer in my opinion.

Of course, you win some, you lose some: when importing a lot of types from the same module, we get a lot of `type` keywords, which can also lead to Prettier wrapping imports onto separate lines, which is why slightly more lines are added than removed in the diff.


.vscode/settings.json
```diff
@@ -16,6 +16,7 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "typescript.preferences.importModuleSpecifier": "relative",
+  "typescript.preferences.preferTypeOnlyAutoImports": true,
```
axelboc, line 19: Oversight: I had this in my user settings instead of in the project's settings.

eslint.config.base.js
axelboc, line 1: Just a comment to surface the ESLint config changes that make all this happen.

loichuder (approved):

PR #1736, opened 2025-01-21
Merge signed and unsigned integer types

axelboc:
The goal is to split `NumericType` into two dtypes: `IntegerType` and `FloatType`, and to replace `DTypeClass.Unsigned` with a `signed` boolean in `IntegerType`.

I had first planned to introduce this change as part of adding support for big integers (cf. https://github.com/silx-kit/h5web/issues/1679#issuecomment-2333937430). At the time I wanted to introduce a new `DType` called `BigIntegerType` and it didn't really make sense to have this dtype together with `NumericType`... I'm now going back on this solution for reasons I'll explain soon, but I find that the `NumericType` refactoring still makes sense. `IntegerType`/`FloatType` brings us closer to the HDF5 types, and representing the sign as a boolean avoids relying on the `DTypeClass` enum's value too much in the metadata viewer (i.e. "Integer (unsigned)").

### DIFF
--- a/packages/app/src/metadata-viewer/utils.ts
+++ b/packages/app/src/metadata-viewer/utils.ts
@@ -1,5 +1,6 @@
 import {
   isH5WebComplex,
+  isIntegerType,
   isNumericType,
   isScalarShape,
 } from '@h5web/shared/guards';
@@ -29,7 +30,15 @@ export function renderShape(shape: Shape): string {
 export function renderType(type: DType): string {
   if (isNumericType(type)) {
     const { endianness, size } = type;
-    return `${type.class}, ${size}-bit${endianness ? `, ${endianness}` : ''}`;
+
+    const endiannessStr = endianness ? `, ${endianness}` : '';
+    const signStr = isIntegerType(type)
+      ? type.signed
+        ? ' (signed)'
+        : ' (unsigned)'
+      : '';
+
+    return `${type.class}${signStr}, ${size}-bit${endiannessStr}`;
   }
 
   if (type.class === DTypeClass.String) {
--- a/packages/shared/src/hdf5-models.ts
+++ b/packages/shared/src/hdf5-models.ts
@@ -102,7 +102,6 @@ export type ScalarShape = [];
 export enum DTypeClass {
   Bool = 'Boolean',
   Integer = 'Integer',
-  Unsigned = 'Integer (unsigned)',
   Float = 'Float',
   Complex = 'Complex',
   String = 'String',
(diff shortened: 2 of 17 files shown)


packages/app/src/metadata-viewer/utils.ts
```diff
@@ -29,7 +30,15 @@ export function renderShape(shape: Shape): string {
...
-    return `${type.class}, ${size}-bit${endianness ? `, ${endianness}` : ''}`;
+
+    const endiannessStr = endianness ? `, ${endianness}` : '';
+    const signStr = isIntegerType(type)
+      ? type.signed
+        ? ' (signed)'
+        : ' (unsigned)'
+      : '';
+
+    return `${type.class}${signStr}, ${size}-bit${endiannessStr}`;
```
axelboc, line 41: A bit more display logic here, but less logic when creating the dtypes in the providers (e.g. removal of the `intOrUint` utility) and in other places (e.g. `type.class === DTypeClass.Signed || type.class === DTypeClass.Unsigned`). You win some you lose some — it's really more of a conceptual change.

packages/shared/src/hdf5-models.ts
```diff
@@ -121,6 +120,7 @@ export type Endianness = (typeof H5T_TO_ENDIANNESS)[H5T_ORDER];
 export type CharSet = (typeof H5T_TO_CHAR_SET)[H5T_CSET];
 export type StrPad = (typeof H5T_TO_STR_PAD)[H5T_STR];
 
+export type NumericType = IntegerType | FloatType;
```
axelboc, line 123: We still keep a `NumericType` as that's still useful throughout the codebase.

loichuder:
I am not awfully convinced by this. Personally, I have always valued descriptive values over flags.

Also
> IntegerType/FloatType brings us closer to the HDF5 types

Is that really true ? As far as I can see, unsigned and signed integers are separate datatypes: https://support.hdfgroup.org/documentation/hdf5/latest/group___p_d_t_s_t_d.html

So unless this lays ground for another future refactoring, I need a little more to go along :smile:

axelboc:
I was referring to the `H5T_class_t` enum: https://support.hdfgroup.org/documentation/hdf5/latest/_h5_tpublic_8h.html#a071841985f647f69516dbe77d93167f2

axelboc:
So the idea is to bring `DTypeClass` closer to `H5T_class_t`. The only two values that don't have equivalent are `DTypeClass.Bool` and `DTypeClass.Complex`, which are compound types with very specific, well-defined fields.

Another argument is that the `class` property is the discriminant property of the [discriminated union type](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions) `DType`. That one of the interfaces in that union has `class: DTypeClass.Integer | DTypeClass.Unsigned | DTypeClass.Float` is a bit nonsensical. (I'd like to split `ArrayType` into two, `ArrayType` and `VLenType`, for the same reason.)

PR #1737, opened 2025-01-21
Make `ArrayValue` type distributive

axelboc:
As a reminder, providers are allowed to return boolean datasets as `boolean[]` (HSDS only) or as `number[]` (h5wasm  #1692 and h5grove  #1709).

The goal here is for `ArrayValue<BooleanType>` to evaluate to `number[] | boolean[]` instead of `(number | boolean)[]` (which was quite wrong). This doesn't change much but it will matter more when introducing `bigint`...

While I'm at it, I give the `Primitive` and `ArrayValue` generics default values, and I use `ArrayValue` in a bunch of places to replace `unknown[] | TypedArray`. This will save me later from adding `BigInt64Array` and `BigUint64Array` to that type everywhere...

### DIFF
--- a/packages/app/src/vis-packs/core/utils.ts
+++ b/packages/app/src/vis-packs/core/utils.ts
@@ -11,7 +11,7 @@ import {
   type NumArray,
 } from '@h5web/shared/vis-models';
 import { createArrayFromView } from '@h5web/shared/vis-utils';
-import ndarray, { type NdArray, type TypedArray } from 'ndarray';
+import ndarray, { type NdArray } from 'ndarray';
 
 import { type DimensionMapping } from '../../dimension-mapper/models';
 import { isAxis } from '../../dimension-mapper/utils';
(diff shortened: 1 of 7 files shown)


packages/app/src/vis-packs/core/utils.ts
```diff
@@ -95,12 +93,16 @@ export function getImageInteractions(keepRatio: boolean): InteractionInfo[] {
   return keepRatio ? BASE_INTERACTIONS : INTERACTIONS_WITH_AXIAL_ZOOM;
 }
 
+function isBoolArray(val: ArrayValue<NumericLikeType>): val is boolean[] {
+  return Array.isArray(val) && typeof val[0] === 'boolean';
+}
+
 export function toNumArray(arr: ArrayValue<NumericLikeType>): NumArray {
-  if (typeof arr[0] === 'boolean') {
```
axelboc, line 99: TypeScript still thinks that `arr` is `number[] | boolean[]` both inside and after this block, so I'm implementing a proper type guard. Since it's so specific (the parameter has type `ArrayValue<NumericLikeType>`), I chose to keep the type guard close by.

packages/shared/src/guards.ts
```diff
@@ -142,14 +142,6 @@ export function isTypedArray(val: unknown): val is TypedArray {
   );
 }
 
-export function assertArrayOrTypedArray(
```
axelboc, line 145: This should be really be called `assertValueArray`, but since it's used only below, I chose to inline it.

packages/shared/src/hdf5-models.ts
```diff
@@ -207,21 +207,19 @@ export interface UnknownType {
 /* ----------------- */
 /* ----- VALUE ----- */
 
-export type Primitive<T extends DType> = T extends NumericType | EnumType
-  ? number
+export type Primitive<T extends DType = DType> = T extends NumericLikeType
+  ? number | (T extends BooleanType ? boolean : never) // let providers choose how to return booleans
```
axelboc, line 211: Just a small refactoring to make use of `NumericLikeType` and avoid duplicating `number`.

loichuder (approved):
Looks good!

PR #1738, opened 2025-01-21
Fix support for enum/bool auxiliary signals in NX Line visualization

axelboc:
`assertNumericLikeNxData` was asserting for `NxData<NumericType>` instead of `NxData<NumericLikeType>` ... but was correctly calling `assertNumericLikeType` internally! So `MappedLineVis` thought it could only receive numeric (nd)arrays (`number[] | TypedArray`) when in fact, at runtime, it could receive `boolean[]`.

No errors were thrown at runtime because JS automatically coerces booleans to numbers in all mathematical operations (e.g. `Math.min(false, 2) => 0`). This typing bug would have been more problematic with `bigint[]`, though.

I say "Fix support", because the fix still has invisible runtime implications: I now convert boolean auxiliary arrays to number arrays in `MappedLineVis` with a new hook called `useToNumArrays`, which avoids type coercion further down the line (e.g. in `useDomains`).

### DIFF
--- a/packages/app/src/vis-packs/core/line/MappedLineVis.tsx
+++ b/packages/app/src/vis-packs/core/line/MappedLineVis.tsx
@@ -8,9 +8,9 @@ import {
 } from '@h5web/lib';
 import {
   type ArrayShape,
-  type ArrayValue,
   type Dataset,
   type NumericLikeType,
+  type Value,
 } from '@h5web/shared/hdf5-models';
 import { type AxisMapping } from '@h5web/shared/nexus-models';
 import { type NumArray } from '@h5web/shared/vis-models';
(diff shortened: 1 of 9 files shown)


packages/app/src/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -24,18 +24,19 @@ import {
...
 interface Props {
-  dataset?: Dataset<ArrayShape, NumericLikeType>;
-  value: ArrayValue<NumericLikeType>;
+  dataset: Dataset<ArrayShape, NumericLikeType>;
+  value: Value<Props['dataset']>;
   valueLabel?: string;
   errors?: NumArray;
   auxLabels?: string[];
-  auxValues?: NumArray[];
+  auxValues?: Value<Props['dataset']>[];
```
loichuder, line 39: I wonder if it might not be clearer to introduce a new type alias here to not reference `dataset` explicitely.

It seems weird since `auxValues` are not stored in `dataset`.

loichuder (approved):

packages/app/src/vis-packs/core/line/MappedLineVis.tsx
axelboc: Yeah, wasn't sure that it was clearer. Two goals here:

1. The types of the `value` and each item in `auxValues` should match (if an `NXdata` group has a boolean signal, then it should be allowed to have boolean auxiliaries). This requires changing `auxValues?: NumArray[]` at least to `ArrayValue<NumericLikeType>` to match the `value` prop.
2. The dtype passed to `Dataset<ArrayShape, DType>` should match the dtype passed to `ArrayValue<Dtype>`. In other words, I'd like to avoid the risk of the two becoming out of sync — e.g. `Dataset<ArrayShape, NumericLikeType>` and `ArrayValue<NumericType>`. This is why I refer to the type of the `dataset` prop with `Value<Prop['dataset'].

Perhaps the second goal is not critical, since TypeScript would show an error in the container if the dtype of `dataset` and `value`/`auxValues` did not match. So should I stick with `ArrayValue<NumericType>`?

axelboc: Opening a PR to propose a normalisation of the mapped components' props.

PR #1739, opened 2025-01-23
Normalise prop types of mapped components with `ArrayValue`

axelboc:
From https://github.com/silx-kit/h5web/pull/1738#discussion_r1924836346

The `value`, `auxValues`, `errors`, `auxErrors` and `axisValues` props of mapped components (`MappedLineVis`, `MappedRgbVis`, etc.) were typed in a few different, inconsistent ways:

 - with `NumArray` (= `number[] | TypedArray`)
 - with `ArrayValue<DType>` (where `DType` matches the one of the `dataset` prop — e.g. `Dataset<ArrayShape, NumericType>` => `ArrayValue<NumericType>`, but it works with any other dtype or dtype union like `NumericLikeType`)
 - with `Value<Props['dataset']>` to avoid duplicating the dtype.

While the third way is the most robust and DRY, it's not the most readable since it adds an extra level of misdirection. And while the first way is the most explicit, it becomes less readable and/or more error-prone with other dtypes, like `NumericLikeType`, `PrintableType` or `ComplexType` (especially once we start supporting big ints).

So I'm going for the middle ground with `ArrayValue`.

---

While I'm at it:

- I remove `NumArrayDataset` completely, which was an alias for `Dataset<ArrayShape, NumericType>`; it adds a couple of imports in a few files, but it's less confusing, especially since we don't have aliases for datasets with other dtypes, like `Dataset<ArrayShape, NumericLikeType>`.
- I rename `Primitive` to `ScalarValue`. This makes it clearer that `ScalarValue` is the value of datasets with `ScalarShape`.

### DIFF
--- a/packages/app/src/providers/mock/mock-api.ts
+++ b/packages/app/src/providers/mock/mock-api.ts
@@ -5,11 +5,12 @@ import {
 } from '@h5web/shared/guards';
 import {
   type ArrayShape,
+  type ArrayValue,
   type AttributeValues,
   type Dataset,
   type Entity,
   type GroupWithChildren,
-  type NumArrayDataset,
+  type NumericType,
   type ProvidedEntity,
   type Value,
 } from '@h5web/shared/hdf5-models';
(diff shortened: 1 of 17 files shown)


loichuder (approved):
Not bad! Yeah I think it improves things

PR #1745, opened 2025-01-29
Allow providers to return bigints

axelboc:
The mock and h5wasm providers now return `bigint` values for int-64 datasets (including compounds with int-64 fields). They no longer have to cast int-64 integers to JS numbers (i.e. float 64), which means that the `bigint` values reach the _Scalar_, _Matrix_ and _Compound_ visualizations without any loss of precision.

This addresses most of #1679. (I'll close it once `H5GroveProvider` also returns bigints.)

The preparatory work I did in #1736, #1737, #1738, #1739 and #1743 bore fruit. I didn't end up needing a separate `BigIntegerType` or even the "generic solution" explained in https://github.com/silx-kit/h5web/issues/1679#issuecomment-2333937430. (This generic solution was flawed anyway: allowing providers to declare which value type they want to return for which dataset shape/type is useless, since the viewer wants to know which value types _any_ provider may return, so either way we end up with a type union.)

#### Summary of the changes

- Mock datasets with bigint values (e.g. `1n`) are now "guessed" to have type `IntegerType` instead of `UnknownType`
- `H5WasmApi` no longer converts bigints to numbers
- `ScalarValue<NumericType>` is now `number | bigint`
- `ArrayValue<NumericType>` is now `TypedArray | BigIntTypedArray | number[] | bigint[]`
- The _Scalar_, _Matrix_ and _Compound_ visualizations now support bigints — they just convert them to string, except when scientific notation is selected (in which case they cast them to numbers and pass them to `d3-format`).
- The mapped components of the WebGL visualizations (which don't support bigints) are now responsible for converting bigint signals (including auxiliaries) to numbers — this is done in `useToNumArray`, which was already responsible for converting booleans to numbers.
- Since axis and error arrays are now also allowed to contain bigints (because of `ArrayValue<NumericType>`), and since D3 and most of our code in `@h5web/lib` doesn't support bigints, those arrays also have to be converted inside mapped components with `useToNumArray` and `useToNumArrays`.

#### How I got everything to work

What mattered the most to not break the entire codebase was to ensure that `NumArray` didn't go from `number[]| TypedArray` to `number[] | bigint[] | TypedArray | BigIntTypedArray`, as `NumArray` is used throughout `@h5web/lib`.

The passage between `Value<Dataset<Shape, DType>>` and `NumArray` happens in the mapped components, so declaring the props of those components with `ArrayValue<NumericArray>` instead of `NumArray` in #1739 was key.

Another trick that makes the implementation a lot simpler in this PR is to ensure that enum and boolean datasets cannot be provided as bigints but only as numbers (even though technically, `EnumType` and `BoolType` can have a 64-bit `IntegerType` as base type). This is enforced by the `ScalarValue` and `ArrayValue` types and by the corresponding assertion functions. 

#### Screenshots

image

image


packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -28,7 +28,15 @@ export const useToNumArray = createMemo(toNumArray);
 
 export function useToNumArrays(
   arrays: ArrayValue<NumericLikeType>[],
-): NumArray[] {
+): NumArray[];
+
+export function useToNumArrays(
+  arrays: (ArrayValue<NumericLikeType> | undefined)[],
```
axelboc, line 34: In mapped components, the `auxErrors` and `axisValues` array props are allowed to have `undefined` values.

packages/app/src/vis-packs/core/matrix/utils.ts
```diff
@@ -23,7 +24,7 @@ import { format } from 'd3-format';
 
 export function createNumericFormatter(
   notation: Notation,
-): (val: ScalarValue<NumericType>) => string {
+): (val: number) => string {
```
axelboc, line 27: I specifically don't want `number | bigint` here.

packages/app/src/vis-packs/core/utils.ts
```diff
@@ -93,11 +93,33 @@ export function getImageInteractions(keepRatio: boolean): InteractionInfo[] {
...
+export function toNumArray<T extends ArrayValue<NumericLikeType> | undefined>(
+  arr: T,
+): T extends ArrayValue<NumericLikeType> ? NumArray : undefined;
+
+export function toNumArray(
+  arr: ArrayValue<NumericLikeType> | undefined,
+): NumArray | undefined {
+  if (!arr) {
+    return undefined;
+  }
```
axelboc, line 117: Adding an `undefined` case here for the same reason as in `useToNumArrays`, and to allow `useToNumArray(errors)` (as the `errors` prop can also be `undefined` in mapped components.)

packages/app/src/vis-packs/core/utils.ts
```diff
@@ -93,11 +93,33 @@ export function getImageInteractions(keepRatio: boolean): InteractionInfo[] {
...
+
+  if (isBigIntTypeArray(arr)) {
+    return Float64Array.from(arr, Number); // cast to float 64
```
axelboc, line 120: Data arrays that are passed to WebGL get converted to `Float32Array` soon after, unfortunately. I don't think it's worth optimising, though.

packages/shared/src/guards.ts
```diff
@@ -412,11 +423,13 @@ export function isComplexValue(
   return type.class === DTypeClass.Complex;
 }
 
-function assertScalarValue(
```
axelboc, line 415: Swapping the two parameters to match `assertArrayValue` and have the `value` being asserted first.

packages/app/src/vis-packs/core/matrix/utils.ts
```diff
@@ -34,6 +35,22 @@ export function createNumericFormatter(
...
+): (val: ScalarValue<NumericType>) => string {
+  const formatter = createNumericFormatter(notation);
+
+  switch (notation) {
+    case Notation.Scientific:
+      return (val) => formatter(Number(val));
+
+    default:
+      return (val) => {
+        return typeof val === 'bigint' ? val.toString() : formatter(val);
```
axelboc, line 49: I've pushed a bug fix here. Int-64 values provided as `number` were being cast to string instead of being formatted with D3. I'm not 100% which is the result we want, actually - e.g. for fixed-point notation, do we want `1.000` or `1`? I assumed `1`, since decimals don't matter, so maybe I should revert and keep the `toString()`?

loichuder: Yeah, it does not make much sense to do more than converting to strings for integers. `toString` should be enough.

loichuder (approved):
:clap: :muscle:

packages/app/src/vis-packs/core/matrix/utils.ts
axelboc: Refactoring done in https://github.com/silx-kit/h5web/pull/1745/commits/6852665dbbcbe77084896fa6c12cfc57f30e9bf8

I've split `createNumericFormatter` into two:

- `createFloatFormatter` which works as before
- `createIntegerFormatter`:
  - with "auto" and "fixed-point" notation: casts numbers/bigints to string with `toString()`
  - with "scientific" notation: casts bigints to numbers with `Number`, and then formats numbers with D3 (same format as before: `'.3e'`)

PR #1747, opened 2025-01-30
Fetch int-64 values from h5grove as binary to maintain precision

### LINKED ISSUE #1679: 64-bit integers not shown correctly in h5web display

axelboc:
Follows #1745 and closes #1679


packages/app/src/providers/utils.ts
```diff
@@ -33,7 +36,7 @@ export function typedArrayFromDType(
         return Int16Array;
       case 32:
         return Int32Array;
-      case 64: // No support for 64-bit integer values in JS
+      case 64:
         return undefined;
```
loichuder, line 40: Why not returning `BigInt64Array` here ?

axelboc: Sorry, should have explained: for the same reason I have a separate type union, a separate type guard, etc. — when both kinds of typed arrays are mixed together in a union, TypeScript goes wild:

- You can't invoke the `Big(U)Int64Array` constructors with `number[]`, only with `bigint[]` (so `new IntOrBigIntTypedArray(arr as number [] | bigint[])` doesn't type-check)
- You can't loop through the array, since `number[]` and `bigint[]` aren't compatible.

Having a single function worked fine in `h5grove-api` since we invoke the typed array constructor with a buffer. However, it failed in `useIgnoreFillValue`, where we pass a `number`:

```ts
  const DTypedArray = typedArrayFromDType(dataset.type);

  // Cast fillValue in the type of the dataset values to be able to use `===` for the comparison
  const fillValue =
    DTypedArray && typeof wrappedFillValue[0] === 'number'
      ? new DTypedArray(wrappedFillValue as number[])[0]
      : undefined;
```

Arguably, perhaps `useIgnoreValue` should be reworked to support bigints as well... Seemed like a can of worms, though, so I went with a compromise.

loichuder: Yeah, ok I see.

loichuder (approved):

PR #1750, opened 2025-01-30
Add mock data and snapshot tests for `_FillValue`

axelboc:
- Add a couple of mock datasets to test and demonstrate the use of the `_FillValue` attribute (and to double check that the fill value is used in the domain computation as expected).

|`_FillValue=400` | `_FillValue=400` | 
|---|---|
| image|image|

- Add snapshot tests for those two datasets to prevent future regressions of the `ignoreValue` feature in the _Line_ and _Heatmap_ visualizations.
- Introduce a type alias for the `ignoreValue` function.
- Tweak the implementation of `useIgnoreFillValue` slightly to memoise a few more computations.
- Fix the dynamic domain computation in some stories by passing `ignoreValue` (and `errors` for that matter) to `useDomain`.


apps/storybook/src/LineVis.stories.tsx
```diff
@@ -65,15 +65,34 @@ type Story = StoryObj<typeof meta>;
...
+        getDomain(
+          dataArray,
+          scaleType,
+          showErrors ? errorsArray : undefined,
+          ignoreValue,
+        ),
+        ...getDomains(
+          auxArrays,
+          scaleType,
+          showErrors ? auxErrorsArrays : undefined,
```
axelboc, line 94: We don't pass `ignoreValue` here, as the expected behaviour for auxiliary signals is not clear. Something to discuss in the future if the user need arises.

packages/lib/src/vis/line/errorBarsGeometry.ts
```diff
@@ -31,10 +31,9 @@ class ErrorBarsGeometry extends H5WebGeometry<'position', Params> {
     } = this.params!;
 
     const value = ordinates[index];
-    const isIgnored = ignoreValue ? ignoreValue(value) : false;
     const bufferIndex = index * 2;
 
-    if (isIgnored) {
```
loichuder, line 37: I liked the intermediate variable: brain computing `.?` can be tricky sometime :sweat_smile:

loichuder (approved):

packages/lib/src/vis/line/errorBarsGeometry.ts
axelboc: Haha, I personally found the ternary more difficult to read. :sweat_smile: With `?.`, the fact that `ignoreValue` is optional doesn't distract as much from the main condition, which is whether or not the value is to be ignored.

PR #1752, opened 2025-02-03
Add line event handlers to `DataCurve`

### LINKED ISSUE #1720: onLineEnter() Function Param to Highlight DataCurve

No.

For the DataCurve components to able to take some function parameter such as `onLineEnter()` to highlight a DataCurve line by increasing its thickness.

This may be more flexible and fit better with existing patterns as 2 separate functions.

1: Similar to the existing `onDataPointEnter` function param, an `onLineEnter` function param to be triggered when a DataCurve line is entered (which would then make sense to also have the corresponding `onLineClick` and `onLineLeave` params).

2: To "highlight" a DataCurve, perhaps a new css custom property i.e. `--h5w-line-thickness` or an integer param on DataCurve i.e. `thickness="2"` (which would be the same pattern as the existing `glyphSize`).

### Additional context

Particularly for multiline plots, it would be very helpful for visual clarity to have some way to highlight a "current" line.

I currently "solve" this issue by increasing the `glyphSize` on `onDataPointEnter`, however, it takes quites a lot of motor dexterity when attempting to hover over the exact pixel of data points.

Being able to visually highlight a particular line (especially when viewing plots with more than 4 lines) would greatly increase the motor accessibility and visual accessibility for all users.

axelboc:
Fix #1720

`[DataCurve]` Add props `onLineClick`, `onLineEnter` and `onLineLeave`

The three new line event handlers all receive the `index` of the line segment that was interacted with, just like the three existing data point event handlers.

I didn't change the [_Interactive_ story](https://h5web-docs.panosc.eu/?path=/story/building-blocks-datacurve--interactive) since the actions add-on already logs the new events by default:

image

### DIFF
--- a/packages/lib/src/vis/line/DataCurve.tsx
+++ b/packages/lib/src/vis/line/DataCurve.tsx
@@ -1,11 +1,11 @@
 import { type IgnoreValue, type NumArray } from '@h5web/shared/vis-models';
 import { type ThreeEvent } from '@react-three/fiber';
-import { useCallback } from 'react';
 
 import ErrorBars from './ErrorBars';
 import Glyphs from './Glyphs';
 import Line from './Line';
 import { CurveType, GlyphType } from './models';
+import { useEventHandler } from './utils';
 
 interface Props {
   abscissas: NumArray;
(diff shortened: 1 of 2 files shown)


chanmSC:
Spectacular! Thank you for this feature!

Issue #1742, opened 2025-01-27
`MatrixVis`: customisable fixed-point precision


axelboc:
### Is your feature request related to a problem?

Received by email:

> It would be nice if the number of decimals were adjustable, currently I seem to get only 3 decimals in fixed-point 'matrix' view, so a number is shown as "3.141" but no more digits are visible.
>
> We work with frequency metrology, at least 12 decimal digits should ideally be available.

![Image](https://github.com/user-attachments/assets/1092fd59-1895-4af7-aabe-1a39c6ec915b)

### Requested solution or feature

- Add a toolbar control to the Matrix visualisation to allow users to customise the number of decimal digits in fixed-point notation.
- Consider using this same precision in scientific notation (cf. `createNumericFormatter`) and for complex numbers (cf. `createMatrixComplexFormatter`).

loichuder:
I could not find the link but I am sure that we discussed this before in a previous issue/MR 😄

axelboc:
Oh I found it: #1646

Issue #1646, opened 2024-05-21
[Matrix] The vis does not allow to distinguish numeric values that are very close


loichuder:
### Is your feature request related to a problem?

I happened to find a dataset (link below) with only 2 values that are extremely close numerically but yet different. When looking at the Line vis (see screenshots below), I see that the two values are different (the line is not horizontal). 

But the numeric precision of the tooltip and of the Matrix vis does not allow me to see this difference.

### Requested solution or feature

Not sure. Perhaps a way of specifying the number of digits in fixed-point notation as suggested earlier in #1175 ?

### Alternatives you've considered

Export as CSV and reading the values there but that is really convoluted.

#### Dataset: 
 `/1.1/instrument/positioners/obpitch` in [strain.hdf5](https://gitlab.esrf.fr/XRD/darfix/-/blob/main/src/orangecontrib/darfix/tutorials/hdf5_dataset/strain.hdf5?ref_type=heads)

#### Screenshots

Line vis: 

![image](https://github.com/silx-kit/h5web/assets/42204205/d8cabf19-c955-4516-b2fe-ac202b2f54b0)

Matrix vis: 

![image](https://github.com/silx-kit/h5web/assets/42204205/71a04cf4-ad73-4c67-be17-54880475e0c6)

axelboc:
I'd be okay with adding a "Precision" field in the toolbar. Only problem I see is that we would have to pick a default precision that works for all notations out of the box. Currently we use the following numeric formats:

- **Fixed-point:** `.3f` (or `.2f` for complex members)
- **Scientific:** `.3e` (or `.3~e` for complex members)
- **Auto:** `.5~g` (or `.3~g` for complex members)

axelboc:
See also #1742, which I've just closed as duplicate.

Here is an overview of our current notations:

| Raw number | _Auto_ (`.5~g`) | _Scientific_ (`.3e`) | _Fixed-point_ (`.3f`) |
|---|---|---|---|
| **`11111111111`** | `"1.1111e+10"` | `"1.111e+10"` | `"11111111111.000"` |
| **`999999`** | `"1e+6"` | `"1.000e+6"` | `"999999.000"` |
| **`99999.9`** | `"1e+5"` | `"1.000e+5"` | `"99999.900"` |
| **`99999`** | `"99999"` | `"1.000e+5"` | `"99999.000"` |
| **`9999.99`** | `"10000"` | `"1.000e+4"` | `"9999.990"` |
[...]

PR #1753, opened 2025-02-04
Replace fixed-point with exact notation in Matrix and Compound vis

axelboc:
As discussed in https://github.com/silx-kit/h5web/issues/1646#issuecomment-2631205677.

Keeping the issue open until I address the formatting of numbers in the visualizations' tooltips.

This should also address the user need of #1742 in a more usable way than the solution proposed of adding a control to the toolbar to increase the fixed-point precision.

While I'm at it, I bring the formatting of complex numbers in line with the formatting of other numbers. We were so far using formatters with less precision to save space in the _Matrix_ vis, but I think consistency should prime.

Here is a more precise list of user-facing changes:

1. `< Scalar >` Display complex numbers with exact precision
2. `< Scalar >` Always show both real and imaginary parts of complex numbers to avoid confusing complex numbers without imaginary parts with real numbers — this applies also when inspecting values of complex attributes in the metadata viewer
3. **`< Matrix, Compound >` Replace _Fixed-point_ notation with _Exact_ notation** to allow inspecting and copying exact values
4. `< Matrix, Compound >` Inherit integer and/or float formatting for complex numbers in _Auto_ and _Scientific_ notations (depending on dtype of real and imaginary parts) — concretely, complex numbers are now displayed with more precision regardless of the notation.

| # | Before | After |
| --- | --- | --- |
| 1 | image | image |
| 2 | image | image |
| 3 | image image | image image |
| 4 | image image | image image |


packages/app/src/vis-packs/core/matrix/utils.ts
```diff
@@ -83,10 +71,13 @@ export function getFormatter(
   }
 
   if (isComplexType(type)) {
-    return createMatrixComplexFormatter(notation);
+    return createComplexFormatter(
+      getFormatter(type.realType, notation),
+      getFormatter(type.imagType, notation),
```
axelboc, line 76: I recurse to inherit the formatting of integers/floats.

packages/shared/src/hdf5-models.ts
```diff
@@ -216,7 +216,7 @@ export interface UnknownType {
 export type ScalarValue<T extends DType = DType> = T extends NumericLikeType
   ?
       | number
-      | (T extends NumericType ? bigint : never) // let providers return bigints
+      | (T extends IntegerType ? bigint : never) // let providers return bigints
```
axelboc, line 219: Unrelated. This is more precise but has no impact on the codebase since we don't have visualizations that accept only integers or only floats.

packages/shared/src/vis-utils.ts
```diff
@@ -64,24 +64,13 @@ export function formatBool(value: ScalarValue<BooleanType>): string {
...
   return (value) => {
     const [real, imag] = value;
-
-    if (imag === 0 && !full) {
-      return formatVal(real);
-    }
-
-    if (real === 0 && !full) {
-      return `${formatVal(imag)} i`;
-    }
```
axelboc, line 81: I'm removing the feature of showing only the real part or only the imaginary part if the other part is equal to 0.

loichuder: Yeah, I don't know why we thought this was a great idea :joy_cat:

PR #1755, opened 2025-02-05
Toggle exact notation in Line vis tooltip with wheel button

axelboc:
Proof of concept for #1646.

While the new _Exact_ notation introduced in #1753 makes sense for the _Matrix_ vis (especially since the _Auto_ notation remains the default), the tooltips of the WebGL-based visualizations are another matter. Showing exact values can quickly add clutter:

image

There's no easy way to know what precision the user expects to see. Finding the smallest difference between two values is of course not an option, so we're basically left with... providing a good default behaviour and letting the user opt in to see more precise values.

In this spirit, here's my attempt at an interaction that consists in clicking the wheel button to toggle "exact" precision on or off for the tooltip:

[Screencast from 2025-02-05 14-40-53.webm](https://github.com/user-attachments/assets/cd6ce54e-02e5-4fa2-9f31-7769c1309dcd)

This is just a draft to request feedback on the interaction — please ignore the implementation for now.

### DIFF
--- a/packages/app/src/vis-packs/core/complex/MappedComplexLineVis.tsx
+++ b/packages/app/src/vis-packs/core/complex/MappedComplexLineVis.tsx
@@ -56,8 +56,14 @@ function MappedComplexLineVis(props: Props) {
   } = props;
 
   const { visType } = config;
-  const { customDomain, yScaleType, xScaleType, curveType, showGrid } =
-    lineConfig;
+  const {
+    customDomain,
+    yScaleType,
+    xScaleType,
+    curveType,
+    showGrid,
+    exactNotation,
+  } = lineConfig;
 
   const numAxisArrays = useToNumArrays(axisValues);
   const [dataArray, ...auxArrays] = useMappedComplexArrays(
@@ -110,6 +116,7 @@ function MappedComplexLineVis(props: Props) {
           label: auxLabels[i],
           array,
         }))}
+        exactNotation={exactNotation}
         testid={dimMapping.toString()}
       />
     </>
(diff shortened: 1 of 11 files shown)


loichuder:
The interaction is fine for me, just not very discoverable. We will have to document it somehow.

Anyway, nobody raised the issue about the exact value in the tooltip (except me) so another sensible alternative is to do nothing.

 But since we have to possibility to do something, even not ideal, that is better than nothing :wink:

t20100:
+1 even if the interaction is not discoverable.

axelboc:
Okay, so I've had another look at the implementation, and I have concerns with:

- the arbitrary middle-click interaction interfering with other interactions (we don't currently allow panning with middle click in `@h5web/lib`, but we might in the future, for instance, or consumers might choose to develop their own custom middle-click interaction)
- the tight coupling with `TooltipMesh`, which makes customising the interaction (button, modifier key, etc.) and persisting the setting between dataset/visualisation challenging.

Here is another approach that consists in simply adding a toggle to the line and heatmap toolbars (and their equivalents for complex numbers):

[Screencast from 2025-04-07 16-34-52.webm](https://github.com/user-attachments/assets/f475a031-b82e-48fc-90ec-d57b499e0f48)

I'm not super psyched about adding a control to the toolbar, but at least we get persistence and discoverability out of the box, and no risk of conflicting interactions.

axelboc:
That's a nice idea as well, though it might be a bit tricky to find a timing that feels reasonable while not being too distracting. There's also the idea of showing the exact value in a status bar of sorts, but this would reduce the space available for the visualization.

As discussed offline, it appears that none of the ideas explored so far are 100% conclusive, so the decision is to remain with the status quo for the time being. I'm going to close this PR, as well as #1646, since we did implement exact notation in the _Matrix_ vis, which was the core request of the issue.

PR #1759, opened 2025-02-17
Export models, type guards and assertions

axelboc:
Consumers using the provider API would benefit from having access to all our models, type guards and assertions. Example inspired from the [Data Portal project](https://gitlab.esrf.fr/icat/data-portal/-/blob/main/packages/h5/src/h5/H5ImageHeatmapViewer.tsx#L54-62):

```ts
const { entitiesStore, valuesStore } = useDataContext();

const entity = entitiesStore.get('/my-dataset'); // ProvidedEntity
assertDataset(entity); // Dataset
assertNonNullShape(entity) // Dataset<ScalarShape | ArrayShape>

// Now it's safe to get the value
const data = valuesStore.get({ dataset: entity }); // unknown

// I can keep going...
assertArrayShape(entity); // Dataset<ArrayShape>
const { shape } = entity; // number[] with at least one element

assertFloatType(entity); // Dataset<ArrayShape, FloatType>
assertDatasetValue(data); // number[] | TypedArray
const dataArray = useMemo(() => ndarray(data, shape), [entity, data]); // NdArray<number[] | TypedArray>
```

### DIFF
--- a/packages/shared/src/guards.ts
+++ b/packages/shared/src/guards.ts
@@ -72,9 +72,12 @@ export function assertNonNull<T>(
   }
 }
 
-function assertNum(val: unknown): asserts val is number {
+export function assertNum(
+  val: unknown,
+  message = 'Expected number',
+): asserts val is number {
   if (typeof val !== 'number') {
-    throw new TypeError('Expected number');
+    throw new TypeError(message);
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/shared/src/guards.ts
```diff
@@ -72,9 +72,12 @@ export function assertNonNull<T>(
   }
 }
 
-function assertNum(val: unknown): asserts val is number {
+export function assertNum(
+  val: unknown,
+  message = 'Expected number',
```
axelboc, line 77: I'm adding the ability to pass a custom error `message` to most assertions.

packages/shared/src/guards.ts
```diff
@@ -144,13 +153,25 @@ export function isBigIntTypedArray(val: unknown): val is BigIntTypedArray {
   return val instanceof BigInt64Array || val instanceof BigUint64Array;
 }
 
+export function assertArrayOrTypedArray(
```
axelboc, line 156: I've added some missing guards, and moved some others around a bit.

loichuder (approved):

PR #1763, opened 2025-02-28
Add CSV export to NX Line (and other export improvements)

### LINKED ISSUE #1454: Data mapping not fully reflected in exports
### Describe the bug

The export feature (to CSV, TIFF, etc.) does not take into account potential transposition of datasets (or dataset slices). See https://github.com/silx-kit/vscode-h5web/pull/23#discussion_r1257530624

The `getExportURL` method that data providers may implement does not receive sufficient information to determine whether users have transposed the dataset/slice they want to export. The current signature of the method is:

```
public getExportURL?<D extends Dataset<ArrayShape>>(
    format: ExportFormat,
    dataset: D,
    value: Value<D>
  ): ExportURL;
```

### To Reproduce

1. Go to https://h5web.panosc.eu/h5grove
2. Select dataset `/entry_0000/1_integration/results/I`
3. Select _Matrix_ visualization
4. Export to CSV => 10 rows x 1000 columns
5. Use dimension mapper sidebar to transpose dataset
6. Export to CSV again => still 10 rows x 1000 columns (expected 1000 rows x 10 columns)

Can also be reproduced with the _Heatmap_ visualization and the TIFF export, for instance.

### Expected behaviour

Exports should reflect potential transposition of the datasets (or dataset slices).

### LINKED ISSUE #1619: Export of NeXus Data visualization should contain all data

Right now when viewing an NXdata group that contains a line plot with auxiliary signals the export function only exports the main dataset which is usually not what you would want as you're interested in all the data.

Customize the export to csv feature so that you can select which data should be included.

### LINKED ISSUE #1639: Export Fieldnames in CSV Export for Compound Datatypes

When exporting a compound datatype dataset to csv there are no column headers

Export column headers for compound datatypes

### LINKED ISSUE #1723: Export 1D NxData as CSV

Scientists at ESRF beamlines often need to provide their users with ASCII data instead of HDF5 because the user software does not read HDF5. 

Most of the time, the ASCII file should contain pyFAI integration data (ex: `/entry_0000/1_integration/results` in https://h5web.panosc.eu/h5grove) : one column for radial values, one for intensity and an optional one for errors.

Such an export can be easily done with a Python script but one needs to know the path of the data to export first, which requires exploration of the HDF5 file with an external tool (H5Web, silx, ...). 

H5Web could bridge the gap here by providing exploration of HDF5 files and an automatic export to CSV within a single application. It would also mean that downloading the file would not be mandatory to export the data, making it very useful in remote viewing situations like the ESRF data portal.

Have an Export as CSV button for the NXSpectrum vis. It should not be that difficult given that we already have similar export capabilities for other vis. 

- Python script: requires prior knowledge of the data path and disk access to the file
- Dedicated GUI for HDF5 exploration and export: requires new development and disk access to the file

### Additional context

Specific instance of #1619. 

Loosely related to https://github.com/silx-kit/h5web/issues/1639 that I never got around to do :disappointed:

axelboc:
## CSV export for _NX Line_

- Fix #1723 
- Fix #1619

image

The following columns are exported, in order, when the corresponding datasets are present in the `NXdata` group: values, abscissas, errors, aux1, aux1 errors, aux2, aux2 errors, ... For instance, with mock dataset `/nexus_entry/spectrum_with_aux`, the CSV file looks like this:

image

> Note that the column names are NeXus labels (potentially with long name and unit) rather than dataset names. I will try to solve this in a subsequent PR.

### Built-in, provider-agnostic CSV and JSON exports

CSV and JSON exports are now built into the viewer, which means that consumer applications (VS Code, myHDF5, etc.) no longer have to implement them themselves. All that's needed is for providers to include a very basic implementation of the `getExportURL` method internally.

Thanks to this, the four provider demos (h5grove, h5wasm, HSDS and mock) now all support exporting to CSV/JSON and do so in a consistent way — i.e. the CSV is always generated locally (it used to be generated remotely in the h5grove demo).

### Export compound scalar/array datasets to CSV

- Fix #1639

image

image

### Respect transposition when exporting 2D dataset to CSV in _Matrix_ vis

- Fix #1454

image

image

## Summary

The tables below summarise export capabilities out of the box before/after this PR for every visualization/format/provider combination. "Out of the box" means without any consumer-provided export implementation.

- :warning: only the signal is exported
- :triangular_flag_on_post: transposition is not respected

<details>
<summary><strong>BEFORE</strong></summary>

| Visualization     | Format | `H5GroveProvider`                | `H5WasmProvider`   | `HsdsProvider` | `MockProvider`     |
|-------------------|--------|----------------------------------|--------------------|----------------|--------------------|
| Raw               | JSON   | :white_check_mark:               | :white_check_mark: | :x:            | :white_check_mark: |
| Matrix            | CSV    | :white_check_mark: :triangular_flag_on_post:             | :x:                | :x:            | :white_check_mark: :triangular_flag_on_post: |
|                   | NPY    | :white_check_mark: :triangular_flag_on_post:              | :x:                | :x:            | :x:                |
| Compound | CSV    | :x:                              | :x:                | :x:            | :x:                |
| Line              | CSV    | :white_check_mark:               | :x:                | :x:            | :white_check_mark: |
|                   | NPY    | :white_check_mark:               | :x:                | :x:            | :x:                |
| NX Line           | CSV    | :warning:         | :x:                | :x:            | :x:                |
|                   | NPY    | :warning:         | :x:                | :x:            | :x:                |
| Heatmap           | NPY    | :white_check_mark: :triangular_flag_on_post:              | :x:                | :x:            | :x:                |
|                   | TIFF   | :white_check_mark: :triangular_flag_on_post:              | :x:                | :x:            | :x:                |
| NX Heatmap        | NPY    | :warning: :triangular_flag_on_post:  | :x:                | :x:            | :x:                |
|                   | TIFF   | :warning: :triangular_flag_on_post:  | :x:                | :x:            | :x:                |

</details>

<details>
<summary><strong>AFTER</strong></summary>

| Visualization     | Format | `H5GroveProvider`                | `H5WasmProvider`   | `HsdsProvider` | `MockProvider`     |
|-------------------|--------|----------------------------------|--------------------|----------------|--------------------|
| Raw               | JSON   | :white_check_mark:               | :white_check_mark: | :white_check_mark:            | :white_check_mark: |
| Matrix            | CSV    | :white_check_mark:              | :white_check_mark:                | :white_check_mark:            | :white_check_mark: |
|                   | NPY    | :white_check_mark: :triangular_flag_on_post:              | :x:                | :x:            | :x:                |
| Compound | CSV    | :white_check_mark:                              | :white_check_mark:                | :white_check_mark:            | :white_check_mark:                |
| Line              | CSV    | :white_check_mark:               | :white_check_mark:                | :white_check_mark:            | :white_check_mark: |
|                   | NPY    | :white_check_mark:               | :x:                | :x:            | :x:                |
| NX Line           | CSV    | :white_check_mark:         | :white_check_mark:                | :white_check_mark:            | :white_check_mark:                |
|                   | NPY    | :warning:         | :x:                | :x:            | :x:                |
| Heatmap           | NPY    | :white_check_mark: :triangular_flag_on_post:              | :x:                | :x:            | :x:                |
|                   | TIFF   | :white_check_mark: :triangular_flag_on_post:              | :x:                | :x:            | :x:                |
| NX Heatmap        | NPY    | :warning: :triangular_flag_on_post:  | :x:                | :x:            | :x:                |
|                   | TIFF   | :warning: :triangular_flag_on_post:  | :x:                | :x:            | :x:                |

</details>

---

### Challenges and early attempts

Providers **shouldn't know about NeXus**, so I discarded the following solutions quickly:

- implementing a specific endpoint in h5grove;
- sending axis/auxiliary/errors values to the provider (e.g. via `getExportURL`) along with the main dataset's `value`;
- sending the `NXdata` metadata and letting the provider get the values again internally.

:point_right: NeXus exports have to be generated **in the viewer** (well, at least the export generation functions have to be prepared in the viewer...)

Currently, providers all accept a `getExportURL` prop, which enables a lot of different scenarios: it allows consumers to generate the exports themselves, to generate one-time download links, etc. It's also what allows exports to work in the VS Code extension, as the webview has to generate the export payload and send it to the extension host so it can be saved to disk.

Being able to implement NPY and TIFF exports in the viewer to get rid of `getExportURL` entirely would have made things a lot easier. After all, we already retrieve the data from the providers, so in a way, they've already done their job. I would then have provided another kind of API to allow consumers of the `App` component to customise the exports and hook onto the exports process — providers would no longer have had anything to do with exports... Of course I discarded this option as well, as generating NPY and TIFF in JS is not trivial, and it's not really something we want to maintain long-term.

So there wasn't really any other option than to find a way to **mix both viewer-side generation and provider-side generation**, while continuing to provide flexibility for consumers.

I tried generating export entries both in the providers and in the viewers and then merging them, but this made it really hard to comprehend which exports would end up available in each visualization.

### Proposed solution

In the end, I settled on a solution that involves **passing an "exporter" function to `getExportURL` instead of the raw `value` array**.

```tsx
// BEFORE
public getExportURL(..., value: Value): ExportURL;

// AFTER
public getExportURL(..., builtInExporter?: () => string): ExportURL;
```

I introduce a hook called `useExportEntries` that I call from the mapped components. Internally, it calls `getExportURL`, passing the exporter for the given format if any:

```tsx
// `MappedHeatmapVis`: no viewer-side exporter
const exportEntries = useExportEntries(['tiff', 'npy'], dataset, selection);

// `MappedRawVis`: simple JSON exporter
const exportEntries = useExportEntries(['json'], dataset, undefined, {
  json: () => JSON.stringify(value, null, 2),
});

// `MappedLineVis`: CSV exporter
const exportEntries = useExportEntries(['npy', 'csv'], dataset, selection, {
  csv: () => generateCsv(...),
});
```

In the providers, I then make sure to return an `ExportURL` function that calls `builtInExporter` if it's defined:

```tsx
public override getExportURL(...): ExportURL | undefined {
    const url = this._getExportURL?.(...); // consumer can still override (e.g. for VS Code)

    if (url) {
      return url;
    }

    if (builtInExporter) {
      return async () => new Blob([builtInExporter()]);
    }
    
    // Provider-side exports go here — e.g.:
    // return new URL(...);
```


packages/app/src/index.ts
```diff
@@ -9,8 +9,12 @@ export { default as H5GroveProvider } from './providers/h5grove/H5GroveProvider'
 export { enableBigIntSerialization } from './utils';
 export { getFeedbackMailto } from './breadcrumbs/utils';
 export type { FeedbackContext } from './breadcrumbs/models';
-export type { ExportFormat, ExportURL } from './providers/models';
 export type GetExportURL = NonNullable<DataProviderApi['getExportURL']>;
+export type {
+  ExportFormat,
+  ExportURL,
+  BuiltInExporter,
+} from '@h5web/shared/vis-models';
```
axelboc, line 17: Probably good to check those types out at the bottom of the diff before reviewing the rest.

packages/app/src/providers/api.ts
```diff
@@ -35,12 +34,12 @@ export abstract class DataProviderApi {
    * - `() => Promise<Blob>`  Export is generated client-side
    * - `undefined`            Export scenario is not supported
    */
-  public getExportURL?<D extends Dataset<ArrayShape>>( // optional, so can't be abstract
```
axelboc, line 38: Removing the arbitrary restriction on `ArrayShape` to support exporting scalar datasets like scalar compounds.

packages/app/src/vis-packs/core/compound/MappedCompoundVis.tsx
```diff
@@ -49,12 +53,19 @@ function MappedCompoundVis(props: Props) {
...
 
-  const { getExportURL } = useDataContext();
-  const selection = getSliceSelection(dimMapping);
+  const exportEntries = useExportEntries(['npy', 'csv'], dataset, selection, {
+    csv: () =>
+      generateCsv(
+        fieldNames,
+        mappedArray,
+        Object.values(fields).map((field) => getCsvFormatter(field)),
+      ),
```
axelboc, line 67: Each mapped component that supports CSV exports implements its own CSV exporter: `MappedLineVis`, `MappedCompoundVis` and `MappedMatrixVis`.

packages/app/src/vis-packs/core/heatmap/HeatmapToolbar.tsx
```diff
@@ -109,14 +107,8 @@ function HeatmapToolbar(props: Props) {
...
 
-      {getExportURL && (
-        <ExportMenu
-          isSlice={isSlice}
-          entries={EXPORT_FORMATS.map((format) => ({
-            format,
-            url: getExportURL(format),
-          }))}
-        />
+      {exportEntries.length > 0 && (
```
axelboc, line 110: If an `ExportEntry` is in the array, it is now sure to be supported (cf. `useExportEntries`). This allows hiding the export menu when there are no export entries (instead of just disabling it).

packages/app/src/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -104,10 +106,31 @@ function MappedLineVis(props: Props) {
...
+
+  const exportEntries = useExportEntries(['npy', 'csv'], dataset, selection, {
+    csv: () =>
+      generateCsv(
+        valueLabel,
+        dataArray,
+        abscissaParams,
+        errorsArray,
+        auxiliaries,
+      ),
```
axelboc, line 132: This is the entry point for _NX Line_ CSV exports. As mentioned, it's not perfect because I don't have easy access to the axis/auxiliary dataset names; I only have access to their label (which can include `long_name` and `units`). I will work on this.

packages/app/src/vis-packs/core/matrix/utils.ts
```diff
@@ -80,6 +82,20 @@ export function getFormatter(
...
+): (val: ScalarValue<T>) => string;
+
+export function getCsvFormatter<T extends PrintableType>(
+  type: T,
+): ValueFormatter<PrintableType> {
+  if (isComplexType(type)) {
+    return createComplexFormatter((val) => val.toString());
+  }
+
+  return (val: string | number | bigint | boolean) => val.toString();
```
axelboc, line 96: I chose to keep the raw number values when exporting enum and boolean datasets. Not sure if it's the right choice of if users would prefer enum value labels and `true`/`false`. :shrug:

packages/app/src/vis-packs/core/raw/MappedRawVis.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { isBinaryImage } from './utils';
+
+interface Props {
+  dataset: Dataset<ScalarShape | ArrayShape>;
+  value: unknown;
+  toolbarContainer?: HTMLDivElement | undefined;
+  config: RawConfig;
+}
+
+function MappedRawVis(props: Props) {
```
axelboc, line 22: I had to create a component to be able to call `useExportEntries`. There's no dimension mapper in the raw vis so it's a bit counter-intuitive, but I may need to add one soon to support stacks of binary images.

packages/shared/src/vis-models.ts
```diff
@@ -66,6 +66,15 @@ export interface Dims {
 
 export type IgnoreValue = (val: number) => boolean;
 
+export interface ExportEntry {
+  format: string;
+  url: ExportURL;
```
axelboc, line 71: `url` used to be option; now it isn't.

loichuder (commented):
That looks quite good!

There are a few things that bother me a bit:
- The implicit coupling between the `formats` and the record of built-in exporters passed to `useExportEntries`
- Depending on where you look in the code, it can be difficult to know the priority between `getExportURL`, `_getExportURL` and the built-in exporter.

But all and all, the fact that this PR addresses several long-standing and difficult issues at once is a testimony of its relevance.

packages/shared/src/vis-models.ts
axelboc: `ExportEntry` was already defined in `ExportMenu` before. It's meant to type an entry in the export menu, so I think the naming is good as is.

`BuiltInExporter` is the term I used for functions that generate the export payload in the mapped components. The provider then generates an `ExportURL` async function that invokes this exporter, and the mapped components use those `ExportURL`s in the export entries they pass to the toolbars.

packages/app/src/providers/mock/mock-api.ts
```diff
@@ -99,45 +93,35 @@ export class MockApi extends DataProviderApi {
...
-
-        const finalCsv = csv.slice(0, -2);
-
-        // Demonstrate both `Blob` and `URL` techniques (cf. `src/providers/api.ts`)
-        return dataset.name === 'oneD'
-          ? new Blob([finalCsv])
-          : new URL(`data:text/plain,${encodeURIComponent(finalCsv)}`);
-      };
-    }
+    return async () => {
```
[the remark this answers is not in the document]
axelboc: Hmm, no because the mock provider relies entirely on the viewer providing a `builtInExporter`. It doesn't care which format the exporter is for; it just invokes it.

To make that clearer, I've renamed the local variable `csv` (used to store the result of calling `builtInExporter`) to `payload`.

axelboc:
I've updated the documentation of the `getExportURL` prop, if you don't mind having a look again @loichuder: https://github.com/silx-kit/h5web/pull/1763/commits/b8864441cb5582a2da37e9cb452a770af50e5627

---

> The implicit coupling between the formats and the record of built-in exporters passed to `useExportEntries`

Agreed, not perfect, but the types ensure that the record cannot contain formats that are not in the `formats` array. I think it's better than to try to force the two parameters into a single `Record`, and to use `null` or `false` values as a way to identify formats that may be handled by the provider.

> Depending on where you look in the code, it can be difficult to know the priority between `getExportURL`, `_getExportURL` and the built-in exporter.

Totally agree as well, it gets quite confusing. This actually prompted me to update the documentation of the `getExportURL` prop, so thanks! :grin:

Eventually I'm hoping to at least remove the `builtInExporter` parameter from `getExportURL`. It's only done this way to support the VS Code use case for now. A better way would be to pass an `onExportPayload` callback to the viewer somehow to get direct access to the built-in export payload when it's generated.

However, simply adding an `onExportPayload` prop to the `App` component and drilling it down to all the mapped components was out of the question. It's part of the broader goal of making the viewer more customisable/extendable. I have a few ideas in mind...

packages/app/README.md
```diff
@@ -269,12 +254,19 @@ getExportURL={(format, dataset, selection) => async () => {
 }}
 ```
 
-</details>
+```tsx
+// Tweak a built-in export payload in some way (round or format numbers, truncate lines, etc.)
+getExportURL={(format, dataset, selection, buitInExporter) => async () => {
```
loichuder, line 259:
```suggestion
getExportURL={(format, dataset, selection, builtInExporter) => async () => {
```

packages/app/README.md
```diff
@@ -269,12 +254,19 @@ getExportURL={(format, dataset, selection) => async () => {
...
+  if (!buitInExporter || format !== 'csv') {
```
loichuder, line 260:
```suggestion
  if (!builtInExporter || format !== 'csv') {
```

loichuder (approved):

PR #1768, opened 2025-03-14
Improve visualization of "raw" datasets with array shapes

axelboc:
This PR improves handling of array datasets by the _Raw_ visualization. The idea is to visualize only **one scalar slice at a time** instead of the entire array, notably to support **stacks of raw JPEG/PNG images**.

It's the first of three steps. I'm planning to then:

- remove _Scalar_ vis and let _Raw_ vis handle all scalars;
- rename _Raw_ back to _Scalar_ (and everything else called "raw").

This way we'll end up with a **single _Scalar_ vis fallback** to visualize **all scalar values** (whether from a scalar dataset or from a scalar slice of an array dataset).

---

The main challenge was dealing with "scalar selections". I'm talking about `selection` strings of the form `[0-9](,[0-9])*` (e.g. `"2"`, `"0,1"`, `"2,0,4"`, etc.) This was previously [discussed here](https://github.com/silx-kit/h5web/pull/1768#discussion_r1998422883).

The first problem came from the providers. h5grove handles these scalar selections as expected — i.e. it returns a scalar value — but h5wasm and HSDS have quirks:

- h5wasm inconsistently unwraps the scalar depending on the dtype of the dataset — I'll endeavour to contribute a fix to the h5wasm repo at some point.
- HSDS never unwraps the scalar — a version 1.0.0 should be coming soon that seems to be a major overhaul, so I'll make sure to adapt the provider and [test it](https://github.com/silx-kit/h5web/pull/1926) thoroughly once it's out.

The second problem came from our `Value<D extends HasShape & HasType>` type, which is not aware of "scalar selections" — i.e. if you give it a dataset with an array shape as parameter, it returns an `ArrayValue` type, full stop. So the `ValueFetcher` component, `useValue` hook and `assertValue` guard would either lie or fail when encountering scalar slices taken from array datasets.

For a while, I considered making `Value<D>`, `ValueFetcher`, `useValue` and `assertValue` aware of selections but this made everything way more complicated and messy. Instead, I ended up creating a **`ScalarFetcher` component** that handles this case specifically in `RawVisContainer` without affecting any of the types, hooks and guards.

---

I've updated `sample.h5` (demos: [h5grove](https://h5web.panosc.eu/h5grove?file=sample.h5), [h5wasm](https://h5web.panosc.eu/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Fsample.h5)) by merging `compound_nested_scalar` and `compound_array_vlen_1D` into a more comprehensive `compound_mixed_1D` dataset that demonstrates everything at once (vlen/array dtypes + nested compound with int64 and complex fields).

| PROVIDER | BEFORE | AFTER |
| --- | --- | --- |
| h5grove | image | image |
| h5wasm | image | image |

> Note that the following datasets in `sample.h5` also show the dimension mapper now: `byte_string_1D`, `vlen_*_1D`.

I've updated the mock file with a similar `oneD_compound_mixed` dataset, as well as with multiple opaque datasets: `oneD_opaque`, `oneD_opaque_png`, and `twoD_opaque`.

image

The mock `oneD_opaque_png` dataset demonstrates how the new _Array_ vis handles stacks of PNG images, but you can also try it with [`frogs.h5`](https://h5web.panosc.eu/h5grove?file=frogs.h5):

image

> ZIP with h5py script, images and `frogs.h5`: [frogs.zip](https://github.com/user-attachments/files/19249129/frogs.zip)


packages/app/src/providers/mock/utils.ts
```diff
@@ -46,19 +43,22 @@ export function findMockEntity(
...
-  dataset: Dataset<ArrayShape | ScalarShape, T>,
+export function sliceValue(
+  value: unknown[],
+  dataset: Dataset<ArrayShape>,
   selection: string,
-): ScalarValue<T>[] {
+): unknown {
   const { shape } = dataset;
-  const dataArray = ndarray(value as ScalarValue<typeof dataset.type>[], shape);
-  const mappedArray = applyMapping(
```
axelboc, line 56: `applyMapping` brings some unnecessary misdirection. Instead, I now call `pick` and `createArrayFromView` directly.

And since `pick` always returns an array even when selecting a scalar, I add a condition to extract the scalar.

packages/app/src/vis-packs/core/visualizations.ts
```diff
@@ -72,7 +75,7 @@ export const CORE_VIS = {
     Icon: FiCpu,
     Container: RawVisContainer,
     ConfigProvider: RawConfigProvider,
-    supportsDataset: hasNonNullShape,
+    supportsDataset: hasScalarShape,
```
axelboc, line 78: This is just to avoid complexifying the fallback visualization code in the `Visualizer` for now. As mentioned, the plan is to merge `RawVis` into `ScalarVis`.

packages/app/src/vis-packs/core/visualizations.ts
```diff
@@ -167,8 +178,7 @@ export const CORE_VIS = {
       return (
         hasCompoundType(dataset) &&
         hasPrintableCompoundType(dataset) &&
-        hasNonNullShape(dataset) &&
-        (hasScalarShape(dataset) || hasMinDims(dataset, 1))
```
axelboc, line 171: The last line actually did the same check as `hasNonNullShape`...

packages/app/src/visualizer/utils.ts
```diff
@@ -82,6 +82,8 @@ function getNxDefaultPath(
   return getImplicitDefaultChild(entity.children, attrValueStore)?.path;
 }
 
+const FALLBACK_VIS = new Set([Vis.Raw, Vis.Array]);
```
axelboc, line 85: This will become `[Vis.Scalar, Vis.Array]` in my next PR.

packages/h5wasm/src/worker-utils.ts
```diff
@@ -321,5 +322,14 @@ export function readSelectedValue(
...
+  const slicedValue = h5wDataset.slice(ranges);
+  assertNonNull(slicedValue);
+
+  /* h5wasm unwraps scalar slices inconsistently - e.g. it does so for opaque
+   * datasets but not for compound datasets */
+  if (isScalarSelection(selection) && Array.isArray(slicedValue)) {
+    return slicedValue[0];
+  }
+
+  return slicedValue;
```
axelboc, line 334: I'll open an issue on the h5wasm repo when I get the time.

packages/shared/src/guards.test.ts
```diff
@@ -133,27 +133,100 @@ describe('assertDatasetValue', () => {
...
+
+    expect(() =>
+      assertDatasetValue(
+        Float32Array.from([0, 1]), // big ints can be returned as any kind of numbers
+        dataset('foo', intType(true, 64), [2]),
+      ),
+    ).not.toThrow();
   });
 
-  describe('assertDatasetValue', () => {
```
axelboc, line 138: Duplicate `describe` block.

packages/shared/src/guards.test.ts
```diff
@@ -133,27 +133,100 @@ describe('assertDatasetValue', () => {
...
+      assertDatasetValue(
+        BigInt64Array.from([0n, 1n]),
+        dataset('foo', intType(), [2]),
+      ),
+    ).toThrow('Expected number');
+  });
+
+  it('should not throw when value shape satisfies selection', () => {
+    expect(() =>
+      assertDatasetValue(
```
axelboc, line 167: `assertDatasetValue` now takes a `selection` string and asserts that the value is a scalar when a scalar selection is passed.

packages/shared/src/guards.ts
```diff
@@ -544,16 +554,22 @@ export function assertScalarValue(
     Object.values(type.fields).forEach((fieldType, index) => {
       assertScalarValue(value[index], fieldType);
     });
+  } else if (isArrayOrVlenType(type)) {
```
axelboc, line 557: Adding a case for `H5T_ARRAY` and `H5T_VLEN` dtypes.

packages/app/src/providers/h5grove/__snapshots__/h5grove-api.test.ts.snap
axelboc, line 1: cf. `create_h5_sample.py` at the bottom of the diff.

axelboc:
> h5grove seems to handle these scalar selections as expected — i.e. it returns a scalar value — but h5wasm and HSDS have some quirks... I wrote some comments in the code that are hopefully clear enough.

I've just realised that this assumption that providers should return scalar values for scalar selections in array datasets is problematic in terms of typing. In `assertDatasetValue`, the return type is `Value<Dataset>`, which only cares about the dataset's shape: if it's `ArrayShape`, we get `ArrayValue`, if it's `ScalarShape`, we get `ScalarValue`.

With the changes in this PR, the return type `Value<Dataset<ArrayShape, T>>` should actually resolve to `ScalarValue<T> | ArrayValue<T>` instead of just `ArrayValue<T>`. So `assertDatasetValue` is lying.

I have to investigate the impact of changing the `Value` type.

loichuder:
Looking at your screenshots, I wonder if the move should not be to extend the `Matrix` vis to support these unsupported arrays.

And to create a dedicated vis for JPEG/PNG ?

axelboc:
> Looking at your screenshots, I wonder if the move should not be to extend the Matrix vis to support these unsupported arrays.

The limitation of the _Matrix_ vis is that the space for each item is restricted: single row, fixed width. So vlens, arrays, complicated compounds, etc. wouldn't fit. We could imagine nesting the _Matrix_ vis: if an item is a vlen/array/compound, allow the user to click on the cell to explore the value, but this would take more clicks than with the dimension mapper.

The other problem is that we can't tell the dataset contains JPEG/PNG image(s) until we actually fetch its value. If we decide to trigger the _Matrix_ vis by default for an opaque 1D dataset, for instance, we end up with a bunch of problems (no slider in the dimension mapper, _Matrix_ tab that doesn't fit the visualization, etc.)

> For example, what if I want to inspect the binary blob for some reason ? I would expect to select the Raw vis in this case.

Yeah, it's true that in the new _Array_ vis, users can only export one slice at a time to JSON. There's no way to see the entire dataset. If that's really a problem for someone, though, perhaps we could make the sliders togglable somehow?

loichuder:
> The limitation of the _Matrix_ vis is that the space for each item is restricted: single row, fixed width. So vlens, arrays, complicated compounds, etc. wouldn't fit. We could imagine nesting the _Matrix_ vis: if an item is a vlen/array/compound, allow the user to click on the cell to explore the value, but this would take more clicks than with the dimension mapper.

Gotcha. Yeah, in essence, you wanted a Raw vis but with a dimension mapper for things that we know are arrays but don't have a suitable visualization .

> The other problem is that we can't tell the dataset contains JPEG/PNG image(s) until we actually fetch its value. 

I didn't remember that, that's unfortunate :slightly_frowning_face:

packages/app/src/dimension-mapper/store.tsx
```diff
@@ -23,7 +23,7 @@ interface DimMappingState {
   ) => void;
 }
 
-function createLineConfigStore() {
```
loichuder, line 26: ??

Why did we name it like that ? :joy:

packages/shared/src/guards.ts
```diff
@@ -523,6 +525,14 @@ export function isComplexValue(
   return type.class === DTypeClass.Complex;
 }
 
+export function isArrayOrVlenType(type: DType): type is ArrayType | VLenType {
+  return type.class === DTypeClass.Array || type.class === DTypeClass.VLen;
+}
+
+export function isScalarSelection(selection: string | undefined): boolean {
+  return selection !== undefined && /^\d+(?:,\d+)*$/u.test(selection);
```
loichuder, line 533: I struggle to understand what you mean by "scalar selection". 

Is it a selection that will produce a scalar value ? If so, the test should check also the dataset shape ? 

Is it a selection that only has scalar values inside ? In that case, perhaps "selected indices" would be a better name ?

axelboc: It's a selection that produces a scalar value, like `2,0,1` or `3`, as opposed to `2,0,:` or `:`.

The dataset shape doesn't matter here: when we have a selection string, the dataset is expected to have an array shape (cf. `assertDatasetValue`). The selection string is also expected to have the same number of elements as the number of dimensions in the dataset.

---

The problem I'm facing (which is why I've switched the PR to draft) is that the `Value` type has to be aware of the selection so it can conditionally return `ScalarValue` for "scalar selections".

I've managed to make it work with:

```ts
type ValueSelection = (number | null)[];
type ScalarSelection = number[];

export type Value<
  D extends Dataset,
  V extends ValueSelection | undefined = undefined,
> =
  D extends Dataset<infer S, infer T>
    ? S extends ScalarShape
      ? ScalarValue<T>
      : S extends ArrayShape
        ? V extends ScalarSelection
          ? ScalarValue<T>
          : ArrayValue<T>
        : never
    : never;
```

It requires keeping the selection as an array for longer, which changes quite a few things. 

I'm now trying to use the dimension mapping array directly instead of introducing a new structure. I'll keep you updated.

loichuder: > It's a selection that produces a scalar value, like 2,0,1 or 3, as opposed to 2,0,: or :.

I see. The issue I had is that I am used to Python where you can do stuff like `threeD[3]` and get a 2D array. In essence, the `3` is a short hand for `3, :, :`. 

But since we enforce the same number of elements as the number of dimensions in the dataset, we should never have this.

packages/h5wasm/src/worker-utils.ts
```diff
@@ -320,5 +319,13 @@ export function readSelectedValue(
     return [Number(member), Number(member) + 1];
   });
 
-  return h5wDataset.slice(ranges);
+  const slicedValue = h5wDataset.slice(ranges);
+
+  /* h5wasm unwraps scalar slices inconsistently - e.g. it does so for opaque
+   * datasets but not for compound, vlen, etc. */
+  if (/^\d+(?:,\d+)*$/u.test(selection) && Array.isArray(slicedValue)) {
```
loichuder: We should perhaps compile the Regexp and import it where it is needed?

packages/h5wasm/src/worker-utils.ts
```diff
@@ -303,7 +302,7 @@ function parseVirtualSources(metadata: Metadata): VirtualSource[] | undefined {
 export function readSelectedValue(
   h5wDataset: H5WasmDataset,
   selection: string | undefined,
-): OutputData | null {
+): unknown {
```
loichuder, line 306: Is this intentional?

packages/app/src/vis-packs/core/raw/RawVisContainer.tsx
```diff
@@ -1,32 +1,56 @@
...
 
 function RawVisContainer(props: VisContainerProps) {
   const { entity, toolbarContainer } = props;
   assertDataset(entity);
   assertNonNullShape(entity);
 
+  const { dims } = entity.shape;
+  const [dimMapping, setDimMapping] = useDimMappingState({
+    dims,
+    axesCount: 0, // slicing only
```
loichuder, line 21: Remind me what is this for?

packages/h5wasm/src/worker-utils.ts
axelboc: Yeah, I couldn't import `isScalarSelection` from the worker because of where it lives, yet I didn't want to move it to the shared package (couldn't really find a good place for it). For now I've moved the regex itself to the shared package as `SCALAR_SELECTION_REGEXP` in `vis-utils.ts`. I'm hoping I'll be able to revert this once h5wasm is fixed.

packages/h5wasm/src/worker-utils.ts
axelboc: Yes, the return type gets in the way a bit and `DataProvider#getValue` has return type `unknown`, so it "swallows" whichever return type we specify here. We assert the value after it's returned by the provider, so we don't need to assert it more than strictly necessary inside this function.

packages/app/src/vis-packs/core/raw/RawVisContainer.tsx
axelboc: It says that the visualization doesn't have any "axes" and therefore that every dimension needs to be "sliced". The returned `dimMapping` is either an empty array if `dims` is empty (i.e. scalar dataset), or an array with as many zeros as there are dimensions in the array dataset.

This is why we are guaranteed to end up with either an `undefined` selection for scalar datasets, or a "scalar selection" for array datasets.

> As a reminder, when `axesCount` is 1, `dimMapping` includes one `'x'`, and when `axesCount` is 2, `dimMapping` includes one `'x'` and one `'y'`. You can check out the [unit tests of `initDimMapping`](https://github.com/silx-kit/h5web/blob/7260a8ca625ef5579052fd67398b5eb88f45f313/packages/lib/src/dimension-mapper/utils.test.ts) for some examples.

PR #1770, opened 2025-03-14
Merge Raw vis into Scalar vis

axelboc:
As discussed, we end up with two fallback visualizations: one for scalar datasets and one for array datasets. If any other visualization is supported (e.g. _Compound_ for printable scalar compound datasets), then _Scalar_ vis is not shown.

I'm planning to clean-up the mock file a fair bit but didn't want to clutter this PR, which is already confusing because of all the files that are moved and renamed.


packages/app/src/vis-packs/core/scalar/MappedScalarVis.tsx
```diff
@@ -41,17 +42,17 @@ function MappedRawVis(props: Props) {
         )}
 
       {isImage ? (
-        <RawImageVis
+        <BinaryImageVis
```
axelboc, line 45: I renamed this component, since the term "Raw" no longer makes much sense.

packages/app/src/vis-packs/core/scalar/utils.ts
```diff
@@ -1,35 +1,56 @@
...
   }
 
-  return (val: number | bigint | string) => val.toString();
+  return (val: unknown) => {
+    if (isTypedArray(val)) {
+      return `${val.constructor.name} [ ${val.toString()} ]`;
+    }
+
+    return JSON.stringify(val, null, 2);
+  };
```
axelboc, line 55: This takes care of formatting non-printable scalar values. This code was previously in `RawVis`.

packages/app/src/vis-packs/core/scalar/ScalarVisContainer.tsx
```diff
@@ -1,36 +1,56 @@
...
+import { useScalarConfig } from './config';
+import MappedScalarVis from './MappedScalarVis';
+import ScalarFetcher from './ScalarFetcher';
 
 function ScalarVisContainer(props: VisContainerProps) {
-  const { entity } = props;
+  const { entity, toolbarContainer } = props;
   assertDataset(entity);
-  assertScalarShape(entity);
-  assertPrintableType(entity);
```
axelboc, line 18: `ScalarVisContainer` now accepts anything with a non-null shape.

packages/app/src/vis-packs/core/visualizations.test.ts
```diff
@@ -35,11 +35,6 @@ const mockStore = {
 const nullInt = dataset('null', nullShape(), intType());
 
 const scalarInt = dataset('int', scalarShape(), intType());
-const scalarUint = dataset('uint', scalarShape(), intType(false));
-const scalarBigInt = dataset('bigint', scalarShape(), intType(true, 64));
-const scalarFloat = dataset('float', scalarShape(), floatType());
-const scalarStr = dataset('float', scalarShape(), strType());
-const scalarBool = dataset('bool', scalarShape(), boolType(intType(true, 8)));
```
axelboc, line 42: Testing `CORE_VIS.Scalar` with all these is now overkill since it doesn't care about the dtype any more.

packages/app/src/vis-packs/core/scalar/ScalarVisContainer.tsx
```diff
@@ -1,36 +1,56 @@
...
+      {dims.length > 0 && (
+        <DimensionMapper
+          className={visualizerStyles.dimMapper}
+          dims={dims}
+          dimMapping={dimMapping}
+          canSliceFast={canSliceFast}
+          onChange={setDimMapping}
+        />
+      )}
+      <VisBoundary resetKey={dimMapping} isSlice={dims.length > 0}>
```
axelboc, line 39: The code comes from `RawVisContainer` with one subtle fix here caught by our _Scalar_ vis tests: `isSlice` is now set dynamically to ensure that the correct loading message is shown ("Loading data" instead of "Loading slice" when fetching the value of a scalar dataset)

PR #1772, opened 2025-03-18
Don't consider negative domain as valid in log/sqrt scale

### LINKED ISSUE #1761: The user can't revert to linear scale after setting the scale to "Log" or "Square root" on an unsupported dataset
### Describe the bug

After switching to "Log" or "Square root" scale on an unsupported dataset, the scale selector disappears, making it impossible to reset the scale.

### To Reproduce

1. Create a dataset that only contains negative values (e. g. a 2D matrix).
Example Python code:
```
import h5py

with h5py.File("test.h5", "w") as hdf5_file:
    hdf5_file.create_dataset("positive_matrix", data=[[1, 2], [3, 4]])
    hdf5_file.create_dataset("mixed_matrix", data=[[-1, 2], [3, -4]])
    hdf5_file.create_dataset("negative_matrix", data=[[-1, -2], [-3, -4]])
```
2. Open that dataset as a line plot or heatmap.

Image

3. Change the scale to "Log" or "Square root"

Image

4. Now we get an error indicating our dataset cannot be displayed: "Expected domain compatible with log scale". This is expected behaviour, as the logarithm and square root of a negative number are undefined (at least in the real number space). However, the dropdown selector that is needed to revert to the linear scale is gone. The only way I found to get the linear scale back is to open another dataset that contains positive values.

Image

### Expected behaviour

Keep displaying the scale selector even if a dataset contains unsupported data.

### Context

- OS: Windows 11 24H2
- H5Web context: [H5Web VS Code extension](https://github.com/silx-kit/vscode-h5web) v0.1.7
- I also tried [H5Wasm](https://h5web.panosc.eu/h5wasm), so it's not a VS Code related issue

axelboc:
Fix #1761

`getValidDomainForScale` used to consider negative domains like `[-20, -10]` as valid. (If I recall, we made this choice initially because the _Line_ vis and D3 could technically handle it.) However, this meant that we ended up with a negative `dataDomain` throughout, and this didn't work well with the `DomainWidget`: the `extendDomain` function used to throw, which triggered the error boundary and unmounted the toolbar.

I could not find a straightforward way to handle negative domains in the domain widget, so I decided to change `getValidDomainForScale` to not consider negative domains as valid at all. This means we typically fall back to the default domain `[0.1, 1]`, which is handled cleanly and doesn't throw any error (thus allowing to change the scale back to linear or whatever).

> I'm not super happy that there's no visual feedback to indicate that the chosen data domain differs from the real data domain, but this is a problem for another day...

[Screencast from 2025-03-18 15-39-20.webm](https://github.com/user-attachments/assets/9027e8fa-d562-4673-8b1c-f8d9cd78bfcb)


packages/lib/src/vis/shared/ViewportCenterer.tsx
```diff
@@ -17,9 +17,14 @@ function ViewportCenterer() {
...
+    // On resize, move camera to the latest saved viewport center coordinates
     if (viewportCenter.current) {
-      // On resize, move camera to the latest saved viewport center coordinates
-      moveCameraTo(dataToWorld(viewportCenter.current));
+      const newPos = dataToWorld(viewportCenter.current);
+
+      // Ignore previous center if no longer valid with current scale
+      if (Number.isFinite(newPos.x) && Number.isFinite(newPos.y)) {
+        moveCameraTo(newPos);
+      }
```
axelboc, line 27: Subtle bug fix here: when switching from linear to log when visualising a dataset with a negative domain in the _Line_ vis, the viewport center saved (in data coordinates) while in linear scale cannot be converted back to world coordinates in log scale (because `ordinateScale(<negativeValue>)` returns `NaN`).

This breaks the visualization, because `ViewportCenterer` moves the camera position to a `NaN` y coordinate:

[Screencast from 2025-03-18 16-11-20.webm](https://github.com/user-attachments/assets/b8b2ca81-9b26-4d52-9a1f-3f50c0876ebc)

packages/shared/src/vis-utils.ts
```diff
@@ -167,16 +167,17 @@ export function getValidDomainForScale(
   }
 
   const { min, max, positiveMin, strictPositiveMin } = bounds;
-  if (scaleType === ScaleType.Log && min * max <= 0) {
```
axelboc, line 170: The fix is to mainly to replace `min * max <= 0` with `min <= 0`.

packages/lib/src/vis/shared/ViewportCenterer.tsx
loichuder: :fishing_pole_and_fish:

loichuder (approved):
Makes sense. Even Matplotlib does not support negative values in log plots: https://matplotlib.org/stable/gallery/scales/log_demo.html

PR #1774, opened 2025-03-18
Allow changing texture mag/min filter of Heatmap vis

### LINKED ISSUE #1721: Support modes other than nearest neighbour downsampling for HeatmapVis
For diffraction applications it would be nice to support some sort of filtering beyond nearest neighbour. If you have a diffraction image being displayed at say 1/4 of its size with single pixel diffraction spots you basically cant see anything in `<HeatmapVis>`. I know @t20100 recently implemented this in silx, i don't know how much work it is for h5web.

axelboc:
Fix #1721

`[HeatmapVis, HeatmapMesh]` Add props `magFilter` and `minFilter` to allow changing the heatmap texture's [magnification and minification filters](https://threejs.org/docs/#api/en/constants/Textures) (`NearestFilter` by default)

### DIFF
--- a/apps/storybook/src/HeatmapMesh.stories.tsx
+++ b/apps/storybook/src/HeatmapMesh.stories.tsx
@@ -40,6 +40,7 @@ const meta = {
   args: {
     invertColorMap: false,
     magFilter: NearestFilter,
+    minFilter: NearestFilter,
   },
   argTypes: {
     scaleType: {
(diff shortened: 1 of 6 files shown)


loichuder (approved):
Cool, thanks !

We will have to see if this is enough or if we should do something more fancy as @t20100 suggested in the issue.

stufisher:
Thanks for this, ill try and check it out

PR #1779, opened 2025-03-27
Don't misuse render prop pattern

axelboc:
I've relaxed the rule [`react/no-unstable-nested-components`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md) to allow the render prop pattern:

```tsx
<ErrorBoundary
  fallbackRender={(args) => (
    <ErrorFallback className={visualizerStyles.vis} {...args} />
  )}
/>
```

The code above is okay (and unavoidable) as long as `fallbackRender` is used as a function inside `ErrorBoundary` - e.g. `{fallbackRender(...args)}`.

It's not okay when the render prop is used as a component. An example of this was the `icon` prop of the `Btn` component:

```tsx
// Btn.tsx
const { icon: Icon } = props;
return <>{Icon && <Icon className={styles.icon} />}</>

<Btn
  // render prop but used as component
  icon={(iconProps) => (
    <FiItalic
      transform="skewX(20)"
      style={{ marginLeft: '-0.25em', marginRight: '0.0625rem' }}
      {...iconProps}
    />
  )}
/>
```

Unfortunately, `react/no-unstable-nested-components` cannot detect these two cases, which leads to false positives. Since the render prop pattern is legitimate and unavoidable in many cases, the rule could lead to worse solutions. So I've decided to turn on the rule's `allowAsProp` option, which allows passing "nested" components as props.

I'm also fixing the `icon` prop by turning it into a component prop:

```tsx
// Btn.tsx
const { Icon } = props;
return <>{Icon && <Icon className={styles.icon} />}</>

<Btn
  // good component prop (icon is extracted into its own component)
  Icon={ErrorsIcon}
/>
```

### DIFF
--- a/apps/demo/package.json
+++ b/apps/demo/package.json
@@ -26,7 +26,7 @@
     "wouter": "3.3.5"
   },
   "devDependencies": {
-    "@esrf/eslint-config": "1.0.3",
+    "@esrf/eslint-config": "1.0.4",
     "@types/node": "^22.12.0",
     "@types/react": "^18.3.3",
     "@types/react-dom": "^18.3.0",
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -28,7 +28,7 @@
     "three": "0.172.0"
   },
   "devDependencies": {
-    "@esrf/eslint-config": "1.0.3",
+    "@esrf/eslint-config": "1.0.4",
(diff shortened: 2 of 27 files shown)


loichuder:
Should we use a specific `propNamePattern` ?

PR #1780, opened 2025-03-31
Put x axis values first in CSV

axelboc:
@Blackclaws in https://github.com/silx-kit/h5web/issues/1619#issuecomment-2082444177 correctly pointed out that axis values are expected to come first in CSV exports, like in `XY` and `XYE` files. Sorry for missing this in #1763.

### NeXus export with axis values

image

### NeXus export without axis values

The indices are used so that signal values remain in the second column; the column is labelled `x`:

image

### Basic 1D export

No axis values, no column headers — does that make sense @t20100 @loichuder?

image


loichuder:
I would perhaps add the dataset name as the column header for the basic 1D export?

packages/app/src/vis-packs/core/line/utils.ts
```diff
@@ -30,8 +31,11 @@ export function generateCsv(
   for (let i = 0; i < dataArray.shape[0]; i += 1) {
     csv += csv.length > 0 ? '\n' : '';
 
+    csv += isNexus
+      ? `${(abscissaParams.value ? abscissaParams.value[i] : i).toString()},`
+      : '';
```
loichuder, line 36:
```suggestion
    if(isNexus) {
        csv += `${(abscissaParams.value ? abscissaParams.value[i] : i).toString()},`
    }
```

loichuder (approved):

axelboc:
> I would perhaps add the dataset name as the column header for the basic 1D export?

Okay, so in that case, I can have the indices in the first column to basically end up with a "bare minimal" NeXus export (i.e. `NXdata` group with only a `signal` attribute):

image

loichuder:
> > I would perhaps add the dataset name as the column header for the basic 1D export?
> 
> Okay, so in that case, I can have the indices in the first column to basically end up with a "bare minimal" NeXus export (i.e. `NXdata` group with only a `signal` attribute):
> 

That looks excellent! Good call :+1:

t20100 (approved):

PR #1782, opened 2025-04-01
Upgrade Vite and related dependencies

axelboc:
- [Vite 6](https://vite.dev/blog/announcing-vite6.html) ([migration](https://vite.dev/guide/migration.html#migration-from-v5), [CHANGELOG](https://github.com/vitejs/vite/blob/v6.2.4/packages/vite/CHANGELOG.md))
  - Only breaking change that concerns us is how Vite names CSS output files. It used to be `style.css` and couldn't be configured. Now, it uses either the package name (e.g. `app`) or the configured JS output `fileName` by default, and can be overridden with a dedicated option called [`cssFileName`](https://vite.dev/config/build-options.html#build-lib). So I made some minor adjustments to the `vite.styles.config.js` files and to the `build:css` scripts.
- [Vitest 3](https://github.com/vitest-dev/vitest/releases?page=2) ([migration](https://vitest.dev/guide/migration.html#migrating-to-vitest-2-0)) and a couple of Vite plugins.
  - Nothing notable here

While I was dealing with the styles, I made some more changes notably to fix hot module replacement. If you recall, in both the app and lib packages we have to files: `src/styles.css` and `src/styles.ts`:

- `src/styles.css` is used to import global styles (`btnClean` and the like, and the React Reflex styles) in the demo and Storybook apps
- `src/styles.ts` is used to build the distribution styles of the NPM packages (as explained in the `CONTRIBUTING` guide).

In this PR:

- I rename the `src/styles.css` files to `src/global-styles.css` to make it clear what we're importing in the demo and Storybooks apps.
- I add `"exports"` definitions to both the app and lib's `package.json` so that styles (global and distributed) can be imported more cleanly — this is what fixes hot module replacement.

### DIFF
--- a/packages/app/README.md
+++ b/packages/app/README.md
@@ -36,7 +36,7 @@ npm install @h5web/app
 ```
 
 ```tsx
-import '@h5web/app/dist/styles.css';
+import '@h5web/app/styles.css';
 
 import React from 'react';
 import { App, MockProvider } from '@h5web/app';
--- a/packages/app/src/styles.css
+++ b/packages/app/src/styles.css
@@ -1,3 +1,0 @@
-/* Global app styles for demo/Storybook */
-@import 'react-reflex/styles.css';
-@import '@h5web/lib/src/styles.css'; /* global lib styles */
--- a/packages/app/src/styles.ts
+++ b/packages/app/src/styles.ts
@@ -3,4 +3,4 @@
    Output is later concatenated with local app styles. */
 
 import 'react-reflex/styles.css';
-import '@h5web/lib/dist/styles.css'; // local + global lib styles
+import '@h5web/lib/dist/styles.css'; // distributed lib styles
--- a/packages/app/vite.styles.config.js
+++ b/packages/app/vite.styles.config.js
@@ -7,7 +7,7 @@ export default defineConfig({
     lib: {
       entry: path.resolve('src/styles.ts'),
       formats: ['es'],
-      fileName: () => 'styles.js',
+      fileName: 'styles',
     },
     outDir: 'dist/temp',
     emptyOutDir: false,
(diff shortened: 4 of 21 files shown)


packages/app/README.md
```diff
@@ -36,7 +36,7 @@ npm install @h5web/app
 ```
 
 ```tsx
-import '@h5web/app/dist/styles.css';
```
axelboc, line 39: The old path is still supported, but the documentation should now assume that consumers use modern bundlers that support `"exports"`.

packages/app/src/styles.css
```diff
@@ -1,3 +0,0 @@
-/* Global app styles for demo/Storybook */
-@import 'react-reflex/styles.css';
-@import '@h5web/lib/src/styles.css'; /* global lib styles */
```
axelboc, line 3: Now cleaned-up to `'@h5web/lib/global-styles.css'`

packages/app/src/styles.ts
```diff
@@ -3,4 +3,4 @@
    Output is later concatenated with local app styles. */
 
 import 'react-reflex/styles.css';
-import '@h5web/lib/dist/styles.css'; // local + global lib styles
+import '@h5web/lib/dist/styles.css'; // distributed lib styles
```
axelboc, line 6: While the import path doesn't change, it is now properly defined in the lib's `package.json` via  `"exports"`.

packages/app/vite.styles.config.js
```diff
@@ -7,7 +7,7 @@ export default defineConfig({
     lib: {
       entry: path.resolve('src/styles.ts'),
       formats: ['es'],
-      fileName: () => 'styles.js',
+      fileName: 'styles',
```
axelboc, line 10: The CSS output file uses the same name: `styles.css`. The JS output file, `styles.js` remains empty as before (and deleted with `rimram` when running the `build:css` script).

loichuder (approved):
Clean :+1:

PR #1787, opened 2025-04-02
Upgrade pnpm to v10

axelboc:
~Draft until [this bug](https://github.com/typescript-eslint/typescript-eslint/issues/8732) is resolved (and released) in `typescript-eslint`.~

~**EDIT**: The issue above is resolved, but upgrading `typescript-eslint` and `eslint` itself has had no effect. I've investigated turning on/off a number of pnpm options, but nothing worked: I still get a type-checking error coming from the `@eslint+config-helpers` package when running`pnpm lint:tsc`. I will keep this as draft and try again in a few weeks.~

**EDIT** Had to use ChatGPT to get me out of this :scream: — it led me to pnpm's [hoisting options](https://pnpm.io/settings#dependency-hoisting-settings). After some investigations, I found `publicHoistPattern: ['@types/*']` to be a good compromise. This installs all `@types` packages to the root of the `node_modules` folder to make sure they're available from everywhere. This is apparently a [common practice](https://github.com/orgs/pnpm/discussions/5535).

I have to be honest, I can't say that I understand all the intricacies of hoisting and why, with pnpm 10, hoisting some `@types` packages "privately" (i.e. to `node_modules/.pnpm/node_modules`) instead of "publicly" (i.e. to `node_modules`) no longer works. :shrug: I will need to investigate this more, notably to double check whether I can hoist only specific `@types` packages.

---

Lots of changes in [pnpm 10](https://github.com/pnpm/pnpm/releases/tag/v10.0.0) and later, like [10.5](https://github.com/pnpm/pnpm/releases/tag/v10.5.0) and [10.6](https://github.com/pnpm/pnpm/releases/tag/v10.6.0). I'm glad I did not upgrade right away when 10.0 came out... :joy:

- Configuring pnpm via `.npmrc` and `package.json` is now deprecated. Everything can now live inside **[`pnpm-workspace.yaml`](https://pnpm.io/settings)**, which is awesome.
  - I took the opportunity to recheck all the available options and discovered a few interesting ones I had missed and might be useful in the future like [`updateConfig`](https://pnpm.io/settings#updateconfig).
  - I also enabled a couple of new options from pnpm@10.x like [`strictDepBuilds`](https://pnpm.io/settings#strictdepbuilds) and [`verifyDepsBeforeRun`](https://pnpm.io/settings#verifydepsbeforerun) (cf. below).
- `pnpm test` no longer requires `--` to pass arguments to the underlying binary, just like with `pnpm run`
   - This is actually an issue I reported [two years ago](https://github.com/pnpm/pnpm/issues/4821).
- pnpm no longer runs dependency build scripts on install for security reasons. Each dependency with a build script has to be pre-approved. If a dependency has not been pre-approved, pnpm prompts us to run [`pnpm approve-builds`](https://pnpm.io/cli/approve-builds), which adds the dependency to the `onlyBuiltDependencies` array in the root `package.json`.
- `pnpm link` now adds overrides to the root `package.json`.
- SHA256 hashing
- pnpm is now capable of managing its own version (and the version of Node for that matter). This is enabled out of the box, so if the version installed doesn't match the `packageManager` version in `package.json`, pnpm installs the right version of itself before continuing.
  - While it's very promising as a way to remove the need for Volta, nvm and the like, I prefer to disable this behaviour for now ([`managePackageManagerVersions`](https://pnpm.io/settings#managepackagemanagerversions)) because `packageManager` doesn't support version ranges like `10.x` — there's an issue [about it](https://github.com/nodejs/corepack/issues/95) in Corepack. It seems silly to force-install an outdated version of pnpm just because the project hasn't updated its `packageManager` field (which is a highly impractical thing to do given how many releases of pnpm there are).
- pnpm can now check if `node_modules` is up to date before running a script or binary (`pnpm run`, `pnpm exec`).
  - I turn this option ([`verifyDepsBeforeRun`](https://pnpm.io/settings#verifydepsbeforerun)) to `"error"`, as I'm often caught by this when switching between branches.
  - I think erroring is best:
    - a warning would be easy to miss
    - we wouldn't want to show a prompt in the CI
    - installing automatically might be a waste of time, as the problem might be that we're still on an old branch/commit and haven't checked out/pulled the latest one.


.npmrc
```diff
@@ -1,14 +0,0 @@
...
-auto-install-peers=false
-strict-peer-dependencies=true
-
-# Save exact dependency versions in `package.json`
-save-prefix=""
-
-# Branch that must be checked out when publishing
-publish-branch=main
-
-scope=@h5web
```
axelboc, line 14: Turns out [`scope`](https://docs.npmjs.com/cli/v9/using-npm/config#scope) is not really relevant for us. The `publish` command just takes the package name from `package.json`.

.npmrc
```diff
@@ -1,14 +0,0 @@
...
-
-# Make installation of peer dependencies more predictible
-auto-install-peers=false
-strict-peer-dependencies=true
-
-# Save exact dependency versions in `package.json`
-save-prefix=""
-
-# Branch that must be checked out when publishing
-publish-branch=main
```
axelboc, line 12: Thought this was an npm config option but actually it's a CLI flag of [`pnpm publish`](https://pnpm.io/cli/publish#--publish-branch-branch), which already defaults to `master` or `main`. It can be set as an [option](https://pnpm.io/cli/publish#configuration) but we clearly no longer need to.

The remaining options move to `pnpm-workspace.yaml`, which means we no longer need an `.npmrc` file at all.

apps/demo/package.json
```diff
@@ -26,7 +26,7 @@
     "wouter": "3.6.0"
   },
   "devDependencies": {
-    "@esrf/eslint-config": "1.0.6",
+    "@esrf/eslint-config": "1.1.0",
```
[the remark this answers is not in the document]
axelboc: I've finally found a workaround which consists in hoisting `@types` packages to the `node_modules` folder. I've updated the PR description. I've also reverted upgrading `@esrf/eslint-config` to keep this PR focused, as it was not the culprit.

loichuder (approved):

PR #1798, opened 2025-04-22
Update lists of HDF5 files on homepage

axelboc:
We received a report that `links.h5` did not actually contain external links and virtual datasets, contrary to what was said on the homepage. This is [now corrected](https://h5web.panosc.eu/h5grove?file=links.h5).

While I was at it, I updated the lists of files. I'll comment on the changes.

### DIFF
--- a/apps/demo/src/Home.tsx
+++ b/apps/demo/src/Home.tsx
@@ -64,16 +64,11 @@ function Home() {
                 <Link to="/h5grove">
                   <strong>water_224.h5</strong>
                 </Link>{' '}
-                (default) - a typical NeXus file with various real-world
-                datasets to demonstrate H5Web’s core visualizations.
+                (default) − a typical NeXus file with various real-world
+                datasets to demonstrate H5Web’s core visualizations;
               </li>
               <li>
-                <Link to="/h5grove?file=compressed.h5">compressed.h5</Link> - a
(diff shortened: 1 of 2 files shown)


apps/demo/src/Home.tsx
```diff
@@ -200,17 +198,14 @@ function Home() {
...
                 <strong>water_224.h5</strong>
               </Link>{' '}
-              (default),{' '}
-              <Link to="/hsds?file=compressed.h5">compressed.h5</Link> (note
-              that bitshuffle is not yet supported by HSDS),{' '}
               <Link to="/hsds?file=epics.h5">epics.h5</Link>,{' '}
               <Link to="/hsds?file=grove.h5">grove.h5</Link>,{' '}
-              <Link to="/hsds?file=links.h5">links.h5</Link>,{' '}
+              <Link to="/hsds?file=sample_hsds.h5">sample_hsds.h5</Link>,{' '}
               <Link to="/hsds?file=tall.h5">tall.h5</Link>.
```
axelboc, line 209: Very few of our test files can be loaded into HSDS without errors, since they are meant to test advanced HDF5 feature and edge cases. I list only those that work, and keep HSDS' test file (`tall.h5`).

loichuder (approved):

Issue #1800, opened 2025-05-12
Provide custom AxiosInstance to H5GroveProvider


KirmTwinty:
### Is your feature request related to a problem?

The current `H5GroveProvider` only allows configuration of the Axios client through the `axiosConfig?: AxiosRequestConfig` property.
However, in some cases, it is more useful to provide a fully customized Axios instance, which may include additional features such as interceptors.

### Requested solution or feature

One simple solution can be to provide a custom derived `H5GroveApiAxiosInstance` that override the client, and then a new `H5GroveCustomProvider` (code not provided here):

``` ts
import axios, {
    AxiosInstance
}
import h5grove, {
    DataProviderApi,
    H5GroveApi
}

export class H5GroveApiAxiosInstance extends H5GroveApi {
    private readonly client: AxiosInstance;

    public constructor(
	url: string,
	filepath: string,
	axiosClient?: AxiosInstance,
    private readonly _getExportURL?: DataProviderApi['getExportURL'],
    ) {
	super(filepath);

	this.client = axiosClient
    }
}

```

### Alternatives you've considered

Directly modify the `H5GroveProvider` & `H5GroveApi`.

### Additional context

In our application, for example, we use interceptors to ensure that the current user has a valid token, refresh it when necessary, and handle other authentication-related logic.

PS: thanks for the actual developments!

axelboc:
I've got something that works in #1806 if you want to take a look @KirmTwinty — there will still be some more clean-up to do afterwards, but do let us know if that fits your needs and if you think the API and documentation makes sense. 🙏

PR #1801, opened 2025-05-15
Throw custom `AbortError` when aborting a fetch

axelboc:
I've got a working solution for removing `axios` from the providers (#1800), but it's going to take a few PRs...

This PR focuses on cleaning up the code related to aborting requests. I won't reexplain how this all works — the important bit is that, in order to show an informative error message with a "retry" button, we have to know when a request is aborted and why (e.g. manually by the user, or automatically when switching dataset).

image

To do this, we catch the error that axios (and the underlying `fetch` call) throws when a request is aborted via its [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal). We then rethrow our own error so that we have a very simple check to do in the generic `ErrorFallback` component (we definitely don't want to check if an error is an axios `CanceledError` in this component...)

In this PR, I simply introduce a new `AbortError` to make this logic even more robust and especially to remove some duplicated logic related to forwarding the "abort reason". See individual commits for more info.

### DIFF
--- a/packages/app/src/providers/utils.ts
+++ b/packages/app/src/providers/utils.ts
@@ -19,8 +19,6 @@ import { type AxiosProgressEvent, isAxiosError } from 'axios';
 
 import { type DataProviderApi } from './api';
 
-export const CANCELLED_BY_USER = 'Cancelled by user';
-
 export function typedArrayFromDType(
   dtype: DType,
 ): TypedArrayConstructor | undefined {
--- a/packages/app/src/providers/utils.ts
+++ b/packages/app/src/providers/utils.ts
@@ -102,3 +102,13 @@ export function createAxiosProgressHandler(
     })
   );
 }
+
+export class AbortError extends Error {
+  public constructor(abortSignal?: AbortSignal, cause?: unknown) {
+    const message =
+      typeof abortSignal?.reason === 'string' ? abortSignal.reason : undefined;
+
+    super(message, { cause });
+    this.name = 'AbortError';
+  }
+}
--- a/packages/app/src/providers/utils.ts
+++ b/packages/app/src/providers/utils.ts
@@ -19,7 +19,7 @@ import { type AxiosProgressEvent, isAxiosError } from 'axios';
 
 import { type DataProviderApi } from './api';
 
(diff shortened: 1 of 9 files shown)


packages/app/src/providers/utils.ts
```diff
@@ -102,3 +102,13 @@ export function createAxiosProgressHandler(
     })
   );
 }
+
+export class AbortError extends Error {
+  public constructor(abortSignal?: AbortSignal, cause?: unknown) {
+    const message =
```
loichuder, line 108: Shouldn't the message be always `CANCELLED_BY_USER` ?

loichuder (approved):

packages/app/src/providers/utils.ts
axelboc: We call `valuesStore.abortAll()` when switching entity and visualization to avoid potentially piling on unnecessary requests.

axelboc: Technically, when this happens, we evict those same abort errors from the store right away so the `ErrorFallback` is never shown – e.g.

```ts
  function onSelectPath(path: string) {
    setSelectedPath(path);
    valuesStore.abortAll('entity changed');
    valuesStore.evictErrors();
  }
```

This means I could potentially remove the check for `message === CANCELLED_BY_USER` in `ErrorFallback` and only check for `error instanceof AbortError`. But I still think it's important to store an abort reason in the `AbortError` for debugging purposes.

axelboc: I've gone ahead and simplified the check in `ErrorFallback` (and also removed the `CANCELLED_BY_USER` constant). Even if "Request cancelled - Retry?" were to be shown for other abort reasons, it would still make sense.

As it turns out, I've found a race condition: abort errors caused by switching entity or visualization are not always properly evicted from the store. I've got a solution in mind.

loichuder: > I've gone ahead and simplified the check in ErrorFallback (and also removed the CANCELLED_BY_USER constant). 

Great, that's indeed simpler I think :+1:

PR #1804, opened 2025-05-19
Refactor `H5GroveApi` to simplify internals + handle errors on all endpoints

axelboc:
Two individual commits to finish cleaning up `H5GroveApi` before I extract the `fetcher`:

1. First, I remove the private `fetch*` methods (like `fetchEntity`, `fetchValue`, etc.), as I no longer think they make the code more readable anymore. The calls to the new `fetcher` are now inlined in the main provider methods (`getEntity`, `getValue`, etc.)

I kept being annoyed at having to go up and down the file to follow what is, in the end, fairly straightforward code. This refactoring also reduces duplication a bit in the `getValue` method: I can prepare a `baseParams` object (with the `file`, `path` and `selection` parameters) and spread it in all three `fetcher` calls.

2. Second, I extract the h5grove error handling logic into a private method and use it on every endpoint.

The h5grove errors that we handle ("File not found", "No entity found at <path>", etc.) can technically occur on every endpoint. While the H5Web viewer (`App`) is likely to trigger those errors only on the initial `getEntity('/')` call, it may not be the case when consumers implement their own viewers.

### DIFF
--- a/packages/app/src/providers/h5grove/h5grove-api.ts
+++ b/packages/app/src/providers/h5grove/h5grove-api.ts
@@ -59,35 +59,7 @@ export class H5GroveApi extends DataProviderApi {
       const buffer = await this.fetcher(`${this.baseURL}/meta/`, { path });
       return parseEntity(path, toJSON(buffer) as H5GroveEntityResponse);
     } catch (error) {
-      if (!(error instanceof FetcherError)) {
-        throw error;
-      }
-
-      const payload = toJSON(error.buffer);
-      if (!isH5GroveErrorResponse(payload)) {
-        throw error;
-      }
-
-      const { message } = payload;
-      if (message.includes('File not found')) {
-        throw new Error(`File not found: '${this.filepath}'`, { cause: error });
-      }
-      if (message.includes('Permission denied')) {
-        throw new Error(
-          `Cannot read file '${this.filepath}': Permission denied`,
-          { cause: error },
-        );
-      }
-      if (message.includes('not a valid path')) {
-        throw new Error(`No entity found at ${path}`, { cause: error });
-      }
-      if (message.includes('Cannot resolve')) {
-        throw new Error(`Could not resolve soft link at ${path}`, {
-          cause: error,
-        });
-      }
-
-      throw error;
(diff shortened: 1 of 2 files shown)


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -50,50 +49,81 @@ export class H5GroveApi extends DataProviderApi {
     this.fetcher = createAxiosFetcher(
       axios.create({
         adapter: 'fetch',
-        baseURL,
```
axelboc, line 53: All `fetch` calls now concatenate `this.baseUrl` to form absolute URLs, so the axios instance no longer needs to hold on to it.

packages/app/src/providers/h5grove/models.ts
```diff
@@ -1,7 +1,6 @@
 import { type AttributeValues, type Filter } from '@h5web/shared/hdf5-models';
 
 export type H5GroveEntityResponse = H5GroveEntity;
-export type H5GroveDataResponse = unknown;
```
axelboc, line 4: I've removed this type, as it was only used as return type to `fetchValue`. Now, if I write `toJSON(...) as H5GroveDataResponse` inside the `getValue` method, ESLint says the casting is unnecessary and removes it.

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -136,91 +166,43 @@ export class H5GroveApi extends DataProviderApi {
...
     }
   }
 
-  private async fetchAttrValues(
-    path: string,
-  ): Promise<H5GroveAttrValuesResponse> {
-    const buffer = await this.fetcher(`/attr/`, { path });
-    return toJSON(buffer) as H5GroveAttrValuesResponse;
-  }
+  private getH5GroveError(error: unknown, path: string): Error | undefined {
```
loichuder, line 177: I would perhaps rename this function with another verb than `get` to distinguish it from the other `get..` functions that do fetchs

loichuder (approved):
Taking shape :muscle:

PR #1806, opened 2025-05-19
Allow consumers of `H5GroveProvider` to pass their own fetcher

### LINKED ISSUE #1800: Provide custom AxiosInstance to H5GroveProvider

axelboc:
It's finally coming together. Closes #1800.

Consumers can now provide their own fetcher and therefore their own axios instance (with custom headers, interceptors, etc.) As documented, some config options of the axios instance are ignored, like `baseURL`, `responseType`, etc.

Now that the `fetcher` is exposed through a prop on `H5GroveProvider` comes the question of whether to make it required or not. I chose to make it **optional** and to provide a basic fetcher implementation internally based on the Fetch API. My reasoning is that I don't want consumers to have to think about this `fetcher` business when they first install `@h5web/app` and try to use `H5GroveProvider` for the first time.

Technically, the only thing that the basic fetcher doesn't do that axios does is to track the progress of the requests. It's mind boggling to me how complicated it is to achieve this: https://github.com/axios/axios/blob/v1.x/lib/helpers/trackStream.js ... so I decided it wasn't worth it. What this means for users is that they see a progress bar that doesn't actually progress — e.g. with network throttling enabled to emphasize the problem:

[Screencast from 2025-05-19 11-46-36.webm](https://github.com/user-attachments/assets/a0e8297f-008f-46f7-bbd2-d53eae4c93eb)

To fix this, I now start the progress state with no value (`undefined`) and show an "indeterminate" progress bar:

[Screencast from 2025-05-19 14-09-52.webm](https://github.com/user-attachments/assets/26ff5e9c-4e83-43ba-942c-183475b4aca8)

This new behaviour is actually also useful with the axios fetcher, as the indeterminate progress bar now shows up while the connection is being established, before the first packets are received:

[Screencast from 2025-05-19 14-16-21.webm](https://github.com/user-attachments/assets/9a6f3814-5ec3-416a-8c9c-cb6053501728)

---

Next up, I'll tweak the implementation of `createAxiosFetcher` slightly so it doesn't rely on axios at all (`isAxiosError`, `isCancel`). Then, I'll implement the `fetcher` API in `HsdsProvider` and remove axios from `@h5web/app` entirely.

### DIFF
--- a/packages/app/README.md
+++ b/packages/app/README.md
@@ -54,13 +54,6 @@ function MyApp() {
 export default MyApp;
 ```
 
-> If your bundler supports it (e.g. webpack 5), you may be able to shorten the
-> stylesheet import path as follows:
->
-> ```ts
-> import '@h5web/app/styles.css';
-> ```
-
 ### Examples
 
 The following code sandboxes demonstrate how to set up and use `@h5web/app` with
@@ -191,18 +184,22 @@ The base URL of the H5Grove server.
 
 The path and/or name of the file to display in the UI.
 
-#### `axiosConfig?: AxiosRequestConfig` (optional)
+#### `fetcher?: Fetcher` (optional)
 
-By default, `H5GroveProvider` does not make any assumption as to how to
-configure its internal [axios](https://axios-http.com/docs/req_config) client.
-If you use one of H5Grove's default API implementations, then you'll need to use
-this prop to pass the `file` query parameter as shown above.
+An optional asynchronous function to fetch data and metadata from an h5grove
+back-end. The function accepts a URL together with some query parameters and
+options, and is expected to return a promise that resolves to an `ArrayBuffer`.
 
-If your API server requires authentication or is on a different domain, you'll
-need to pass the necessary request headers and configuration as well.
+If `fetcher` is not provided, `H5GroveProvider` creates one based on the native
+Fetch API using [`createBasicFetcher`](#createbasicfetcher). While this fetcher
+is sufficient to start with, it has significant limitations (e.g. it doesn't
+track requests progress, doesn't support passing authentication headers, etc.)
+For better user experience and security, we recommend using a fetcher that
+relies on a fetching library like [axios](https://axios-http.com/) (cf.
+[`createAxiosFetcher`](#createaxiosfetcher)).
 
-> Remember to memoise or extract your `axiosConfig` object so the fetching cache
-> does not get cleared if/when your app re-renders.
+> If you have to initialise the `fetcher` during render, make sure to memoise it
+> so the fetching cache does not get cleared if/when your app re-renders.
 
 #### `getExportURL?: (...args) => URL | (() => Promise<URL | Blob>) | undefined` (optional)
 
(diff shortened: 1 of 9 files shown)


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -35,28 +34,23 @@ import {
...
-        ...axiosConfig,
-      }),
-    );
   }
 
   public override async getEntity(path: string): Promise<ProvidedEntity> {
     try {
-      const buffer = await this.fetcher(`${this.baseURL}/meta/`, { path });
+      const buffer = await this.fetcher(`${this.baseURL}/meta/`, {
+        file: this.filepath,
```
axelboc, line 50: We no longer expect consumers to configure axios with the `file` parameter, so we have to specify it explicitly. (Note that I had already specified it in the `getValue` and `getAttrValues` methods in my previous PR.)

packages/app/src/vis-packs/ValueLoader.module.css
```diff
@@ -34,6 +46,25 @@
   background-color: var(--primary);
 }
 
+.progress:indeterminate::-webkit-progress-value {
+  width: 100% !important;
+  background-image: var(--indeterminate-bg);
+  background-size: 3rem 3rem;
+  animation: stripes 1s linear infinite;
```
axelboc, line 53: Full disclosure, I couldn't get this `animation` to work in Chrome (the stripes are static). I've left the code as is hoping that it's just a passing bug in Chrome... Hopefully we'll have a cross-browser way to style progress bars at some point...

packages/app/README.md
```diff
@@ -191,18 +184,19 @@ The base URL of the H5Grove server.
...
+#### `fetcher?: Fetcher` (optional)
 
-If your API server requires authentication or is on a different domain, you'll
-need to pass the necessary request headers and configuration as well.
+An optional asynchronous function to fetch data and metadata from an h5grove
+back-end. The function accepts a URL together with some query parameters and
+options, and is expected to return a promise that resolves to an `ArrayBuffer`.
 
-> Remember to memoise or extract your `axiosConfig` object so the fetching cache
-> does not get cleared if/when your app re-renders.
```
loichuder, line 205: If I understand correctly, this warning still relevant for the `fetcher`. So I think it should appear here as well.

packages/app/README.md
```diff
@@ -191,18 +184,19 @@ The base URL of the H5Grove server.
...
+If `fetcher` is not provided, `H5GroveProvider` creates one based on the native
+Fetch API using [`createBasicFetcher`](#createbasicfetcher). While this fetcher
+is sufficient to start with, it has significant limitations. If you need to pass
```
loichuder, line 195:
```suggestion
is sufficient to start with, it has significant limitations (e.g. it does not request progress tracking). A custom fetcher can be given to support, for example, authentication.
Note that H5Web provides such as a fetcher (cf. [`createAxiosFetcher`](#createaxiosfetcher)) that can be used to show request progress when fetching data.
```

packages/app/src/providers/h5grove/h5grove-api.ts
loichuder: Makes sense

packages/app/src/vis-packs/ValueLoader.module.css
loichuder: Could you remove the progress bar altogether in this case ?

loichuder (approved):

packages/app/src/vis-packs/ValueLoader.module.css
axelboc: It's just the animation; the stripes still give a sense that something is ongoing (and better than no progress bar or an empty progress bar like before, in my opinion):

[Screencast from 2025-05-20 15-36-32.webm](https://github.com/user-attachments/assets/02e1c36a-caa2-4746-bc04-61fcac165772)

PR #1807, opened 2025-05-21
Decouple axios from `HsdsProvider`

axelboc:
To follow up on #1800, I now decouple axios from `HsdsApi` like I did in `H5GroveApi`.

I worked on improving error handling in a separate commit. Every internal `fetch` method now handles HSDS' "file not found" errors, and when HSDS doesn't return valid JSON (e.g. when a compression is not supported), I no longer silence the JSON parsing error:

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/apps/demo/src/HsdsApp.tsx
+++ b/apps/demo/src/HsdsApp.tsx
@@ -1,4 +1,11 @@
-import { App, assertEnvVar, HsdsProvider } from '@h5web/app';
+import {
+  App,
+  assertEnvVar,
+  createAxiosFetcher,
+  HsdsProvider,
+} from '@h5web/app';
+import axios from 'axios';
+import { useMemo } from 'react';
 import { useSearchParams } from 'wouter';
 
 import { getFeedbackURL } from './utils';
@@ -19,13 +26,17 @@ function HsdsApp() {
   const [searchParams] = useSearchParams();
   const filepath = `${SUBDOMAIN}${searchParams.get('file') || FILEPATH}`;
 
+  const fetcher = useMemo(() => {
+    return createAxiosFetcher(
+      axios.create({
+        adapter: 'fetch',
+        auth: { username: USERNAME, password: PASSWORD },
+      }),
+    );
+  }, []);
+
   return (
-    <HsdsProvider
-      url={URL}
-      username={USERNAME}
-      password={PASSWORD}
-      filepath={filepath}
-    >
+    <HsdsProvider url={URL} filepath={filepath} fetcher={fetcher}>
       <App
         sidebarOpen={!searchParams.has('wide')}
         getFeedbackURL={getFeedbackURL}
--- a/packages/app/src/providers/hsds/HsdsProvider.tsx
+++ b/packages/app/src/providers/hsds/HsdsProvider.tsx
@@ -2,31 +2,30 @@ import { type PropsWithChildren, useMemo } from 'react';
 
 import { type DataProviderApi } from '../api';
 import DataProvider from '../DataProvider';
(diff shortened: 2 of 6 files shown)


packages/app/src/providers/hsds/HsdsProvider.tsx
```diff
@@ -2,31 +2,30 @@ import { type PropsWithChildren, useMemo } from 'react';
...
+import { type Fetcher } from '../models';
 import { HsdsApi } from './hsds-api';
 
 interface Props {
   url: string;
-  username: string;
-  password: string;
   filepath: string;
   resetKeys?: unknown[];
+  fetcher: Fetcher;
```
axelboc, line 12: Note that the fetcher is required. HSDS requires authentication, so we can't initialise a basic fetcher like in `H5GroveProvider`.

packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -51,46 +50,20 @@ import {
...
-        } catch {
-          // https://github.com/HDFGroup/hsds/issues/87
-          try {
-            return JSON.parse(
-              data.replaceAll(/-?Infinity|NaN/gu, '"$&"'),
-            ) as unknown;
-          } catch {
-            return data;
-          }
-        }
```
axelboc, line 91: Moved to a new utility called `toExtendedJSON`.

packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -202,78 +175,99 @@ export class HsdsApi extends DataProviderApi {
...
-          signal: abortSignal,
-          onDownloadProgress: createAxiosProgressHandler(onProgress),
+          domain: this.filepath,
+          ...(selection && { select: `[${selection}]` }),
         },
+        { abortSignal, onProgress },
       );
 
-      return data.value;
+      return (toExtendedJSON(buffer) as HsdsValueResponse).value;
```
axelboc, line 268: Note that I call `toExtendedJSON` only when it makes sense to do so — i.e. when fetching dataset and attribute values. Otherwise, I call the more basic `toJSON`.

packages/app/src/test-utils.tsx
```diff
@@ -147,7 +147,7 @@ export async function assertListeningAt(
 ) {
   try {
     await fetch(url);
-  } catch {
-    throw new Error(message);
+  } catch (error) {
+    throw new Error(message, { cause: error });
```
axelboc, line 151: Unrelated change. This just improves error logging in the API tests.

axelboc (commented):
I'll document the new `fetcher` prop of `HsdsProvider` in a separate PR, as I'm planning to improve `createBasicFetcher` at the same time.

apps/demo/src/HsdsApp.tsx
```diff
@@ -19,13 +26,17 @@ function HsdsApp() {
   const [searchParams] = useSearchParams();
   const filepath = `${SUBDOMAIN}${searchParams.get('file') || FILEPATH}`;
 
+  const fetcher = useMemo(() => {
+    return createAxiosFetcher(
+      axios.create({
+        adapter: 'fetch',
+        auth: { username: USERNAME, password: PASSWORD },
```
axelboc, line 33: I had to keep the initialization of the `fetcher` inside the component. If I move it out, I have to move two of the `assertEnvVar` calls out as well, which then forces the `VITE_HSDS_USERNAME` and `VITE_HSDS_PASSWORD` env vars to be provided, which is annoying when running tests or the demo for the first time.

loichuder (approved):
:heart:

PR #1809, opened 2025-05-21
Make axios a peer/dev dep in app package

axelboc:
I was planning on removing axios from the app package completely, but we need to keep it if we want `createAxiosFetcher` to be properly typed and type-checked.

So in the end, I turn it into a dev dependency and add it as an optional peer dependency. And to make sure this doesn't cause issues in consumer packages that don't install axios, I remove the two remaining non-type imports of `axios` from the app package (`isAxiosError` and `isCancel`).

### DIFF
--- a/packages/app/src/providers/utils.ts
+++ b/packages/app/src/providers/utils.ts
@@ -15,7 +15,7 @@ import {
   type BigIntTypedArrayConstructor,
   type TypedArrayConstructor,
 } from '@h5web/shared/vis-models';
-import { type AxiosInstance, isAxiosError, isCancel } from 'axios';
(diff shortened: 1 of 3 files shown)


packages/app/src/providers/utils.ts
```diff
@@ -142,7 +142,7 @@ export function createAxiosFetcher(axiosInstance: AxiosInstance): Fetcher {
     const { abortSignal, onProgress } = opts;
 
     try {
-      const { data } = await axiosInstance.get<ArrayBuffer>(url, {
```
axelboc, line 145: Making the implicit `ArrayBuffer` type cast explicit below.

PR #1810, opened 2025-05-22
Fix abort error handling

axelboc:
This fixes an issue with the basic fetcher when requests are aborted (via `abortSignal.abort(<reason>)`).

In the h5grove demo, where the basic fetcher is configured without any headers, the browser throws a `DOMException`. But in the HSDS demo, where we pass an `Authentication` header for basic auth, the browser throws the abort reason as is (i.e. the reason string itself). Our current error handling logic expects a `DOMException`, so in the HSDS demo, we just see "Unknown error".

Both Firefox and Chrome behave the same way, so there must be an explanation but I could not find it. :disappointed: 

The fix is to simply look at the `abortSignal.aborted` flag in the `catch` block... It's basically what axios does as well (though in a much more convoluted way).

Additionally, I realised that `abortSignal.abort()` is meant to be called with an `Error`, not a `string`. So I now pass `AbortError` directly, and then I simply rethrow it from the fetcher with `throw abortSignal.reason`.

All in all, it cleans things up very nicely.

### DIFF
--- a/packages/app/src/providers/mock/mock-api.ts
+++ b/packages/app/src/providers/mock/mock-api.ts
@@ -67,7 +67,7 @@ export class MockApi extends DataProviderApi {
       throw new Error('error');
     }
 
-    if (dataset.name.startsWith('slow')) {
+    if (dataset.name.startsWith('slow') && abortSignal) {
       await cancellableDelay(abortSignal);
     }
 
--- a/packages/app/src/providers/mock/utils.ts
+++ b/packages/app/src/providers/mock/utils.ts
@@ -17,7 +17,6 @@ import { getChildEntity } from '@h5web/shared/hdf5-utils';
 import ndarray from 'ndarray';
 
 import { applyMapping } from '../../vis-packs/core/utils';
-import { AbortError } from '../utils';
 
 export const SLOW_TIMEOUT = 3000;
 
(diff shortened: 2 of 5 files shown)


packages/app/src/providers/mock/mock-api.ts
```diff
@@ -67,7 +67,7 @@ export class MockApi extends DataProviderApi {
       throw new Error('error');
     }
 
-    if (dataset.name.startsWith('slow')) {
+    if (dataset.name.startsWith('slow') && abortSignal) {
```
axelboc, line 70: I've made the `abortSignal` required in `cancellableDelay` to simplify the code. In practice, we never request the `'slow_value'` dataset without an abort signal (it's optional to simplify testing the h5grove and h5wasm APIs).

packages/app/src/providers/mock/utils.ts
```diff
@@ -82,20 +81,20 @@ export function getChildrenPaths(
...
+      abortSignal.removeEventListener('abort', handleAbort);
       resolve();
     }, SLOW_TIMEOUT);
 
     function handleAbort() {
       clearTimeout(timeout);
-      abortSignal?.removeEventListener('abort', handleAbort);
-      reject(new AbortError(abortSignal));
+      abortSignal.removeEventListener('abort', handleAbort);
+      reject(abortSignal.reason); // eslint-disable-line @typescript-eslint/prefer-promise-reject-errors
```
axelboc, line 95: It's the responsibility of the code that calls `abortSignal.abort()` to pass an error (i.e. the suspense fetch store).

packages/app/src/providers/utils.ts
```diff
@@ -124,8 +124,8 @@ export function createBasicFetcher(
 
       throw new FetcherError(response.status, response.statusText, buffer);
     } catch (error) {
-      if (error instanceof DOMException && error.name === 'AbortError') {
-        throw new AbortError(abortSignal, error);
+      if (abortSignal?.aborted && abortSignal.reason instanceof AbortError) {
+        throw abortSignal.reason;
```
axelboc, line 128: As explained, we have access to the `abortSignal`, so we can just check for `abortSignal.aborted`!

The `instanceof` check is just in case the signal is aborted in an unexpected way (i.e. not by the suspense fetch store).

PR #1815, opened 2025-06-06
Add linewidth parameter to DataCurve

CPrescher:
This tries to add a linewidth parameter to the DataCurve.

It works for values of 1 to 6 inside of chrome, but does not work for larger values and firefox completely ignores it. Looking into the LineBasicMaterials [docs](https://threejs.org/docs/#api/en/materials/LineBasicMaterial), it is clear, that a reliable implementation would need to use [Line2](https://threejs.org/docs/index.html#examples/en/lines/Line2). 

However, this would be a large change.

I would be happy, with it at least working correctly in chrome.

What is your take on this?

### DIFF
--- a/packages/lib/src/vis/line/Line.tsx
+++ b/packages/lib/src/vis/line/Line.tsx
@@ -8,7 +8,6 @@ import { hasR3FEventHandlers } from '../utils';
 import LineGeometry from './lineGeometry';
 
 // Alias Three's `Line` to `Line_` to avoid conflict with SVG `line` in JSX
-// https://docs.pmnd.rs/react-three-fiber/tutorials/typescript#extending-threeelements
 class Line_ extends R3FLine {}
 extend({ Line_ });
 declare module '@react-three/fiber' {
@@ -21,6 +20,7 @@ interface Props extends Object3DNode<R3FLine, typeof R3FLine> {
   abscissas: NumArray;
   ordinates: NumArray;
   color: string;
+  lineWidth?: number;
   visible?: boolean;
   ignoreValue?: IgnoreValue;
 }
@@ -30,6 +30,7 @@ function Line(props: Props) {
     abscissas,
     ordinates,
     color,
+    lineWidth = 1,
     visible = true,
     ignoreValue,
     ...lineProps
@@ -55,7 +56,7 @@ function Line(props: Props) {
 
   return (
     <line_ geometry={geometry} visible={visible} {...lineProps}>
-      <lineBasicMaterial color={color} />
+      <lineBasicMaterial color={color} linewidth={lineWidth} />
     </line_>
   );
 }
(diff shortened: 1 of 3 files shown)


loichuder:
> It works for values of 1 to 6 inside of chrome, but does not work for larger values and firefox completely ignores it. 

Yes, I remember that we pursed this but decided the result was not worth the effort at the time. For some reason, I can't find the related discussion/issue/MR again.

> What is your take on this?

Personally, in essence, I have nothing against having a `linewidth` parameter for the low-level components. Or even moving to `Line2` if it does not raise additional problems (in terms of implementation or performance)

@axelboc ?

packages/lib/src/vis/line/Line.tsx
```diff
@@ -53,10 +57,26 @@ function Line(props: Props) {
...
+  const line2Material = new LineMaterial({
+    color: color,
+    linewidth: lineWidth,
+    resolution: new Vector2(size.width, size.height),
+  });
+
   return (
-    <line_ geometry={geometry} visible={visible} {...lineProps}>
-      <lineBasicMaterial color={color} />
-    </line_>
```
axelboc, line 59: I would have thought that you'd only have to change `line_` to `line2` and the material to `lineMaterial` (with additional props). The buffer geometry generated by `useGeometry` should work as it, no?

CPrescher: well, i am not used to work with three, but just replacing with geometry generated by useGeometry gave the following error:

Type 'H5WebGeometry<"position", Params>' is missing the following properties from type 'LineGeometry': isLineGeometry, fromLine, isLineSegmentsGeometry, setPositions, and 7 more.ts(2740)

axelboc: Right okay, it's a typing issue but looking into it more, `Line2` expects a buffer geometry with **line segments**, so I'm gonna have to write a new `LineSegmentsGeometry` compatible with our [`useGeometry` hook](https://h5web-docs.panosc.eu/?path=/docs/utilities--docs#usegeometry). (ThreeJS' geometries are not very optimised, they tend to loop over the data points multiple times and create various intermediary arrays and buffers in the process, which isn't great when dealing with large amounts of data)

axelboc:
I think we should keep the `Line` component with the default ThreeJS `Line` and develop a separate `SegmentedLine` (or whatever) based on `Line2`. The `Line` component will still be slightly more performant and memory efficient than `SegmentedLine`, so I think it's worth keeping it around (especially for the viewer).

Since I may not have time to work on this `SegmentedLine` (and `LineSegmentsGeometry`) for a while, may I suggest that we focus this pull request on adding a `materialProps` prop to `Line` and `DataCurve`? This way you'll still have a solution that works in Chrome if you like.

Might be worth also opening an issue to track the work on `SegmentedLine` if you feel like it (otherwise I'll do it, no worries).

PR #1816, opened 2025-06-06
add materialProps as properties for Line and DataCurve

CPrescher:
adds materialsProps to Line and DataCurve

this allows to pass extra information like linewidth or opacity to the line creation process.

e.g. 

materialsProbs = { linewidth: 3}

This PR is result of discussion in #1815


axelboc:
Published in `15.0.0-beta.1`

PR #1820, opened 2025-06-13
Extract and test dim mapping state initialisation

axelboc:
I'm planning to move the dimension mapper and related utilities/styles/models to the library, because I'd like to use them in Braggy. As a first step, I extract the dim mapping state intialisation into a dedicated utility function.

While I'm at it, I add support for initialising a dim mapping state with 0 axes (for slicing scalar values, cf. my attempt in https://github.com/silx-kit/h5web/pull/1768/files#diff-123aa107dfdd1a6159a9a847361690b69a161a6551acf1c537ad0239eaa6699f). It's not needed yet, but it actually  makes the new `initDimMapping` function more robust, IMO. I also add some unit tests to cover all edge cases.

### DIFF
--- a/packages/app/src/dimension-mapper/store.tsx
+++ b/packages/app/src/dimension-mapper/store.tsx
@@ -10,6 +10,7 @@ import { createStore, type StoreApi, useStore } from 'zustand';
 
 import { areSameDims } from '../vis-packs/nexus/utils';
 import { type DimensionMapping } from './models';
+import { initDimMapping } from './utils';
 
 interface DimMappingState {
   dims: ArrayShape;
@@ -23,7 +24,7 @@ interface DimMappingState {
   ) => void;
 }
 
-function createLineConfigStore() {
+function createDimMappingStore() {
   return createStore<DimMappingState>((set) => ({
     dims: [],
     axesCount: 0,
@@ -41,7 +42,7 @@ interface Props {}
 export function DimMappingProvider(props: PropsWithChildren<Props>) {
   const { children } = props;
 
-  const [store] = useState(createLineConfigStore);
+  const [store] = useState(createDimMappingStore);
 
   return (
     <StoreContext.Provider value={store}>{children}</StoreContext.Provider>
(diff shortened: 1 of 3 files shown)


packages/app/src/dimension-mapper/utils.ts
```diff
@@ -2,10 +2,31 @@ import { type Axis } from '@h5web/shared/vis-models';
...
+  if (axesCount < 0 || axesCount > 2) {
+    throw new RangeError('Expected 0, 1 or 2 axes');
+  }
+
+  // Cap number of axes to number of dimensions
+  const safeAxesCount = Math.min(axesCount, dims.length);
+
+  return [
+    ...Array.from({ length: dims.length - safeAxesCount }, () => 0),
+    ...(safeAxesCount > 0
```
axelboc, line 24: This check used to be `dims.length > 0` and we weren't capping the number of axes to the number of dimensions, so we could end up with the following invalid cases:

- `initDimMapping([], 2)` => `['y', 'x']` even though there are 0 dimensions
- `initDimMapping([1], 2)` => `['y', 'x']` even though there is only 1 dimension
- `initDimMapping([1, 1], 0)` => `['y', 'x']` even though we're specifying that we want a mapping for 0 axes...

Note that I decided to cap the number of axes instead of throwing an error because it's useful for instance in the _Compound_ visualization, which supports both scalar and nD compound datasets. It saves having to compute the right number of axis in the container.

packages/app/src/dimension-mapper/store.tsx
```diff
@@ -23,7 +24,7 @@ interface DimMappingState {
   ) => void;
 }
 
-function createLineConfigStore() {
```
loichuder, line 26: ??? :joy_cat:

packages/app/src/dimension-mapper/utils.ts
loichuder: > because it's useful for instance in the Compound visualization, which supports both scalar and nD compound datasets. It saves having to compute the right number of axis in the container.

Could you elaborate on that ?

axelboc: https://github.com/silx-kit/h5web/blob/5d323afecadf673dbd803632fee4bfc25d3dda97/packages/app/src/vis-packs/core/compound/CompoundVisContainer.tsx#L26

Above, `dims` can be:
- `[]` for scalar compound datasets, in which case we want `[]` as dim mapping (and therefore an `undefined` fetch selection to fetch the whole dataset — i.e. the scalar value);
- `[<dim1>]` for 1D compound dataset, in which case we want `['x']` as dim mapping (and therefore `':'` as fetch selection);
- `[<dim1>, <dim2>, ...]` for nD compound datasets, in which case we want `[..., 0, 'x']` as dim mapping (and therefore `'...,0,:'` as fetch selection).

If I threw instead of capping the number of axes in `initDimMapping`, I'd have to write:

```ts
const [dimMapping, setDimMapping] = useDimMappingState(dims, dims.length === 0 ? 0 : 1); 
// or
const [dimMapping, setDimMapping] = useDimMappingState(dims, Math.min(dims.length, 1));
```

So it's just a bit of flexibility for convenience. Since I couldn't think of a negative effect of the capping behaviour, I went for it.

PR #1825, opened 2025-06-19
Allow customising dimension mapper with CSS variables

axelboc:
Per our convention, any component inside the library has to:

- be customisable through CSS variables prefixed with `--h5w`;
- look decent by default, when the viewer's CSS variables (`--primary-bg`, etc.) are not defined.

So in this PR, I define a bunch of CSS variables for the dimension mapper and give them default values, which are consistent with the ones used for the domain widget, notably. I also make sure all font sizes are relative rather than absolute, and I add a bit more horizontal padding to the thumbs, for when the numbers they contain are large..

The screenshots need updating mostly because of the very slightly different font sizes.


packages/lib/src/dimension-mapper/DimensionMapper.module.css
```diff
@@ -2,9 +2,8 @@
   display: flex;
   flex-direction: column;
   padding: 1rem 0.375rem 1rem 0.375rem;
-  background-color: var(--primary-light-bg);
+  background-color: var(--h5w-dimMapper--bgColor, ghostwhite);
   overflow: hidden auto;
-  font-size: 0.875em;
```
axelboc, line 7: This was making the switch to relative font sizes in `SlicingSlider.module.css` a bit tricky and led to awkward `em` values. I move it to `axisMapper` so I can adjust the font size of the slider elements more granularly.

packages/lib/src/dimension-mapper/SlicingSlider.module.css
```diff
@@ -3,19 +3,22 @@
   display: flex;
   flex-direction: column;
   padding: 0 0.5rem;
-  --thumb-height: 1.5rem;
-  --mark-height: 0.5rem;
```
axelboc, line 7: I considered making those customisable as well but changing those values signficantly caused some alignment issues. Didn't seem worth digging into.

packages/lib/src/dimension-mapper/SlicingSlider.module.css
```diff
@@ -29,38 +32,34 @@
...
   align-items: center;
   justify-content: center;
-  padding: 0.125rem;
-  background-color: var(--primary-bg);
-  border: 2px solid var(--secondary);
-  border-radius: 1.5em;
-  height: var(--thumb-height);
-  min-width: 1.5rem;
-  font-size: 0.625rem;
+  padding: 0.125rem 0.25rem;
```
axelboc, line 44: Increasing horizontal padding inside the thumb a tiny bit.

packages/lib/src/dimension-mapper/SlicingSlider.module.css
```diff
@@ -29,38 +32,34 @@
...
 
-.thumb:global(.active) {
-  border-color: var(--secondary-dark);
-}
-
-.thumb:active {
-  cursor: none;
+.thumbValue {
+  font-size: 0.625em;
 }
```
axelboc, line 55: The `active` class was no longer being applied (probably a regression from a long time ago). I chose not to fix it because `react-slider` adds this class when the thumb has focus, so the class remains after you've interacted with the thumb (until you click somewhere else) and it looks weird if the thumb keeps a different color after you're done interacting with it...

packages/lib/src/dimension-mapper/SlicingSlider.module.css
loichuder: And therefore triggering a cascade of new screenshots :joy_cat:

loichuder (approved):

PR #1829, opened 2025-06-23
Document and test dimension mapping utilities

axelboc:
Final PR about moving the dimension mapper and related utilities to the lib.

- Extract and expose a non-memoised utility from `useSlicedDimsAndMapping`: `getSlicedDimsAndMapping`.
- Document `initDimMapping`, `getSliceSelection`, `getSlicedDimsAndMapping` (and `useSlicedDimsAndMapping`)
- Unit test `getSliceSelection` and `getSlicedDimsAndMapping` (I already wrote tests for `initDimMapping` in a previous PR)

### DIFF
--- a/packages/lib/src/dimension-mapper/AxisMapper.tsx
+++ b/packages/lib/src/dimension-mapper/AxisMapper.tsx
@@ -1,9 +1,8 @@
 import { ToggleGroup } from '@h5web/lib';
 import { type AxisMapping } from '@h5web/shared/nexus-models';
-import { type Axis } from '@h5web/shared/vis-models';
+import { type Axis, type DimensionMapping } from '@h5web/shared/vis-models';
 
 import styles from './DimensionMapper.module.css';
-import { type DimensionMapping } from './models';
 
 interface Props {
   axis: Axis;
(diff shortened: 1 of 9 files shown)


packages/lib/src/dimension-mapper/AxisMapper.tsx
```diff
@@ -1,9 +1,8 @@
 import { ToggleGroup } from '@h5web/lib';
 import { type AxisMapping } from '@h5web/shared/nexus-models';
-import { type Axis } from '@h5web/shared/vis-models';
+import { type Axis, type DimensionMapping } from '@h5web/shared/vis-models';
 
 import styles from './DimensionMapper.module.css';
-import { type DimensionMapping } from './models';
```
axelboc, line 6: I've moved the `DimensionMapping` type to the shared package. It seemed more logical and it was lonely in a file all by itself... :grin:

packages/lib/src/dimension-mapper/DimensionMapper.tsx
```diff
@@ -1,14 +1,13 @@
...
 import { type HTMLAttributes } from 'react';
 
 import AxisMapper from './AxisMapper';
 import styles from './DimensionMapper.module.css';
-import { type DimensionMapping } from './models';
 import SlicingSlider from './SlicingSlider';
 
 interface Props extends Omit<HTMLAttributes<HTMLDivElement>, 'onChange'> {
   dims: number[];
-  dimHints?: AxisMapping<string>;
```
axelboc, line 11: `AxisMapping` was created as a NeXus concept (it's in `nexus-models.ts` in the shared package). Now that the `DimensionMapper` is exposed, I prefer to remove this subtle coupling with NeXus. It makes the type of the `dimHints` prop more explicit at the same time.

loichuder (approved):

PR #1830, opened 2025-06-25
Support right mouse button and preventing default context menu

### LINKED ISSUE #1813: @h5web/lib limitations on mouse buttons and modifier keys
Hi,

would it be possible, to also add the following interactions as possibilities:

- right mouse button as a choice? (disabling the right click context menu of course)
- have 'space' as modifier key 

or are there specific reasons why this is currently limited in the chosen way?

Best wishes,
Clemens

axelboc:
Fix #1813

- Interactions like `Pan`, `SelectToZoom`, etc. can now be configured with the right mouse button.
- A new component called `PreventDefaultContextMenu` prevents the browser from showing its native context menu on right-click. Prop `when` controls this behaviour:
  - with `'as-needed'` (the default), the context menu is prevented only when an interaction based on the right mouse button is registered and enabled;
  - with `'always'`, the context menu is always prevented regardless of registered interactions;
  - with `'never'`, the context menu is never prevented regardless of registered interactions.
- In order for `PreventDefaultContextMenu` to be able to access registered interactions, I expose a `getInteractions` function through the existing `InteractionsContext`. It returns all registered interactions by default, but it can also filter them when given a specific mouse button/modifier key.
- I add `PreventDefaultContextMenu` to `DefaultInteractions` so it works out of the box in all high-level visualization components (`LineVis`, etc.).
- In the Storybook, I document `InteractionsContext`, `getInteractions` and `PreventDefaultContextMenu`, and I add a `Pan` story with the right mouse button.

[Screencast from 2025-06-25 14-49-38.webm](https://github.com/user-attachments/assets/94e10145-5929-41d5-b9aa-adc0920ffb0d)

### DIFF
--- a/packages/lib/src/interactions/InteractionsProvider.tsx
+++ b/packages/lib/src/interactions/InteractionsProvider.tsx
@@ -1,3 +1,4 @@
+import { castArray } from '@h5web/shared/vis-utils';
 import {
   createContext,
   type ReactNode,
@@ -12,6 +13,10 @@ import { type InteractionConfig } from './models';
 export interface InteractionsContextValue {
   registerInteraction: (id: string, config: InteractionConfig) => void;
   unregisterInteraction: (id: string) => void;
+  getInteractions: (
+    button?: InteractionConfig['button'],
+    modifierKey?: InteractionConfig['modifierKey'],
+  ) => Interaction[];
   shouldInteract: (id: string, event: MouseEvent) => boolean;
 }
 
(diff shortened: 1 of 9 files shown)


packages/lib/src/interactions/InteractionsProvider.tsx
```diff
@@ -38,14 +43,41 @@ function InteractionsProvider(props: { children: ReactNode }) {
...
+    ): Interaction[] => {
+      const interactions = [...interactionMap.values()].filter(
+        (inter) => inter.isEnabled,
+      );
+
+      if (button === undefined) {
+        return interactions;
+      }
+
+      const buttons = button === 'Wheel' ? button : castArray(button);
```
loichuder: Sounds like `button === 'Wheel` should be treated separately:

```ts
if(button === 'Wheel') {
  return interactions.filter(
        (inter) => inter.isWheel && modifierKeys.every((k) => inter.modifierKeys.includes(k)),
      );
}
```

loichuder (approved):
Well thought-out :clap:

PR #1831, opened 2025-06-26
Make N+1 bin optional in histogram

axelboc:
It's the same issue we had with the heatmap vis: the x axis needs to go up to the right-hand edge of the last bin; otherwise we can't compute the width of the last bar and we get a `NaN` width warning in the browser console.

This PR makes the extra value in the `bins` array optional. If it's not present, we assume we want the same step value as between the last and second last bins.

### DIFF
--- a/packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
+++ b/packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
@@ -4,7 +4,7 @@ import { AxisBottom, AxisLeft } from '@visx/axis';
 import { scaleLinear } from '@visx/scale';
 import { type ReactNode } from 'react';
 
-import { useSafeDomain } from '../../../vis/heatmap/hooks';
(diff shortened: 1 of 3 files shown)


packages/lib/src/toolbar/controls/Histogram/Histogram.tsx
```diff
@@ -41,6 +41,7 @@ function Histogram(props: Props) {
   } = props;
   const { colorMap, invertColorMap } = props;
 
+  const bins = usePixelEdgeValues(rawBins, values.length);
```
axelboc, line 44: Perhaps no longer the best name for this hook :sweat_smile:

loichuder:
Sorry, I don't get why it should be optional from the description of the issue:

> the x axis needs to go up to the right-hand edge of the last bin; otherwise we can't compute the width of the last bar and we get a NaN width warning in the browser console.

Sounds like you **do** want the last bin, then ? How making it optional helps ?

PR #1837, opened 2025-08-07
fix zoomOnSelection with mouse dragging outside of the graph

CPrescher:
Hi,

this pull requests fixes an issue when using the SelectToZoom interaction. It was behaving erratically when the mouse was dragged beyond the visible graphics boundaries (e.g. outside of the axes)

The attempt here clamps the box to the visible Sizeand results in correct zooming behavior even at the boundaries of the graph.

Best wishes,
Clemens


CPrescher:
I apologize, I will go back to drawing board, I just realized, that this is not the solution.

It seems to be related to something else....

Issue #1727, opened 2025-01-13
Dimension mapping strategy


axelboc:
So something like `<App dimMappingStrategy="first" />` (with the default strategy being `"last"` and the typing being `type DimMappingStrategy = 'first' | 'last';`). I don't think we need anything more complicated, at least for your use case, right?

_Originally posted by @axelboc in https://github.com/silx-kit/h5web/discussions/1502#discussioncomment-7325510_

PR #1840, opened 2025-08-14
Improve handling of RGB dataset dimensions

### LINKED ISSUE #1838: Truncated selection prevents fetching RGB stack with HSDS
### Describe the bug

First identified in https://github.com/silx-kit/h5whale/issues/9#issuecomment-2940186858

When fetching an RGB image from a stack, H5Web forgets to specify the last dimension (the one with the three channels) in the selection string. With h5grove and h5wasm this doesn't cause an issue but with HSDS this fails with an error "400: Invalid Selection".

This doesn't occur when fetching a single RGB image because no selection string is passed in that case (the entire dataset is fetched).

### To Reproduce

1. https://h5web.panosc.eu/hsds?file=rgb_stack.h5
2. `/OutputDeck/Camera_1/Images/Images`
3. Error "400: Invalid Selection" when requesting slice with selection `"0,:,:"` instead of `0,:,:,:`

### Expected behaviour

The selection string should include the final RGB channel dimension for correctness and compatibility with HSDS.

### Screenshots

Image

Note that the displayed error doesn't match the error received from the server. This will need to be corrected as well.

axelboc:
Fix #1838 

This PR fixes the issue above by now sending a complete selection string for RGB stacks with all providers, therefore moving past the fetching error in HSDS.

This is done by making a more general improvement to how dimensions of RGB datasets are handled: when the _RGB_ vis is selected, the **last dimension** (the one with the channels) is now **visible in the dimension mapper** but locked so it cannot be mapped to an axis:

image

[Screencast from 2025-08-14 09-29-05.webm](https://github.com/user-attachments/assets/1f38207b-74aa-423d-b49f-6a3f181c7500)

---

I've done this by introducing the concept of "locked dimensions" to the dimension mapper. When a dimension is locked, it is marked as `null` in the dim mapping state.


packages/app/src/dim-mapping-store.tsx
```diff
@@ -51,18 +54,23 @@ export function DimMappingProvider(props: PropsWithChildren<Props>) {
 export function useDimMappingState(
   dims: number[],
   axesCount: number,
+  lockedDimsCount = 0,
```
axelboc, line 57: This is the entry point for vis containers to specify a number of locked dimensions. It is passed down to `initDimMapping`. The other changes in this file are to ensure that the dim mapping is reset when switching between the _RGB_ vis and the _Heatmap_ or _Matrix_ vis for instance (i.e. so that a locked dimension is never mapped to an axis).

packages/lib/src/dimension-mapper/utils.ts
```diff
@@ -27,20 +38,18 @@ export function getSliceSelection(
...
 
-  // Create slice selection string from dim mapping - e.g. [0, 'y', 'x'] => "0,:,:"
-  return dimMapping.map((dim) => (isAxis(dim) ? ':' : dim)).join(',');
-}
-
-export function isAxis(elem: number | Axis): elem is Axis {
-  return typeof elem !== 'number';
+  // Create slice selection string from dim mapping - e.g. [0, 'y', 'x', null] => "0,:,:,:"
+  return dimMapping
+    .map((dim) => (typeof dim === 'number' ? dim : ':'))
```
axelboc, line 43: Here and below, I'm just inverting the logic: instead of looking for axes (i.e. `'x'` and `'y'`), I now look for numbers (i.e. sliceable dimensions) — it's simpler than to add a check for `null` after `isAxis(dim)`.

axelboc (commented):
The screenshot and recording were actually taken with h5grove because I have another issue with HSDS: the RGB images are now properly fetched with full selection strings, but they're all black because I'm only getting zeros... I'm not sure if it's specific to the file I'm testing with (cf. #1838) or something in HSDS. We're not on the latest version, so I'm hoping @Flydroid won't get the issue. Otherwise, I'll open a dedicated issue.

Note that the mandrill image works fine with HSDS:

image

packages/lib/src/dimension-mapper/utils.ts
```diff
@@ -1,21 +1,32 @@
-import { type Axis, type DimensionMapping } from '@h5web/shared/vis-models';
+import { type DimensionMapping } from '@h5web/shared/vis-models';
 
 export function initDimMapping(
   dims: number[],
   axesCount: number,
+  lockedDimsCount = 0,
```
loichuder, line 6: So the dims can only be locked starting from the deepest ones ?

axelboc: Yep, same as the fact that the axes are always mapped to the deepest dimensions. There's an issue open to make that configurable, which will likely require giving `initDimMapping` a more complex signature: #1727. We can deal with locked dimensions as part of this future work.

Note that the RGB visualization currently expects the last dimension to be the channels dimension (otherwise the vis doesn't show up at all), and the new locked dimension feature is used only by the RGB vis for now, so I think the decision to always lock the deepest dimensions is safe and sufficient for a first implementation.

loichuder: Ok, just making sure that I understood well :wink:

loichuder (approved):

Issue #1834, opened 2025-07-25
Mixing complex and real data in the same NXGroup prevents visualisating the data


loichuder:
### Describe the bug

When multiple data fields, real and complex, are placed in the same NXdata group (as `signal` and `auxiliary_signals`), an error “Expected dataset to have numeric, boolean or enum type” is raised at the group level.

The signal attribute is not taken into account and the default image is not displayed when the NXdata group is selected. 

However the data are correctly displayed , with correct axes, when the individual data fields are selected.

### To Reproduce

Load this file in the demo: https://h5web.panosc.eu/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Ffile_fail.nxs

The error should appear when opening the file and the NXGroup contained inside

### Expected behaviour

The vis should allow to choose between each real signal and complex signal component. Alike what is displayed when not mixing real and complex signals: https://h5web.panosc.eu/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Ffile_ok.nxs

### Context
- Version: v14
- H5Web context: demo, myHDF5, VSCode

axelboc:
Just to confirm: we're talking about the _NX Heatmap_ visualization, right? So having the radio buttons visible even when mixing real and complex signal and auxiliaries. And then the phase/amplitude drop-down in the toolbar would appear and disappear (or get enabled/disabled) depending on the selected signal?

It's tricky because complex datasets require special handling (fetching complex values, rendering the complex toolbar with the phase/amplitude drop-down, etc.) We rely on the type of the signal to decide whether to render `NxImageContainer` and `NxComplexImageContainer`, so once that decision is made, it's tricky to mix things up further down the React tree.

We would need to maybe merge `NxComplexImageContainer` into a single `NxImageContainer` but that would complexify things a bit. Not saying it's not doable.

PR #1842, opened 2025-08-18
[NX Line] Support NXdata group with mixed real and complex signals

axelboc:
Part 1 to resolve issue #1834 

I'm starting with the _Line_ and _NX Line_ visualisations. Since the idea is to plot both real and complex signals together, it's actually the most challenging case and I wanted to make sure I could find a solution for it.

Working example with [file_fail.nxs](https://h5web.panosc.eu/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Ffile_fail.nxs):

[Screencast from 2025-08-18 16-42-11.webm](https://github.com/user-attachments/assets/88ba1988-d948-4df3-8eab-683d55340555)

> Note that the _NX Line_ visualisation has to be the "preferred" visusalisation for the file to work (since _NX Heatmap_ still throws an error).

---

### Implementation

The core idea is to merge all the relevant complex line vis components (container, mapped, toolbar, config) into the main line vis components by expanding the latter's supported dtypes from `NumericLikeType` to `NumericLikeType | ComplexType`.

> I considered adding `ComplexType` to the `NumericLikeType` union, but I'm not yet sure if it's the right call... It's on my mind, though.

1. The changes start in the core and NeXus visualization definitions, where I now direct complex datasets to `LineVisContainer` (instead of `ComplexLineVisContainer`), and `NXdata` groups with a complex signal to `NxSpectrumContainer` (instead of `NxComplexSpectrumContainer`).
2. In `LineVisContainer` and `NxSpectrumContainer`, I let complex datasets, signals and auxiliaries go through the various assertions (`assertNumericLikeOrComplexType` instead of just `assertNumericLikeType`).
3. In the line toolbar config, I bring in the `complexVisType` state, which was previously part of the complex line config.
4. In `MappedLineVis`, I type various props with `NumericLikeType | ComplexType` and I change the implementation of `useToNumArray(s)` to support complex arrays.
 
This last change is key to keeping changes to a minimum in `MappedLineVis`. The `useToNumArray(s)` hooks now receive `complexVisType` as parameter and compute the amplitude or phase values (accordingly) of any complex input arrays they receive. This means that they still return `NumArray`(s) just as they did before, regardless of whether the input arrays are real or complex, allowing the rest of the `MappedLineVis` component to remain unchanged.

The only minor downside to this implementation is that the amplitude/phase values of complex arrays are now recomputed on demand, whenever the user switches from _Amplitude_ to _Phase_ or back in the toolbar. Without this small compromise, the code of `MappedLineVis` quickly became littered with complicated, unreadable `useMemo`s... I think it is an acceptable compromise, though, since the amplitude and phase computations are both very fast.


packages/app/src/providers/mock/mock-file.ts
```diff
@@ -248,8 +248,11 @@ export function makeMockFile(): GroupWithChildren {
             signal: withNxAttr(array('twoD_cplx'), {
               interpretation: 'spectrum',
             }),
-            auxiliary: { secondary_cplx: array('secondary_cplx') },
-            auxAttr: ['secondary_cplx'],
+            auxiliary: {
+              secondary_cplx: array('secondary_cplx'),
+              tertiary_float: array('tertiary_float'),
```
axelboc, line 253: Adding a float auxiliary signal to the `complex_spectrum` group in the mock demo.

Screenshot from 2025-08-18 16-51-39

packages/app/src/vis-packs/core/complex/ComplexLineToolbar.tsx
```diff
@@ -1,99 +0,0 @@
...
-        options={AXIS_SCALE_TYPES}
-      />
-
-      <Separator />
-
-      <ComplexVisTypeSelector
-        value={visType}
-        onChange={setVisType}
-        options={[ComplexVisType.Amplitude, ComplexVisType.Phase]}
-      />
```
axelboc, line 70: Moved to `LineToolbar` but rendered conditionally

packages/app/src/vis-packs/core/complex/lineConfig.tsx
```diff
@@ -1,45 +0,0 @@
...
-} from '@h5web/shared/vis-models';
-import { createContext, useContext, useState } from 'react';
-import { createStore, type StoreApi, useStore } from 'zustand';
-import { persist } from 'zustand/middleware';
-
-import { type ConfigProviderProps } from '../../models';
-
-export interface ComplexLineConfig {
-  visType: ComplexLineVisType;
-  setVisType: (visType: ComplexLineVisType) => void;
```
axelboc, line 13: Moved to the main line config.

packages/app/src/vis-packs/core/hooks.ts
```diff
@@ -27,17 +30,23 @@ import { applyMapping, getBaseArray, toNumArray } from './utils';
 export const useToNumArray = createMemo(toNumArray);
```
axelboc, line 30: The changes are in the underlying utility `toNumArray`.

packages/app/src/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -141,6 +144,9 @@ function MappedLineVis(props: Props) {
             dataDomain={combinedDomain}
             isSlice={selection !== undefined}
             disableErrors={!errors}
+            withComplexSelector={
+              isComplexType(dataset.type) || auxValues.some(isComplexArray)
```
axelboc, line 153: This is what controls whether to show or hide the amplitude/phase drop-down in the toolbar. Not 100% happy with this. I might pass the auxiliary signals definitions down from the containers in a future PR so I can check the dtypes instead of the values.

packages/app/src/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -157,7 +163,7 @@ function MappedLineVis(props: Props) {
         abscissaParams={abscissaParams}
         ordinateLabel={valueLabel}
         title={title}
-        dtype={formatNumLikeType(dataset.type)}
+        dtype={isComplexType(type) ? complexVisType : formatNumLikeType(type)}
```
axelboc, line 171: `dtype` is the string displayed in the tooltip next to the signal value, in parens, to hint at the underlying dtype. It wasn't passed at all in `MappedComplexLineVis`, so I decided to pass the complex vis type itself ("amplitude" or "phase").

Again, there's potential for improvement here, since the signal's dtype may not correspond to the dtype of the auxiliary signals any more. We may need to add dtype hints for auxiliaries as well...

packages/app/src/vis-packs/core/line/config.tsx
```diff
@@ -32,25 +40,28 @@ function createLineConfigStore() {
     persist(
       (set): LineConfig => ({
         customDomain: [null, null],
-        setCustomDomain: (customDomain: CustomDomain) => set({ customDomain }),
+        setCustomDomain: (customDomain) => set({ customDomain }),
 
         curveType: CurveType.LineOnly,
-        setCurveType: (type: CurveType) => set({ curveType: type }),
```
axelboc, line 38: Just removing some unnecessary types.

packages/app/src/vis-packs/core/utils.ts
```diff
@@ -86,17 +100,27 @@ function isBoolArray(val: ArrayValue<NumericLikeType>): val is boolean[] {
...
 ): NumArray | undefined {
   if (!arr) {
     return undefined;
   }
 
+  if (isComplexArray(arr)) {
+    return complexVisType === ComplexVisType.Amplitude
+      ? arr.map(([real, imag]) => Math.hypot(real, imag))
+      : arr.map(([real, imag]) => Math.atan2(imag, real));
+  }
```
axelboc, line 122: This is the key change as explained: `toNumArray` now turns complex arrays into number arrays just like other numeric-like arrays (booleans, bigint, etc.)

While this makes a lot of sense in my opinion, it does mean a slight performance downgrading because `useToNumArray(s)` now memoises `toNumArray` not only on the input array but also on the `complexVisType`.

packages/app/src/vis-packs/core/line/MappedLineVis.tsx
```diff
@@ -116,6 +120,10 @@ function MappedLineVis(props: Props) {
     scaleType: xScaleType,
   };
 
+  const ordinateLabel = isComplexType(type)
+    ? `${valueLabel} (${COMPLEX_VIS_TYPE_LABELS[complexVisType].toLowerCase()})`
+    : valueLabel;
+
```
axelboc, line 126: I had forgotten to merge in the `ordinateLabel` logic from `MappedComplexLineVis`. However, I've reworked it slightly to make it simpler and more readable. 

Old logic:

- If `valueLabel` is defined (i.e. `NxSpectrumContainer`): `"<valueLabel> (amplitude)"` or `"<valueLabel> (phase)"`
- Otherwise (i.e. `LineVisContainer`): `"Amplitude"` or `"Phase"`

Adding this logic as is to `MappedLineVis` was a bit cumbersome... So now, instead:

1. `valueLabel` always defaults to `dataset.name`;
2. when the dataset is complex, amplitude/phase is always added as a suffix to `valueLabel` (first bullet point above);
3. when the ordinate label ends up being the same as the title, it is no longer shown.

This ensures that we always have a decent ordinate label that we can use (e.g. in the CSV export) and that is independent of the title (so that we never lose the amplitude/phase info when applicable, as the title can be overridden via NeXus attributes)

As a result of all this, you'll noticed that in the screenshots, the ordinate label is removed when it used to duplicate the title, and in one of them, "Amplitude" becomes "one_D_cplx (amplitude)".

---

I forgot to mention also that the old complex line vis (and NeXus complex line vis) didn't used to support CSV export. So **now we get 1D CSV export for complex datasets and signals for free!**

Of course, what is exported is the amplitude or phase values — to export whole complex numbers, users still have to open the relevant complex dataset and export it from the _Matrix_ vis.

PR #1852, opened 2025-08-21
Add `Tooltip` and `Guides` building blocks

### LINKED ISSUE #1849: Use crosshair without tooltip
### Describe the bug

I am not able to use the crosshair without a tooltip.

What could i do?

A crosshair is something related to the the mouse, not really related to any tooltip.

### To Reproduce

Actually i do that
```
            <TooltipMesh
              guides={config.crossHair ? 'both' : undefined}
              renderTooltip={(x, y) => {
                return <></>;
              }}
            />
```

Returning `undefined` hide the crosshair.

But this displays anyway a tooltip shadow.

Image

### Context

- OS: na
- Browser: na
- Version: na
- H5Web context: @h5web/lib

axelboc:
Fix #1849

I'm extracting two low-level building blocks from `TooltipMesh` (more precisely `TooltipOverlay`): `Tooltip` and `Guides`, to open up more scenarios.

The API of  `TooltipMesh` remains unchanged and it still takes care of registering the pointer event listeners. The only difference is that it now renders two separate overlays instead of one (but still at the same `z-index`).

I add stories for the two new components, as well as one story for `TooltipMesh` to demonstrate rendering only the guides without the tooltip.


packages/lib/src/vis/shared/Guides.tsx
```diff
@@ -0,0 +1,32 @@
...
+import styles from './Guides.module.css';
+import Overlay from './Overlay';
+
+interface Props extends SVGAttributes<SVGSVGElement> {
+  show?: boolean;
+  top?: number | false;
+  left?: number | false;
+}
+
+function Guides(props: Props) {
```
axelboc, line 12: I decided to keep using the term "guides" rather than "crosshairs". My thinking is that crosshairs with only a horizontal line or only a vertical line ... are no longer crosshairs. :grin:

packages/lib/src/vis/shared/Guides.tsx
```diff
@@ -0,0 +1,32 @@
+import { type SVGAttributes } from 'react';
+
+import styles from './Guides.module.css';
+import Overlay from './Overlay';
+
+interface Props extends SVGAttributes<SVGSVGElement> {
+  show?: boolean;
```
axelboc, line 7: The `show` prop here and in the `Tooltip` component are to avoid rendering the components conditionally and therefore unmounting/remounting the overlays (which is inefficient).

packages/lib/src/vis/shared/Guides.tsx
```diff
@@ -0,0 +1,32 @@
...
+  top?: number | false;
+  left?: number | false;
```
axelboc, line 9: Pass `undefined` or `false` to hide the corresponding guide:

```
left={<condition> ? <number> : undefined}
left={<condition> && <number>}
```

packages/lib/src/vis/shared/Tooltip.tsx
```diff
@@ -0,0 +1,37 @@
...
+  left: number | undefined;
+  top: number | undefined;
+}
+
+function Tooltip(props: PropsWithChildren<Props>) {
+  const { className = '', style, show = true, left, top, children } = props;
+
+  return (
+    <Overlay className={styles.overlay}>
+      {show && children !== undefined && children !== false && (
```
axelboc, line 19: Just as before, the tooltip won't show at all if it doesn't have children. Since the `children` can be any `ReactNode` now (including `0`), I check for `undefined` and `false` explicitly.

packages/lib/src/vis/shared/TooltipMesh.tsx
```diff
@@ -1,16 +1,17 @@
...
+import Guides from './Guides';
+import Tooltip from './Tooltip';
 import { useVisCanvasContext } from './VisCanvasProvider';
 import VisMesh from './VisMesh';
 
 interface Props {
   size?: Size;
   guides?: 'horizontal' | 'vertical' | 'both';
-  renderTooltip: (x: number, y: number) => ReactElement | undefined;
+  renderTooltip: (x: number, y: number) => ReactNode;
```
axelboc, line 14: This gives a bit more flexibility (`undefined` still but also `false`, number, string, etc.)

PR #1854, opened 2025-08-21
Move relevant @types packages from dev to prod/peer dependencies

axelboc:
When some of H5Web's APIs are used in a consumer project, the types that come from external dependencies like `ndarray` become `any`. The fix is to install the relevant `@types` packages in the consumer project. While most consumers should already have `@types/react` installed, they may not want to install `@types/ndarray` if they don't use the library directly.

In this PR, I move all the relevant `@types` packages (all except `@types/node` I think) from `devDependencies` to either `peerDependencies` (when the corresponding package is also a peer dependency—e.g. `@types/react`) or "production" `dependencies` (e.g. `@types/ndarray` in `@h5web/app`).

When I move an `@types` package to `peerDependencies`, I also mark it as optional via `peerDependenciesMeta` so that non-TypeScript users don't get a warning from their package manager.

### DIFF
--- a/packages/h5wasm/package.json
+++ b/packages/h5wasm/package.json
@@ -37,10 +37,14 @@
   },
   "peerDependencies": {
     "@h5web/app": "workspace:*",
-    "react": ">=18",
+    "@types/react": "18.x",
+    "react": "18.x",
     "typescript": ">=4.5"
   },
   "peerDependenciesMeta": {
+    "@types/react": {
+      "optional": true
+    },
     "typescript": {
       "optional": true
     }
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -29,14 +29,34 @@
     "lint:tsc": "tsc"
   },
   "peerDependencies": {
-    "d3-array": "3.2.4",
-    "d3-format": "3.1.0",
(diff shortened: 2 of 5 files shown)


packages/h5wasm/package.json
```diff
@@ -37,10 +37,14 @@
   },
   "peerDependencies": {
     "@h5web/app": "workspace:*",
-    "react": ">=18",
```
axelboc, line 40: Fixing a few incorrect peer deps ranges in the process.

packages/shared/package.json
```diff
@@ -29,14 +29,34 @@
     "lint:tsc": "tsc"
   },
   "peerDependencies": {
-    "d3-array": "3.2.4",
-    "d3-format": "3.1.0",
-    "ndarray": "1.0.19",
-    "ndarray-ops": "1.2.2",
-    "react": ">=18",
-    "zustand": "5.0.3"
```
axelboc, line 37: These version numbers would often become out of sync because they don't get updated by pnpm when running `pnpm up -rL` and the like. They don't really matter anyway; the important part is that the peer dependencies are listed so that Vite and rollup can identify them as "external imports".

PR #1846, opened 2025-08-19
[NX Heatmap] Support NXdata group with mixed real and complex signals

axelboc:
Part 2 of #1834 

Similarly to part 1, the relevant work is mostly in `MappedHeatmapVis`. I'm not super happy with the added complexity brought on by the combined phase/amplitude heatmap visualization (with the amplitude as an alpha layer)... I will try to come up with a better implementation in a follow-up PR.

As with part 1, this merger leads to another huge removal of code.

[Screencast from 2025-08-19 15-09-40.webm](https://github.com/user-attachments/assets/08bfac7a-fc66-4540-95ee-bf0efdff1fcd)

Now [`file_fail.nxs`](http://localhost:5173/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Ffile_fail.nxs) works as expected:

image

Another side benefit: when viewing a complex dataset with _Heatmap_ or _NX Heatmap_, it is now possible to flip the image along the `x` and `y` axis. The old `ComplexToolbar` didn't have those controls.

PR #1856, opened 2025-08-25
[NX Heatmap] Support NXdata group with mixed real and complex signals

axelboc:
This addresses the main feature request of #1834, that is for the _NX Heatmap_ visualization to support `NXdata` groups with mixed real and complex signals (main and auxiliary signals alike). Demonstration with [`file_fail.nx`](http://localhost:5173/h5wasm?url=https%3A%2F%2Fwww.silx.org%2Fpub%2Fh5web%2Ffile_fail.nxs):

[Screencast from 2025-08-25 14-49-30.webm](https://github.com/user-attachments/assets/0d665a12-77b5-4e42-ab3e-a471d46ca806)

I'm basically combining, fixing and improving upon #1846 and #1855. (I've also closed #1842 because contrary to what I initially thought, the _NX Line_ case is more complicated.)

The idea remains the same: merging all the complex heatmap components into the main heatmap components. So now:

- a complex dataset goes through `HeatmapVisContainer` and `MappedHeatmapVis` (instead of `ComplexVisContainer` and `MappedComplexVis`), and;
- an `NXdata` group with a complex signal and/or complex auxiliaries goes through `NxImageContainer` and `MappedHeatmapVis` (instead of `NxComplexImageContainer` and `MappedComplexVis`)

While this still means that `MappedHeatmapVis` increases in complexity significantly, since it now has to support a `value` array of type `NumericLikeType | ComplexType`, I think I found an implementation that I'm happy with — with less misdirection and cleverness than in #1846 and #1855, and no compromises on performance when interacting with the phase/amplitude drop-down in the toolbar or when transposing the data.

As mentioned in #1846, as a side benefit of this work, complex images can now be flipped along the `x` and `y` axes.


packages/app/src/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -58,22 +64,47 @@ function MappedHeatmapVis(props: Props) {
...
+  const isComplex = isComplexArray(value);
+
+  const numArray = useMemo(
+    () => (!isComplex ? toNumArray(value) : undefined),
+    [isComplex, value],
+  );
+  const [phaseArray, amplitudeArray] = useMemo(
+    () => (isComplex ? getPhaseAmp(value) : []),
+    [isComplex, value],
+  );
```
axelboc, line 86: All the complexity comes from having to memoise computations while not being allowed to call hooks conditionally... In previous attempts, I tried to modify `toNumArray` to handle complex values by computing and returning phase or amplitude values directly. This had the downside of slowing down interactions with the phase/amplitude drop-down and of making it very hard to name variables like `numArray`.

Now I have three separate array variables, but the key is that `numArray` and `phaseArray`/`amplitudeArray` cannot both be defined at the same time:

- If we're dealing with a numeric-like dataset, only `numArray` (and therefore `numMapped`) is defined—we end up with `numMapped` and `numDomain` as the data array/domain.
- If we're dealing with a complex dataset, only `phaseArray` and `amplitudeArray` are defined—we end up with `phaseMapped`/`phaseDomain` or `amplitudeMapped`/`amplitudeDomain` or both depending on `complexVisType`.

packages/app/src/vis-packs/core/heatmap/MappedHeatmapVis.tsx
```diff
@@ -58,22 +64,47 @@ function MappedHeatmapVis(props: Props) {
...
+
+  const [dataArray, dataDomain] = numMapped
+    ? [numMapped, numDomain]
+    : complexVisType === ComplexVisType.Amplitude
+      ? [amplitudeMapped, amplitudeDomain]
+      : [phaseMapped, phaseDomain];
 
-  const dataDomain =
-    useDomain(dataArray, scaleType, undefined, ignoreValue) || DEFAULT_DOMAIN;
+  assertDefined(dataArray);
```
axelboc, line 107: TypeScript is not clever enough to realise that `dataArray` is always defined here.

packages/app/src/vis-packs/core/utils.ts
```diff
@@ -112,6 +119,20 @@ export function toNumArray(
...
+export function getPhaseAmp(
+  arr: ArrayValue<ComplexType>,
+): [phase: NumArray, amplitude: NumArray] {
+  const phase: number[] = Array.from({ length: arr.length });
+  const amplitude: number[] = Array.from({ length: arr.length });
+
+  arr.forEach(([real, imag], i) => {
+    phase[i] = Math.atan2(imag, real);
+    amplitude[i] = Math.hypot(real, imag);
+  });
```
axelboc, line 131: Trying to be a bit more efficient here by initialising the two arrays to the right length and assigning values to them instead of growing the arrays by pushing new values to them.

packages/lib/src/vis/hooks.ts
```diff
@@ -33,7 +33,7 @@ export const useValueToIndexScale = createMemo(getValueToIndexScale);
 export const useAxisDomain = createMemo(getAxisDomain);
 
 export function useDomain(
-  valuesArray: AnyNumArray,
+  valuesArray: AnyNumArray | undefined,
```
axelboc, line 36: Subtle change: I need to be able to pass a potentially `undefined` array to `useDomain`. For consistency, I make the same change in `useDomains` and `getDomain`.

PR #1862, opened 2025-08-27
Remove complex toolbars and configs

axelboc:
I move the phase/amplitude selector into the `HeatmapToolbar` and `LineToolbar`, and render it conditionally based on a new `isComplex` prop. I also move the `visType` state into the heatmap and line config stores (and rename it to `complexVisType`).

This allows removing `ComplexToolbar` and `ComplexLineToolbar` and their corresponding config stores/providers.

As a side benefit, the _Heatmap_ and _NX Heatmap_ visualizations now allow users to flip the heatmap along the `x` and `y` axes for complex datasets:

image


packages/app/src/vis-packs/core/line/LineToolbar.tsx
```diff
@@ -65,15 +80,28 @@ function LineToolbar(props: Props) {
...
-      />
+      {!isComplex && (
+        <ToggleBtn
+          label="Errors"
+          Icon={ErrorsIcon}
+          value={!disableErrors && showErrors}
+          onToggle={toggleErrors}
+          disabled={disableErrors}
+        />
+      )}
```
axelboc, line 104: We don't currently support errors in `MappedComplexLineVis` (i.e. when the main NXdata signal is complex). I debated adding a dedicated prop, but thought this was sufficient.

PR #1864, opened 2025-08-27
Refactor mapped complex components to improve memoisation

axelboc:
Memoisation was not quite right in the mapped complex vis components, especially in `MappedComplexLineVis`. The choice between using the phase or amplitude arrays and domains based on `complexVisType` has to be the last step—we have to compute and memoise them separately all the way otherwise we end up recomputing them or even re-transposing the arrays when the user interacts with the dimension mapper or the toolbar...

This neatens things up quite nicely, actually! :raised_hands:  We used to call the low-level utilities `getValidDomainForScale` and `getBaseArray` from the mapped complex vis components instead of the higher-level `useDomain` and `useMappedArray`. Now it's all consistent with the non-complex mapped vis components. :tada:

### DIFF
--- a/packages/app/src/vis-packs/core/complex/MappedComplexHeatmapVis.tsx
+++ b/packages/app/src/vis-packs/core/complex/MappedComplexHeatmapVis.tsx
@@ -1,9 +1,9 @@
 import {
   type DimensionMapping,
   HeatmapVis,
+  useDomain,
   useSafeDomain,
   useSlicedDimsAndMapping,
-  useValidDomainForScale,
   useVisDomain,
 } from '@h5web/lib';
 import {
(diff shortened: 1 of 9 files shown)


packages/app/src/vis-packs/core/complex/MappedComplexHeatmapVis.tsx
```diff
@@ -58,26 +58,20 @@ function MappedComplexHeatmapVis(props: Props) {
...
   const numAxisArrays = useToNumArrays(axisValues);
 
-  const [slicedDims, slicedMapping] = useSlicedDimsAndMapping(dims, dimMapping);
-  const mappedArray = useMappedArray(value, slicedDims, slicedMapping);
+  const slicedArgs = useSlicedDimsAndMapping(dims, dimMapping);
+  const phaseArray = useMappedArray(phase, ...slicedArgs);
+  const amplitudeArray = useMappedArray(amplitude, ...slicedArgs);
 
-  const { phaseValues, phaseBounds, amplitudeValues, amplitudeBounds } =
-    useMemo(() => getPhaseAmplitudeValues(mappedArray.data), [mappedArray]);
```
axelboc, line 67: Note that we now compute the phase/amplitude domains in separate loops. But this happens on first render and never again, so it's acceptable.

It's definitely better than to recompute the phase/amplitude values and domains every time the user remaps the dimensions (since previously `useMappedArray` was called before `getPhaseAmplitudeValues`).

packages/app/src/vis-packs/core/complex/MappedComplexLineVis.tsx
```diff
@@ -62,19 +63,26 @@ function MappedComplexLineVis(props: Props) {
     showGrid,
   } = config;
 
+  const { phaseArrays, amplitudeArrays } = usePhaseAmplitudeArrays([
+    value,
+    ...auxValues,
+  ]);
   const numAxisArrays = useToNumArrays(axisValues);
-  const [dataArray, ...auxArrays] = useMappedComplexArrays(
```
axelboc, line 66: This was a very complicated hook that tried to do way too many things. It really didn't help me keep my sanity when I was looking for a solution to the mixed `NXdata` problem... Looking at it now, I wish I had done this refactoring first...

PR #1865, opened 2025-09-01
Fix aux colors in dark mode

axelboc:
We had chosen custom, brighter colors for auxiliary lines in dark mode but they were no longer applied due to an outdated CSS variable name.

### BEFORE

image

### AFTER

image


packages/app/src/App.module.css
```diff
@@ -61,9 +61,6 @@
   --h5w-slicingSlider-track--color: var(--secondary-dark);
   --h5w-slicingSlider-thumb--borderColor: var(--secondary);
   --h5w-slicingSlider-thumb--bgColor: var(--primary-bg);
-
-  --h5w-line--color: darkblue;
-  --h5w-line--colorAux: orangered, forestgreen, red, mediumorchid, olive;
```
axelboc, line 66: They're the default values, so they don't need to be declared.

PR #1871, opened 2025-09-15
Remember better the last known dimension mapping state

axelboc:
I came across this article on the week-end: https://tkdodo.eu/blog/deriving-client-state-from-server-state — turns out that this trick can be applied to our dimension mapping state.

### Context

In order to remember the dimension mapping state (e.g. when switching between _Inspect_ and _Visualize_, or when selecting another dataset with the same dimensions and the visualization stays the same), we hold it in a zustand store. 

When comes the time to initialise a new mapping state in `useDimMappingState`, we first check if the new mapping is compatible with the previous mapping in the store (same dims, etc.) If it is, we return the mapping from the store; otherwise it is "stale" so we return the new mapping and reset the store via a `useEffect`.

The `useEffect` triggers an extra render, which, as it turns out, can be avoided.

### Improvement

When the dimension mapping in the store is stale, I return the new mapping but I no longer reset the store via `useEffect`. Instead, I keep the stale state in the store **until the user interacts with the `DimensionMapper` again**.

When this happens, I update the store with the new mapping of course, but I also update the context attached to that new mapping (i.e. `dims`, `axesCount` and `lockedDimsCount`).

### Benefits

As mentioned, this removes a render at the visualization container level — i.e. whenever we select a dataset, change visualization, etc.

However, there's another unexpected benefit: the user's last mapping state is stored **until the user interacts with the dimension mapper again**. So even if the user switches to a dataset with different dimensions and back, they will get back their mapping state as long as they don't interact with the dimension mapper in between. Here's a recording demonstrating this:

[Screencast from 2025-09-15 15-24-25.webm](https://github.com/user-attachments/assets/e2c0a5c9-4e05-4cbe-b087-281916cc5a0c)

<details>
<summary>Old behaviour for reference</summary>

[Screencast from 2025-09-15 15-26-01.webm](https://github.com/user-attachments/assets/61ef5306-7ff0-4bde-b5f5-b2220b364f56)

</details>

### DIFF
--- a/packages/app/src/dim-mapping-store.tsx
+++ b/packages/app/src/dim-mapping-store.tsx
@@ -3,8 +3,9 @@ import { type ArrayShape } from '@h5web/shared/hdf5-models';
 import {
   createContext,
   type PropsWithChildren,
+  useCallback,
   useContext,
-  useEffect,
+  useMemo,
   useState,
 } from 'react';
 import { createStore, type StoreApi, useStore } from 'zustand';
@@ -16,25 +17,14 @@ interface DimMappingState {
   axesCount: number;
   lockedDimsCount: number;
   mapping: DimensionMapping;
-  setMapping: (mapping: DimensionMapping) => void;
-  reset: (
-    dims: ArrayShape,
-    axesCount: number,
-    lockedDimsCount: number,
-    mapping: DimensionMapping,
-  ) => void;
 }
 
 function createDimMappingStore() {
-  return createStore<DimMappingState>((set) => ({
+  return createStore<DimMappingState>(() => ({
     dims: [],
     axesCount: 0,
     lockedDimsCount: 0,
     mapping: [],
-    setMapping: (mapping) => set({ mapping }),
-    reset: (dims, axesCount, lockedDimsCount, mapping) => {
-      set({ dims, axesCount, lockedDimsCount, mapping });
-    },
   }));
 }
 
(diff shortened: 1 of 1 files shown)


packages/app/src/dim-mapping-store.tsx
```diff
@@ -56,22 +46,31 @@ export function useDimMappingState(
...
+    () => initDimMapping(dims, axesCount, lockedDimsCount),
+    [axesCount, dims, lockedDimsCount],
+  );
 
-  const mapping = isStale
-    ? initDimMapping(dims, axesCount, lockedDimsCount)
-    : state.mapping;
+  const setMapping = useCallback(
+    (newMapping: DimensionMapping) => {
+      store.setState({ dims, axesCount, lockedDimsCount, mapping: newMapping });
```
axelboc, line 60: `setMapping` is kind of both a setter and a reset now.

Note that I now use `store.setState` to avoid declaring a setter in the store.

packages/app/src/dim-mapping-store.tsx
```diff
@@ -56,22 +46,31 @@ export function useDimMappingState(
...
 
-  useEffect(() => {
-    state.reset(dims, axesCount, lockedDimsCount, mapping);
-  }, [isStale]); // eslint-disable-line react-hooks/exhaustive-deps
+  if (
+    areSameDims(dims, state.dims) &&
+    axesCount === state.axesCount &&
+    lockedDimsCount === state.lockedDimsCount
+  ) {
+    // Previous dimenion mapping is still valid, so keep using it
```
loichuder:
```suggestion
    // Previous dimension mapping is still valid, so using it
```

PR #1872, opened 2025-09-16
Support NeXus `default_slice` attribute

### LINKED ISSUE #1844: NeXus/HDF5 "default_slice" attribute for NXdata instances is not interpreted (correctly)
### Describe the bug

NeXus/HDF5 files with NXdata instances which use the "default_slice" attribute do not cause H5Web to interpret that default slice. C-style 0th orthogonal slice is shown by default while 7th it should be (both slicing along axis_y).

### To Reproduce

1. Take this example: https://box.hu-berlin.de/d/2c4091ddda3249d2b3f9/
2. Display it using a e.g. the latest h5web, we observed with jupyterlab 12.4.0 and h5grove 2.3.0
3. View that group /entry1/atom_probe/reconstruction/naive_discretization/data an instance of NXdata
4. It has "default_slice ".", "7", "." instructing to slice to the 7th layer of the 12^3 intensity 3D matrix
5. However that information is not interpreted.

### Screenshots
Below screenshot for another example, in FAIRmat we observe this throughout several different use cases that all use NXdata, so it is not specific just for the dataset abovementioned

Image

Image

### Context

- OS: [e.g. Ubuntu 24.x]
- Browser: [Firefox latest]
- H5Web context: plain pypi jupyter, jupyterlab, jupyterlab_h5web install from one week ago, i.e., latest

axelboc:
Fix #1844 

[Screencast from 2025-09-16 15-57-57.webm](https://github.com/user-attachments/assets/730f8d2e-ef16-4d6c-8927-db5772f11e8d)

---

Note that the validation of the `default_slice` is done in two stages:

- First when retrieving the attribute from the `NXdata` group (`getDefaultSlice`) — at this stage, I check that the value of the attribute is an array of strings, that the array has the same size of the signal's dimensions, and that every value is either `'.'` or a valid number within the bounds of the corresponding dimension.
- Second when actually applying the default slice (`applyDefaultSlice`) — at this stage, I check that the number of `'.'` in the default slice arrays corresponds to the number of "mappable" dimensions (i.e. the number of axes + the number of locked dimensions for the selected visualization)

If at either stage the default slice doesn't pass the validity checks, it is simply ignored.


packages/app/src/__tests__/NexusPack.test.tsx
```diff
@@ -202,8 +216,32 @@ test('show error/fallback for malformed NeXus entity', async () => {
...
+  expect(screen.getByRole('slider', { name: 'D1' })).toHaveValue(0);
+
+  // Slicing index out of bound
+  await selectExplorerNode('default_slice_out_of_bounds');
+  expect(screen.getByRole('slider', { name: 'D0' })).toHaveValue(0);
+  expect(screen.getByRole('slider', { name: 'D1' })).toHaveValue(0);
+});
+
 test('ignore malformed `SILX_style` attribute', async () => {
-  const errorSpy = mockConsoleMethod('error');
```
axelboc, line 206: Unrelated: no point in mocking the console in this test since we just check that it's not called... If `console.error` is invoked, the test would fail anyway.

packages/app/src/dim-mapping-store.tsx
```diff
@@ -41,19 +42,25 @@ export function DimMappingProvider(props: PropsWithChildren<Props>) {
...
-  dims: number[],
-  axesCount: number,
-  lockedDimsCount = 0,
-): [DimensionMapping, (mapping: DimensionMapping) => void] {
+export function useDimMappingState(opts: {
+  dims: number[];
+  axesCount: number;
+  lockedDimsCount?: number;
+  defaultSlice?: DefaultSlice;
+}): [DimensionMapping, (mapping: DimensionMapping) => void] {
```
axelboc, line 50: I am changing the signature of this hook to accept a single `opts` object instead of four separate parameters. Since `lockedDimsCount` is optional, the calls would have become unreadable.

packages/app/src/dim-mapping-store.tsx
```diff
@@ -41,19 +42,25 @@ export function DimMappingProvider(props: PropsWithChildren<Props>) {
...
-  const mapping = useMemo(
-    () => initDimMapping(dims, axesCount, lockedDimsCount),
-    [axesCount, dims, lockedDimsCount],
-  );
+  const mapping = useMemo(() => {
+    const newMapping = initDimMapping(dims, axesCount, lockedDimsCount);
+
+    return defaultSlice
+      ? applyDefaultSlice(newMapping, defaultSlice)
+      : newMapping;
```
axelboc, line 62: Instead of trying to do everything at once, I decided to keep `initDimMapping` unchanged and apply the default slice to the initial mapping as a separate step. I think it makes the algorithm simpler, or at least easier to understand. Moreover, `initDimMapping` is exported in `@h5web/lib` and I've already used it in a consumer application — it's useful as is, so I'm reluctant to change it.

cypress/snapshots/app.cy.ts/default_slice.snap.png
axelboc, line 1: New screenshot here (as usual, the others are just diffs because of the new `NXdata` group in the explorer)

packages/app/src/dim-mapping-store.tsx
loichuder: Agreed :+1:

loichuder (commented):
Perhaps I missed it but should we not print warnings in the console if the `default_slice` is malformed?

I think that if the `default_slice` attribute is present, there is a will from the file creator to use this feature so that any issue should probably be raised to understand what is going on

axelboc:
Sorry, I just forgot about it!

I've now added two warnings:

- when the length doesn't match the number of dimensions
- when the indices are out of bounds (or NaN)

Like with `SILX_style`, I do not warn if the shape/type of the attribute is not what we expect (i.e. an array of strings), as this seems highly unlikely to happen. I've actually also removed the mock `NXdata` group that used to test `default_slice` with a mixed array of strings and numbers, as I doubt anyone would create a `default_slice` attribute with a compound type in practice... :smile:

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -231,6 +237,48 @@ export function findTitleDataset(
...
+    !Array.isArray(defaultSliceRaw) ||
+    !defaultSliceRaw.every((v) => typeof v === 'string')
+  ) {
+    return undefined;
+  }
+
+  if (defaultSliceRaw.length !== signalDims.length) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      "Malformed 'default_slice' attribute: expected same length as signal dimensions",
```
loichuder, line 257: Might be useful to give the numbers here. Something like

```suggestion
      "Malformed 'default_slice' attribute: mismatch between `default_slice` length (${defaultSliceRaw.length}) and signal dimensions (${signalDims.length})",
```

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -231,6 +237,48 @@ export function findTitleDataset(
...
+  });
+
+  const isValid = defaultSlice.every((v, i) => {
+    // Look for invalid or out-of-bounds indices
+    return v === '.' || (!Number.isNaN(v) && v >= 0 && v < signalDims[i]);
+  });
+
+  if (!isValid) {
+    // eslint-disable-next-line no-console
+    console.warn(
```
loichuder, line 273: Same as before. The `warn` could be in the `isValid` callback to get information about `i` and `v`

axelboc: I'm not fussed to be honest. It adds a little bit of complexity for very little gain. Same for the other warning. As discussed, H5Web is not a NeXus format validator and users have to know to open their dev tools anyway, so very few are likely to see these messages... Informing that the `default_slice` attribute has been considered but rejected, and roughly why, is good enough in my opinion. If they really don't understand why from the warning, they can always open an issue.

loichuder: > users have to know to open their dev tools anyway, so very few are likely to see these messages...

Fair

PR #1878, opened 2025-09-22
Add possiblity to have piecewise constant interpolation in LineVis

### LINKED ISSUE #1875: Request more interpolation methods: Piecewise constant, linear, and PCHIP for 1-D signal
Currently, h5web show 1-D plot using piecewise linear interpolation. However, in some cases (e.g. digital signals processing), we prefer it to be piecewise constant. 

The comparison is shown below (the blue is from H5web, the red dot line is what I expected):

Image

### Requested solution

So it would be nice that H5Web can support interpolation method other than piecewise linear. 
Some solutions I can think of:
1.  Add some button in GUI that allows user to specify which interpolation method it can be used for 1D signal. 
2. H5web can hornor some attribute in the signal group which tells the interpolation method, e.g. `@1D_interpolation="piecewise-constant"`.

loichuder:
Fix #1875


packages/lib/src/vis/line/piecewiseConstantLineGeometry.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    this.attributes.position.setXYZ(posIndex, x, y, 0);
+
+    const nextX = abscissaScale(abscissas[index + 1]);
+    const nextY = ordinateScale(ordinates[index + 1]);
+    if (!Number.isFinite(nextX) || !Number.isFinite(nextY)) {
+      this.attributes.position.setXYZ(posIndex + 1, 0, 0, Z_OUT);
+      return;
+    }
```
loichuder, line 57: This specific part is to handle edge cases with the log scale for example:

image

If the next point is not visible, the line segment should not be added

axelboc:
Yeah no, just a random failure I can't get my head around... I'm hoping React 19 will solve it... :crossed_fingers:

packages/lib/src/vis/line/piecewiseConstantLineGeometry.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    if (!Number.isFinite(x) || !Number.isFinite(y)) {
+      this.attributes.position.setXYZ(posIndex, 0, 0, Z_OUT);
+      this.attributes.position.setXYZ(posIndex + 1, 0, 0, Z_OUT);
+      return;
+    }
+
+    this.attributes.position.setXYZ(posIndex, x, y, 0);
+
+    const nextX = abscissaScale(abscissas[index + 1]);
```
axelboc, line 52: I think `abscissaScale` and `ordinateScale` are called with `undefined` for the last point — i.e. when `index === length - 1`. Might be good to add an early return before this line to make this more explicit.

axelboc:
@NAThompson thanks, we just need to find a slightly more space-efficient way to integrate this feature into the toolbar.

axelboc:
> NeXus might already have such a standard...

Unfortunately we don't believe that's the case, and we're not very keen on adding a new NX attribute ourselves (or a field to the already-non-standard `SILX_style` attribute)

@loichuder is currently working on an "aspect" drop-down menu which would combine both the interpolation and the Line/Points/Both options. That should actually take up less space overall. :tada:

PR #1880, opened 2025-09-24
Replace pieceWiseConstant bool by Interpolation enum

loichuder:
Early step for the new line aspect selector


packages/lib/src/toolbar/controls/LineAspectSelector.tsx
axelboc, line 1: Did you mean to commit this? 😁

loichuder: Oops ! Quick, avert your eyes :rofl:

PR #1881, opened 2025-09-25
Replace Line/Points and Piecewise toggle buttons by a dropdown

### LINKED ISSUE #1875: Request more interpolation methods: Piecewise constant, linear, and PCHIP for 1-D signal

loichuder:
Fix #1875 

[Screencast from 2025-09-25 13-48-01.webm](https://github.com/user-attachments/assets/a7488aa2-da4e-40b8-a740-d6cf5ca753cc)

Side boon: interpolation is disabled when displaying only points (since it is not relevant).

### DIFF
--- a/packages/lib/src/toolbar/controls/LineAspectSelector/LineAspectSelector.tsx
+++ b/packages/lib/src/toolbar/controls/LineAspectSelector/LineAspectSelector.tsx
@@ -1,0 +1,91 @@
+import { useClick, useInteractions } from '@floating-ui/react';
+import { useId } from 'react';
+import { MdArrowDropDown, MdAutoGraph } from 'react-icons/md';
+
+import { CurveType, Interpolation } from '../../../vis/line/models';
+import toolbarStyles from '../../Toolbar.module.css';
+import { useFloatingDismiss, useFloatingMenu } from '../hooks';
+import RadioGroup from './RadioGroup';
+
+const CURVE_TYPE_LABELS: Record<CurveType, string> = {
+  [CurveType.LineOnly]: 'Line',
+  [CurveType.GlyphsOnly]: 'Points',
+  [CurveType.LineAndGlyphs]: 'Both',
+};
+
+interface Props {
+  curveType: CurveType;
+  onCurveTypeChanged: (curveType: CurveType) => void;
+  interpolation: Interpolation;
+  onInterpolationChanged: (interpolation: Interpolation) => void;
+}
+
+function LineAspectSelector(props: Props) {
+  const {
+    curveType,
+    onCurveTypeChanged,
+    interpolation,
+    onInterpolationChanged,
+  } = props;
+
+  const { context, refs, floatingStyles } = useFloatingMenu();
+  const { open: isOpen, floatingId } = context;
+
+  const referenceId = useId();
+
+  const { getReferenceProps, getFloatingProps } = useInteractions([
+    useClick(context),
+    useFloatingDismiss(context),
+  ]);
+
+  return (
+    <>
+      <button
+        ref={refs.setReference}
+        id={referenceId}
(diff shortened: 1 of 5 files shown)


packages/lib/src/toolbar/controls/LineAspectSelector/LineAspectSelector.tsx
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    <>
+      <button
+        ref={refs.setReference}
+        id={referenceId}
+        className={toolbarStyles.btn}
+        type="button"
+        role="combobox"
+        aria-label="Line aspect"
+        aria-haspopup="listbox"
```
axelboc:
```suggestion
        aria-haspopup="menu"
```

packages/lib/src/toolbar/controls/LineAspectSelector/LineAspectSelector.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return (
+    <>
+      <button
+        ref={refs.setReference}
+        id={referenceId}
+        className={toolbarStyles.btn}
+        type="button"
+        role="combobox"
+        aria-label="Line aspect"
```
axelboc:
```suggestion
```

packages/lib/src/toolbar/controls/LineAspectSelector/LineAspectSelector.tsx
```diff
@@ -0,0 +1,94 @@
...
+        </span>
+      </button>
+
+      {isOpen && (
+        <div
+          ref={refs.setFloating}
+          id={floatingId}
+          className={toolbarStyles.menu}
+          style={floatingStyles}
+          role="listbox"
```
axelboc:
```suggestion
          role="menu"
```

packages/lib/src/toolbar/controls/LineAspectSelector/RadioGroup.tsx
```diff
@@ -0,0 +1,54 @@
...
+            htmlFor={option}
+            aria-disabled={disabled}
+          >
+            <input
+              id={option}
+              name={name}
+              type="radio"
+              checked={option === value}
+              onChange={() => onValueChanged(option)}
+              aria-labelledby={`${option}-label`}
```
axelboc, line 42:
```suggestion
              role="menuitemradio"
              aria-labelledby={`${option}-label`}
```

packages/lib/src/toolbar/controls/LineAspectSelector/RadioGroup.module.css
```diff
@@ -0,0 +1,27 @@
...
+  display: flex;
+  flex-direction: column;
+}
+
+.option {
+  composes: btnClean from global;
+  display: flex;
+  align-items: center;
+  padding: 0.375rem 0.75rem;
+  transition: background-color 0.05s ease-in-out;
```
axelboc:
```suggestion
```

axelboc (approved):
Brilliant, I just made a couple of suggestions with the ARIA markup to use the `menu` role instead of the `combobox` role (which is more for "select" and "autocomplete" drop-downs)

PR #1882, opened 2025-09-25
Fix wrapping of attribute value buttons

axelboc:
### Before

image

### After

image


packages/app/src/visualizer/utils.ts
```diff
@@ -32,62 +32,41 @@ import { NEXUS_VIS, NexusVis } from '../vis-packs/nexus/visualizations';
 export function resolvePath(
   path: string,
   entitiesStore: EntitiesStore,
-  attrValueStore: AttrValuesStore,
```
axelboc, line 35: While I'm at it, I'm fixing typos in this module and moving the order of a function.

loichuder (approved):

PR #1883, opened 2025-09-25
Always show all supported NeXus vis instead of just the primary one

### LINKED ISSUE #1876: Slicing only available on raw waveform, not on NeXus-labelled groups

When a heatmap/matrix is labeled with the requisite NeXus metadata, H5Web displays an image with correct axis labels:

Image

Now, there is a really nice feature of H5Web, where you can slice 1D lines out of images and create "timeseries":

Image

However, this option is not available on the NeXus group-rather it is only available on the matrix itself. This means that the axes labels are lost, and indexing is simply integers at this point.

It would be nice if the slices kept the dimensioned axes-the "Matrix Line Heatmap" option just needs to be "promoted" to the Nexus group:

Image

Data to reproduce, with a .txt extension to get it past the upload filter:

[h5web_volume_example.h5.txt](https://github.com/user-attachments/files/22415597/h5web_volume_example.h5.txt)

axelboc:
Fix #1876 

[Screencast from 2025-09-25 15-36-05.webm](https://github.com/user-attachments/assets/8351020d-1a26-4506-a2af-5a1794683161)

### DIFF
--- a/packages/app/src/__tests__/NexusPack.test.tsx
+++ b/packages/app/src/__tests__/NexusPack.test.tsx
@@ -13,28 +13,36 @@ import { NexusVis } from '../vis-packs/nexus/visualizations';
 test('visualize NXdata group with explicit signal interpretation', async () => {
   // Signal with "spectrum" interpretation
   const { selectExplorerNode } = await renderApp('/nexus_entry/spectrum');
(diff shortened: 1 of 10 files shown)


axelboc:
I've improved the implementation but it's now a bit much for one PR. I'm gonna split it up so it's more digestible.

PR #1885, opened 2025-09-25
Split `NXnote` and `NXdata` vis definitions

axelboc:
I'm refactoring the `resolvePath` function, which finds the supported visualizations for the current entity (following any default NeXus path as needed).

To solve #1883, my goal is for each NeXus visualization to decide on its own if it supports the current entity or not (like we do with core visualizations).

This was made very challenging by the fact that the _NX Note_ visualization definition was mixed in with the other `NXdata` visualization definitions. So in this PR, I define the _NX Note_ visualization separately from the others.

I then rename a few things, like the `NEXUS_VIS` dictionary to `NX_DATA_VIS`, and I refactor `resolvePath` to simplify the algorithm a little bit and make it more explicit — notably based on the entity kind (dataset, `NXdata` group, etc.)

### DIFF
--- a/packages/app/src/test-utils.tsx
+++ b/packages/app/src/test-utils.tsx
@@ -11,12 +11,12 @@ import { expect, type MockInstance, vi } from 'vitest';
 import App from './App';
 import MockProvider from './providers/mock/MockProvider';
 import { type Vis } from './vis-packs/core/visualizations';
-import { type NexusVis } from './vis-packs/nexus/visualizations';
+import { type NxDataVis } from './vis-packs/nexus/visualizations';
 
 interface RenderAppResult extends RenderResult {
   user: ReturnType<typeof userEvent.setup>;
   selectExplorerNode: (name: string) => Promise<void>;
-  selectVisTab: (name: Vis | NexusVis) => Promise<void>;
+  selectVisTab: (name: Vis | NxDataVis) => Promise<void>;
 }
 
 type InitialPath = `/${string}`;
--- a/packages/app/src/visualizer/utils.ts
+++ b/packages/app/src/visualizer/utils.ts
@@ -8,6 +8,8 @@ import {
 } from '@h5web/shared/guards';
 import {
   type ChildEntity,
+  type Dataset,
+  type GroupWithChildren,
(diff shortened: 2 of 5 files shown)


packages/app/src/visualizer/utils.ts
```diff
@@ -36,9 +41,24 @@ export function resolvePath(
...
+    return supportedVis.length > 0 ? { entity, supportedVis } : undefined;
+  }
+
+  if (!isGroup(entity)) {
+    return undefined;
+  }
+
+  if (isNxNoteGroup(entity, attrValuesStore)) {
+    return { entity, supportedVis: [NX_NOTE_VIS] };
+  }
```
axelboc, line 56: I now consider `NXnote` and `NXdata` groups as distinct entity "kinds" at the top level of the algorithm. This will help me refactor `getSupportedNxDataVis` much more easily.

packages/app/src/test-utils.tsx
```diff
@@ -11,12 +11,12 @@ import { expect, type MockInstance, vi } from 'vitest';
...
 import MockProvider from './providers/mock/MockProvider';
 import { type Vis } from './vis-packs/core/visualizations';
-import { type NexusVis } from './vis-packs/nexus/visualizations';
+import { type NxDataVis } from './vis-packs/nexus/visualizations';
 
 interface RenderAppResult extends RenderResult {
   user: ReturnType<typeof userEvent.setup>;
   selectExplorerNode: (name: string) => Promise<void>;
-  selectVisTab: (name: Vis | NexusVis) => Promise<void>;
+  selectVisTab: (name: Vis | NxDataVis) => Promise<void>;
```
loichuder, line 19: Is there not typing problems now that `NxNote` is not part of `NxDataVis` anymore?

axelboc: We don't ever need to select the _NX Note_ vis tab in the tests. :smile_cat:

loichuder: Oh shoot, missed that it was in the test-utils...

loichuder (approved):

Issue #1817, opened 2025-06-10
Line thickness


axelboc:
### Is your feature request related to a problem?

Due to a WebGL limitation, `LineBasicMaterial` cannot have a [`lineWidth`](https://threejs.org/docs/#api/en/materials/LineBasicMaterial.linewidth) greater than 1. This affects the `Line` and `DataCurve` components.

### Requested solution or feature

Provide equivalents to `Line` and `DataCurve` based on Three's [`Line2`](https://threejs.org/docs/index.html#examples/en/lines/Line2), which does support rendering thick lines.

### Additional context

Requested in #1815

PR #1888, opened 2025-09-30
Refactor `useGeometry` to support any Three.js geometry class

axelboc:
This is a non-trivial refactoring ahead of addressing #1817 

### Previously

```ts
const geometry = useGeometry(
  SomeGeometry,
  dataLength,
  { ...someParams }
)

class SomeGeometry extends H5WebGeometry { ... }

abstract class H5WebGeometry extends BufferGeometry { ... }
```

The problem is that for [`Line2`](https://threejs.org/docs/#examples/en/lines/Line2), we need a more advanced geometry than `BufferGeometry` — we need a [`LineGeometry`](https://threejs.org/docs/#examples/en/lines/LineGeometry). Since `useGeometry` only accepts a class that extends `H5WebGeometry` and `H5WebGeometry` extends `BufferGeometry`, we're stuck.

### Now

```ts
const geometry = useMemo(() => new SomeGeometry({ ...someParams }), [...]);
useUpdateGeometry(geometry)

class SomeGeometry extends BufferGeometryOrSubClass implements H5WebGeometry { ... }
interface H5WebGeometry { ... }
```

`SomeGeometry` can now extend any subclass of `BufferGeometry` including `LineGeometry`.

Due to some typing complexity related to constructors that I would struggle to explain, keeping the same signature for `useGeometry` was a bit messy. I also disliked the separate `prepare` method and the fact that `dataLength` is passed separately (since it's typically the length of one of the array parameters passed to the geometry).

I found that letting consumers instantiate the geometry themselves with the required parameters was much less obscure and made the types a lot simpler. Additionally:

- The parameters are now passed to the geometry constructors instead of via a `prepare` method.
- The geometry constructors retrieve the data length from whichever parameter they like.
- The simplified `useUpdateGeometry` hook now calls an `update()` method and it's up to this method to loop through the data however it wants.


packages/lib/src/vis/hooks.ts
```diff
@@ -111,59 +111,36 @@ export function useCssColors(
...
-  );
+  useLayoutEffect(() => {
+    if (skipUpdates) {
+      return;
+    }
+
+    geometry.update();
+
+    if (isInteractive) {
+      geometry.computeBoundingBox();
```
axelboc, line 132: `computeBoundingSphere` calls this internally if needed, but I think it's best to be explicit.

packages/lib/src/vis/hooks.ts
```diff
@@ -111,59 +111,36 @@ export function useCssColors(
...
-      isInteractive,
-      invalidate,
-    ],
-  );
+  useLayoutEffect(() => {
+    if (skipUpdates) {
+      return;
+    }
+
+    geometry.update();
```
axelboc, line 129: Each geometry's `update` method is now in charge of looping through the data. This means that this hook no longer needs to know the `dataLength`.

packages/lib/src/vis/hooks.ts
```diff
@@ -111,59 +111,36 @@ export function useCssColors(
...
+  geometry: H5WebGeometry & BufferGeometry,
   config: {
     skipUpdates?: boolean; // set to `true` when R3F object is hidden
     isInteractive?: boolean; // set to `true` to keep bounding sphere up to date for raycaster
   } = {},
-): H5WebGeometry<AttributeNames, Params> {
+): void {
   const { skipUpdates = false, isInteractive = false } = config;
-
-  const geometry = useMemo(() => new Ctor(dataLength), [Ctor, dataLength]);
```
axelboc, line 128: This hook no longer has to instantiate a geometry class.

packages/lib/src/vis/line/ErrorBars.tsx
```diff
@@ -17,31 +18,33 @@ interface Props {
...
-
-  const geometryParams = {
+  const {
     abscissas,
     ordinates,
     errors,
-    abscissaScale,
-    ordinateScale,
+    color,
+    visible = false,
```
axelboc, line 26: Very subtle fix here: without this default value, I was seeing a subtle empty error bar in the middle of the canvas:

image

packages/lib/src/vis/hooks.ts
```diff
@@ -111,59 +111,36 @@ export function useCssColors(
...
-        geometry.index.needsUpdate = true;
-      }
-
-      invalidate();
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [
-      geometry,
-      dataLength,
-      ...Object.values(params), // eslint-disable-line react-hooks/exhaustive-deps
```
axelboc, line 161: The `useMemo` calls that instantiate the geometries in the components now list the params explicitly, so these `react-hooks/exhaustive-deps` errors goes away for good.

packages/lib/src/vis/line/Line.tsx
```diff
@@ -49,25 +49,26 @@ function Line(props: Props) {
 
   const Geometry =
     interpolation === Interpolation.Constant
-      ? PieceWiseConstantLineGeometry
```
axelboc, line 52: I took the liberty of renaming this geometry as it was a bit of a mouthful.

packages/lib/src/vis/line/errorBarsGeometry.ts
```diff
@@ -1,62 +1,63 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
axelboc, line 1: We no longer have to disable this linting error in any of the geometry file because by getting rid of the `prepare` method, the `params` instance variable is now always defined.

packages/lib/src/vis/line/errorBarsGeometry.ts
```diff
@@ -1,62 +1,63 @@
...
+  implements H5WebGeometry
+{
+  public constructor(private readonly params: Params) {
     super();
+    const { length } = params.ordinates;
     this.setAttribute('position', createBufferAttr(length * 2));
   }
 
-  public update(index: number): void {
+  public update(): void {
```
axelboc, line 22: Sorry for the noise in the diff, I renamed `bufferIndex` and destructured `this.attributes.position`... The notable change is of course the `for...of` loop and the use of `continue` instead of `return`.

packages/lib/src/vis/line/glyphsGeometry.ts
```diff
@@ -1,44 +1,40 @@
...
 import { createBufferAttr, Z_OUT } from '../utils';
+import { type LineGeometryParams } from './lineGeometry';
 
-interface Params {
-  abscissas: NumArray;
-  ordinates: NumArray;
-  abscissaScale: AxisScale;
-  ordinateScale: AxisScale;
-  ignoreValue?: IgnoreValue;
-}
```
axelboc, line 14: I removed a bit of duplication by exporting `LineGeometryParams` from the line geometry.

apps/storybook/src/Utilities.mdx
```diff
@@ -422,43 +422,33 @@ const { delta, isDragging, startDrag } = useDrag({
...
 );
+
+useUpdateGeometry(lineGeometry, { skipUpdates: !visible });
 ```
 
 The following built-in geometry classes are available and can be used as examples for
 writing your own geometries:
 
 - `LineGeometry`, used in [`Line`](https://h5web-docs.panosc.eu/?path=/docs/building-blocks-line--docs)
+- `LineConstantGeometry`, used in [`Line`](https://h5web-docs.panosc.eu/?path=/docs/building-blocks-line--docs) with constant interpolation
```
loichuder, line 451: Thanks 😅

loichuder (approved):
Looks good ! Well done for the great refactoring !

PR #1889, opened 2025-10-02
Support thick lines

### LINKED ISSUE #1817: Line thickness

axelboc:
At long last, fix #1817 

I've decided to completely replace `Line` and `LineBasicMaterial` with [`Line2`](https://threejs.org/docs/#examples/en/lines/Line2) and [`LineMaterial`](https://threejs.org/docs/?q=linema#examples/en/lines/LineMaterial) from `three/addons/lines`, since they no longer require installing a separate package and seem to now be official add-ons rather than just examples.

Technically, since the viewer doesn't use this feature yet, there's a slight impact on memory with no actual gain in `@h5web/app` — indeed, the position buffers are now twice as long because the new geometries (based on [`LineGeometry`](https://threejs.org/docs/?q=linegeo#examples/en/lines/LineGeometry)) require line segments rather than points. However, I think it's an acceptable compromise considering that the alternative would probably be to duplicate the `Line` component and its two (for now) geometries.

It's worth noting that by default, `LineMaterial` assumes that the specified `linewidth` is in CSS pixels, which means that it remains constant regardless of zoom level. This is exactly what we want, which is great. It's possible to change this behaviour by setting the [`worldUnits`](https://threejs.org/docs/?q=linema#examples/en/lines/LineMaterial.linewidth) boolean prop, but I didn't document it as I'm not sure there's a valid use case for it.

[Screencast from 2025-10-02 12-46-18.webm](https://github.com/user-attachments/assets/85db8a85-fd3e-4a87-966c-d985d3a1c543)

The new constant interpolation added a bit more work, but it was manageable.

---

So overall, the only consumer-facing changes for now are as follows:
- `[DataCurve, Line]` Add new `width` prop to control the thickness of the line
- :warning: `[DataCurve, Line]` Prop `materialProps` now accepts properties of Three's [`LineMaterial`](https://threejs.org/docs/?q=lin#examples/en/lines/LineMaterial) instead of [`LineBasicMaterial`](https://threejs.org/docs/?q=lin#api/en/materials/LineBasicMaterial)
- :warning: `[Line]` Accept extra props from Three's [`Line2`](https://threejs.org/docs/?q=line2#examples/en/lines/Line2) instead of [`Line`](https://threejs.org/docs/?q=line#api/en/objects/Line)

---

**EDIT** I tried to add support for dashed lines. The idea is to compute the [cumulative distances](https://github.com/mrdoob/three.js/blob/master/examples/jsm/lines/LineSegments2.js#L283) to the start and end of every segment. The problem, which we had identified a long time ago, is that these distances are in world space so the size of the dashes and gaps increase when zooming. `Line2` solves this problem for the line width, but `LineSegments2` doesn't solve it for dashes... and I'm not fond of the idea of updating the geometry on every R3F frame, obviously. So I'm putting this work on the back burner.


packages/lib/src/vis/line/lineConstantGeometry.ts
```diff
@@ -1,58 +1,84 @@
...
 
 import { type H5WebGeometry } from '../models';
 import { createBufferAttr, Z_OUT } from '../utils';
 import { type LineGeometryParams } from './lineGeometry';
 
-class LineConstantGeometry
-  extends BufferGeometry<Record<'position', BufferAttribute>>
-  implements H5WebGeometry
-{
+class LineConstantGeometry extends ThreeLineGeometry implements H5WebGeometry {
```
axelboc, line 12: I recommend reviewing the diff of `LineGeometry` first :wink:

packages/lib/src/vis/line/lineGeometry.ts
```diff
@@ -12,38 +17,70 @@ interface Params {
...
-    const { length } = params.ordinates;
-    this.setAttribute('position', createBufferAttr(length));
+
+    this.length = params.ordinates.length;
+    this.positions = createBufferAttr((this.length - 1) * 2); // two positions per line segment
+
+    // Parts of `LineGeometry#setPositions` that don't need to be called on every update
+    // https://github.com/mrdoob/three.js/blob/40728556ec00833b54be287807cd6fb04a897313/examples/jsm/lines/LineSegmentsGeometry.js#L97
+    const { array } = this.positions;
+    const instancedBuffer = new InstancedInterleavedBuffer(array, 2 * 3); // two sets of `xyz` coords per line segment
```
axelboc, line 33: In `LineGeometry`, the `positions` buffer attribute has to contain line segments rather than connected points. It is wrapped in an `InstancedInterleavedBuffer`, which is exposed via two `InterleavedBufferAttribute`: `instanceStart` and `instanceEnd`, which provide `LineMaterial` with access to the start and end points of the line segments.

packages/lib/src/vis/line/lineGeometry.ts
```diff
@@ -12,38 +17,70 @@ interface Params {
...
 
     for (const [index, value] of ordinates.entries()) {
-      if (ignoreValue?.(value)) {
-        positions.setXYZ(index, 0, 0, Z_OUT);
+      const posIndex = index * 2;
+
+      if (index >= this.length - 1) {
+        this.setInvalidSegment(posIndex);
+        continue;
+      }
```
axelboc, line 51: Skipping the last data point, since it's the end of the second last data point's segment.

packages/lib/src/vis/line/lineGeometry.ts
```diff
@@ -12,38 +17,70 @@ interface Params {
...
+
+      const nextValue = ordinates[index + 1];
+
+      if (ignoreValue?.(value) || ignoreValue?.(nextValue)) {
```
axelboc, line 55: Don't render the segment either if:

- its start and/or end value is ignored
- its computed positions are not finite (see further down).

PR #1892, opened 2025-10-13
Tweak and document line aspect menu

axelboc:
The changes are very subtle in the UI but basically:

- the spacing between the "Aspect" label and the arrow is consistent with controls based on the `Selector` component
- the radio groups are a bit more spaced out from the top/bottom of the container
- options are highlighted on hover

| BEFORE | AFTER |
| --- | --- |
| image | image |

Additionally:

- I rename `LineAspectSelector` to `LineAspectMenu`
- I export `RadioGroup`
- I move the two components to the root of the `toolbar/controls` folder, since `RadioGroup` is quite generic
- I demonstrate the two new components in Storybook

### DIFF
--- a/apps/storybook/src/RadioGroup.stories.tsx
+++ b/apps/storybook/src/RadioGroup.stories.tsx
@@ -1,0 +1,37 @@
+import { RadioGroup } from '@h5web/lib';
+import { type Meta, type StoryObj } from '@storybook/react-vite';
+import { useState } from 'react';
+
+type Option = 'Bar' | 'Baz';
+const OPTIONS: Option[] = ['Bar', 'Baz'];
+
+const meta = {
+  title: 'Toolbar/RadioGroup',
+  component: RadioGroup,
+  argTypes: {
+    options: { control: false },
+    value: { control: false },
+  },
+} satisfies Meta<typeof RadioGroup>;
+
+export default meta;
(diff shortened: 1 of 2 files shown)


packages/lib/src/toolbar/controls/LineAspectMenu.tsx
```diff
@@ -29,7 +30,7 @@ function LineAspectSelector(props: Props) {
   } = props;
 
   const { context, refs, floatingStyles } = useFloatingMenu();
-  const { open: isOpen, floatingId } = context;
+  const { floatingId, open: isOpen } = context;
```
axelboc, line 33: Just bringing the implementation closer to `ExportMenu` for consistency.

Reusing the `Btn` component for the trigger like in `ExportMenu` is what fixed the spacing of the "Aspect" label with the arrow in the UI.

packages/lib/src/toolbar/controls/LineAspectSelector/RadioGroup.tsx
axelboc, line 1: Minor changes to this file are visible in https://github.com/silx-kit/h5web/pull/1892/commits/8b4d7f6727395c79b932e03e336876484fc9d372 — I basically move the label of the group out of the `div` that wraps the options so I can better control the spacing.

Issue #1867, opened 2025-09-11
Toggle display of NeXus auxiliary signals


NAThompson:
Currently, h5web supports the very important usecase of linking multiple timeseries in such a way that they are automatically plotted together. However, if n timeseries are linked as auxiliary signals, then n waveforms are unconditionally plotted.

Feature request: Allow the user to toggle the timeseries display on or off at their convenience.

Use cases:

- Plotting a timeseries and the same timeseries passed through a linear filter. The ability to screenshot one or the other, or both would be very useful.
- Plotting a waveform as well as an associated residual (e.g. from a linear solve). Again turning off the residual would be useful once the sanity check passes.

PR #1894, opened 2025-10-15
Move NX Heatmap signal picker to vis bar

axelboc:
To prepare for #1867, I move the _NX Heatmap_ signal picker to the vis bar, next to the visualization tabs.

I could not move the picker inside the toolbar because the toolbar is inside a `Suspense` boundary, so the picker would unmount and remount every time a signal was picked.

Fortunately, this implementation is much simpler because we still render the picker directly in the visualization container as before, and it's where we manage the selected signal state and where we already have a reference to the `toolbaContainer` element, so we don't need to drill down any props through the mapped vis components.

[Screencast from 2025-10-15 10-29-34.webm](https://github.com/user-attachments/assets/8efd777b-6cc3-487e-97b2-2848adee89c9)

---

For reference, this is what the picker looked like before:

image


packages/app/src/visualizer/Visualizer.module.css
```diff
@@ -9,7 +9,6 @@
 .visBar {
   flex: none;
   display: flex;
-  justify-content: space-between;
```
axelboc, line 12: This was causing the picker to move to the right edge of the vis bar while the toolbar was suspending. The toolbar grows to fill the space so it's not needed anyway.

packages/app/src/__tests__/NxSignalPicker.test.tsx
```diff
@@ -0,0 +1,17 @@
+import { screen } from '@testing-library/react';
```
t20100, line 1: There's a typo in the filename:
`packages/app/src/__tests__/SignalPicke.test.tsx` -> `packages/app/src/__tests__/SignalPicker.test.tsx`

t20100 (approved):
Position of the drop down menu looks good.
When not selected, there looks to be very little visual difference between the NXview buttons and the Signals drop-down: what about adding a separator as in the toolbar? or another visual clue to differentiate the view buttons from the drop-down?

PR #1897, opened 2025-10-16
NXdata can now visualize data given as path and using arithmetic operations in axes & signals

A-D-Fuchs:
## Motivation

We are the developers of [NOMAD CAMELS](https://fau-lap.github.io/NOMAD-CAMELS/). Our software allows you to customize measurements and saves the measured data in HDF5 files (using some of the NeXus classes as a convenience). 
We have the option to display plots for each measurement run and save.
Our measurement data can be nested, so NXdata (`data`) can contain a sub-tree of different datasets. image
We would like the nice visualization to be possible for axes and signals that are not only located directly in NXdata, but below it as well. It seems like a fairly arbitrary limitation due to the fact that HDF5 by design is built around arbitrary nesting and paths to information located somewhere else. 

As our software allows to display arithmetic operations of the recorded data, so something like `data_array**2-1/2` we would love to see this support for the saved data in the HDF5 web viewer as well. 

Here an example NXdata entry:
image

## Changes

Therefore we added code to allow NXdata's `signal`, `axes `and `auxiliary_signal ` to have paths to the actual data. The paths can now also contain arithmetic operations like multiplication(*), division(/), exponents (**), addition & subtraction (+-).

Tests run without errors.

We'd love to hear feedback on this. Maybe you have a much simpler/better suggestion on how we could achieve this?

### DIFF
--- a/packages/app/src/vis-packs/nexus/utils.ts
+++ b/packages/app/src/vis-packs/nexus/utils.ts
@@ -257,11 +257,16 @@ export function findAssociatedDatasets(
     if (name === '.') {
       return undefined;
     }
-  const ent = resolveNxEntity(group, name, entitiesStore);
-  assertDefined(ent, `Expected entity "${name}" to exist`);
-  assertDataset(ent, `Expected "${name}" to be a dataset`);
-  assertArrayShape(ent);
-  return ent;
+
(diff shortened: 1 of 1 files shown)


A-D-Fuchs:
Thanks for the response. We understand that you do not want to diverge from the NeXus spec.

PR #1900, opened 2025-10-21
Upgrade `react-window` and refactor `MatrixVis`

axelboc:
[`react-window` v2](https://github.com/bvaughn/react-window/releases) is a complete rewrite. I was able to remove a lot of complexity, like the `SettingsProvider`.

To simplify things even further, I dropped the ability to turn off the "sticky" behaviour of row/column header cells (the cells with the row/column indices). Now they are always sticky.

I also replaced the `RenderedItemsContext` with a zustand store as it's just simpler to work with and understand.

[Screencast from 2025-10-22 10-17-33.webm](https://github.com/user-attachments/assets/8cad8906-3b73-49e9-a3cf-e787679765d4)


packages/lib/src/vis/matrix/Grid.tsx
```diff
@@ -1,46 +0,0 @@
...
-          className={styles.grid}
-          innerElementType={StickyGrid}
-          width={wrapperSize.width}
-          height={wrapperSize.height}
-          rowHeight={cellSize.height}
-          rowCount={rowCount}
-          columnWidth={cellSize.width}
-          columnCount={columnCount}
-          overscanColumnCount={8}
-          overscanRowCount={16}
```
axelboc, line 36: Subtle change: there's now a single `overscanCount` prop. I've set it to 10 and it seems to work well (cells are rendered sufficiently in advance before they come into view, as shown in the recording).

packages/lib/src/vis/matrix/HeaderCells.tsx
```diff
@@ -1,33 +1,30 @@
...
-      style={{
-        width: width || cellSize.width,
-        height: cellSize.height,
-        transform,
-      }}
+      style={{ width, height, transform }}
       data-bg={index % 2 === 1 ? '' : undefined}
+      {...(headers
+        ? { role: 'columnheader', 'aria-colindex': index + 1 }
+        : { 'aria-hidden': 'true' })}
```
axelboc, line 25: I've added ARIA markup for the compound case (when the columns have headers that are not just indices).

PR #1906, opened 2025-10-23
Upgrade Cypress to v15 and always run tests in Firefox

axelboc:
So my local issue with Firefox was caused by the deprecation of the old CDP protocol that allowed Cypress to communicate with Firefox. The new protocol, [WebDriver BiDi](https://www.cypress.io/blog/announcing-cypress-support-for-firefox-over-webdriver-bidi), became the default in Cypress v14.1.0.

The good news is that after upgrading to Cypress 15 and switching to FF (even in the CI), everything works perfectly! The old issue I had with FF in the CI with flaky screenshots doesn't seem to be happening any more. :tada:

The resolution of a very long standing issue at last.


cypress.config.ts
```diff
@@ -9,7 +9,8 @@ export default defineConfig({
       addMatchImageSnapshotPlugin(on);
     },
   },
-  retries: process.env.CYPRESS_TAKE_SNAPSHOTS ? 3 : null,
+  defaultBrowser: 'firefox',
+  retries: process.env.CYPRESS_TAKE_SNAPSHOTS ? 2 : null,
```
axelboc, line 13: Decreasing the retries to 2 to speed things up, since I haven't seen a flaky test in a while.

PR #1909, opened 2025-10-23
Upgrade Node to v24 as well as all GitHub actions

axelboc:


.github/workflows/deploy-storybook.yml
```diff
@@ -16,16 +16,15 @@ jobs:
...
       - name: Checkout 🏷️
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Set up Node 🕹️
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
           node-version-file: 'package.json'
-          registry-url: 'https://registry.npmjs.org'
```
axelboc, line 25: I don't think this is needed now that I've enabled [trusted publishing](https://docs.npmjs.com/trusted-publishers) to NPM. I'll restore it next time otherwise.

PR #1926, opened 2025-12-15
Bump HSDS support to v1 and test HSDS API

axelboc:
- Create Poetry project at `support/hsds` with [hsds](https://github.com/HDFGroup/hsds) and [h5pyd](https://github.com/HDFGroup/h5pyd)
- Configure HSDS (cf. `support/hsds/config`)
- Add `--hsds` flag to `create_h5_sample.py` to create `sample-hsds.h5` without datasets not supported by `hsload`
- Add `support:hsds` and `support:hsds:load` scripts to respectively start the HSDS support server and load `sample-hsds.h5` into HSDS
- Create `hsds-api.test.ts` file and run test to create Vitest snapshot file
- Update documentation and CI workflow
- Update HSDS provider to support [HSDS v1](https://github.com/HDFGroup/hsds/releases)


.gitignore
```diff
@@ -7,10 +7,13 @@ dist*/
 /cypress/snapshots/**/__diff_output__
 /cypress/snapshots/**/__received_output__
 
+/support/hsds/root/
```
axelboc: `root` is the name of the default bucket,a s configured in `support/hsds/config/override.yml`

package.json
```diff
@@ -30,9 +30,11 @@
     "lint:root:eslint": "eslint --max-warnings=0",
     "lint:root:tsc": "tsc",
     "test": "vitest",
-    "support:setup": "poetry -C support/sample install && poetry -C support/h5grove install",
+    "support:setup": "poetry -C support/sample install && poetry -C support/h5grove install && poetry -C support/hsds install",
     "support:sample": "poetry -C support/sample run python create_h5_sample.py",
     "support:h5grove": "poetry -C support/h5grove run python tornado_app.py --basedir ../sample/dist",
+    "support:hsds": "rm -rf support/hsds/hs.log && poetry -C support/hsds run hsds --root_dir . --config_dir ./config --password_file ./config/passwd.txt",
+    "support:hsds:load": "rm -rf support/hsds/root/* && poetry -C support/hsds run hsload -e http://localhost:5101 -u test -p test --link ../sample/dist/sample-hsds.h5 /",
```
axelboc: - The `admin` user's password is defined in `config/passwd.txt`
- The file is loaded into the root folder of the bucket `/` and `hsload` automatically creates a new "domain" for the file called `/sample-hsds.h5`
- From what I understand, `--link` means that the data stays in the file

packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -38,7 +38,7 @@ export class H5GroveApi extends DataProviderApi {
   public constructor(
     private readonly baseURL: string,
     filepath: string,
-    private readonly fetcher: Fetcher = createBasicFetcher(),
```
axelboc, line 42: Unnecessary type here

packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -60,7 +60,7 @@ export class HsdsApi extends DataProviderApi {
   public constructor(
     private readonly baseURL: string,
     filepath: string,
-    private readonly fetcher: Fetcher,
+    private readonly fetcher = createBasicFetcher(),
```
axelboc, line 64: I've managed to configure the HSDS support server to allow unauthenticated requests, so I'm making the `fetcher` parameter optional so I don't have to pass it in `hsds-api.ts`.

packages/app/src/providers/hsds/__snapshots__/hsds-api.test.ts.snap
axelboc: This is the snapshot file. I've already identified a few issues/inconsistencies with other providers, like a few "Unknown" dtypes, which I'll try to fix.

.gitignore
t20100: maybe use a more explicit name e.g. `test-bucket`

CONTRIBUTING.md
```diff
@@ -412,13 +416,24 @@ pyenv and Poetry:
...
+pyenv exec pnpm support:sample # for h5grove and h5wasm
+pyenv exec pnpm support:sample --hsds # for HSDS
+
+# Start support servers in separate terminals
 pyenv exec pnpm support:h5grove
-pnpm test api
+pyenv exec pnpm support:hsds
+
+# Load sample file into HSDS
+pyenv exec pnpm support:hsds:load
```
t20100: Can't the start server also create the sample files if they don't exists and load it for hsds?

t20100 (approved):
Nice!

CONTRIBUTING.md
axelboc: I like the predictability of the two separate `package.json` scripts and the fact that they each have a single responsibility. For instance, the `load` script can be run multiple times after starting the server in case changes are made to `sample-hsds.h5`.

I would also prefer to keep the two one-liners reasonably short and not have to create a separate bash script.

packages/app/src/providers/hsds/models.ts
```diff
@@ -140,5 +140,5 @@ export interface HsdsCompoundType {
 export interface HsdsEnumType {
   class: 'H5T_ENUM';
   base: HsdsType;
-  mapping: Record<string, number>;
+  members: { name: string; value: number }[];
```
axelboc, line 143: Breaking change for enums in HSDS v1.0. Note that we don't care about the ordering of enum members, so our internal type remains a record.

axelboc:
This is finally ready to go. :tada:  I've updated the HSDS provider to support HSDS v1, so HSDS will need to be upgraded on Bosquet and configured with `max_task_count: 10000` @loichuder

packages/app/src/providers/hsds/__snapshots__/hsds-api.test.ts.snap
```diff
@@ -0,0 +1,1787 @@
...
+    "shape": {
+      "class": "Scalar",
+      "dims": [],
+    },
+    "type": {
+      "charSet": "ASCII",
+      "class": "String",
+      "strPad": "null-terminated",
+    },
+    "value": 0,
```
axelboc, line 42: HSDS v1 no longer seems to support `H5S_NUL` shape — it always returns `H5T_SCALAR` and returns `0` as the value instead of `null`.

axelboc (commented):
I'll do a deeper comparison of the test snapshot with h5grove/h5wasm separately and open relevant issues in the HSDS repo.

loichuder (approved):
Could we somehow reuse the support server you implemented on Bosquet?

PR #1930, opened 2025-12-22
Upgrade dependencies

axelboc:
I'm also removing a couple of `memo` calls on components that don't need them.


packages/lib/src/vis/matrix/Cell.tsx
```diff
@@ -27,4 +26,4 @@ function Cell(props: CellComponentProps<Props>) {
   );
 }
 
-export default memo(Cell);
```
axelboc, line 30: This was actually causing a typing issues with `react-window@2.2.2`

PR #1934, opened 2025-12-22
Increase test coverage of explorer and search

axelboc:
The search wasn't tested at all, and in the explorer, we weren't testing the keyboard navigation.

Both the `search` and `explorer` folders now have 100% test coverage, which adds two percentage points overall. We're now on **62.66%**.

While testing the keyboard navigation in the explorer, I noticed a subtle bug: when pressing the **right arrow key** on an already expanded **empty group**, the focus would move to the sibling of the group instead of remaining on the group. Below is a screen recording of the bug where I press the right arrow key five times in a row:

[Screencast from 2025-12-22 16-08-08.webm](https://github.com/user-attachments/assets/5de8c424-4a02-4083-9b7c-e53be652f00c)

Now the focus remains on the group — here I press the right arrow key 10 times:

[Screencast from 2025-12-22 16-10-22.webm](https://github.com/user-attachments/assets/e51c9555-0a2b-416f-a553-beb5a77a7be5)

### DIFF
--- a/packages/app/src/explorer/EntityItem.tsx
+++ b/packages/app/src/explorer/EntityItem.tsx
@@ -61,30 +61,30 @@ function EntityItem(props: Props) {
   }, [btnRef, isSelected]);
 
   const handleKeyDown = useCallback(
-    (e: KeyboardEvent<HTMLButtonElement>) => {
-      switch (e.key) {
+    (evt: KeyboardEvent<HTMLButtonElement>) => {
+      switch (evt.key) {
         case 'Home':
-          focusFirst(e);
+          focusFirst(evt);
           return;
 
         case 'End':
-          focusLast(e);
+          focusLast(evt);
           return;
 
         case 'ArrowDown':
-          focusNext(e);
+          focusNext(evt);
           return;
 
         case 'ArrowUp':
-          focusPrevious(e);
+          focusPrevious(evt);
           return;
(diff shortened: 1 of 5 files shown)


packages/app/src/explorer/EntityItem.tsx
```diff
@@ -93,10 +93,10 @@ function EntityItem(props: Props) {
             return;
           }
           if (isExpanded) {
-            focusNext(e);
+            focusNext(evt, true);
```
axelboc, line 96: The fix starts here; I pass `true` to `focusNext` to restrict the search of the next button to focus to the children only.

packages/app/src/explorer/utils.ts
```diff
@@ -30,61 +35,66 @@ export function needsNxBadge(
...
 }
 
-export function focusParent(e: KeyboardEvent<HTMLButtonElement>): void {
-  const activeElement = e.currentTarget;
+export function focusParent(evt: KeyboardEvent<HTMLButtonElement>): void {
+  const activeElement = evt.currentTarget;
   const { path } = activeElement.dataset;
-  if (!path) {
-    return;
-  }
```
axelboc, line 43: If the `data-path` attribute is not present on an explorer `button`, it's a bug, so replace with an assertion.

PR #1936, opened 2025-12-23
Improve feedback when copying path and test breadcrumbs bar

axelboc:
- Show confirmation icon for 3s after copying path regardless of pointer (and focus) — previously it was shown until the pointer left the crumb
- Add aria label to last crumb to convey that path can be copied
- Show copy icon when last crumb is focused with the keyboard

[Screencast from 2025-12-23 10-43-24.webm](https://github.com/user-attachments/assets/33293e1d-19f5-4717-9855-4bf08038d9cc)

As mentioned, I also test the breadcrumbs bar a little (but I leave out the fullscreen and feedback buttons, which aren't easy to test in JSDOM, at least for now).

### DIFF
--- a/packages/app/src/__tests__/MetadataViewer.test.tsx
+++ b/packages/app/src/__tests__/MetadataViewer.test.tsx
@@ -3,18 +3,6 @@ import { expect, test } from 'vitest';
 
 import { renderApp } from '../test-utils';
 
-test('switch between "display" and "inspect" modes', async () => {
-  const { user } = await renderApp();
-
-  // Switch to "inspect" mode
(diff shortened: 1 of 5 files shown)


packages/app/src/__tests__/MetadataViewer.test.tsx
```diff
@@ -3,18 +3,6 @@ import { expect, test } from 'vitest';
 
 import { renderApp } from '../test-utils';
 
-test('switch between "display" and "inspect" modes', async () => {
```
axelboc, line 6: Moved to `BreadcrumbsBar.test.tsx`

PR #1940, opened 2026-01-07
Use `PropsWithChildren` consistently

axelboc:
We still had a few `children?: ReactNode` lying around. I'm also adding an empty `NoProps` interface in the shared package as, if you recall, `PropsWithChildren` alone or `PropsWithChildren<{}>` actually allows any props to be passed instead of _only_ the `children` prop.

### DIFF
--- a/apps/storybook/src/Annotation.stories.tsx
+++ b/apps/storybook/src/Annotation.stories.tsx
@@ -151,9 +151,11 @@ export const FollowPointer = {
   },
 } satisfies Story;
 
-function PointerTracker(props: {
+interface PointerTrackerProps {
   children: (x: number, y: number) => ReactNode;
-}) {
+}
+
+function PointerTracker(props: PointerTrackerProps) {
   const { children } = props;
   const [coords, setCoords] = useRafState<[number, number]>();
 
--- a/packages/app/src/vis-packs/VisBoundary.tsx
+++ b/packages/app/src/vis-packs/VisBoundary.tsx
@@ -1,4 +1,4 @@
-import { type ReactNode, Suspense } from 'react';
+import { type PropsWithChildren, Suspense } from 'react';
 import { ErrorBoundary } from 'react-error-boundary';
 
 import ErrorFallback from '../ErrorFallback';
@@ -9,10 +9,9 @@ import ValueLoader from './ValueLoader';
 interface Props {
   resetKey?: unknown;
   isSlice?: boolean;
-  children: ReactNode;
 }
 
-function VisBoundary(props: Props) {
+function VisBoundary(props: PropsWithChildren<Props>) {
   const { resetKey, isSlice, children } = props;
   const { valuesStore } = useDataContext();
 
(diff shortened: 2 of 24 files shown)


apps/storybook/src/Annotation.stories.tsx
```diff
@@ -151,9 +151,11 @@ export const FollowPointer = {
   },
 } satisfies Story;
 
-function PointerTracker(props: {
+interface PointerTrackerProps {
```
axelboc, line 154: Unrelated extraction of a props interface here and in `TooltipMesh.stories.tsx`

packages/app/src/vis-packs/VisBoundary.tsx
```diff
@@ -9,10 +9,9 @@ import ValueLoader from './ValueLoader';
 interface Props {
   resetKey?: unknown;
   isSlice?: boolean;
-  children: ReactNode;
```
axelboc, line 12: Technically `PropsWithChildren` declares the `children` prop as optional, but since `ReactNode` allows `undefined` it doesn't change much...

packages/lib/src/interactions/svg/SvgElement.tsx
```diff
@@ -17,5 +16,4 @@ function SvgElement(props: PropsWithChildren<Props>) {
   return <Html>{createPortal(children, svgOverlay)}</Html>;
 }
 
-export type { Props as SvgElementProps };
```
axelboc, line 20: No point in exporting this... I've also removed it from `index.ts`.

packages/app/src/vis-packs/models.ts
```diff
@@ -1,19 +1,16 @@
...
 import { type IconType } from 'react-icons';
 
 export interface VisContainerProps {
   entity: ProvidedEntity;
   toolbarContainer: HTMLDivElement | undefined;
 }
 
-export interface ConfigProviderProps {
-  children: ReactNode;
-}
```
axelboc, line 12: No point in having a custom type for this — I'm replacing every occurrence with `PropsWithChildren<NoProps>`

axelboc: I thought about having a type alias for `PropsWithChildren<NoProps>` but I like the explicitness of it and the fact that it makes it easier to add props to the component later if needed.

packages/app/src/vis-packs/models.ts
```diff
@@ -1,19 +1,16 @@
...
-
 export interface VisDef {
   name: string;
   Icon: IconType;
   Container: ElementType<VisContainerProps>;
-  ConfigProvider?: ElementType<ConfigProviderProps>;
+  ConfigProvider?: ElementType<NoProps>;
```
loichuder:
```suggestion
  ConfigProvider?: ElementType<PropsWithChildren<NoProps>>;
```

axelboc: :fishing_pole_and_fish:

PR #1941, opened 2026-01-08
Maintain camera zoom when applicable

### LINKED ISSUE #1572: Keep zoom level and axis positions fixed between files

axelboc:
Fix #1572 

Contrary to what I initially suggested in the issue above, I went with a solution that's fully transparent for users. There's no "pin" button; everything is automatic. The solution relies on the axis domains to decide whether to restore the camera scale and position:

- The position and scale of the R3F camera are saved in a zustand store every time the user zooms or pans (with some debouncing), along with a `visContextKey` string computed from the axis domain(s).
- When switching dataset, if the current key and the key saved in the store match, the camera position and scale are restored.

The store can keep the camera states of multiple visualizations at once (cf. `visKey` prop of `KeepZoom`).

[Screencast from 2026-01-14 10-25-19.webm](https://github.com/user-attachments/assets/511297ca-7d12-4907-9dd5-78d8734446d2)

For the Line visualization, the zoom gets saved as long the user zooms only on the `x` axis (cf `xOnly` prop):

[Screencast from 2026-01-14 13-32-56.webm](https://github.com/user-attachments/assets/286fd313-fdcc-4d46-ae1a-f9c880fabc9a)

### DIFF
--- a/packages/app/src/vis-packs/core/complex/MappedComplexHeatmapVis.tsx
+++ b/packages/app/src/vis-packs/core/complex/MappedComplexHeatmapVis.tsx
@@ -1,6 +1,7 @@
 import {
   type DimensionMapping,
   HeatmapVis,
+  KeepZoom,
   useDomain,
   useSafeDomain,
   useSlicedDimsAndMapping,
@@ -115,7 +116,9 @@ function MappedComplexHeatmapVis(props: Props) {
               }
             : undefined
         }
-      />
+      >
+        <KeepZoom visKey="heatmap" />
+      </HeatmapVis>
     </>
   );
 }
--- a/packages/lib/src/vis/shared/RatioEnforcer.tsx
+++ b/packages/lib/src/vis/shared/RatioEnforcer.tsx
@@ -1,5 +1,5 @@
 import { useThree } from '@react-three/fiber';
-import { useEffect } from 'react';
+import { useLayoutEffect } from 'react';
 
 import { useMoveCameraTo } from '../../interactions/hooks';
 import { useVisCanvasContext } from './VisCanvasProvider';
@@ -9,8 +9,8 @@ function RatioEnforcer() {
   const camera = useThree((state) => state.camera);
   const moveCameraTo = useMoveCameraTo();
 
-  useEffect(() => {
-    if (!visRatio || camera.scale.x === camera.scale.y) {
+  useLayoutEffect(() => {
+    if (visRatio === undefined || camera.scale.x === camera.scale.y) {
       return;
     }
 
--- a/packages/lib/src/vis/tiles/store.ts
+++ b/packages/lib/src/vis/tiles/store.ts
@@ -7,8 +7,5 @@ interface TooltipStore {
 
 export const useTooltipStore = create<TooltipStore>((set) => ({
   val: undefined,
-  setTooltipValue: (x: number, y: number, v: number) =>
-    set(() => ({
-      val: { x, y, v },
-    })),
+  setTooltipValue: (x, y, v) => set(() => ({ val: { x, y, v } })),
 }));
--- a/packages/app/src/App.tsx
+++ b/packages/app/src/App.tsx
@@ -1,5 +1,6 @@
-import '@h5web/lib'; // make sure lib styles come first in CSS bundle
+import '@h5web/lib'; // eslint-disable-line import/no-duplicates -- make sure lib styles come first in CSS bundle
 
+import { KeepZoomProvider } from '@h5web/lib'; // eslint-disable-line import/no-duplicates
 import { useToggle } from '@react-hookz/web';
 import { Suspense, useState } from 'react';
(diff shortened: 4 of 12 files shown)


packages/lib/src/vis/shared/RatioEnforcer.tsx
```diff
@@ -9,8 +9,8 @@ function RatioEnforcer() {
   const camera = useThree((state) => state.camera);
   const moveCameraTo = useMoveCameraTo();
 
-  useEffect(() => {
-    if (!visRatio || camera.scale.x === camera.scale.y) {
+  useLayoutEffect(() => {
+    if (visRatio === undefined || camera.scale.x === camera.scale.y) {
```
axelboc, line 13: Unrelated, `visRatio` is a number so it's a bit more strict. I also switch to `useLayoutEffect` so it runs a bit sooner; not sure it really makes a difference.

packages/lib/src/vis/tiles/store.ts
```diff
@@ -7,8 +7,5 @@ interface TooltipStore {
 
 export const useTooltipStore = create<TooltipStore>((set) => ({
   val: undefined,
-  setTooltipValue: (x: number, y: number, v: number) =>
```
axelboc, line 10: Unrelated, removing some unnecessary type annotations.

packages/app/src/vis-packs/core/complex/MappedComplexHeatmapVis.tsx
```diff
@@ -115,7 +117,9 @@ function MappedComplexHeatmapVis(props: Props) {
               }
             : undefined
         }
-      />
+      >
+        <KeepZoom visKey={Vis.Heatmap} />
```
axelboc: Note that I've put the component in the lib, as it's generic enough to be reused in other apps, but I make use of it in the viewer, since it relies on a top-level provider.

loichuder: I am unsure if the key should be `Heatmap` or something else but I don't think it matters much

loichuder (commented):
Reading back #https://github.com/silx-kit/h5web/issues/1572, it may be useful to also add the feature to `Line` as well.

I am also wondering about something. Consider the following sequence:
1. I view `Image1` with Heatmap. I zoom in.
2. I view `Image2` with Heatmap`. It has the same dimensions so the zoom is conserved.
3. I view another unrelated dataset (say a line). The zoom is not conserved since the key is different.
4. I view `Image2` again. Is the zoom reset or because the entry is still in the store, I get a zoomed-in display?

packages/app/src/vis-packs/core/complex/MappedComplexHeatmapVis.tsx
axelboc: Same :grin:

loichuder:
> Regarding adding KeepZoom to the line visualization [...]

I see your point. Perhaps let's not add anything to `Line` for now and wait for user feedback with perhaps some relevant data.

packages/app/src/vis-packs/core/complex/MappedComplexHeatmapVis.tsx
```diff
@@ -115,7 +116,9 @@ function MappedComplexHeatmapVis(props: Props) {
               }
             : undefined
         }
-      />
+      >
+        <KeepZoom visKey="heatmap" />
```
axelboc, line 120: I didn't like importing `Vis` (and `NxDataVis` for `MappedScatterVis`) from the mapped components, so I've switched to simple strings.

Issue #1943, opened 2026-01-13
Support “regular axis” metadata (t0/Δt) to avoid storing full axis vectors in NXdata line plots


NAThompson:
**Problem statement**

In NeXus NXdata, H5Web can auto-plot a line when the group contains a signal dataset and an axis dataset referenced by @axes.
For a uniformly sampled time series, storing a `{times, values}` datastructure is unnecessarily expensive because the times array is large and redundant-`times[i] = t0 + i Δt`.

I’d would like to store a `{t0, Δt, values}` datastructure and have H5Web still auto-plot vs physical time.

**Proposed behavior**

Add support for an axis that is implicitly defined as a regular grid. If an axis dataset is missing, but metadata indicates regular sampling, H5Web should synthesize the axis and plot with the correct metadata.

I imagine the `h5ls -d`looking something like:

```
/entry/data (NXdata)
  values         (float[N])   @signal=1
  t0             (float)      @units="s"
  dt             (float)      @units="s"
  @axes = "time"              (virtual axis name)
```

There is a "checkdown" available: Use a [scaling transform](https://www.nexusformat.org/Scaled_Data.html)  in order to map the integers `[0, 1, 2, 3, ..., N-1]` (which is quite compressible) to `[t0, t0 + Δt, ...]`.

loichuder:
After discussion with @axelboc and @t20100, we feel like the `axes` may not be the limiting factor to visualize the data if there is 100 million of points.

Is it a 1D dataset that you are viewing ? With the Line vis ? In which context (myHDF5, VSCode, JupyterLab) ?

Because from quick tests, even without axes, the visualization feels already a bit laggy with 10**6 points in VSCode...

PR #1944, opened 2026-01-15
Support scaling_factor and offset for axis

loichuder:
For #1943 

POC for the support of `scaling_factor` and `offset`. According to the [NXdata](https://manual.nexusformat.org/classes/base_classes/NXdata.html) spec, these are a bit like `errors`: they are stored as datasets in the `NXdata` group with the name `AXISNAME_scaling_factor` and `AXISNAME_offset`.

So I had to work a bit around this design choice by allowing my `getScalingInfo` function to get data from `valuesStore`. Not sure if I should have made a hook out of it?

Also, scaling could have unintended side effects since the scaled value may have a different type than the value store in the dataset itself (ex: with a scaling factor of `0.1`, integers get converted to floats). I also have to do a bit of dirty conversions in `NxValuesFetcher` to work around this.

Tell me what you think.

### DIFF
--- a/packages/app/src/vis-packs/nexus/hooks.ts
+++ b/packages/app/src/vis-packs/nexus/hooks.ts
@@ -49,8 +49,7 @@ export function useNxData(group: GroupWithChildren): NxData {
       (dataset) =>
         dataset && {
           dataset,
-          ...getDatasetInfo(dataset, attrValuesStore),
-          ...getScalingInfo(group, dataset.name, valuesStore),
+          ...getAxisDef(group, dataset, attrValuesStore, valuesStore),
         },
     ),
     defaultSlice: useDefaultSlice(group, signalDataset.shape, attrValuesStore),
--- a/packages/app/src/vis-packs/nexus/utils.ts
+++ b/packages/app/src/vis-packs/nexus/utils.ts
@@ -4,6 +4,7 @@ import {
   assertArrayShape,
   assertDataset,
   assertDefined,
+  assertNum,
   assertNumericLikeOrComplexType,
   assertNumericType,
   assertScalarShape,
@@ -131,7 +132,7 @@ export function findAuxErrorDataset(
   return dataset;
 }
 
-export function findScalingFactor(
+export function findScalingFactorDataset(
   group: GroupWithChildren,
   datasetName: string,
 ): Dataset<ScalarShape, NumericType> | undefined {
@@ -147,7 +148,7 @@ export function findScalingFactor(
   return dataset;
 }
 
-export function findOffset(
+export function findOffsetDataset(
   group: GroupWithChildren,
   datasetName: string,
 ): Dataset<ScalarShape, NumericType> | undefined {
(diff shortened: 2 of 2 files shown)


packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
```diff
@@ -43,7 +43,22 @@ function NxValuesFetcher<T extends NumericLikeType | ComplexType>(
...
+    const axisScalingFactor = axisDefs[i]?.scalingFactor;
+    const axisOffset = axisDefs[i]?.offset;
+    if (value && (axisScalingFactor || axisOffset)) {
+      const offset = axisOffset ?? 0;
+      const scalingFactor = axisScalingFactor ?? 1;
+      const scaledValue: number[] = [];
+      value.forEach((v) => {
+        scaledValue.push(scalingFactor * Number(v) + offset);
+      });
+      return scaledValue;
```
loichuder: I tried to use `value.map` but to no avail: I got errors saying that `BigInt` could not be converted to number.

I think it has to do with the savage conversions I am doing in this part but I am not versed in TS enough to do better 😓

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -312,6 +345,27 @@ export function getSilxStyle(
...
+  valuesStore: ValuesStore,
+): ScalingInfo {
+  const scalingFactorDataset = findScalingFactor(group, datasetName);
+  const scalingFactor = scalingFactorDataset
+    ? Number(valuesStore.get({ dataset: scalingFactorDataset }))
+    : undefined;
+
+  const offsetDataset = findOffset(group, datasetName);
+  const offset = offsetDataset
+    ? Number(valuesStore.get({ dataset: offsetDataset }))
```
loichuder: The cast as `Number` could probably be replaced by an `assert` instead ?

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
axelboc: Might be best to skip bigint arrays, since it will not work anyway? I'll push something.

packages/app/src/vis-packs/nexus/utils.ts
axelboc: Yes, a couple of assertions would be best I think.

packages/app/src/vis-packs/nexus/utils.ts
```diff
@@ -130,6 +131,38 @@ export function findAuxErrorDataset(
   return dataset;
 }
 
+export function findScalingFactor(
```
axelboc: `findScalingFactorDataset` and `findOffsetDataset`

packages/app/src/vis-packs/nexus/hooks.ts
```diff
@@ -46,7 +47,11 @@ export function useNxData(group: GroupWithChildren): NxData {
     })),
     axisDefs: axisDatasets.map(
       (dataset) =>
-        dataset && { dataset, ...getDatasetInfo(dataset, attrValuesStore) },
+        dataset && {
+          dataset,
+          ...getDatasetInfo(dataset, attrValuesStore),
+          ...getScalingInfo(group, dataset.name, valuesStore),
+        },
```
axelboc, line 53: Maybe time to make a `getAxisDef` function instead of adding `getScalingInfo`?

packages/app/src/vis-packs/nexus/models.ts
```diff
@@ -38,7 +43,7 @@ type WithError<T extends DatasetDef> = T & {
   errorDataset?: Dataset<ArrayShape, NumericType>;
 };
 
-export type AxisDef = DatasetDef<NumericType>;
+export type AxisDef = DatasetDef<NumericType> & ScalingInfo;
```
axelboc, line 46: How about:

```ts
interface AxisDef extends DatasetDef<NumericType> {
  scalingFactor: number | undefined;
  offset: number | undefined;
}
```

axelboc:
I also meant to say: good thinking putting the scaling computation in `NxValuesFetcher`. Definitely the simplest place for it :+1:

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
```diff
@@ -43,7 +48,21 @@ function NxValuesFetcher<T extends NumericLikeType | ComplexType>(
...
+      return value;
+    }
+
+    assertDefined(axisDefs[i]);
+    const { scalingFactor, offset } = axisDefs[i];
+    if (scalingFactor === undefined || offset === undefined) {
+      return value;
+    }
+
+    return value.map((v) => scalingFactor * v + offset);
```
axelboc, line 64: Oh I'm just realising that the [NeXus spec](https://manual.nexusformat.org/classes/base_classes/NXdata.html) says this should be: :thinking: 

```
corrected values = (FIELDNAME + offset) * scaling_factor`
```

loichuder: Oopsie-daisy ! Thanks for checking :joy_cat:

packages/app/src/vis-packs/nexus/models.ts
loichuder: Yep, makes sense if we replace `getScalingInfo` by `getAxisDef` that will return `AxisDef`

PR #1947, opened 2026-01-16
Support unwrapping complex phase values

### LINKED ISSUE #1925: Phase should(?) be unwrapped in complex number visualization
### Describe the bug

If a list of complex numbers is provided to H5Web via an HDF5 file, there is a beautiful interface that allows it to be visualized as a magnitude or a phase. However, the phase is always taken in the range [-π, π]:

Image

For any function of roughly linear phase, this behavior is not quite what you want. Instead, you would like to use (say) [np.unwrap](https://numpy.org/doc/2.3/reference/generated/numpy.unwrap.html) to make the phase decrease monotonically.

### To Reproduce

The following script generates the visualization screenshotted above:

```python
#!/usr/bin/env -S uv run --script
# /// script
# dependencies = ["numpy", "h5py"]
# ///
"""
Generate a small HDF5/NeXus file demonstrating that naive visualization
of complex phase (without np.unwrap) produces discontinuities.

Layout:

/ (root)
  @default = "entry"
  /entry
    @NX_class = "NXentry"
    @default  = "data"
    title
    /data
      @NX_class = "NXdata"
      @signal   = "spectrum"
      @axes     = "frequency"

      frequency       (1D, Hz)
      spectrum        (1D, complex128)
      amplitude       (|spectrum|)
      phase_wrapped   (angle(spectrum) in [-π, π])
      phase_unwrapped (np.unwrap(phase_wrapped))
"""

import numpy as np
import h5py
from pathlib import Path


def write_complex_phase_unwrap_demo(
    filename: str | Path = "phase_unwrap_complex.h5",
    n_points: int = 512,
    f_min: float = -5e3,     # Hz
    f_max: float = 5e3,      # Hz
    delta_t: float = 150e-6, # s, controls phase slope
    f0: float = 2e3,         # Hz, Gaussian width scale
) -> None:
    """
    Create an HDF5 file with a complex NeXus signal for H5Web.

    We define:
        f: frequency axis
        A(f) = exp(-0.5 * (f / f0)^2)
        spectrum(f) = A(f) * exp(-1j * 2π f Δt)

    The true phase is -2π f Δt, i.e. a straight line in f.
    The wrapped phase (np.angle) has 2π jumps, which np.unwrap fixes.
    """
    filename = Path(filename)

    # Frequency axis
    f = np.linspace(f_min, f_max, n_points)  # Hz

    # Amplitude envelope (Gaussian in frequency)
    A = np.exp(-0.5 * (f / f0) ** 2)

    # Linear phase: -2π f Δt (negative slope)
    spectrum = A * np.exp(-1j * 2 * np.pi * f * delta_t)  # complex128

    with h5py.File(filename, "w") as h5:
        # Root: default entry
        h5.attrs["default"] = b"entry"

        # NXentry group
        entry = h5.create_group("entry")
        entry.attrs["NX_class"] = b"NXentry"
        entry.attrs["default"] = b"data"

        entry.create_dataset(
            "title",
            data="Complex spectrum phase unwrap demo",
        )

        # NXdata group
        data = entry.create_group("data")
        data.attrs["NX_class"] = "NXdata"

        # Tell NeXus/H5Web which dataset is the signal and what the axis is
        data.attrs["signal"] = "spectrum"
        data.attrs["axes"] = "frequency"

        # Frequency axis
        d_f = data.create_dataset("frequency", data=f.astype("float64"))
        d_f.attrs["units"] = "Hz"
        d_f.attrs["long_name"] = "Frequency"

        # Complex spectrum (the main point of the demo)
        d_spec = data.create_dataset("spectrum", data=spectrum.astype("complex128"))
        d_spec.attrs["units"] = "a.u."
        d_spec.attrs["long_name"] = "A(f) exp(-i 2π f Δt)"

    print(f"Wrote complex phase unwrap demo to {filename.resolve()}")


if __name__ == "__main__":
    write_complex_phase_unwrap_demo()

```

### Expected behaviour

I believe that _almost always_ an array of complex numbers should have its phase unwrapped before visualization.

Obviously this is a context dependent claim, but personally I can't think of an example where you _wouldn't_ want this behavior, and can think of many examples where you do want it.

### Context

- H5Web in VSCode, version 0.2.1.

### Aside

- This interface is absolutely gorgeous-this is a nitpick.
- Can the signal "long_name" be placed in `| |` for amplitude visualization or prepended with `arg` for phase?

axelboc:
Fix #1925

I add an option called "Phase (unwrapped)" to the complex vis type selector in the toolbar of the complex line visusalisation:

[Screencast from 2026-01-16 13-06-22.webm](https://github.com/user-attachments/assets/4c8fedb5-f14c-4e6f-a7d5-ba5021f6d53d)

In the proposed implementation, when selecting the unwrapped option in the drop-down, the phase/amplitude values are recomputed, which kind of breaks the optimisation I had of precomputing the arrays on initial render. The alternative would be for `usePhaseAmplitudeArrays` to return three arrays (of arrays because of auxiliary signals): `phaseArrays`, `unwrappedPhaseArrays` and `amplitudeArrays`, but this felt messier.

I'm tempted to go back to having no optimisation at all and to just compute the phase/amplitude values when changing  the complex vis type... Not sure...

### DIFF
--- a/packages/app/src/vis-packs/core/complex/MappedComplexLineVis.tsx
+++ b/packages/app/src/vis-packs/core/complex/MappedComplexLineVis.tsx
@@ -69,26 +69,34 @@ function MappedComplexLineVis(props: Props) {
     interpolation,
   } = config;
 
-  const { phaseArrays, amplitudeArrays } = usePhaseAmplitudeArrays([
-    value,
-    ...auxValues,
-  ]);
+  const { phaseArrays, unwrappedPhaseArrays, amplitudeArrays } =
+    usePhaseAmplitudeArrays([value, ...auxValues]);
   const numAxisArrays = useToNumArrays(axisValues);
 
   const mappingArgs = useSlicedDimsAndMapping(dims, dimMapping);
   const mappedPhaseArrays = useMappedArrays(phaseArrays, ...mappingArgs);
+  const mappedUnwrappedPhaseArrays = useMappedArrays(
+    unwrappedPhaseArrays,
+    ...mappingArgs,
+  );
   const mappedAmplitudeArrays = useMappedArrays(
     amplitudeArrays,
     ...mappingArgs,
   );
 
   const phaseDomains = useDomains(mappedPhaseArrays, yScaleType);
+  const unwrappedPhaseDomains = useDomains(
+    mappedUnwrappedPhaseArrays,
+    yScaleType,
+  );
   const amplitudeDomains = useDomains(mappedAmplitudeArrays, yScaleType);
 
   const [pickedArrays, pickedDomains] =
-    complexVisType === ComplexVisType.Phase
-      ? [mappedPhaseArrays, phaseDomains]
-      : [mappedAmplitudeArrays, amplitudeDomains];
+    complexVisType === ComplexVisType.Amplitude
+      ? [mappedAmplitudeArrays, amplitudeDomains]
+      : complexVisType === ComplexVisType.Phase
+        ? [mappedPhaseArrays, phaseDomains]
+        : [mappedUnwrappedPhaseArrays, unwrappedPhaseDomains];
 
   const [dataArray, ...auxArrays] = pickedArrays;
   const [dataDomain, ...auxDomains] = pickedDomains;
--- a/packages/app/src/vis-packs/core/complex/hooks.ts
+++ b/packages/app/src/vis-packs/core/complex/hooks.ts
@@ -1,36 +1,6 @@
 import { createMemo } from '@h5web/shared/createMemo';
-import { isComplexArray } from '@h5web/shared/guards';
-import {
-  type ArrayValue,
-  type ComplexType,
-  type NumericLikeType,
-} from '@h5web/shared/hdf5-models';
-import { type NumArray } from '@h5web/shared/vis-models';
 
-import { toNumArray } from '../utils';
-import { getPhaseAmplitude } from './utils';
+import { getPhaseAmplitude, getPhaseAmplitudeArrays } from './utils';
 
 export const usePhaseAmplitude = createMemo(getPhaseAmplitude);
-
-export function usePhaseAmplitudeArrays(
-  values: ArrayValue<NumericLikeType | ComplexType>[],
-): { phaseArrays: NumArray[]; amplitudeArrays: NumArray[] } {
-  const phaseArrays: NumArray[] = [];
-  const amplitudeArrays: NumArray[] = [];
-
-  values.forEach((arr) => {
-    if (isComplexArray(arr)) {
-      const { phase, amplitude } = getPhaseAmplitude(arr);
-      phaseArrays.push(phase);
-      amplitudeArrays.push(amplitude);
-      return;
-    }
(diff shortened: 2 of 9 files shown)


packages/app/src/vis-packs/core/complex/hooks.ts
```diff
@@ -1,36 +1,6 @@
...
-} from '@h5web/shared/hdf5-models';
-import { type NumArray } from '@h5web/shared/vis-models';
 
-import { toNumArray } from '../utils';
-import { getPhaseAmplitude } from './utils';
+import { getPhaseAmplitude, getPhaseAmplitudeArrays } from './utils';
 
 export const usePhaseAmplitude = createMemo(getPhaseAmplitude);
-
-export function usePhaseAmplitudeArrays(
```
axelboc, line 15: I think this was meant to be memoised — I've moved it to `utils.ts` and just added the `unwrapPhase` parameter.

packages/app/src/vis-packs/core/complex/MappedComplexLineVis.tsx
```diff
@@ -86,9 +86,9 @@ function MappedComplexLineVis(props: Props) {
   const amplitudeDomains = useDomains(mappedAmplitudeArrays, yScaleType);
 
   const [pickedArrays, pickedDomains] =
-    complexVisType === ComplexVisType.Phase
-      ? [mappedPhaseArrays, phaseDomains]
-      : [mappedAmplitudeArrays, amplitudeDomains];
+    complexVisType === ComplexVisType.Amplitude
+      ? [mappedAmplitudeArrays, amplitudeDomains]
+      : [mappedPhaseArrays, phaseDomains];
```
axelboc: We want the phase arrays when either `ComplexVisType.Phase` or `ComplexVisType.PhaseUnwrapped` is selected, so I have to swap the logic.

packages/app/src/vis-packs/core/heatmap/HeatmapToolbar.tsx
```diff
@@ -82,7 +92,7 @@ function HeatmapToolbar(props: Props) {
         <ComplexVisTypeSelector
           value={complexVisType}
           onChange={setComplexVisType}
-          options={Object.values(ComplexVisType)}
```
axelboc, line 85: The heatmap doesn't have the unwrapped option, so I'm decoupling the enum from the options like in the line toolbar. I've also added a `ComplexHeatmapVisType` union type.

packages/shared/src/mock-values.ts
```diff
@@ -30,10 +30,7 @@ const oneD_enum = () => ndarray([0, 2, 2, 1, 1, 0, 2, 2, 1, 1]);
 const oneD_cplx = () =>
   ndarray(
     range9().map((val) =>
-      cplx(
-        val * Math.cos((val * 3.14) / 10),
-        val * Math.sin((val * 3.14) / 10),
```
axelboc, line 35: Removing the divisions to test the unwrapping. See the `oneD_cplx` mock dataset.

PR #2019, opened 2026-06-01
Upgrade pnpm to v11

axelboc:
https://github.com/pnpm/pnpm/releases/tag/v11.0.0

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -22,5 +22,5 @@ runs:
         restore-keys: ${{ runner.os }}-
 
     - name: Install dependencies ⚙️
-      run: pnpm install
+      run: pnpm ci
       shell: bash
--- a/package.json
+++ b/package.json
@@ -11,9 +11,9 @@
   },
   "engines": {
     "node": "24.x",
-    "pnpm": ">=10.26"
+    "pnpm": "11.x"
   },
-  "packageManager": "pnpm@10.33.0",
+  "packageManager": "pnpm@11.5.0",
   "type": "module",
   "scripts": {
     "start": "pnpm --filter demo start",
@@ -49,7 +49,7 @@
     "cypress": "15.13.1",
     "cypress-wait-for-stable-dom": "0.1.0",
     "eslint": "9.39.2",
-    "playwright": "1.59.1",
+    "playwright": "1.60.0",
     "prettier": "3.8.1",
     "typescript": "5.9.3",
     "vitest": "4.1.3",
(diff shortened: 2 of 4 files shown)


.github/actions/install/action.yml
```diff
@@ -22,5 +22,5 @@ runs:
         restore-keys: ${{ runner.os }}-
 
     - name: Install dependencies ⚙️
-      run: pnpm install
+      run: pnpm ci
```
axelboc, line 25: This cleans `node_modules` (noop) and runs `pnpm install --frozen-lockfile` (which was already the case with `pnpm install` when `CI=true` is detected). So no actual change but just stricter.

pnpm-workspace.yaml
```diff
@@ -7,30 +7,25 @@ packages:
...
-managePackageManagerVersions: false
+pmOnFail: ignore
 
 # Manage peer dependencies explicitly and predictibly
 strictPeerDependencies: true
 autoInstallPeers: false
 resolvePeersFromWorkspaceRoot: false
 
-# Fail install if any dependency has unapproved build scripts
-strictDepBuilds: true
```
axelboc, line 18: Now the default

pnpm-workspace.yaml
```diff
@@ -7,30 +7,25 @@ packages:
...
+# Approve specific build scripts
 allowBuilds:
   cypress: true
   esbuild: true
 
-# Save exact dependency versions in `package.json`
-savePrefix: ''
-
-# Only install packages that are at least 2 days old
-minimumReleaseAge: 2880
```
axelboc, line 27: Now it defaults to 1 day, which seems good enough for us. We pin packages and manually check release notes when upgrading.

pnpm-workspace.yaml
```diff
@@ -7,30 +7,25 @@ packages:
...
+# Relax minimum release age protection for owned packages
 minimumReleaseAgeExclude:
   - '@h5web/*'
   - '@esrf/eslint-config'
 
-# Run install automatically if anything is out of date before running a script
-verifyDepsBeforeRun: 'install'
```
axelboc, line 33: Now the default

package.json
```diff
@@ -49,7 +49,7 @@
     "cypress": "15.13.1",
     "cypress-wait-for-stable-dom": "0.1.0",
     "eslint": "9.39.2",
-    "playwright": "1.59.1",
+    "playwright": "1.60.0",
```
axelboc, line 52: Playwright browser installation was stuck in CI — see https://github.com/microsoft/playwright/issues/40998

PR #2024, opened 2026-07-10
Expose createArrayFromView and getAxisOffsets

PeterC-DLS:
1. `@h5web/lib` does not handle views of `ndarray` so can end up using the wrong portion of the underlying 1D typed array and passing it to WebGL. Exposing `createArrayFromView` allows devs to handle this case.
2. The space left when displaying images, etc is a problem when a dev wants to pack in several plots close together in an UI. Exposing `getAxisOffsets` allows them to control the aspect ratio and match the length of the scale axes to the image.

### DIFF
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -112,7 +112,7 @@ export type { VisCanvasContextValue } from './vis/shared/VisCanvasProvider';
 export type { InteractionsContextValue } from './interactions/InteractionsProvider';
 
 // Utilities
-export { toTypedNdArray } from '@h5web/shared/vis-utils';
+export { toTypedNdArray, createArrayFromView } from '@h5web/shared/vis-utils';
 
 export {
   getDomain,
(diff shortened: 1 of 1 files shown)


axelboc:
I've pushed a commit to document `createArrayFromView` and move `getAxisOffsets` to the experimental section: https://github.com/silx-kit/h5web/commit/b35a9348a28be38827e549211d7a78071f094c67

Could you please expand on the second problem you're facing (packing several plots close together)? Maybe we can find a solution to save you from implementing what seems like a complicated workaround.

PeterC-DLS:
Thanks for the quick merge!

Please see https://github.com/DiamondLightSource/davidia/pull/226 for the change we added so page resizes do not add padding to a `RgbVis` - this is demonstrated [here](https://diamondlightsource.github.io/davidia/?path=/story/plots-image--tight-axes).

axelboc:
I see, thanks a lot. This was the initial equal-sizing behaviour right at the start but we ended up removing it to avoid the container measuring step and because we wanted to maximise the space available to the visualization when zooming in.

I considered bringing this sizing behaviour back, but looking at https://github.com/DiamondLightSource/davidia/pull/226, your toolbar complicates things a bit. Let's keep the `getAxisOffsets` export then. I'll document it. **EDIT:** done in 079c7008239724603a7456058e32914d5e729a53

PR #2027, opened 2026-07-29
Add peer dep bounds for react (app) and three (lib)

### LINKED ISSUE #2025: Adding supported/required "react-three/fiber" and "three" versions to peerDependencies in package.json

The current `h5web/lib` version `17.0.0` doesn't working with the newest versions of `react-three/fiber` and `three`.

In particular `react-three/fiber` is on a full major version ahead (currently `@9.6.1`), and it appears `h5web/lib` requires the last full major version behind to work (`@8.18.0`), otherwise weird things happen like getting a deconstructed version of the plot and interactions not interacting as intended (i.e. zooming doesn't actually zoom).

Adding these version dependencies/supported versions to `peerDependencies` in `package.json` would greatly help.

Testing on my local, these are the versions that seem to be supported/work, but it would be good to be reconfirmed:

```yaml
"@h5web/lib": "^17.0.0"
"@react-three/fiber": "8.18.0"
"three": "0.182.0"
```

Alternative would be to just add these version constraints to the readme instructions, like how the `react` version 18 dependency is.

Otherwise, just following the current instructions of `npm install @h5web/lib three @react-three/fiber ndarray` will install the newest unsupported versions of `three` and `@react-three/fiber`.

### Additional context

I tested this with customised stacked `<VisCanvas><HeatmapMesh><ScatterPoints>` and also with just plain `<HeatmapVis>`, and both ended up deconstructed with broken interactions like the below screenshot:

Image

loichuder:
Fix #2025 

`@react-three/fiber` is already set to `8.x`

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -34,8 +34,8 @@
     "@types/react": "18.x",
     "@types/react-dom": "18.x",
     "axios": ">=1",
-    "react": ">=18",
-    "react-dom": ">=18",
+    "react": "18.x",
+    "react-dom": "18.x",
     "typescript": ">=4.5"
   },
   "peerDependenciesMeta": {
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -34,10 +34,10 @@
     "@react-three/fiber": "8.x",
     "@types/react": "18.x",
     "@types/react-dom": "18.x",
-    "@types/three": ">=0.138",
+    "@types/three": ">=0.138 <=0.182",
     "react": "18.x",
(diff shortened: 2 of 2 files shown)


axelboc (approved):

axelboc:
It'd be worth also updating the install commands in the [README](https://www.npmjs.com/package/@h5web/lib) of the lib and the [_Getting Started_](https://h5web-docs.panosc.eu/?path=/docs/getting-started--docs) page in Storybook

loichuder:
I'll do a follow-up PR

PR #2029, opened 2026-08-03
Use fixed width in matrix vis for space-padded fixed-length string arrays

### LINKED ISSUE #2026: [Matrix] Fixed-length string datasets may lead to a lot of blank space in cells
### Describe the bug

When visualizing a dataset with fixed-length strings inside, the cell width grows to accomodate the fixed length. Since this length can be big and that the value is centered in the cell, the cell can appear to show only blank space:

https://github.com/user-attachments/assets/3770bb9c-266b-4979-8199-6107fee18a15

### To Reproduce

```python
import h5py

with h5py.File("fixed_len.h5", "w") as file:
    dset = file.create_dataset(
        "dataset", dtype="|S256", data=["A", "AAAAA", "AAAAAAAAAAAAAAAAAAa"]
    )
```

Then open in H5Web and try to visualize `dataset`

### Expected behaviour

One could say that this is expected and that it is the fixed length that should be better tuned to the data.

However, I feel we could alleviate this very easily by aligning the value to the left instead of centering it.

axelboc:
Fix #2026 

I added a check for space padding when computing the default cell width.

In all of our test files, I couldn't find one with an array of fixed-length strings! :joy:

image


loichuder:
The issue is still there for null-padded strings (see the dataset I provide in https://github.com/silx-kit/h5web/issues/2026)

Should we do the same for these?

axelboc:
Yeah I guess relying on the length at all is not very reliable. I'll just use a default width, then?

loichuder:
> Yeah I guess relying on the length at all is not very reliable. I'll just use a default width, then?

We can compromise by computing the expected length and clamp it to a maximum value so the cell still looks like a cell on most screens (e.g. using the 300px value).

The user can anyway tune the width manually afterwards.

loichuder (approved):

PR #2031, opened 2026-08-04
doc: use correct peer deps versions in Getting Started sections

loichuder:
https://github.com/silx-kit/h5web/pull/2027#issuecomment-5175752200

### DIFF
--- a/packages/lib/README.md
+++ b/packages/lib/README.md
@@ -30,8 +30,16 @@ separate `@types/` package.
 
 ## Getting started 🚀
 
+> [!WARNING]
+>
+> `@h5web/lib` requires the following peer deps:
+>
+> - React v18.*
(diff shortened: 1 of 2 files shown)


loichuder:
Not sure what's up with the e2e test

packages/lib/README.md
```diff
@@ -30,8 +30,16 @@ separate `@types/` package.
 
 ## Getting started 🚀
 
+> [!WARNING]
```
loichuder, line 33: Tried to be fancy here but perhaps it does not display well outside of Github:

Image

axelboc: Nice, I guess we'll see what that looks like on NPM

axelboc (approved):

PR #2032, opened 2026-08-04
Fix zooming by forcing the use of the scale in the Canvas' Camera

### LINKED ISSUE #2028: `three` `0.183.0` breaks the zooming
_Originally posted by @loichuder in [#2025](https://github.com/silx-kit/h5web/issues/2025#issuecomment-5117200665)_

Probably due to https://github.com/mrdoob/three.js/pull/32805

loichuder:
Fix #2028 

As I foresaw, the issue stemmed from https://github.com/mrdoob/three.js/pull/32805 that removes the scale when computing  `matrixWorldInverse` of the `Camera`.

It seems using scales in the matrix (even more so non-uniform scales) is unsupported in some parts of Threejs (https://github.com/mrdoob/three.js/issues/26659#issuecomment-1697145086), leading to problems when computing lightning for example (https://github.com/mrdoob/three.js/issues/26659#issuecomment-1696898864).

I'd like to say this is not relevant for us but the discussion at https://github.com/mrdoob/three.js/issues/26659 worries me a bit: perhaps our use of scale to implement non-uniform zooming is actually more a bug than a feature of three.js. To be discussed with @t20100 perhaps.

Anyway, in this PR, I make a quickfix by creating a new camera that computes `matrixWorldInverse` as three was doing it before v0.183 and it fixes the issue.


packages/app/src/setupTests.ts
```diff
@@ -7,7 +7,12 @@ vi.mock(import('./providers/mock/utils'), { spy: true });
 
 enableBigIntSerialization();
 
-failOnConsole();
+failOnConsole({
```
loichuder, line 10: We have to ignore this deprecation warning since it is only fixed in R3F v10: https://github.com/pmndrs/react-three-fiber/issues/3741

axelboc (approved):
<3

PR #2034, opened 2026-08-10
Upgrade linting config for new import-x plugin

axelboc:
`eslint-plugin-import` is no longer actively maintained and has been holding up the upgrade to ESLint 10 for a few months now. Some of its rules were also painfully slow and had to be disabled. The plugin has been forked and the new plugin, [`eslint-plugin-import-x`](https://github.com/un-ts/eslint-plugin-import-x) has a Rust-based resolver, which means it's a lot faster. See https://gitlab.esrf.fr/ui/eslint-config/-/merge_requests/18

The new plugin caught a few import issues, which is great. :tada:


apps/demo/src/h5wasm/UrlForm.tsx
```diff
@@ -1,9 +1,8 @@
-import { createBasicFetcher } from '@h5web/app';
+import { createBasicFetcher, FetcherError } from '@h5web/app';
 import { type FormEvent, useCallback, useEffect, useState } from 'react';
 import { FiLoader } from 'react-icons/fi';
 import { useSearchParams } from 'wouter';
 
-import { FetcherError } from '../../../../packages/app/src/providers/utils';
```
axelboc, line 6: Cross-package relative import

apps/storybook/src/TiledHeatmapMesh/TiledHeatmapMesh.stories.tsx
```diff
@@ -7,8 +7,7 @@ import {
   TiledTooltipMesh,
   VisCanvas,
 } from '@h5web/lib';
-import { ScaleType } from '@h5web/shared/vis-models';
-import { COLOR_SCALE_TYPES } from '@h5web/shared/vis-utils';
```
axelboc, line 11: I moved `COLOR_SCALES_TYPES` and `AXIS_SCALE_TYPES` from `vis-utils` to `vis-models` to fix a minor import cycle. `vis-models` already contains enums, and `*_SCALE_TYPES` are used to generate types and type guards, so I feel it's reasonable to consider them as "models".

packages/app/src/explorer/EntityItem.tsx
```diff
@@ -12,7 +12,7 @@ import {
 } from 'react';
 import { FiRefreshCw } from 'react-icons/fi';
 
-import EntityList from './EntityList';
+import EntityList from './EntityList'; // eslint-disable-line import-x/no-cycle -- safe
```
axelboc: Recursive rendering, can't do much about it. Import cycles are usually safe anyway — when they're not, the bundler complains.

packages/app/package.json
```diff
@@ -89,6 +89,7 @@
     "vite": "8.1.2",
     "vite-css-modules": "1.16.0",
     "vitest": "4.1.9",
-    "vitest-browser-react": "2.2.0"
+    "vitest-browser-react": "2.2.0",
+    "vitest-fail-on-console": "0.10.1"
```
axelboc, line 93: Moved from root `package.json`

packages/app/rollup.config.js
```diff
@@ -1,8 +1,8 @@
+import { aliasEntries } from '@h5web/shared/rollup.utils.js';
```
axelboc, line 1: I'm now exposing `rollup.utils.js` explicitly from `packages/shared/package.json` to avoid cross-package relative imports.

packages/app/rollup.config.js
```diff
@@ -1,8 +1,8 @@
...
 import alias from '@rollup/plugin-alias';
 import { defineConfig } from 'rollup';
-import dts from 'rollup-plugin-dts';
+import { dts } from 'rollup-plugin-dts';
```
axelboc, line 4: `dts` is both a named export and a default export — the named export is preferred.

packages/lib/src/dimension-mapper/AxisMapper.tsx
```diff
@@ -1,7 +1,7 @@
-import { ToggleGroup } from '@h5web/lib';
```
axelboc, line 1: Importing from the same package, so switching to a relative import

packages/lib/src/interactions/DefaultInteractions.tsx
```diff
@@ -1,14 +1,12 @@
-import Pan, { type PanProps } from '../interactions/Pan';
```
axelboc, line 1: Unnecessary `../interactions` segments

packages/shared/src/guards.ts
```diff
@@ -705,7 +706,7 @@ export function assertLength(
     return;
   }
 
-  const { length: arrLength } = getValues(arr);
```
axelboc, line 708: Last step to remove the import cycle between `guards` and `vis-utils`. Since `getValues` is so simple, I just inline it here.

packages/app/src/explorer/EntityItem.tsx
loichuder: Extend the comment about it to mention it?

packages/shared/src/guards.ts
loichuder: A bit unfortunate but all right.

loichuder (approved):

loichuder:
Good clean-up :+1:

PR #2035, opened 2026-08-14
Plot datasets against their HDF5 dimension scales

NAThompson:
H5Web derives plot axes from NeXus metadata only, so a file that labels its
axes the native HDF5 way -- `make_scale` plus `attach_scale` -- is plotted
against the sample index. A non-uniform abscissa can't be recovered from
indices, so such plots are wrong in shape, not merely unlabelled; today the
only workaround is to duplicate the data into hand-written `NXdata` groups.

Scales are resolved behind a new optional `getDimensionScales` on
`DataProviderApi`, implemented here for h5wasm, which can lean on libhdf5 to
dereference `DIMENSION_LIST`. Providers that leave it undefined keep plotting
against indices, as does any dimension whose scale is missing, non-numeric or
the wrong length -- the spec leaves that mismatch to the application.

Towards silx-kit#1313.

### DIFF
--- a/packages/app/src/__tests__/DimensionScales.browser.test.ts
+++ b/packages/app/src/__tests__/DimensionScales.browser.test.ts
@@ -1,0 +1,65 @@
+import { expect, test } from 'vitest';
+
+import { getAxis, getDimMappingBtn, renderApp } from '../test-utils';
+
+test('plot 1D dataset against its attached dimension scale', async () => {
+  await renderApp('/dimension_scales/oneD');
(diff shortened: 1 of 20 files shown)


NAThompson:
@axelboc : Force pushed a fix to green up the CI-you'll need to re-approve the run.

I believe the e2e might have been victim to drift without my work breaking it?

NAThompson:
@axelboc : Rebased and simplified.

PR #2037, opened 2026-08-14
Cut the main-thread cost of opening and re-rendering a visualisation

NAThompson:
Profiling the h5wasm demo on a 2068x2162 uint16 frame put most of the time between clicking a dataset and seeing pixels in avoidable allocation and shader recompilation. Click-to-figure goes from $211$ ms to $102$ ms.

Per-element allocation: `getBounds` and `getBoundsWithErrors` walked values with `for..of` and `.entries()`, which allocate per element in V8 and JSC -- scanning a 34 MB `Float64Array` triggered 4037 young-generation collections where an indexed loop triggers 4 ($3.25\times$ in V8, $1.25\times$ in SpiderMonkey, which already elides most of it). The complex visualisations allocated their outputs with `Array.from({ length: n })`, whose `PACKED_ELEMENTS` kind boxes every value written in; a `Float64Array` and indexed iteration take $4.19$M values from $572$ ms to $160$ ms. Both are tested against naive reference implementations, since hand-written expectations would not pin down code written this way.

Shader recompilation: `HeatmapMaterial` and `GlyphMaterial` rebuilt their `args` object every render, and R3F compares it element by element, so each re-render replaced the `ShaderMaterial` and relinked its GLSL -- $1.9$ ms in Chrome, $4.7$ ms in Firefox. Both now memoize on what actually appears in the shader source and update the rest in place. `useWarmUpWebGL` moves the first context's $9$ ms of driver setup into idle time.

### DIFF
--- a/packages/app/src/App.tsx
+++ b/packages/app/src/App.tsx
@@ -22,6 +22,7 @@ import { useDataContext } from './providers/DataProvider';
 import Sidebar from './Sidebar';
 import VisConfigProvider from './VisConfigProvider';
 import Visualizer from './visualizer/Visualizer';
+import { useWarmUpWebGL } from './webgl-warm-up';
 
 const SIDEBAR_ID = 'h5w-sidebar';
 const MAIN_AREA_ID = 'h5w-main-area';
@@ -46,6 +47,8 @@ function App(props: Props) {
 
   const [selectedPath, setSelectedPath] = useState<string>(initialPath);
   const [isInspecting, setInspecting] = useState(false);
+
+  useWarmUpWebGL();
 
   const { valuesStore } = useDataContext();
   function onSelectPath(path: string) {
--- a/packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
+++ b/packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
@@ -1,7 +1,8 @@
 import { type Domain, ScaleType } from '@h5web/shared/vis-models';
+import { useThree } from '@react-three/fiber';
 import { rgb, type RGBColor } from 'd3-color';
 import { type NdArray } from 'ndarray';
-import { memo, useMemo } from 'react';
+import { memo, useLayoutEffect, useMemo } from 'react';
 import {
   DataTexture,
   DoubleSide,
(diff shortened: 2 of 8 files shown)


packages/app/src/App.tsx
```diff
@@ -47,6 +48,8 @@ function App(props: Props) {
   const [selectedPath, setSelectedPath] = useState<string>(initialPath);
   const [isInspecting, setInspecting] = useState(false);
 
+  useWarmUpWebGL();
```
axelboc, line 51: I'm not convinced this is worth the added code. It's only a few ms when the first visualization loads.

packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
```diff
@@ -91,32 +92,35 @@ function HeatmapMaterial(props: Props) {
...
-    vertexShader: VERTEX_SHADER,
-    fragmentShader: `
+  const badColorVector = useMemo(() => new Vector4(), []); // updated in place, like the other uniform values
+
+  /* The scale type is the only prop that appears in the shader source, so the
+     material only has to be re-created when it changes. Everything else is a
+     uniform, updated in place below: passing a new object in `args` would make
+     R3F re-instantiate the material, which recompiles and relinks the GLSL
+     program — on every slice of a 3D dataset, for instance. */
+  const shader = useMemo(
```
axelboc, line 106: From the R3F v9 docs at https://r3f.docs.pmnd.rs/api/objects#shader-material-uniforms:

> ShaderMaterial, RawShaderMaterial and subclasses keep the material's [uniforms object](https://threejs.org/docs/#ShaderMaterial.uniforms) stable. When you pass a new uniforms object, React Three Fiber merges the incoming uniforms into the existing target instead of replacing it.
> This preserves three.js internal uniform caching, even if your component creates a fresh uniforms object on each render. Existing uniform entries are updated in place and new entries are added as needed.

So once we upgrade, passing a new `uniforms` object on every render should be perfectly fine and we'll be able to replace the `args={[shader]}` syntax with explicit props (`uniforms`, `fragmentShader`, etc.)

At least that's my hope. I'd prefer to revisit this after the upgrade.

axelboc (commented):
The `for` loop optimisations and unit tests have been integrated in #2054 and #2055.

@loichuder not sure about the other optimisations — please see my comments.

NAThompson:
@axelboc : Thanks!

loichuder:
> @loichuder not sure about the other optimisations — please see my comments.

It seems the majority of the speed-up came from the `for` loops optimizations. The rest can wait the R3F upgrade.

PR #2045, opened 2026-08-28
Visualize float16 datasets

NAThompson:
`isTypedArray` did not recognize `Float16Array`, so half-precision values were rejected outright even though `parseDType` already labelled the datasets float16. The check goes through `globalThis` because naming the constructor directly would throw a `ReferenceError` on older runtimes.

Heatmaps view the halves as uint16 rather than widening them to float32, so they upload as a HALF_FLOAT texture -- already supported here -- with no conversion pass and half the texture memory.


NAThompson:
@axelboc : Had to alias to Uint16 until [this](https://github.com/KhronosGroup/WebGL/issues/3531) is resolve.

Also this PR depends on https://github.com/usnistgov/h5wasm/pull/141, so no merge until then.

NAThompson:
@loichuder , @axelboc : The h5wasm MR has been merged-this is ready for review.

packages/lib/src/vis/heatmap/HeatmapMaterial.tsx
```diff
@@ -20,12 +20,12 @@ import { type ColorMap, type TextureSafeTypedArray } from './models';
 import { getInterpolator, scaleDomain } from './utils';
 
 interface Props extends VisMeshProps {
-  values: NdArray<TextureSafeTypedArray | Uint16Array>; // uint16 values are treated as half floats
+  values: NdArray<TextureSafeTypedArray>;
```
axelboc, line 23: `TextureSafeTypedArray` now includes `Uint16Array`

axelboc:
image

axelboc:
Next steps will be to upgrade h5wasm once released, and to make sure the h5grove and HSDS providers return `Float16Array` (I already have a stash that's almost ready).

PR #2046, opened 2026-09-01
Migrate values store to React Query

axelboc:
This PR starts the migration to [Tanstack Query](https://tanstack.com/query/latest) (also referred to as "React Query").

The goal is to replace our suspense store implementation, which is not fully compatible with React 19. Our current solution also had a lot of limitations in terms of caching, cancelling, etc. React Query already improves on some of these aspects (I'll point out how inline) and opens the door to further improvements.

I start with migrating the values store. It's good to see that our architecture allows migrating this store independently from the others (entities and attributes). The migration was mostly straightforward, as it's still the same suspense-based architecture. Most calls to our `react-suspense-fetch` store are replaced one to one — the main exception is with `useNxValues`, which I'll discuss inline.

The main challenges were that `useSuspenseQuery(ies)` cannot be called conditionally (unlike `valuesStore.get()`), and that suspense-enabled queries cannot be disabled/skipped (it's not possible to pass an `undefined` query, and the `enabled` flag is not allowed unlike with `useQuery(ies)`).

All the tests remain unchanged and still pass, so there's really no noticeable change in behaviour for end users, which is also a big relief.

### DIFF
--- a/apps/demo/src/h5wasm/H5WasmApp.tsx
+++ b/apps/demo/src/h5wasm/H5WasmApp.tsx
@@ -1,5 +1,6 @@
 import { App } from '@h5web/app';
 import { H5WasmBufferProvider, H5WasmLocalFileProvider } from '@h5web/h5wasm';
+import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
 import { useState } from 'react';
 import { useSearchParams } from 'wouter';
 
@@ -7,6 +8,8 @@ import { getFeedbackURL } from '../utils';
 import DropZone from './DropZone';
 import { type RemoteFile } from './models';
 import { getPlugin } from './plugin-utils';
+
+const DEVTOOLS = import.meta.env.VITE_QUERY_DEVTOOLS === 'true';
 
 function H5WasmApp() {
   const [searchParams] = useSearchParams();
@@ -23,10 +26,12 @@ function H5WasmApp() {
       {h5File instanceof File ? (
         <H5WasmLocalFileProvider file={h5File} getPlugin={getPlugin}>
           <App sidebarOpen={!isWide} getFeedbackURL={getFeedbackURL} />
+          {DEVTOOLS && <ReactQueryDevtools />}
         </H5WasmLocalFileProvider>
       ) : (
         <H5WasmBufferProvider {...h5File} getPlugin={getPlugin}>
           <App sidebarOpen={!isWide} getFeedbackURL={getFeedbackURL} />
+          {DEVTOOLS && <ReactQueryDevtools />}
         </H5WasmBufferProvider>
       )}
     </DropZone>
--- a/packages/app/src/ErrorFallback.tsx
+++ b/packages/app/src/ErrorFallback.tsx
@@ -1,4 +1,5 @@
 import { AbortError } from '@h5web/shared/react-suspense-fetch';
+import { CancelledError } from '@tanstack/react-query';
 import { type FallbackProps } from 'react-error-boundary';
 
 import styles from './App.module.css';
@@ -11,7 +12,7 @@ interface Props extends FallbackProps {
 function ErrorFallback(props: Props) {
   const { className = '', error, resetErrorBoundary } = props;
 
-  if (error instanceof AbortError) {
+  if (error instanceof AbortError || error instanceof CancelledError) {
     return (
       <p className={`${styles.error} ${className}`}>
         Request cancelled
--- a/packages/app/src/hooks.ts
+++ b/packages/app/src/hooks.ts
@@ -4,6 +4,7 @@ import {
   assertShape,
   assertType,
   assertValue,
+  isDefined,
 } from '@h5web/shared/guards';
 import {
   type ArrayShape,
@@ -14,6 +15,7 @@ import {
   type ScalarShape,
   type Value,
 } from '@h5web/shared/hdf5-models';
+import { useSuspenseQueries } from '@tanstack/react-query';
 
 import { useDataContext } from './providers/DataProvider';
 import { type ValuesStoreParams } from './providers/models';
@@ -65,17 +67,18 @@ export function useValue<D extends Dataset<ArrayShape | ScalarShape>>(
   dataset: D | undefined,
   selection?: string,
 ): Value<D> | undefined {
-  const { valuesStore } = useDataContext();
+  const { queries } = useDataContext();
+
+  const [result] = useSuspenseQueries({
+    queries: [...(dataset ? [queries.value(dataset, selection)] : [])],
+  });
 
   if (!dataset) {
     return undefined;
   }
 
-  // If `selection` is undefined, the entire dataset will be fetched
-  const value = valuesStore.get({ dataset, selection });
-
-  assertValue(value, dataset);
-  return value;
+  assertValue(result.data, dataset);
+  return result.data;
 }
 
 type ValueFromParams<
@@ -85,35 +88,42 @@ type ValueFromParams<
 export function useValues<
   R extends Record<string, ValuesStoreParams['dataset'] | ValuesStoreParams>,
 >(datasets: R): { [K in keyof R]: ValueFromParams<R[K]> } {
-  const { valuesStore } = useDataContext();
+  const { queries } = useDataContext();
 
-  const storeParams = Object.entries(datasets).map(
-    ([key, datasetOrStoreParams]) =>
-      [
-        key,
-        'dataset' in datasetOrStoreParams
-          ? datasetOrStoreParams // already store params => keep as is
-          : { dataset: datasetOrStoreParams, selection: undefined }, // dataset => prepare store params
-      ] as const,
-  );
+  const keys = Object.keys(datasets);
+  const storeParams = Object.values(datasets).map((datasetOrStoreParams) => {
+    return 'dataset' in datasetOrStoreParams
+      ? datasetOrStoreParams // already store params => keep as is
+      : { dataset: datasetOrStoreParams }; // dataset => prepare store params
+  });
 
-  storeParams.forEach(([, params]) => {
-    valuesStore.prefetch(params);
+  const results = useSuspenseQueries({
+    queries: storeParams.map(({ dataset, selection }) => {
+      return queries.value(dataset, selection);
+    }),
   });
 
   return Object.fromEntries(
-    storeParams.map(([key, params]) => [key, valuesStore.get(params)]),
+    results.map(({ data }, i) => {
+      assertValue(data, storeParams[i].dataset);
+      return [keys[i], data];
+    }),
   ) as { [K in keyof R]: ValueFromParams<R[K]> };
 }
 
 export function useValuesInCache(
   ...datasets: (Dataset<ScalarShape | ArrayShape> | undefined)[]
 ): (dimMapping: DimensionMapping) => boolean {
-  const { valuesStore } = useDataContext();
+  const { queries, queryClient } = useDataContext();
+
+  const definedDatasets = datasets.filter(isDefined);
+
   return (nextMapping) => {
     const selection = getSliceSelection(nextMapping);
-    return datasets.every(
-      (dataset) => !dataset || valuesStore.has({ dataset, selection }),
-    );
+
+    return definedDatasets.every((dataset) => {
+      const { queryKey } = queries.value(dataset, selection);
+      return queryClient.getQueryData(queryKey) !== undefined;
+    });
   };
 }
--- a/packages/app/src/providers/DataProvider.tsx
+++ b/packages/app/src/providers/DataProvider.tsx
@@ -1,25 +1,62 @@
 import { isGroup } from '@h5web/shared/guards';
+import {
+  type ArrayShape,
+  type Dataset,
+  type ScalarShape,
+} from '@h5web/shared/hdf5-models';
 import { getNameFromPath } from '@h5web/shared/hdf5-utils';
 import { createFetchStore } from '@h5web/shared/react-suspense-fetch';
+import {
+  hashKey,
+  QueryClient,
+  QueryClientProvider,
+  queryOptions,
+} from '@tanstack/react-query';
 import {
   createContext,
   type PropsWithChildren,
   useContext,
   useMemo,
+  useState,
 } from 'react';
 
 import { type DataProviderApi } from './api';
+import { type AttrValuesStore, type EntitiesStore } from './models';
 import {
-  type AttrValuesStore,
-  type EntitiesStore,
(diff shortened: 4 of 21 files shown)


apps/demo/src/h5wasm/H5WasmApp.tsx
```diff
@@ -8,6 +9,8 @@ import DropZone from './DropZone';
 import { type RemoteFile } from './models';
 import { getPlugin } from './plugin-utils';
 
+const DEVTOOLS = import.meta.env.VITE_QUERY_DEVTOOLS === 'true';
```
axelboc, line 12: The React Query devtools are a godsend:

Image

packages/app/src/providers/DataProvider.tsx
```diff
@@ -1,25 +1,62 @@
...
-  type AttrValuesStore,
-  type EntitiesStore,
-  type ValuesStore,
-} from './models';
+  createProgressStore,
+  type ProgressStore,
+  trackProgress,
+} from './progress-store';
+
+function getQueryOptionsFactories(
```
axelboc, line 31: For a bit of context, the goal is to end up with this in `DataContext` consumers:

```ts
const { queries } = useDataContext();
const { data } = useSuspenseQuery(queries.value(dataset, selection));
```

This pattern comes out of reading the docs and various resources they link to, notably:
- https://tanstack.com/query/latest/docs/framework/react/guides/query-keys
- https://tanstack.com/query/latest/docs/framework/react/guides/query-options
- https://tanstack.com/query/latest/docs/framework/react/typescript#typing-query-options
- https://tkdodo.eu/blog/effective-react-query-keys
- https://github.com/lukemorales/query-key-factory

---

So this `getQueryOptionsFactories` function creates the `queries` object, which contains "query options" factories. The first such factory is `value()`, which wraps `DataProviderApi#getValue`.

To not lose type inference and keep the linter happy, I had to keep this function in this file and not type its return explicitly — maybe I can find a way to improve this later.

---

The `queryKey` is an array that starts with a `scope`: the `filepath` passed to the provider. This is more of a safeguard since we recommend [unmounting/remounting](https://github.com/silx-kit/h5web/tree/main/packages/app#resetkeys-unknown-optional) the provider when the HDF5 file changes, which would recreate the `QueryClient` and therefore clear the cache completely.

The scope is followed by a namespace (`"value"`) to allow accessing/manipulating all the "value" queries at once (for instance to cancel them).

The rest of the `queryKey` is made of the "dependencies" of the `queryFn`: the dataset's path and the `selection` — these are the same as the keys that were passed to the `react-suspense-fetch` store.

packages/app/src/providers/DataProvider.tsx
```diff
@@ -40,6 +77,23 @@ interface Props {
...
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
+        defaultOptions: {
+          queries: {
+            networkMode: 'always', // always fire queries regardless of online/offline status
+            staleTime: 'static', // never ever refetch cached data
+            gcTime: 15 * 60 * 1000, // keep cached data with no active queries for 15 min
+            retry: false, // never retry fetching automatically on error
+            structuralSharing: false, // not relevant since data is never refetched
```
axelboc, line 89: `networkMode` is crucial since providers can very well work offline (h5wasm obviously, but also h5grove if run locally for instance)

`staleTime`, `gcTime` and `retry` are particularly crucial as well since React Query is _very_ aggressive by default: https://tanstack.com/query/latest/docs/framework/react/guides/important-defaults

The `gcTime` option changes the behaviour that we had previously, which was to keep everything in the cache forever — now the memory is freed 15 min after a query value is last used. This basically comes for free, which is awesome — no need to implement our own LRU cache...

packages/app/src/providers/models.ts
```diff
@@ -17,7 +17,7 @@ export type AttrValuesStore = FetchStore<Entity, AttributeValues>;
 
 export interface ValuesStoreParams {
   dataset: Dataset<ScalarShape | ArrayShape>;
-  selection?: string | undefined;
+  selection?: string | undefined; // if omitted or `undefined`, provider should return the full dataset
```
axelboc, line 20: This comment was previously in `useValue` but I think it makes more sense here.

packages/app/src/providers/progress-store.ts
```diff
@@ -0,0 +1,46 @@
...
+interface ProgressState {
+  ongoing: Map<string, number | undefined>;
+  setProgress: (queryHash: string, value: number | undefined) => void;
+  clearProgress: (queryHash: string) => void;
+}
+
+export type ProgressStore = StoreApi<ProgressState>;
+
+export function createProgressStore(): ProgressStore {
+  return createStore<ProgressState>((set): ProgressState => ({
```
axelboc, line 12: Very similar to the implementation in `react-suspense-fetch.ts`: https://github.com/silx-kit/h5web/blob/main/packages/shared/src/react-suspense-fetch.ts#L42

The difference is that the progress store's keys are now the hashed React Query `queryKey`s. 

We now also have a single progress store for all the queries instead of one progress store per suspense fetch store.

packages/app/src/providers/progress-store.ts
```diff
@@ -0,0 +1,46 @@
...
+        const next = new Map(ongoing);
+        next.delete(queryHash);
+        return { ongoing: next };
+      }),
+  }));
+}
+
+export type OnProgress = (value: number) => void;
+
+export async function trackProgress<TResult>(
```
axelboc, line 32: This is a small abstraction to keep the query options factories readable — it reimplements and decouples the logic that was previously hidden away inside the suspense fetch store: https://github.com/silx-kit/h5web/blob/main/packages/shared/src/react-suspense-fetch.ts#L154-L164

packages/app/src/vis-packs/nexus/hooks.ts
```diff
@@ -62,43 +57,6 @@ export function useNxData(group: GroupWithChildren): NxData {
   };
 }
 
-export function usePrefetchNxValues(
```
axelboc, line 65: `usePrefetchNxValues` and `useNxValues` were used by `NxValuesFetcher` to fetch all the various `NXdata` datasets in parallel. At first I tried refactoring the two hooks with `queryClient.prefetchQuery` and `useSuspenseQueries` but ran into issues with cancelling ongoing requests.

The new implementation in `NxValuesFetcher` relies on a single call to `useSuspenseQueries`, which removes the need for prefetching entirely.

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
axelboc, line 1: This was the most complicated part of the migration.

As mentioned in the PR description, the challenge is that `useSuspenseQueries` doesn't accept `undefined` queries. Constructing an array of defined queries is easy — the difficulty is to then map the array of query results back to the original data structure (especially for auxiliary and axis arrays that can have `undefined` gaps).

The good news is that I found a way to make use of the existing `useValues` hook, which accepts a record of datasets and returns a record of typed dataset values. I made this hook for Hibou, and so far I had only found a single use for it the H5Web repo itself, in `NxNoteFetcher`, so I'm glad to be able to use it here as well.

I started with my own implementation which involved keeping the indices of the queries in an object that matched the `NxData` interface. Claude then had the idea of constructing an object with keys like `auxValues-${i}` — this then made me think of `useValues`. Claude also helped shorten the construction of the `NxValues` object at the end, which was convoluted because of the need to re-assert every value — it came up with the two `optional` and `required` helpers, which cleaned things up nicely.

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
```diff
@@ -20,27 +21,47 @@ function NxValuesFetcher<T extends NumericLikeType | ComplexType>(
...
+  });
+
+  axisDefs.forEach((def, i) => {
+    if (def) {
+      datasets[`axisValues-${i}`] = { dataset: def.dataset };
+    }
+  });
+
+  // 2. Fetch the values
+  const values = useValues(datasets);
```
axelboc, line 48: I haven't found a way to take full advantage of `useValues`'s typing power — `datasets` is not strictly typed, so `values` becomes `Record<string, unknown>` even though the values are asserted internally. So I have to assert the values again below with Claude's `required` and `optional` helpers (which are basically versions of `assertValue` that simply return the value being asserted).

packages/app/src/vis-packs/ValueLoader.tsx
```diff
@@ -40,19 +40,20 @@ function ValueLoader(props: Props) {
             {[...ongoing.entries()]
               .slice(0, MAX_PROGRESS_BARS)
               .map(([key, val]) => (
-                <progress
-                  className={styles.progress}
-                  key={`${key.dataset.path}_${key.selection || ''}`}
-                  value={val}
-                />
+                <progress key={key} className={styles.progress} value={val} />
```
axelboc, line 43: No need to reconstruct a `key` any more here for React, since the progress store now already holds serialised keys (the `queryHash`es).

packages/app/src/visualizer/VisManager.tsx
```diff
@@ -29,10 +29,13 @@ function VisManager(props: Props) {
...
   const [visBarElem, setVisBarElem] = useState<HTMLDivElement>();
 
-  const { valuesStore } = useDataContext();
+  const { filepath, queryClient } = useDataContext();
   function onVisChange(index: number) {
     setActiveVis(index);
-    valuesStore.abortAll('visualization changed', true);
+    void queryClient.cancelQueries(
+      { queryKey: [filepath, 'value'] },
+      { silent: true },
```
axelboc, line 37: Here we don't want to trigger the error boundary or store the `CancelledError` in the query cache. The `silence` option does exactly that: it cancels the query silently.

Note how only the "value" queries are cancelled — this is really powerful stuff.

packages/app/src/ErrorFallback.tsx
```diff
@@ -11,7 +12,7 @@ interface Props extends FallbackProps {
 function ErrorFallback(props: Props) {
   const { className = '', error, resetErrorBoundary } = props;
 
-  if (error instanceof AbortError) {
+  if (error instanceof AbortError || error instanceof CancelledError) {
```
axelboc, line 15: Obviously, I'll need to do a bit of clean up at the end of the migration to remove `react-suspense-fetch` and this `AbortError`. I'll also have to update the documentation.

packages/app/src/hooks.ts
```diff
@@ -65,17 +67,18 @@ export function useValue<D extends Dataset<ArrayShape | ScalarShape>>(
   dataset: D | undefined,
   selection?: string,
 ): Value<D> | undefined {
-  const { valuesStore } = useDataContext();
+  const { queries } = useDataContext();
+
+  const [result] = useSuspenseQueries({
+    queries: [...(dataset ? [queries.value(dataset, selection)] : [])],
```
axelboc, line 73: `useSuspenseQuery` doesn't accept an undefined or disabled query, so to avoid triggering `Suspense` unnecessarily when the input dataset is undefined, I somewhat abuse `useSuspenseQueries` a little...

To be honest, with the refactoring of `NxValuesFetcher`, `useValue` is now only called from `ValueFetcher` with a defined dataset. So maybe the `undefined` case can be removed. I chose to keep it for now in case there's a valid use case in a consumer project I can't remember.

packages/app/src/hooks.ts
```diff
@@ -85,35 +88,42 @@ type ValueFromParams<
...
+      return queries.value(dataset, selection);
+    }),
   });
 
   return Object.fromEntries(
-    storeParams.map(([key, params]) => [key, valuesStore.get(params)]),
+    results.map(({ data }, i) => {
+      assertValue(data, storeParams[i].dataset);
+      return [keys[i], data];
+    }),
```
axelboc, line 110: The input `datasets` record cannot contain `undefined` values, which means we don't have to deal with `undefined` gaps when reconstructing the returned record from the query results — we can just map by index.

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
loichuder: Not a fan of `required`/`optional` wording.

`getAssertedValue`?

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
loichuder: I see.

Would that help to use records all the way instead of lists? Is that even possible?

packages/app/src/hooks.ts
loichuder: It was originally added for the `title` dataset as a convenience thing: https://github.com/silx-kit/h5web/commit/40503ea0f10ee7f69448e47c3ffc85b74d7d5a9b

I think it stayed only for NeXus datasets that may or not be here.

We can probably get rid of it in the next major.

loichuder (approved):
Looks good! Impatient to do how it will turn out with all stores migrated

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
axelboc: There's definitely potential for rethinking the NX data structures through the whole chain, from the containers down to the visualizations. Removing those arrays with `undefined` gaps would simplify things a lot...

loichuder: A battle for another day :wink:

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
axelboc: Done.

I tried again to strengthen the type of the `datasets` record to avoid the double value assertion. I got somewhere but it felt a bit too hacky. I'll revisit this at some point along with https://github.com/silx-kit/h5web/pull/2046#discussion_r3912375810

axelboc:
I'll continue the migration in a PR stack on top of this one and merge everything at the end, so that we don't have outdated documentation on `main`.

packages/app/src/vis-packs/nexus/NxValuesFetcher.tsx
loichuder: I feel like sometimes double assertions are needed as a compromise for too complex typings.

Did you push the changes? I can't see them.

axelboc: Now I have, thanks! :sweat_smile:

loichuder: Good thinking to combine the two functions with a typing override 👍

PR #2049, opened 2026-09-03
Migrate entities and attribute values stores

axelboc:
For the simplest example usage of the new **`queries.entity(path)`**, see `useEntity` and `useDatasets` in `packages/app/src/hooks.ts` (`useEntity` is called in the explorer and the metadata viewer; `useDatasets` is used only in Hibou). Those are already both hooks and they both called `entitiesStore.get()` directly without any conditions, so migrating to `useSuspenseQuery(ies)` was trivial.

For the simplest example usage of the new **`queries.attrValues(entity)`**, see the metadata viewer's `AttributesInfo` component — it called `attrValuesStore.get()` directly and unconditionally, so trivial also.

---

The more challenging cases are when `store.get()` is called **conditionally**, sometimes **deep inside a stack of synchronous functions**...

Let's take `useNxData`, which is used by all the NeXus vis containers:
1. it calls `findSignalDataset`
2. `findSignalDataset` checks if the NXdata group has a "signal" attribute
3. If it does (and only if), then it asks the provider for the value of the "signal" attribute with `attrValuesStore.get(group)["signal"]` (via `getAttributeValue`).

And that's just one of the many attribute values that `useNxData` requests, conditionally or not, but always a few functions deep.

At this point it's good to remind ourselves why we request attribute values conditionally: we want to avoid making unnecessary requests to remote providers like h5grove. So removing the conditions and turning every function into a custom hook so we can call `useSuspenseQuery(queries.attrValues())` deep into the call stack is not an option. For the same reason, we also can't retrieve all the attribute values in `useNxData` and pass them down.

The only option is to call `useSuspenseQuery` from `useNxData`. But then how do we replace the calls to `attrValuesStore.get()` within nested functions like `getAttributeValue`?

Fortunately, Tanstack Query provides a way to invoke a query without `useSuspenseQuery`: it's called [`queryClient.query()`](https://tanstack.com/query/latest/docs/reference/QueryClient#queryclient-query) ([very recently](https://github.com/TanStack/query/pull/10658) renamed from `fetchQuery`).

The problem is that it's async so it means that every function has to be made async (`async` keyword, `Promise` return type, and `await` at call site). This is what most of the diff is about.

The other similar case to `useNxData` is the `resolvePath` function that is called from the `Visualizer` to find the supported visualization for a given entity.

---

Finally, there are the few intermediate cases where it's not quite clear whether a new dedicated query like "nxData" or "resolution" is needed or not, and whether efforts should be put into trying not to trigger the query when possible (with `useSuspenseQueries` and an empty queries array).

---

Hopefully, this gives a good overview. For the remaining explanations, I'll comment inline.


packages/app/src/explorer/NxBadge.tsx
```diff
@@ -10,20 +11,28 @@ interface Props {
...
+        );
+
+        return !!resolution?.supportedVis.some((vis) =>
+          vis.name.startsWith('NX'),
+        );
+      } catch {
+        return false; // no badge if malformed NeXus metadata
+      }
+    },
+  });
```
axelboc, line 32: In `Visualizer`, we call `resolvePath` to find the supported visualisations for the given entity. If you look there first, you'll see that `resolvePath` is now `async` and wrapped in a `"resolution"` query.

Here we use `resolvePath` to determine whether to show the `NX` badge. So I invoke the `"resolution"` query inside the `queryFn` of another `"nxBadge"` query.

Side benefit is that the "nxBadge" query effectively warms up the "resolution" query, so when the user clicks on a group, the resolution is returned instantly from the cache.

---

We could probably do without `Suspense` in this component. I might replace `useSuspenseQuery` with `useQuery` in a future MR, and even drop the "nxBadge" query and call the "resolution" query directly.

packages/app/src/providers/DataProvider.tsx
```diff
@@ -31,9 +30,29 @@ import {
...
+        queryKey: [scope, 'entity', path] as const,
+        queryFn: async () => {
+          const entity = await api.getEntity(path);
+
+          if (isGroup(entity)) {
+            // Cache non-group children (datasets, datatypes and links)
+            entity.children
+              .filter((c) => !isGroup(c))
+              .forEach((c) => {
+                queryClient.setQueryData([scope, 'entity', c.path], c);
```
axelboc, line 48: Pretty straightforward!

packages/app/src/vis-packs/core/heatmap/HeatmapVisContainer.tsx
```diff
@@ -31,7 +30,6 @@ function HeatmapVisContainer(props: VisContainerProps) {
 
   const config = useHeatmapConfig();
   const selection = getSliceSelection(dimMapping);
-  const ignoreValue = useNcIgnoreValue(entity);
```
axelboc, line 34: `useNcIgnoreValue` now calls `useSuspenseQuery`, so by calling the hook here, the whole container gets unmounted/remounted when the "ncIgnoreValue" query for the given entity is not in the cache.

In this container, it's not really a problem, but in the `NxHeatmapVisContainer`, the `entity` passed to `useNcIgnoreValue` changes when the user selects an auxiliary signal from the drop-down menu. And now when that happens, the auxiliary signals drop-down menu itself gets unmounted. This means that the user has to re-open the menu to select another signal, whereas previously, the menu would remain open when switching between signals.

The menu would previously remain open because selected signal's attribute values would always be fetched at this point, always, so the suspense boundary would not get triggered. The difference now is the overarching `"nxIgnoreValue"` query, which always triggers the suspense boundary once.

The fix is to move `useNcIgnoreValue()` inside the mapped components so the suspense boundary that gets triggered is `VisBoundary` (the same that is triggered by the `ValueFetcher` (or `NxValuesFetcher`). It kind of breaks the fetching boundary that we had (nothing fetched in the mapped components and below), but I think that was kind of an artificial boundary anyway.

For consistency I make the same change in `MappedLineVis` and its containers.

packages/app/src/vis-packs/netcdf/hooks.ts
```diff
@@ -12,24 +12,29 @@ import { createIgnoreFillValue, getFillValue, getValidRange } from './utils';
...
 
-  return useMemo(() => {
-    if (!hasNumericType(dataset)) {
-      return undefined;
-    }
+  const { data: nxIgnoreValue } = useSuspenseQuery({
+    queryKey: [dataContext.filepath, 'ncIgnoreValue', dataset.path],
+    queryFn: async () => {
+      if (!hasNumericType(dataset)) {
+        return null;
```
axelboc, line 21: Returning `undefined` inside a `queryFn` is not allowed.

packages/app/src/vis-packs/nexus/hooks.ts
```diff
@@ -22,38 +25,54 @@ import {
...
     },
-    auxDefs: auxSignals.map((auxSignal) => ({
-      dataset: auxSignal,
-      errorDataset: findAuxErrorDataset(group, auxSignal.name),
-      ...getFieldInfo(auxSignal, attrValuesStore),
-    })),
-    axisDefs: axisDatasets.map(
-      (dataset) =>
-        dataset && { dataset, ...getFieldInfo(dataset, attrValuesStore) },
+    auxDefs: await Promise.all(
```
axelboc, line 57: Side benefit here: the definitions are now retrieved in parallel instead of sequentially. We could probably make a lot more things parallel, but there's also readability to take into account 😅

packages/app/src/vis-packs/nexus/hooks.ts
```diff
@@ -22,38 +25,54 @@ import {
...
-const useDefaultSlice = createMemo(getDefaultSlice);
-
 export function useNxData(group: GroupWithChildren): NxData {
-  const { attrValuesStore } = useDataContext();
+  const dataContext = useDataContext();
+
+  return useSuspenseQuery({
+    queryKey: [dataContext.filepath, 'nxData', group.path],
+    queryFn: async () => getNxData(group, dataContext),
+  }).data;
```
axelboc, line 34: It's worth noting that we now trigger the suspense boundary only once, wait for the query's promise to resolve and then re-render. Before we would trigger the boundary on every `store.get()` call and re-render as many times, which was, in hindsight, highly inefficient.

packages/app/src/vis-packs/ValueLoader.tsx
```diff
@@ -51,7 +51,11 @@ function ValueLoader(props: Props) {
               onClick={() => {
                 void queryClient.cancelQueries(
                   { queryKey: [filepath, 'value'] },
-                  { revert: false },
+                  {
+                    /* Make sure the error stays in the query cache so the error boundary appears.
+                     * The boundary's `onError` callback then takes care of evicting errored queries from the cache. */
+                    revert: false,
+                  },
```
axelboc, line 58: Forgot to push this comment in my previous PR.

packages/app/src/visualizer/queries.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import { queryOptions } from '@tanstack/react-query';
+
+import { type DataContextValue } from '../providers/DataProvider';
+import { resolvePath } from './utils';
+
+export function resolvePathQuery(path: string, dataContext: DataContextValue) {
```
axelboc, line 7: I didn't put this query in the `DataProvider` since it's part of the viewer logic.

packages/app/src/vis-packs/core/heatmap/HeatmapVisContainer.tsx
axelboc: I may be able to refactor `useNcIgnoreValue` with `useSuspenseQueries` to avoid the `"nxIgnoreValue"` query. That being said, it might still make sense to move `useNcIgnoreValue` inside the mapped components. Would be interested in your opinion @loichuder.

**EDIT** I've experimented a bit but I'm not getting anywhere.

packages/app/src/hooks.ts
```diff
@@ -127,3 +131,32 @@ export function useValuesInCache(
...
+  entity: Entity,
+  attr: A,
+): Value<A>;
+
+export function useAttrValue<A extends Attribute<ScalarShape | ArrayShape>>(
+  entity: Entity,
+  attr: A | undefined,
+): Value<A> | undefined;
+
+export function useAttrValue<A extends Attribute<ScalarShape | ArrayShape>>(
```
axelboc, line 145: Only used to get the optional `IMAGE_SUBCLASS` attribute in `RgbVisContainer` and `NxRgbContainer`. If `attr` is undefined, the "attrValues" query doesn't run and the suspense boundary isn't triggered.

packages/app/src/providers/DataProvider.tsx
```diff
@@ -47,6 +66,12 @@ function getQueryOptionsFactories(
         },
       });
     },
+    attrValues: (entity: Entity) => {
+      return queryOptions({
+        queryKey: [scope, 'attrValues', entity.path] as const,
+        queryFn: async () => api.getAttrValues(entity),
```
axelboc, line 72: I'm considering adding an "attrValue" query to retrieve the value of a single attribute, and then to call `setQueryData` in the "attrValues" query function to prefill the "attrValue" cache. **EDIT** doesn't add much in the end — it simplifies `getAttributeValue` and `useAttrValue`, but adding an extra layer of caching for each attribute's value is completely redundant.

I'm also considering moving the `assertValue` calls inside the query functions of the "value" and "attrValues" queries. **EDIT**: this is not currently possible for dataset values because of scalar selections.

~~But that's for another day. For now, the goal is to migrate as directly as possible.~~

packages/app/src/explorer/NxBadge.tsx
```diff
@@ -10,20 +11,28 @@ interface Props {
...
-  } catch (error: unknown) {
-    if (error instanceof Error) {
-      return null; // no badge if malformed NeXus metadata
-    }
-
-    throw error; // most likely a promise; rethrow to trigger Suspense boundary
+  const dataContext = useDataContext();
+  const { queryClient } = dataContext;
+
+  const { data: show } = useSuspenseQuery({
```
loichuder, line 17:
```suggestion
  const { data } = useSuspenseQuery({
  ...
  const show = !!data;
  // OR
  const hasData = !!data;
```

packages/app/src/providers/DataProvider.tsx
loichuder: Yeah I think using `getAttrValues` is fine. We may have problems if heavy data is stored in attributes but this is not supposed to happen.

packages/app/src/vis-packs/core/heatmap/HeatmapVisContainer.tsx
loichuder: I know it does not come from this PR but I found this part quite confusing when reading it again: there is `nxIgnoreValue`/`ncIgnoreValue` and each look like a typo of the other.

Could we simply have `useIgnoreValue`?

loichuder: Now for the problem at hand, the fact that behaviour changes like that depending on where we put the hook is kind of a smell. Ideally containers should not worry about that.

Perhaps a refacto is indeed needed. Personally, I'd like to have similar fetching structures for NeXus and NetCDF attributes if possible.

I mean, we can merge like this and rethink it afterwards.

packages/app/src/vis-packs/netcdf/hooks.ts
```diff
@@ -12,24 +12,29 @@ import { createIgnoreFillValue, getFillValue, getValidRange } from './utils';
...
   dataset: Dataset<ArrayShape>,
 ): IgnoreValue | undefined {
-  const { attrValuesStore } = useDataContext();
+  const dataContext = useDataContext();
 
-  return useMemo(() => {
-    if (!hasNumericType(dataset)) {
-      return undefined;
-    }
+  const { data: nxIgnoreValue } = useSuspenseQuery({
```
loichuder:
```suggestion
  const { data: ncIgnoreValue } = useSuspenseQuery({
```

But has I mentioned above, we should probably go with `ignoreValue` (no prefix)

loichuder (approved):
Cool !

We can discuss the `ignoreValue` thingy later if you wish.

packages/app/src/explorer/NxBadge.tsx
axelboc: `data` is already `boolean` — but agreed that it was a bit unclear. I'll merge as is to avoid conflicts, though as this changes again in https://github.com/silx-kit/h5web/pull/2053/changes#diff-988ea6600759bc33cedb54af4c6a10cb8f8b4cf3d0d497ea76a7bcc2d5fb4b6e

packages/app/src/vis-packs/core/heatmap/HeatmapVisContainer.tsx
axelboc: > I know it does not come from this PR but I found this part quite confusing when reading it again: there is nxIgnoreValue/ncIgnoreValue and each look like a typo of the other.

Yeah sorry, nothing to do with NeXus, it was a typo. I've renamed to `useIgnoreValue` :+1:

axelboc: > Now for the problem at hand, the fact that behaviour changes like that depending on where we put the hook is kind of a smell. Ideally containers should not worry about that.

Agreed

PR #2051, opened 2026-09-07
Migrate tiled heatmap

axelboc:
The tiled visualization (`TiledHeatmapMesh`) expects an instance of `TilesApi`. This API has an abstract `get` method that is supposed to return the ndarray for a given tile synchronously — i.e. thanks to `react-suspense-fetch`.

This PR makes the `get` method async and invokes it via `useSuspenseQuery`. It also adds a `TiledHeatmapProvider` so consumers can easily render the required `QueryClientProvider` with the right query client configuration.

We have to add `@tanstack/react-query` to the lib, unfortunately, but since everything is self-contained, we don't have to make it a peer dep. :tada:

I've confirmed that, when the tiled heatmap is not used, `@tanstack/react-query` doesn't appear in the bundle. To test this, I brought the [lib demo](https://codesandbox.io/p/devbox/h5weblib-vite-xru04) from Codesandbox into the repo — I'll open a separate PR for this, though. I've been dragging my feet adding a second demo to the repo, but I think it's time.

### DIFF
--- a/packages/lib/src/vis/tiles/Tile.tsx
+++ b/packages/lib/src/vis/tiles/Tile.tsx
@@ -1,5 +1,6 @@
 import { useThrottledCallback } from '@react-hookz/web';
 import { type ThreeEvent } from '@react-three/fiber';
+import { useSuspenseQuery } from '@tanstack/react-query';
 import { memo } from 'react';
 import { type MagnificationTextureFilter, Vector2 } from 'three';
 
@@ -20,7 +21,11 @@ function Tile(props: Props) {
   const { api, layer, x, y, magFilter, onPointerMove, ...colorMapProps } =
     props;
 
-  const array = api.get(layer, new Vector2(x, y));
+  const { data: array } = useSuspenseQuery({
+    queryKey: ['tile', layer, x, y],
+    queryFn: async () => api.get(layer, new Vector2(x, y)),
+  });
+
   const [height, width] = array.shape;
 
   const handlePointerMove = useThrottledCallback(
--- a/packages/lib/src/vis/tiles/TiledHeatmapProvider.tsx
+++ b/packages/lib/src/vis/tiles/TiledHeatmapProvider.tsx
@@ -1,0 +1,27 @@
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { type PropsWithChildren, useState } from 'react';
+
+function TiledHeatmapProvider(props: PropsWithChildren) {
+  const { children } = props;
+
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
+        defaultOptions: {
+          queries: {
+            networkMode: 'always', // always fire queries regardless of online/offline status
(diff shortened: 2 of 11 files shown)


packages/lib/src/vis/tiles/Tile.tsx
```diff
@@ -20,7 +21,11 @@ function Tile(props: Props) {
   const { api, layer, x, y, magFilter, onPointerMove, ...colorMapProps } =
     props;
 
-  const array = api.get(layer, new Vector2(x, y));
+  const { data: array } = useSuspenseQuery({
+    queryKey: ['tile', layer, x, y],
```
axelboc: The query key doesn't have a scope, so it's important that each `TiledHeatmapMesh` gets its its own `QueryClient` instance and query cache.

This is why I wrap each `TiledHeatmapMesh` in its own `TiledHeatmapProvider` in `TiledHeatmapMesh.stories.tsx‎`.

packages/lib/src/vis/tiles/Tile.tsx
```diff
@@ -20,7 +21,11 @@ function Tile(props: Props) {
...
+    queryFn: async () => api.get(layer, new Vector2(x, y)),
+  });
+
```
axelboc, line 28: Note that the query key doesn't have a scope, so it's important that each `TiledHeatmapMesh` gets its its own `QueryClient` instance and query cache.

This is why I wrap each `TiledHeatmapMesh` in a `TiledHeatmapProvider` in `TiledHeatmapMesh.stories.tsx‎`.

This approach is, I think, simpler to understand. It's also more flexible, as we could imagine adding props to `TiledHeatmapProvider` to allow configuring each query client independently (e.g. the garbage collection time).

packages/lib/src/vis/tiles/TiledHeatmapProvider.tsx
```diff
@@ -0,0 +1,27 @@
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { type PropsWithChildren, useState } from 'react';
+
+function TiledHeatmapProvider(props: PropsWithChildren) {
+  const { children } = props;
+
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
```
loichuder, line 9: If instancating the `QueryClient` becomes more widespread, we could create a subclass that sets the options accordingly

packages/lib/src/vis/tiles/Tile.tsx
loichuder: > This is why I wrap each TiledHeatmapMesh in a TiledHeatmapProvider in TiledHeatmapMesh.stories.tsx‎.

Yep, good call

loichuder (approved):

packages/lib/src/vis/tiles/TiledHeatmapProvider.tsx
axelboc: Agreed, chose not to for now since it's quite the isolated case.

PR #2052, opened 2026-09-07
Remove `react-suspense-fetch` and export/document new React Query API

axelboc:
I've made separate commits if it's easier.


packages/app/src/providers/h5grove/h5grove-api.ts
```diff
@@ -58,11 +59,11 @@ export class H5GroveApi extends DataProviderApi {
   }
 
   public override async getValue(
-    storeParams: ValuesStoreParams,
```
axelboc, line 61: `ValuesStoreParams` no longer makes sense at the level of the provider APIs, so I make the two parameters explicit.

packages/app/src/providers/models.ts
```diff
@@ -1,21 +1,10 @@
...
-  type FetchStore,
-  type OnProgress,
-} from '@h5web/shared/react-suspense-fetch';
-
-export type EntitiesStore = FetchStore<string, ProvidedEntity>;
-export type ValuesStore = FetchStore<ValuesStoreParams, unknown>;
-export type AttrValuesStore = FetchStore<Entity, AttributeValues>;
 
-export interface ValuesStoreParams {
+export interface ValueQueryParams {
```
axelboc, line 7: I keep the interface, because it's still used by `useValues`, but I give it a better name.

packages/app/src/providers/models.ts
```diff
@@ -30,3 +19,5 @@ export interface FetcherOptions {
   abortSignal?: AbortSignal;
   onProgress?: OnProgress;
 }
+
+export type OnProgress = (value: number) => void;
```
axelboc, line 23: Moved and deduplicated from `shared/src/react-suspense-fetch.ts` and `packages/app/src/providers/progress-store.ts`

packages/app/src/providers/utils.ts
```diff
@@ -114,27 +110,19 @@ export function createBasicFetcher(
...
-      if (abortSignal?.aborted && abortSignal.reason instanceof AbortError) {
-        throw abortSignal.reason;
-      }
+    const tracked = trackProgress(response, onProgress);
+    const buffer = await tracked.arrayBuffer();
 
-      throw error;
+    if (tracked.ok) {
+      return buffer;
     }
```
axelboc, line 137: React Query's hooks throw their own `CancelledError`. The underlying error that `fetch` throws when aborted doesn't bubble up. So we don't need to catch it and rethrow our own error — we can just test for `CancelledError` in the error boundary.

Note that I re-export `CancelledError` from `@h5web/app` so consumers can use it in their own error boundaries.

packages/app/src/index.ts
```diff
@@ -38,16 +37,31 @@ export type {
...
+} from '@tanstack/react-query';
+
 // Hooks
-export { useEntity, useDatasets, useValue, useValues } from './hooks';
+export {
+  useEntity,
+  useDatasets,
+  useValue,
+  useValues,
+  useAttrValue,
```
axelboc, line 63: Exporting the new `useAttrValue` hook.

packages/app/src/index.ts
```diff
@@ -38,16 +37,31 @@ export type {
...
+// React Query
+export {
+  useQuery,
+  useQueries,
+  useSuspenseQuery,
+  useSuspenseQueries,
+  queryOptions,
+  hashKey,
+  CancelledError,
+} from '@tanstack/react-query';
```
axelboc, line 55: I really don't want to make `@tanstack/react-query` a peer dep, as I don't want to force consumers of the built-in `App` viewer to install it.

That's why I re-export the React Query hooks, as well as a couple other things that consumers who build their own app may need. We can export more things later if need be.

loichuder: Yeah makes sense. This is a bit unnecessary to make consumers aware of the fetching machinery.

packages/app/README.md
```diff
@@ -510,19 +510,23 @@ function MyApp() {
...
-The three stores are created with the
-[react-suspense-fetch](https://github.com/dai-shi/react-suspense-fetch) library,
-which relies on [React Suspense](https://react.dev/reference/react/Suspense). A
-component that uses one of these stores (e.g.
-`entitiesStore.get('/path/to/entity')`) must have a `Suspense` ancestor to
-manage the loading state.
+You can use [React Query](https://tanstack.com/query/latest) to query data in
+your components, for instance with the
+[`useSuspenseQuery`](https://tanstack.com/query/latest/docs/framework/react/guides/suspense)
+hook:
```
loichuder:
```suggestion
hook (available from `@h5web/app`):
```

I think it is still worth mentioning even if it is written in full below.

packages/app/README.md
```diff
@@ -623,32 +626,88 @@ const hasAttr = hasAttribute(entity, 'my_attr'); // boolean
...
 
 ```ts
-const strAttr = findScalarStrAttr(entity, 'my_str_attr');
 const numAttr = findScalarNumAttr(entity, 'my_num_attr');
+assertDefined(numAttr);
+
+const num = useAttrValue(entity, numAttr); // number | bigint
+```
+
+If you do not assert the attribute object, the type of the value will be loser:
```
loichuder: Loser yourself!

```suggestion
If you do not assert the attribute object, the type of the value will be lost:
```

packages/app/README.md
```diff
@@ -623,32 +626,88 @@ const hasAttr = hasAttribute(entity, 'my_attr'); // boolean
...
+});
+```
+
+If you need to run a query conditionally, the simplest is to use
+`useQuery(ies)`'s
+[`enabled`](https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries)
+option. If you're set on using `useSuspenseQuery(ies)`, it is typically simpler
+to write a custom query and return `null` conditionally from the `queryFn`.
+
+When the conditionnal logic becomes complex, write a custom query:
```
loichuder:
```suggestion
When the conditional logic becomes complex, write a custom query:
```

packages/app/README.md
```diff
@@ -623,32 +626,88 @@ const hasAttr = hasAttribute(entity, 'my_attr'); // boolean
...
+    queries.attrValue(entity, numAttr),
+  ],
+});
+```
+
+If you need to run a query conditionally, the simplest is to use
+`useQuery(ies)`'s
+[`enabled`](https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries)
+option. If you're set on using `useSuspenseQuery(ies)`, it is typically simpler
+to write a custom query and return `null` conditionally from the `queryFn`.
```
loichuder: I feel like this sentence could be linked with the next one.

In both cases, you need to write a custom query and the example demonstrate the conditional `null` return.

packages/app/README.md
```diff
@@ -623,32 +626,88 @@ const hasAttr = hasAttribute(entity, 'my_attr'); // boolean
...
+When the conditionnal logic becomes complex, write a custom query:
+
+```ts
+const { filepath, queryClient, queries } = useDataContext();
+
+const entity = useEntity('/');
+const defaultDatasetAttr = findScalarStringAttr(entity, 'default_dataset');
+
+const { data: defaultDatasetValue } = useSuspenseQuery({
+  queryKey: [filepath, 'defaultDatasetValue'], // always start the `queryKey` with the `filepath`
```
loichuder:
```suggestion
  queryKey: [filepath, 'defaultDatasetValue'], // always start the `queryKey` with the `filepath` to scope the query to the file
```

loichuder (approved):

packages/app/README.md
axelboc: lol, I meant looser — i.e. not as strict

PR #2056, opened 2026-09-11
Add example project with `@h5web/lib`

axelboc:
Bringing the [`h5weblib-vite` codesandbox](https://codesandbox.io/p/devbox/h5weblib-vite-xru04) into the repo so we can more easily test installing and using `@h5web/lib` in a consumer project. I'm adding Sonda, the bundle analyzer, so we can check that unused parts of the lib are correctly tree-shaken.

image

### DIFF
--- a/apps/example-lib/src/styles.css
+++ b/apps/example-lib/src/styles.css
@@ -1,0 +1,23 @@
+@import 'modern-normalize/modern-normalize.css';
+/* In a real-world project, remember to import the lib's styles: */
+/* @import '@h5web/lib/styles.css'; */
+
+body {
+  font-family:
+    -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
+    'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
+}
+
+h2 {
+  margin-bottom: 0;
+  font-size: 1.125rem;
+  font-weight: normal;
+}
+
+#root {
+  display: flex;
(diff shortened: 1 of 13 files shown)


apps/example-lib/src/styles.css
```diff
@@ -0,0 +1,23 @@
+@import 'modern-normalize/modern-normalize.css';
+/* In a real-world project, remember to import the lib's styles: */
+/* @import '@h5web/lib/styles.css'; */
```
axelboc, line 3: Just a reminder that this is not a perfect real-world example since we import the lib's source code directly, not the distributed code.

.github/workflows/lint-test.yml
```diff
@@ -153,3 +153,18 @@ jobs:
           author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com> # https://github.com/peter-evans/create-pull-request/blob/v8/action.yml#L27
           title: 'Update Cypress reference snapshots'
           commit-message: 'Update Cypress reference snapshots'
+
+  example:
+    runs-on: ubuntu-latest
+    # Skip `pull_request` runs on local PRs for which `push` runs are already triggered
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.fork
```
loichuder: Do we really need to run this for every PR? I'd say running on pushes on `main` should be sufficient no?

loichuder (approved):

.github/workflows/lint-test.yml
axelboc: Good call :+1:

PR #2059, opened 2026-09-14
Don't query child groups for NX badges

### LINKED ISSUE #2057: [h5grove] Opening a group fires as many `/meta` requests as there are children
This can be problematic when dealing with groups with ~500+ children.

The reason is architectural and was traced back by @axelboc to the `NXBadge` that needs metadata of the children to check whether the badge should be added or not.

After discussion, there could be a way of alleviating this since the metadata of the parent group already contains the metadata of the children (albeit without the children of the said children).

axelboc:
Fix #2057

Don't query every child group just to decide whether they need an NX badge or not. This reduces the number of requests to remote providers like h5grove _significantly_. For `water_224.h5`:

| Before: 37 requests | After: 16 requests |
|--------|--------|
| image | image | 
| image | image |

This optimisation comes at the cost of a slightly less accurate display of the NX badge as shown in the screenshots above.

#### Cases where a badge is no longer shown

- an `NXdata` group with an "old-style" signal ( = a child dataset with a `signal` attribute instead of a `signal` attribute on the group);
- a group with an implicit NX default child (= a child `NXdata`, `NXentry` or `NXprocess` group) but no actual `default` attribute;
- a group with a default attribute that points to one of the two cases above.

> The silver lining is that this might encourage/help HDF5 file creators to use NeXus better.

#### Cases where a badge is now shown

- an `NXdata` group that isn't actually supported by any of the visualizations (e.g. because it's malformed);
- a group with a default attribute that points to the case above.


packages/app/src/explorer/utils.ts
```diff
@@ -107,3 +112,37 @@ export function focusLast(evt: KeyboardEvent<HTMLButtonElement>): void {
...
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function nxBadgeQuery(group: Group, dataContext: DataContextValue) {
+  return queryOptions({
+    queryKey: [dataContext.filepath, 'nxBadge', group.path] as const,
+    queryFn: async () => needsNxBadge(group, dataContext),
+  });
+}
+
+async function needsNxBadge(
```
axelboc, line 124: For reference, this is what we had before we started showing NX badges more consistently: https://github.com/silx-kit/h5web/commit/45f10a0ebdf169585d767451b9c681745571e7de#diff-77b6d1b517c134033a0435e23202ac1bdef2f1691305dc327f7f197b81517ced

So we don't go back quite as far — the logic remains fairly well aligned with `resolvePath`. Notably, we still query the default entity and recurse to make sure it's an `NXdata` group (or `NXnote`).

packages/app/src/explorer/utils.ts
```diff
@@ -107,3 +112,37 @@ export function focusLast(evt: KeyboardEvent<HTMLButtonElement>): void {
...
+  }
+
+  const defaultEntity = await queryClient.query(queries.entity(nxDefaultPath));
+  if (!isGroup(defaultEntity)) {
+    return false;
+  }
+
+  return dataContext.queryClient.query(
+    nxBadgeQuery(defaultEntity, dataContext),
+  );
```
axelboc, line 147: Note that I invoke a nested query here instead of recursing on `needsNxBadge` directly.

packages/app/src/visualizer/queries.ts
```diff
@@ -1,12 +0,0 @@
-/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
-import { queryOptions } from '@tanstack/react-query';
-
-import { type DataContextValue } from '../providers/DataProvider';
-import { resolvePath } from './utils';
-
-export function resolvePathQuery(path: string, dataContext: DataContextValue) {
```
axelboc, line 7: Moved to `utils.ts` so I can invoke it recursively.

packages/app/src/visualizer/utils.ts
```diff
@@ -79,12 +88,17 @@ export async function resolvePath(
...
+    (await getImplicitNxDefaultPath(entity.children, dataContext));
+
+  if (!nxDefaultPath) {
+    return null;
   }
 
-  return null;
+  return dataContext.queryClient.query(
+    resolvePathQuery(nxDefaultPath, dataContext),
+  );
```
axelboc, line 101: Like in `needsNxBadge`, I now invoke a nested query here instead of recursing on `resolvePath` directly. This avoids running the resolution again when selecting the default group in the explorer.

packages/app/src/visualizer/utils.ts
```diff
@@ -79,12 +88,17 @@ export async function resolvePath(
     }
   }
 
-  const nxDefaultPath = await getNxDefaultPath(entity, dataContext);
-  if (nxDefaultPath) {
-    return resolvePath(nxDefaultPath, dataContext);
+  const nxDefaultPath =
+    (await getNxDefaultPath(entity, dataContext)) ||
+    (await getImplicitNxDefaultPath(entity.children, dataContext));
```
axelboc, line 93: Splitting the implicit default child resolution out of `getNxDefaultPath` so I can reuse the latter for the NX badge.

packages/app/src/explorer/utils.ts
```diff
@@ -107,3 +112,37 @@ export function focusLast(evt: KeyboardEvent<HTMLButtonElement>): void {
...
+async function needsNxBadge(
+  group: Group,
+  dataContext: DataContextValue,
+): Promise<boolean> {
+  const { queryClient, queries } = dataContext;
+
+  const nxClass = await getNxClass(group, dataContext);
+  if (nxClass === 'NXdata' || nxClass === 'NXnote') {
+    return true;
+  }
```
axelboc, line 133: `resolvePath` calls `isNxDataGroup`, which checks for the presence of a `signal` attribute (on the group or one of its children). I could have kept the check for the `signal` attribute on the group, but since a missing `signal` attribute is only one of the many ways one can malform an `NXdata` group, I preferred to be consistent and not check for malformation at all.

packages/app/src/explorer/utils.ts
```diff
@@ -107,3 +112,37 @@ export function focusLast(evt: KeyboardEvent<HTMLButtonElement>): void {
   buttonList[buttonList.length - 1]?.focus();
   evt.preventDefault();
 }
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
loichuder: ?

loichuder (approved):

loichuder:
From 500 to 1 request when dealing with a group with 500 children but no NeXus metadata :tada:

packages/app/src/explorer/utils.ts
axelboc: React Query recommends using automatic type inference for query options: https://tanstack.com/query/latest/docs/framework/react/typescript#typing-query-options — an explicit return type would be quite ugly... something like:

```ts
 OmitKeyof<
  UseQueryOptions<
    boolean,
    Error,
    boolean,
    readonly [string, 'nxBadge', string]
  >,
  'queryFn'
> & {
  queryFn?:
    | QueryFunction<
        boolean,
        readonly [string, 'nxBadge', string]
      >
    | undefined;
} & QueryKeyWithDataTag<
    readonly [string, 'nxBadge', string],
    boolean
  >
```

axelboc: I've added `-- prefer automatic type inference` at the end of the disable comments

loichuder: Thanks :)

PR #2061, opened 2026-09-17
Rewrite HSDS provider

axelboc:
I'm rewriting `hsds-api.ts` from the ground up using the latest API endpoints documented at https://github.com/HDFGroup/hsds/blob/v1.0.1/openapi.yml

## Batching

The screenshots below show the number of requests going from 64 to only 2 requests when opening the `sample-hsds.h5` sample file:

image

image

This is thanks to the following endpoints and features:

- `GET /` with `h5path`, `include_links` and `include_attrs` query params — this combines fetching the domain, then the root group, then the attributes of the group, and then the links of the group into a single request
- `POST `/` with `parent_id` and `include_attrs` query params, and `{ h5paths : string[] }` request body — this allows us to fetch the metadata of all the children of a group in a single request
- `GET /{collection}/{id}/attributes` with `IncludeData` — this allows getting the values of all the attributes of an entity in a single request

## Typed arrays

`GET /datasets/{id}/value` supports the `Accept: application/octet-stream` header to request a binary response instead of JSON. So we replicate the `getValue` logic from `h5grove-api.ts` to return typed arrays when possible.

The screenshots below show that we now get true `Infinty` and `NaN` values instead of strings, and typed arrays for opaque datasets:

image

image

image

---

Opening as draft, as I'd like to keep testing with other HDF5 files, notably soft/external links, NeXux attributes, binary images, etc.


packages/app/src/providers/hsds/HsdsProvider.tsx
```diff
@@ -7,25 +7,25 @@ import { HsdsApi } from './hsds-api';
 
 interface Props {
   url: string;
-  filepath: string;
+  domain: string;
```
axelboc, line 10: Using the HSDS lingo is less confusing I think. The `domain` is still exposed as `filepath` through the `DataContext`, though.

packages/app/src/providers/hsds/HsdsProvider.tsx
```diff
@@ -7,25 +7,25 @@ import { HsdsApi } from './hsds-api';
...
   resetKeys?: unknown[];
-  fetcher: Fetcher;
+  fetcher?: Fetcher;
```
axelboc, line 12: The API works with a default basic fetcher, without authentication. This allows testing with the internal HSDS support server instead of Bosquet.

packages/app/src/providers/utils.ts
```diff
@@ -107,11 +107,12 @@ export function createBasicFetcher(
     params: Record<string, string>,
     opts: FetcherOptions = {},
   ): Promise<ArrayBuffer> => {
-    const { abortSignal, onProgress } = opts;
+    const { abortSignal, onProgress, ...moreFetchOpts } = opts;
     const queryParams = new URLSearchParams(params);
 
     const response = await fetch(`${url}?${queryParams.toString()}`, {
       ...fetchOpts,
+      ...moreFetchOpts,
```
axelboc, line 115: This lets me pass other options like `method: 'POST'` and `headers` with each request.

packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -27,86 +23,83 @@ import {
...
-  convertHsdsShape,
-  convertHsdsType,
-  isHsdsGroup,
+  convertHsdsEntity,
+  convertHsdsSymbolicLink,
   toExtendedJSON,
 } from './utils';
 
 export class HsdsApi extends DataProviderApi {
-  private readonly entities = new Map<string, HsdsEntity<ProvidedEntity>>();
```
axelboc, line 58: `GET /` with the `h5path` query param allows fetching the metadata of an entity at a given path. We don't need to walk the entire tree from the root any more, so we don't need to cache the entities along the way either.

packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -27,86 +23,83 @@ import {
...
 
-    if (path === '/') {
-      const rootId = await this.fetchRootId();
-      const root = await this.processGroup({
-        id: rootId,
-        collection: 'groups',
-        path: '/',
-        name: this.filepath,
-      });
+    if (response.class === 'group') {
```
axelboc, line 66: This block handles groups specifically as we need to fetch the metadata of the children. `GET /?h5path=` with `include_links=true` gives us the links of the group but doesn't resolve them.

packages/app/src/providers/hsds/hsds-api.ts
```diff
@@ -148,14 +172,8 @@ export class HsdsApi extends DataProviderApi {
       return {};
     }
 
-    const attrsPromises = attributes.map(async (attr) =>
-      this.fetchAttributeWithValue(collection, id, attr.name),
-    );
-
-    const attrsWithValues = await Promise.all(attrsPromises);
```
axelboc, line 155: We used to make one value request per attribute!

